PR #8, opened 2020-08-25
Change duzzt dependency

dougwettlaufer:


persistence-api/src/main/java/io/stargate/db/datastore/query/QueryBuilderImpl.java
```diff
@@ -53,7 +53,7 @@
 /**
  * Convenience builder for creating queries.
  */
-@GenerateEmbeddedDSL(reusable = true, modifier = AFModifier.DEFAULT, autoVarArgs = false, name = "QueryBuilder", syntax = "(<keyspace>|<table>|<insert>|<update>|<delete>|<select>|<index>|<type>) consistencyLevel? (future|prepare|execute)", where = {
```
EricBorczuk, line 56: I assume that `reusable` and `displayed` were related somehow to our version of duzzt with graph?

dougwettlaufer: Correct

PR #11, opened 2020-08-28
Update readme

dougwettlaufer:


csplinter:
Can we structure this more like Istio's README? https://github.com/istio/istio

We want to show
1. Introduction: description of the different components with an architecture diagram of the modules
2. Repositories: layout of the github org and a description of the contents of each repo
3. Issue tracking: tell people how we track issues and how they can file them

Most of the content in this README as written is relevant for the contributors / developers guide, which we should link to from this main README

dougwettlaufer:
@csplinter this isn't intended to be the final readme, still working to just get things out here. Plan is to get something more polished out here later. Please feel free to file a ticket with your thoughts.

PR #13, opened 2020-09-02
Adding graphql module

dougwettlaufer:


mpenick:
Conflicts here.

graphqlapi/src/main/java/graphql/kickstart/servlet/CustomGraphQLServlet.java
```diff
@@ -0,0 +1,172 @@
...
+        if (this.isAsyncServletModeEnabled) {
+            AsyncContext asyncContext = request.startAsync(request, response);
+            HttpServletRequest asyncRequest = (HttpServletRequest)asyncContext.getRequest();
+            HttpServletResponse asyncResponse = (HttpServletResponse)asyncContext.getResponse();
+            this.asyncExecutor.execute(() -> {
+                this.doRequest(asyncRequest, asyncResponse, handler, asyncContext);
+            });
+        } else {
+            this.doRequest(request, response, handler, (AsyncContext)null);
+        }
```
olim7t, line 158: I'm not convinced by this async mode. It just schedules the exact same task on another executor. Granted, this frees up the HTTP thread to process another query, but that query will end up scheduling its own task on `asyncExecutor`, so we're not addressing the contention, only moving it one thread pool away.

Async works well when we deal with an async backend, for example if we used `DataStore` directly, we could complete the HTTP response in a callback on the `CompletionStage` returned by `query()`.

And in fact it looks like GraphQL supports that. From [the docs](https://www.graphql-java.com/documentation/v11/execution/):

> By default the “query” execution strategy is graphql.execution.AsyncExecutionStrategy which will dispatch each field as CompletableFuture objects and not care which ones complete first. This strategy allows for the most performant execution.
> 
> **The data fetchers invoked can themselves return CompletionStage values and this will create fully asynchronous behaviour.**

If I understand that correctly, if our data fetchers return futures, all we'd have to do in `doGet` and `doPost` is invoke `handler.handle(request, response)` and let it do its magic.

graphqlapi/src/main/java/io/stargate/graphql/core/DataFetchers.java
```diff
@@ -0,0 +1,452 @@
...
+            DataStore dataStore = persistence.newDataStore(queryState, null);
+
+            String statement = buildStatement(table, environment, dataStore);
+            CompletableFuture<ResultSet> rs = dataStore.query(statement);
+            ResultSet resultSet = rs.get();
+
+            ImmutableMap.Builder resultMap = ImmutableMap.builder();
+            resultMap.put("value", environment.getArgument("value"));
+
+            return resultMap.build();
```
olim7t, line 83: Per my other comment on making data fetchers async, this is how we would transform this:
```suggestion
            return dataStore.query(statement)
                .thenApply(resultSet -> ImmutableMap.of("value", environment.getArgument("value")));
```

olim7t (changes requested):
First round of comments. I think it would be a big improvement to take advantage of GraphQL's async capabilities, I'll give it a try tomorrow.

graphqlapi/src/main/java/graphql/kickstart/servlet/CustomGraphQLServlet.java
olim7t: Upon further investigation, GraphQL supports async but graphql-java-servlet doesn't 😢 . It blocks on the query result in [GraphQLInvoker](https://github.com/graphql-java-kickstart/graphql-java-servlet/blob/master/graphql-java-kickstart/src/main/java/graphql/kickstart/execution/GraphQLInvoker.java#L38).

There's even a pending feature request for that: graphql-java-kickstart/graphql-java-servlet#259.

I'll check if I can contribute a PR. In the meantime, I suggest we still remove `asyncExecutor`, because as I was explaining previously I don't think it really improves performance.

graphqlapi/pom.xml
```diff
@@ -0,0 +1,160 @@
...
+  <artifactId>graphqlapi</artifactId>
+  <properties>
+    <osgi.bundle.version>${project.version}</osgi.bundle.version>
+    <jetty.version>9.4.30.v20200611</jetty.version>
+  </properties>
+  <dependencies>
+    <dependency>
+      <groupId>io.stargate.db</groupId>
+      <artifactId>persistence-api</artifactId>
+      <version>0.0.5-SNAPSHOT</version>
```
tub, line 20: This module is depending on an older snapshot of `persistence-api` and `authentication`
```suggestion
      <version>${project.version}</version>
```

olim7t: Good catch, addressing this as I'm merging the latest changes from the base branch.

graphqlapi/src/main/java/graphql/kickstart/servlet/CustomGraphQLServlet.java
```diff
@@ -0,0 +1,172 @@
+package graphql.kickstart.servlet;
```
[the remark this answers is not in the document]
dougwettlaufer: Sure. I don't believe there's any significance having it here

olim7t (approved):
Approving to keep things moving.

There's still a lot to improve in terms of organizing the code, but this can be addressed in follow-up tickets.

olim7t:
Sonar still reports bugs but I think they are false positives. E.g. in `AlterTableFetcher`, it's not smart enough to figure out that `table` can't be null if the list is not empty (IntelliJ figures it out, and warns me if I add `assert table != null`).

olim7t (approved):

graphqlapi/src/main/java/graphql/kickstart/servlet/CustomGraphQLServlet.java
```diff
@@ -0,0 +1,303 @@
...
+
+  protected void doPost(HttpServletRequest request, HttpServletResponse response) {
+    handle(request, response);
+  }
+
+  private void handle(HttpServletRequest request, HttpServletResponse response) {
+    String keyspaceName = getKeyspaceName(request);
+    if (keyspaceName == null) {
+      failOnNoDefaultKeyspace(response);
+    } else if (!KEYSPACE_NAME_PATTERN.matcher(keyspaceName).matches()) {
```
dougwettlaufer, line 66: Shouldn't we check this farther up?

graphqlapi/src/main/java/io/stargate/graphql/core/GqlKeyspaceSchema.java
```diff
@@ -0,0 +1,581 @@
...
+                    .name("options")
+                    .type(new GraphQLTypeReference("QueryOptions")))
+            .type(buildEntityResultOutput(table))
+            .dataFetcher(fetcherFactory.new QueryDataFetcher(table))
+            .build();
+
+    GraphQLFieldDefinition filterQuery =
+        GraphQLFieldDefinition.newFieldDefinition()
+            .name(CaseUtil.toLowerCamel(table.name()) + "Filter")
+            .deprecate("No longer supported. Use root type instead.")
```
jorgebay, line 196: What's the origin of this deprecation?

PR #22, opened 2020-09-10
Create CONTRIBUTING.md

dougwettlaufer:


EricBorczuk (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,58 @@
...
+## <a name="feature"></a> Want a Feature?
+You can *request* a new feature by [submitting an issue](#submit-issue) to the GitHub
+Repository. If you would like to *implement* a new feature, please submit an issue with
+a proposal for your work first, to be sure that we can use it.
+
+* **Small Features** can be crafted and directly [submitted as a Pull Request](#submit-pr).
+
+## <a name="submit"></a> Contribution Guidelines
+
+### <a name="submit-issue"></a> Submitting an Issue
```
dimas-b, line 22: Why not `<a name="submit-issue"/>`?.. would be simpler and easier to read since it's just an anchor.

dimas-b: Also, AFAIK, markdown support auto-generating anchors for section header... at least IDEA has a auto-suggestion for those... I think they are more readable (something like `#submitting-an-Issue) in this case. What is the advantage of explicit `<a/>` anchors?

dougwettlaufer: @csplinter ^

dougwettlaufer: Really they both render appropriately in github which is what really matters.

csplinter: I don't have a preference

csplinter (approved):

PR #24, opened 2020-09-10
Add a settings.xml when deploying to GH packages

dougwettlaufer:


.github/workflows/release.yml
```diff
@@ -72,6 +72,22 @@ jobs:
       - uses: actions/setup-java@v1.4.2
         with:
           java-version: 1.8
+      - name: Setup Github Package Registry
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          mkdir -p ~/.m2
+          cat <<EOF > ~/.m2/settings.xml
+          <settings>
```
EricBorczuk, line 81: Is there any issue of escaping characters here?

EricBorczuk (approved):

.github/workflows/release.yml
dougwettlaufer: Not sure what you mean. This is just creating a file.

EricBorczuk: like when I just copy your `cat <<EOF ...` statement to my terminal, it doesn't quite work, needs some kind of escaping or close quote or smth

PR #30, opened 2020-09-11
Add logback.xml

dougwettlaufer:


stargate-lib/logback.xml
```diff
@@ -0,0 +1,15 @@
...
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%-5level [%thread] %date{ISO8601} %F:%L - %msg%n</pattern>
+        </encoder>
+    </appender>
+    <root level="${stargate.logging.level.root:-INFO}">
+        <appender-ref ref="STDOUT" />
+    </root>
+    <logger name="org.apache.cassandra" level="${stargate.logging.level.cassandra:-INFO}"/>
+    <logger name="com.datastax.bdp.db" level="${stargate.logging.level.dse:-INFO}"/>
```
dimas-b, line 12: Is `com.datastax.bdp.db` also OSS under Stargate?

dougwettlaufer: Yep. See stargate/persistence-extensions.

mpenick (approved):

stargate-lib/logback.xml
dimas-b: Ah, I see. thx! Still, since it's an "extension" it is necessary to configure its logging in the default "generic" config?

dougwettlaufer: Definitely room for improvement with logging. As we get more sophisticated we can look into adding different files but for now I really just want to cut down on the spam :).

PR #32, opened 2020-09-11
Support developer mode

jorgebay:
When developer mode is set, the stargate node joins the ring and defaults the seed list to itself.


jorgebay:
You can use the following command to start in developer mode:

```
./starctl --cluster-name stargate --cluster-version 3.11 --listen 127.0.0.1 --simple-snitch --developer-mode
```

jorgebay:
I've simplified it a little further by setting simple snitch in developer mode, the new command w/ the required parameters is:

```
./starctl --cluster-name stargate --cluster-version 3.11 --listen 127.0.0.1 --developer-mode
```

jorgebay:
I've rolled in a change here to make binding to listen address opt-in.

stargate-starter/src/main/java/io/stargate/starter/Starter.java
```diff
@@ -125,6 +123,16 @@
...
     private boolean emulateDbaasDefaults = false;
 
+    @Order(value = 14)
+    @Option(name = {"--developer-mode"}, description = "Defines whether the stargate node should also behave as a " +
+        "regular node, joining the ring with tokens assigned in order to facilitate getting started quickly and not " +
+        "requiring additional nodes or existing cluster")
+    boolean developerMode = false;
+
+    @Order(value = 15)
+    @Option(name = {"--bind-to-listen-address"}, description = "When set, it binds web services to listen address only")
```
dougwettlaufer, line 133: Please add these to starctl as well since we also support env vars.

dougwettlaufer (commented):
Looks good, just needs the script update.

stargate-starter/src/main/java/io/stargate/starter/Starter.java
jorgebay: Good catch, I've added them.

dougwettlaufer (approved):

mpenick:
@jorgebay These changes need to also be made to the other persistence backends: https://github.com/stargate/persistence-extensions

jorgebay:
There was a [small discussion](https://github.com/riptano/c2/pull/173#issuecomment-691874030) on the private repo, tldr: only support it in the built-in C* 3.11 for now.

I think it it's valuable for users to have a developer mode that doesn't require an additional cluster. Having 3.11 support for dev mode is good enough for me, but if you think that we need it for dse and C* 4.0 we should file a ticket.

PR #35, opened 2020-09-11
Rev version 0.0.6-SNAPSHOT

dougwettlaufer:
Still working on getting the GH action to do this for us.


mpenick (dismissed):

dimas-b (dismissed):

auth-api/pom.xml
```diff
@@ -5,7 +5,7 @@
     <parent>
         <groupId>io.stargate</groupId>
         <artifactId>stargate</artifactId>
-        <version>0.0.5-SNAPSHOT</version>
+        <version>0.0.6-SNAPSHOT</version>
```
dimas-b, line 8: Sorry, @dougwettlaufer , could you give a TL;DL on why the previous approach of using `${revision}` for version properties did not work? thx!

dimas-b (approved):

auth-api/pom.xml
dougwettlaufer: Using the property for building/running worked because maven substituted the value on the fly but when it came time to publish to GH packages instead of replacing `${revision}` with `0.0.1` it was instead keeping the variable and using that in the deployed pom. So whenever you pulled in the dependency into another project it broke saying `${revision} doesn't exist`.

PR #38, opened 2020-09-12
Update rest and auth apis

dougwettlaufer:


restapi/src/main/java/io/stargate/web/resources/v2/schemas/ColumnsResource.java
```diff
@@ -34,7 +34,7 @@
 import io.stargate.web.resources.Db;
 import io.stargate.web.resources.RequestHandler;
 
-@Path("/schemas/keyspaces/{keyspaceName}/tables/{tableName}/columns")
+@Path("/v2/schemas/keyspaces/{keyspaceName}/tables/{tableName}/columns")
```
tomekl007, line 37: if we are changing the api in a non-backward compatible way, shouldn't we also keep the previous version? (To have old `"v1"` and new `v2`)

dougwettlaufer: This is to correct a misunderstanding with the API spec. The correct path for these new APIs is `/v2`.

EricBorczuk (approved):

PR #39, opened 2020-09-13
Add health-checker module and centralize metrics

olim7t:
Demo: launch Stargate, then browse to http://127.0.0.2:8085/metrics
Metrics are prefixed by the name of their module, e.g. `restapi_*`, `cql_*`, etc. There is also a common set of `jvm_*` metrics.


health-checker/src/main/java/io/stargate/health/metrics/impl/PrefixingMetricRegistry.java
```diff
@@ -0,0 +1,349 @@
...
+/**
+ * Wraps an existing registry to prefix every new metric with a given string on the fly.
+ *
+ * <p>For clients of this class, it looks like they're interacting with a regular registry, when
+ * they are in fact only seeing one part of a bigger registry, that has other metrics under
+ * different prefixes.
+ *
+ * <p>Implementation note: unfortunately {@link MetricRegistry} is a class, not an interface. So we
+ */
+public class PrefixingMetricRegistry extends MetricRegistry {
```
olim7t, line 32: I'm not overjoyed with this, but it's a convenient way to quickly integrate with modules that already work with a metric registry. Adding the prefix in the client code requires significantly more work, DropWizard in particular does not offer a built-in way to do it.

If we don't scope the metrics, the different modules collide and an error is thrown.

health-checker/src/main/java/io/stargate/health/Server.java
```diff
@@ -0,0 +1,43 @@
...
+            }
+        });
+        environment.jersey().register(CheckerResource.class);
+    }
+
+    @Override
+    public void initialize(final Bootstrap<ApplicationConfiguration> bootstrap) {
+        super.initialize(bootstrap);
+        bootstrap.setConfigurationSourceProvider(new ResourceConfigurationSourceProvider());
+        bootstrap.setMetricRegistry(metrics.getRegistry());
```
olim7t, line 41: Currently health-checker doesn't prefix its metrics, because what we pass here is both the registry where the module will publish its own metrics, and the one that is used to expose the `/metrics` endpoint (where we want every metric).

I'll probably need to switch to a custom servlet to switch to Prometheus format anyway, so I figured I'd address that later.

health-checker/src/main/java/io/stargate/health/Server.java
```diff
@@ -0,0 +1,88 @@
...
+            @Override
+            protected void configure() {
+                bind(bundleService).to(BundleService.class);
+            }
+        });
+        environment.jersey().register(CheckerResource.class);
+
+        // Export all DropWizard metrics to Prometheus, they will be picked up later by the
+        // Prometheus MetricsServlet.
+        CollectorRegistry.defaultRegistry.register(new DropwizardExports(metrics.getRegistry()));
```
olim7t, line 66: Contrary to what I thought, Prometheus _will_ pick new DropWizard metrics as they get added dynamically, so we have nothing more to do beyond this.

health-checker/src/main/java/io/stargate/health/Server.java
olim7t: Fixed in faee7d8.

persistence-cassandra-3.11/src/main/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
```diff
@@ -0,0 +1,420 @@
...
+/**
+ * Patches the Cassandra metric implementation to allow injecting our own registry.
+ *
+ * <p>This is messy because of how the original class was designed (no interface contract, relies on
+ * static state...): {@link #actualRegistry} must be set manually before hitting any Cassandra code
+ * that might attempt to register a metric.
+ *
+ * <p>This class must mimic the exact public API of the original one.
+ */
+public class CassandraMetricsRegistry extends MetricRegistry {
```
olim7t, line 23: This is really ugly, but it's the simplest way to intercept metrics without too many impacts on the Cassandra code itself. I hope redefining `org.apache.cassandra` types in our own codebase is OK from a licensing perspective.

The original class also creates a MBean for each metric out of the box. I removed that because I don't think we want JMX by default, but that could be readded if necessary.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/CassandraPersistenceActivator.java
```diff
@@ -106,6 +109,10 @@ private Config makeConfig() throws IOException {
 
     @Override
     public void start(BundleContext context) {
+        ServiceReference<?> metricsReference = context.getServiceReference(Metrics.class.getName());
```
olim7t, line 112: Contrary to what I did in other modules, I'm not registering for `serviceChanged` events here. I wonder if that's really necessary for the health-checker module, given that it's always present (downstream modules have an API dependency to it), and it won't be swapped at runtime.

I can adapt this or the other modules to make it uniform.

jorgebay:
heads up, I don't think I'll be able to review this one before EOD.

dougwettlaufer:
Any idea what's up with the build failure?

olim7t:
> Any idea what's up with the build failure?

I don't know, the VM forked by the failsafe plugin returns with an error. Rebasing didn't solve it, so it's not a transient issue. It's kind of hard to investigate further because I can't run the tests locally.

health-checker/pom.xml
```diff
@@ -0,0 +1,201 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <!-- inherit common settings -->
+  <parent>
+    <groupId>io.stargate</groupId>
+    <artifactId>stargate</artifactId>
+    <version>0.0.6-SNAPSHOT</version>
```
dougwettlaufer, line 8:
```suggestion
    <version>0.0.7-SNAPSHOT</version>
```

cql/pom.xml
```diff
@@ -27,6 +27,12 @@
       <version>0.0.6-SNAPSHOT</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>io.stargate.health</groupId>
+      <artifactId>health-checker</artifactId>
+      <version>0.0.6-SNAPSHOT</version>
```
dougwettlaufer, line 33:
```suggestion
      <version>0.0.7-SNAPSHOT</version>
```

these all should be bumped to 0.0.7 now.

cql/src/main/java/io/stargate/cql/CqlActivator.java
```diff
@@ -62,25 +64,37 @@ public void start(BundleContext context) {
...
+        this.cql.setPersistence((Persistence<?, ?, ?>) context.getService(persistenceReference));
+      }
+
+      metricsReference = context.getServiceReference(Metrics.class.getName());
+      if (metricsReference != null) {
+        log.info("Setting metrics in CQL");
+        this.cql.setMetrics((Metrics) context.getService(metricsReference));
+      }
+
+      if (this.cql.getPersistence() != null && this.cql.getMetrics() != null) {
```
dougwettlaufer, line 88: Should we have a follow on ticket for making metrics optional?

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -144,10 +145,10 @@ private int countConnectedClients() {
   }
 
   private <T> Gauge<T> registerGauge(String name, Gauge<T> gauge) {
-    return Metrics.register(factory.createMetricName(name), gauge);
```
dougwettlaufer, line 147: What was this existing `CassandraMetricsRegistry.Metrics` doing? Might this impact any of the metrics in the persistence modules?

health-checker/src/main/java/io/stargate/health/metrics/impl/PrefixingMetricRegistry.java
```diff
@@ -0,0 +1,348 @@
...
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * Wraps an existing registry to prefix every new metric with a given string on the fly.
+ *
+ * <p>For clients of this class, it looks like they're interacting with a regular registry, when
+ * they are in fact only seeing one part of a bigger registry, that has other metrics under
+ * different prefixes.
+ *
+ * <p>Implementation note: unfortunately {@link MetricRegistry} is a class, not an interface. So we
```
dougwettlaufer, line 29: Missing the rest of this sentence?

persistence-cassandra-3.11/src/main/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
dougwettlaufer: Is there a way to do this generically or will we need to reimplement for each persistence module?

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/CassandraPersistenceActivator.java
```diff
@@ -121,6 +124,10 @@ private Config makeConfig() throws IOException {
 
   @Override
   public void start(BundleContext context) {
+    ServiceReference<?> metricsReference = context.getServiceReference(Metrics.class.getName());
```
dougwettlaufer, line 127: Doesn't this need a service listener?

dougwettlaufer (commented):
Looks good! Will approve pending passing build

persistence-cassandra-3.11/src/main/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
olim7t: The class's public contract changes across Cassandra versions, so we'll have to redefine it every time.

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
olim7t: It was wrapping a static `MetricRegistry` instance, and also publishing each metric as a JMX MBean.

I replaced the static registry by the one injected from the health-checker module. I also disable JMX because I don't think we want to do that by default, but that could be easily re-added if necessary.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/CassandraPersistenceActivator.java
olim7t: I had a comment asking about that, but it got lost in the formatting changes: do we need a service listener for every bundle, or only those that might be changed at runtime?

mpenick: The problem is initialization ordering can cause other modules to not be able during the initialization of this module. Also, it'll enable hot reloading.

jorgebay (commented):
It's looking great.

I've noticed that stargate nodes with `./starctl ... --bind-to-listen-address` stopped working w/ this change. Probably because how host address was set originally, which is not ideal...

dougwettlaufer (approved):

PR #42, opened 2020-09-14
CQL: Add token-based authentication

mpenick:


testing/src/test/java/io/stargate/it/CQLTest.java
```diff
@@ -88,8 +87,8 @@ public void setup() {
     table = testName;
   }
 
-  @BeforeClass
```
mpenick, line 91: We shouldn't be using static state here because it relies on non-static state `stargateHost` in `BaseOsgiIntegrationTest`. If a test is not run beforehand then `stargateHost` will be `null`.

auth-api/pom.xml
```diff
@@ -5,7 +5,7 @@
   <parent>
     <groupId>io.stargate</groupId>
     <artifactId>stargate</artifactId>
-    <version>0.0.8-SNAPSHOT</version>
+    <version>0.0.9-SNAPSHOT</version>
```
dimas-b, line 8: Are we not doing module versions in sync with the parent POM?.. just wondering.

cql/src/main/java/io/stargate/cql/CqlActivator.java
```diff
@@ -57,30 +62,42 @@ private static Config makeConfig() {
...
-          log.info("Started CQL....");
+      ServiceReference[] refs =
+          context.getServiceReferences(AuthenticationService.class.getName(), null);
+      if (refs != null) {
+        for (ServiceReference ref : refs) {
+          // Get the service object.
+          Object service = context.getService(ref);
+          if (service instanceof AuthenticationService
+              && ref.getProperty("AuthIdentifier") != null
+              && ref.getProperty("AuthIdentifier").equals(AUTH_IDENTIFIER)) {
```
dimas-b, line 82: Check this out (just an idea): https://github.com/stargate/api-extensions/blob/add-sql-connection-auth/sql/server/src/main/java/io/stargate/api/sql/server/SqlActivator.java#L47

persistence-api/src/main/java/io/stargate/db/Persistence.java
```diff
@@ -44,6 +44,8 @@
 
   ClientState newClientState(String name);
 
+  AuthenticatedUser<?> newAuthenticatedUser(String name);
```
dimas-b, line 47: Opening this API does not sound right to me. Since we validate credentials (token) via `AuthenticationService` service and by-pass the negotiator from `Persistence` in this case, would it be preferable to have a local implementation of `AuthenticatedUser` (in the `cql` module) for this use case?

dimas-b, line 47: Opening this API does not sound right to me. Since we validate credentials (token) via `AuthenticationService` service and by-pass the negotiator from `Persistence` in this case, would it be preferable to have a local implementation of `AuthenticatedUser` (in the `cql` module) for this use case?

dimas-b (changes requested):

persistence-api/src/main/java/io/stargate/db/Persistence.java
dimas-b: Having thought a bit more about this, I guess the `AuthenticationService` should perhaps negotiate with `Persistence` and establish an authenticated `ClientState`... WDYT?

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/PlainTextTokenSaslNegotiator.java
```diff
@@ -0,0 +1,113 @@
...
+      Integer.parseInt(System.getProperty("stargate.cql_token_max_length", "36"));
+
+  static final byte NUL = 0;
+
+  private final Persistence persistence;
+  private final AuthenticationService authentication;
+  private final Authenticator.SaslNegotiator wrapped;
+  private StoredCredentials credentials;
+  private String username;
+  private String password;
```
tomekl007, line 28: shouldn't this be a `char[]` for security reasons?

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/PlainTextTokenSaslNegotiator.java
```diff
@@ -0,0 +1,113 @@
...
+
+  @Override
+  public AuthenticatedUser<?> getAuthenticatedUser() throws AuthenticationException {
+    if (credentials != null) return persistence.newAuthenticatedUser(credentials.getRoleName());
+    else return wrapped.getAuthenticatedUser();
+  }
+
+  private boolean attemptTokenAuthentication(byte[] clientResponse) {
+    try {
+      decodeCredentials(clientResponse);
```
tomekl007, line 58: I find this design a bit misleading. It seems that `decodeCredentials` is called every time the `attemptTokenAuthentication` is called, but it sets the `password` and `username` fields on this class. Those fields are no used outside of this context, so I think that this should be reworked to stateless logic. The `decodeCredentials` should return username/password (maybe wrapped in some light class) instead of setting class fields. It will make logic simpler and easier to test. It would make sense to have class fields if the results of the `decodeCredentials` will be memoized.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/PlainTextTokenSaslNegotiator.java
```diff
@@ -0,0 +1,113 @@
...
+    return credentials != null || wrapped.isComplete();
+  }
+
+  @Override
+  public AuthenticatedUser<?> getAuthenticatedUser() throws AuthenticationException {
+    if (credentials != null) return persistence.newAuthenticatedUser(credentials.getRoleName());
+    else return wrapped.getAuthenticatedUser();
+  }
+
+  private boolean attemptTokenAuthentication(byte[] clientResponse) {
```
tomekl007, line 56: I think that we should unit test this method (or `decodeCredentials`) at least.

persistence-api/src/main/java/io/stargate/db/Persistence.java
mpenick: The problem with this is you end up duplicating the logic over and over in all the different persistence implementation.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/PlainTextTokenSaslNegotiator.java
mpenick: Yes, that makes sense.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/PlainTextTokenSaslNegotiator.java
mpenick: Is this related to string duplication? I removed these fields based the suggestion below. `password` will need to be a string at some point though for the `Authentication` API. Maybe we should consider changing that API in another ticket.

cql/src/main/java/io/stargate/cql/CqlActivator.java
mpenick: That's awesome, thanks for the suggestion.

persistence-api/src/main/java/io/stargate/db/Persistence.java
mpenick: Let me take a look a this a bit more.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/PlainTextTokenSaslNegotiator.java
tomekl007: Yes, it is related to the java string pool (https://www.baeldung.com/java-string-pool). So there is a risk that such a `String password` may leak.

cql/src/main/java/io/stargate/cql/CqlActivator.java
mpenick: @dimas-b These checks are incorrect:

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/PlainTextTokenSaslNegotiator.java
dimas-b: +1 on `char[]` but realistically, it will probably become a `String` anyway... either up to down the call chain :)

cql/src/main/java/io/stargate/cql/CqlActivator.java
dimas-b: yes, I hit that already :) thanks for the pointer though :) should be fixed in the latest commit

dimas-b: ah.... I forgot to push :facepalm:

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/PlainTextTokenSaslNegotiator.java
mpenick: The values are going to be converted to `String` anyway in the case of the wrapped negotiator. I'm not sure this is worth the effort right now. I'm not even sure you can convert a byte array to a char array without using an intermediate `String` anyway.
https://github.com/apache/cassandra/blob/cassandra-3.11.8/src/java/org/apache/cassandra/auth/PasswordAuthenticator.java#L214-L215

persistence-api/src/main/java/io/stargate/db/Persistence.java
mpenick: The issue here I need to interoperate with an existing `ClientState`:
https://github.com/stargate/stargate/blob/v0.0.8/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java#L78-L79

Also, I would prefer not refactor `AuthReponse` as part of this feature.

mpenick:
Matching persistence extensions PR: https://github.com/stargate/persistence-extensions/pull/14

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -115,7 +115,7 @@ public void initialize(Config config) {
 
     daemon.start();
 
-    waitForSchema(StorageService.RING_DELAY);
+    waitForSchema(5 * StorageService.RING_DELAY);
```
mpenick, line 118: This fixes the schema agreement issue because it waits (a maximum, probably less) 2.5 minutes which is guaranteed to be longer than the migration of schema which can be 2 minutes in some cases. Anyway, I have a follow up ticket for this called: "Schema agreement times out for DDL queries (usually after startup)".

persistence-api/src/main/java/io/stargate/db/Persistence.java
dimas-b: Understood, but in principle, having an API that just creates an authenticated user looks dangerous to me. In the long run, would it make sense to direct all authentication call through the `AuthenticationService` only? `Persistence` implementations could connect to it through a separate API following the model similar to JRE's `javax.security.auth.spi.LoginModule`, WDYT?

dimas-b: Would it be worth opening a follow-up ticket?

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/PlainTextTokenSaslNegotiator.java
```diff
@@ -0,0 +1,127 @@
...
+      return persistence.newAuthenticatedUser(storedCredentials.getRoleName());
+    else return wrapped.getAuthenticatedUser();
+  }
+
+  @VisibleForTesting
+  boolean attemptTokenAuthentication(byte[] clientResponse) {
+    try {
+      Credentials credentials = decodeCredentials(clientResponse);
+
+      if (!credentials.username.equals(TOKEN_USERNAME)) return false;
```
dougwettlaufer, line 61: Can we guarantee username is not null?

dougwettlaufer (dismissed):

stargate-starter/src/main/java/io/stargate/starter/Starter.java
```diff
@@ -288,6 +288,11 @@ void setStargateProperties() {
     }
   }
 
+  public Starter withAuthEnabled(boolean enabled) {
+    enableAuth = enabled;
```
olim7t, line 292: I'm a bit confused by this. It's only used to set PasswordAuthenticator and CassandraAuthorizer on the Cassandra daemon that we start in the persistence module.

Is that a separate mechanism from auth-table-based-service? When would you use both?

olim7t (commented):
How do you manually test this? I tried building locally, but I can still run cqlsh without authentication.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/PlainTextTokenSaslNegotiator.java
olim7t: Yes: decodeCredentials throws if the user is null.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/ServerConnection.java
```diff
@@ -119,11 +123,16 @@ void applyStateTransition(Message.Type requestType, Message.Type responseType) {
...
+    if (saslNegotiator == null) {
       saslNegotiator =
           persistence
               .getAuthenticator()
               .newSaslNegotiator(queryState.getClientAddress(), certificates());
+
+      if (authentication != null)
+        saslNegotiator =
+            new PlainTextTokenSaslNegotiator(saslNegotiator, persistence, authentication);
+    }
```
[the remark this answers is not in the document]
mpenick: Thanks, nice catch.

stargate-starter/src/main/java/io/stargate/starter/Starter.java
mpenick: auth-table-based-service relies on CassandraAuthorizer as it converts auth tokens to C* roles.

persistence-api/src/main/java/io/stargate/db/Persistence.java
mpenick: > Understood, but in principle, having an API that just creates an authenticated user looks dangerous to me

dimas-b (dismissed):

mpenick:
@olim7t To run:

Create a 3 node cluster (needed for the table-based token auth RF) with auth enabled:
```
ccm create stargate -v 3.11.6 -n 3 -s -b --pwd-auth
```

Startup stargate with a `cql_auth_id`:
```
java -Dstargate.libdir=./stargate-lib  -Dstargate.cql_auth_id=AuthTableBasedService -jar stargate-lib/stargate-starter-0.0.9-SNAPSHOT.jar --cluster-name stargate --cluster-seed 127.0.0.1 --cluster-version 3.11 --listen 127.0.0.10 --simple-snitch --enable-auth
```

mpenick:
You'll need to generate a token to test token-based auth through cqlsh.

POST to `http://127.0.0.1:8081/v1/auth/token/generate` with the body:

```json
{
    "key": "<some username>",
    "secret": "<some password>"
}
```

w/ a `ccm` cluster you can use the `cassandra`/`cassandra` credentials.

olim7t:
Why do we need to pass `--stargate.cql_auth_id` or `--enable-auth` at startup? For other modules (REST, GraphQL...), token-based works out of the box without additional options.

dougwettlaufer (approved):

olim7t:
> I made `--stargate.cql_auth_id` distinct from `--stargate.auth_id`

Does that mean that they can be set to different values?

Doesn't the `AuthenticationService` you fetch in `CqlActivator` have to match the one that was set on the `apiServer` in `AuthApiActivator`?

mpenick:
@olim7t Yes, fix incoming. Use `-Dstargate.cql_use_auth_servicecql_use_auth_service=true` to enable/disable. Now using regular `stargate.auth_id`.

dougwettlaufer (approved):

PR #46, opened 2020-09-15
More CQL tests

jorgebay:
Added some more CQL tests for compression, paging and UDTs.

I've taken out the paging tests until the page state issue is resolved.

### DIFF
--- a/testing/src/test/java/io/stargate/it/CQLTest.java
+++ b/testing/src/test/java/io/stargate/it/CQLTest.java
@@ -503,7 +503,6 @@ public class CQLTest extends BaseOsgiIntegrationTest
             .withBoolean(DefaultDriverOption.METADATA_TOKEN_MAP_ENABLED, false)
             .withString(DefaultDriverOption.LOAD_BALANCING_POLICY_CLASS, DcInferringLoadBalancingPolicy.class.getName())
             .withDuration(DefaultDriverOption.REQUEST_TIMEOUT, Duration.ofSeconds(5))
-            .withDuration(DefaultDriverOption.CONNECTION_INIT_QUERY_TIMEOUT, Duration.ofSeconds(5))
             .withDuration(DefaultDriverOption.CONTROL_CONNECTION_TIMEOUT, Duration.ofSeconds(5));
     }
 }
--- a/testing/src/test/java/io/stargate/it/CQLTest.java
+++ b/testing/src/test/java/io/stargate/it/CQLTest.java
@@ -20,6 +20,8 @@ import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
 import java.time.Duration;
+import java.util.Arrays;
+import java.util.Date;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
@@ -27,20 +29,21 @@ import java.util.function.Supplier;
 
 import com.datastax.oss.driver.api.core.cql.BoundStatement;
 import com.datastax.oss.driver.api.core.cql.Statement;
+import org.junit.AfterClass;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TestName;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import com.datastax.oss.driver.api.core.CqlIdentifier;
 import com.datastax.oss.driver.api.core.CqlSession;
 import com.datastax.oss.driver.api.core.config.DefaultDriverOption;
 import com.datastax.oss.driver.api.core.config.DriverConfigLoader;
+import com.datastax.oss.driver.api.core.config.ProgrammaticDriverConfigLoaderBuilder;
 import com.datastax.oss.driver.api.core.cql.BatchStatement;
 import com.datastax.oss.driver.api.core.cql.BatchType;
 import com.datastax.oss.driver.api.core.cql.PreparedStatement;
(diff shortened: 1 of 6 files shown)


jorgebay:
Here's the branch for the paging test: https://github.com/stargate/stargate/compare/more-cql-tests...paging-test

We can merge that one after paging is fixed @tomekl007

testing/src/test/java/io/stargate/it/CQLTest.java
```diff
@@ -30,52 +33,63 @@
...
+
+    @BeforeClass
+    public static void beforeAll()
+    {
+        DriverConfigLoader loader =
+            DriverConfigLoader.programmaticBuilder()
+                .withBoolean(DefaultDriverOption.METADATA_TOKEN_MAP_ENABLED, false)
+                .withString(DefaultDriverOption.LOAD_BALANCING_POLICY_CLASS, DcInferringLoadBalancingPolicy.class.getName())
+                .withDuration(DefaultDriverOption.REQUEST_TIMEOUT, Duration.ofSeconds(5))
+                .withDuration(DefaultDriverOption.CONNECTION_INIT_QUERY_TIMEOUT, Duration.ofSeconds(5))
```
tomekl007, line 79: The `init-query-timeout` default is `5 seconds` (but we would need to update driver to `4.8`

testing/src/test/java/io/stargate/it/CQLTest.java
```diff
@@ -350,6 +364,52 @@ public void tupleTest()
...
 
+    @Test
+    public void compressionTest()
+    {
+        Arrays.stream(new String[] {"lz4", "snappy"}).forEach(compression -> {
+            DriverConfigLoader loader =
+                DriverConfigLoader.programmaticBuilder()
+                    .withBoolean(DefaultDriverOption.METADATA_TOKEN_MAP_ENABLED, false)
+                    .withString(DefaultDriverOption.LOAD_BALANCING_POLICY_CLASS, DcInferringLoadBalancingPolicy.class.getName())
+                    .withDuration(DefaultDriverOption.REQUEST_TIMEOUT, Duration.ofSeconds(5))
```
tomekl007, line 375: maybe extract this logic to a separate method? (since it is duplicated)

jorgebay: good call, will do.

jorgebay: done!

testing/src/test/java/io/stargate/it/CQLTest.java
```diff
@@ -350,6 +355,46 @@ public void tupleTest()
         assertThat(tupleReturnValue.getInt(2)).isEqualTo(2);
     }
 
+    @Test
+    public void compressionTest()
+    {
+        Arrays.stream(new String[] {"lz4", "snappy"}).forEach(compression -> {
```
dimas-b, line 361: What do you think about using JUnit 5 parameters? It should be possible to apply them on a per-test method basis.

olim7t: +1
And in this case the parameters are compile-time constant so you can provide them with a simple annotation:
```java
    @ParameterizedTest
    @ValueSource(strings = { "lz4", "snappy" })
    public void compressionTest(String compression) { ... }
```

testing/src/test/java/io/stargate/it/CQLTest.java
```diff
@@ -31,51 +34,53 @@
...
 {
-    private static final Logger logger = LoggerFactory.getLogger(PersistenceTest.class);
-
     @Rule
     public TestName name = new TestName();
 
     private String table;
     private String keyspace;
-    private CqlSession session;
+    private static CqlSession session;
```
olim7t, line 59: Nit: this module pulls the Java driver as a transitive dependency of `persistence-api`. Since it uses the driver directly, I think it would be better to declare it in its own POM.

olim7t (approved):

dougwettlaufer (approved):

testing/src/test/java/io/stargate/it/CQLTest.java
jorgebay: Makes sense, will do.

testing/src/test/java/io/stargate/it/CQLTest.java
```diff
@@ -31,51 +34,53 @@
...
+            .withConfigLoader(getDriverConfigLoaderBuilder()
+                .withDuration(DefaultDriverOption.REQUEST_TRACE_INTERVAL, Duration.ofSeconds(1))
+                .build())
+            .addContactPoint(new InetSocketAddress(stargateHost, 9043)).build();
+    }
+
+    @AfterClass
+    public static void afterAll()
+    {
+        session.close();
```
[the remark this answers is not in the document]
jorgebay: This can be addressed when we improve the test harness and how we create integration tests (https://datastax.jira.com/browse/C2-286). 

If its ok by you, I'll leave this suggestion out.

testing/src/test/java/io/stargate/it/CQLTest.java
jorgebay: Nice catch!

jorgebay:
Rebased against master.

testing/src/test/java/io/stargate/it/CQLTest.java
```diff
@@ -440,9 +483,18 @@ public void shouldUsePaginationForSimpleStatement()
             Thread.sleep(100);
```
tomekl007, line 483: maybe we could import the Awaitality library for test blocking code?
https://github.com/awaitility/awaitility
Then it could be changed to 
`await().atMost(100, MILLISECONDS).until(actionToWait());`

jorgebay: ok, I've filed a ticket for it.

https://datastax.jira.com/browse/C2-307

jorgebay:
I've addressed most of the suggestions from the initial reviews.

testing/src/test/java/io/stargate/it/CQLTest.java
```diff
@@ -76,12 +77,19 @@
...
+    private static final int KEYSPACE_NAME_MAX_LENGTH = 48;
+
     @Before
     public void setup()
     {
         String testName = name.getMethodName();
         testName = testName.substring(0, testName.indexOf("["));
         keyspace = "ks_" + new Date().getTime() + "_" + testName;
+
+        if (keyspace.length() > KEYSPACE_NAME_MAX_LENGTH) {
```
tomekl007, line 89: It may be dangerous - if there will be 2 long tests with the same suffix, they will end up in the same keyspace. Maybe we can randomize the last 2 characters or something similar?

testing/src/test/java/io/stargate/it/CQLTest.java
tomekl007: nice, thx

testing/src/test/java/io/stargate/it/CQLTest.java
```diff
@@ -440,9 +493,18 @@ public void shouldUsePaginationForSimpleStatement()
...
         return null;
     }
 
+    private static ProgrammaticDriverConfigLoaderBuilder getDriverConfigLoaderBuilder()
+    {
+        return DriverConfigLoader.programmaticBuilder()
+            .withBoolean(DefaultDriverOption.METADATA_TOKEN_MAP_ENABLED, false)
+            .withString(DefaultDriverOption.LOAD_BALANCING_POLICY_CLASS, DcInferringLoadBalancingPolicy.class.getName())
+            .withDuration(DefaultDriverOption.REQUEST_TIMEOUT, Duration.ofSeconds(5))
+            .withDuration(DefaultDriverOption.CONNECTION_INIT_QUERY_TIMEOUT, Duration.ofSeconds(5))
```
tomekl007, line 506: if you upgraded to java-driver `4.9`, you don't need to override this setting any more - the default is `5 seconds`.

testing/src/test/java/io/stargate/it/CQLTest.java
jorgebay: It uses the timestamp as prefix. The table will still be different, I'm just limiting when reaching maximum.

testing/src/test/java/io/stargate/it/CQLTest.java
jorgebay: Removed it.

tomekl007 (approved):

mpenick (approved):

testing/src/test/java/io/stargate/it/CQLTest.java
```diff
@@ -49,53 +53,63 @@
...
 import com.datastax.oss.driver.api.core.type.DataTypes;
 import com.datastax.oss.driver.api.core.type.TupleType;
+import com.datastax.oss.driver.api.core.type.UserDefinedType;
 import com.datastax.oss.driver.internal.core.loadbalancing.DcInferringLoadBalancingPolicy;
 import net.jcip.annotations.NotThreadSafe;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
 
 @RunWith(Parameterized.class)
```
dimas-b, line 69: Do we need this at class level with JUnit 5? Method-level annotations seem to be sufficient.

testing/src/test/java/io/stargate/it/CQLTest.java
```diff
@@ -20,27 +20,31 @@
...
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Supplier;
 
 import com.datastax.oss.driver.api.core.cql.BoundStatement;
 import com.datastax.oss.driver.api.core.cql.Statement;
+import org.junit.AfterClass;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.BeforeClass;
```
dimas-b, line 34: Could we use only JUnit 5 annotations just for the sake of clarify and consistency? WDYT?

testing/src/test/java/io/stargate/it/CQLTest.java
jorgebay: Can we create a ticket for further enhancements, I fear this PR for adding a couple of tests it's getting out of control.

dimas-b: #94

jorgebay: Thanks @dimas-b , appreciate!

PR #89, opened 2020-09-15
Remove last GH packages reference

dougwettlaufer:


.github/workflows/release.yml
```diff
@@ -79,14 +79,18 @@ jobs:
           java-version: 1.8
           server-username: MAVEN_USERNAME
           server-password: MAVEN_PASSWORD
+      - id: install-secret-key
```
dougwettlaufer, line 82: Instructions per https://gist.github.com/sualeh/ae78dc16123899d7942bc38baba5203c

.github/workflows/release.yml
```diff
@@ -79,14 +79,18 @@ jobs:
...
+        name: Install gpg secret key
+        run: |
+          cat <(echo -e "${{ secrets.OSSRH_GPG_SECRET_KEY }}") | gpg --batch --import
```
dimas-b, line 85: I suppose these keys remain installed after the build is over... Do we wipe build hosts after publishing?

olim7t (approved):

.github/workflows/release.yml
dougwettlaufer: I'm not entirely sure what GH does with their infra. I'm assuming after the build is complete they tear everything down.

dougwettlaufer: And the number of secrets the actions are already passed I think gpg key is the least of our problems :)

PR #91, opened 2020-09-15
Enforce Java and XML source formatting

olim7t:
Java: Google style, the rules are not configurable. There are plugins for Intellij IDEA and Eclipse, and you can format from the command line with `mvn fmt:format`.

XML: the plugin has configuration options, but currently I've kept the default, I don't know if it's worth spending time fine-tuning it. There are no IDE plugins so you'd have to configure your IDE to match at least the indent. `mvn xml-format:xml-format` to format from the command line.

I haven't formatted the sources yet, I think it's better to do it in a separate commit, and then maybe there will be a clever way to rebase existing PRs without too much disruption.

There are 2 ways we can configure this:
1. the build fails if the code is not properly formatted.
2. the build formats the code, and we add some sort of github action that commits the changes if any.

I like 1 and this is how it's currently configured, but it makes the build a bit longer, and gets in the way (you'll find out that you forget to format *very often*). I don't like 2 because it pollutes the git history.


pom.xml
```diff
@@ -70,6 +70,37 @@
                     <autoReleaseAfterClose>false</autoReleaseAfterClose>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>com.coveo</groupId>
+                <artifactId>fmt-maven-plugin</artifactId>
+                <version>2.10</version>
```
olim7t, line 76: Version 2.10 requires Java 11 at least.
If we want to build with Java 8, we need to downgrade to 2.9.

dimas-b:
> I like 1 and this is how it's currently configured, but it makes the build a bit longer, and gets in the way (you'll find out that you forget to format very often)

+1 to option 1.

I'm not sure about this plugin, but `spotless` can reformat all code with an explicit `mvn spotless:apply` I found it was not too hard when one makes a formatting mistake (IDEA catches a lot of those)

dimas-b (approved):

pom.xml
dougwettlaufer: Java 11 is not currently an option

dougwettlaufer:
> There are 2 ways we can configure this:

What about option 3, commit hook that runs formatter plus build fails on bad format? 

> I don't like 2 because it pollutes the git history.

Shouldn't be a problem with our "Squash and Merge" right?


I'm not sure what the convention is but can you put something quick in either CONTRIBUTING.md or DEV_GUIDE.md concerning the formatting? A link or two for setting this up in IDEA/Eclipse plus a "hey format your code with `mvn ...` or else the build breaks"

dougwettlaufer:
Might also be worth looking into

https://github.com/marketplace/actions/google-java-format
https://github.com/marketplace/actions/super-linter
https://peterevans.dev/posts/github-actions-how-to-automate-code-formatting-in-pull-requests/
https://github.com/michalrus/git-hooks-code-autoformat

tomekl007 (approved):
Maybe we could also add an XML file formatter in this PR?

jorgebay (approved):
> Maybe you could run the fmt to format all existing files

+1 on reformatting all existing files as part of this PR. We can rebase and fix conflicts on the other open PRs.

pcmanus (approved):
Happy to get existing code reformatted with this ASAP as well. Don't mind if it's a separate commit in this PR or a quick followup PR though.

pom.xml
```diff
@@ -70,6 +70,37 @@
...
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
+                        <phase>process-sources</phase>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>au.com.acegi</groupId>
+                <artifactId>xml-format-maven-plugin</artifactId>
```
dimas-b, line 88: This formatter seems too strict to me. For example, it does not allow multi-line XML tags, which are often convenient and more readable. For example, for OSGi:
```
            <Import-Package>
              org.osgi.*,
              io.stargate.db.*,
              org.apache.cassandra.stargate.db.*,
              !*
            </Import-Package>
```
vs.
```
            <Import-Package>org.osgi.*, io.stargate.db.*, org.apache.cassandra.stargate.db.*, !*</Import-Package>
```

dimas-b: seems to work with `CDATA`:
```
            <Import-Package><![CDATA[
              org.osgi.*,
              io.stargate.db.*,
              org.apache.cassandra.stargate.db.*,
              !*
            ]]></Import-Package>
```

pom.xml
dougwettlaufer: Probably why the build is failing

olim7t:
> Shouldn't be a problem with our "Squash and Merge" right?

Right, I hadn't thought about that, but if the formatting commits are limited to PRs, it's not a problem indeed.

pom.xml
olim7t: OK, I downgraded to 2.9. I hope this won't introduce incompatibilities between the versions of google-formatter used by this plugin and IDE plugins.

dougwettlaufer (approved):

dougwettlaufer (approved):

PR #93, opened 2020-09-16
Create integration test that with recommended driver settings Stargate nodes are uniformly loaded with requests

tomekl007:
todo: 
- [x] simulate scenario when one stargate node is down - **won't fix:** it will be implemented as a part of a separate task


tomekl007:
I still need to handle the one more test case: https://github.com/stargate/stargate/pull/93#issue-488030600 but feel free to review for faster feedback.

testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
```diff
@@ -384,26 +447,63 @@ public void baseSetup() throws BundleException, InterruptedException, IOExceptio
...
+        } finally {
+          logger.error("Successful starting stargate node nr: {}", i);
+        }
+      }
     }
   }
+
+  private void startStargateInstance(String seedHost, Integer seedPort, int stargateNodeNumber)
+      throws IOException, BundleException {
+    int jmxPort = new ServerSocket(0).getLocalPort();
```
tomekl007, line 483: Sometimes, for the last stargate instance (the 3rd one in this case), I am getting:
`Port already in use:` even that this port is picked just before the https://github.com/stargate/stargate/pull/93/files#diff-2c65e0c4892da78e91b8b4d3d357e25aR492.
To solve that, I am retrying the start action just once, and it seems that it helps. Let me know if you have a better idea.

testing/src/test/java/io/stargate/it/backends/Cassandra311Backend.java
```diff
@@ -24,6 +24,6 @@
 public class Cassandra311Backend extends AllBackendSuite {
   @Parameterized.Parameters(name = "{index}: {0}")
   public static Iterable<Object[]> functions() {
-    return Collections.singletonList(new Object[] {"cassandra:3.11.6", false, "3.11"});
+    return Collections.singletonList(new Object[] {"cassandra:3.11.6", false, "3.11", 3});
```
tomekl007, line 27: With this approach, all our tests will run with 3 stargate nodes, it is not ideal but other solutions turn out to be not feasible/optimal: 
i.e.: creating `Cassandra311BackendMulitpleNodes` and `BackandSuiteMultipleNodes` with only one test: `MultipleStargateInstancesTest` will lead to code duplication and two suites to extend by the other persistence implementations.

testing/src/test/java/io/stargate/it/backends/AllBackendSuite.java
```diff
@@ -36,4 +43,7 @@
 
   @Parameterized.Parameter(2)
   public String version;
+
+  @Parameterized.Parameter(3)
+  public Integer numberOfStargateNodes;
```
tomekl007, line 48: I think that this parameter would be useful for other persistence implementations.

persistence-api/src/main/java/io/stargate/db/datastore/DataStore.java
```diff
@@ -80,9 +80,9 @@ default void waitForSchemaAgreement() {
       if (isInSchemaAgreement()) {
         return;
       }
-      Uninterruptibles.sleepUninterruptibly(100, TimeUnit.MILLISECONDS);
```
mpenick, line 83: Maybe the is should be configurable? Using something like a `stargate.schema_agreement_timeout` setting.

tomekl007: I agree, but it does not look strictly related to this ticket. Maybe we can do it in a separate JIRA task? (I can create one)

jorgebay:
> simulate scenario when one stargate node is down

I think that should be out of scope from this task. Should we create a separate ticket for it?

tomekl007:
> > simulate scenario when one stargate node is down
> 
> I think that should be out of scope from this task. Should we create a separate ticket for it?

This scenario was in the ticket description

testing/src/test/java/io/stargate/it/backends/Cassandra311Backend.java
mpenick: If there are limitations in passing parameters to the suite, then maybe there could be a `Starter addStargateNode()` method that the inherited test class could use to create additional stargate nodes?

mpenick: This would be useful for CQL tests that I need to write to test `system.peers` functionality.

tomekl007: right now the suite with 3 nodes seems working (there is sometimes the race condition (bind address already in use) on Jenkins when starting the last node). The `addStargate()` called by the actual test method is problematic because of the stateful logic and changing the IPs if:
```
  if (Boolean.getBoolean(SKIP_HOST_NETWORKING_FLAG)) {
```
I am not sure if this MAC code is working, I was not able to make the Mac part work for me, @jorgebay neither.
The 2nd issue with this approach is static state and the fact that test are not isolated. So starting new starter host will impact other ITs. Ideally, we would have `Junit rule` or something similar for `Stargate` that will be spin per test class suite and isolated from eachother.

tomekl007:
@mpenick I've adapted this PR to the new CCM, but there is a new problem with `health-checker.port`:
```
server:
  type: health-checker
  applicationConnectors:
    - type: http
      port: 8084
```
If there are N stargate nodes started on local, there is a port clash:
```
Caused by: java.io.IOException: Failed to bind to /127.0.0.13:8084
	at org.eclipse.jetty.server.ServerConnector.openAcceptChannel(ServerConnector.java:346)
	at org.eclipse.jetty.server.ServerConnector.open(ServerConnector.java:307)
	at org.eclipse.jetty.setuid.SetUIDListener.lifeCycleStarting(SetUIDListener.java:216)
	... 88 more
Caused by: java.net.BindException: Address already in use
```
do you know if there is a way to configure it programmatically (to get a free port similar to what I did for jmx port)?

dougwettlaufer:
> @mpenick I've adapted this PR to the new CCM, but there is a new problem with `health-checker.port`:
> 
> ```
> server:
>   type: health-checker
>   applicationConnectors:
>     - type: http
>       port: 8084
> ```
> 
> If there are N stargate nodes started on local, there is a port clash:
> 
> ```
> Caused by: java.io.IOException: Failed to bind to /127.0.0.13:8084
> 	at org.eclipse.jetty.server.ServerConnector.openAcceptChannel(ServerConnector.java:346)
> 	at org.eclipse.jetty.server.ServerConnector.open(ServerConnector.java:307)
> 	at org.eclipse.jetty.setuid.SetUIDListener.lifeCycleStarting(SetUIDListener.java:216)
> 	... 88 more
> Caused by: java.net.BindException: Address already in use
> ```
> 
> do you know if there is a way to configure it programmatically (to get a free port similar to what I did for jmx port)?
> 
> **update** : this does not seems to manifest on Jenkins, only on local.

The health-checker is probably just the first to clash, odds are all of the http services will fails since they bind to a specific port. To fix it we'll need to make the ports configurable through a `-Dstargate.restapi_port=8XXX`. It is odd that GH action is passing.

tomekl007:
> Failed to bind to /127.0.0.13:8084

@dougwettlaufer I think that we should handle this using the approach that @jorgebay implemented in https://github.com/stargate/stargate/pull/18/files but the health-checker module is using a different mechanism for registering connectors:

I observe the port clash on the health-checker module:
```
Failed to bind to /127.0.0.13:8084
```
The graphapi and auth are using this mechanism to prevent that:

```
ServerConnector connector = new ServerConnector(server);
    if (Boolean.parseBoolean(System.getProperty("stargate.bind_to_listen_address"))) {
      connector.setHost(System.getProperty("stargate.listen_address"));
    }
    connector.setPort(port);
    server.addConnector(connector);
```
then, if I will do `server.getConnectors` it returns one connector:
```
ServerConnector@39d37b3d{HTTP/1.1,[http/1.1]}{127.0.0.11:8081}
```

Unfortunately, the health-checker api uses a different configuration mechanism for creating connectors. It uses the `config.yaml`:
```
server:
  type: health-checker
  applicationConnectors:
    - type: http
      port: 8084
  adminConnectors:
    - type: http
      port: 8085
```
Because of that, when I do `server.getConnectors` I get two (application and admin):
```
{ServerConnector@6306} "application@aee05f4{HTTP/1.1,[http/1.1]}{127.0.0.11:8084}"
{ServerConnector@6307} "admin@2b4829aa{HTTP/1.1,[http/1.1]}{127.0.0.11:8085}"
```
They both exposed under the stargate specific address (not `127.0.0.1`) so it should work out-of-the-box but somehow there is a port clash between stargate instance started on `127.0.0.11:8084` and `127.0.0.13:8084`

I am able to plug into the server creation lifecycle https://github.com/stargate/stargate/pull/93/commits/28ede9c49982fa2ee98c141d95ceed7a4154b8e1
but it results in 3 connectors:
```
{Connector[3]@6304} 
{ServerConnector@6203} "ServerConnector@4ab455e2{HTTP/1.1,[http/1.1]}{127.0.0.11:8085}"
{ServerConnector@6306} "application@aee05f4{HTTP/1.1,[http/1.1]}{127.0.0.11:8084}"
{ServerConnector@6307} "admin@2b4829aa{HTTP/1.1,[http/1.1]}{127.0.0.11:8085}"
```
that they also clash with each other (because there are 2 connectors exposed on the port 8085.
@olim7t could you tell me why we needed to pick a different config mechanism for health-checker API (comparing to the previous graph and auth)?

Finally, when I rebase my branch today (changes from 25.09) the port clash does not happen anymore (I run this multiple times on my local and there is no a problem, the Jenkins build does not fail with the address already in use anymore)

testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
```diff
@@ -73,15 +78,43 @@
...
 
-  static Starter stargateStarter;
-  public static String stargateHost = "127.0.0.11";
   public boolean enableAuth;
+
   protected final ClusterConnectionInfo backend;
 
+  public static List<Starter> stargateStarters = new ArrayList<>();
+  private static List<String> stargateHosts = new ArrayList<>();
+  public static final Integer numberOfStargateNodes = 3;
```
tomekl007, line 87: Making it parameterizable will require refactoring the whole testing code to make the `Stargate` runners not shared between test classes. Such a rework will be a substantial effort and will be done in a separate ticket (we have the ticket for it already)
The three, as a default, seems like a reasonable choice for now. It will allow us to validate more scenarios. I followed the same pattern as for CCM now, which is sharing the cluster between tests.

testing/src/test/java/io/stargate/it/CQLTest.java
```diff
@@ -308,8 +301,8 @@ public void traceQuery() throws UnknownHostException {
     assertThat(rs.getExecutionInfo().getTracingId()).isNotNull();
 
     QueryTrace trace = rs.getExecutionInfo().getQueryTrace();
-    assertThat(trace.getCoordinatorAddress().getAddress())
-        .isEqualTo(InetAddress.getByName(stargateHost));
+    assertThat(trace.getCoordinatorAddress().getAddress()).isIn(getStargateInetSocketAddresses());
+    ;
```
mpenick, line 305: Extra `;`.

tomekl007: done

stargate-starter/src/main/java/io/stargate/starter/Starter.java
```diff
@@ -190,6 +190,21 @@
...
+  int jmxPort = 7199;
+
+  @Order(value = 17)
+  @Option(name = {"--dynamic_snitch", "Set if the dynamic snitch should wrap the actual snitch."})
+  boolean dynamicSnitch = true;
+
+  @Order(value = 18)
+  @Option(
+      name = {"--disable_mbean_registration", "Whether the mbean registration should be disabled"})
+  boolean disableMBeanRegistration = false;
```
dougwettlaufer, line 206: The `starctl` script needs updated to support these.

testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
```diff
@@ -73,15 +78,43 @@
...
+  public static String getStargateHost(int stargateInstanceNumber) {
+    return stargateHosts.get(stargateInstanceNumber);
+  }
+
+  public static List<String> getStargateHosts() {
+    return stargateHosts;
+  }
+
+  public static List<InetAddress> getStargateInetSocketAddresses() throws UnknownHostException {
+    return Arrays.asList(
```
mpenick, line 112: minor: This is fragile if the number of stargate nodes is changed. Maybe iterate and convert instead?

```java
    return stargateHosts.stream().map(h -> {
      try {
        return InetAddress.getByName(h);
      } catch (UnknownHostException e) {
        throw new RuntimeException(e);
      }
    }).collect(Collectors.toList());
```

dimas-b (changes requested):

testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
dougwettlaufer, line 112: Should we build this list dynamically using `numberOfStargateNodes` ? That way we don't have to worry about this getting out of sync if that number is changed.

testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
```diff
@@ -246,35 +279,76 @@ Object proxy(
...
+            .get(stargateInstanceNumber)
+            .getService(name)
+            .orElseThrow(() -> new AssertionError("Missing " + name));
 
     return (T) proxy(p, p.getClass().getClassLoader(), clazz.getClassLoader(), clazz);
   }
 
+  /**
+   * Returns a proxy facade for a OSGi service loaded in stargate By default, it retrieves the
+   * facase from the 1st starter node.
```
dougwettlaufer, line 298: > facase

Typo?

testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
tomekl007: good catch, done

testing/src/test/java/io/stargate/it/CQLTest.java
```diff
@@ -182,14 +182,7 @@ public void querySystemLocal() throws UnknownHostException {
...
     assertThat(rows).hasNext();
-    assertThat(rows.next().getInetAddress("listen_address"))
-        .isEqualTo(InetAddress.getByName(stargateHost));
-  }
-
-  @Test
-  public void querySystemPeers() {
-    ResultSet rs = session.execute("SELECT * FROM system.peers");
-    assertThat(rs.all()).isEmpty();
+    assertThat(rows.next().getInetAddress("listen_address")).isIn(getStargateInetSocketAddresses());
```
dimas-b, line 185: Should we validate all rows to be an exhaustive list of configured addresses?

tomekl007: system local can contain any address from there, the request can go to any node because of the `LoadBalancingPolicy` so such a test would be more fragile

dimas-b: Also, this test case seems to be a covered by `MultipleStargateInstancesTest`. Why would we have both?

stargate-starter/src/main/java/io/stargate/starter/Starter.java
tomekl007: done in: https://github.com/stargate/stargate/pull/93/commits/d00e650a0f6c2768b3e43ef588fbc354c29f4496 but one condition: `dynamic_snitch` has default `true`. So passing this `--dynamic_snitch` will not change the default. Do we have a clear way to pass the negation?

testing/src/test/java/io/stargate/it/CQLTest.java
dimas-b: but doesn't `system.peers` reflect all Stargate nodes?

testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
```diff
@@ -73,15 +78,43 @@
...
+
   protected final ClusterConnectionInfo backend;
 
+  public static List<Starter> stargateStarters = new ArrayList<>();
+  private static List<String> stargateHosts = new ArrayList<>();
+  public static final Integer numberOfStargateNodes = 3;
+
+  static {
+    for (int i = 1; i <= numberOfStargateNodes; i++) {
+      stargateHosts.add("127.0.0.1" + i);
```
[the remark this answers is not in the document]
tomekl007: done

testing/src/test/java/io/stargate/it/CQLTest.java
dimas-b: oops, sorry, I got confused by the diff - it does apply to `system.local` now - please ignore

testing/src/test/java/io/stargate/it/MultipleStargateInstancesTest.java
```diff
@@ -0,0 +1,144 @@
...
+    // system.peers should have N records (all stargate nodes - 1)
+    assertThat(all.size()).isEqualTo(runningStargateNodes - 1);
+  }
+
+  @Test
+  public void shouldDistributeTrafficUniformly() {
+    // given
+    createKeyspaceAndTable();
+    long totalNumberOfRequests = 300;
+    long numberOfRequestPerNode = totalNumberOfRequests / runningStargateNodes;
```
dimas-b, line 101: nit: would it be more robust from the math perspective to fix `numberOfRequestPerNode` and compute `totalNumberOfRequests = numberOfRequestPerNode * runningStargateNodes`?

testing/src/test/java/io/stargate/it/MultipleStargateInstancesTest.java
```diff
@@ -0,0 +1,144 @@
...
+    // then
+    Collection<Node> nodes = session.getMetadata().getNodes().values();
+    assertThat(nodes.size()).isEqualTo(runningStargateNodes);
+    for (Node n : nodes) {
+      long cqlMessages =
+          ((Timer)
+                  session.getMetrics().get().getNodeMetric(n, DefaultNodeMetric.CQL_MESSAGES).get())
+              .getCount();
+      assertThat(cqlMessages)
+          .isBetween(numberOfRequestPerNode - tolerance, numberOfRequestPerNode + tolerance);
```
dimas-b, line 124: Where does the variance come from?... just wondering :thinking: shouldn't the driver do strict round-robin?

testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
```diff
@@ -73,15 +78,50 @@
...
+
   protected final ClusterConnectionInfo backend;
 
+  public static List<Starter> stargateStarters = new ArrayList<>();
+  private static List<String> stargateHosts = new ArrayList<>();
+  public static final Integer numberOfStargateNodes = 3;
+
+  static {
+    for (int i = 1; i <= numberOfStargateNodes; i++) {
+      int portSuffix = 1 + i;
```
dimas-b, line 91: `portSuffix = 10 + i`? I think we need to reserve a few lower address for `ccm`, which starts at `127.0.0.1` in our setup

testing/src/test/java/io/stargate/it/MultipleStargateInstancesTest.java
tomekl007: I totally agree, unfortunately, there are still sometimes cases when one of the Startate nodes does not start (schema agreement problem, timeout). In such a case this test will pass asserting the traffic amongst two nodes, whereas the fixed approach won't. So I've done a less strict approach to not introduce a non-deterministic test. I don't have more ideas on how to make the Stargate startup more robust, I will welcome any inputs.

dimas-b (changes requested):

testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
dimas-b: Please add a comment about that as well :)

testing/src/test/java/io/stargate/it/MultipleStargateInstancesTest.java
tomekl007: I didn't analyze it more deeply, in a variety of cases it is equal to the exact number but I observed some cases when it was not even. It can be for example retry or speculative execution.

testing/src/test/java/io/stargate/it/MultipleStargateInstancesTest.java
dimas-b: sorry, I do not understand... the numbers should be the same regardless of how well the nodes run, it's just a matter of rounding errors in case the division is not exact.

stargate-starter/src/main/java/io/stargate/starter/Starter.java
dougwettlaufer: In the script if the value is not provided then you pass in `true` to the starter?

stargate-starter/src/main/java/io/stargate/starter/Starter.java
```diff
@@ -274,6 +299,12 @@ void setStargateProperties() {
     System.setProperty("stargate.emulate_dbaas_defaults", emulateDbaasDefaults ? "true" : "false");
     System.setProperty("stargate.developer_mode", String.valueOf(developerMode));
     System.setProperty("stargate.bind_to_listen_address", String.valueOf(bindToListenAddressOnly));
+    System.setProperty("cassandra.jmx.remote.port", String.valueOf(jmxPort));
+    System.setProperty("cassandra.jmx.local.port", String.valueOf(jmxPort));
+    System.setProperty("stargate.dynamic_snitch", String.valueOf(!disableMBeanRegistration));
```
dougwettlaufer, line 304: Shouldn't this be 

```suggestion
    System.setProperty("stargate.dynamic_snitch", String.valueOf(! disableDynamicSnitch));
```

tomekl007:
I run 10 consecutive builds and there is no race condition (https://github.com/stargate/stargate/pull/93/commits) - all builds from `Sep 29, 2020` so this should be ready to merge now

health-checker/src/main/java/io/stargate/health/HealthCheckerActivator.java
```diff
@@ -10,10 +10,10 @@
   private static final Logger log = LoggerFactory.getLogger(HealthCheckerActivator.class);
 
   private BundleContext context;
-  private boolean started;
+  private volatile boolean started;
```
jorgebay, line 13: NIT: its used within a synchronized block so there's no need to use `volatile`.

jorgebay (commented):
Supporting multiple stargate nodes in the tests is an important milestone 💪 !

I've added some comments below, mostly minor stuff.

stargate-starter/src/main/java/io/stargate/starter/Starter.java
```diff
@@ -190,6 +190,21 @@
...
 
+  @Order(value = 16)
+  @Option(
+      name = {"--jmx-port"},
+      description = "The port on which JMX should start")
+  int jmxPort = 7199;
+
+  @Order(value = 17)
+  @Option(name = {"--dynamic_snitch", "Set if the dynamic snitch should wrap the actual snitch."})
+  boolean dynamicSnitch = true;
```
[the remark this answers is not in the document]
tomekl007: those two settings are different, the:
- `disable_dynamic_snitch` - states if the dynamic snitch should wrap the underlying snitch
- `--simple-snitch` - states the type of this underlying snitch

stargate-starter/src/main/java/io/stargate/starter/Starter.java
```diff
@@ -220,6 +236,11 @@ public Starter(
     this.simpleSnitch = isSimpleSnitch;
     this.cqlPort = cqlPort;
     this.watchBundles = false;
+    // bind to listen address only to allow multiple starters to start on the same host
+    this.bindToListenAddressOnly = true;
```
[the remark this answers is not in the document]
tomekl007: default is false:
```
  @Order(value = 15)
  @Option(
      name = {"--bind-to-listen-address"},
      description = "When set, it binds web services to listen address only")
  boolean bindToListenAddressOnly = false;
```
the code that you mentioned is used only in Integration tests - when creating a new stargate instance using the constructor.

stargate-starter/src/main/java/io/stargate/starter/Starter.java
```diff
@@ -190,6 +190,25 @@
...
+      description = "The port on which JMX should start")
+  int jmxPort = 7199;
+
+  @Order(value = 17)
+  @Option(
+      name = {
+        "--disable_dynamic_snitch",
+        "Whether the dynamic snitch should wrap the actual snitch."
+      })
+  boolean disableDynamicSnitch = false;
```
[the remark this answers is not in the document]
tomekl007: see: https://github.com/stargate/stargate/pull/93#discussion_r496505524

stargate-starter/src/main/java/io/stargate/starter/Starter.java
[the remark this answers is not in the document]
jorgebay: Wouldn't it be better to disable wrapping when `--simple-snitch` is set?

stargate-starter/src/main/java/io/stargate/starter/Starter.java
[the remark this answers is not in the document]
jorgebay: oh, snap! thanks for the clarification.

jorgebay (approved):
Looks great!

I posted a suggestion about snitch settings, but it doesn't have to be taken care of in this PR.

stargate-starter/src/main/java/io/stargate/starter/Starter.java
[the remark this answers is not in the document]
tomekl007: when running one instance it is ok to have a dynamic snitch wrapper without `simple-snitch`
So I think those should be set separately

mpenick (approved):
+1

dimas-b (approved):

PR #95, opened 2020-09-16
Fix schema migration for non-member nodes

mpenick:
This will likely fix the schema agreement issues we've been
experiencing. Schema needs to be migrated regardless of the node type.
The previous logic was only pulling the schema if the other node was a
stargate (non-member) node. That was incorrect schema should be pulled
regardless of the node type.

### DIFF
--- a/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
+++ b/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
@@ -16,6 +16,7 @@
 package io.stargate.db.cassandra.impl;
 
 import java.io.IOException;
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.net.UnknownHostException;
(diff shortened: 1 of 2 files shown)


persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -110,6 +113,15 @@ public void initialize(Config config)
         daemon = new CassandraDaemon(true);
 
         DatabaseDescriptor.daemonInitialization(() -> config);
+
+        root = new InternalDataStore();
+        authenticator = new AuthenticatorWrapper(DatabaseDescriptor.getAuthenticator());
+        handler = org.apache.cassandra.service.ClientState.getCQLQueryHandler();
+        interceptor = new DefaultQueryInterceptor();
+        interceptor.initialize();
```
mpenick, line 121: Move this initialization before because if a `IEndpointStateChangeSubscriber` isn't registered before `Gossiper.start()` the events might be missed.

mpenick:
No! `java.lang.IllegalStateException: Failed to reach schema agreement after 10 seconds.`

PR #102, opened 2020-09-17
Remove pretty param

dougwettlaufer:


dimas-b:
Why are we removing the pretty param?

dougwettlaufer:
> Why are we removing the pretty param?

A couple of reasons

1. With any modern tool, formatting json is trivial. Postman and httpie will do it by default, curl can pipe to jq, and in code there's really no benefit.
2. The parameter breaks client generation from Open API Spec yaml files

dimas-b (approved):

PR #105, opened 2020-09-17
Run mvn fmt on release

dougwettlaufer:


.github/workflows/release.yml
```diff
@@ -92,6 +92,7 @@ jobs:
           mvn versions:set -DremoveSnapshot versions:commit
           mvn -B -Dgpg.passphrase=${{ secrets.OSSRH_GPG_SECRET_KEY_PASSWORD }} clean deploy -DskipTests -P deploy && \
           mvn --batch-mode release:update-versions -DautoVersionSubmodules=true versions:commit
+          mvn xml-format:xml-format fmt:format
```
mpenick, line 95: Ha, maybe this is a dumb question. :) What's running format (on a local copy) during CI for?

dougwettlaufer: Right above this we change the version and the format is super finicky so that version change breaks the format. To fix this I unfortunately have to 1) up the version, 2) format, and 3) finally commit and make the PR to prepare for next release.

PR #111, opened 2020-09-21
Add endpoint for token by username

dougwettlaufer:


auth-api/src/main/java/io/stargate/auth/api/AuthResource.java
```diff
@@ -87,29 +90,83 @@ public Response createToken(Secret body) {
...
-    if (body == null) {
+  public Response createToken(Credentials credentials) {
+    if (credentials == null) {
       return Response.status(Response.Status.BAD_REQUEST)
           .entity(new Error("Must provide a body to the request"))
           .build();
     }
 
-    if (body.getUsername() == null || body.getUsername().equals("")) {
+    if (credentials.getUsername() == null || credentials.getUsername().equals("")) {
```
tomekl007, line 100: this logic should be extracted to a separate method because is occurring in multiple places.

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthTableBasedService.java
```diff
@@ -119,29 +145,50 @@ public String createToken(String key, String secret) throws UnauthorizedExceptio
...
+        dataStore
+            .query()
+            .select()
+            .column("role")
+            .from("system_auth", "roles")
+            .where("role", WhereCondition.Predicate.Eq, key)
+            .execute();
+
+    if (resultSet.isEmpty()) {
+      throw new RuntimeException(String.format("Provided username %s is incorrect", key));
```
tomekl007, line 161: it can be flattened to one if:
```
Row row = resultSet.one();
if (row != null && row.has("role") ) {
   row.getString("role");
}else {
  throw new RuntimeException(String.format("Provided username %s is incorrect", key));
}

File: testing/src/test/java/io/stargate/it/http/AuthApiTest.java [hunk 3]
```diff
@@ -0,0 +1,181 @@
...
+            String.format("%s:8081/v1/auth/token/generate", host),
+            "",
+            HttpStatus.SC_BAD_REQUEST);
+
+    Error error = objectMapper.readValue(body, Error.class);
+    assertThat(error.getDescription()).isEqualTo("Must provide a body to the request");
+  }
+
+  @Test
+  public void auth() throws IOException {
```
tomekl007, line 91: `auth` I think that that the test name should be more descriptive, i.e:
`performAuthSuccesfully`.

testing/src/test/java/io/stargate/it/http/AuthApiTest.java
```diff
@@ -0,0 +1,181 @@
...
+            String.format("%s:8081/v1/auth/usernametoken", host),
+            objectMapper.writeValueAsString(new UsernameCredentials("cassandra")),
+            HttpStatus.SC_CREATED);
+
+    AuthTokenResponse authTokenResponse = objectMapper.readValue(body, AuthTokenResponse.class);
+    assertThat(authTokenResponse.getAuthToken()).isNotNull();
+  }
+
+  @Test
+  public void authUsernameToken_BadUsername() throws IOException {
```
tomekl007, line 156: do we have a convention for test names?
I saw in other places he `camelCaseConvention`, but here there are mixed with `_`

auth-api/src/main/java/io/stargate/auth/api/AuthResource.java
```diff
@@ -32,9 +33,12 @@
 
 @Path("/v1")
 public class AuthResource {
+
   private static final Logger logger = LoggerFactory.getLogger(AuthResource.class);
 
   final AuthenticationService authService;
+  private static final boolean shouldEnableUsernameToken =
+      Boolean.parseBoolean(System.getProperty("stargate.auth_api_enable_username_token", "false"));
```
tomekl007, line 41: where this property is set? I see it only set in the test:
https://github.com/stargate/stargate/pull/111/files#diff-2c65e0c4892da78e91b8b4d3d357e25aR406
but not in the actual `Starter` logic.

auth-api/src/main/java/io/stargate/auth/api/AuthResource.java
```diff
@@ -87,29 +90,83 @@ public Response createToken(Secret body) {
...
+
+    return Response.status(Response.Status.CREATED)
+        .entity(new AuthTokenResponse().authToken(token))
+        .build();
+  }
+
+  /**
+   * NOTE: This method is intended to be used in conjunction with another authentication service.
+   * For example, the request has already been authenticated through a proxy but a token is still
+   * needed to make requests to Stargate.
```
dimas-b, line 134: This sounds like a kerberos use case to me.

auth-api/src/main/java/io/stargate/auth/api/AuthResource.java
```diff
@@ -87,29 +90,83 @@ public Response createToken(Secret body) {
...
+   *
+   * @param usernameCredentials A username to generate a token for IFF the username exists in the
+   *     database.
+   * @return On success will return a token to be used with later requests to Stargate.
+   */
+  @POST
+  @Path("/auth/usernametoken")
+  @Consumes(MediaType.APPLICATION_JSON)
+  @Produces(MediaType.APPLICATION_JSON)
+  public Response createTokenFromUsername(UsernameCredentials usernameCredentials) {
```
dimas-b, line 148: How do we make sure ordinary users cannot abuse this method? Is it not exposed to the outside traffic? (sorry, I do not know how to check)

auth-api/src/main/java/io/stargate/auth/api/AuthResource.java
dougwettlaufer: Right. It's just a system property

auth-api/src/main/java/io/stargate/auth/api/AuthResource.java
dougwettlaufer: These are different objects. This uses Credentials but down below it's UsernameCredentials

auth-api/src/main/java/io/stargate/auth/api/AuthResource.java
dougwettlaufer: Hence the note up above. This should be used in a secured manner such as if you already have auth behind a proxy but you still need to generate a token for some username. Also why it's behind a flag that's off by default.

dimas-b (approved):

testing/src/test/java/io/stargate/it/http/AuthApiTest.java
dougwettlaufer: These are just following the names of the endpoints they're calling. In this case `/auth`.

testing/src/test/java/io/stargate/it/http/AuthApiTest.java
dougwettlaufer: No clue. I've been using `baseTestName_TestCondition`

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthTableBasedService.java
dougwettlaufer: This throws a "NoSuchElement" error

testing/src/test/java/io/stargate/it/http/AuthApiTest.java
EricBorczuk: I should probably change my tests to match :) since I'm going through schemaless tests soon I can standardize

auth-api/src/main/java/io/stargate/auth/model/Credentials.java
```diff
@@ -43,10 +44,21 @@ public void setPassword(String password) {
...
-    if (this == o) return true;
-    if (o == null || getClass() != o.getClass()) return false;
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
     Credentials that = (Credentials) o;
     return Objects.equals(username, that.username) && Objects.equals(password, that.password);
```
EricBorczuk, line 63: does Objects.equals do the same thing as `StringUtils.equals` (tests equality, null-safe)?

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthTableBasedService.java
```diff
@@ -119,29 +145,50 @@ public String createToken(String key, String secret) throws UnauthorizedExceptio
...
             .column("salted_hash")
             .from("system_auth", "roles")
-            .where("role", WhereCondition.Predicate.Eq, key)
+            .where("role", WhereCondition.Predicate.Eq, hashedPassword)
             .execute();
 
     if (resultSet.isEmpty()) {
       throw new RuntimeException(
-          String.format("Provided username %s and/or password are incorrect", key));
+          String.format("Provided username %s and/or password are incorrect", hashedPassword));
```
EricBorczuk, line 185: Is it intentional to expose the hashedPassword in these error messages?

auth-api/src/main/java/io/stargate/auth/api/AuthResource.java
```diff
@@ -87,29 +90,83 @@ public Response createToken(Secret body) {
...
     }
 
-    // TODO: [doug] 2020-06-18, Thu, 0:42 handle potential errors from authService
     String token;
     try {
-      token = authService.createToken(body.getUsername(), body.getPassword());
+      token = authService.createToken(credentials.getUsername(), credentials.getPassword());
+    } catch (UnauthorizedException e) {
+      return Response.status(Response.Status.UNAUTHORIZED)
+          .entity(new Error("Failed to create token: " + e.getMessage()))
```
EricBorczuk, line 117: Is there any situation where `e.getMessage()` would have sensitive data that you wouldn't want to return back?

EricBorczuk (approved):
Looks good! I had a few questions that are mostly for my own knowledge

auth-api/src/main/java/io/stargate/auth/model/Credentials.java
dougwettlaufer: Objects.equals was an IntelliJ recommendation for null safety.

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthTableBasedService.java
dougwettlaufer: That variable was actually misnamed, it's the username. I'll fix that.

auth-api/src/main/java/io/stargate/auth/api/AuthResource.java
dougwettlaufer: No because it's just going to contain what the authService passes it. So as long as we don't throw any exceptions with sensitive data we're fine.

EricBorczuk (approved):

Issue #50, opened 2020-09-15
Support running integration tests on macOS


blerghh:
We should support running integration tests on macOS w/ or w/o test containers (i.e. use ccm directly for the backend cluster).

PR #114, opened 2020-09-22
Use ccm for managing test cluster during IT runs

dimas-b:
* The cluster is setup at the test class init time for now.
  This is mainly to deal with JUnit 4 and 5 code uniformly,
  but also the OSGi container is global for now and depends
  on the backend cluster to be ready.

* Removed C*-related test parameters - this can be controlled
  by java system properties during test runs now.


dimas-b:
Related CCM PR: https://github.com/riptano/ccm/pull/719

@pcmanus : How do I request reviews in CCM? The repo does not allow me to add reviewers :thinking:

testing/src/test/java/io/stargate/it/CQLTest.java
```diff
@@ -111,7 +107,7 @@ public void setup() {
     table = testName;
   }
 
-  @Before
+  @BeforeEach
```
tomekl007, line 110: I am not sure if this is necessary to create a stop session after and before every test suite

testing/src/test/java/io/stargate/it/BaseStorageIntegrationTest.java
```diff
@@ -0,0 +1,86 @@
...
+        CcmBridge.builder()
+            .withCassandraConfiguration("cluster_name", CLUSTER_NAME)
+            .withNodes(clusterNodes)
+            .build();
+
+    if (!EXTERNAL_BACKEND) {
+      ccm.create();
+      ccm.start();
+
+      Runtime.getRuntime()
```
tomekl007, line 51: In driver we have:
```
  Runtime.getRuntime()
        .addShutdownHook(
            new Thread(
                () -> {
                  try {
                    ccmBridge.remove();
                  } catch (Exception e) {
                    // silently remove as may have already been removed.
                  }
                }));
```
so maybe add this `try catch` block

tomekl007 (commented):
This is a nice improvement, I've run the `mvn clean verify` and it passes on my mac.
I wonder how hard it will be to migrate the java-driver mechanism of JUnit rules, for example see this:
- https://github.com/datastax/java-driver/tree/9f0d89799a8a1e4cd1022dd7c43333924c36a648/test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm
- https://github.com/datastax/java-driver/blob/9f0d89799a8a1e4cd1022dd7c43333924c36a648/test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CustomCcmRule.java
Please not the fluency of this approach, so for example to create a 3 node cluster:
```
@ClassRule
public static final CustomCcmRule CCM_RULE = CustomCcmRule.builder().withNodes(3).build();
```
or to retrieve the shared one-node cluster:
```
private static CcmRule ccmRule = CcmRule.getInstance();
```
Maybe we could do it in a separate ticket?

testing/src/test/java/io/stargate/it/BaseStorageIntegrationTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    ccm =
+        CcmBridge.builder()
+            .withCassandraConfiguration("cluster_name", CLUSTER_NAME)
+            .withNodes(clusterNodes)
+            .build();
+
+    if (!EXTERNAL_BACKEND) {
+      ccm.create();
+      ccm.start();
```
tomekl007, line 49: we will start now the ccm for every test case, right now we don't have many ITs but in the future it may impact build time. We may consider creating one shared instance, but also give the possiblity for more advanced setups, see:
https://github.com/stargate/stargate/pull/114#pullrequestreview-494231979

.github/workflows/ci.yml
```diff
@@ -24,9 +24,18 @@ jobs:
...
+          fetch-depth: 1
+          path: ccm
+      - name: Install CCM Dependencies
+        run: pip install --user setuptools pyYaml six tailer ldap3 lxml
       - name: run tests and analyze
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
-        run: mvn -B verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar --file pom.xml
+        run: env PATH=$PWD/ccm:$PATH mvn -B verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar --file pom.xml
```
jorgebay, line 40: Shouldn't we use `sudo ./setup.py install` ?

testing/src/test/java/io/stargate/it/CQLTest.java
jorgebay: +1, this should be `@BeforeAll` and `@AfterAll`.

jorgebay (approved):
I've added some comments below but it's such an improvement from the current situation it may be easier to merge now and tackle individual remarks later.

testing/src/test/java/io/stargate/it/BaseStorageIntegrationTest.java
dimas-b: sure. will do.

testing/src/test/java/io/stargate/it/BaseStorageIntegrationTest.java
dimas-b: Actually, it starts one ccm cluster per test JVM. I'm planning to combine all `test` runs in one JVM in a follow-up PR... did not want to make too many changes at once :)

.github/workflows/ci.yml
dimas-b: Good point. I'll give it a try... I basically copied python commands from some Jenkins job I had before, but this approach looks cleaner.

testing/src/test/java/io/stargate/it/CQLTest.java
dimas-b: It used to be `@BeforeAll`, but was [changed](https://github.com/stargate/stargate/commit/210c929d0024530be9e73d59f8cebbc312386ee0#diff-f2cee11f058387da4ff5d1b14fe11887L91-L92). I'm not sure why. @mpenick : WDYT?

dimas-b:
@tomekl007 : Regarding `@ClassRule` at this point the main difficulty is that we have a mixture of JUnit 4 and 5 tests. I spent some time on this, but could not figure out a way to satisfy both call paths :) I think we will be migrating all tests to JUnit 5 so we can use whatever is the natural way in JUnit 5 to do rules (extensions, IIRC)

.github/workflows/ci.yml
```diff
@@ -24,9 +24,18 @@ jobs:
           path: ~/.m2
           key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-m2
+      - name: Clone CCM
+        uses: actions/checkout@v2
+        with:
+          repository: dimas-b/ccm
```
dougwettlaufer, line 30: Why the special repo? Do we think this'll be removed by the time the PR is merged?

testing/src/test/java/io/stargate/it/CQLTest.java
dougwettlaufer: Was that this change?
https://github.com/stargate/stargate/pull/42#discussion_r491104263

testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
```diff
@@ -283,129 +242,28 @@ Object proxy(
...
-    if (backendContainer != null && !backendContainer.getDockerImageName().equals(dockerImage)) {
-      logger.info("Docker image changed {} {}", dockerImage, backendContainer.getDockerImageName());
-
-      if (stargateStarter != null) stargateStarter.stop();
-
-      backendContainer.stop();
-
-      backendContainer = null;
-      stargateStarter = null;
+  @BeforeEach
```
dougwettlaufer, line 246: `@Before` and `@BeforeEach`?

dougwettlaufer (dismissed):

testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
dimas-b: That is required because some of the sub-classes use JUnit 4, others JUnit 5... we'll sort it out in a follow-up PR.

.github/workflows/ci.yml
dimas-b: yeah, because my CCM PR is not merged yet. I'll fix before merging.

testing/src/test/java/io/stargate/it/CQLTest.java
dimas-b: Yes, I think it was #42

testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
dougwettlaufer: Got it, thanks

testing/src/test/java/io/stargate/it/CQLTest.java
dimas-b: looks like we can fix this now that all the connection info is static.

dimas-b: I tried doing this, but unfortunately, `CQLTest` also relies on non-static `enableAuth`, which in turn required starting the OSGi container after class-level initializers... I'd rather file a separate ticket to refactor this since it's not directly related to `ccm`

.github/workflows/ci.yml
dimas-b: Updated and force pushed - seems to work fine. thanks for the tip :)

testing/src/test/java/io/stargate/it/BaseStorageIntegrationTest.java
dimas-b: Updated and force pushed

tomekl007 (dismissed):

.github/workflows/ci.yml
dimas-b: riptano/ccm#719 has been merged, we're using that repo in CI now.

dimas-b:
I did a few forced pushes while sorting out CCM issues. That should be stable now (and the related CCM PR riptano/ccm/pull/719 merged).

@dougwettlaufer : I also made some changes to `GraphqlTest` that was not part of this PR before - please take a look.

dougwettlaufer (approved):

tomekl007 (approved):

jorgebay (approved):

dougwettlaufer:
Fixes #50

PR #117, opened 2020-09-24
REST/v1 with compound partition key /rows/{pk} of mixed types leads to 400

dougwettlaufer:


testing/src/test/java/io/stargate/it/http/RestApiTest.java
```diff
@@ -362,6 +362,45 @@ public void getRowWithClustering() throws IOException {
     assertThat(rowResponse.getRows().get(0).get("firstName")).isEqualTo("Jane");
   }
 
+  @Test
+  public void getRowWithMixedClustering() throws IOException {
```
tlasica, line 366: nit: I think mixed types are essential, so maybe `getRowWithMixedClusteringKeyTypes` ?

restapi/src/main/java/io/stargate/web/resources/RowResource.java
```diff
@@ -410,10 +416,8 @@ private static Object filterToValue(Object val, String column, Table tableData)
         .collect(Collectors.toList());
   }
 
-  private List<String> idFromPath(PathSegment id) {
-    MultivaluedMap<String, String> matrixParameters = id.getMatrixParameters();
-    List<String> values = new ArrayList<>(matrixParameters.keySet());
-    values.add(0, id.getPath());
-    return values;
+  private List<String> idFromPath(String path) {
+    String id = path.substring(path.lastIndexOf("/") + 1);
```
tlasica, line 420: I assume that it is not possible to get `/keyspaces/x/tables/y/1;2;3/` here?

testing/src/test/java/io/stargate/it/http/RestApiTest.java
dougwettlaufer: As in renaming just renaming the method?

restapi/src/main/java/io/stargate/web/resources/RowResource.java
dougwettlaufer: The path for this is `/keyspaces/ks1/tables/tbl1/rows/1;2;3`. For `/keyspaces/ks1/tables/tbl1` that would be picked up by `getAll()` and not go through here.

dimas-b (approved):

restapi/src/main/java/io/stargate/web/resources/RowResource.java
olim7t: What happens if one of the PK components is a string that contains the character `/`? Would it have to be URL-encoded, or is there a chance that it might cause a bug here?

dougwettlaufer: It should be URL encoded.

restapi/src/main/java/io/stargate/web/resources/RowResource.java
```diff
@@ -410,10 +416,8 @@ private static Object filterToValue(Object val, String column, Table tableData)
...
+    return Arrays.asList(id.split(";"));
```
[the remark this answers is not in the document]
olim7t: Since I had the same question about `/`, I think the answer is you need to URL-encode it.

I tested both with the branch:
```
$ curl -X GET 'http://127.0.0.2:8082/v1/keyspaces/test/tables/foo/rows/a%2Fb;1' -H 'Content-Type: application/json' -H"x-cassandra-token: $TOKEN"
{"count":1,"rows":[{"ks1":"a/b","ks2":1}]}

$ curl -X GET 'http://127.0.0.2:8082/v1/keyspaces/test/tables/foo/rows/a%3Bb;1' -H 'Content-Type: application/json' -H"x-cassandra-token: $TOKEN"
{"count":1,"rows":[{"ks1":"a;b","ks2":1}]}
```

olim7t (approved):

PR #146, opened 2020-09-24
Use JUnit 5 for testing

dimas-b:
* Also move restapi tests to their proper location in the tree

* Run Integration Tests in `testing` in one JVM to save on
  restarting the backend cluster.

* Also run `testing` suites in parallel. Being backbox tests
  they should be able to run concurrently.

* Note: testcontainers are removed because they transitively
  pull in JUnit 4 (they are no longer used anyway).


restapi/src/test/java/io/stargate/web/resources/WhereParserTest.java
```diff
@@ -0,0 +1,267 @@
+/*
```
tomekl007, line 1: why does it show this file as changed?

dimas-b: I moved it and it got reformatted. It was previously "invisible" to the build.

dougwettlaufer (approved):

PR #147, opened 2020-09-24
Enforce license headers in the build

olim7t:
Currently this only checks production Java files.

It could be extended to any of the following if deemed necessary: test Java, XML, properties... Or any file type really, the plugin has flexible support for custom comment delimiters.

### DIFF
--- a/persistence-api/src/main/java/io/stargate/db/datastore/DataStore.java
+++ b/persistence-api/src/main/java/io/stargate/db/datastore/DataStore.java
@@ -1,5 +1,5 @@
 /*
- * Copyright DataStax, Inc. and/or The Stargate Authors
+ * Copyright The Stargate Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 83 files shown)


persistence-api/src/main/java/io/stargate/db/datastore/DataStore.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright DataStax, Inc. and/or The Stargate Authors
+ * Copyright The Stargate Authors
```
csplinter, line 2: For files that we took from DSE, we need to have 
```
Copyright DataStax, Inc. and/or The Stargate Authors
```

olim7t: OK I missed that, I'll add another license set.

dimas-b: Can this be configured / overridden in sub-modules?.. might be worth keeping the DataStax info local to the modules that use DS code.

olim7t:
Closing, I don't see a way to make this work.

We use 3 different headers, with no way to automatically tell which one each file should get: there are discrepancies even within packages.

PR #151, opened 2020-09-25
Cleanup of schema handling in persistence-api

pcmanus:
This is a first step towards making `Persistence` the only interface for persistence extensions to implement, and making `DataStore` be on top of `Persistence`. For now, this move the handling of schema from `DataStore` to `Persistence` (`DataStore` still has methods to access that schema, so users of `DataStore` are unaffected, but those methods just delegate to the underlying `Persistence`.

The PR also fix some issue regarding the check for schema agreement (in short, the existing `SchemaTool` facility that was in `persistence-common` was using the `Gossiper` imported by `persistence-common`, which was wrong).


pcmanus:
For info, the changes necessary for this on persistence extensions is at https://github.com/stargate/persistence-extensions/pull/19.

dougwettlaufer (approved):

mpenick (approved):
+1 looks good.

mpenick:
minor issue: There are some missing license headers.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -77,29 +84,61 @@
...
   private static final Logger logger = LoggerFactory.getLogger(CassandraPersistence.class);
 
-  private DataStore root;
   private CassandraDaemon daemon;
   private Authenticator authenticator;
   private QueryHandler handler;
   private QueryInterceptor interceptor;
 
+  // C* listener that ensures that our Stargate schema remains up-to-date with the internal C* one.
+  private MigrationListener migrationListener;
```
olim7t, line 105: Not strictly related to this PR, but what's the concurrency policy for this class? It initializes those non-final, non-volatile fields outside of the constructor, and CassandraPersistenceActivator doesn't use any synchronization when it calls `initialize`.

We should at least make the fields volatile, that will work if there won't be concurrent calls to `initializePersistence` (and I think that's a fair assumption).

olim7t (dismissed):

dougwettlaufer (approved):

PR #152, opened 2020-09-25
Add core module

jorgebay:
Adds a core module that has no dependencies to other modules.

It registers the metrics service to be used across any module.


core/src/main/java/io/stargate/core/CoreActivator.java
```diff
@@ -0,0 +1,28 @@
...
+import org.osgi.framework.ServiceEvent;
+import org.osgi.framework.ServiceListener;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CoreActivator implements BundleActivator, ServiceListener {
+  private static final Logger logger = LoggerFactory.getLogger(CoreActivator.class);
+
+  @Override
+  public void start(BundleContext context) {
```
tomekl007, line 16: what is a long term plan for this module? Would it have only a metrics module? If yes, then I think it should be called just MetricsActivator.

core/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+        <groupId>io.stargate</groupId>
+        <version>0.0.10-SNAPSHOT</version>
+    </parent>
+    <groupId>io.stargate.core</groupId>
+    <artifactId>core</artifactId>
+    <properties>
+        <dropwizard.version>2.0.10</dropwizard.version>
+    </properties>
+    <dependencies>
+        <dependency>
```
tomekl007, line 17: I am not sure if we should include `slf4j` in the core? What if other modules that are using it provide dependencies to the logging library? All sub-module will need to exclude any logging library, right?

core/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+                        <Bundle-Name>Core</Bundle-Name>
+                        <Bundle-Description>Core API for Stargate project</Bundle-Description>
+                        <Bundle-SymbolicName>io.stargate.core</Bundle-SymbolicName>
+                        <Bundle-Activator>io.stargate.core.CoreActivator</Bundle-Activator>
+                        <Import-Package><![CDATA[
+                          com.codahale.metrics,
+                          org.osgi.framework,
+                        ]]></Import-Package>
+                        <Export-Package><![CDATA[
+                          io.stargate.core.*,
```
tomekl007, line 85: do we need to export whole `io.stargate.core.*,`? maybe the metrics `package` is enough?

health-checker/src/main/java/io/stargate/health/HealthCheckerActivator.java
```diff
@@ -19,8 +15,13 @@ public void start(BundleContext context) {
     this.context = context;
     log.info("Starting healthchecker....");
 
-    Metrics metrics = new MetricsImpl();
-    context.registerService(Metrics.class, metrics, null);
+    ServiceReference<?> metricsReference = context.getServiceReference(Metrics.class.getName());
+
+    if (metricsReference == null) {
```
tomekl007, line 20: is there a possibility that the metrics module cannot be loaded but this is expected? For example, if the client doesn't want to include metrics on their classpath? Maybe in this situation, we shouldn't fail fast?

jorgebay: Metrics gathering is always enabled.
Health checks depends on metrics being gathered.

core/src/main/java/io/stargate/core/CoreActivator.java
jorgebay: afaik, there can be one activator per module, so all core services will be registered here. For now, only the metrics...

core/pom.xml
jorgebay: This makes the whole package to be available to be use by other bundles.
For now, only a single interface will be exported.

core/pom.xml
jorgebay: Without this dependency, core module won't activate: 

```
Starting bundle io.stargate.core
Unexpected error: Activator start error in bundle io.stargate.core [8].
org.osgi.framework.BundleException: Activator start error in bundle io.stargate.core [8].
	at org.apache.felix.framework.Felix.activateBundle(Felix.java:2452)
	at org.apache.felix.framework.Felix.startBundle(Felix.java:2308)
	at org.apache.felix.framework.BundleImpl.start(BundleImpl.java:998)
	at org.apache.felix.framework.BundleImpl.start(BundleImpl.java:984)
	at io.stargate.starter.Starter.start(Starter.java:368)
	at io.stargate.starter.Starter.main(Starter.java:477)
Caused by: java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory
	at io.stargate.core.CoreActivator.<clinit>(CoreActivator.java:13)
```

dimas-b: Should the scope be `provided` perhaps? I believe `slf4j` is part of the system class path in our OSGi container (if I'm not mistaken)

cql/pom.xml
```diff
@@ -141,10 +147,10 @@
                             !org.slf4j.spi,
                             !sun.misc,
                             !org.eclipse.jetty.security,
```
dimas-b, line 149: a bit tangential to this PR, but in general, would it make sense to import only specific required packages and then have a wildcard `!*` at the end of the imports section?

core/pom.xml
jorgebay: I'll give it a try.

cql/pom.xml
jorgebay: I _think_ that by default when specifying `<Import-Package>`, it won't include `*`, so negating here might be unnecessary, but 🤷

health-checker/src/main/java/io/stargate/health/HealthCheckerActivator.java
dimas-b: I'm not sure OSGi guarantees any particular order of bundle activation. So even though we start the metrics service unconditionally, it may not be available yet when this method runs.... I think :thinking:

dimas-b: I believe I saw something like a helper lib to manage inter-service dependencies on the web, but I lost the tab for it :)

core/pom.xml
jorgebay: With `provided`, I also get:

```
Starting bundle io.stargate.core
Unexpected error: Activator start error in bundle io.stargate.core [8].
org.osgi.framework.BundleException: Activator start error in bundle io.stargate.core [8].
	at org.apache.felix.framework.Felix.activateBundle(Felix.java:2452)
	at org.apache.felix.framework.Felix.startBundle(Felix.java:2308)
	at org.apache.felix.framework.BundleImpl.start(BundleImpl.java:998)
	at org.apache.felix.framework.BundleImpl.start(BundleImpl.java:984)
	at io.stargate.starter.Starter.start(Starter.java:368)
	at io.stargate.starter.Starter.main(Starter.java:477)
Caused by: java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory
	at io.stargate.core.CoreActivator.<clinit>(CoreActivator.java:13)
```

health-checker/src/main/java/io/stargate/health/HealthCheckerActivator.java
jorgebay: oh, interesting...

I was thinking OSGi used the import packages as a way to reorder activation. We should find a way then...

dimas-b: something like this perhaps? https://www.osgi.org/wp-content/uploads/AutoServDependencyMgmt_byHall_Cervantes.pdf

dimas-b: also: https://felix.apache.org/documentation/subprojects/apache-felix-dependency-manager/guides/bundles-and-dependencies.html

core/pom.xml
dimas-b: hmmm, interesting, I guess we can leave it for now, IIRC we have a dedicated issue for improving logging

health-checker/src/main/java/io/stargate/health/HealthCheckerActivator.java
dimas-b: AFAIK, imports are just for managing class loading :thinking: but I'm not an OSGi expert :)

jorgebay: oh my, no easy way to set (ordered) dependencies between service instances, I'm hating OSGi more each day... 

I think the shorter path for now would be to use `serviceChanged()`... I'll adapt the PR.

dimas-b: Check out `ServiceTracker`, cf. https://github.com/stargate/api-extensions/blob/select-by-pk/sql/server/src/main/java/io/stargate/api/sql/server/SqlActivator.java#L47

dougwettlaufer: Bundle activation will occur as dependencies are resolved. So persistence for example will activate first since it has no dependencies but something like graphql will activate later because it waits on persistence -> auth -> graphql.

dougwettlaufer (dismissed):
Looks good. Slightly apprehensive about having a "core" module as that tends to be a recipe for a catch-all but I think we should have some sufficiently crosscutting dependencies like metrics (maybe logging?) to warrant this.

health-checker/src/main/java/io/stargate/health/HealthCheckerActivator.java
jorgebay: Thanks for the clarification @dougwettlaufer !

jorgebay:
I've addressed the issue related to service registration.

I've extracted a small utility function listening to registered services `serviceChanged()`.

> Slightly apprehensive about having a "core" module

I agree that in general it should be avoided... In this case, given that we have a bunch modules on top of each other (and osgi adds some complexity), I think it will enable us implement crosscutting functionality and utilities faster (compared to one per functionality), but in any case we should keep an eye to avoid making it a catch-all :)

cql/pom.xml
mpenick: I don't think we need the negating. We should probably just import what we use.

core/src/main/java/io/stargate/core/BundleUtils.java
```diff
@@ -0,0 +1,21 @@
...
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceEvent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class BundleUtils {
+  private static final Logger logger = LoggerFactory.getLogger(CoreActivator.class);
+
+  /** Gets the instance of the newly registered service. */
+  public static Object getRegisteredService(BundleContext context, ServiceEvent serviceEvent) {
```
dimas-b, line 12: Could we make it `public static <T> T getRegisteredService(BundleContext context, ServiceEvent serviceEvent, Class<T> serviceType)` to avoid type checks at calling sites?

Also, if the caller does not like the object's type it will have to "unget" the service to avoid dangling references (I think OSGi requires that).

health-checker/src/main/java/io/stargate/health/HealthCheckerActivator.java
```diff
@@ -36,24 +49,10 @@ public void stop(BundleContext context) {}
...
-        break;
-      case (ServiceEvent.MODIFIED):
-        // TODO: [doug] 2020-06-15, Mon, 12:58 do something here...
-        log.info("Service of type " + objectClass[0] + " modified.");
-        break;
-      default:
-        break;
+    Object service = BundleUtils.getRegisteredService(context, serviceEvent);
+
+    if (service instanceof Metrics) {
```
dimas-b, line 54: cf. https://github.com/stargate/stargate/pull/152/files#diff-5c75ac240c063eb5580c85cce8f7cc3cR12

dimas-b:
LGTM in general. My only concern is ungetting ignored services in places like https://github.com/stargate/stargate/pull/152/files#diff-8098ebff610571479f528519ef20d1bbR54

core/src/main/java/io/stargate/core/BundleUtils.java
jorgebay: tbh, I extracted the method from `persistence-cassandra-3.11`, `restapi`, ...

Good point, I've modified the utility method. Is that what you had in mind?

dimas-b (approved):

health-checker/pom.xml
```diff
@@ -180,12 +186,12 @@
...
+              io.dropwizard.*,
+              org.osgi.framework,
+              com.codahale.metrics,
+              io.stargate.core.*
             ]]></Import-Package>
-            <Export-Package><![CDATA[
-              io.stargate.health.metrics.api,
-              com.codahale.metrics
-            ]]></Export-Package>
+            <Export-Package>io.stargate.health</Export-Package>
```
dougwettlaufer, line 194: Actually this no longer needs to export anything since it's now just a top level http service.

dougwettlaufer (approved):

health-checker/pom.xml
jorgebay: good catch! I'll fix it.

dougwettlaufer (approved):

PR #156, opened 2020-09-25
Convert BaseStorageIntegrationTest to a JUnit 5 extension

dimas-b:
* Better control over test lifecycle.

* Does not require extending any particular base test class.


dimas-b:
Let's merge this part now to enable repo reorg. I will do another PR for follow-up test improvements

testing/src/test/java/io/stargate/it/storage/ExternalStorage.java
```diff
@@ -0,0 +1,243 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(ExternalStorage.class);
+
+  private static final String CCM_VERSION = "ccm.version";
+  private static final boolean EXTERNAL_BACKEND =
+      Boolean.getBoolean("stargate.test.backend.use.external");
+
+  private static final AtomicBoolean executing = new AtomicBoolean();
+
+  private static final int clusterNodes = Integer.getInteger("stargate.test.backend.nodes", 1);
```
dougwettlaufer, line 45: How flexible will this be for the tests @tomekl007 is working on where he needs multiple nodes?

testing/src/test/java/io/stargate/it/storage/ExternalStorage.java
```diff
@@ -0,0 +1,243 @@
...
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ParameterContext;
+import org.junit.jupiter.api.extension.ParameterResolutionException;
+import org.junit.jupiter.api.extension.ParameterResolver;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** JUnit 5 extension for tests that need a backend database cluster managed by {@code ccm}. */
+public class ExternalStorage implements BeforeAllCallback, AfterAllCallback, ParameterResolver {
```
dougwettlaufer, line 35: In the future how hard will it be to include non-ccm managed storage?

dougwettlaufer (approved):
Looks good just a few minor questions.

testing/src/test/java/io/stargate/it/storage/ExternalStorage.java
dimas-b: Ah, I actually missed this one - it's supposed to come from `ClusterSpec`.

My thinking is that each class will define how many nodes it needs. We could support per-method specs too.

Do we foresee setting cluster size from outside of the test code scope (e.g. in maven properties / run config)?

testing/src/test/java/io/stargate/it/storage/ExternalStorage.java
dimas-b: Just update this class to support it.

dimas-b: We could use different `ClusterSpec` -like annotations too.

dimas-b:
Merging now to unblock repo reorg. I'll fix the cluster size handling in a follow-up RP.

testing/src/test/java/io/stargate/it/storage/ExternalStorage.java
dougwettlaufer: No, I'd imagine cluster sizes will be determined internally by the tests themselves.

testing/src/test/java/io/stargate/it/storage/ExternalStorage.java
dougwettlaufer: Got it, thanks.

PR #159, opened 2020-09-28
[ON HOLD] Basic CDC Service implementation

jorgebay:
It contains a basic CDC Service implementation and producer interface with activator.

Ideally, the service will be invoked by listeners that would be added to the Persistence API. It's still to be defined how this integration will be done.

For now, this serves as a skeleton to allow further collaboration on a basic set of interfaces.


persistence-api/pom.xml
```diff
@@ -16,6 +16,17 @@
     </repository>
   </repositories>
   <dependencies>
+    <dependency>
+      <groupId>org.apache.cassandra</groupId>
+      <artifactId>cassandra-all</artifactId>
+      <version>4.0-alpha4</version>
```
jorgebay, line 22: This is required by to use C*'s `org.apache.cassandra.metrics.DefaultNameFactory`, similar to how we are using in the `cql` module.

I'll create a ticket to get rid of this dependency and `cql`'s.

persistence-api/src/main/java/io/stargate/db/DbActivator.java
```diff
@@ -0,0 +1,71 @@
...
+import org.slf4j.LoggerFactory;
+
+public class DbActivator implements BundleActivator, ServiceListener {
+  private static final Logger logger = LoggerFactory.getLogger(DbActivator.class);
+  public CDCService cdcService;
+  private boolean started;
+  private BundleContext context;
+
+  @Override
+  public void start(BundleContext context) throws InvalidSyntaxException {
```
tomekl007, line 18: I would add synchronized to all methods: those will be called very rarely so overhead would be low.

persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java
```diff
@@ -0,0 +1,142 @@
...
+    } catch (Exception e) {
+      logger.info("There was an issue releasing resources of CDC Producer", e);
+    }
+  }
+
+  @VisibleForTesting
+  interface CDCConfig {
+    boolean isTrackedByCDC(MutationEvent mutation);
+
+    long getLatencyErrorMs();
```
tomekl007, line 140: nit, the name of this method: `getLatencyErrorMs`  does not seem clear to me. Can we change it or add docs?

persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java
```diff
@@ -0,0 +1,142 @@
...
+
+  private static final CompletableFuture<Void> completedFuture =
+      CompletableFuture.completedFuture(null);
+  private static final CompletableFuture<Void> unhealthyFuture = new CompletableFuture<>();
+  private static final Logger logger = LoggerFactory.getLogger(CDCService.class);
+
+  private final CDCProducer producer;
+  private final CDCHealthChecker healthChecker;
+  private final CDCConfig config;
+  private final ScheduledExecutorService timeoutScheduler =
```
tomekl007, line 40: `1` thread for this tread pool maybe not enough. There will be high traffic on `publish` but `orTimeout` will have only one thread. The default `CompletableFuture` uses a fork-join pool that has more threads so maybe we should use a similar number here.

persistence-api/src/main/java/io/stargate/db/DbActivator.java
```diff
@@ -0,0 +1,71 @@
...
+
+import com.codahale.metrics.MetricRegistry;
+import io.stargate.db.cdc.CDCProducer;
+import io.stargate.db.cdc.CDCService;
+import io.stargate.db.cdc.CDCServiceImpl;
+import org.osgi.framework.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DbActivator implements BundleActivator, ServiceListener {
```
tomekl007, line 11: Question for the future, how it will look like if there will be > 1 underlying `CDCProducers` that we need to support and register (i.e. kafka and kinesis).

persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java
```diff
@@ -0,0 +1,142 @@
...
+import org.slf4j.LoggerFactory;
+
+public final class CDCServiceImpl implements CDCService {
+
+  private static final CompletableFuture<Void> completedFuture =
+      CompletableFuture.completedFuture(null);
+  private static final CompletableFuture<Void> unhealthyFuture = new CompletableFuture<>();
+  private static final Logger logger = LoggerFactory.getLogger(CDCService.class);
+
+  private final CDCProducer producer;
```
tomekl007, line 37: Similar to
>Question for the future, how it will look like if there will be > 1 underlying `CDCProducers` that we need to support and register (i.e. Kafka and kinesis).

Maybe we should handle `List<CDCProducer>` from the beginning?

persistence-api/src/main/java/org/apache/cassandra/stargate/schema/TableMetadata.java
```diff
@@ -0,0 +1,34 @@
...
+  List<ColumnMetadata> getClusteringKeys();
+
+  /** A collection of static and regular columns. */
+  List<ColumnMetadata> getColumns();
+
+  /**
+   * Gets an object that is used as identity for this instance. This object will be used to
+   * represent this version of the table metadata for caching.
+   */
+  Object getIdentity();
```
tomekl007, line 33: Isn't it possible to create a class that represents this entity? (with the table name, schema name)

persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.cassandra.metrics.MetricNameFactory;
+
+public final class CDCMetrics {
+  public static final CDCMetrics instance = new CDCMetrics();
+
+  private static final MetricNameFactory factory = new DefaultNameFactory("CDC");
+
+  private boolean initialized = false;
+  private MetricRegistry metricRegistry;
+
```
tomekl007, line 35: I think it will be useful to have two additional `Meter` with the sum of send and failures 
`Meter successfullySendMessages`
`Meter errors`

persistence-api/src/main/java/org/apache/cassandra/stargate/schema/ColumnMetadata.java
```diff
@@ -0,0 +1,29 @@
...
+    PARTITION_KEY,
+    CLUSTERING,
+    REGULAR,
+    STATIC
+  }
+
+  Kind getKind();
+
+  /** Case-sensitive column name. */
+  String getName();
```
tomekl007, line 26: Maybe we could create a `CqlIdentifier` type - the same that is used in the driver? It is handling quotations, case-sensitivity, etc. I know that adding this dependency may be problematic, but maybe we can port this portion of code to stargate repo?

persistence-api/src/test/java/io/stargate/db/cdc/CDCServiceTest.java
```diff
@@ -0,0 +1,109 @@
...
+    assertThat(runtimeEx).hasCause(ex);
+  }
+
+  @Test
+  public void shouldTimeoutAndReportFailure() {
+    ServiceBuilder builder =
+        new ServiceBuilder().withProducerFuture(new CompletableFuture<>()).withTimeout(20);
+    CDCService service = builder.build();
+
+    ExecutionException runtimeEx =
```
tomekl007, line 68: shouldn't this be wrapped in `await`? (https://github.com/awaitility/awaitility)

persistence-api/src/main/java/org/apache/cassandra/stargate/db/DeleteEvent.java
```diff
@@ -0,0 +1,10 @@
+package org.apache.cassandra.stargate.db;
+
+import java.util.List;
+
+/** Represents a delete of one or multiple rows. */
+public interface DeleteEvent extends MutationEvent {
```
tomekl007, line 6: I understand that this will be handling delete of the whole row, right?
What about the deletion of one field in a row? Will it be handled as the `RowUpdateEvent` with the cell having `null` value? (denoting clear/delete)?

jorgebay: > Will it be handled as the RowUpdateEvent with the cell having null value? (denoting clear/delete)?

yes!

`DeleteEvent` can affect one or multiple rows depending on the amount of keys specified.

tomekl007: Does it mean that it will be possible to get only PKs without CK for delete? Such a solution is problematic because the Kafka key should contain all PKs and Clustering Keys. 
If we omit any of those keys, sending null as any Kafka field value will mean that the delete event will be sent to a different Kafka partition than an update. This will break the time ordering because data for the same record will not land on the same Kafka partition. 
Debezium treats all PKs and Clustering Keys as required for that reason:
https://debezium.io/documentation/reference/connectors/cassandra.html#cassandra-change-events-value

jorgebay: ok, so we should leave out partition delete (when CKs exist) and range deletes for now.

jorgebay: Now that I think of it, I think the record key should only be composed by the partition keys (and not the CKs).

tomekl007: This may be problematic - Kafka will have performance degradation because of the partition skew

jorgebay: Having partition keys would be easier to reason about and it would equate partitioning on both Kafka and C*: skew on one side would mean skew on the other.

tomekl007: agree, I see one more small disadvantage of not having clustering keys in the `kafka.key`: it will be a bit harder for clients of CDC to locate the record in db based only on the `kafka.key` but maybe that is not a problem. 
I'll try to rework and see how it will look like; so to sum up:
- for `RowUpdateEvent` the `getPartitionKeys` and `getClusteringKeys` will always return all columns. PKs will be in `kafka.key` whereas CKs in `kafka.value`
- for the `DeleteEvent` CKs may or may not be present. The value schema seems already refecting that.
The most important bit for the client to react accordingly will be (`op: u` or `op: d`)

persistence-api/src/main/java/org/apache/cassandra/stargate/db/MutationEvent.java
```diff
@@ -0,0 +1,12 @@
+package org.apache.cassandra.stargate.db;
+
+import org.apache.cassandra.stargate.schema.TableMetadata;
+
+/** Represents a change made in the database. */
+public interface MutationEvent {
```
tomekl007, line 6: can we introduce an intermidiete interface between  `MutationEvent` -> `DeleteEvent` AND `RowUpdateEvent` smth like:
```
class PartitionedMutationEvent{
   List<CellValue> getPartitionKeys();
    List<CellValue> getClusteringKeys();
}
```
?
Those methods are in both implementations, and having a common interface will make the CDC producer logic easier

persistence-api/src/main/java/org/apache/cassandra/stargate/schema/TableMetadata.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * <p>Backend agnostic representation equivalent to {@code
+ * org.apache.cassandra.schema.TableMetadata}
+ */
+public interface TableMetadata {
+  /** Gets the table id. */
+  UUID getId();
+
+  /** Gets the case-sensitive name of the keyspace. */
+  String getKeyspace();
```
tomekl007, line 17: Maybe we could annotate return types of those methods with `@NonNull` and `@Nullable` to allow the client's code to reason about return type and guard against `null` if needed?  We did that in the java-driver and it was very useful for our clients.

persistence-api/src/main/java/io/stargate/db/DbActivator.java
jorgebay: I think there should be only 1 cdc producer. CDC is part of the hot write path and it fails the request when the producer fails, so there shouldn't be multiple producers.

I think there's the idea of multiple streams integrations, like push this a stream A, push this to stream B, but its not necessarily related to CDC. We may share the same persistence primitives but not much (i.e. it will not block the write request).

dougwettlaufer:


```diff

Issues
======
- Added 14
           

Complexity increasing per file
==============================
- persistence-api/src/test/java/io/stargate/db/cdc/TableSchemaManagerTest.java  1
- persistence-api/src/test/java/io/stargate/db/cdc/CDCServiceTest.java  1
- persistence-api/src/test/java/io/stargate/db/cdc/CDCTestUtil.java  2
- persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java  4
- persistence-api/src/main/java/io/stargate/db/DbActivator.java  2
- persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java  3
- persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java  2
- persistence-api/src/main/java/org/apache/cassandra/stargate/schema/CQLType.java  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6263920&bid=20396688)

persistence-api/src/main/java/io/stargate/db/DbActivator.java
jorgebay: Done!

dougwettlaufer:


```diff

Issues
======
- Added 14
           

Complexity increasing per file
==============================
- persistence-api/src/main/java/io/stargate/db/DbActivator.java  4
- persistence-api/src/test/java/io/stargate/db/cdc/CDCTestUtil.java  2
- persistence-api/src/test/java/io/stargate/db/cdc/CDCServiceTest.java  1
- persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java  3
- persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java  2
- persistence-api/src/main/java/org/apache/cassandra/stargate/schema/CQLType.java  4
- persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java  4
- persistence-api/src/test/java/io/stargate/db/cdc/TableSchemaManagerTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6263920&bid=20396688)

persistence-api/src/main/java/io/stargate/db/DbActivator.java
```diff
@@ -0,0 +1,71 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(DbActivator.class);
+  public CDCService cdcService;
+  private boolean started;
```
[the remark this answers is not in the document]
jorgebay: It shouldn't as it's used from synchronized blocks.

persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java
jorgebay: My understanding is that openjdk's `orTimeout()` uses 1 thread: https://github.com/openjdk/jdk/blob/928da494a853c164452c96d01318c94d115ca946/src/java.base/share/classes/java/util/concurrent/CompletableFuture.java#L2825-L2829

We usually shouldn't care about the timeout accuracy and as long as the caller is async all the way, the invoker thread should do little effort.

persistence-api/src/main/java/org/apache/cassandra/stargate/schema/TableMetadata.java
jorgebay: I think this can be resolved once we know what the persistence api would look like.

The idea here is to provide a caching mechanism for `TableMetadata`. In C*, `TableMetadata` is immutable, so it should be straightforward, but if we wrap it each time in a new instance for the backend agnostic API, we loose that immutability. If we use the wrapping approach, returning the private instance on `getIdentity()` would solve that.

persistence-api/src/main/java/org/apache/cassandra/stargate/schema/ColumnMetadata.java
jorgebay: I think that as an end user, it would be easy to reason a case sensitive value: `users_table` or `UserTable`.

persistence-api/src/test/java/io/stargate/db/cdc/CDCServiceTest.java
jorgebay: I have no experience with `awaitability` library.

Using a new dependency for something that can be solved by `.get()` and unwrap exception seems unnecessary to me. We can create an utility method if we see the need.

dougwettlaufer:


```diff

Issues
======
- Added 14
           

Complexity increasing per file
==============================
- persistence-api/src/test/java/io/stargate/db/cdc/TableSchemaManagerTest.java  1
- persistence-api/src/main/java/io/stargate/db/DbActivator.java  4
- persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java  2
- persistence-api/src/test/java/io/stargate/db/cdc/CDCTestUtil.java  2
- persistence-api/src/main/java/org/apache/cassandra/stargate/schema/CQLType.java  4
- persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java  4
- persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java  3
- persistence-api/src/test/java/io/stargate/db/cdc/CDCServiceTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6263920&bid=20396688)

persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java
jorgebay: I've changed it to a more clear name.

persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java
jorgebay: I'll add one for successful.

persistence-api/src/main/java/org/apache/cassandra/stargate/schema/TableMetadata.java
jorgebay: We could have the discussion outside of this PR.

persistence-api/src/test/java/io/stargate/db/cdc/CDCServiceTest.java
tomekl007: We are using this library for testing async code in java-driver and it has proven to be very robust. But in this situation you are right, `get()` is better

persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java
tomekl007: `errors` are also very important, so I think we should have both. For the kafka-sink-connector https://github.com/datastax/kafka-sink/ we have metrics for both and they are very useful

persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java
```diff
@@ -0,0 +1,88 @@
...
+  private Counter producerMessagesInFlight;
+  private Meter producerFailures;
+  private Meter producerTimedOut;
+  private Timer producerLatency;
+
+  private CDCMetrics() {}
+
```
tomekl007, line 39: I think it will be useful to have low granularity metrics for types of `MutationEvents`:
- `deletion-event-count`
- `update-event-count`
etc.
and `schema-changed-count` when there `createTableSchemaAsync` called

dougwettlaufer:


```diff

Issues
======
- Added 14
           

Complexity increasing per file
==============================
- persistence-api/src/test/java/io/stargate/db/cdc/TableSchemaManagerTest.java  1
- persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java  3
- persistence-api/src/main/java/org/apache/cassandra/stargate/schema/CQLType.java  4
- persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java  4
- persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java  2
- persistence-api/src/test/java/io/stargate/db/cdc/CDCServiceTest.java  1
- persistence-api/src/main/java/io/stargate/db/DbActivator.java  4
- persistence-api/src/test/java/io/stargate/db/cdc/CDCTestUtil.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6263920&bid=20396688)

persistence-api/pom.xml
```diff
@@ -16,6 +16,33 @@
     </repository>
   </repositories>
   <dependencies>
+    <dependency>
+      <groupId>io.stargate.core</groupId>
+      <artifactId>core</artifactId>
+      <version>0.0.12-SNAPSHOT</version>
```
dougwettlaufer, line 22:
```suggestion
      <version>0.0.15-SNAPSHOT</version>
```

dougwettlaufer (dismissed):

dougwettlaufer:


```diff

Issues
======
- Added 14
           

Complexity increasing per file
==============================
- persistence-api/src/test/java/io/stargate/db/cdc/CDCServiceTest.java  1
- persistence-api/src/main/java/org/apache/cassandra/stargate/schema/CQLType.java  4
- persistence-api/src/test/java/io/stargate/db/cdc/CDCTestUtil.java  2
- persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java  3
- persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java  2
- persistence-api/src/test/java/io/stargate/db/cdc/TableSchemaManagerTest.java  1
- persistence-api/src/main/java/io/stargate/db/DbActivator.java  4
- persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6263920&bid=20396688)

dougwettlaufer:


```diff

Issues
======
- Added 14
           

Complexity increasing per file
==============================
- persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java  4
- persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java  3
- persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java  2
- persistence-api/src/main/java/org/apache/cassandra/stargate/schema/CQLType.java  4
- persistence-api/src/test/java/io/stargate/db/cdc/CDCTestUtil.java  2
- persistence-api/src/test/java/io/stargate/db/cdc/TableSchemaManagerTest.java  1
- persistence-api/src/main/java/io/stargate/db/DbActivator.java  4
- persistence-api/src/test/java/io/stargate/db/cdc/CDCServiceTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6263920&bid=20396688)

dougwettlaufer:


```diff

Issues
======
- Added 14
           

Complexity increasing per file
==============================
- persistence-api/src/test/java/io/stargate/db/cdc/TableSchemaManagerTest.java  1
- persistence-api/src/main/java/org/apache/cassandra/stargate/schema/CQLType.java  4
- persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java  2
- persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java  4
- persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java  3
- persistence-api/src/main/java/io/stargate/db/DbActivator.java  4
- persistence-api/src/test/java/io/stargate/db/cdc/CDCServiceTest.java  1
- persistence-api/src/test/java/io/stargate/db/cdc/CDCTestUtil.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6263920&bid=20396688)

dougwettlaufer:


```diff

Issues
======
- Added 14
           

Complexity increasing per file
==============================
- persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java  4
- persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java  2
- persistence-api/src/main/java/io/stargate/db/DbActivator.java  4
- persistence-api/src/test/java/io/stargate/db/cdc/CDCServiceTest.java  1
- persistence-api/src/test/java/io/stargate/db/cdc/CDCTestUtil.java  2
- persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java  3
- persistence-api/src/main/java/org/apache/cassandra/stargate/schema/CQLType.java  4
- persistence-api/src/test/java/io/stargate/db/cdc/TableSchemaManagerTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6263920&bid=20396688)

dougwettlaufer (approved):

dougwettlaufer:


```diff

Issues
======
- Added 14
           

Complexity increasing per file
==============================
- persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java  4
- persistence-api/src/main/java/io/stargate/db/DbActivator.java  4
- persistence-api/src/test/java/io/stargate/db/cdc/TableSchemaManagerTest.java  1
- persistence-api/src/test/java/io/stargate/db/cdc/CDCTestUtil.java  2
- persistence-api/src/test/java/io/stargate/db/cdc/CDCServiceTest.java  1
- persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java  2
- persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java  3
- persistence-api/src/main/java/org/apache/cassandra/stargate/schema/CQLType.java  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6263920&bid=20396688)

dougwettlaufer:


```diff

Issues
======
- Added 4
           

Complexity increasing per file
==============================
- persistence-api/src/main/java/io/stargate/db/DbActivator.java  4
- persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java  4
- persistence-api/src/main/java/org/apache/cassandra/stargate/schema/CQLType.java  4
- persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java  3
- persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6263920&bid=20396688)

dougwettlaufer:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- persistence-api/src/main/java/org/apache/cassandra/stargate/schema/CQLType.java  4
- persistence-api/src/main/java/io/stargate/db/DbActivator.java  4
- persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java  2
- persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java  3
- persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6263920&bid=20396688)

persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java
```diff
@@ -0,0 +1,143 @@
...
+import com.datastax.oss.driver.shaded.guava.common.util.concurrent.ThreadFactoryBuilder;
+import io.stargate.core.metrics.api.Metrics;
+import io.stargate.db.metrics.CDCMetrics;
+import java.util.concurrent.*;
+import org.apache.cassandra.stargate.db.MutationEvent;
+import org.apache.cassandra.stargate.exceptions.CDCWriteException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class CDCServiceImpl implements CDCService {
```
dimas-b, line 31: Could it be preferable to move the "impl" service to different (new?) module so that the "API" module defines just java classes, but does not start services automatically?

persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java
```diff
@@ -0,0 +1,143 @@
...
+    try {
+      final ScheduledFuture<Boolean> scheduledTimeout =
+          timeoutScheduler.schedule(
+              () ->
+                  f.completeExceptionally(
+                      new TimeoutException(String.format("Timed out after %sms", timeoutMs))),
+              timeoutMs,
+              MILLISECONDS);
+
+      f.thenAccept(r -> scheduledTimeout.cancel(false));
```
dimas-b, line 118: shouldn't we return the result of `thenAccept`?

persistence-api/src/main/java/io/stargate/db/cdc/CDCProducer.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import org.apache.cassandra.stargate.db.MutationEvent;
+
+/** Represents a publisher of mutations events for Change Data Capture (CDC). */
+public interface CDCProducer {
+  /**
+   * Initializes the {@link CDCProducer}. It will be invoked once in the lifetime of the instance
+   * before any calls to send.
```
dimas-b, line 26: `send` -> `publish`?

persistence-api/src/main/java/io/stargate/db/cdc/CDCProducer.java
```diff
@@ -0,0 +1,40 @@
...
+   *
+   * @param options The passthrough options.
+   */
+  CompletableFuture<Void> init(Map<String, Object> options);
+
+  /**
+   * Publishes the mutation event. It will only be invoked for mutations that should be tracked by
+   * the CDC.
```
dimas-b, line 34: Is it worth mentioning that `publish` should be non-blocking?

persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java
```diff
@@ -0,0 +1,143 @@
...
+      MetricRegistry registry) {
+    this.producer = producer;
+    this.healthChecker = healthChecker;
+    this.config = config;
+
+    CDCMetrics.instance.init(registry);
+  }
+
+  @Override
+  public CompletableFuture<Void> publish(MutationEvent mutation) throws CDCWriteException {
```
dimas-b, line 69: Do we actually throw `CDCWriteException` directly? Would it not be wrapped in the `CompletableFuture`?

persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java
```diff
@@ -0,0 +1,128 @@
...
+  public CompletableFuture<Void> publish(MutationEvent mutation) {
+    TableSchemaManager schemaManager =
+        tableSchemaManager.computeIfAbsent(
+            mutation.getTable().getId(), k -> new TableSchemaManager(this::createTableSchemaAsync));
+
+    return schemaManager
+        .ensureCreated(mutation.getTable())
+        // Invoke send() with the version that later affected this table,
+        // not the most recent version number.
+        .thenCompose(originalSchemaVersion -> send(mutation));
```
dimas-b, line 71: I guess this can lead to mutation events piling up in the `CompletableFuture` dependency graph if the schema creation is taking a long time and incoming traffic is heavy... should we perhaps limit this somehow? Should we perhaps skip mutations until the schema has been created?

persistence-api/src/test/java/io/stargate/db/cdc/CDCTestUtil.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.stream.Collectors;
+
+public final class CDCTestUtil {
```
dimas-b, line 13: Is this class used? I could not find any references to it in this PR :thinking:

dougwettlaufer:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java  3
- persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java  2
- persistence-api/src/main/java/org/apache/cassandra/stargate/schema/CQLType.java  4
- persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java  4
- persistence-api/src/main/java/io/stargate/db/DbActivator.java  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6263920&bid=20396688)

persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java
jorgebay: I think the service should be started automatically as it's possible to enable/disable cdc per table.
Ideally, a user will set up a CDC producer, deploy and later in time enable a table.

persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java
jorgebay: Good catch, it will be wrapped in the future (no sync throw). It's a leftover from the C* patch, I'll remove it.

persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java
jorgebay: We shouldn't skip mutations because from the user pov, if cdc is enabled on a table it should either succeed or throw.

We can add mitigations in the future, some of them we already discussed w/ @tomekl007:
- Set creation timeouts.
- Reject after n queued mutations.
- Try to warmup the schema creation when a new stargate node is added (to minimize creations).

I think this mitigations are out of scope for the first iteration of CDC.

persistence-api/src/test/java/io/stargate/db/cdc/CDCTestUtil.java
jorgebay: It's used only by `TableSchemaManagerTest`, it's also a leftover from the port.

I'll move those methods to the test that is actually using it :)

persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java
jorgebay: I think the caller of this method shouldn't care about the result of `scheduledTimeout.cancel()`. But if cancelling in the background feels strange, I can change it.

persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java
jorgebay: About moving the service to a new module, I think we can discuss it after the query objects + persistence wrapper (C2-587) is merged. I think CDC service implementation should live wherever the persistence wrapper lives.

dougwettlaufer:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java  3
- persistence-api/src/main/java/io/stargate/db/DbActivator.java  4
- persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java  4
- persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java  2
- persistence-api/src/main/java/org/apache/cassandra/stargate/schema/CQLType.java  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6263920&bid=20396688)

persistence-api/src/test/java/io/stargate/db/cdc/CDCTestUtil.java
jorgebay: I've addressed most of the issues from your review.

dougwettlaufer:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- persistence-api/src/main/java/io/stargate/db/metrics/CDCMetrics.java  2
- persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java  4
- persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java  3
- persistence-api/src/main/java/org/apache/cassandra/stargate/schema/CQLType.java  4
- persistence-api/src/main/java/io/stargate/db/DbActivator.java  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6263920&bid=20396688)

persistence-api/src/test/java/io/stargate/db/cdc/TableSchemaManagerTest.java
```diff
@@ -0,0 +1,125 @@
...
+    when(t.getIdentity()).thenReturn(identity);
+    return t;
+  }
+
+  public static <T> List<T> invokeParallel(Callable<T> task, int times) {
+    return invokeParallel(task, times, times);
+  }
+
+  public static <T> List<T> invokeParallel(Callable<T> task, int times, int nThreads) {
+    ExecutorService executor = Executors.newFixedThreadPool(nThreads);
```
tomekl007, line 104: Creating a new executor for every call to `invokeParallel` is not a good pattern. We are risking creating too many threads. Maybe in a test, it is not a problem 🤷

persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java
dimas-b: good point :+1:

persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java
dimas-b: sgtm

dimas-b (approved):

persistence-api/src/main/java/io/stargate/db/cdc/CDCServiceImpl.java
```diff
@@ -0,0 +1,188 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(CDCService.class);
+
+  private final CDCProducer producer;
+  private final CDCHealthChecker healthChecker;
+  private final CDCConfig config;
+  private final ScheduledExecutorService timeoutScheduler =
+      Executors.newScheduledThreadPool(
+          1, new ThreadFactoryBuilder().setDaemon(true).setNameFormat("failAfter-%d").build());
+
+  static {
```
tomekl007, line 45: why this is in static block

persistence-api/pom.xml
```diff
@@ -99,9 +126,15 @@
             <Bundle-Name>Persistence</Bundle-Name>
             <Bundle-Description>DB component for project</Bundle-Description>
             <Bundle-SymbolicName>io.stargate.db</Bundle-SymbolicName>
-            <Import-Package>org.osgi.framework</Import-Package>
+            <Bundle-Activator>io.stargate.db.DbActivator</Bundle-Activator>
```
tomekl007, line 129: Why this is called `DdActivator` and not for example `CDCActivator`?

jorgebay: Because it's the activator for the `persistence-api` module.

jorgebay:
I've pushed a change to wrap the producer error in `CDCWriteException`.

PR #160, opened 2020-09-28
Wait for schema migration to occur during startup

mpenick:
Added property configuration to make this configurable; however, it
should wait greater than the twice migration manager's delay value.

### DIFF
--- a/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
+++ b/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
@@ -101,10 +101,7 @@ public class CassandraPersistence
    * live token owner doesn't become live within MigrationManager.MIGRATION_DELAY_IN_MS.
    */
   private static final int STARTUP_DELAY_MS =
-      Integer.parseInt(
-          System.getProperty(
-              "stargate.startup_delay_ms",
-              Integer.toString(3 * MigrationManager.MIGRATION_DELAY_IN_MS)));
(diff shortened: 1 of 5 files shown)


mpenick:
Related persistence ext. changes: https://github.com/stargate/persistence-extensions/pull/21

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -96,6 +96,16 @@
         ViewDefinition> {
   private static final Logger logger = LoggerFactory.getLogger(CassandraPersistence.class);
 
+  /*
+   * Initial schema migration can take greater than 2 * MigrationManager.MIGRATION_DELAY_IN_MS if a
+   * live token owner doesn't become live within MigrationManager.MIGRATION_DELAY_IN_MS.
+   */
+  private static final int STARTUP_DELAY_MS =
+      Integer.parseInt(
+          System.getProperty(
```
dimas-b, line 105: Would it be simpler to use `Integer.getInt(name, default)`?

dimas-b (dismissed):

dougwettlaufer (approved):

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
mpenick: TIL, thanks!

tomekl007 (approved):

PR #161, opened 2020-09-28
Don't build iterable from stream with `#iterator`

pcmanus:
Some call was building an iterable from a stream using the
`<stream>#iterator`. This is dodgy because such iterable can in fact
only be used once or will throw.

In this case, the `MaterializedView#create` method was iterating twice
on its `columns` argument which had been created just like that (once
to check a precondition, the 2nd time to truly use the columns).

The patch avoid this "pattern" by building a true iterable (a list) in
those cases.

### DIFF
--- a/persistence-common/src/main/java/io/stargate/db/datastore/common/AbstractCassandraSchemaConverter.java
+++ b/persistence-common/src/main/java/io/stargate/db/datastore/common/AbstractCassandraSchemaConverter.java
@@ -1,5 +1,6 @@
 package io.stargate.db.datastore.common;
 
+import com.google.common.collect.Iterables;
 import io.stargate.db.schema.CollectionIndexingType;
 import io.stargate.db.schema.Column;
 import io.stargate.db.schema.ImmutableCollectionIndexingType;
@@ -103,7 +104,7 @@ public abstract class AbstractCassandraSchemaConverter<K, T, C, U, I, V> {
   protected abstract boolean isBaseTableOf(T table, V view);
 
   public Schema convertCassandraSchema(Iterable<K> cassandraKeyspaces) {
-    return Schema.create(Streams.of(cassandraKeyspaces).map(this::convertKeyspace)::iterator);
+    return Schema.create(Iterables.transform(cassandraKeyspaces, this::convertKeyspace));
   }
 
   private Keyspace convertKeyspace(K keyspace) {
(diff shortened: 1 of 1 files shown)


dougwettlaufer (approved):

persistence-common/src/main/java/io/stargate/db/datastore/common/AbstractCassandraSchemaConverter.java
```diff
@@ -112,8 +113,8 @@ private Keyspace convertKeyspace(K keyspace) {
     Stream<UserDefinedType> userDefinedTypes = convertUserTypes(name, userTypes(keyspace));
     return Keyspace.create(
         name,
-        tables::iterator,
-        userDefinedTypes::iterator,
+        tables.collect(Collectors.toList()),
```
tomekl007, line 116: To make it even less error-prone, I suggest to not use `Stream` as a variable at all, so this code:
```
    String name = keyspaceName(keyspace);
    Stream<Table> tables = convertTables(name, tables(keyspace), views(keyspace));
    Stream<UserDefinedType> userDefinedTypes = convertUserTypes(name, userTypes(keyspace));
    return Keyspace.create(
        name,
        tables.collect(Collectors.toList()),
        userDefinedTypes.collect(Collectors.toList()),
        replicationOptions(keyspace),
        Optional.of(usesDurableWrites(keyspace)));
  }
```
would be:
```
    String name = keyspaceName(keyspace);
    List<Table> tables = convertTables(name, tables(keyspace), views(keyspace)).collect(Collectors.toList());
    List<UserDefinedType> userDefinedTypes = convertUserTypes(name, userTypes(keyspace)).collect(Collectors.toList());
    return Keyspace.create(
        name,
        tables,
        userDefinedTypes,
        replicationOptions(keyspace),
        Optional.of(usesDurableWrites(keyspace)));
  }
```
having the latter solution, we won't risk that someone else will pass this Stream as an argument in the future.
(same for other occurrences of this pattern)

tomekl007 (changes requested):

dimas-b (approved):

persistence-common/src/main/java/io/stargate/db/datastore/common/AbstractCassandraSchemaConverter.java
pcmanus: Well, it's not necessary wrong to use streams as arguments, depends what's you're trying to do. The care is only when you need to convert a stream to an iterable, and I could make the counter-argument that the current code does the conversion exactly when it's needed. I think an argument can be made that it's more self-documented as it is than if when-we-convert is detached from why-we-need-to-convert.

tomekl007: sure, up to you :) But in this case, we are not reusing `Stream` anywhere (not chaining any subsequent operations) so maybe it is not needed to keep the type.

tomekl007 (approved):

PR #165, opened 2020-09-28
Add GraphQL createKeyspace mutation (fixes #155)

### LINKED ISSUE #155: Allow users to createKeyspace with GraphQL
Users need to be able to create keyspaces via the GraphQL API w/o having to cross over into another API (ie CQL).

This should be configurable such that a user can disable this functionality at runtime in the case that they want to restrict allowing this capability on a cluster.

olim7t:
TODO:
- [ ] add a config option to disable the feature(? or use permissions as suggested by Doug)
- [x] also support durable_writes option
- [x] check IT execution (currently running into a CCM issue in my workspace)


graphqlapi/README.md
```diff
@@ -60,8 +60,12 @@ create a keyspace by executing:
...
   createKeyspace(
     name:"library", # The name of your keyspace
-    # Controls how your data is replicated,
-    dcs: { name:"dc1", replicas: 1 }  # Use at least 3 replicas in production
+    # Controls how your data is replicated (accepts the same key/value options as the
+    # `CREATE KEYSPACE` statement in CQL.
+    replication: [
+      { key: "class", value: "SimpleStrategy" },
+      { key: "replication_factor", value: "1" }
+    ]
```
olim7t, line 68: The feature was briefly mentioned here but not implemented.

I changed the syntax for replication option: CQL can in theory accept any arbitrary map of string options, so I think we should reflect that.

dougwettlaufer: Should this follow the same pattern as REST v2?

dougwettlaufer (dismissed):
Looks good just a small comment on syntax consistency with REST.

graphqlapi/README.md
olim7t: It can.

The downside of that pattern is that it's pretty rigid: it limits the choice to SimpleStrategy and NetworkTopologyStrategy, with a predetermined set of options. If new options or new strategies get added, they won't be supported.

I agree that these two defaults should be good for most cases, but in principle it bothers me to artificially limit what's available. If we already do it in other places, then maybe it's a more global discussion to have.

olim7t: Case in point: [transient replication](https://cassandra.apache.org/blog/2018/12/03/introducing-transient-replication.html) (Cassandra 4) introduces a new format for the per-DC value of NetworkTopologyStrategy.
```
{'class' : 'NetworkTopologyStrategy', 'DC1' : '3/1'};
```
REST v2 doesn't support it, it expects a single integer.

mpenick: Why are we exposing these options through graphql/rest at all? :) Isn't the point of these APIs to simplify using C*?  I'm wondering if this shouldn't be defaulted to a sane value and be a server-side configuration option.

olim7t: I'm not opposed to the idea, but it's a fine line between simplifying usage and hiding useful options. Someone who does have specific replication options in mind will get frustrated when they can't find how to set them via GraphQL, then they need to realize that it's a server-side setting, and figure out how to change it.

As I said that's a more general debate, for now I'll reflect the REST v2 options.

dougwettlaufer (approved):

PR #167, opened 2020-09-29
Fix building of QueryOptions ignoring some options

pcmanus:
Some of the native protocol query options (mainly names of bind markers
and client-provided timestamps) where not passed down properly to the
persistence layers.

And indeed, handling them requires reflection as the involved classes
and constructores are not currently public in C*/DSE.

This patch adds the required reflection magic to pass those options
properly. We also ensure the `QueryOptions#prepare` method is called
on all path, so named bound values are properly re-ordered.

### DIFF
--- a/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
+++ b/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
@@ -338,7 +338,7 @@ public class CassandraPersistence
               throw new PreparedQueryNotFoundException(id);
             }
 
-            // Please note that this need to happen _before_ the beginTraceExecute, because when
+            // Please note that this needs to happen _before_ the beginTraceExecute, because when
             // we add bound values to the trace, we rely on the values having been re-ordered by
             // the following prepare (if named values were used that is).
             internalOptions.prepare(prepared.boundNames);
--- a/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
+++ b/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
@@ -207,8 +207,7 @@ public class CassandraPersistence
 
   @Override
   public DataStore newDataStore(
-      QueryState<org.apache.cassandra.service.QueryState> state,
-      QueryOptions<org.apache.cassandra.service.ClientState> queryOptions) {
+      QueryState<org.apache.cassandra.service.QueryState> state, QueryOptions queryOptions) {
     return new InternalDataStore(
         this, Conversion.toInternal(state), Conversion.toInternal(queryOptions));
   }
@@ -285,7 +284,9 @@ public class CassandraPersistence
               beginTraceQuery(cql, state, options);
             }
 
-            CQLStatement statement = QueryProcessor.parseStatement(cql, internalState).statement;
+            ParsedStatement.Prepared prepared = QueryProcessor.parseStatement(cql, internalState);
+            internalOptions.prepare(prepared.boundNames);
+            CQLStatement statement = prepared.statement;
 
             Result result =
                 interceptor.interceptQuery(
@@ -337,9 +338,15 @@ public class CassandraPersistence
               throw new PreparedQueryNotFoundException(id);
             }
 
+            // Please note that this need to happen _before_ the beginTraceExecute, because when
+            // we add bound values to the trace, we rely on the values having been re-ordered by
+            // the following prepare (if named values were used that is).
+            internalOptions.prepare(prepared.boundNames);
+
             if (internalState.traceNextQuery()) {
               internalState.createTracingSession(customPayload);
-              beginTraceExecute(prepared, state, options, internalOptions.getProtocolVersion());
+              beginTraceExecute(
+                  prepared, state, internalOptions, internalOptions.getProtocolVersion());
             }
 
             CQLStatement statement = prepared.statement;
@@ -552,7 +559,7 @@ public class CassandraPersistence
   private void beginTraceExecute(
       ParsedStatement.Prepared prepared,
       QueryState state,
-      QueryOptions options,
+      org.apache.cassandra.cql3.QueryOptions options,
       ProtocolVersion version) {
     ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
     if (options.getPageSize() > 0) {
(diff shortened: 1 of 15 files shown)


persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
```diff
@@ -74,6 +76,54 @@
...
+  // DefaultQueryOptions(ConsistencyLevel consistency, List<ByteBuffer> values, boolean
+  // skipMetadata, QueryOptions.SpecificOptions options, ProtocolVersion protocolVersion)
+  private static final Constructor<?> defaultOptionsCtor;
+  // OptionsWithNames(QueryOptions.DefaultQueryOptions wrapped, List<String> names)
+  private static final Constructor<?> optionsWithNameCtor;
+
+  static {
+    try {
+      Class<?> defaultOptionsClass =
+          Class.forName("org.apache.cassandra.cql3.QueryOptions$DefaultQueryOptions");
```
dimas-b, line 94: nit: would it be possible to declare this as `Class<org.apache.cassandra.cql3.QueryOptions>` to avoid type casts when we invoke this contructor?

pcmanus: Well, unfortunately no. The point here is that the `QueryOptions.DefaultQueryOptions` which we want to construct is not visible.

dimas-b (approved):

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -337,9 +338,15 @@ public Authenticator getAuthenticator() {
               throw new PreparedQueryNotFoundException(id);
             }
 
+            // Please note that this need to happen _before_ the beginTraceExecute, because when
```
mpenick, line 341: typo: "needs"

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
```diff
@@ -86,17 +136,57 @@
...
-    // TODO(mpenick): timestamp and nowInSeconds?
-    return org.apache.cassandra.cql3.QueryOptions.create(
+    return createOptions(
         toInternal(options.getConsistency()),
         options.getValues(),
+        options.getNames(),
         options.skipMetadata(),
         options.getPageSize(),
-        PagingState.deserialize(options.getPagingState(), protocolVersion),
+        // Note that PagingState.deserialize modify its input, so we duplicate to avoid nasty
```
mpenick, line 148: typo: "modify" --> "modifies"

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
```diff
@@ -86,17 +136,57 @@
...
+      if (boundNames != null) {
+        options =
+            (org.apache.cassandra.cql3.QueryOptions)
+                optionsWithNameCtor.newInstance(defaultOptions, boundNames);
+      }
+    } catch (Exception e) {
+      // We can't afford to ignore that: the values wouldn't be in the proper order, and worst
+      // case scenario, this could end up inserting values in the wrong columns, which essentially
+      // boils down to corrupting the use DB.
+      throw new RuntimeException(
```
mpenick, line 186: Should we throw a `ServerError` here?

mpenick:
Should we add a couple quick integration tests to `CQLTests.java` for named markers and timestamps?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -311,6 +312,13 @@ public Authenticator getAuthenticator() {
                   }
 
                   final CQLStatement statement = prepared.statement;
+                  // Please note that this need to happen _before_ the beginTraceExecute, because
```
mpenick, line 315: aside: Is this a bug in the formatter? (it's done this to me too)

pcmanus:
> Should we add a couple quick integration tests to `CQLTests.java` for named markers and timestamps?

Certainly. I'll add some. Thanks for the pointer.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
pcmanus: Maybe? I expected that we catch all exceptions somewhere and transform unknown ones into `ServerError` anyway. That's kind of how it works in C*: `ServerError` means "any exception that we didn't really expected to be throw but there is a bug", but we don't go and throw `ServerError` all over the place.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
pcmanus: Yeah, weird. I cleaned it up and the formatter seems to leave it away now.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -337,9 +338,15 @@ public Authenticator getAuthenticator() {
               throw new PreparedQueryNotFoundException(id);
             }
 
+            // Please note that this needs to happen _before_ the beginTraceExecute, because when
+            // we add bound values to the trace, we rely on the values having been re-ordered by
+            // the following prepare (if named values were used that is).
```
jorgebay, line 343: Good idea to include a note.

jorgebay (approved):
Looks good, given the circumstances of having to rely on reflection :)

Not directly related to this patch but should we create a ticket on the DSE side (and also on C*, after 4) to ask for API changes and get rid of some of the hacks?

pcmanus:
>  should we create a ticket on the DSE side (and also on C*, after 4) to ask for API changes and get rid of some of the hacks?

Yes, we should. At least on the DSE side, there is absolutely no reason to not expose what we need. And even for C*, since we're only asking for minor visibility changes, I suspect we might even get C* 3.11 to change. I'll put on my TODO list to create those ticket, and maybe even submit patches.

mpenick (approved):

PR #181, opened 2020-10-01
Fix native protocol version negotiation for DSE persistence

mpenick:
The java-driver overrides the negotiated protocol version based on `dse_version`. If it's null then it uses the `cluster_version` (Cassandra's version) instead.


mpenick:
The java-driver will override the negotiated protocol version (via native protocol) using a calculated protocol version considering the `dse_version`:

https://github.com/datastax/java-driver/blob/6e39fb62928e6ab149ae5b89785a6d85373fa803/core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java#L402-L413

via the method `highestCommon`():

https://github.com/datastax/java-driver/blob/4.9.0/core/src/main/java/com/datastax/oss/driver/internal/core/DefaultProtocolVersionRegistry.java#L110-L209

dimas-b (approved):

PR #183, opened 2020-10-01
Fix dse bind by name

mpenick:


dimas-b (dismissed):

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -269,26 +269,40 @@ public Authenticator getAuthenticator() {
...
-      org.apache.cassandra.cql3.QueryOptions internalOptions = Conversion.toInternal(options);
-
       return Conversion.toFuture(
           Single.defer(
               () -> {
                 try {
-                  final UUID tracingId =
-                      beginTraceQuery(
-                          cql, internalState, internalOptions, customPayload, isTracingRequested);
+                  org.apache.cassandra.service.QueryState internalState =
```
tomekl007, line 276: Moving this inside of the `Single.defer` is fixing the race-condition, right?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -269,26 +269,40 @@ public Authenticator getAuthenticator() {
...
+                      state,
+                      options,
+                      internalState,
+                      internalOptions,
+                      customPayload,
+                      queryStartNanoTime,
+                      tracingId);
                 } catch (Exception e) {
-                  return Single.error(Conversion.handleException(e));
+                  return Tracing.instance
```
tomekl007, line 303: I know that this is happening inside of the async flow, but I wonder if we shouldn't make a `stop()` (blocking) instead of the async version?
If there are an exception and the client will not chain the operations properly on the returned `CompletableFuture` we may risk the situation in which the session will not be closed when needed, wdyt?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -269,26 +269,40 @@ public Authenticator getAuthenticator() {
...
+                      .stopSessionAsync()
+                      .andThen(Single.error(Conversion.handleException(e)));
```
jorgebay, line 305: This `stopSessionAsync()` call on sync catch is confusing... looking at DSE codebase, it's added in `BatchMessage` and `QueryMessage` but not in `ExecuteMessage`.

jorgebay (approved):
lgtm, I've asked a question below.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
mpenick: Great catch. Thanks!

dougwettlaufer:


```diff

Clones added
============
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6270855&bid=20413467)

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
mpenick: You're right. This doesn't make sense because it wouldn't be on the TPC thread processing the request so the thread local state for the tracing session would be invalid.

tomekl007 (approved):

pcmanus (approved):

PR #171, opened 2020-09-30
[ON HOLD] add producer-kafka module

tomekl007:
it contains:
- Create a test harness for Kafka producer integration tests
- Register the Kafka producer as an OSGI bundle.	
- Create change event key
- Create change event value
- Handle DeleteEvent
- Create a mapping from keyspace.table to Kafka topic
- Integrate with schema registry
- Handle Cassandra/DSE schema changes.
- Handle Conversions from Dse/Cassandra data type to Avro
- Publish Kafka specific metrics
- Add documentation for Kafka CDC producer
- Test that KafkaCDCProducer support sending to N topics 
- Separate Unit and Integration tests

PR #187, opened 2020-10-02
Integrate with schema registry && Handle Cassandra/DSE schema changes.

tomekl007:


producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java
```diff
@@ -0,0 +1,235 @@
...
+
+    int schemaId = registerSchema(tableMetadata, keySchema, subject);
+
+    logger.info(
+        "Registered keySchema: {}, for subject: {} and id: {}", keySchema, subject, schemaId);
+  }
+
+  private int registerSchema(TableMetadata tableMetadata, ParsedSchema schema, String subject) {
+    try {
+      return schemaRegistryClient.register(subject, schema);
```
jorgebay, line 121: There are no async counterparts in the `SchemaRegistryClient`?

producer-kafka/src/main/java/io/stargate/producer/kafka/schema/CqlToAvroTypeConverter.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class CqlToAvroTypeConverter {
+  public static Schema toAvroType(CQLType type) {
+    if (type.equals(TEXT)) {
+      return Schema.create(Type.STRING);
+    } else if (type.equals(INT)) {
+      return Schema.create(Type.INT);
+    } else {
+      // todo handle other types
+      throw new UnsupportedOperationException(String.format("The type: %s is not supported", type));
```
jorgebay, line 33: Is there a different ticket for this?

producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java
```diff
@@ -0,0 +1,235 @@
...
+      keyBuilder.name(columnMetadata.getName()).type(avroFieldSchema).noDefault();
+    }
+    return keyBuilder.endRecord();
+  }
+
+  Schema constructValueSchema(TableMetadata tableMetadata) {
+    List<Schema> partitionKeys =
+        constructRequiredValueFieldsSchema(tableMetadata.getPartitionKeys());
+    List<Schema> clusteringKeys =
+        constructRequiredValueFieldsSchema(tableMetadata.getClusteringKeys());
```
jorgebay, line 143: Clustering keys might or might not be present in a DELETE event.

producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java
```diff
@@ -0,0 +1,235 @@
...
+              .endRecord();
+      fields.add(SchemaBuilder.unionOf().nullType().and().type(partitionKey).endUnion());
+    }
+    return fields;
+  }
+
+  private List<Schema> constructOptionalValueFieldsSchema(List<ColumnMetadata> tableMetadata) {
+    List<Schema> partitionKeys = new ArrayList<>();
+    for (ColumnMetadata columnMetadata : tableMetadata) {
+      Schema partitionKey =
```
jorgebay, line 203: I think the variable names "partitionKeys" don't apply here, probably a leftover.

producer-kafka/src/test/java/io/stargate/producer/kafka/KafkaCDCProducerIntegrationTest.java
```diff
@@ -16,76 +16,78 @@
...
 
   @BeforeAll
-  public static void setup() {
+  public static void setup() throws Exception {
     Network network = Network.newNetwork();
-    kafkaContainer = new KafkaContainer().withNetwork(network);
-    toxiproxyContainer = new ToxiproxyContainer().withNetwork(network);
-    toxiproxyContainer.start();
-    kafkaProxy = toxiproxyContainer.getProxy(kafkaContainer, PORT);
+    kafkaContainer = new KafkaContainer().withNetwork(network).withEmbeddedZookeeper();
```
jorgebay, line 74: oh, the testcontainers API for kafka is super comfortable.

jorgebay (commented):
On `KeyValueConstructor`, it uses the "latest schema for topic".

Given that there are calls from multiple threads to `CDCProducer#publish()`, I think this might be an issue.

producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java
tomekl007: No, there is only a sync version.
But on the other hand, we are wrapping calls to schema-registry within `CompletableFuture`, i.e.:
`return CompletableFuture.runAsync(() -> schemaProvider.createOrUpdateSchema(tableMetadata)); `
so there willl be no blocking from the perspective of the stargate core CDC framework

producer-kafka/src/main/java/io/stargate/producer/kafka/schema/CqlToAvroTypeConverter.java
tomekl007: Yes, I have the ticket for it (see the streaming table changes epic). I plan to work on it soon.

producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java
tomekl007: and that is perfectly ok, this `required` means not-null value, but the field can be missing because it is a union type:
`SchemaBuilder.unionOf().nullType().and().type(partitionKey).endUnion()`
so this:
`CK : null` will not be possible, but absence of the `CK` is.
I've renamed those methods, let me know if it is more clear now.

producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java
tomekl007: good catch, done

producer-kafka/src/test/java/io/stargate/producer/kafka/KafkaCDCProducerIntegrationTest.java
tomekl007: yes, it has great Kafka integration (only schema-registry is missing, but it is very easy to setup and start embedded version)

dougwettlaufer:


```diff

Issues
======
- Added 11
           

Complexity increasing per file
==============================
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderTest.java  1
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java  4
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/MappingProviderKeyValueConstructorIntegrationTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/EmbeddedSchemaRegistryServer.java  3
- producer-kafka/src/test/java/io/stargate/producer/kafka/mapping/MappingServiceTest.java  1
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/CqlToAvroTypeConverter.java  4
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
         

Clones added
============
- producer-kafka/src/test/java/io/stargate/producer/kafka/KafkaCDCProducerIntegrationTest.java  6
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6274566&bid=20422428)

producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java
jorgebay: oh, got it, thanks for the info.

tomekl007:
> `KeyValueConstructor` uses the "latest schema for topic".
> 
> Given that there are calls from multiple threads to `CDCProducer#publish()`, would this be an issue?
> 
> I left other comments below.

The schema evolution will be forward and backward compatible. The forward is a crucial part here: it means that if some of the producer instances will migrate to the new schema (latest) and some do not yet both producers can still publish data that will be readable by consumers. So consumers will be able to read the new field, but ignore it when it is missing. 
It is achieved by union with `default = null`.
My only concern with the `latest` is the performance of it. Fetching the latest will mean more rest calls to schema-registry. (it should be cached by some time)  But at this point in time, this is probably a premature optimization - I will create a followup ticket in case we will have performance problems.

dougwettlaufer:


```diff

Issues
======
- Added 10
           

Complexity increasing per file
==============================
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/mapping/MappingServiceTest.java  1
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/CqlToAvroTypeConverter.java  4
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/EmbeddedSchemaRegistryServer.java  3
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java  4
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/MappingProviderKeyValueConstructorIntegrationTest.java  1
         

Clones added
============
- producer-kafka/src/test/java/io/stargate/producer/kafka/KafkaCDCProducerIntegrationTest.java  6
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6274566&bid=20422428)

jorgebay:
> The schema evolution will be forward and backward compatible. 

I'm not sure I understand. Consider the following scenario:
- Thread A wants to publish a mutation w/ schema version 1
- Thread B wants to publish a mutation w/ schema version 1
- Thread C wants to publish a mutation w/ schema version 2

In case thread B gets behind thread C in the race, thread B will get version 2 (incorrect).

Could that scenario still occur or am I missing something?

tomekl007:
> > The schema evolution will be forward and backward compatible.
> 
> I'm not sure I understand. Consider the following scenario:
> 
> * Thread A wants to publish a mutation w/ schema version 1
> * Thread B wants to publish a mutation w/ schema version 1
> * Thread C wants to publish a mutation w/ schema version 2
> 
> In case thread B gets behind thread C in the race, thread B will get version 2 (incorrect).
> 
> Could that scenario still occur or am I missing something?

Yes, this scenario is possible - this is a common scenario in the microservices world - where you have N producers. Let's assume that a rolling deployment of the new producer is happening, there will be a point in time, where some producers are publishing events with schema 1 and some with schema 2. 
This is why schema-registry has an ability to enforce the schema evolution validation, please see here:
https://docs.confluent.io/current/schema-registry/avro.html
The `FULL_TRANSITIVE` type of schema evolution means that `All previous versions` are backward and forward compatible.
It means that a producer can publish data with any version of feature schema. So in your example, the B will get version `2`, but it is not problematic - it will leave new fields unset.
 
This is a common validation type for microservices/multi-thread deployment.
The kafka-producer-extension will use the schema-registry that should be configured to enforce `FULL_TRANSITIVE`.
Otherwise, keeping the schema in sync between N thread producers and N stargate instances would require a global synchronization. 

Now that I think of it more, I see a problem with schema-registry and it's validation using `FULL_TRANSITIVE` on types that are unions. 
There are two changes that we should treat as non-backward compatible:
- the rename of a column - if this is happening, we should add a new column to a schema instead of replacing it. Would it be possible to send the new and old column name in the `createTableSchemaAsync(TableMetadata)`? If yes, we should add a new column to the schema when rename is happening.
- delete of a column - this should be basically no-op in the schema; the removed field with a default null must be kept.
When new producers will create an Avro event without the deleted field, they will simply omit setting it. 

Right now, they will both be problematic in the scenario that you did describe. 
Just to double-check, we need to support both, right?
If yes, we need to find a way how to propagate the rename and delete column event with information that happened to be able to modify schema in a compatible way. Is it possible to have this info?

jorgebay:
We need to support both and there's no way (that I can think of) of getting that metadata change "event".

I think some kind of map of `TableMetadata` and kafka schema id has to be maintained, to use the value for a given table version.

tomekl007:
> We need to support both and there's no way (that I can think of) of getting that metadata change "event".
> 
> I think some kind of map of `TableMetadata` and kafka schema id has to be maintained, to use the value for a given table version.

I think that map with ids is not solving the problem. The `createTableSchemaAsync` can be executed on any ONE stargate node, isn't it? If yes, then the new schema will be created and id maintained in one producer instance. But the `publish` can be sent to N stargate nodes. In such a setup the newest schema will be visible to only one node that gets `createTableSchemaAsync`.

jorgebay:
I think it should work using multiple nodes. Each node will try to obtain a schema version for a given table. Even if schema registry does not support idempotent operations, the nodes will post mutations with different schema identifiers, but those schemas will be valid and submitted beforehand.

tomekl007:
> I think it should work using multiple nodes. Each node will try to obtain a schema version for a given table. Even if schema registry does not support idempotent operations, the nodes will post mutations with different schema identifiers, but those schemas will be valid and submitted beforehand.

**Yes, this is true as long as the `createTableSchemaAsync` will be executed on all stargate nodes. Is it true?** 
If yes, then we will have an N version of the same schema - each will have a different id, and each stargate instance will fetch its own version.

jorgebay:
that's correct, `createTableSchemaAsync()` will be invoked in every stargate that gets a mutation.

The guarantee is that, on every node, `send()` will be invoked only after `createTableSchemaAsync()` is called (and awaited to complete) for a given table metadata.

tomekl007:
> that's correct, `createTableSchemaAsync()` will be invoked in every stargate that gets a mutation.
> 
> The guarantee is that, on every node, `send()` will be invoked only after `createTableSchemaAsync()` is called (and awaited to complete) for a given table metadata.

It's still problematic because the map of schema id per subject will not be persistent. If we will restart the node, this map will be empty. In such a case we still need to fetch the latest schema and we will risk the race-condition once again. Do we have the ability in CDC core to call the `createTableSchemaAsync()` always when it starts getting CDC changes? Especially when the node is restarting.

tomekl007:
When I think about it more, I think that we must keep the schema consistent and never remove a field in the schema. We should also always treat the rename as adding a new field. 
The reason for it is that consumer will read the data using the recent schema, but they might consume data from Kafka from the earliest offset (so data that was written by the producer before some schema change):
- if we will rename a field, there will be some records with a new field, but some older records with the older field name. If the consumer will use the latest schema to read this data, the latest schema will not have the field before rename. It will break the consumer logic.
I am proposing this solution in `createOrUpdateSchema`:
```
updateSchemaAsync(metadata)                                               
v = getLatest 
newSchema = addWithoutDuplicates(allfields(v), allNewFields(metadata))
registerSchema(newSchema) 
```
the addWithoutDuplicates will construct a new schema based on `allFIelds` in the latest schema appending new fields from metadata. It will not keep duplicates. 
I think I've analyzed all multi-thread scenarios, and I think that we are safe during race condition, here is a detailed analysis:
1. no schema yet scenario:
```
1. no race
// thread-1
updateSchemaAsync(metadata)                                               
v = getLatest // v == null
newSchema = addWithoutDuplicates(allfields(v), allNewFields(metadata))
registerSchema(newSchema) // the same schema with id 1

// thread-2
updateSchemaAsync(metadata)                                               
v = getLatest // v == null
newSchema = addWithoutDuplicates(allfields(v), allNewFields(metadata))
registerSchema(newSchema) // the same schema with id 2

2. race
// thread-1
updateSchemaAsync(metadata)                                               
v = getLatest // v == null
newSchema = addWithoutDuplicates(allfields(v), allNewFields(metadata))
registerSchema(newSchema) // the same schema with id 1

// thread-2
updateSchemaAsync(metadata)                                               
v = getLatest // v == thread-1 schema
newSchema = addWithoutDuplicates(allfields(v), allNewFields(metadata)) // no new fields appended because metadata is the same for thread-1 and thread-2
registerSchema(newSchema) // the same schema with id 2
```

2. schema-exists, add a new field or rename - both treat the same way:
```
// no race 
// thread-1                                                               
updateSchemaAsync(metadata) // metadata has one new field                                               
v = getLatest 
newSchema = addWithoutDuplicates(allfields(v), allNewFields(metadata)) // append one addiional field
registerSchema(newSchema)

// thread-2                                                              
updateSchemaAsync(metadata)                                               
v = getLatest
newSchema = addWithoutDuplicates(allfields(v), allNewFields(metadata)) // append one addiional field
registerSchema(newSchema)

// race
// thread-1                                                               
updateSchemaAsync(metadata) // metadata has one new field (same situation for rename)                                                
v = getLatest 
newSchema = addWithoutDuplicates(allfields(v), allNewFields(metadata)) // append one addiional field
registerSchema(newSchema)

// thread-2                                                              
updateSchemaAsync(metadata)                                               
v = getLatest // get latest wih new field
newSchema = addWithoutDuplicates(allfields(v), allNewFields(metadata)) // no-op, it does not add duplicate
registerSchema(newSchema)
```

3. schema-exists, remove a field
```
// no race 
// thread-1                                                               
updateSchemaAsync(metadata) // metadata has one less field                                              
v = getLatest 
newSchema = addWithoutDuplicates(allfields(v), allNewFields(metadata)) // do nothing
registerSchema(newSchema)

// thread-2                                                              
updateSchemaAsync(metadata)                                               
v = getLatest
newSchema = addWithoutDuplicates(allfields(v), allNewFields(metadata)) // do nothing
registerSchema(newSchema)

// race
// thread-1                                                               
updateSchemaAsync(metadata) // metadata has one less field                                              
v = getLatest 
newSchema = addWithoutDuplicates(allfields(v), allNewFields(metadata)) // do nothing
registerSchema(newSchema)

// thread-2                                                              
updateSchemaAsync(metadata)                                               
v = getLatest 
newSchema = addWithoutDuplicates(allfields(v), allNewFields(metadata)) // do nothing
registerSchema(newSchema)
```
 @jorgebay wdyt?

jorgebay:
The record has the schema id and the consumer can read the schema by id, then there's no need for any specific logic. Am I missing something?

tomekl007:
> The record has the schema id and the consumer can read the schema by id, then there's no need for any specific logic. Am I missing something?

It depends on the consumer logic. From my experience, some of the clients tend to use only the latest schema to read the data (`use.latest.version=ture`) instead of fetching the schema by id that is in the Avro field. 
If we are ok with restricting our clients from using it, then we should be ok without constructing schema based on the old one.

In that case, there is still an open question about the first start (persistence):
https://github.com/stargate/stargate/pull/187#issuecomment-703693553

dougwettlaufer:


```diff

Issues
======
- Added 10
           

Complexity increasing per file
==============================
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/CqlToAvroTypeConverter.java  4
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/MappingProviderKeyValueConstructorIntegrationTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/mapping/MappingServiceTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/EmbeddedSchemaRegistryServer.java  3
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java  4
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
         

Clones added
============
- producer-kafka/src/test/java/io/stargate/producer/kafka/KafkaCDCProducerIntegrationTest.java  6
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6274566&bid=20422428)

jorgebay:
> If we will restart the node, this map will be empty. In such a case we still need to fetch the latest schema and we will risk the race-condition once again.

Restarting a node is not something that occurs every day... We have to manage it gracefully, but its ok to add some overhead because of it.

We can also mitigate having additional schema versions by trying to compare the latest version against the new version (other node might created it already). In any case those should be compatible.

dougwettlaufer:


```diff

Issues
======
- Added 9
           

Complexity increasing per file
==============================
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/EmbeddedSchemaRegistryServer.java  3
- producer-kafka/src/test/java/io/stargate/producer/kafka/mapping/MappingServiceTest.java  1
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java  5
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/CqlToAvroTypeConverter.java  4
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/MappingProviderKeyValueConstructorIntegrationTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
         

Clones added
============
- producer-kafka/src/test/java/io/stargate/producer/kafka/KafkaCDCProducerIntegrationTest.java  6
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6274566&bid=20422428)

dougwettlaufer:


```diff

Issues
======
- Added 9
           

Complexity increasing per file
==============================
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/CqlToAvroTypeConverter.java  4
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/MappingProviderKeyValueConstructorIntegrationTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/EmbeddedSchemaRegistryServer.java  3
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/mapping/MappingServiceTest.java  1
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java  4
         

Clones added
============
- producer-kafka/src/test/java/io/stargate/producer/kafka/KafkaCDCProducerIntegrationTest.java  6
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6274566&bid=20422428)

tomekl007:
I've implemented what we discussed, please take a look. 

> We can also mitigate having additional schema versions by trying to compare the latest version against the new version (other node might created it already). In any case those should be compatible.

I agree but I think that this can be done separately if there will be a need for such performance improvement. I can create a followup ticket for it.

dougwettlaufer:


```diff

Issues
======
- Added 9
           

Complexity increasing per file
==============================
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/MappingProviderKeyValueConstructorIntegrationTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/mapping/MappingServiceTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java  4
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/EmbeddedSchemaRegistryServer.java  3
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/CqlToAvroTypeConverter.java  4
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderTest.java  1
         

Clones added
============
- producer-kafka/src/test/java/io/stargate/producer/kafka/KafkaCDCProducerIntegrationTest.java  6
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6274566&bid=20422428)

dougwettlaufer:


```diff

Issues
======
- Added 6
           

Complexity increasing per file
==============================
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/EmbeddedSchemaRegistryServer.java  3
- producer-kafka/src/test/java/io/stargate/producer/kafka/mapping/MappingServiceTest.java  1
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java  4
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/MappingProviderKeyValueConstructorIntegrationTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/CqlToAvroTypeConverter.java  4
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryException.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderTest.java  1
         

Clones added
============
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/KafkaCDCProducerIntegrationTest.java  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6274566&bid=20422428)

dougwettlaufer:


```diff

Issues
======
- Added 6
           

Complexity increasing per file
==============================
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/CqlToAvroTypeConverter.java  4
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/mapping/MappingServiceTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/MappingProviderKeyValueConstructorIntegrationTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/EmbeddedSchemaRegistryServer.java  3
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryException.java  1
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java  4
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
         

Clones added
============
- producer-kafka/src/test/java/io/stargate/producer/kafka/KafkaCDCProducerIntegrationTest.java  6
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6274566&bid=20422428)

dougwettlaufer:


```diff

Issues
======
- Added 7
           

Complexity increasing per file
==============================
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/EmbeddedSchemaRegistryServer.java  3
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/MappingProviderKeyValueConstructorIntegrationTest.java  1
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderTest.java  1
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryException.java  1
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java  4
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
- producer-kafka/src/main/java/io/stargate/producer/kafka/schema/CqlToAvroTypeConverter.java  4
- producer-kafka/src/test/java/io/stargate/producer/kafka/mapping/MappingServiceTest.java  1
         

Clones added
============
- producer-kafka/src/test/java/io/stargate/producer/kafka/KafkaCDCProducerIntegrationTest.java  6
- producer-kafka/src/test/java/io/stargate/producer/kafka/schema/SchemaRegistryProviderIntegrationTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6274566&bid=20422428)

dougwettlaufer (approved):

tomekl007:
I need to merge it because it blocks my work.
If someone wants to review it, look at `cdc-basic-kafka` branch and this commit:
https://github.com/stargate/stargate/pull/171/commits/7ff47f96ecea25abd21bcee861e1a8d6d2aa4a71

producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java
```diff
@@ -78,21 +84,42 @@ public void createOrUpdateSchema(TableMetadata tableMetadata) {
     createOrUpdateValueSchema(tableMetadata);
   }
 
+  private Schema getSchemaBySubject(String subjectName) {
+    Integer schemaId = schemaIdPerSubject.get(subjectName);
+    if (schemaId == null) {
```
jorgebay, line 89: I'm not sure I understand when this condition will hit.

jorgebay (commented):
lgtm, just one doubt below.

producer-kafka/src/main/java/io/stargate/producer/kafka/schema/SchemaRegistryProvider.java
tomekl007: When there is a restart of the stargate node and the `schemaIdPerSubject` is empty.

tomekl007: Ok, nice. In that case, this can be simplified. I will remove this code branch, and throw `IllegalStateException` if this invariant is not kept.

PR #213, opened 2020-10-06
Swagger for REST w/o `dropwizard-swagger`

mpenick:


mpenick:
Never mind, It could be a pretty big undertaking to switch to 3.0. I'm not sure it's that beneficial right now.

restapi/pom.xml
```diff
@@ -172,6 +138,75 @@
...
+            <id>copy-resources</id>
+            <phase>prepare-package</phase>
+            <goals>
+              <goal>copy-resources</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${project.build.directory}/classes/swagger-ui</outputDirectory>
+              <resources>
+                <resource>
+                  <directory>${project.build.directory}/swagger-ui/META-INF/resources/webjars/swagger-ui/${swagger-ui.version}</directory>
```
dougwettlaufer, line 202: What does this mean for when we need to pass in tokens and display models based on db content?

restapi/src/main/java/io/stargate/web/impl/Server.java
```diff
@@ -104,7 +116,21 @@ protected void configure() {
     environment.jersey().register(TablesResource.class);
     environment.jersey().register(KeyspacesResource.class);
     environment.jersey().register(ColumnsResource.class);
-    environment.jersey().register(SwaggerStaticResource.class);
+
+    environment.jersey().register(ApiListingResource.class);
```
dougwettlaufer, line 120: Does this still return raw json/yaml with `/swagger.json` and `/swagger.yaml`?

restapi/pom.xml
mpenick: Not sure what you mean here.

restapi/pom.xml
```diff
@@ -172,6 +138,75 @@
...
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <file>${project.build.directory}/swagger-ui/META-INF/resources/webjars/swagger-ui/${swagger-ui.version}/index.html</file>
+          <replacements>
+            <replacement>
+              <token>https://petstore.swagger.io/v2/swagger.json</token>
+              <value>/swagger.json</value>
+            </replacement>
```
dougwettlaufer, line 183: We'll need to be careful about this when we upgrade versions.

restapi/src/main/java/io/stargate/web/impl/Server.java
mpenick:
```

dougwettlaufer: Sweet

dougwettlaufer (approved):

PR #217, opened 2020-10-07
Use a profile to build DSE

### LINKED ISSUE #202: Don't try to build persistence-dse-6.8 by default
Currently persistence-dse-6.8 is in the default build process.  It requires authenticated access to repo.datastax.com.  To encourage community development, we should either open up the dependencies for the dse jars or exclude persistence-dse-6.8 from being built by default so that people don't have to do that manually to simply build it for Cassandra.

dougwettlaufer:
Resolves #202

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,4 +52,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
           CCM_CLUSTER_START_TIMEOUT_OVERRIDE: 600
-        run: mvn -B verify --file pom.xml -P it-${{ matrix.backend }}
+        run: mvn -B verify --file pom.xml -P it-${{ matrix.backend }},dse
--- a/pom.xml
+++ b/pom.xml
@@ -8,24 +8,6 @@
   <name>${project.groupId}:${project.artifactId}</name>
   <description>Core modules for Stargate</description>
   <url>http://github.com/stargate/stargate</url>
-  <modules>
-    <module>core</module>
-    <module>stargate-starter</module>
-    <module>persistence-api</module>
-    <module>persistence-common</module>
-    <module>persistence-cassandra-3.11</module>
-    <module>persistence-cassandra-4.0</module>
-    <module>persistence-dse-6.8</module>
-    <module>cql</module>
-    <module>graphqlapi</module>
-    <module>restapi</module>
-    <module>auth-api</module>
-    <module>authentication</module>
-    <module>auth-table-based-service</module>
-    <module>health-checker</module>
-    <!-- testing last -->
-    <module>testing</module>
-  </modules>
   <properties>
     <osgi.bundle.version>${project.version}</osgi.bundle.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -52,4 +52,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
           CCM_CLUSTER_START_TIMEOUT_OVERRIDE: 600
-        run: mvn -B verify --file pom.xml -P it-${{ matrix.backend }}
+        run: mvn -B verify --file pom.xml -P it-${{ matrix.backend }},dse
```
jorgebay, line 55: It would be nice to test without token and without the `dse` profile.

jorgebay (approved):
Looks good!

I've added a suggestion below but it doesn't have to be included as part of this patch.

.github/workflows/ci.yml
dougwettlaufer: What do you mean "without token"?

jorgebay: Sorry, meant without a custom artifactory.

pom.xml
```diff
@@ -115,6 +97,53 @@
...
+        <module>cql</module>
+        <module>graphqlapi</module>
+        <module>restapi</module>
+        <module>auth-api</module>
+        <module>authentication</module>
+        <module>auth-table-based-service</module>
+        <module>health-checker</module>
+        <!-- testing last -->
+        <module>testing</module>
+      </modules>
```
dimas-b, line 145: Do we need to have default modules in a profile? Could we put them in the main (top-level) `modules` block? Something like that works with configs... I wonder if it might work with modules too. Having an unconditional section, plus the `dse` module _addition_ in a profile would be more readable I think :thinking:

testing/pom.xml
```diff
@@ -359,5 +353,19 @@
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>dse</id>
```
dimas-b, line 357: Could it break the default "it" profile selection? AFAIK, if any profile is specified _explicitly_ all the "default" profiles will not be active anymore :thinking:

dimas-b: Might be worth refactoring those `it-*` profiles into plain `cassandra-3.11`, `cassandra-4.0` and `dse-6.8`... and reuse them both for compilation and testing. WDYT?

As for the `-6.8` suffix I think it matches the persistence modules name well (it also has the version suffix).

.github/workflows/ci.yml
dougwettlaufer: Ah, sure. Was able to do that locally but didn't want to increase our build times by splitting this out more. In a follow up I can add some logic based on the `matrix.backend`.

testing/pom.xml
dougwettlaufer: > Could it break the default "it" profile selection? AFAIK, if any profile is specified explicitly all the "default" profiles will not be active anymore

From what I've seen so far, no.

> Might be worth refactoring those it-* profiles into plain cassandra-3.11, cassandra-4.0 and dse-6.8... and reuse them both for compilation and testing. WDYT?

I think that might be a bit overkill. I'd prefer to be able to build as much as possible without profiles.

.github/workflows/ci.yml
dougwettlaufer: https://github.com/stargate/stargate/pull/218

PR #222, opened 2020-10-08
Cleanup persistence API and move DataStore on top of it

pcmanus:
At a high level, this PR does the following:
- it "cleans up" the `Persistence` interface. The existing interface is mimicking pretty closely how the C* code is organized, including a number of "wrapping" classes. But while the C* code may have reasons to be how it is (through sometimes it's simply ugly because), it doesn't make for a very friendly API. This PR attempts to improve this.
- it re-implements `DataStore` entirely on top of the (updated) `Persistence` interface. This avoid having multiple entry points into the persistence and save quite a bit of code for persistence implementations.
- it updates the persistence implementations, both to the interface changes, but also to build and execute transport message directly, instead of inlining the code of those messages. This reduce the code needed quite a bit, but also limit the risk of us forgetting something when inlining said messages execution code.

This is a relatively big code change (though a decent part of it is pretty simple), so for reviews, I suggest to first look at the new `Persistence` interface to see the differences as everything else is tied one way or another.


dougwettlaufer:


```diff

Issues
======
+ Solved 17
- Added 31
           

Complexity increasing per file
==============================
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedResultSet.java  10
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedDataStore.java  1
- persistence-api/src/main/java/io/stargate/db/SimpleStatement.java  1
- cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/BatchMessage.java  2
- cql/src/main/java/org/apache/cassandra/stargate/transport/internal/QueryOptions.java  19
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/ReflectionUtils.java  11
- persistence-api/src/main/java/org/apache/cassandra/stargate/exceptions/AuthenticationException.java  1
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  6
- persistence-api/src/main/java/io/stargate/db/BoundStatement.java  1
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  3
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedPreparedStatement.java  15
- persistence-common/src/main/java/io/stargate/db/datastore/common/util/UncheckedExecutionException.java  1
- persistence-api/src/main/java/io/stargate/db/datastore/ArrayListBackedRow.java  4
- persistence-api/src/main/java/io/stargate/db/ClientInfo.java  5
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/ClientStateWithPublicAddress.java  1
- persistence-api/src/main/java/io/stargate/db/Statement.java  1
- persistence-api/src/main/java/io/stargate/db/Parameters.java  16
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateQueryHandler.java  3
- persistence-api/src/main/java/io/stargate/db/Batch.java  1
         

Complexity decreasing per file
==============================
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  -4
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
+ persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java  -2
+ persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
         

Clones added
============
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  4
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  2
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  2
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  4
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java  5
         

Clones removed
==============
+ graphqlapi/src/main/java/io/stargate/graphql/fetchers/AlterTableAddFetcher.java  -1
+ graphqlapi/src/main/java/io/stargate/graphql/fetchers/AlterTableDropFetcher.java  -1
+ graphqlapi/src/main/java/io/stargate/graphql/fetchers/CreateTableDataFetcher.java  -1
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
+ graphqlapi/src/main/java/io/stargate/graphql/fetchers/DropTableFetcher.java  -1
+ persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6304491&bid=20501494)

dougwettlaufer:


```diff

Issues
======
+ Solved 17
- Added 24
           

Complexity increasing per file
==============================
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/ClientStateWithPublicAddress.java  1
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateQueryHandler.java  3
- persistence-api/src/main/java/io/stargate/db/Batch.java  1
- persistence-api/src/main/java/io/stargate/db/BoundStatement.java  1
- cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/BatchMessage.java  2
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedResultSet.java  10
- persistence-api/src/main/java/io/stargate/db/Statement.java  1
- persistence-common/src/main/java/io/stargate/db/datastore/common/util/UncheckedExecutionException.java  1
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedDataStore.java  1
- persistence-api/src/main/java/io/stargate/db/ClientInfo.java  5
- cql/src/main/java/org/apache/cassandra/stargate/transport/internal/QueryOptions.java  19
- persistence-api/src/main/java/io/stargate/db/datastore/ArrayListBackedRow.java  4
- persistence-api/src/main/java/org/apache/cassandra/stargate/exceptions/AuthenticationException.java  1
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  3
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  6
- persistence-api/src/main/java/io/stargate/db/SimpleStatement.java  1
- persistence-api/src/main/java/io/stargate/db/Parameters.java  16
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/ReflectionUtils.java  11
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedPreparedStatement.java  15
         

Complexity decreasing per file
==============================
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  -4
+ persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java  -2
+ persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
         

Clones added
============
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  4
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java  5
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  4
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  2
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  2
         

Clones removed
==============
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
+ graphqlapi/src/main/java/io/stargate/graphql/fetchers/CreateTableDataFetcher.java  -1
+ persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java  -1
+ graphqlapi/src/main/java/io/stargate/graphql/fetchers/DropTableFetcher.java  -1
+ graphqlapi/src/main/java/io/stargate/graphql/fetchers/AlterTableAddFetcher.java  -1
+ graphqlapi/src/main/java/io/stargate/graphql/fetchers/AlterTableDropFetcher.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6304491&bid=20501494)

jorgebay:
-7K loc, I'm already liking this. I'm going through it...

persistence-common/src/main/java/io/stargate/db/datastore/common/AbstractCassandraPersistence.java
```diff
@@ -13,18 +20,15 @@
...
  * @param <T> the concrete class for table metadata in the persistence layer.
  * @param <C> the concrete class for column metadata in the persistence layer.
  * @param <U> the concrete class for user types in the persistence layer.
  * @param <I> the concrete class for secondary indexes metadata in the persistence layer.
  * @param <V> the concrete class for materialized views metadata in the persistence layer.
  */
-public abstract class AbstractCassandraPersistence<
-        Config, ClientState, QueryState, K, T, C, U, I, V>
-    implements Persistence<Config, ClientState, QueryState> {
+public abstract class AbstractCassandraPersistence<Config, K, T, C, U, I, V>
```
jorgebay, line 30: Not related to this changeset, but can we rename `schemaConverter()` method to `newSchemaConverter()` ?

jorgebay (commented):
I took a first pass at it, its looking great.

I've added a few comments below, mostly minor suggestions. I'm seeing some issues on the persistence tests related to `DataStore`, probably those need adapting.

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -56,32 +51,39 @@
...
-        org.apache.cassandra.service.QueryState,
         KeyspaceMetadata,
         TableMetadata,
         ColumnMetadata,
         UserType,
         IndexMetadata,
         ViewMetadata> {
+  static {
+    System.setProperty(
+        "cassandra.custom_query_handler_class", StargateQueryHandler.class.getName());
```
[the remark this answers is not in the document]
pcmanus: I can check if this work. But unfortunately the way this work in C* is a tad fragile. Namely, that property is checked during static initialization of the `ClientState` class. So my reasoning was to put it as early as possible to ensure that `ClientState` don't get loaded in some unexpected way before we've set it (thinking, for instance, the code generating the config, will runs before `initializePersistence`, and C* ends up with weird class dependencies sometimes).

In any case, I can probably add a comment :).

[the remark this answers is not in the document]
pcmanus: I moved it at the beginning of `initializePersistence` with a comment. This appears to work just fine. We can revisit if this become a problem (though if it does, we may have some fun figuring it out...).

pcmanus:
> I'm seeing some issues on the persistence tests related to `DataStore`, probably those need adapting.

Can you be more specific? Do you mean compilation issues, or test issues? I don't seem to be getting any.

dougwettlaufer:


```diff

Issues
======
+ Solved 17
- Added 24
           

Complexity increasing per file
==============================
- persistence-api/src/main/java/io/stargate/db/Parameters.java  16
- persistence-common/src/main/java/io/stargate/db/datastore/common/util/UncheckedExecutionException.java  1
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateQueryHandler.java  3
- cql/src/main/java/org/apache/cassandra/stargate/transport/internal/QueryOptions.java  19
- persistence-api/src/main/java/org/apache/cassandra/stargate/exceptions/AuthenticationException.java  1
- persistence-api/src/main/java/io/stargate/db/ClientInfo.java  5
- cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/BatchMessage.java  2
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  3
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedResultSet.java  10
- persistence-api/src/main/java/io/stargate/db/SimpleStatement.java  1
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedPreparedStatement.java  15
- persistence-api/src/main/java/io/stargate/db/Batch.java  1
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/ReflectionUtils.java  11
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/ClientStateWithPublicAddress.java  1
- persistence-api/src/main/java/io/stargate/db/Statement.java  1
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedDataStore.java  1
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  6
- persistence-api/src/main/java/io/stargate/db/BoundStatement.java  1
- persistence-api/src/main/java/io/stargate/db/datastore/ArrayListBackedRow.java  4
         

Complexity decreasing per file
==============================
+ persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java  -2
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  -4
+ persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
         

Clones added
============
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  4
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  2
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  4
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java  5
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  2
         

Clones removed
==============
+ persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java  -1
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
+ graphqlapi/src/main/java/io/stargate/graphql/fetchers/AlterTableAddFetcher.java  -1
+ graphqlapi/src/main/java/io/stargate/graphql/fetchers/DropTableFetcher.java  -1
+ graphqlapi/src/main/java/io/stargate/graphql/fetchers/AlterTableDropFetcher.java  -1
+ graphqlapi/src/main/java/io/stargate/graphql/fetchers/CreateTableDataFetcher.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6304491&bid=20501494)

jorgebay:
> Do you mean compilation issues, or test issues?

Test issues. You can reproduce it if you run `CassandraPersistenceIT`, it also occurs on CI.

pcmanus:
> Test issues. You can reproduce it if you run `CassandraPersistenceIT`, it also occurs on CI.

My bad, I see those now. I hadn't realized how to run those. I'll check.

dimas-b:
As for running `CassandraPersistenceIT`, it should just work from IntelliJ, but also in maven:

```
$ ./mvnw -f persistence-cassandra-3.11 failsafe:integration-test@it-cassandra-3.11
```

persistence-api/src/main/java/io/stargate/db/Persistence.java
```diff
@@ -15,96 +15,162 @@
...
   Authenticator getAuthenticator();
 
-  DataStore newDataStore(QueryState<Q> state, QueryOptions queryOptions);
+  /**
+   * Creates a new connection for an "external" client identified by the provided info.
+   *
+   * @param clientInfo info identifying the client for which the connection is created.
+   * @return the newly created persistence connection.
+   */
+  Connection newConnection(ClientInfo clientInfo);
```
dimas-b, line 62: `ClientInfo` appears to require an `InetSocketAddress`... it will be difficult to provide for SQL clients, for example (because socket info is hidden behind Avatica driver APIs).

How critical is having the client's address available?

persistence-api/src/main/java/io/stargate/db/Persistence.java
```diff
@@ -15,96 +15,162 @@
  */
 package io.stargate.db;
 
+import com.datastax.oss.driver.shaded.guava.common.util.concurrent.Uninterruptibles;
```
dimas-b, line 18: Do we want to expose driver class in persistence API? I believe it would be preferable to hide them (even if we use the internally). For SQL, for example, C* drivers may not be relevant at all.

persistence-api/src/main/java/io/stargate/db/Persistence.java
```diff
@@ -15,96 +15,162 @@
...
-      long queryStartNanoTime);
-
   boolean isInSchemaAgreement();
 
-  void captureClientWarnings();
-
-  List<String> getClientWarnings();
+  /** Wait for schema to agree across the cluster */
+  default void waitForSchemaAgreement() {
+    for (int count = 0; count < 100; count++) {
```
dimas-b, line 85: Could we change the timeout into a parameter? (perhaps add another method that take a timeout parameter?)

dimas-b: ... and have a system property to control the default timeout perhaps?

persistence-api/src/main/java/io/stargate/db/Persistence.java
dougwettlaufer: Well, on the flip side we really don't want to use guava at this layer. Is there an alternative to `Uninterruptibles` or maybe we should shade our own guava?

persistence-api/src/main/java/io/stargate/db/Persistence.java
pcmanus: > `ClientInfo` appears to require an `InetSocketAddress`

First, I'll note it's not really new to this patch (it was before in the stargate `ClientState` wrapper, that is replace by this `ClientInfo`). It's due to C* underlying `ClientState` class using `InetSocketAddress`.

Now, why does C* cares about he client address you may ask? It's a good question, and I'd need to check in details. But at a minimum, I believe it uses it in some logging/audit logging, and possibly tracing.

But I wouldn't worry for SQL. Simplest solution is we just don't provide a `ClientInfo` at all (check `newConnection`, there is an argument-less version), and that's likely just fine. And maybe longer term we figure out it's neat to have "some" client info for SQL, and I'm sure we can figure out a way to do it (we can hardcode the port for instance).

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -207,318 +221,40 @@ public InetAddressAndPort getNativeAddress(InetAddressAndPort endpoint) {
...
-  public CompletableFuture<? extends Result> prepare(
-      String cql,
-      QueryState state,
-      Map<String, ByteBuffer> customPayload,
-      boolean isTracingRequested) {
-    CompletableFuture<Result> future = new CompletableFuture<>();
-
-    InternalDataStore.EXECUTOR.submit(
+  private <T extends Result> CompletableFuture<T> runOnExecutor(Supplier<T> supplier) {
+    assert executor != null : "This persistence has not be initialized";
```
dougwettlaufer, line 245:
```suggestion
    assert executor != null : "This persistence has not been initialized";
```

persistence-api/src/main/java/io/stargate/db/Persistence.java
pcmanus: That method is, contrary to what it looks like, not new to this patch; it's just moved as-is from `DataStore`. And I didn't wanted to mess up with schema agreement in that patch so I left it as is. Which isn't to say I disagree in principle, and in fact I need to dig why we have both this method and the different (but essentially doing the same thing) `SchemaAgreement` class in the `cql` module.

So if that's ok, I'll create a quick follow-up issue to clean up schema agreement code.

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -194,330 +209,42 @@ public InetAddressAndPort getNativeAddress(InetAddressAndPort endpoint) {
...
-  public CompletableFuture<? extends Result> prepare(
-      String cql,
-      QueryState state,
-      Map<String, ByteBuffer> customPayload,
-      boolean isTracingRequested) {
-    CompletableFuture<Result> future = new CompletableFuture<>();
-
-    EXECUTOR.submit(
+  private <T extends Result> CompletableFuture<T> runOnExecutor(Supplier<T> supplier) {
+    assert executor != null : "This persistence has not be initialized";
```
dougwettlaufer, line 233:
```suggestion
    assert executor != null : "This persistence has not been initialized";
```

dougwettlaufer:


```diff

Issues
======
+ Solved 17
- Added 24
           

Complexity increasing per file
==============================
- persistence-api/src/main/java/io/stargate/db/datastore/ArrayListBackedRow.java  4
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedDataStore.java  1
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  3
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedPreparedStatement.java  15
- cql/src/main/java/org/apache/cassandra/stargate/transport/internal/QueryOptions.java  19
- persistence-common/src/main/java/io/stargate/db/datastore/common/util/UncheckedExecutionException.java  1
- persistence-api/src/main/java/io/stargate/db/SimpleStatement.java  1
- persistence-api/src/main/java/io/stargate/db/Statement.java  1
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/ReflectionUtils.java  11
- cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/BatchMessage.java  2
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateQueryHandler.java  3
- persistence-api/src/main/java/io/stargate/db/Parameters.java  16
- persistence-api/src/main/java/io/stargate/db/BoundStatement.java  1
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedResultSet.java  10
- persistence-api/src/main/java/io/stargate/db/ClientInfo.java  5
- persistence-api/src/main/java/org/apache/cassandra/stargate/exceptions/AuthenticationException.java  1
- persistence-api/src/main/java/io/stargate/db/Batch.java  1
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  6
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/ClientStateWithPublicAddress.java  1
         

Complexity decreasing per file
==============================
+ persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java  -2
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  -4
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
+ persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
         

Clones added
============
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  2
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java  5
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  4
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  4
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  2
         

Clones removed
==============
+ persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java  -1
+ graphqlapi/src/main/java/io/stargate/graphql/fetchers/DropTableFetcher.java  -1
+ graphqlapi/src/main/java/io/stargate/graphql/fetchers/AlterTableAddFetcher.java  -1
+ graphqlapi/src/main/java/io/stargate/graphql/fetchers/AlterTableDropFetcher.java  -1
+ graphqlapi/src/main/java/io/stargate/graphql/fetchers/CreateTableDataFetcher.java  -1
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6304491&bid=20501494)

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
```diff
@@ -369,48 +360,31 @@ public static Result toResult(
         return new Result.SetKeyspace(((ResultMessage.SetKeyspace) resultMessage).keyspace);
       case SCHEMA_CHANGE:
         return new Result.SchemaChange(toSchemaChangeMetadata(resultMessage));
+      case PREPARED:
```
dougwettlaufer, line 363: Was this breaking something since it was missing?

persistence-api/src/main/java/io/stargate/db/Persistence.java
dimas-b: I see that it's used a lot in persistence API, but comes from a dedicated artifact (`java-driver-shaded-guava`), so it's probably fine.

persistence-api/src/main/java/io/stargate/db/Persistence.java
dimas-b: sure. thx!

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java
```diff
@@ -108,38 +119,51 @@
...
-    if (options == null) {
-      return org.apache.cassandra.cql3.QueryOptions.DEFAULT;
-    }
+  public static QueryOptions toInternal(
+      List<ByteBuffer> values, List<String> boundNames, Parameters parameters) {
     org.apache.cassandra.transport.ProtocolVersion protocolVersion =
-        toInternal(options.getProtocolVersion());
+        toInternal(parameters.protocolVersion());
+    // Note that PagingState.deserialize below modifies its input, so we duplicate to avoid nasty
+    // surprises down the line
```
dougwettlaufer, line 145: Duplicate comment
```suggestion
```

dougwettlaufer (approved):

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
pcmanus: No. There was a separate method to handle prepared results previously. As the code is more factored now, this has to move here.

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthTableBasedService.java
```diff
@@ -157,12 +154,12 @@ private String queryUsername(String key) throws ExecutionException, InterruptedE
             .where("role", WhereCondition.Predicate.Eq, key)
             .execute();
 
-    if (resultSet.isEmpty()) {
+    if (resultSet.hasNoMoreFetchedRows()) {
       throw new RuntimeException(String.format("Provided username %s is incorrect", key));
```
dimas-b, line 158: This is not related this PR again, but I believe in general it is not advisable to disclose information about user name existence if the credentials have not been verified... It would be preferable to report `Provided username %s and/or password are incorrect`, WDYT?

dimas-b:
I did not go through all the changes, but overall LGTM :) 

+1 to moving `DataStore` out of persistence implementations!

dimas-b (approved):

persistence-api/src/main/java/io/stargate/db/Persistence.java
```diff
@@ -15,96 +15,162 @@
...
+     *
+     * @param batch the batch to execute.
+     * @param parameters the parameters for the batch execution.
+     * @return a future that, on completion of the execution, provides its result. Please see the
+     *     {@link #execute} method for warnings regarding the completion of that future. The same
+     *     warnings apply for this method as well.
+     */
+    CompletableFuture<Result> batch(Batch batch, Parameters parameters, long queryStartNanoTime);
+
+    void captureClientWarnings();
```
mpenick, line 170: Removing the inline request execution code is a good idea. That approach may give us another opportunity. We can probably remove these warning methods from `Persistence`. 

How about pulling the warnings from `ResultMessage` using `ResultMessage.getWarnings()`  (it's not private, so no reflection needed :partying_face:)? This might work as long as there's no warnings on `ErrorMessage`, otherwise we'd need to handle that path too.

```diff
diff --git a/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Message.java b/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Message.java
index 579e10f..5346498 100644
--- a/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Message.java
+++ b/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Message.java
@@ -658,9 +658,6 @@ public abstract class Message {
 
         final Persistence.Connection persistenceConnection = connection.persistenceConnection();
 
-        if (connection.getVersion().isGreaterOrEqualTo(ProtocolVersion.V4))
-          persistenceConnection.captureClientWarnings();
-
         connection.validateNewMessage(request.type, connection.getVersion());
 
         logger.trace("Received: {}, v={}", request, connection.getVersion());
@@ -675,7 +672,6 @@ public abstract class Message {
               } else {
                 try {
                   response.setStreamId(request.getStreamId());
-                  response.setWarnings(persistenceConnection.getClientWarnings());
                   response.attach(connection);
                   connection.applyStateTransition(request.type, response.type);
 
@@ -691,8 +687,6 @@ public abstract class Message {
                       "Failed to reply, got another error whilst writing reply: {}",
                       t.getMessage(),
                       t);
-                } finally {
-                  persistenceConnection.resetClientWarnings();
                 }
               }
             });
diff --git a/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/ResultMessage.java b/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/ResultMessage.java
index 29a7982..96bf109 100644
--- a/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/ResultMessage.java
+++ b/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/ResultMessage.java
@@ -67,6 +67,7 @@ public class ResultMessage extends Message.Response {
     super(Message.Type.RESULT);
     this.result = result;
     this.tracingId = result.getTracingId();
+    this.warnings = result.getWarnings();
   }
 
   public static class VoidSubCodec implements CBCodec<Result> {
diff --git a/persistence-api/src/main/java/io/stargate/db/Persistence.java b/persistence-api/src/main/java/io/stargate/db/Persistence.java
index 963fb22..24202da 100644
--- a/persistence-api/src/main/java/io/stargate/db/Persistence.java
+++ b/persistence-api/src/main/java/io/stargate/db/Persistence.java
@@ -166,11 +166,5 @@ public interface Persistence {
      *     warnings apply for this method as well.
      */
     CompletableFuture<Result> batch(Batch batch, Parameters parameters, long queryStartNanoTime);
-
-    void captureClientWarnings();
-
-    List<String> getClientWarnings();
-
-    void resetClientWarnings();
   }
 }
diff --git a/persistence-api/src/main/java/io/stargate/db/Result.java b/persistence-api/src/main/java/io/stargate/db/Result.java
index ca5b8e6..3ae091e 100644
--- a/persistence-api/src/main/java/io/stargate/db/Result.java
+++ b/persistence-api/src/main/java/io/stargate/db/Result.java
@@ -50,6 +50,7 @@ public abstract class Result {
 
   public final Kind kind;
   private UUID tracingId;
+  private List<String> warnings;
 
   private Result(Kind kind) {
     this.kind = kind;
@@ -64,6 +65,15 @@ public abstract class Result {
     return tracingId;
   }
 
+  public Result setWarnings(List<String> warnings) {
+    this.warnings = warnings;
+    return this;
+  }
+
+  public List<String> getWarnings() {
+    return warnings;
+  }
+
   public static class Rows extends Result {
     public final List<List<ByteBuffer>> rows;
     public final ResultMetadata resultMetadata;
diff --git a/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java b/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
index 6ed6a08..214d658 100644
--- a/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
+++ b/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
@@ -433,20 +433,5 @@ public class CassandraPersistence
         return Conversion.toInternal(((BoundStatement) statement).preparedId());
       }
     }
-
-    @Override
-    public void captureClientWarnings() {
-      ClientWarn.instance.captureWarnings();
-    }
-
-    @Override
-    public List<String> getClientWarnings() {
-      return ClientWarn.instance.getWarnings();
-    }
-
-    @Override
-    public void resetClientWarnings() {
-      ClientWarn.instance.resetWarnings();
-    }
   }
 }
diff --git a/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java b/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
index ec60cf8..1ff87b4 100644
--- a/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
+++ b/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
@@ -437,7 +437,9 @@ public class Conversion {
 
   public static Result toResult(
       ResultMessage resultMessage, org.apache.cassandra.transport.ProtocolVersion version) {
-    return toResultInternal(resultMessage, version).setTracingId(resultMessage.getTracingId());
+    return toResultInternal(resultMessage, version)
+        .setTracingId(resultMessage.getTracingId())
+        .setWarnings(resultMessage.getWarnings());
   }
 
   public static Result toResultInternal(
diff --git a/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java b/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
index a092ebd..8875647 100644
--- a/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
+++ b/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
@@ -424,20 +424,5 @@ public class CassandraPersistence
         return Conversion.toInternal(((BoundStatement) statement).preparedId());
       }
     }
-
-    @Override
-    public void captureClientWarnings() {
-      ClientWarn.instance.captureWarnings();
-    }
-
-    @Override
-    public List<String> getClientWarnings() {
-      return ClientWarn.instance.getWarnings();
-    }
-
-    @Override
-    public void resetClientWarnings() {
-      ClientWarn.instance.resetWarnings();
-    }
   }
 }
diff --git a/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java b/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
index 919a6bf..8b4eb0b 100644
--- a/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
+++ b/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
@@ -343,7 +343,8 @@ public class Conversion {
   public static Result toResult(
       ResultMessage resultMessage, org.apache.cassandra.transport.ProtocolVersion version) {
     return toResultInternal(resultMessage, version)
-        .setTracingId(ReflectionUtils.getTracingId(resultMessage));
+        .setTracingId(ReflectionUtils.getTracingId(resultMessage))
+        .setWarnings(resultMessage.getWarnings());
   }
 
   private static Result toResultInternal(
diff --git a/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java b/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java
index f2807ed..e1238b6 100644
--- a/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java
+++ b/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java
@@ -445,7 +445,9 @@ public class Conversion {
 
   public static Result toResult(
       ResultMessage resultMessage, org.apache.cassandra.transport.ProtocolVersion version) {
-    return toResultInternal(resultMessage, version).setTracingId(resultMessage.getTracingId());
+    return toResultInternal(resultMessage, version)
+        .setTracingId(resultMessage.getTracingId())
+        .setWarnings(resultMessage.getWarnings());
   }
 
   private static Result toResultInternal(
diff --git a/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java b/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
index d26c583..e08a664 100644
--- a/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
+++ b/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
@@ -448,20 +448,5 @@ public class DsePersistence
         return Conversion.toInternal(((BoundStatement) statement).preparedId());
       }
     }
-
-    @Override
-    public void captureClientWarnings() {
-      ClientWarn.instance.captureWarnings();
-    }
-
-    @Override
-    public List<String> getClientWarnings() {
-      return ClientWarn.instance.getWarnings();
-    }
-
-    @Override
-    public void resetClientWarnings() {
-      ClientWarn.instance.resetWarnings();
-    }
   }
 }
```

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -157,6 +152,11 @@ protected void unregisterInternalSchemaListener() {
 
   @Override
   protected void initializePersistence(Config config) {
+    // C* picks this property during the static loading of the ClientState class. So we set it
+    // early, to make sure that class i not loaded before we've set it.
```
mpenick, line 156: "i" --> "is"

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -143,6 +138,11 @@ protected void unregisterInternalSchemaListener() {
 
   @Override
   protected void initializePersistence(Config config) {
+    // C* picks this property during the static loading of the ClientState class. So we set it
+    // early, to make sure that class i not loaded before we've set it.
```
mpenick, line 142: same typo as above.

olim7t (approved):

persistence-api/src/main/java/io/stargate/db/Persistence.java
olim7t: That artifact is intended for internal use in the driver only, so I would not rely on it. It's not guaranteed to stay there forever (at some point there was some talk of forking the required classes into the driver codebase directly and removing it).

persistence-api/src/main/java/io/stargate/db/Persistence.java
```diff
@@ -15,96 +15,162 @@
...
+     * Prepare a query on this connection.
+     *
+     * @param query the query to prepare.
+     * @param parameters the parameters for the preparation (note that preparation only use a subset
+     *     of parameters).
+     * @return a future that, on completion of the preparation, provides its result. Please see the
+     *     {@link #execute} method for warnings regarding the completion of that future. The same
+     *     warnings apply for this method as well.
+     */
+    CompletableFuture<Result.Prepared> prepare(String query, Parameters parameters);
```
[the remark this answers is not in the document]
olim7t: +1, from a purely theoretical standpoint we don't want to expose `complete()` to the caller.

But in practice CompletionStage has a `toCompletableFuture()` method so this is a bit idealistic.

persistence-api/src/main/java/io/stargate/db/datastore/DataStore.java
```diff
@@ -45,19 +68,11 @@ default QueryBuilder query() {
   CompletableFuture<ResultSet> query(
       String cql, Optional<ConsistencyLevel> consistencyLevel, Object... parameters);
```
olim7t, line 69: GraphQL (and other APIs I imagine) needs to set other execution parameters: fetch size, paging state... And it's a bit more convenient to do it at this level, rather than when we create the DataStore.

Instead of just `consistencyLevel`, this method could take a `UnaryOperator<Parameters>` that gives the caller a chance to customize whatever they want.

persistence-api/src/main/java/io/stargate/db/Persistence.java
pcmanus: So there is 2 sub-parts here:

The class you're pointing to is indeed a guava class, imported from the shaded version inside the driver. I agree using that is quite dodgy, but again, this is actually copied code from `DataStore` and I simply didn't change it. Imo, if we want to use Guava, we should add guava as a direct dependency and use it (but the other option could be to avoid it altogether, I don't think it's that used in `persistence-api` so far; don't know if we care about minimizing dependencies). Anyway, I suggested in another comment to delay cleaning up `waitForSchemaAgreement` to a followup and that's what use it, so maybe leave that to that followup? Or if that bother you guys, happy to add guava as dependency and replace.

The other point though is that we _do_ expose driver classes in the persistence API. And it is not new to this patch either. For instance, the `Column.ColumnType` classes exposes `TypeCodec` from the driver. Admittedly, this patch does expose driver classes a bit more because I've reused `GettableByIndex`/`GettableByName` in `Row` to allow reusing quite a bit of boring code. But to answer the more general question, I do think we should absolutely reuse the abstraction that already exists in the drivers, are public, and have already had the test of time. In fact, I suggest we use more of the driver code: the `DataStore` is essentially a driver-like interface to persistence, so it uses a lot of the same concepts (Row, ResultSet, Columns, ...) and I think it's a shame we've rewritten quite a few of them. In fact, I plan to look soonish if we can consider reusing much more of the driver interfaces for things like `ResulSet` and all the schema classes.

persistence-api/src/main/java/io/stargate/db/Persistence.java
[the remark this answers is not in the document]
pcmanus: I don't think it's that obvious. Granted, using `CompletionStage` avoid exposing `complete()` and that's good (even if, not that air tight as @olim7t mentions). On the other side, `CompletionStage` is not a `Future` so it's effectively less feature-rich for clients of those method (of course, here again, the `toCompletableFuture` makes both class basically equivalent if you're ok calling that).

Now I know that "blocking" is something though as a curse word in our day and age, but I for one think that the `Future` interface is sometimes nice to have, and it should be up to the client to decide what he wants, but using `CompletionStage` feels to me like suggesting it should be avoided.

And in general, as far as return type goes, returning the most precise type is more "general" for the client (of course, for implementation of the method, using the least precise type is a bit better in that it allows to switch the concrete implementation; it's a balancing act. But in that case, I don't think there is really any other implementation on `CompletionStage` other than `CompletableFuture` so...).

In any case, I'll dodge this a bit for now by saying that the use of `CompletableFuture` is not new or specific to this patch, it's the type we use kind of everywhere. So if we want to change (and tbc, I'm not saying doing so would be bad; just that there is argument for keeping it as is), let's consider that in a followup.

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthTableBasedService.java
pcmanus: Agreed, and that one is simple enough that I'll sneak it in.

persistence-api/src/main/java/io/stargate/db/Persistence.java
pcmanus: Actually, re-shaded-guava: I just realized how often we were accessing it in the `persistence-api` module, and we use it in almost every other module anyway, so I'm going to add the dependency and replace the imports now.

persistence-api/src/main/java/io/stargate/db/Statement.java
```diff
@@ -0,0 +1,24 @@
...
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+public abstract class Statement {
+  private final List<ByteBuffer> values;
+  private final @Nullable List<String> boundNames;
+
+  protected Statement(List<ByteBuffer> values, @Nullable List<String> boundNames) {
```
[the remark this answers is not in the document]
pcmanus: Good point, I'll add a precondition.

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -658,4 +295,149 @@ private void waitForSchema(int delayMillis) {
...
+    public Optional<String> usedKeyspace() {
+      return Optional.ofNullable(clientState.getRawKeyspace());
+    }
+
+    private <T extends Result> CompletableFuture<T> executeRequestOnExecutor(
+        Parameters parameters, long queryStartNanoTime, Supplier<Request> requestSupplier) {
+      return runOnExecutor(
+          () -> {
+            QueryState queryState = new QueryState(clientState);
+            Request request = requestSupplier.get();
```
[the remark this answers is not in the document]
pcmanus: Two benefits:
1. as you mentioned, this move the work on the executor. To be fair, it probably doesn't matter terribly, but feels a tad cleaner maybe (?).
2. this simplify exception handling slightly. Genuine exceptions can be thrown while building the `Request` object, and this way it gets properly handled by the catchall of `runOnExecutor`.

persistence-api/src/main/java/io/stargate/db/Persistence.java
pcmanus: Well well. Turns out there is a reason the `persistence-api` was relying on the shaded jar. C* uses a antiquated version of guava (18.0) and having a newer version in the classpath breaks it (and if I instead "force" the guava version of C* for the `persistence-api`, this doesn't work either because it doesn't have some of the method we use, like `Preconditions#checkArgument`, which is a shame).

Tl;dr, I'll revert to the shaded guava for now. As @dimas-b mentions, we explicitly import it, so it's not "that" terrible. We can find a better option later if we're bothered.

persistence-api/src/main/java/io/stargate/db/Persistence.java
mpenick: We could do this in a follow up ticket. It's probably more important to get this merged.

mpenick (approved):
I think we could improve how warnings are done and make the persistence API even smaller, but that can be done in a follow up ticket.

Really nice improvements. Well done! I'll re-approve again after the conflicts are resolved.

persistence-api/src/main/java/io/stargate/db/datastore/DataStore.java
mpenick: +1 It'd be nice if `DataStore` was a long-lived object and `Parameter` was pass per method instead of a constructor parameter. Maybe this is a follow up issue?

persistence-api/src/main/java/io/stargate/db/Persistence.java
pcmanus: Good idea, that's cleaner. I'll do that (but I'll deal with `ErrorMessage`; I don't know of the top of my mind if it's possible today, but in theory there is no reason a query that errors out "late" can't also have warnings so better be on the safe side).

persistence-api/src/main/java/io/stargate/db/datastore/DataStore.java
pcmanus: Good idea, I'll add that.

dougwettlaufer:


```diff

Issues
======
+ Solved 20
- Added 32
           

Complexity increasing per file
==============================
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  6
- persistence-api/src/main/java/io/stargate/db/datastore/ArrayListBackedRow.java  4
- persistence-api/src/main/java/org/apache/cassandra/stargate/exceptions/PersistenceException.java  1
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedResultSet.java  10
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  3
- cql/src/main/java/org/apache/cassandra/stargate/transport/internal/QueryOptions.java  19
- persistence-api/src/main/java/io/stargate/db/datastore/query/Value.java  2
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/ClientStateWithPublicAddress.java  1
- persistence-api/src/main/java/io/stargate/db/ClientInfo.java  5
- persistence-api/src/main/java/io/stargate/db/Parameters.java  16
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateQueryHandler.java  3
- persistence-api/src/main/java/io/stargate/db/BoundStatement.java  1
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/ReflectionUtils.java  11
- persistence-api/src/main/java/io/stargate/db/Statement.java  3
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedDataStore.java  1
- persistence-api/src/main/java/org/apache/cassandra/stargate/exceptions/AuthenticationException.java  1
- persistence-api/src/main/java/io/stargate/db/Batch.java  1
- cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/BatchMessage.java  2
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/FakeConnection.java  1
- persistence-api/src/main/java/io/stargate/db/SimpleStatement.java  1
- persistence-api/src/main/java/io/stargate/db/datastore/query/MixinPreparedStatement.java  2
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedPreparedStatement.java  15
         

Complexity decreasing per file
==============================
+ persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java  -2
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
+ persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
         

Clones added
============
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  5
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  2
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java  6
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  2
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  6
         

Clones removed
==============
+ persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java  -1
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6304491&bid=20501494)

persistence-api/src/main/java/io/stargate/db/datastore/DataStore.java
pcmanus: I'll note that I kept the `Parameters` in the ctor and the `UnaryOperator<Parameters>` approach. I think it's useful to pass your own "long lived default" (and some parameters like the protocol version probably don't ever need to be specified per-query) but be able to modify a few things per-query, which this allows.

dougwettlaufer:


```diff

Issues
======
+ Solved 21
- Added 31
           

Complexity increasing per file
==============================
- cql/src/main/java/org/apache/cassandra/stargate/transport/internal/QueryOptions.java  19
- persistence-api/src/main/java/io/stargate/db/datastore/query/Value.java  2
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateQueryHandler.java  3
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/ReflectionUtils.java  11
- persistence-api/src/main/java/io/stargate/db/datastore/query/MixinPreparedStatement.java  2
- persistence-api/src/main/java/io/stargate/db/datastore/ArrayListBackedRow.java  4
- persistence-api/src/main/java/io/stargate/db/SimpleStatement.java  1
- cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/BatchMessage.java  2
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedResultSet.java  10
- persistence-api/src/main/java/io/stargate/db/Batch.java  1
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  6
- persistence-api/src/main/java/io/stargate/db/BoundStatement.java  1
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/FakeConnection.java  1
- restapi/src/test/java/io/stargate/web/docsapi/dao/DocumentDBTest.java  1
- persistence-api/src/main/java/org/apache/cassandra/stargate/exceptions/PersistenceException.java  1
- persistence-api/src/main/java/io/stargate/db/Statement.java  3
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/ClientStateWithPublicAddress.java  1
- persistence-api/src/main/java/io/stargate/db/ClientInfo.java  5
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedDataStore.java  8
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedPreparedStatement.java  15
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  3
- persistence-api/src/main/java/org/apache/cassandra/stargate/exceptions/AuthenticationException.java  1
- persistence-api/src/main/java/io/stargate/db/Parameters.java  16
- restapi/src/test/java/io/stargate/web/docsapi/service/DocumentServiceTest.java  2
         

Complexity decreasing per file
==============================
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
+ persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java  -2
+ persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
         

Clones added
============
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  2
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java  7
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  5
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  7
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  2
         

Clones removed
==============
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
+ persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java  -1
+ restapi/src/test/java/io/stargate/web/docsapi/service/DocumentServiceTest.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6304491&bid=20501494)

dougwettlaufer:


```diff

Issues
======
+ Solved 21
- Added 31
           

Complexity increasing per file
==============================
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  6
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/FakeConnection.java  1
- persistence-api/src/main/java/io/stargate/db/datastore/query/MixinPreparedStatement.java  2
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedPreparedStatement.java  15
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/ClientStateWithPublicAddress.java  1
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateQueryHandler.java  3
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedDataStore.java  8
- persistence-api/src/main/java/io/stargate/db/ClientInfo.java  5
- cql/src/main/java/org/apache/cassandra/stargate/transport/internal/QueryOptions.java  19
- persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedResultSet.java  10
- cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/BatchMessage.java  2
- restapi/src/test/java/io/stargate/web/docsapi/dao/DocumentDBTest.java  1
- persistence-api/src/main/java/org/apache/cassandra/stargate/exceptions/AuthenticationException.java  1
- restapi/src/test/java/io/stargate/web/docsapi/service/DocumentServiceTest.java  2
- persistence-api/src/main/java/io/stargate/db/Parameters.java  16
- persistence-api/src/main/java/io/stargate/db/datastore/query/Value.java  2
- persistence-api/src/main/java/io/stargate/db/Statement.java  3
- persistence-api/src/main/java/io/stargate/db/datastore/ArrayListBackedRow.java  4
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/ReflectionUtils.java  11
- persistence-api/src/main/java/io/stargate/db/SimpleStatement.java  1
- persistence-api/src/main/java/io/stargate/db/Batch.java  1
- persistence-api/src/main/java/org/apache/cassandra/stargate/exceptions/PersistenceException.java  1
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  3
- persistence-api/src/main/java/io/stargate/db/BoundStatement.java  1
         

Complexity decreasing per file
==============================
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
+ persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
+ persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java  -2
         

Clones added
============
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  2
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  7
- persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java  7
- persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java  5
- persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java  2
         

Clones removed
==============
+ persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java  -1
+ restapi/src/test/java/io/stargate/web/docsapi/service/DocumentServiceTest.java  -2
+ persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6304491&bid=20501494)

PR #223, opened 2020-10-08
Bump guava on tests

jorgebay:
Fix minor security vulnerability.

### DIFF
--- a/testing/pom.xml
+++ b/testing/pom.xml
@@ -165,7 +165,7 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>19.0</version>
+      <version>29.0-jre</version>
       <scope>test</scope>
     </dependency>
     <dependency>


testing/pom.xml
```diff
@@ -165,7 +165,7 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>19.0</version>
+      <version>29.0-jre</version>
```
tomekl007, line 168: I see in a lot of places in code that we are using shaded`com.datastax.oss.driver.shaded.guava` and that was a deliberate decision. In that case, why do we even need to declare the guava dependency expliclity?

jorgebay: Probably different uses through shaded and direct dependencies and have grown organically in the code, without much attention.

In this case, I'm trying to [fix this SV](https://github.com/stargate/stargate/network/alerts).
We can create a different ticket for unifying dependencies.

tomekl007 (approved):

testing/pom.xml
tomekl007: Ok, maybe we should create a dedicated JIRA ticket for it?

jorgebay: yeah (we are using GH issues now)

PR #234, opened 2020-10-09
Use multiple jobs for CI

dougwettlaufer:


.github/workflows/ci.yml
```diff
@@ -37,6 +34,67 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1.4.2
+        with:
+          java-version: 1.8
+      - name: Setup Maven
+        env:
+          ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_USERNAME }}
+          ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_PASSWORD }}
+        run: |
```
tomekl007, line 74: isn't there a way to extract this artifactory setup logic and share between builds? (it seems that it is the same for this and `dse-6_8`)

tomekl007 (approved):

.github/workflows/ci.yml
dimas-b: GH actions do not support YAML anchors :shrug:

.github/workflows/ci.yml
```diff
@@ -3,11 +3,8 @@ name: Java CI
 on: [push]
 
 jobs:
-  build:
+  build-cassandra-3_11:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        backend: [ "cassandra-3.11", "cassandra-4.0", "dse-6.8" ]
```
dimas-b, line 10: Are multiple jobs that much different from a matrix job? I'd imagine the execution is going to be the same... it certainly looks very similar in the build console :thinking:

dougwettlaufer: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

> A matrix allows you to create multiple jobs by performing variable substitution in a single job definition

So I guess if the matrix is creating multiple jobs then, no, multiple jobs and matrix are not much different. Maybe the improvement I was seeing was from using the cache?

dougwettlaufer:
Closing per https://github.com/stargate/stargate/pull/234#discussion_r503333937

PR #235, opened 2020-10-09
Centralize versions for common dependencies

olim7t:
`<dependencyManagement>` in the parent POM sets the versions globally.
Modules still have to declare their dependencies, but they can omit the versions.

Note that this applies to transitive dependencies as well: if you depend on A, which depends on B:1.5, but you had B:1.3 in the management section, then 1.3 would be used.
Because of that I only factored common dependencies (logging, testing, etc). I don't want to generalize this too much because we'll run into weird cases where two modules have transitive dependencies to different versions of the same library.


pom.xml
```diff
@@ -285,4 +298,188 @@
...
+      <dependency>
+        <groupId>io.stargate.auth.table</groupId>
+        <artifactId>auth-table-based-service</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.stargate.auth.table</groupId>
+        <artifactId>authentication</artifactId>
+        <version>${project.version}</version>
+      </dependency>
```
dougwettlaufer, line 322: I'm not sure these ones belong up here

pom.xml
```diff
@@ -285,4 +298,188 @@
...
+      <dependency>
+        <groupId>io.stargate.graphql</groupId>
+        <artifactId>graphqlapi</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.stargate.health</groupId>
+        <artifactId>health-checker</artifactId>
+        <version>${project.version}</version>
+      </dependency>
```
dougwettlaufer, line 342: Why are we also including top level services up here?

pom.xml
```diff
@@ -285,4 +298,188 @@
...
+      <dependency>
+        <groupId>io.stargate.starter</groupId>
+        <artifactId>stargate-starter</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.stargate.it</groupId>
+        <artifactId>testing</artifactId>
+        <version>${project.version}</version>
+      </dependency>
```
dougwettlaufer, line 387: Do any of the children use these as dependencies?

pom.xml
```diff
@@ -15,6 +15,16 @@
     <doclint>none</doclint>
     <driver.version>4.9.0</driver.version>
     <xml-format.skip>true</xml-format.skip>
+    <dropwizard.version>2.0.12</dropwizard.version>
+    <!--
+      Note that this is currently aligned with the Jetty version that DropWizard transitively depends on.
+      It's probably a good idea to keep the two versions in sync.
+    -->
+    <jetty.version>9.4.30.v20200611</jetty.version>
```
olim7t, line 23: There was also some lack of uniformity across modules: some were using DropWizard 2.0.10. I've aligned everything, and the modules that use Jetty directly use the same version as DropWizard.

pom.xml
```diff
@@ -285,4 +298,188 @@
...
+        <artifactId>testing</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.dropwizard</groupId>
+        <artifactId>dropwizard-core</artifactId>
+        <version>${dropwizard.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.dropwizard</groupId>
```
dougwettlaufer, line 394: Should we maybe create an http pom/bom for these sort of things?

pom.xml
```diff
@@ -15,6 +15,16 @@
...
+    <dropwizard.version>2.0.12</dropwizard.version>
+    <!--
+      Note that this is currently aligned with the Jetty version that DropWizard transitively depends on.
+      It's probably a good idea to keep the two versions in sync.
+    -->
+    <jetty.version>9.4.30.v20200611</jetty.version>
+    <slf4j.version>1.7.30</slf4j.version>
+    <logback.version>1.2.3</logback.version>
+    <junit.version>5.7.0</junit.version>
+    <mockito.version>3.5.13</mockito.version>
```
olim7t, line 27: I only declared properties for versions that are repeated more than once. Otherwise there's no reason to add an extra level of indirection.

auth-api/pom.xml
```diff
@@ -10,15 +10,13 @@
...
   <properties>
-    <jettyVersion>9.4.24.v20191120</jettyVersion>
     <jerseyVersion>2.30.1</jerseyVersion>
     <jacksonVersion>2.10.1</jacksonVersion>
   </properties>
   <dependencies>
     <dependency>
       <groupId>io.stargate.auth</groupId>
       <artifactId>authentication</artifactId>
-      <version>0.0.15-SNAPSHOT</version>
```
dougwettlaufer, line 21: Couldn't we just use `${project.version}`?

testing/pom.xml
```diff
@@ -42,19 +37,11 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
-      <version>${slf4j.version}</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>ch.qos.logback</groupId>
-      <artifactId>logback-core</artifactId>
-      <version>${logback.version}</version>
```
olim7t, line 51: In this module there is no OSGi, so we definitely don't need to declare logback-core, it's a transitive dependency of logback-classic.

auth-api/pom.xml
olim7t: Right that would work too. Either one works for me, I can make the change.

pom.xml
olim7t: Yeah I was debating that, "leaf" modules that no-one depend on don't need to be there. Maybe I was too lazy to find out which these are 🙂 

Possibly a moot point if we switch to `${project.version}` per your other suggestion.

pom.xml
olim7t: Yeah also if there's any common code we can factor we could consider extracting an http utility module.

auth-api/pom.xml
olim7t: Done in e3907ba

pom.xml
dougwettlaufer: https://github.com/stargate/stargate/issues/239

PR #236, opened 2020-10-09
Add parameterMacro to swagger to support prefilling the auth header

dougwettlaufer:


restapi/src/main/resources/swagger-ui-cust/index.html
```diff
@@ -0,0 +1,65 @@
...
+        presets: [
+          SwaggerUIBundle.presets.apis,
+          SwaggerUIStandalonePreset
+        ],
+        plugins: [
+          SwaggerUIBundle.plugins.DownloadUrl
+        ],
+        layout: "StandaloneLayout",
+        parameterMacro: function (operation, parameter) {
+          urlParams = new URLSearchParams(window.location.search);
```
mpenick, line 51: could these two lines move inside the if statement?

mpenick: is it a good idea to have a token credential in the url? it's essentially a limited time password.

dougwettlaufer: Doesn't hurt anything for them to be out here but I can move them.

mpenick (approved):

PR #249, opened 2020-10-12
add document api diagram change

polandll:
Replace current stargate architecture diagram with one that includes Document API for the Document API release.


EricBorczuk:
This addition looks correct! I actually had messed up something when I initially made this image though; the `cql` square in the diagram should connect to only `persistence-api`, just like graphql/rest/documents

dougwettlaufer (approved):

Issue #197, opened 2020-10-05
Publish tests so new persistence backends can be tested


tjake:
Ideally these could be run with https://junit.org/junit5/docs/current/user-guide/#running-tests-console-launcher

Specifying `-Dstargate.test.backend.use.external=true`

dimas-b:
Should be resolved by #327

Issue #188, opened 2020-10-02
Type check in ExternalStorage#supportsParameter is too permissive


olim7t:
Not an issue currently, but I ran into this on a temporary local branch.

```java
  @Override
  public boolean supportsParameter(
      ParameterContext parameterContext, ExtensionContext extensionContext)
      throws ParameterResolutionException {
    return parameterContext.getParameter().getType().isAssignableFrom(ClusterConnectionInfo.class);
  }
```
According to the API contract, this will match if the parameter type "is either the same as, or a superclass or interface of" `ClusterConnectionInfo`. In other words the connection info will be injected into any `Object` parameter.

This is too permissive: if you have parameterized tests, it's possible that one of them could use an `Object` parameter. Either explicitly, or via erasure if the test method is parameterized (that's how I ran into it, I can provide an example if this is not clear).

If Junit finds two parameter resolvers that handle the same type, it throws an exception.

I think exact equality would be sufficient in this case, `Cluster` is private so we know it can't be referenced from a test class:
```java
return parameterContext.getParameter().getType() == ClusterConnectionInfo.class;
```

jeffreyscarpenter:
can't repro

PR #254, opened 2020-10-12
Add JUnit 5 extension for managing Stargate nodes

dimas-b:
* The new @StargateContainer extension can be used in
  conjunction with @ExternalStorage to start one or
  more Stargate nodes during integrations tests.

* @StargateSpec controls the Stargate node configuration

* BaseOsgiIntegrationTest is mostly empty now, but it is
  handy to holding the @UseStargateContainer and
  @ClusterSpec annotations

----

Also dump storage logs to build STDOUT in case of test failure. Hopefully, this will help with diagnosing errors.


dimas-b:
This PR also configures most of the tests to use 1 Stargate node. `MultipleStargateInstancesTest` will use 3 nodes.

dougwettlaufer:


```diff

Issues
======
+ Solved 3
- Added 4
           

Complexity increasing per file
==============================
- testing/src/test/java/io/stargate/it/storage/StargateContainer.java  5
         

Complexity decreasing per file
==============================
+ testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java  -4
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6326779&bid=20560471)

stargate-starter/src/main/java/io/stargate/starter/Starter.java
```diff
@@ -400,6 +408,8 @@ public void start() throws BundleException {
       bundle.start();
     }
 
+    System.out.println(STARTED_MESSAGE);
```
tomekl007, line 411: maybe use `logger.info` instead of `system.out`?

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
```diff
@@ -0,0 +1,528 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(StargateContainer.class);
+
+  private static final File LIB_DIR = new File(System.getProperty("stargate.libdir"));
+  private static final int PROCESS_WAIT_MINUTES =
+      Integer.getInteger("stargate.test.process.wait.timeout.minutes", 10);
+
+  private static final AtomicBoolean suiteExecuting = new AtomicBoolean();
+  private static final AtomicBoolean testExecuting = new AtomicBoolean();
+
+  // the first 10 addresses are reserved for storage nodes
```
tomekl007, line 81: I see one problem with limiting it to 10 nodes only for SG. 
Assuming that we will have 3 test classes that start 3 stargate nodes, and the rest of the test classes start 1 SG node. If the ordering of the tests will be unfortunate, that is:
- test with 3 nodes
- test with 1 node
- test  with 3 nodes
- test with 1 node
- test with 3 nodes
We will start and stop 12 SG nodes, exceeding this limit of 10 nodes. Right now it is not a problem, but maybe in the future?
Can we increase this limit of addresses reserved for SG?

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
```diff
@@ -0,0 +1,528 @@
...
+                        .orElse(DefaultSpecHolder.get()));
+
+    StargateParameters params = parameters(spec, context);
+
+    Container c = container;
+    if (c != null) {
+      if (!c.matches(backend, spec, params)) {
+        c.stop();
+      } else {
+        // reuse container
```
tomekl007, line 163: Maybe we could add a debug log to be able to trace this (if there is a need)

stargate-starter/src/main/java/io/stargate/starter/Starter.java
dimas-b: This class already uses `System.out` to print package activation messages... so I thought this would fix the pattern.

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
dimas-b: 10 addresses are reserved for C* or DSE, which we do not run concurrently. I do not imagine us needing more than 10 nodes of C* for tests :thinking:

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
dimas-b: Sure. will do. I think I'll and an INFO message for visibility.

dimas-b: Added `LOG.info` in 3d0f45473fa69678a78113cf2d4c8d0f49753329

dougwettlaufer:


```diff

Issues
======
+ Solved 3
- Added 4
           

Complexity increasing per file
==============================
- testing/src/test/java/io/stargate/it/storage/StargateContainer.java  5
         

Complexity decreasing per file
==============================
+ testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java  -4
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6326779&bid=20560471)

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
```diff
@@ -0,0 +1,538 @@
...
+
+  @Override
+  public void afterAll(ExtensionContext context) {
+    leave(suiteExecuting);
+  }
+
+  @Override
+  public void beforeEach(ExtensionContext context) throws Exception {
+    enter(testExecuting);
+    start(context);
```
olim7t, line 189: The first test method we hit will restart an instance if its config doesn't match the test class's. That's a bit wasteful when you consider this example:
```java
@UseStargateContainer public class ExtensionTest {
  @Test @StargateSpec(nodes = 2) public void doSomething1() {}
  @Test @StargateSpec(nodes = 1) public void doSomething2() {}
}
```

The extension will perform the following operations:
```
>>> Enter start for Optional[class io.stargate.it.ExtensionTest], 1 nodes needed
>>> Start new Stargate
>>> Enter start for Optional[public void io.stargate.it.ExtensionTest.doSomething1()], 2 nodes needed
>>> Stop current Stargate because it doesn't match config (1 nodes)
>>> Start new Stargate
>>> Enter start for Optional[public void io.stargate.it.ExtensionTest.doSomething2()], 1 nodes needed
>>> Stop current Stargate because it doesn't match config (2 nodes)
>>> Start new Stargate
```

Note that we start an instance at the beginning of the class, but it's immediately shut down as we enter the first method.

One way to fix this is to not start anything in `beforeAll()`. Instead, just memorize the class's `StargateSpec`, and use it as the fallback in the next `beforeEach` if the test method doesn't declare its own. Also with this approach you would only need one AtomicBoolean for mutual exclusion.
The only downside is that if the test class declares its own `@BeforeAll` method, the Stargate instance won't be started when it executes, which might be a bit counter-intuitive.

Another way is to let the caller decide how they want to use the extension: we'd only do "per method" if the extension is declared as an [instance field](https://junit.org/junit5/docs/current/user-guide/#extensions-registration-programmatic-instance-fields); we can detect that in the extension because beforeAll/afterAll will not be called. I think it would be more correct, but a bit more complicated.

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
```diff
@@ -0,0 +1,538 @@
...
+      Class<?> testClass = context.getRequiredTestClass();
+      Method method = testClass.getMethod(customizer, Builder.class);
+      method.invoke(testInstance, builder);
+    }
+
+    return builder.build();
+  }
+
+  private static synchronized void start(ExtensionContext context) throws Exception {
+    ClusterConnectionInfo backend = ExternalStorage.connectionInfo();
```
olim7t, line 144: Can we use JUnit's parameter injection instead of relying on static state? It would be nice to say "use the ExternalStorage that is active for this test".

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
```diff
@@ -0,0 +1,538 @@
...
+    start(context);
+  }
+
+  @Override
+  public void afterEach(ExtensionContext context) throws Exception {
+    leave(testExecuting);
+  }
+
+  private boolean isStargateConnectionInfo(ParameterContext parameterContext) {
+    return parameterContext.getParameter().getType().isAssignableFrom(StargateConnectionInfo.class);
```
olim7t, line 198: I think that's too coarse-grained, see #188.

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
```diff
@@ -0,0 +1,538 @@
...
+      return id.toString();
+    }
+
+    @Override
+    public List<? extends StargateConnectionInfo> nodes() {
+      return nodes;
+    }
+  }
+
+  private static class Node extends Thread implements StargateConnectionInfo {
```
olim7t, line 300: It feels redundant to start our own thread, when commons-exec is already doing things asynchronously.
We could use `execute(CommandLine, ExecuteResultHandler)`, and do `exit.countDown()` in the handler.

Kind of similar to what you're already doing for `ready`.

olim7t (changes requested):

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
dimas-b: I'm thinking to just drop the default `@StargateSpec` completely. If a test needs a SG node in `@BeforeAll` it will have to have an annotation at the class level. If the annotation is missing, the extension will simply not start SG. Consequently, if SG connection info parameters are required, but there is not spec in scope, we can produce a nice error message.

Also, I'll probably drop the boolean mutex check and replace it with a check against the cluster ID :thinking:

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
dimas-b: I do not think parameter injection works between extensions. I use the `ExtensionContex.Store` for that initially, but the code looked actually messier that static accessors :) 

I'll try something else presently :thinking:

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
dimas-b: `DefaultExecutor` will create a new thread in that case anyway, but we won't control the thread's name :thinking: 

We can avoid creating a thread with `CommandLauncher`, but then we'd have to do a lot more work managing I/O streams... I'm not sure it's worth the trouble. We do not have that many SG nodes anyway... WDYT?

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
olim7t: It's kind of disappointing if there's no easy way to share information between extensions.

For the Java driver we passed the actual rule instance to the dependent rule ([example](https://github.com/datastax/java-driver/blob/4.x/integration-tests/src/test/java/com/datastax/oss/driver/core/cql/SimpleStatementCcmIT.java#L50-L58)). Maybe we could do something similar with JUnit 5 extensions, but you'd have to declare every rule programmatically as a static or instance field (no more `@ExtendWith`).

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
```diff
@@ -0,0 +1,538 @@
...
+    StargateSpec spec =
+        context
+            .getElement()
+            .flatMap(e -> Optional.ofNullable(e.getAnnotation(StargateSpec.class)))
+            .orElseGet(
+                () ->
+                    context
+                        .getTestClass()
+                        .flatMap(c -> Optional.ofNullable(c.getAnnotation(StargateSpec.class)))
+                        .orElse(DefaultSpecHolder.get()));
```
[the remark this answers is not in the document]
dimas-b: I'm not sure `AnnotationUtils` will be helpful in this context. The trick is that "element" may not be present, in which case we default to class, which again may not be present. As far as I understand `AnnotationUtils` will simply return nothing if the argument is not present :thinking:

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
olim7t: `DefaultExecutor` already does everything asynchronously. The only reason why you're holding a thread is to wait until the command has exited, and run those 3 instructions:
```java
LOG.info("Stargate node {} existed with return code {}", nodeIndex, retValue);
FileUtils.deleteDirectory(cacheDir);
exit.countDown();
```
What I'm saying is, you can move that into an `ExecuteResultHandler` and call the async `execute()` variant. You don't need your own thread anymore, just launch the command and forget, the callback will be invoked when the command returns.

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
olim7t: Right, that would work too. The only drawback is that `@StargateSpec` would now have to be explicitly set everywhere.

dimas-b: What if we introduced a `scope` for `@StargateSpec` (similar to `@ClusterSpec`), and added a new scope of `VOID`. This way if methods had a variety of specs the class could have the `VOID` spec to effectively stop any previous SG containers without starting new ones. WDYT?

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
dimas-b: I'll have another try with `ExtensionContex.Store`. We could put the key name into the "spec" annotation in case we need to support multiple storage clusters in the same test env. :thinking:

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
dimas-b: yes, but if we call `execute(CommandLine, ExecuteResultHandler)` a thread will be created by that method anyway. So the only difference would be in who creates that thread... Did I miss something else?

dimas-b: I think I found a way to set our own thread name with `execute(CommandLine, ExecuteResultHandler)`... I'll switch to that.

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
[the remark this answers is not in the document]
olim7t: I think `element' is the class if you're coming from `beforeAll`, or the method if you're coming from `beforeEach`. That's what I saw in my debug logs [here](https://github.com/stargate/stargate/pull/254#discussion_r504893102).

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
olim7t: > if we call `execute(CommandLine, ExecuteResultHandler)` a thread will be created by that method anyway. So the only difference would be in who creates that thread... Did I miss something else?

I hadn't looked at the commons-exec code, I thought it always started a thread and the synchronous execute was just a blocking wrapper. But it's actually the other way around, so fair enough.

So my only argument is to not reinvent the wheel if the library already does it. I don't think the thread name is hugely important, but if you've found a way to change it then all good 👍

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
olim7t: I don't know, I feel that JUnit already has a pretty clear scoping model to execute extensions per class or per method, so why add our own on top.

dimas-b: Do you mean that we derive scope from the element that has `@StargateSpec`? So, e.g. if the annotation is on the class, we start/stop in before/afterAll callbacks, if it's on the method we start/stop in before/afterEach callbacks? Concequently, if it's default (no `@StargateSpec` annotation) - it's a shared instance stopped at the JVM exit?

dougwettlaufer:


```diff

Issues
======
+ Solved 3
- Added 4
           

Complexity increasing per file
==============================
- testing/src/test/java/io/stargate/it/storage/StargateContainer.java  5
         

Complexity decreasing per file
==============================
+ testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java  -4
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6326779&bid=20560471)

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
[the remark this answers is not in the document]
dimas-b: I've completely rework this section. Please re-review.

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
```diff
@@ -0,0 +1,538 @@
...
+    }
+
+    c = new Container(backend, spec, params);
+    c.start();
+    container = c;
+  }
+
+  @Override
+  public void beforeAll(ExtensionContext context) throws Exception {
+    enter(suiteExecuting);
```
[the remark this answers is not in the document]
dimas-b: this section has been reworked.

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
dimas-b: start/stop logic has been reworked. Please review again.

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
dimas-b: Storage into objects are passed through JUnit contexts now.

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
dimas-b: fixed

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
dimas-b: I changes this code to use execution listeners.

dimas-b:
rebased to the latest `master`

dougwettlaufer (approved):

olim7t (approved):
👍 

Could you add a short how-to in `DEV_GUIDE.md`? I think the most useful bits are:
* `@StargateContainer` annotation always go on the class
* `@StargateSpec` goes on the class to get a per-class Stargate instance, or on the test methods to get per-method instances.
* `StargateEnvironmentInfo` can be injected: if class-level, into `@BeforeAll`, `@BeforeEach` and `@Test` methods; if method-level, only into `@BeforeEach` and `@Test` methods.

I'll work on adding a `CqlSession` rule in the near future, this will be a very useful example to follow, thanks.

dimas-b:
Thanks @olim7t . I'll do DEV_GUIDE improvements in a follow-up PR. I've got one in the queue for minor changes to allow JUnit 5 console launcher (#197) and it has some DEV_GUILDE changes already.

dimas-b:
DEV_GUIDE notes proposed in #327

PR #269, opened 2020-10-13
Add basic SQL execution code

dimas-b:
* Using Apache Calcite 1.25.0 for query planning
* Using Avatica for client communication

Note: Cassandra query plans are rudimentary at this time.
Most queries will do a full scan.

Note: Authentication is trivial - it attempts to create a
new Stargate auth token given JDBC username / password.


dougwettlaufer:


```diff

Issues
======
- Added 42
           

Complexity increasing per file
==============================
- sql/server/src/test/java/io/stargate/api/sql/ListBackedResultSet.java  2
- sql/server/src/main/java/io/stargate/api/sql/schema/StargateSchema.java  2
- sql/server/src/main/java/io/stargate/api/sql/server/StargateConnection.java  3
- sql/server/src/main/java/io/stargate/api/sql/server/StargateStatement.java  4
- sql/server/src/main/java/io/stargate/api/sql/schema/StargateTable.java  1
- sql/server/src/main/java/io/stargate/api/sql/plan/rel/CassandraSingleRowQuery.java  2
- sql/server/src/main/java/io/stargate/api/sql/plan/exec/StatementExecutor.java  3
- sql/server/src/main/java/io/stargate/api/sql/plan/exec/RuntimeContext.java  2
- sql/server/src/main/java/io/stargate/api/sql/plan/rel/CassandraModify.java  7
- sql/server/src/main/java/io/stargate/api/sql/plan/rule/CassandraScanRule.java  1
- sql/server/src/main/java/io/stargate/api/sql/server/StargateMeta.java  3
- sql/server/src/main/java/io/stargate/api/sql/schema/TypeUtils.java  25
- sql/server/src/main/java/io/stargate/api/sql/plan/rule/CassandraPrimaryKeyQueryRule.java  9
- sql/server/src/test/java/io/stargate/api/sql/server/StargateMetaTest.java  1
- sql/server/src/main/java/io/stargate/api/sql/plan/exec/DataSourceStatement.java  2
- sql/server/src/main/java/io/stargate/api/sql/server/AvaticaServer.java  1
- sql/server/src/test/java/io/stargate/api/sql/MapBackedRow.java  2
- sql/server/src/main/java/io/stargate/api/sql/plan/QueryPlanner.java  4
- sql/server/src/main/java/io/stargate/api/sql/plan/exec/NoOpSpakHandler.java  2
- sql/server/src/test/java/io/stargate/api/sql/plan/QueryPlannerTest.java  1
- testing/src/test/java/io/stargate/it/sql/AvaticaJdbcTest.java  1
- sql/server/src/test/java/io/stargate/api/sql/AbstractDataStoreTest.java  1
- sql/server/src/main/java/io/stargate/api/sql/plan/rule/CassandraModifyRule.java  1
- sql/server/src/test/java/io/stargate/api/sql/server/SerializingTestDriver.java  4
- sql/server/src/main/java/io/stargate/api/sql/plan/CalciteUtils.java  1
- sql/server/src/main/java/io/stargate/api/sql/server/SqlActivator.java  7
- sql/server/src/main/java/io/stargate/api/sql/plan/exec/ResultSetEnumerator.java  4
- sql/server/src/main/java/io/stargate/api/sql/plan/PreparedSqlQuery.java  5
- sql/server/src/main/java/io/stargate/api/sql/server/QueryOptionsImpl.java  1
- sql/server/src/main/java/io/stargate/api/sql/plan/rel/CassandraFullScan.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6334590&bid=20579117)

sql/pom.xml
```diff
@@ -0,0 +1,28 @@
...
+  <parent>
+    <groupId>io.stargate</groupId>
+    <artifactId>stargate</artifactId>
+    <version>0.0.16-SNAPSHOT</version>
+  </parent>
+  <groupId>io.stargate.api.sql</groupId>
+  <artifactId>sql-base</artifactId>
+  <packaging>pom</packaging>
+  <properties>
+    <junit.version>5.6.2</junit.version>
```
dougwettlaufer, line 13: You don't want `<junit.version>5.7.0</junit.version>` from the root pom?

sql/pom.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>io.stargate</groupId>
+    <artifactId>stargate</artifactId>
+    <version>0.0.16-SNAPSHOT</version>
+  </parent>
+  <groupId>io.stargate.api.sql</groupId>
+  <artifactId>sql-base</artifactId>
```
dougwettlaufer, line 10: Why not call this just "sql"?

dougwettlaufer: Never mind, I see it now that this is the parent pom for sql

sql/server/pom.xml
```diff
@@ -0,0 +1,113 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>io.stargate.api.sql</groupId>
+    <artifactId>sql-base</artifactId>
+    <version>0.0.16-SNAPSHOT</version>
+  </parent>
+  <groupId>io.stargate.api.sql.server</groupId>
+  <artifactId>stargate-avatica-server</artifactId>
```
dougwettlaufer, line 10: Nit: Maybe drop the "avatica" to avoid leaking implementation and go with `io.stargate.api.sql.server.sql-server` or just `io.stargate.api.sql.server`

sql/server/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.stargate.db</groupId>
+      <artifactId>persistence-api</artifactId>
+      <version>0.0.16-SNAPSHOT</version>
```
dougwettlaufer, line 19: Using this now
```suggestion
      <version>${project.version}</version>
```

sql/server/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+          </dependency>
+        </dependencies>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <configuration>
+          <unpackBundle>true</unpackBundle>
+          <instructions>
+            <Bundle-Name>Stargate-Avatica-Server</Bundle-Name>
```
dougwettlaufer, line 81: Same suggestion down here concerning naming.

sql/sqlline-client/pom.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>io.stargate.api.sql</groupId>
+    <artifactId>sql-base</artifactId>
+    <version>0.0.16-SNAPSHOT</version>
+  </parent>
+  <groupId>io.stargate.api.sql.client</groupId>
+  <artifactId>stargate-sqlline-client</artifactId>
```
dougwettlaufer, line 10: Similar to above, thoughts about `sql-client`?

sql/pom.xml
dimas-b: Missed that on copy-paste :)

sql/server/pom.xml
dimas-b: Yeah, naming is hard :)

Avatica is leaked invariably by us forcing the client to use the Avatica JDBC driver, so I figured we might as well make it explicit in the server artefact name.

In theory we could split the server into the "core" dealing purely with Calcite and one (or more) connectors, the first being Avatica. I just did not want to introduce too many modules in the first commit, plus is it not certain at this point whether we are going to support other wire protocols.

I wanted to use "sql-server" at some point, but did not want it to possibly be confused with the MS SqlServer.

I'll probably drop the "stargate" prefix anyway.

How about `io.stargate.api.sql:avatica-server`

sql/sqlline-client/pom.xml
dimas-b: ok.

sql/server/pom.xml
pcmanus: It has been suggested that one alternative to Avatica would be to implement the [PostgreSQL wire protocol](https://www.postgresql.org/docs/9.3/protocol.html). This is a bit more effort than Avatica server side, but it would advantages in term of number of languages supported (there is more postgresSQL clients than Avatica ones), and possibly in term of clients maturity.

I don't know how and if this impact this particular comment because we probably want to stick with Avatica a bit longer (or ever, really, not saying going the PostgreSQL wire protocol route is a given) for convenience as SQL is far from being production ready, and `avatica-server` is probably just fine here for now. But wanted to add that context.

pcmanus:
Are we sure we want to commit this to `master`? In my mind, SQL is still very much in early stage of development so it would feel cleaner to continue developing it in a long-running dev branch. Granted, it's a tiny bit more work rebasing that branch regularly, but given the nature of this being a separate extension, this should be pretty free.

Typically, publishing the modules here on central feels premature to me (and if we don't, we can probably set aside module naming discussions for now).

sql/server/src/main/java/io/stargate/api/sql/plan/exec/NoOpSpakHandler.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+package io.stargate.api.sql.plan.exec;
+
+import org.apache.calcite.jdbc.CalcitePrepare;
+import org.apache.calcite.linq4j.tree.ClassDeclaration;
+import org.apache.calcite.plan.RelOptPlanner;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.runtime.ArrayBindable;
+
+public class NoOpSpakHandler implements CalcitePrepare.SparkHandler {
```
pcmanus, line 24: Typo: `Spak` -> `Spark`

sql/server/src/main/java/io/stargate/api/sql/plan/rel/CassandraFullScan.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.calcite.linq4j.tree.BlockBuilder;
+import org.apache.calcite.linq4j.tree.Expression;
+import org.apache.calcite.linq4j.tree.Expressions;
+import org.apache.calcite.plan.*;
+import org.apache.calcite.rel.core.TableScan;
+import org.apache.calcite.rel.metadata.RelMetadataQuery;
+import org.apache.calcite.schema.Table;
+
+/** Relational operation representing a full Cassandra table scan. */
+public class CassandraFullScan extends TableScan implements EnumerableRel {
```
pcmanus, line 33: I'd avoid the `Cassandra` prefix (here and other similar places), since storage is supposedly somewhat abstracted (I mean, storage pretty much has to be CQL-based, but it doesn't have to be "Cassandra"). Imo, `FullScan` is probably good enough, though if we're worried this might be confused by a "SQL full scan", maybe `StorageFullScan`.

sql/server/src/main/java/io/stargate/api/sql/schema/StargateSchema.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Collections;
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.apache.calcite.linq4j.tree.Expression;
+import org.apache.calcite.rel.type.RelProtoDataType;
+import org.apache.calcite.schema.*;
+import org.immutables.value.Value;
+
+@Value.Immutable(prehash = true)
+public abstract class StargateSchema implements Schema {
```
pcmanus, line 28: I think the naming is confusing, because stargate already has a `Schema` (in `persistence-api`). As this exists (afaict) only because Calcite wants _its_ version of the schema, I'd call it `CalciteSchema`, but `SQLSchema` probably works as well (avoids any confusion).

sql/server/src/main/java/io/stargate/api/sql/server/StargateStatement.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import org.apache.calcite.adapter.java.JavaTypeFactory;
+import org.apache.calcite.avatica.AvaticaParameter;
+import org.apache.calcite.avatica.ColumnMetaData;
+import org.apache.calcite.avatica.Meta;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeField;
+
+public class StargateStatement extends Meta.StatementHandle {
```
pcmanus, line 34: Here again, since stargate is full of stuff called "statement", I think it's worth finding a name more useful. In general, I'd argue that putting "stargate" into any naming is rarely gonna provide useful information.

pcmanus (commented):
Did a very quick scan, and left a few minor remarks that came to me. And from a very high level, this clearly look like a good start. A careful code review would take me much more time, but as hinted in my prior comment, I'd be ok keeping review light on the assumption this is an early POC and that potential early rough edges are likely going to be ironed out incrementally anyway.

sql/server/pom.xml
dimas-b: Right now, there's no clean line between Stargate and Avatica because Calcite itself has ties to Avatica in many places. For example, we have to produce values of certain types in result sets for Avatica to be able to serialize them correctly, even though we do not explicitly depend on Avatica at that level. It should certainly be possible to support multiple drivers, but at this stage I'm just not sure it is worth the effort.

sql/server/src/main/java/io/stargate/api/sql/plan/rel/CassandraFullScan.java
dimas-b: Yeah, "Cassandra" does feel awkward in this context. At the same time we certainly rely of C* storage model (PK/CC, etc.) in these classes. I'm fine with `StorageFullScan`.

sql/server/src/main/java/io/stargate/api/sql/schema/StargateSchema.java
dimas-b: Calcite already has a `CalciteSchema` in its JDBC packages :thinking: I'll go with `SQLSchema`

dougwettlaufer:


```diff

Issues
======
- Added 47
           

Complexity increasing per file
==============================
- sql/server/src/main/java/io/stargate/api/sql/plan/rule/CassandraModifyRule.java  1
- sql/server/src/main/java/io/stargate/api/sql/plan/rel/CassandraFullScan.java  1
- sql/server/src/test/java/io/stargate/api/sql/AbstractDataStoreTest.java  1
- sql/server/src/test/java/io/stargate/api/sql/server/StargateMetaTest.java  1
- sql/server/src/main/java/io/stargate/api/sql/plan/exec/StatementExecutor.java  3
- sql/server/src/main/java/io/stargate/api/sql/plan/rel/CassandraSingleRowQuery.java  2
- testing/src/test/java/io/stargate/it/sql/AvaticaJdbcTest.java  1
- sql/server/src/main/java/io/stargate/api/sql/plan/rule/CassandraScanRule.java  1
- sql/server/src/main/java/io/stargate/api/sql/schema/StargateTable.java  1
- sql/server/src/main/java/io/stargate/api/sql/server/SqlActivator.java  7
- sql/server/src/main/java/io/stargate/api/sql/schema/StargateSchema.java  2
- sql/server/src/main/java/io/stargate/api/sql/server/AvaticaServer.java  1
- sql/server/src/main/java/io/stargate/api/sql/plan/QueryPlanner.java  4
- sql/server/src/main/java/io/stargate/api/sql/plan/rule/CassandraPrimaryKeyQueryRule.java  9
- sql/server/src/main/java/io/stargate/api/sql/plan/PreparedSqlQuery.java  5
- sql/server/src/main/java/io/stargate/api/sql/plan/CalciteUtils.java  1
- sql/server/src/main/java/io/stargate/api/sql/plan/exec/ResultSetEnumerator.java  4
- sql/server/src/main/java/io/stargate/api/sql/server/StargateMeta.java  3
- sql/server/src/test/java/io/stargate/api/sql/plan/QueryPlannerTest.java  1
- sql/server/src/main/java/io/stargate/api/sql/plan/exec/RuntimeContext.java  2
- sql/server/src/main/java/io/stargate/api/sql/schema/TypeUtils.java  25
- sql/server/src/main/java/io/stargate/api/sql/server/StargateStatement.java  4
- sql/server/src/test/java/io/stargate/api/sql/server/SerializingTestDriver.java  4
- sql/server/src/test/java/io/stargate/api/sql/ListBackedResultSet.java  2
- sql/server/src/test/java/io/stargate/api/sql/ValidatingDataStore.java  2
- sql/server/src/main/java/io/stargate/api/sql/server/StargateConnection.java  3
- sql/server/src/test/java/io/stargate/api/sql/MapBackedRow.java  2
- sql/server/src/main/java/io/stargate/api/sql/plan/exec/NoOpSpakHandler.java  2
- sql/server/src/main/java/io/stargate/api/sql/plan/exec/DataSourceStatement.java  3
- sql/server/src/main/java/io/stargate/api/sql/plan/rel/CassandraModify.java  7
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6334590&bid=20579117)

sql/server/src/main/java/io/stargate/api/sql/plan/rel/CassandraFullScan.java
pcmanus: >  we certainly rely on C* storage model (PK/CC, etc.) in these classes

Right, but that's what Stargate does. I mean, `Persistence` is the core of Stargate, so something unqualified like `FullScan` would imo naturally imply in the reader mind "a full scan related to the persistence storage". But I'm happy with adding `Storage`.

sql/server/pom.xml
pcmanus: > at this stage I'm just not sure it is worth the effort

Everything's in the definition of "this stage" ;).

As a said, I agree we should ride on Avatica for a bit longer, because before worrying what exact client driver is better for users, we need to make the whole thing usable.

But the more I think about it, the more I think switching to the PostgreSQL wire protocol is something we should seriously consider (again, in due time), because it will give use substantially more clients. Including things like ODBC, which opens quite a few doors.

But again, not something for this PR. Just to keep in mind that if we see opportunities to avoid tying things too much to Avatica as we work on this moving on, then it's worth a bit of extra effort.

sql/server/src/main/java/io/stargate/api/sql/plan/rel/CassandraFullScan.java
dimas-b: Another aspect of this is that these classes show up in `EXPLAIN PLAN`. We can probably keep `FullScan` as the class name, but it may be too ambiguous for the plan explanation... I'll see if we have a way to override what name is used in `EXPLAIN PLAN`.

sql/server/pom.xml
dimas-b: By "this stage" I meant "this PR" :)

+1 to trying the PostgreSQL protocol.

sql/pom.xml
dimas-b: fixed in the latest push

sql/sqlline-client/pom.xml
dimas-b: renamed to `sql-shell` because it is really a shell :)

sql/server/pom.xml
dimas-b: renamed artifact to `io.stargate.api.sql:avatica-server` 

and bundle to `Avatica-Server` with symbolic name `io.stargate.api.sql.server.avatica`

sql/server/src/main/java/io/stargate/api/sql/plan/rel/CassandraFullScan.java
dimas-b: A did some renaming in these and related classes. Please have another look.

As for the "Stargate" prefix, the only remaining instance is `StargateMeta`... I really could not find a better fit. `MetaImpl` is already in its inheritance chain, plus it's a bit terse, and I wanted to emphasize that this impl. is a "Stargate" way of integrating with Avatica.

"Cassandra" prefixes are gone. That actually makes `EXPLAIN PLAN` output more readable I think (we already have physical table names there, so the association with "storage" is pretty clear.. see `QueryPlannerTest` for examples, if interested).

sql/server/src/main/java/io/stargate/api/sql/schema/StargateSchema.java
dimas-b: renamed to `SqlSchema`

sql/server/src/main/java/io/stargate/api/sql/server/StargateStatement.java
dimas-b: renamed to `StatementHolder`

dimas-b:
@pcmanus if you're ok with using this as the first cut of the SQL module on the feature branch, could you please approve to enable the merge button?

pcmanus (approved):
Add a quick look. Changes lgtm and I'm happy to merge that as the beginning of our sql branch!

PR #276, opened 2020-10-14
Make `replicas` always default to 1 when missing, for SimpleStrategy

EricBorczuk:
Seems that when NetworkTopologyStrategy is chosen in the REST API, a default of 3 is always chosen when `replicas` is missing, but no default was supplied for the SimpleStrategy.  This PR adds that default of 1.

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/NamespacesResource.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/NamespacesResource.java
@@ -173,10 +173,10 @@ public class NamespacesResource {
           String token,
       @ApiParam(
               value =
-                  "A map representing a namespace with SimpleStrategy or NetworkTopologyStrategy \n"
+                  "A map representing a namespace with SimpleStrategy or NetworkTopologyStrategy with default replicas of 1 and 3 respectively \n"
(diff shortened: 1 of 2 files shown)


restapi/src/main/java/io/stargate/web/docsapi/resources/NamespacesResource.java
```diff
@@ -176,20 +176,19 @@ public Response createNamespace(
...
                       + "  \"name\": \"killrvideo\",\n"
                       + "   \"datacenters\":\n"
                       + "      [\n"
-                      + "         { \"name\": \"dc1\", \"replicas\": 3 },\n"
+                      + "         { \"name\": \"dc1\", \"replicas\": 3 // default },\n"
                       + "         { \"name\": \"dc2\", \"replicas\": 3 },\n"
                       + "      ],\n"
                       + "}\n"
-                      + "```",
-              defaultValue = "false")
```
EricBorczuk, line 192: idk why the default value of false was here, I think it was a mistake

dougwettlaufer: Yeah probably a bad copy paste.

PR #282, opened 2020-10-14
Unauthorized response

EricBorczuk:
Because I wrapped the POST in two `handle`s inadvertently, the inner one bubbles up a Response of Unauthorized, but then that response got ignored and a 201 was returned.  This fixes that.

Technically this was not a mistake with authz, since the action was never performed when a token was not supplied; however, this put the fear of god in me and so I added some tests to test Unauthorized responses in the docs API


restapi/src/test/java/io/stargate/web/docsapi/resources/DocumentResourceV2Test.java
```diff
@@ -45,151 +45,10 @@
...
-        DocumentResourceV2.class.getDeclaredMethod("wrapResponse", JsonNode.class, String.class);
-    wrapResponseId.setAccessible(true);
-    wrapResponseIdPageState =
-        DocumentResourceV2.class.getDeclaredMethod(
-            "wrapResponse", JsonNode.class, String.class, String.class);
-    wrapResponseIdPageState.setAccessible(true);
-  }
-
-  @Test
-  public void wrapResponse_nodeIdDefined()
```
EricBorczuk, line 63: Mm I think unit tests aren't running in CI...this didn't run properly because `wrapResponse` was removed in a previous PR.

testing/src/test/java/io/stargate/it/http/DocumentApiV2Test.java
```diff
@@ -2139,65 +2256,70 @@ public void deleteNamespace() throws IOException {
...
+      request = request.addHeader("X-Cassandra-Token", authToken);
+    }
 
-    return client.newCall(request).execute();
+    return client.newCall(request.build()).execute();
   }
 
-  private Response patch(String path, Object arg) throws IOException {
-    Request request =
+  private Response patch(String path, Object arg, boolean authorize) throws IOException {
```
dimas-b, line 2299: Could we keep the old method signature as an overload to avoid the many change with `true` for the new parameter?

EricBorczuk: yeaaa I was thinking about doing that

Will do :)

testing/src/test/java/io/stargate/it/http/DocumentApiV2Test.java
```diff
@@ -138,24 +138,55 @@ public void testIt() throws IOException {
...
+            true);
     assertThat(r.code()).isEqualTo(200);
     ObjectNode expected = objectMapper.createObjectNode();
     expected.set("1", obj);
     assertThat(objectMapper.readTree(r.body().string()))
         .isEqualTo(wrapResponse(expected, null, null));
   }
 
+  @Test
+  public void testUnauthorized() throws IOException {
```
dougwettlaufer, line 151: What about a token that isn't valid?

EricBorczuk: Ah yea, fair - I'll add that

restapi/src/test/java/io/stargate/web/docsapi/resources/DocumentResourceV2Test.java
dimas-b: It's because it uses JUnit 4, which is not supported in SG now

dimas-b: So try converting to JUnit 5 :)

EricBorczuk: oh noooooo

alright - I'm not sure if powermock is supported by junit 5 but I sure hope so

EricBorczuk: I'll take a look at this tomorrow

testing/src/test/java/io/stargate/it/http/DocumentApiV2Test.java
EricBorczuk: I added an overloading for each with signature `String path, Object arg, String token`

So that I could test garbage tokens too (if it's null, the header will get left off)

dougwettlaufer (approved):

dougwettlaufer:


```diff

Issues
======
+ Solved 12
           
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6341495&bid=20598607)

restapi/src/test/java/io/stargate/web/docsapi/resources/DocumentResourceV2Test.java
dimas-b: powermock should be avoided in the first place IMHO :)

PR #284, opened 2020-10-16
Support logged batches in GraphQL

jorgebay:
Resolves #133 .

It should be rebased once #271 and #279 are merged.

Issue #256, opened 2020-10-12
Improve error handling when building GraphQL schema


olim7t:
Error handling leaves a lot to be desired. For example, the code sometimes swallows exceptions:
```java
try {
  additionalTypes = buildTypesForTable(table);
  tableQueryField = buildQuery(table);
  tableMutationFields = buildMutations(table);
} catch (Exception e) {
  log.warn("Skipping table " + table.name());
  continue;
}
```
Also, end users will not always have access to the logs. We should find a way to surface something to them: not necessarily the exact error but at least that something went wrong for table X or column Y. I'm not sure what the best approach would be, maybe a special GraphQL query that would return a list of errors.

PR #216, opened 2020-10-07
Revisit GraphQL naming conventions (fixes #118)

olim7t:
Name GraphQL types directly after tables. Only change them if there are
special characters, or they clash with our internal type names.

Fix case-sensitivity in generated CQL queries.

Fixes #118
Fixes #176
Fixes #365

PR #286, opened 2020-10-16
GraphQL: add collection filters

olim7t:
Early draft, but I want to give a preview of the field cache refactoring. It will greatly simplify this ticket, as well as future ones like tuple support.


dougwettlaufer:


```diff

Issues
======
+ Solved 3
- Added 4
           

Complexity increasing per file
==============================
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldOutputTypeCache.java  9
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldInputTypeCache.java  9
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldFilterInputTypeCache.java  4
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldTypeCache.java  22
- graphqlapi/src/test/java/io/stargate/graphql/schema/FieldTypeCachesTest.java  1
         

Complexity decreasing per file
==============================
+ graphqlapi/src/main/java/io/stargate/graphql/schema/DmlSchemaBuilder.java  -23
         

Clones added
============
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldOutputTypeCache.java  1
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldInputTypeCache.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6353286&bid=20628486)

dougwettlaufer:


```diff

Issues
======
+ Solved 3
- Added 8
           

Complexity increasing per file
==============================
- graphqlapi/src/test/java/io/stargate/graphql/schema/SchemaAssertions.java  3
- graphqlapi/src/test/java/io/stargate/graphql/schema/fetchers/dml/DeleteMutationFetcherCollectionsTest.java  1
- graphqlapi/src/test/java/io/stargate/graphql/schema/fetchers/dml/QueryFetcherCollectionsTest.java  1
- graphqlapi/src/test/java/io/stargate/graphql/schema/DmlSchemaUdtTest.java  1
- graphqlapi/src/test/java/io/stargate/graphql/schema/DmlSchemaCollectionsTest.java  1
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldFilterInputTypeCache.java  6
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldOutputTypeCache.java  7
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldInputTypeCache.java  7
- graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/DmlFetcher.java  4
- graphqlapi/src/test/java/io/stargate/graphql/schema/FieldFilterInputTypeCacheTest.java  1
- graphqlapi/src/test/java/io/stargate/graphql/schema/FieldTypeCachesTest.java  1
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldTypeCache.java  22
         

Complexity decreasing per file
==============================
+ graphqlapi/src/main/java/io/stargate/graphql/schema/DmlSchemaBuilder.java  -23
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6353286&bid=20628486)

dougwettlaufer:


```diff

Issues
======
+ Solved 3
- Added 8
           

Complexity increasing per file
==============================
- graphqlapi/src/test/java/io/stargate/graphql/schema/fetchers/dml/QueryFetcherCollectionsTest.java  1
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldInputTypeCache.java  7
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldOutputTypeCache.java  7
- graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/DmlFetcher.java  4
- graphqlapi/src/test/java/io/stargate/graphql/schema/FieldTypeCachesTest.java  1
- graphqlapi/src/test/java/io/stargate/graphql/schema/DmlSchemaCollectionsTest.java  1
- graphqlapi/src/test/java/io/stargate/graphql/schema/FieldFilterInputTypeCacheTest.java  1
- graphqlapi/src/test/java/io/stargate/graphql/schema/fetchers/dml/DeleteMutationFetcherCollectionsTest.java  1
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldTypeCache.java  22
- graphqlapi/src/test/java/io/stargate/graphql/schema/DmlSchemaUdtTest.java  1
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldFilterInputTypeCache.java  6
- graphqlapi/src/test/java/io/stargate/graphql/schema/SchemaAssertions.java  3
         

Complexity decreasing per file
==============================
+ graphqlapi/src/main/java/io/stargate/graphql/schema/DmlSchemaBuilder.java  -23
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6353286&bid=20628486)

graphqlapi/src/main/java/io/stargate/graphql/schema/types/GqlMapBuilder.java
```diff
@@ -54,19 +47,10 @@ private GraphQLType buildKeyValueType() {
     String keyTypeName = getTypeName(keyType);
     String valueTypeName = getTypeName(valueType);
     String name =
-        String.format("%sKey%sValue%s", isInput ? "Input" : "", keyTypeName, valueTypeName);
+        String.format("Entry%sKey%sValue%s", keyTypeName, valueTypeName, isInput ? "Input" : "");
```
olim7t, line 50: I revisited naming here for a few reasons:
* `Input` at the end to match what we already did for tables
* `Key` and `Value` _after_ their respective types, to be closer to natural language: "an entry with an int key and a string value".
* additional `Entry` marker at the beginning, to avoid a corner case where  two different FilterInputs ended up using the same name. See `FieldFilterInputTypeCacheTest`.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/DmlFetcher.java
```diff
@@ -101,6 +101,21 @@ protected DmlFetcher(
...
+          relation = relationStart.contains(toCqlValueTerm(column, condition.getValue()));
+        } else if (condition.getKey().equals("containsEntry")) {
+          Column.ColumnType mapType = column.type();
+          assert mapType != null && mapType.isMap();
+          Map<String, Object> entry = (Map<String, Object>) condition.getValue();
+          Column.ColumnType keyType = mapType.parameters().get(0);
+          Term keyTerm = toCqlTerm(keyType, entry.get("key"));
+          Column.ColumnType valueType = mapType.parameters().get(1);
+          Term valueTerm = toCqlTerm(valueType, entry.get("value"));
+          relation = Relation.mapValue(column.name(), keyTerm).isEqualTo(valueTerm);
```
olim7t, line 118: There's a `FilterOperator` enum screaming to be extracted here, and reused in `FieldFilterInputTypeCache`. But this PR is already big, I'll probably keep it for a follow-up ticket.

olim7t (commented):
Collection filters are in, this is now ready for review.

dougwettlaufer:


```diff

Issues
======
+ Solved 3
- Added 8
           

Complexity increasing per file
==============================
- graphqlapi/src/test/java/io/stargate/graphql/schema/DmlSchemaUdtTest.java  1
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldInputTypeCache.java  7
- graphqlapi/src/test/java/io/stargate/graphql/schema/fetchers/dml/QueryFetcherCollectionsTest.java  1
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldFilterInputTypeCache.java  6
- graphqlapi/src/test/java/io/stargate/graphql/schema/fetchers/dml/DeleteMutationFetcherCollectionsTest.java  1
- graphqlapi/src/test/java/io/stargate/graphql/schema/SchemaAssertions.java  3
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldOutputTypeCache.java  7
- graphqlapi/src/test/java/io/stargate/graphql/schema/FieldFilterInputTypeCacheTest.java  1
- graphqlapi/src/test/java/io/stargate/graphql/schema/DmlSchemaCollectionsTest.java  1
- graphqlapi/src/test/java/io/stargate/graphql/schema/FieldTypeCachesTest.java  1
- graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/DmlFetcher.java  4
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldTypeCache.java  22
         

Complexity decreasing per file
==============================
+ graphqlapi/src/main/java/io/stargate/graphql/schema/DmlSchemaBuilder.java  -23
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6353286&bid=20628486)

graphqlapi/src/main/java/io/stargate/graphql/schema/FieldInputTypeCache.java
```diff
@@ -0,0 +1,69 @@
...
+        GraphQLInputObjectType.newInputObject().name(nameMapping.getUdtNames().get(udt) + "Input");
+    for (Column column : udt.columns()) {
+      try {
+        builder.field(
+            GraphQLInputObjectField.newInputObjectField()
+                .name(nameMapping.getFieldNames(udt).get(column))
+                .type(get(column.type()))
+                .build());
+      } catch (Exception e) {
+        // TODO find a better way to surface errors
```
dougwettlaufer, line 63: Did you mean to swallow the error here?

graphqlapi/src/main/java/io/stargate/graphql/schema/FieldFilterInputTypeCache.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  @Override
+  protected GraphQLInputType compute(Column.ColumnType columnType) {
+    GraphQLInputType gqlInputType = inputTypeCache.get(columnType);
+    if (columnType.isMap()) {
+      return mapFilter(gqlInputType);
+    } else if (columnType.isList() || columnType.isSet()) {
+      return listFilter(gqlInputType);
+    } else {
```
dougwettlaufer, line 54: We don't need special UDT handling over here?

graphqlapi/src/main/java/io/stargate/graphql/schema/FieldOutputTypeCache.java
```diff
@@ -0,0 +1,69 @@
...
+        GraphQLObjectType.newObject().name(nameMapping.getUdtNames().get(udt));
+    for (Column column : udt.columns()) {
+      try {
+        builder.field(
+            new GraphQLFieldDefinition.Builder()
+                .name(nameMapping.getFieldNames(udt).get(column))
+                .type(get(column.type()))
+                .build());
+      } catch (Exception e) {
+        // TODO find a better way to surface errors
```
dougwettlaufer, line 63: Same, okay to swallow this?

graphqlapi/src/main/java/io/stargate/graphql/schema/FieldTypeCache.java
```diff
@@ -0,0 +1,95 @@
+package io.stargate.graphql.schema;
```
dougwettlaufer, line 1: Should these 4 classes maybe go under a `cache` package?

graphqlapi/src/test/java/io/stargate/graphql/schema/DmlSchemaCollectionsTest.java
```diff
@@ -0,0 +1,61 @@
...
+    // CQL sets are translated to GraphQL lists, so these tables reuse the same type too:
+    assertField(graphQlSchema.getType("PkSetTableFilterInput"), "s", listFilterInput);
+    assertField(graphQlSchema.getType("RegularSetTableFilterInput"), "s", listFilterInput);
+  }
+
+  @Test
+  @DisplayName("Should surface map types")
+  public void mapTest() {
+    GraphQLInputObjectType mapFilterInput =
+        ((GraphQLInputObjectType) graphQlSchema.getType("ListEntryIntKeyStringValueFilterInput"));
```
dougwettlaufer, line 38: > ListEntry

Isn't this a map? Or am I reading this wrong?

dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/FieldInputTypeCache.java
olim7t: That's what the TODO is about. I did it that way because that's how it's already done everywhere, but it's bad to swallow the exception.
The problem is that those errors happen while parsing a keyspace at startup or in response to a schema change, not during a user request, so it's not as simple as just rethrowing an exception. See also #256.

graphqlapi/src/main/java/io/stargate/graphql/schema/FieldTypeCache.java
olim7t: Yes, the root package is starting to get crowded, I was thinking of some reorganizing. Maybe something like `schema/dml` and `schema/ddl`, each with their own `fetchers` subpackage and maybe others.

graphqlapi/src/test/java/io/stargate/graphql/schema/DmlSchemaCollectionsTest.java
olim7t: There is no map type in GraphQL, we represent them as lists of entries. For this particular example it looks like this:
```
// create table pk_map_table(m frozen<map<int,text>>)

type EntryIntKeyStringValue {
  key: Int!
  value: String
}

// Output type for results
type PkMapTable {
  m: [EntryIntKeyStringValue]
}

// Filter type for queries
input ListEntryIntKeyStringValueFilterInput {
  containsEntry: EntryIntKeyStringValueInput
  containsKey: Int
  containsValue: String
  eq: [EntryIntKeyStringValueInput]
  gt: [EntryIntKeyStringValueInput]
  gte: [EntryIntKeyStringValueInput]
  in: [[EntryIntKeyStringValueInput]]
  lt: [EntryIntKeyStringValueInput]
  lte: [EntryIntKeyStringValueInput]
  notEq: [EntryIntKeyStringValueInput]
}
```
The code that generates the filter name (`FieldFilterInputTypeCache.format()`) is not aware of those details, it just sees a "list of something". I think it would be possible to special case that to say "Map" instead of "List", but it seems low value, the name of the type should not be directly relevant to end users.

graphqlapi/src/main/java/io/stargate/graphql/schema/FieldFilterInputTypeCache.java
olim7t: Right now UDT filters look just like scalar filters:
```
input AUdtFilterInput {
  eq: AUdtInput
  gt: AUdtInput
  gte: AUdtInput
  in: [AUdtInput]
  lt: AUdtInput
  lte: AUdtInput
  notEq: AUdtInput
}
input IntFilterInput {
  eq: Int
  gt: Int
  gte: Int
  in: [Int]
  lt: Int
  lte: Int
  notEq: Int
}
```
So there is no special logic, we just pull an existing "simple" type from `inputTypeCache` and build upon it. 

That being said, I think there might be a UDT field predicate `WHERE a.i = 1` at least in Cassandra 4. I'm going to check and if that's the case I'll extend the basic filter like for collections.

olim7t: Hrmm yeah, field predicates are a thing in Cassandra 4 and DSE, they're not allowed in WHERE clauses but they are valid in IF conditions:
```
cqlsh:test> CREATE TYPE t(i int, s text);
cqlsh:test> CREATE TABLE foo(k int PRIMARY KEY, t t);
cqlsh:test> DELETE FROM foo WHERE k=1 IF t.i=2;
```
We can extend the UDT filter with a nested filter for each field:
```
input TUdtFilterInput {
  ... basic operators
  i: IntFilterInput
  s: StringFilterInput
}
```
Which would allow us to write something like:
```
deleteFoo(
  value: {k: 1},
  ifCondition: { t: { i: { eq: 2 } }
)
```
One catch is that a UDT could have fields with names that clash with basic operators: "eq", "gt", etc. We can add another level of nesting `{ t: { field: { i: { eq: 2 } } }`, but that complicates the syntax and we'll have to generate another graphql type.

graphqlapi/src/main/java/io/stargate/graphql/schema/DmlSchemaBuilder.java
```diff
@@ -63,61 +57,29 @@
...
     this.persistence = persistence;
     this.authenticationService = authenticationService;
     this.keyspace = keyspace;
 
     this.nameMapping = new NameMapping(keyspace.tables(), keyspace.userDefinedTypes());
-    this.filterInputTypes = buildFilterInputTypes();
+    this.fieldInputTypes = new FieldInputTypeCache(this.nameMapping);
+    this.fieldOutputTypes = new FieldOutputTypeCache(this.nameMapping);
+    this.fieldFilterInputTypes =
+        new FieldFilterInputTypeCache(this.fieldInputTypes, this.nameMapping);
```
jorgebay, line 77: I think we have to consider that names should be "granted" or generated by the `NameMapping` instance.

Given that UDTs, map entries, tables GraphQL names can collide, a central entity generating those names in a deterministic way (i.e. UDTs first, then maps and last tables), with suffixes and prefixes included.

We did something like that on the golang project: see [here](https://github.com/datastax/cassandra-data-apis/blob/0f24c6a5f6ca28c0086c49ad0ce2997994e7f841/config/naming.go#L72), [here](https://github.com/datastax/cassandra-data-apis/blob/0f24c6a5f6ca28c0086c49ad0ce2997994e7f841/config/naming.go#L63) and [this unit test](https://github.com/datastax/cassandra-data-apis/blob/0f24c6a5f6ca28c0086c49ad0ce2997994e7f841/config/naming_test.go#L50-L61) for example .

I see a strong relationship between naming and the field cache you propose in this PR, so maybe it makes sense to address them in one big bang, i.e. go over DB objects, generate names and graphql (input/output) types for them.

jorgebay (commented):
Some remarks below.

There are some conflicts between this PR and #284, so let me know if you want to land this one first, I can rebase mine.

graphqlapi/src/main/java/io/stargate/graphql/schema/FieldTypeCache.java
```diff
@@ -0,0 +1,95 @@
...
+
+  GraphqlT get(ColumnType type) {
+    type = normalize(type);
+    return types.computeIfAbsent(type, this::compute);
+  }
+
+  private ColumnType normalize(ColumnType type) {
+    // Frozen-ness does not matter. We want frozen and non-frozen versions of a CQL type to be
+    // mapped to the same GraphQL type.
+    type = type.frozen(false);
```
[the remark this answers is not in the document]
olim7t: `frozen()` does not modify the type in place, it creates a copy. So this will only affect the cache key.

Currently we generate the same GraphQL type whether the CQL type is frozen or not, so we have to do this otherwise we have duplicate types and GraphQL will crash (try to comment this line, you'll see that some unit tests break).

If frozen-ness starts to leak in the GraphQL schema, we'll have to modify the caches anyway, so we'll consider what information to add in the key at that moment.

graphqlapi/src/main/java/io/stargate/graphql/schema/FieldTypeCache.java
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Caches a category of GraphQL field types, corresponding to table columns or UDT fields.
+ *
+ * <p>There are different categories, each implemented by a subclass. Note that this cache does not
+ * contain table types, they are handled as top-level entities in {@link DmlSchemaBuilder}.
+ *
+ * @param <GraphqlT> the returned GraphQL type.
+ */
+abstract class FieldTypeCache<GraphqlT> {
```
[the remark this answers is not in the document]
olim7t: I generally follow [those rules](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case) for naming. Camel-casing acronyms gives the most consistent results if you have multiple of them in a row.

But indeed the GraphQL library doesn't, so maybe we should be consistent with it, I'm not sure what's more important...

graphqlapi/src/main/java/io/stargate/graphql/schema/FieldFilterInputTypeCache.java
```diff
@@ -0,0 +1,140 @@
...
+    super(nameMapping);
+    this.inputTypeCache = inputTypeCache;
+  }
+
+  @Override
+  protected GraphQLInputType compute(Column.ColumnType columnType) {
+    GraphQLInputType gqlInputType = inputTypeCache.get(columnType);
+    if (columnType.isMap()) {
+      return mapFilter(gqlInputType);
+    } else if (columnType.isList() || columnType.isSet()) {
```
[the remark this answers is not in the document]
olim7t: Really a matter of taste, but FWIW I find that the `else` emphasize that we have 3 different branches. For example if I forgot the `return` in the second branch, the code would not compile, instead of failing silently.

graphqlapi/src/main/java/io/stargate/graphql/schema/DmlSchemaBuilder.java
olim7t: > I think we have to consider that names should be "granted" or generated by the NameMapping instance

They already are, at least the parts that are relevant to the CQL->GraphQL conversion. The fact that we have "Input" and "FilterInput" types, and generate "EntryXKeyYValue" types for maps, are purely specific to GraphQL, so I'm less concerned having these rules directly in the GraphQL code. But if you prefer to extract them into utility methods, I'm not opposed either.

> UDTs, map entries, tables GraphQL names can collide

There are different levels of collision:
1. two tables `foo_bar` and `"FooBar"` can collide, that's the discussion I started in #216
2. a table `foo_bar` and a UDT `foo_bar` do not collide, thanks to the `Udt` suffix.
3. I think the current naming rules do not allow different CQL column types to collide, see [this above comment](https://github.com/stargate/stargate/pull/286#discussion_r506771839).
4. then there are the corner cases where someone might name a table `foo_bar_udt` or `entry_int_key_string_value`, I think we can skip those and log a warning somewhere.

> We did something like that on the golang project

Note that the strategy used in that  project is not resilient to schema changes, one thing to keep in mind if we adapt it.

graphqlapi/src/main/java/io/stargate/graphql/schema/FieldFilterInputTypeCache.java
```diff
@@ -0,0 +1,140 @@
...
+    // - when the map column is the primary key
+    GraphQLInputObjectType.Builder builder = basicComparisons(gqlInputType);
+
+    // The 'contains*' operators are allowed for regular columns with an SAI index
+    // (some of them require ALLOW FILTERING depending on whether the index is on keys(m), values(m)
+    // or entries(m), so if we enable the operators depending on the context, we might want to
+    // consider that).
+    builder.field(GraphQLInputObjectField.newInputObjectField().name("containsKey").type(keyType));
+    builder.field(
+        GraphQLInputObjectField.newInputObjectField().name("containsValue").type(valueType));
```
olim7t, line 79: Note to self: rename this to "contains" to align with the CQL syntax.

graphqlapi/src/main/java/io/stargate/graphql/schema/FieldTypeCache.java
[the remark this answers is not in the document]
jorgebay: oh snap! I haven't checked that `frozen()` method returned a new type instance.
Thanks for the explanation, nvm then.

graphqlapi/src/main/java/io/stargate/graphql/schema/DmlSchemaBuilder.java
jorgebay: I've posted a new comment on the naming discussion https://github.com/stargate/stargate/pull/216#issuecomment-713408183 .

This is not directly related to this change, IMO a refactoring like this would have been an opportunity to also tackle cross cutting concept like graphql naming. Otherwise, it's very likely than a new refactor is needed to deliver that feature.

mpenick (approved):

dougwettlaufer:


```diff

Issues
======
+ Solved 3
- Added 5
           

Complexity increasing per file
==============================
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldInputTypeCache.java  7
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldOutputTypeCache.java  7
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldTypeCache.java  22
- graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/DmlFetcher.java  3
- graphqlapi/src/main/java/io/stargate/graphql/schema/FieldFilterInputTypeCache.java  6
         

Complexity decreasing per file
==============================
+ graphqlapi/src/main/java/io/stargate/graphql/schema/DmlSchemaBuilder.java  -23
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6353286&bid=20628486)

jorgebay:
We forgot to update the `schema.json` and integration tests in this PR, I'll fix that in the scalars PR.

PR #287, opened 2020-10-16
Re-login after creating table on the fly to refresh permissions

EricBorczuk:
When you create a table and then immediately create an index, the permissions for the user aren't necessarily updated in time to reflect the new table's creation (a cache is hit instead).  This PR forces a re-login after the table creation is complete to force refreshing permissions


dougwettlaufer:


```diff

Complexity increasing per file
==============================
- restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6354483&bid=20631506)

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -126,9 +129,23 @@ public void maybeCreateTable(String keyspaceName, String tableName) {
                   String.join(", ", allPathColumnNames)))
           .get();
 
-      dataStore
-          .query(String.format("CREATE INDEX ON \"%s\".\"%s\"(leaf)", keyspaceName, tableName))
-          .get();
+      // After creating the table, it can take up to 2 seconds for permissions cache to be updated,
```
dimas-b, line 132: Would it make sense to introduce a new method into `persistence-api` to allow waiting for this intelligently or perhaps to receive a notification of some sort?

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -126,9 +129,23 @@ public void maybeCreateTable(String keyspaceName, String tableName) {
...
+      // So retry this first index creation until it succeeds.  The other index creations should not
+      // need this behavior.
+      boolean success = false;
+      while (!success) {
```
dimas-b, line 136: This had the potential to cause an infinite look if the index creation fails for an non-transient reason.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
EricBorczuk: yes, that's a really good point... exposing some method that creates a table and blocks until permissions are updated would be really good

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
EricBorczuk: This diff is actually inside a bigger `try` block - which exceptions would cause an infinite loop?

EricBorczuk: either way though it probably makes sense to only retry a finite number of times

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
EricBorczuk: I have no idea how I'd do that though ^_^

dimas-b: Let's file an issue :)

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
dimas-b: > which exceptions would cause an infinite loop?

those that we catch on line 143 and ignore.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -126,9 +129,23 @@ public void maybeCreateTable(String keyspaceName, String tableName) {
...
+        try {
+          Uninterruptibles.sleepUninterruptibly(2, TimeUnit.SECONDS);
+          dataStore
+              .query(String.format("CREATE INDEX ON \"%s\".\"%s\"(leaf)", keyspaceName, tableName))
+              .get();
+          success = true;
+        } catch (ExecutionException e) {
```
dimas-b, line 143: What do you think about logging a debug message for this exception?

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
EricBorczuk: true, duh - thanks

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -126,9 +129,23 @@ public void maybeCreateTable(String keyspaceName, String tableName) {
...
+      // After creating the table, it can take up to 2 seconds for permissions cache to be updated,
+      // So retry this first index creation until it succeeds.  The other index creations should not
+      // need this behavior.
+      boolean success = false;
+      while (!success) {
+        try {
+          Uninterruptibles.sleepUninterruptibly(2, TimeUnit.SECONDS);
+          dataStore
+              .query(String.format("CREATE INDEX ON \"%s\".\"%s\"(leaf)", keyspaceName, tableName))
+              .get();
```
tomekl007, line 141: consider passing a timeout duration to `get()` because it may block indefinitely.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -126,9 +129,23 @@ public void maybeCreateTable(String keyspaceName, String tableName) {
...
-      dataStore
-          .query(String.format("CREATE INDEX ON \"%s\".\"%s\"(leaf)", keyspaceName, tableName))
-          .get();
+      // After creating the table, it can take up to 2 seconds for permissions cache to be updated,
+      // So retry this first index creation until it succeeds.  The other index creations should not
+      // need this behavior.
+      boolean success = false;
+      while (!success) {
+        try {
+          Uninterruptibles.sleepUninterruptibly(2, TimeUnit.SECONDS);
```
tomekl007, line 138: You did mention during the standup that this is configurable. Can we extract the configuration value and use it here?
Currently, if someone will change this setting, it will not be reflected in this wait code.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
EricBorczuk: Upon looking into this, I could use an implementation of QueryInterceptor to intercept a Create Table statement and then sleep for a bit...that still does a sleep though which is really not preferable.

I think maybe the key to this would be having closer control over the perms cache like Jeremiah mentioned is available in DSE...however, that probably will not work in C* 3.11/4, which is not so great...

For now, I decided to force a re-login to force refreshing of permissions.  I believe this will work across all three versions we care about.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
EricBorczuk: It is true that it's configurable in cassandra.yaml; from questions I had asked last week I had thought that stargate doesn't have access to those configs at runtime though?

Either way, I chose to take a different approach that doesn't sleep :) so hopefully I've skirted that issue

dougwettlaufer (approved):

PR #290, opened 2020-10-16
Fix initializing auth in auth-table-based-service (fixes #288)

### LINKED ISSUE #288: Unable to initialize auth when using DSE backend
When starting up in developer mode with DSE auth-table-based-service is unable to create the keyspace and table since there isn't a user in the QueryState. This works in both C* 3.11 and C* 4.0 because there's no user check when executing the query.

dougwettlaufer:


dougwettlaufer:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- testing/src/test/java/io/stargate/it/http/HealthCheckerTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6354980&bid=20633065)

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -348,7 +348,12 @@ protected void loginInternally(io.stargate.db.AuthenticatedUser user) {
 
     private Single<QueryState> newQueryState() {
       if (clientState.getUser() == null) {
-        return Single.just(new QueryState(clientState, UserRolesAndPermissions.ANONYMOUS));
+        // This is here to maintain consistent behavior with C* 3.11 and C* 4.0 since DSE requires
```
tomekl007, line 351: What if someone from the outside will know this `anonymous` user name?

dougwettlaufer: The only time there's no user on the clientState is during internal calls, elsewhere all requests go through an authentication (e.g. cql, rest, and graphql).

EricBorczuk: So there is no way for a requester to log in as `ANONYMOUS_USER`? (does the role have `with login = false` or smth)

EricBorczuk (approved):

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
dougwettlaufer: Well if they somehow logged in with the anonymous user then the clientState would have a user and you wouldn't drop into this block.

Issue #278, opened 2020-10-14
Upgrade swagger to 1.5.8+ to support `@PATCH` annotation from `javax.ws.rs`


EricBorczuk:
On 1.5.0, only the `@PATCH` annotation from io.swagger.jaxrs.PATCH is supported, because for a while PATCH was considered non-standard. Was fixed by https://github.com/swagger-api/swagger-core/pull/1679, which I believe is tagged as version 1.5.8

PR #294, opened 2020-10-20
Upgrade swagger-jersey2-jaxrs

FRosner:
fixes https://github.com/stargate/stargate/issues/278

Is there more to it? :D

Would appreciate if this could be labeled `hacktoberfest-accepted` :)


dougwettlaufer:
> Is there more to it? :D

Also needs upgraded in restapi. Anything else @EricBorczuk ?

EricBorczuk:
Hey @FRosner !

This looks right, I'm pretty sure - however, could you make sure by doing the following:
1) change all usages of `import io.swagger.jaxrs.PATCH;` in our code to `import javax.ws.rs.PATCH;` (there should be 3 in all)
2) Fire up stargate locally and navigate in your browser to `localhost:8082/swagger-ui`
3) Ensure that there are exactly 3 endpoints that say `PATCH` (and not zero)

If that works, then keep the changes you made in 1), push that up, and I think the work here is done! And thanks for tackling this!

FRosner:
> Hey @FRosner !
> 
> This looks right, I'm pretty sure - however, could you make sure by doing the following:
> 
> 1. change all imports of `import io.swagger.jaxrs.PATCH;` in the code to `import javax.ws.rs.PATCH;` (there should be 2 files with that import - DocumentResourceV2 and RowsResource)
> 2. Fire up stargate locally and navigate in your browser to `localhost:8082/swagger-ui`
> 3. Ensure that there are exactly 3 endpoints that say `PATCH` (and not zero, which was the bug before)
> 
> If that works, then keep the changes you made in 1), push that up, and I think the work here is done! And thanks for tackling this!

Sounds like a plan! I'm not sure when I have time. If this is critical, lmk. Otherwise I will try to find some time in the next days 👍 

Thanks for your help!

FRosner:
image
image
image
image

dougwettlaufer (approved):

PR #297, opened 2020-10-20
Add pipefail to test script

dougwettlaufer:


olim7t (approved):

olim7t:
Actually it would be better not to fix the REST resource at first, to check that the build failure is reported correctly.

dougwettlaufer:
> Actually it would be better not to fix the REST resource at first, to check that the build failure is reported correctly.

Yeah realized that right after I pushed my commit... I'll see what this does and then comment out and try again

Issue #303, opened 2020-10-21
Config Store - Mechanism for supporting reload of settings at the runtime


tomekl007:
We need to allow clients of the config-store to see the latest change.
We can implement it using one of these two approaches:
1) "pull" mode: the client will refresh the setting value at regular intervals
2) "push" mode: listener api where clients subscribe to the setting change modification.

If we want 1, we should be ok with a cache. Client’s are interacting with the config-store using the ConfigStore interface so that we can hide the cache implementation there. The refresh interval can be exposed using the config-store API, registering default to for example 1 second.

If we want 2, we need a thread that polls the config for changes, an event API, etc. It's more complicated and maybe not needed for a V_1

tomekl007:
After some more thoughts, I think we may not need to implement any strategy for reloading from the YAML config store.
The `ConfigStoreYaml#getConfigForModule(String)` loads the underlying `stargate-config.yaml` file every time its called. Because of that, the caller will get the fresh version of it every time it's accessed. It's up to the caller to decide when this method should be called. If the `ConfigStoreYaml` client needs to have runtime reload, it should call `getConfigForModule` every time the settings are accessed.

On the other hand, reading the file every time may be problematic from the performance perspective:
```
Map<String, Map<String, Object>> result =
          mapper.readValue(configFilePath.toFile(), yamlConfigType);
```
and we may consider adding a cache atop of it. I've created a simple JMH micro-benchmark comparing loading the file every time this method is called versus caching the file for 30 seconds (exact duration yet to be decided):
```

@Fork(1)
@Warmup(iterations = 5)
@Measurement(iterations = 10)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.MILLISECONDS)
public class ConfigStoreYamlBenchmark {
  private static final int NUMBER_OF_ITERATIONS = 100_000;


  @Benchmark
  public void directLoad(Blackhole blackhole) {
	  // given
	  Path path =
			  Paths.get(
					  Objects.requireNonNull(getClass().getClassLoader().getResource("stargate-config.yaml"))
							  .getPath());
	  ConfigStoreYaml configStoreYaml = new ConfigStoreYaml(path);

	  for(int i = 0; i < NUMBER_OF_ITERATIONS; i++) {


		// when
		ConfigWithOverrides configForModule1 = configStoreYaml.getConfigForModule("extension-1");
[...]

Issue #302, opened 2020-10-21
Config-Store API supports overriding settings


tomekl007:
We need to decide on the hierarchy and priority of configuration. 
We can use a similar pattern to https://docs.spring.io/spring-boot/docs/1.0.1.RELEASE/reference/html/boot-features-external-config.html.
The setting from the config store may have the lowest priority. It can be overridden at lower levels (i.e., YAML file, Java System properties, Command line arguments, etc.). We need to decide how many levels of overriding do we want to support.
There is one caveat here because once we support overrides, the GUI/REST will not reflect the actual value loaded by the extension (if it provides its own override). We would need to educate our users about it.

It has one drawback: if the setting is provided at the lowest level (i.e., system override), we cannot change it at the runtime.
We could solve it by having a "get if not present in config-store" semantic. If the setting is not set in the config-store, then take the lowest level setting. If it is set in the config-store, take the config-store setting ignoring the system override. 

Both approaches have different semantics, and we need to decide which one do we want to support.

tomekl007:
1. I think that we can achieve both goals by inverting priorities.
All settings from the config-store will have the highest priority. In the current design, it is the `stargate-config.yaml` mounted by the k8s. 
The lower priority will have settings that are set using `System.setProperty`. If they are set, and the setting is not present in the `stargate-config.yaml` they will be taken into account - that will be a current situation.
If someone wants to override those at runtime, it will be possible via the `stargate-config.yaml` file.

At some point, we can migrate settings to the `stargate-config.yaml` - that will have a higher priority.
So to sum up:
1. P0 - settings from config store.
2. P1 setting from `System.setProperty`
3. P3 environment variable using `System.getenv`

It seems that the spring cloud does a similar thing for priorities of settings:
https://cloud.spring.io/spring-cloud-commons/multi/multi__spring_cloud_context_application_context_services.html

If we decide that this is a correct solution, it will mean that all `System.getProperty` should be migrated to use the config-store instead. For example, in `CassandraPersistenceActivator`:
```
String clusterName = System.getProperty("stargate.cluster_name", "stargate-cassandra");
```
will be replaced with:
```
String clusterName = Optional.ofNullable(configStore.getConfigForModule("stargate").get("cluster_name")).orElse("stargate-cassandra");
```
to have the same read path (through config-store API) for every setting. (It can be done in a separate ticket)

[...]

PR #305, opened 2020-10-21
Implement Config-Store API based on the K8s config-map API

tomekl007:


config-store-k8s-config-map/src/test/java/io/stargate/config/store/k8s/config/map/ConfigStoreK8sConfigMapTest.java
```diff
@@ -0,0 +1,136 @@
...
+            .build();
+
+    client.configMaps().inNamespace(namespace).create(configMap1);
+    client.configMaps().inNamespace(namespace).create(configMap2);
+
+    // when
+    ConfigStore configStore =
+        new ConfigStoreBuilder()
+            .withNamespace(namespace)
+            .withConfig(client.getConfiguration())
```
tomekl007, line 77: We would need to construct this `Config` object based on the k8s configuration. Does anyone know if we will have access to k8s configuration at runtime to construct config (we will need for sure at least the master URL)?

dougwettlaufer:


```diff

Issues
======
- Added 3
           

Complexity increasing per file
==============================
- config-store-k8s-config-map/src/test/java/io/stargate/config/store/k8s/config/map/ConfigStoreK8sConfigMapTest.java  1
- config-store-k8s-config-map/src/main/java/io/stargate/config/store/k8s/config/map/ConfigStoreActivator.java  1
- config-store-k8s-config-map/src/main/java/io/stargate/config/store/k8s/config/map/ConfigStoreBuilder.java  1
- config-store-k8s-config-map/src/main/java/io/stargate/config/store/k8s/config/map/ConfigStoreK8sConfigMap.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

config-store-k8s-config-map/src/main/java/io/stargate/config/store/k8s/config/map/ConfigStoreK8sConfigMap.java
```diff
@@ -0,0 +1,74 @@
...
+              .build();
+      client.configMaps().inNamespace(namespace).create(newConfigMap);
+    } else {
+      // create a config map based on the existing one
+      ConfigMap newConfigMap =
+          new ConfigMapBuilder()
+              .withNewMetadata()
+              .withName(extensionName)
+              .endMetadata()
+              .addToData(configMap.getData())
```
tomekl007, line 63: We may have a race condition here. One node may see `configMap == null` whereas others may see already created `configMap.` But I think that this is not a problem - the add to data operation is idempotent because it does not keep duplicates. Duplicate overrides the previous value, so the last one wins.
The 2nd case is more problematic - if one node already creates a configMap, the other will create a new empty map overriding previous values. I am not sure if k8s has a pattern for doing that with the `config-map` API.

config-store-k8s-config-map/src/main/java/io/stargate/config/store/k8s/config/map/ConfigStoreK8sConfigMap.java
```diff
@@ -0,0 +1,74 @@
...
+  @Override
+  public Map<String, String> getConfigForExtension(String extensionName) {
+    Resource<ConfigMap, DoneableConfigMap> configMaps =
+        client.configMaps().inNamespace(namespace).withName(extensionName);
+    return configMaps.get().getData();
+  }
+
+  @Override
+  public void addOrReplaceConfigValueForExtension(String extensionName, String key, String value) {
+    ConfigMap configMap = client.configMaps().inNamespace(namespace).withName(extensionName).get();
```
tomekl007, line 45: I think that we should consider creating ConfigMap per extension name (as it is now). 
So every extension has a dedicated config map, wdyt?

config-store-k8s-config-map/src/main/java/io/stargate/config/store/k8s/config/map/ConfigStoreK8sConfigMap.java
```diff
@@ -0,0 +1,74 @@
...
+
+  public ConfigStoreK8sConfigMap(String namespace, NamespacedKubernetesClient client) {
+    this.namespace = namespace;
+    this.client = client;
+  }
+
+  @Override
+  public Map<String, String> getConfigForExtension(String extensionName) {
+    Resource<ConfigMap, DoneableConfigMap> configMaps =
+        client.configMaps().inNamespace(namespace).withName(extensionName);
```
tomekl007, line 39: the `configMaps()` API returns `Map<String, String>`.

config-store-k8s-config-map/pom.xml
```diff
@@ -0,0 +1,171 @@
...
+        <groupId>io.stargate.config-store</groupId>
+        <artifactId>config-store-api</artifactId>
+        <version>${project.version}</version>
+        <scope>provided</scope>
+      </dependency>
+      <dependency>
+        <groupId>io.fabric8</groupId>
+        <artifactId>kubernetes-client</artifactId>
+        <version>4.12.0</version>
+      </dependency>
```
dougwettlaufer, line 24: This dependency shouldn't be needed since you're just working with a yaml file.

pom.xml
```diff
@@ -154,6 +154,8 @@
         <module>authentication</module>
         <module>auth-table-based-service</module>
         <module>health-checker</module>
+        <module>config-store-api</module>
+        <module>config-store-k8s-config-map</module>
```
dougwettlaufer, line 158: This can just be `config-store-file-based` or something like that. In k8s a configmap is just a way to attach a file to a pod. Once the file is attached then your application uses the file as if it were any other local file.

config-store-k8s-config-map/src/main/java/io/stargate/config/store/k8s/config/map/ConfigStoreK8sConfigMap.java
dougwettlaufer: There should just be a single configmap that contains a file. Inside the yaml file you can divide up config however you want.

config-store-k8s-config-map/src/main/java/io/stargate/config/store/k8s/config/map/ConfigStoreK8sConfigMap.java
dougwettlaufer: One issue with config maps is that they are eventually consistent. So if I were to create a config map and bind it to 3 pods then they're not all going to see the change immediately but as they poll each of them will see the file based on their polling.

config-store-k8s-config-map/pom.xml
tomekl007: Yes, I am discussing this with Jim Dickinson. The approach that I took here was to use the client to get the config-map, but I will rework it to the YAML approach. To sum up, my finding at this moment:
1. The `extension-a.yaml` needs to be created somehow and provided by the k8s admin at deployment time
2. Then we would need to do smth like kubectl `create configmap from-yaml --from-file=./extension-a.yaml`
3. next, the actual stargate pod will need to mount this in the fs and use it in volumes:
```
      volumeMounts:
      - name: config-volume
        mountPath: /etc/config
  volumes:
    - name: config-volume
      configMap:
        # Provide the name of the ConfigMap containing the files you want
        # to add to the container
        name: from-yaml
```
4. the `extension-a.yaml` is available within to the stragate process.

tomekl007: If the end-user wants to reload the file, he can do:
```
kubectl apply -f my-configmap.yaml
emacs my-configmap.yaml
kubectl apply -f my-configmap.yaml
```
it will refresh the file, and the change will be available within the file system (this will be implemented as a part of https://github.com/stargate/stargate/issues/303)

config-store-k8s-config-map/src/main/java/io/stargate/config/store/k8s/config/map/ConfigStoreK8sConfigMap.java
tomekl007: yes, with the YAML approach one file may be easier from the deployment and maintenance perspective.

dougwettlaufer:


```diff

Issues
======
- Added 3
           

Complexity increasing per file
==============================
- config-store-yaml/src/main/java/io/stargate/config/store/k8s/config/map/ConfigStoreYaml.java  5
- config-store-yaml/src/test/java/io/stargate/config/store/k8s/config/map/ConfigStoreYamlTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/k8s/config/map/YamlConfigException.java  1
- config-store-api/src/main/java/io/stargate/config/store/MissingExtensionSettingsException.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/k8s/config/map/ConfigStoreActivator.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

dougwettlaufer:


```diff

Issues
======
- Added 3
           

Complexity increasing per file
==============================
- config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreYamlTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreActivator.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreYaml.java  5
- config-store-api/src/main/java/io/stargate/config/store/MissingExtensionSettingsException.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/yaml/YamlConfigException.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreYamlTest.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreActivatorYamlExistsTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreActivator.java  2
- config-store-yaml/src/main/java/io/stargate/config/store/yaml/YamlConfigException.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreYaml.java  5
- config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreActivatorYamlDoesNotExistsTest.java  1
- config-store-api/src/main/java/io/stargate/config/store/MissingExtensionSettingsException.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreActivatorYamlDoesNotExistsTest.java  1
- config-store-api/src/main/java/io/stargate/config/store/MissingExtensionSettingsException.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreYaml.java  5
- config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreActivator.java  2
- config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreActivatorYamlExistsTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/yaml/YamlConfigException.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreYamlTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreActivatorYamlDoesNotExistsTest.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreYamlTest.java  1
- config-store-api/src/main/java/io/stargate/config/store/MissingExtensionSettingsException.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreActivatorYamlExistsTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/yaml/YamlConfigException.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreActivator.java  2
- config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreYaml.java  5
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-yaml/src/main/java/io/stargate/config/store/yaml/YamlConfigException.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreYamlTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreActivator.java  2
- config-store-api/src/main/java/io/stargate/config/store/MissingExtensionSettingsException.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreActivatorYamlDoesNotExistsTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreYaml.java  5
- config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreActivatorYamlExistsTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

config-store-api/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+				<configuration>
+					<unpackBundle>true</unpackBundle>
+					<instructions>
+						<Bundle-Name>Config-Store-API</Bundle-Name>
+						<Bundle-Description>API for Config Store</Bundle-Description>
+						<Bundle-SymbolicName>io.stargate.config.store</Bundle-SymbolicName>
+						<Import-Package>org.osgi.framework</Import-Package>
+						<Export-Package><![CDATA[
+              io.stargate.config.store,
+            ]]></Export-Package>
```
dimas-b, line 28: nit: odd formatting.... why not use zero indent for `io.stargate.config.store`?

config-store-api/src/main/java/io/stargate/config/store/ConfigStore.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.Map;
+
+public interface ConfigStore {
+
+  /**
+   * It retrieves the generic {@code Map<String, Object>} for a given extension name.
+   *
+   * <p>If the caller will try to look up the config for an extension that does not have any
+   * setting, it will throw the {@link MissingExtensionSettingsException}. This is a {@link
```
dimas-b, line 26: Why throw instead of returning an empty `Map`? Is it really an error for an extension to have no config settings defined?

Since the whole config is dynamic, settings may appear later... Would it be more appropriate to just log an error message in this case?

config-store-api/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+		<plugins>
+			<plugin>
+				<groupId>org.apache.felix</groupId>
+				<artifactId>maven-bundle-plugin</artifactId>
+				<configuration>
+					<unpackBundle>true</unpackBundle>
+					<instructions>
+						<Bundle-Name>Config-Store-API</Bundle-Name>
+						<Bundle-Description>API for Config Store</Bundle-Description>
+						<Bundle-SymbolicName>io.stargate.config.store</Bundle-SymbolicName>
```
dimas-b, line 24: `io.stargate.config.store.api`?

config-store-api/pom.xml
dimas-b: Alternatively, since we're not likely to export any other package here... do we really need `CDATA`?

config-store-yaml/README.adoc
```diff
@@ -0,0 +1,74 @@
+= Using Config Store Yaml
+
+This extension provides support for loading stargate configuration from a YAML file.
+
+== YAML File Location
+
+When the `ConfigStoreYaml` is activated (registered in the OSGi), it will try to lookup the stargate config
```
dimas-b, line 7: To me it looks like `activated` and `registered` are conflicting in this context... `registered` is also mentioned below as an _optional_ event :thinking:

config-store-yaml/README.adoc
```diff
@@ -0,0 +1,74 @@
...
+in the `/etc/stargate/stargate-config.yaml` location. Please note that the name of the actual YAML file is `stargate-config.yaml`
+If it does not exists, the error log is produced, and `ConfigStoreYaml` is not registered.
```
dimas-b, line 9: What about local runs? Is the config location overridable?

config-store-api/src/main/java/io/stargate/config/store/ConfigStore.java
tomekl007: From the `ConfigStore` perspective, this is an abnormal situation that the client asks for the config, but the config does not exist. It is up to the caller to decide what to do with that fact. By being explicit in the API, we are allowing clients to provide any missing setting handling. If we would hide it and log that fact, callers will not be able to decide anymore.

config-store-api/pom.xml
tomekl007: I am following the convention from `persistence-api` module. We are not including `api` in the package name there.

config-store-api/pom.xml
tomekl007: good catch, I will remove `CDATA`

config-store-yaml/README.adoc
tomekl007: right, I've unified it

config-store-yaml/pom.xml
```diff
@@ -0,0 +1,181 @@
...
+				<configuration>
+					<unpackBundle>true</unpackBundle>
+					<instructions>
+						<Bundle-Name>Config-Store-Yaml</Bundle-Name>
+						<Bundle-Description>Config store YAML implementation</Bundle-Description>
+						<Bundle-SymbolicName>io.stargate.config.store.yaml</Bundle-SymbolicName>
+						<Bundle-Activator>io.stargate.config.store.yaml.ConfigStoreActivator</Bundle-Activator>
+						<Import-Package><![CDATA[
+              org.osgi.framework,
+              io.stargate.config.store,
```
dimas-b, line 156: What do you think about moving API into `io.stargate.config.store.api`? If API and impl. are sibling packages it should make it easier to deal with package selectors, etc... it does seem to be an issue now, but might become prominent at the code evolves.

config-store-yaml/pom.xml
```diff
@@ -0,0 +1,181 @@
...
+              !com.codahale.metrics.graphite.*,
```
dimas-b, line 157: Why do we need this? Aren't packages excluded by default?

config-store-yaml/pom.xml
```diff
@@ -0,0 +1,181 @@
...
+						<version>${junit.version}</version>
+					</dependency>
+				</dependencies>
+			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-dependency-plugin</artifactId>
+				<executions>
+					<execution>
+						<id>unpack-sjk</id>
```
dimas-b, line 101: Do we actually use `sjk` in this module?

config-store-yaml/pom.xml
tomekl007: I was anticipating that I will use metrics, but I don't. Thx for pointing out - I will remove.

config-store-yaml/pom.xml
```diff
@@ -0,0 +1,181 @@
...
+      <dependency>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-api</artifactId>
+      </dependency>
+      <dependency>
+        <groupId>org.slf4j</groupId>
+        <artifactId>log4j-over-slf4j</artifactId>
+        <version>1.7.30</version>
+        <scope>runtime</scope>
+      </dependency>
```
dimas-b, line 52: Do we actually use `log4j` in this module? Shouldn't logging go directly to whatever `slf4j` implementation is defined at the container level without the extra `log4j` bridge?

config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreActivator.java
```diff
@@ -0,0 +1,61 @@
...
+
+  @Override
+  public void start(BundleContext context) {
+    logger.info("Starting Config Store YAML...");
+    Path yamlFilePath = Paths.get(CONFIG_YAML_LOCATION);
+    if (!Files.exists(yamlFilePath)) {
+      logger.error(
+          "The yaml file does not exists, please check the path: {}. The ConfigStoreYaml will not be registered.",
+          yamlFilePath);
+      return;
```
dimas-b, line 45: What's our mechanism for reload config files? If we exit here without setting up any refresh tasks, we'll never get the config... might as well throw an exception in this case? WDYT?

config-store-yaml/pom.xml
tomekl007: you are right, we are not using persistence - I'll remove it.

config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreActivator.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  @Override
+  public void stop(BundleContext context) {
+    // no-op
+  }
+
+  @Override
+  public void serviceChanged(ServiceEvent event) {
+    // no-op because ConfigStoreYaml does not depend on any other service
```
dimas-b: Do we need to implement `ServiceListener` then?

config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreActivator.java
tomekl007: It will be implemented as a separate task:
https://github.com/stargate/stargate/issues/303

config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreYaml.java
```diff
@@ -0,0 +1,62 @@
...
+          mapper.readValue(configFilePath.toFile(), yamlConfigType);
+      if (!result.containsKey(extensionName)) {
+        throw new MissingExtensionSettingsException(
+            String.format(
+                "The loaded configuration map: %s, does not contain settings from a given extension: %s",
+                result, extensionName));
+      }
+      return result.get(extensionName);
+    } catch (IOException e) {
+      throw new YamlConfigException("Problem when processing yaml file from: " + configFilePath, e);
```
dimas-b: What's the purpose of using a private exception class in this case? Callers will not be able to examine it because it is not part of the API... :thinking:

dimas-b: Why not `MissingExtensionSettingsException`?.. If we cannot load the config it's a good as missing, isn't it?

tomekl007: This is because of our code-quality plugin :D It reports throwing raw `RuntimeException` as a code problem.

config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreActivatorYamlDoesNotExistsTest.java
```diff
@@ -0,0 +1,49 @@
...
+  @BeforeAll
+  public static void setup() {
+    System.setProperty("stargate.config_store.yaml.location", "non_existing_path");
+  }
+
+  @Test
+  public void shouldNotRegisterConfigStoreWhenYamlLocationHasNotExistingStargateConfig() {
+    // given
+    BundleContext bundleContext = mock(BundleContext.class);
+    ConfigStoreActivator kafkaProducerActivator = new ConfigStoreActivator();
```
dimas-b: nit: why `kafka`?

config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreActivatorYamlDoesNotExistsTest.java
```diff
@@ -0,0 +1,49 @@
...
+import io.stargate.config.store.ConfigStore;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.osgi.framework.BundleContext;
+
+class ConfigStoreActivatorYamlDoesNotExistsTest {
+
+  @BeforeAll
+  public static void setup() {
+    System.setProperty("stargate.config_store.yaml.location", "non_existing_path");
```
dimas-b: Should we have a test for starting with no config file? I think it's worth adding it since we apparently expect this situation to happen at some point.

config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreActivatorYamlDoesNotExistsTest.java
```diff
@@ -0,0 +1,49 @@
...
+    // given
+    BundleContext bundleContext = mock(BundleContext.class);
+    ConfigStoreActivator kafkaProducerActivator = new ConfigStoreActivator();
+
+    // when
+    kafkaProducerActivator.start(bundleContext);
+
+    // then
+    verify(bundleContext, times(0))
+        .registerService(eq(ConfigStore.class), any(ConfigStoreYaml.class), any());
```
dimas-b: Should we have a test for starting with no config file? I think it's worth adding it since we apparently expect this situation to happen at some point.

tomekl007: The test in this class validates exactly this case.  The file does not exist - there is no config file.

config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYamlDoesNotExistsTest.java
```diff
@@ -0,0 +1,49 @@
...
+import io.stargate.config.store.api.ConfigStore;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.osgi.framework.BundleContext;
+
+class ConfigStoreActivatorYamlDoesNotExistsTest {
+
+  @BeforeAll
+  public static void setup() {
+    System.setProperty("stargate.config_store.yaml.location", "non_existing_path");
```
dimas-b, line 33: This will affect of the other unit tests in the same JVM, I think... could lead to odd failures.

config-store-api/src/main/java/io/stargate/config/store/ConfigStore.java
dimas-b: +1 to explicit API, but from the practical perspective, can we be sure that config and jars are always in sync during deployment?

config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreYaml.java
dimas-b: Hmm, could we use `IllegalStateException` or `MissingExtensionSettingsException`? Adding a whole exception class just for one case seems like an overkill to me :)

config-store-yaml/src/test/java/io/stargate/config/store/yaml/ConfigStoreActivatorYamlDoesNotExistsTest.java
dimas-b: you're right, I got confused, sorry about the noise :facepalm:

config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYamlDoesNotExistsTest.java
dimas-b: Why not add another constructor to allow unit tests to pass specific settings? The default constructor would use system properties... WDYT?

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java  2
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreYaml.java  5
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYamlExistsTest.java  1
- config-store-api/src/main/java/io/stargate/config/store/api/MissingExtensionSettingsException.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYamlDoesNotExistsTest.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/YamlConfigException.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

config-store-api/src/main/java/io/stargate/config/store/ConfigStore.java
tomekl007: The `stargate-config.yaml` will be mounted in a pod before the deployment of the stargate, see:
https://github.com/riptano/cloud-ondemand/pull/2339
So I think it guarantees the happens-before and files will be in sync.
When the file will be refreshed we would need to only load it again.

config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYamlDoesNotExistsTest.java
tomekl007: good idea, done.

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreYaml.java  5
- config-store-api/src/main/java/io/stargate/config/store/api/MissingExtensionSettingsException.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYaml.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

config-store-yaml/pom.xml
```diff
@@ -0,0 +1,148 @@
...
+        <groupId>org.mockito</groupId>
+        <artifactId>mockito-core</artifactId>
+        <version>3.5.10</version>
+        <scope>test</scope>
+      </dependency>
+    </dependencies>
+	<build>
+		<plugins>
+			<plugin>
+				<artifactId>maven-failsafe-plugin</artifactId>
```
dimas-b, line 57: nit: since we do not have integration tests in this module, `maven-failsafe-plugin` should not be necessary.

config-store-yaml/pom.xml
```diff
@@ -0,0 +1,148 @@
...
+				</executions>
+				<configuration>
+					<parallel>suites</parallel>
+					<threadCountSuites>1</threadCountSuites>
+					<reuseForks>false</reuseForks>
+					<forkCount>1</forkCount>
+				</configuration>
+			</plugin>
+			<plugin>
+				<artifactId>maven-surefire-plugin</artifactId>
```
dimas-b, line 81: Is it not inherited (just wondering)?

dimas-b:
LGTM :+1: still I think there's some clean-up in pom files that's worth doing (e.g. sjk removal) and possibly failsafe/surefire removal - I guess the latter should work without an explicit declaration.

config-store-api/src/main/java/io/stargate/config/store/api/ConfigStore.java
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+   * It retrieves the generic {@code Map<String, Object>} for a given extension name.
+   *
+   * <p>If the caller will try to look up the config for an extension that does not have any
+   * setting, it will throw the {@link MissingExtensionSettingsException}. This is a {@link
+   * RuntimeException} and it's the caller's responsibility to catch it and handle gracefully or
+   * propagate upper in the call stack.
+   */
+  Map<String, Object> getConfigForExtension(String extensionName)
```
dougwettlaufer, line 30: We should probably be working with "module" or "component" as those are more encompassing names. In the event that we store config for something like `core` it wouldn't make sense to call `getConfigForExtension()`

config-store-yaml/README.adoc
```diff
@@ -0,0 +1,74 @@
+= Using Config Store Yaml
```
dougwettlaufer, line 1: We need to be consistent about documentation. Please make sure everyone is on the same page and if so create the necessary issues get our markdown converted to asciidoc. 

Personally, I don't think we're going to be writing anything complicated enough here to warrant the need for asciidoc, it probably makes more sense over in stargate/docs. Plus markdown will probably be more contributor friendly (and if docs are easier to write then people are more likely to do it).

config-store-yaml/README.adoc
```diff
@@ -0,0 +1,74 @@
...
+  b: "value"
+
+extension-2:
+  a: 2
+  b: "value_2"
+----
+
+Every extension should have its own dedicated section.
+When client is using the `ConfigStore#getConfigForExtension(String extensionName)` method to retrieve the config for a given extension,
+the `ConfigStoreYaml` loads only a dedicated section.
```
dougwettlaufer, line 29: I need to check the impl but why not load the whole file at startup and just create a map of maps where the key is the name of the component?

config-store-yaml/README.adoc
```diff
@@ -0,0 +1,74 @@
...
+So for example, when using `ConfigStore#getConfigForExtension("extension-1)` it will return Java `Map<String, Object>` with two elements: `a:1` and `b:"value"`.
+It will allow isolation between settings from various extensions.
+
+If the caller will try to look up the config for an extension that does not have a dedicated section in the YAML file,
+it will throw the `MissingExtensionSettingsException`. This is a `RuntimeException` and it's the caller's responsibility to catch it and handle gracefully or propagate upper in the call stack.
```
dougwettlaufer, line 34:
```suggestion
it will throw the `MissingExtensionSettingsException`. This is a `RuntimeException` and it's the caller's responsibility to catch it and handle gracefully or propagate higher in the call stack.
```

config-store-yaml/pom.xml
```diff
@@ -0,0 +1,148 @@
...
+      <dependency>
+        <groupId>org.mockito</groupId>
+        <artifactId>mockito-core</artifactId>
+        <version>3.5.10</version>
+        <scope>test</scope>
+      </dependency>
+    </dependencies>
+	<build>
+		<plugins>
+			<plugin>
```
dougwettlaufer, line 56: Interesting, I'm surprised the formatter doesn't complain about all of this whitespace.

config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @Override
+  public void start(BundleContext context) {
+    logger.info("Starting Config Store YAML...");
+    Path yamlFilePath = Paths.get(configYamlLocation);
+    if (!Files.exists(yamlFilePath)) {
+      logger.error(
+          "The yaml file does not exists, please check the path: {}. The ConfigStoreYaml will not be registered.",
+          yamlFilePath);
+      return;
```
dougwettlaufer, line 51: Shouldn't we throw an exception and fail startup? Otherwise this is just going to cause stargate init to hang, right?

config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+    ConfigStoreYaml configStoreYaml = new ConfigStoreYaml(yamlFilePath);
+    context.registerService(ConfigStore.class, configStoreYaml, new Hashtable<>());
```
dougwettlaufer, line 54: This needs a name so in the future we can select between different config stores with a system property.

config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+    // when
+    Map<String, Object> configForExtension1 = configStoreYaml.getConfigForExtension("extension-1");
+    Map<String, Object> configForExtension2 = configStoreYaml.getConfigForExtension("extension-2");
+
+    // then
+    assertThat(configForExtension1)
+        .containsOnly(new SimpleEntry<>("a", 1), new SimpleEntry<>("b", "value"));
+    assertThat(configForExtension2)
+        .containsOnly(new SimpleEntry<>("a", 2), new SimpleEntry<>("b", "value_2"));
```
dougwettlaufer, line 49: If the value for a key is a list this handles it too, right?

config-store-yaml/README.adoc
tomekl007: We are doing just that. This API is returning an nested map for a given component to provide isolation between extensions from the client perspective.

config-store-yaml/pom.xml
tomekl007: There is something wrong with the `xml` formatting plugin. I am unable to make it run and format files.
Are you able to run XML formatting phase?

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreYaml.java  5
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYaml.java  1
- config-store-api/src/main/java/io/stargate/config/store/api/MissingExtensionSettingsException.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java
tomekl007: yes, I've added a unit test for it.

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-api/src/main/java/io/stargate/config/store/api/MissingExtensionSettingsException.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreYaml.java  5
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java  2
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYaml.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

config-store-api/src/main/java/io/stargate/config/store/api/ConfigStore.java
tomekl007: I've changed it to a `module`.

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-api/src/main/java/io/stargate/config/store/api/MissingModuleSettingsException.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreYaml.java  5
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYaml.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java  2
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreYaml.java  5
- config-store-api/src/main/java/io/stargate/config/store/api/MissingModuleSettingsException.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYaml.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java  2
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreYaml.java  5
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java  2
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java  1
- config-store-api/src/main/java/io/stargate/config/store/api/MissingModuleSettingsException.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYaml.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java  2
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYaml.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreYaml.java  5
- config-store-api/src/main/java/io/stargate/config/store/api/MissingModuleSettingsException.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java
tomekl007: Why would it block? I've tested and it does not - it just not registering the extension. 
I think that we can change it to throw exception depending on how the implementation here:
https://github.com/riptano/cloud-ondemand/pull/2339/files#diff-15daad6661ac657b68d9b5ddf39c9e77bd60f7840b447a264e22fc98f9c9c648R9
will look like

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreYaml.java  5
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java  1
- config-store-api/src/main/java/io/stargate/config/store/api/MissingModuleSettingsException.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java  2
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYaml.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

config-store-yaml/pom.xml
dimas-b: Indeed, the xml formatter does not detect errors in `pom.xml` files :thinking: 

I does not reformat them even if I run `./mvnw com.coveo:fmt-maven-plugin:format`

config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java
dimas-b: I guess Doug meant that if some other module depends on the config service, it will never really register in this case too effectively making those other services/APIs unavailable.

I think it would be preferable to throw an exception is this case and fail the whole Stargate startup process for a clear and unambiguous failure mode.

config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYaml.java
```diff
@@ -0,0 +1,70 @@
...
+
+import io.stargate.config.store.api.ConfigStore;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Hashtable;
+import java.util.Objects;
+import org.junit.jupiter.api.Test;
+import org.osgi.framework.BundleContext;
+
+class ConfigStoreActivatorYaml {
```
dimas-b, line 32: Just wondering do we need to have the `Test` class name suffix for the `surefire` plugin to be able to find it?

config-store-api/pom.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+		 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<parent>
+		<artifactId>stargate</artifactId>
+		<groupId>io.stargate</groupId>
+		<version>0.0.18-SNAPSHOT</version>
```
dimas-b, line 8: I guess this version may need to be adjusted... it does not find the parent module in my local run:

```
$ ./mvnw -f config-store-yaml/ test
[...]
[ERROR] Failed to execute goal on project config-store-yaml: Could not resolve dependencies for project io.stargate.config.store.yaml:config-store-yaml:jar:0.0.18-SNAPSHOT: Could not find artifact io.stargate.config-store:config-store-api:jar:0.0.18-SNAPSHOT -> [Help 1]

```

dimas-b: I guess this is because the two new modules are not added to the parent pom, therefore they do not get published from the top-level `mvn install` command :thinking:

config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYaml.java
dimas-b: I did a local run of `./mvnw -f config-store-yaml/ test` and `ConfigStoreYamlTest` executed, but this class did not execute.

pom.xml
```diff
@@ -154,6 +154,8 @@
         <module>authentication</module>
         <module>auth-table-based-service</module>
         <module>health-checker</module>
+        <module>config-store-api</module>
+        <module>config-store-yaml</module>
```
dimas-b, line 158: I think we need to add these modules to the default section too (line 133)

config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java
tomekl007: Yes, I also think that being explicit is a good choice but there are two problems with throwing here:
1. If we will merge it and throw the exception, then people will not be able to start the stargate any more without providing this new `stargate-config.yaml` file. I think that it may be too intrusive.
2. We need to coordinate with the cloud-on-demand changes (https://github.com/riptano/cloud-ondemand/pull/2339) - only after our deployment process will be adapted, we can throw an exception.

So my proposition is to log an error for now to allow easier transition and change this logic to throw an exception once the deployment process will be adapted. wdyt?

dimas-b: Why not add an empty default config yaml file? This way we will defer errors to specific modules that need config settings while allowing ordinary modules to activate normally?

tomekl007: We will have one config file provided by the kubernetes; see this:
https://github.com/riptano/cloud-ondemand/pull/2339
If we provide the default file here, there will be a conflict with the file provided by the k8s.
We may decide to add a default file or file per module once we will decide about overrides; for more info, please see this discussion:
https://github.com/stargate/stargate/issues/302#issuecomment-716535829
but also this should be done separately.

dimas-b: my concern is that k8s does not apply to local builds and to our public docker images... those will effectively not have a config service unless someone adds a config file manually.... I'm not 100% sure what the best solution could be in this case, but in my previous experience with OSGi debugging missing service instances is very tricky and tedious... dependent services just wait without any visible error messages.

config-store-yaml/pom.xml
dougwettlaufer: Hmm, that's weird.

config-store-api/src/main/java/io/stargate/config/store/api/MissingModuleSettingsException.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.config.store.api;
+
+/** Denotes that the underlying config-store does not have any setting for a given extension. */
```
dougwettlaufer, line 18:
```suggestion
/** Denotes that the underlying config-store does not have any setting for a given module. */
```

config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java
dougwettlaufer: We can add a temporary feature flag for this with a system prop that defaults to disabled. But like Dmitri I've had enough fun debugging OSGI, we should fail fast if at all possible

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java  2
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYamlTest.java  1
- config-store-api/src/main/java/io/stargate/config/store/api/MissingModuleSettingsException.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreYaml.java  5
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-api/src/main/java/io/stargate/config/store/api/MissingModuleSettingsException.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYamlTest.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreYaml.java  5
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java  4
- config-store-api/src/test/java/io/stargate/config/store/api/ConfigWithOverridesTest.java  4
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYamlTest.java  1
- config-store-api/src/main/java/io/stargate/config/store/api/ConfigWithOverrides.java  3
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreYaml.java  5
- config-store-api/src/main/java/io/stargate/config/store/api/MissingModuleSettingsException.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-api/src/test/java/io/stargate/config/store/api/ConfigWithOverridesTest.java  4
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java  4
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYamlTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreYaml.java  5
- config-store-api/src/main/java/io/stargate/config/store/api/ConfigWithOverrides.java  3
- config-store-api/src/main/java/io/stargate/config/store/api/MissingModuleSettingsException.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java  4
- config-store-api/src/main/java/io/stargate/config/store/api/ConfigWithOverrides.java  3
- config-store-api/src/main/java/io/stargate/config/store/api/MissingModuleSettingsException.java  1
- config-store-api/src/test/java/io/stargate/config/store/api/ConfigWithOverridesTest.java  4
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYamlTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreYaml.java  5
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

config-store-yaml/pom.xml
dimas-b: I vaguely recall someone disabling the automatic XML format check... but I'm not certain.

config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java
```diff
@@ -0,0 +1,78 @@
...
+      return;
+    }
+    logger.info("Starting Config Store YAML...");
+    Path yamlFilePath = Paths.get(configYamlLocation);
+    if (!Files.exists(yamlFilePath)) {
+      throw new IllegalArgumentException(
+          String.format(
+              "The yaml file does not exists, please check the path: %s. The ConfigStoreYaml will not be registered.",
+              yamlFilePath));
+    }
```
dimas-b, line 64: Sorry to be a pain on this issue, but I think in this situation if/when we have another service depending on the config service, that other service will keep waiting on startup if the config service is not enabled. I believe this situation would be hard to debug due to lack of explicit error messages.

What I'm proposing is to _always_ start and _register_ the config service implementation, but throw exceptions about missing config data when that data is requested in run time. This way configuration errors/mistakes will be reported in the context of the service that actually uses that configuration and the error will be explicit. WDYT?

tomekl007: The main reason why I decided to validate that is that the file under `/etc/stargate/stargate-config.yaml"` must always be available when deploying stargate using k8s. The lack of this file is an abnormal situation. 
So if the config-store is enabled (i.e., on production), the file must be present. If it is not, it means that the deployment process is broken - it didn't create a config map.

On the other hand, throwing will make the development harder (starting stargate locally). So postponing this to the runtime will solve this problem, but it also means that we will detect the fact that deployment is broken very late on production. I am not sure if we want such behavior.
Another solution would be to change this:
```
  System.getProperty(
            "stargate.config_store.yaml.location", "/etc/stargate/stargate-config.yaml"));
```
to a location in the stargate repository (it can contain the empty file for now) and change this to `etc/stargate/stargate-config.yaml` when deploying using k8s.

dimas-b: Do we expect any other services to use the config in local builds or public docker images?

dimas-b: Just to clarify: I'm proposing to throw "missing config" exceptions from `ConfigStoreYaml.getConfigForModule`

dougwettlaufer: > What I'm proposing is to always start and register the config service implementation, but throw exceptions about missing config data when that data is requested in run time

I agree with @dimas-b here. If we think about the simplest case of developer-mode here it's perfectly valid to start up stargate without a config file and just a few command line args. In that case everything will register and run just fine. But, lets say you're running in prod with the kafka producer and the file is missing. In that case the _kafka producer_ will cause startup to fail because it's missing config it needs. Yes, it's not 100% explicit as you'll then have to dig through and figure out why the config is missing but it'll be more obvious than not starting configstore or blindly creating a new file.

tomekl007: Ok, if we have a quorum here, I am also ok with dropping that extra validation. 
Implemented in: https://github.com/stargate/stargate/pull/305/commits/43ed0fd33eaeca453beb6d4f7a2bf555dc52bad5

@dimas-b The throw missing config works as you described, so there is no need for change there

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java  1
- config-store-api/src/main/java/io/stargate/config/store/api/MissingModuleSettingsException.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java  1
- config-store-api/src/main/java/io/stargate/config/store/api/ConfigWithOverrides.java  3
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYamlTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreYaml.java  5
- config-store-api/src/test/java/io/stargate/config/store/api/ConfigWithOverridesTest.java  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

dimas-b (approved):

dougwettlaufer:


```diff

Complexity increasing per file
==============================
- config-store-api/src/main/java/io/stargate/config/store/api/ConfigWithOverrides.java  3
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreActivator.java  1
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreActivatorYamlTest.java  1
- config-store-yaml/src/main/java/io/stargate/config/store/api/yaml/ConfigStoreYaml.java  5
- config-store-api/src/test/java/io/stargate/config/store/api/ConfigWithOverridesTest.java  4
- config-store-yaml/src/test/java/io/stargate/config/store/api/yaml/ConfigStoreYamlTest.java  1
- config-store-api/src/main/java/io/stargate/config/store/api/MissingModuleSettingsException.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6377901&bid=20690326)

dougwettlaufer (approved):

config-store-api/pom.xml
```diff
@@ -0,0 +1,69 @@
...
```
jorgebay, line 8: Note that this needs to be updated to latest `0.0.20-SNAPSHOT`.

config-store-api/pom.xml
```diff
@@ -0,0 +1,69 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+		 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<parent>
```
jorgebay, line 5: Use of tabs instead of spaces, like the rest of pom files.

jorgebay (commented):
Late for the party but added a few comments.

config-store-api/pom.xml
tomekl007: I think that we should re-enable XML plugin and tackle all XML formatting globally.

PR #315, opened 2020-10-23
Increase timeouts due to slow CI

dimas-b:


dougwettlaufer (approved):

jorgebay (approved):
Thanks for looking into this 👍

jorgebay:
Not directly related to this, but shouldn't we set the log level for C*/DSE on the `testing` module to `WARNING` to reduce noise?

persistence-api/src/main/java/io/stargate/db/Persistence.java
```diff
@@ -34,6 +34,9 @@
  */
 public interface Persistence {
 
+  int SCHEMA_AGREEMENT_WAIT_RETRIES =
```
tomekl007, line 37: I think that it will be more user-friendly to have one global timeout setting and then split it into iterations, where every iteration lasts `200ms`.
Right now, to understand how this parameter translates to the actual timeout you need to `200 * SCHEMA_AGREEMENT_WAIT_RETRIES`, wdyt?

dimas-b: Sure, but I also think we ought to clean-up the "wait for agreement API" in more substantial ways. Ideally, I think the timeout should be delegated to the caller. I'll file an issue for this.

dimas-b: #318

dimas-b:
> Not directly related to this, but shouldn't we set the log level for C*/DSE on the testing module to WARNING to reduce noise?

AFAIK, C* / DSE logs do not go to the build STDOUT at all. Most of the messages there are from Stargate itself.

I personally find them very useful despite the noise, especially now that the tests are not very stable in CI. Perhaps we could have different settings in local builds and CI... WDYT?

PR #316, opened 2020-10-23
GraphQL: various fixes

olim7t:
There were corner cases where some types are frozen when we construct the name mapping, but not when we query (or vice-versa).

This also better encapsulates the logic in `NameMapping` behind specific methods, instead of exposing maps.

### DIFF
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/NameMapping.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/NameMapping.java
@@ -27,15 +27,10 @@ import java.util.Map;
 import java.util.Set;
 
 public class NameMapping {
-  private final BiMap<Table, String> entityNames = HashBiMap.create();
-  // TODO: using Column as map key is a bit dodgy. First, because Column objects are currently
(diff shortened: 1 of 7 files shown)


graphqlapi/src/main/java/io/stargate/graphql/schema/NameMapping.java
```diff
@@ -27,15 +27,10 @@
 import java.util.Set;
 
 public class NameMapping {
-  private final BiMap<Table, String> entityNames = HashBiMap.create();
-  // TODO: using Column as map key is a bit dodgy. First, because Column objects are currently
-  //   abused a bit by Result.Rows (in the persistence-api module), which can lead to subtle
-  //   issues easily. Second because in general, Column's equality is a tad complex and include
-  //   things like the type, which could possibly change over time for a given "column" (in the
-  //   sense of "defined in a table"). Using the column name here would be more reliable.
```
olim7t, line 35: Sylvain called that one 😉

jorgebay (approved):
lgtm!

dougwettlaufer (approved):

PR #319, opened 2020-10-23
Use tokens to authenticate SQL connections.

dimas-b:


sql/server/src/main/java/io/stargate/api/sql/server/avatica/StargateMeta.java
```diff
@@ -34,31 +36,33 @@
...
+          String username;
           String password = info.get("password");
-          if (username == null || password == null) {
+          if (password == null) {
             throw new IllegalArgumentException("Missing credentials in connection properties.");
           }
 
           try {
-            authenticator.createToken(username, password);
+            StoredCredentials credentials = authenticator.validateToken(password);
```
dougwettlaufer, line 58: Don't we want this to support both? Similar to how cql module works?

dimas-b: CQL validates passwords through SPNEGO, IIRC... but it does not look like it maps well to JDBC... there's no API to negotiate with the real client, AFAIK... We could simulate that on the server side, of course.

dimas-b: Ah, CQL uses SASL, but still it does not map well to JDBC :)

dimas-b: I'd prefer to defer actual password validation if tokens are ok short term. WDYT?

dougwettlaufer: Sorry that should have been phrased better. I meant don't we want to support both traditional username/password auth _and_ token based auth?

dimas-b: Another option is to add an new method to `AuthenticationService` to validate passwords without creating a token. This looks goable short term. Shall I open a PR to `master` for that?

dimas-b: As discussed offline, we're going to defer password-validation for SQL connections until we have a core API in Stargate for password checks. Previously (before this PR) passwords were validated by _creating a token_, which looks like an overkill.

sql/server/src/main/java/io/stargate/api/sql/server/avatica/StargateMeta.java
```diff
@@ -34,31 +36,33 @@
...
     this.authenticator = authenticator;
   }
 
   private void newConnection(ConnectionHandle ch, Map<String, String> info) {
     connections.computeIfAbsent(
         ch.id,
         sid -> {
-          String username = info.get("user");
+          String username;
           String password = info.get("password");
```
mpenick, line 52: Could the be called "token" as to differentiate it from user/pass authentication in the future?

mpenick: Ah, nm. I see how it's being passed.

dimas-b: The name "password" is kind of fixed by JDBC, IIRC. We could require "token" to be the user name for clarity, WDYT?

mpenick: This is what CQL does. It uses a configurable setting for the token user's name, but it defaults to "token".

dimas-b: added code to enforce the "token" user name in JDBC connections.

dougwettlaufer (approved):

PR #321, opened 2020-10-23
Use ClientStateWithPublicAddress to support proxy on internal actions in DSE

dougwettlaufer:


dimas-b (approved):

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -352,7 +353,10 @@ protected void loginInternally(io.stargate.db.AuthenticatedUser user) {
         // at least some user to be present but you cannot login with the "system" user.
         return Single.just(
             new QueryState(
-                ClientState.forExternalCalls(AuthenticatedUser.ANONYMOUS_USER),
+                new ClientStateWithPublicAddress(
+                    AuthenticatedUser.ANONYMOUS_USER,
+                    new InetSocketAddress("127.0.0.1", 0),
+                    new InetSocketAddress("127.0.0.1", 0)),
                 UserRolesAndPermissions.SYSTEM));
       } else {
```
dimas-b, line 361: Should we use `ClientStateWithPublicAddress` in the `else` branch too?

dougwettlaufer: Good catch, was assuming it worked but looks like it doesn't.

dimas-b: Given that `ProxyProtocolQueryInterceptor` expects all client states to be `ClientStateWithPublicAddress` I wonder if we should convert all places when `ClientState` is created :thinking:

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -294,10 +297,11 @@ private void waitForSchema(int delayMillis) {
 
   private static ClientState clientStateForExternalCalls(@Nonnull ClientInfo clientInfo) {
     if (clientInfo.publicAddress().isPresent()) {
-      new ClientStateWithPublicAddress(
+      return new ClientStateWithPublicAddress(
           clientInfo.remoteAddress(), clientInfo.publicAddress().get());
     }
-    return ClientState.forExternalCalls(clientInfo.remoteAddress(), null);
+    return new ClientStateWithPublicAddress(
```
mpenick, line 303: I'd prefer we fix downstream handling of `ClientState`. Something like? Or we could just return empty tables.

```diff
diff --git a/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptor.java b/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptor.java
index ec63191..71a0ebe 100644
--- a/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptor.java
+++ b/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptor.java
@@ -120,15 +120,14 @@ public class ProxyProtocolQueryInterceptor implements QueryInterceptor {
       QueryOptions options,
       Map<String, ByteBuffer> customPayload,
       long queryStartNanoTime) {
-    if (!isSystemLocalOrPeers(statement)) {
+    ClientState clientState = state.getClientState();
+    if (!isSystemLocalOrPeers(statement) || !(clientState instanceof ClientStateWithPublicAddress)) {
       return null;
     }
     SelectStatement selectStatement = (SelectStatement) statement;
     List<List<ByteBuffer>> rows;
-    ClientState clientState = state.getClientState();
-    assert clientState instanceof ClientStateWithPublicAddress;
     InetSocketAddress publicAddress = ((ClientStateWithPublicAddress) clientState).publicAddress();
     if (publicAddress == null) {
```

mpenick (approved):

PR #325, opened 2020-10-23
GraphQL: deserialize floating-point JSON values to BigDecimal

olim7t:
When a query is parameterized and the variables are sent separately (e.g. "Query variables" tab in Playground web UI), the variables are in JSON format.

graphql-java-servlet uses Jackson to deserialize them. By default, the object mapper used deserializes floating-point literals to `Double`, this can lead to loss of precision issues.

We should only do the conversion until we know the target CQL type. It's ok to lose precision for `float` or `double`, but not for `decimal`.

### DIFF
--- a/graphqlapi/src/main/java/graphql/kickstart/servlet/CustomGraphQLServlet.java
+++ b/graphqlapi/src/main/java/graphql/kickstart/servlet/CustomGraphQLServlet.java
@@ -15,6 +15,7 @@
  */
 package graphql.kickstart.servlet;
 
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import graphql.execution.AsyncExecutionStrategy;
 import graphql.kickstart.execution.GraphQLObjectMapper;
 import graphql.kickstart.execution.GraphQLQueryInvoker;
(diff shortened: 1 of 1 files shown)


jorgebay:
The JSON literals supported for each GraphQL type are:

- `Float32`: `Number`
- `Float` (built-in): `Number`
- `Int` (built-in): `Number`
- `Decimal`: `String`
- `BigInt`: `String` (or `Number` but that's an addition that can be discussed separately)

We should not support JSON Numbers for decimals, even if we can parse them correctly, it will create interop issues that will be very hard to troubleshoot. I think there's no need for special handling, JSON `Number` values are IEEE 754 `double` values, the json parser should handle them as `double`.

olim7t:
> We should not support JSON Numbers for decimals

We currently allow number literals for both GraphQL and JSON:
```
cqlsh:test> create table foo(d decimal primary key);
cqlsh:test> insert into foo(d) values (1.0);
```
GraphQL (the query works):
```
query {
  foo(value: { d: 1.0 }) { values { d } }
}
```
GraphQL, parameterized (the query executes but strangely returns no results):
```
query($d: Decimal) {
  foo(value: { d: $d }) { values { d } }
}
// query variables:
{"d": 1.0}
```
Accepting strings only is fine for me, but then we should make sure those queries throw a validation error.

jorgebay:
> we should make sure those queries throw a validation error

👍

olim7t:
OK, closing this then, I've created #337.

PR #335, opened 2020-10-27
Create a Junit 5 extension to manage driver sessions

olim7t:
See the javadocs of `CqlSessionExtension` for an overview.


testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
```diff
@@ -41,8 +50,30 @@
...
+            .map(
+                n -> {
+                  try {
+                    return InetAddress.getByName(n.seedAddress());
+                  } catch (UnknownHostException e) {
+                    throw new RuntimeException(e);
+                  }
+                })
+            .collect(Collectors.toList());
+  }
```
olim7t, line 66: Pulled up from `JavaDriverTestBase` (which is gone now). It's a bit awkward but its used in a few places and I don't want every test to reimplement it. Maybe we could just turn it into a method on `StargateEnvironmentInfo` and tests that need it can inject it.

testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
```diff
@@ -41,8 +50,30 @@
...
   public void init(ClusterConnectionInfo backend) {
     this.backend = backend;
   }
+
+  // TODO generalize this to an ExecutionCondition that reads custom annotations, like
+  // @CassandraRequirement/@DseRequirement in the Java driver tests
+  public boolean isCassandra4() {
+    return !backend.isDse()
+        && Version.parse(backend.clusterVersion()).nextStable().compareTo(Version.V4_0_0) >= 0;
+  }
```
olim7t, line 78: Same as above, this used to be on `JavaDriverTestBase`. There's nothing driver-specific here actually.
It should be possible to bake support for `@CassandraRequirement` / `@DseRequirement` directly into `StargateContainer`, we don't even need to redefine them really.

testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
olim7t: Nevermind, I've fixed that.

testing/src/test/java/io/stargate/it/driver/CqlSessionExtension.java
```diff
@@ -0,0 +1,405 @@
...
+    }
+  }
+
+  @Override
+  public void afterEach(ExtensionContext context) {
+    assert sessionLifecycle != null;
+    if (sessionLifecycle == Lifecycle.PER_METHOD) {
+      LOG.debug("Stopping per-method execution for {}", context.getElement());
+      destroySession(context);
+    }
```
dimas-b, line 171: If we store something implementing `ExtensionContext.Store.CloseableResource`, JUnit 5 will automatically close that thing when the context is closed. `ExternalResource` uses this approach. I do not insist, though :)

testing/src/test/java/io/stargate/it/driver/CqlSessionExtension.java
```diff
@@ -0,0 +1,405 @@
...
+        BeforeEachCallback,
+        AfterEachCallback,
+        AfterAllCallback,
+        ParameterResolver {
+
+  private static final Logger LOG = LoggerFactory.getLogger(CqlSessionExtension.class);
+  private static final int KEYSPACE_NAME_MAX_LENGTH = 48;
+
+  // Tracks whether the extension was configured per-class or per-method.
+  private volatile Lifecycle sessionLifecycle;
```
dimas-b, line 140: `volatile` may be unnecessary... my understanding is that JUnit will create different contexts (and consequently extension instances) for each thread during concurrent execution.

testing/src/test/java/io/stargate/it/driver/CqlSessionExtension.java
```diff
@@ -0,0 +1,405 @@
...
+    LOG.debug("Starting per-class execution for {}", context.getElement());
+    // beforeAll and afterAll are never called if we're per-method, that's how we figure out the
+    // lifecycle
+    sessionLifecycle = Lifecycle.PER_CLASS;
+    createSession(context);
+  }
+
+  @Override
+  public void beforeEach(ExtensionContext context) {
+    if (sessionLifecycle == null) {
```
dimas-b, line 158: If the test method has `@ExtendWith(CqlSessionExtension.class)` then we'll get a new instance of this class for each test method, AFAIK, so checking instance variables may not be reliable... WDYT?

dimas-b: It may be preferable to store session data in the `ExtrensionContext.Store`... :thinking:

testing/src/test/java/io/stargate/it/driver/CqlSessionExtension.java
```diff
@@ -0,0 +1,405 @@
...
+    return builder;
+  }
+
+  /**
+   * Waits until a CQL session sees all the Stargate nodes in its metadata. The CI environment is
+   * slow and it may take a while until system.peers is up to date.
+   *
+   * <p>Note: this is exposed publicly to be accessible from tests that create their own sessions.
+   */
+  public static void awaitAllNodes(CqlSession session, StargateEnvironmentInfo stargate) {
```
dimas-b, line 310: Would do you think about adding an injected interface for this API? This way we can avoid test code calling static methods (any methods in fact) on the extension class.

testing/src/test/java/io/stargate/it/driver/CqlSessionExtension.java
```diff
@@ -0,0 +1,405 @@
...
+    createSession(context);
+  }
+
+  @Override
+  public void beforeEach(ExtensionContext context) {
+    if (sessionLifecycle == null) {
+      LOG.debug("Starting per-method execution for {}", context.getElement());
+      sessionLifecycle = Lifecycle.PER_METHOD;
+      createSession(context);
+    }
```
dimas-b, line 162: If the test method has `@ExtendWith(CqlSessionExtension.class)` then we'll get a new instance of this class for each test method, AFAIK, so checking instance variables may not be reliable... It may be preferable to store session data in the `ExtrensionContext.Store`... :thinking:  WDYT?

testing/src/test/java/io/stargate/it/driver/CqlSessionExtension.java
olim7t: Right, that's more elegant, and not too complicated to add. I'll do it.

testing/src/test/java/io/stargate/it/driver/CqlSessionExtension.java
olim7t: The goal of `sessionLifeCycle` is to skip method-level work when we are in "per-class" mode: if `@ExtendWith` is on the test class, there is a single extension instance and both `beforeAll` and `beforeEach` will be invoked. We don't want to start another session in the `beforeEach`. Conversely, we don't want to close the session in `afterEach` because it's needed for the following methods.

If `@ExtendWith` is on the method, there is indeed one extension instance per method. `beforeAll` and `afterAll` are never called, so each extension instance will hit the condition `sessionLifecycle == null` and start/stop its own session.

testing/src/test/java/io/stargate/it/driver/CqlSessionExtension.java
olim7t: Did you find a definitive answer for that? Unfortunately it's not really covered in the documentation...

It should be pretty easy to test with a dummy extension.

testing/src/test/java/io/stargate/it/cql/BoundStatementTest.java
```diff
@@ -14,42 +15,51 @@
...
-    session.execute(
-        "CREATE TABLE IF NOT EXISTS test3 "
-            + "(pk1 int, pk2 int, v int, "
-            + "PRIMARY KEY ((pk1, pk2)))");
   }
 
   @Test
   @DisplayName("Should execute statement with positional values")
-  public void positionalValuesTest() {
+  public void positionalValuesTest(CqlSession session) {
```
tomekl007, line 62: can't we inject the `CqlSession session` in the test constructor and use it in methods instead of injecting for every test method?
(similar to what we are doing in java-driver tests)

testing/src/test/java/io/stargate/it/driver/CqlSessionExtension.java
dimas-b: true, but what if the annotation is both on the class and on a method?

testing/src/test/java/io/stargate/it/cql/BoundStatementTest.java
dimas-b: Constructor parameter injection should work technically. Still, it worth with `@ParameterizedTest` as well.

As for me, I kind of like having per-method parameters, personally :)

testing/src/test/java/io/stargate/it/driver/CqlSessionExtension.java
dimas-b: not definitive, but I'm pretty sure I was context creation being done after assigning test task to threads when I was digging around JUnit 5 code.... but it's worth double-checking. Having thought about this some more, I'm not sure about the case when test methods are executed by different threads.

testing/src/test/java/io/stargate/it/cql/BoundStatementTest.java
olim7t: I think either approach will work when using a class-level annotation. Not so sure about method-level though, I don't think JUnit creates a new test class instance for every method.

I like passing the session as a parameter because it lifts any ambiguity. It's not so bad with parameterized tests, you can put it last.

testing/src/test/java/io/stargate/it/driver/CqlSessionExtension.java
olim7t: OK, after a few tests it turns out that extensions can be accessed concurrently.

Method-level extensions: obviously not a problem because each method gets its own extension instance.
Class-level extension, across different test classes: same.

BUT with class-level extension, across test methods in the same class: if we allow parallel method execution (`junit.jupiter.execution.parallel.mode.default = concurrent`), then the same extension instance is accessed from different threads:
```
beforeAll, extension=ThreadPrintingExtension@4ea56a23, thread=ForkJoinPool-1-worker-1
beforeEach, extension=ThreadPrintingExtension@4ea56a23, thread=ForkJoinPool-1-worker-1
beforeEach, extension=ThreadPrintingExtension@4ea56a23, thread=ForkJoinPool-1-worker-2
```
So I'm keeping the fields volatile for safe publication. Note that we don't need stronger synchronization, because beforeAll clearly happens-before beforeEach of any method.

We probably don't want parallel method execution though, because this could easily have unintended side-effects if the tests are not carefully built around it. For example two methods writing and deleting rows in the same table. But there's nothing in the extension that formally prevents it. I'll add a note in the javadocs.

dimas-b: Nice. Thanks for the research @olim7t ! :)

testing/src/test/java/io/stargate/it/driver/CqlSessionExtension.java
olim7t: I don't see a strong advantage either way, so I'll keep it like this, it's more explicit.

testing/src/test/java/io/stargate/it/driver/CqlSessionExtension.java
olim7t: Then the method-level annotations are ignored. I think that's something JUnit does.

testing/src/test/java/io/stargate/it/driver/CqlSessionHelper.java
```diff
@@ -0,0 +1,22 @@
+package io.stargate.it.driver;
+
+import com.datastax.oss.driver.api.core.CqlSession;
+
+/**
+ * Utility methods to work with sessions in Stargate tests. To use this object, declare it as a
+ * parameter of a test method and {@link CqlSessionExtension} will inject it.
+ */
+public interface CqlSessionHelper {
```
olim7t, line 9: Keeping the name generic, in case we find other stuff to add here in the future.

testing/src/test/java/io/stargate/it/driver/CqlSessionExtension.java
dimas-b: Interesting :thinking: good point :+1:

dimas-b (approved):

PR #349, opened 2020-10-29
CDC schema information: Use classes in io.stargate.db.schema

jorgebay:
Use schema representation classes from `io.stargate.db.schema`.

Given that the schema is immutable with precomputed hashes, there's no need for a separate "identity".


dougwettlaufer:


```diff

Issues
======
+ Solved 1
- Added 1
           

Complexity decreasing per file
==============================
+ persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java  -3
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6422029&bid=20805747)

persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java
```diff
@@ -61,7 +60,7 @@
   public CompletableFuture<Void> publish(MutationEvent mutation) {
     TableSchemaManager schemaManager =
         tableSchemaManager.computeIfAbsent(
-            mutation.getTable().getId(), k -> new TableSchemaManager(this::createTableSchemaAsync));
+            mutation.getTable(), k -> new TableSchemaManager(this::createTableSchemaAsync));
```
dimas-b, line 63: Just wondering: if we get several schema changes in quick succession, how do we guarantee that the producer makes its schema changes in the same order? Does it matter?

dimas-b (approved):

persistence-api/src/main/java/io/stargate/db/cdc/SchemaAwareCDCProducer.java
jorgebay: We will use the `Schema` that is volatile and swapped each time there's a change, so it can cause race conditions when dealing with schema changes and mutations.

I think ordering doesn't matter though. In the CDC service, each mutation will get a table schema "version". The same will be occur on the producer implementation (kafka/pulsar): each mutation event record will have a corresponding schema version, more than one version can coexist in schema registries.

tomekl007: The ordering doesn't matter as long as changes are backward compatibility, isn't it?

PR #350, opened 2020-10-29
Config-store overrides at a higher level.

tomekl007:


config-store-api/src/main/java/io/stargate/config/store/api/ConfigWithOverrides.java
```diff
@@ -45,20 +47,35 @@ public ConfigWithOverrides(@Nonnull Map<String, Object> configMap) {
...
+   * <p>Prefixing with module name is done to avoid conflicts of overrides between modules.
+   *
    * @return the value with the highest priority or null if there is no value associated with the
    *     given settingName.
    */
   @Nullable
   public Object getWithOverrides(String settingName) {
-    String systemProperty = System.getProperty(settingName);
+    String settingNameWithModulePrefix = withModulePrefix(settingName);
+    String systemProperty = System.getProperty(settingNameWithModulePrefix);
```
dimas-b, line 66: System property values are always `String`, but this method returns `Object`. Is there an implicit assumption about the value type? Given that the base config is represented in YAML, non-String types are quite possible.

tomekl007: Yes, that is up to the caller to assert the actual type. 
YAML can return non-strings. If this will become a problem, we can restrict it to use only String values, I think.

dimas-b: I tend to think it might be safer to go the other way - change the return type to `String` here and if we have a use case for non-string values then we rework the API and the overriding mechanism... WDYT?

dimas-b: (sorry about late comments, somehow I did not notice this in previous PRs)

config-store-api/src/main/java/io/stargate/config/store/api/ConfigWithOverrides.java
```diff
@@ -45,20 +47,35 @@ public ConfigWithOverrides(@Nonnull Map<String, Object> configMap) {
    *
    * <p>3. Underlying config map
    *
+   * <p>When trying to retrieve the system property and environment variable, it will use the full
```
dougwettlaufer, line 50: System props are prefixed with `stargate.` which I think is important since you can have several `-D` for stargate, C*, logback, ...

config-store-api/src/main/java/io/stargate/config/store/api/ConfigWithOverrides.java
```diff
@@ -45,20 +47,35 @@ public ConfigWithOverrides(@Nonnull Map<String, Object> configMap) {
...
+   * System.getProperty("m_1.s")}, then {@code System.getenv("m_1.s")} and finally get the config
+   * from the underlying map using {@code Map.get("s")}.
+   *
+   * <p>Prefixing with module name is done to avoid conflicts of overrides between modules.
+   *
    * @return the value with the highest priority or null if there is no value associated with the
    *     given settingName.
    */
   @Nullable
   public Object getWithOverrides(String settingName) {
```
dougwettlaufer, line 64: Can we have two versions of this? One that adds the module prefix and another that uses the exact settingName as provided?

dougwettlaufer: Some of these are also not module specific like `stargate.auth_id` and `stargate.persistence_id`

tomekl007: I think that we should teat those setting as settings per the stargate module, so the client of this API would do:
`ConfigStore.getConfigForModule("stargate"). getWithOverrides("auth_id")`
by using this mechanism we can provide a map of setting global to stargate in the `stargate-config.yaml`.
So I don't see a need for the 2nd version of this method.

config-store-api/src/main/java/io/stargate/config/store/api/ConfigWithOverrides.java
tomekl007: See my answer here:
https://github.com/stargate/stargate/pull/350#discussion_r514933351

config-store-api/src/main/java/io/stargate/config/store/api/ConfigWithOverrides.java
tomekl007: We already have 2 use cases for non-string values for CDC. Enable/disable that is a `boolean` and `list` of CDC enabled tables. Maybe we can have a middle ground:
make the `getWithOverrrides` method returning `String` and wrap the `configMap` value within the:
`return String.valueOf(configMap.get(settingName));`
Using this mechanism we will allow overrides for properties that are supported using `System.setProperty` and `env_variables`. The caller will need to handle string and parsing. On the other hand, we will allow clients to `getConfigMap` directly if they are not expecting overrides. 
The other solution is to leave the `getWithOverrides` returning `Object` as it is now and handle that client-side. 
wdyt?

dimas-b: With `List` values, I wonder how we're going to handle overrides? A system property per list position or parse and reinterpret the `String` property value? Do we support lists of maps?

This is getting tricky :thinking: 

Do we intent overrides to be able to provide definitive values when the corresponding yaml element is not defined?

If we assume that yaml data is present, we could use its type information to correctly parse overriding `String` values.

Alternatively, what do you think about using JSON? I guess it would be easier to provide overrides at any level in that case.

tomekl007: yes, it is getting tricky - I will document the current limitations.

PR #351, opened 2020-10-29
CQL: Add integration tests for `system.local` and `system.peers`

mpenick:


dimas-b:
It looks like CI is hitting #365 (not related to these changes, apparently)

testing/src/test/java/io/stargate/it/cql/SystemTablesTest.java
```diff
@@ -1,29 +1,94 @@
...
+            .execute(
+                SimpleStatement.builder("SELECT * FROM system.local").setNode(localNode).build())
+            .one();
+    assertThat(localRow).isNotNull();
+    assertThat(localRow.getInetAddress("listen_address"))
+        .isEqualTo(localNode.getListenAddress().map(s -> s.getAddress()).orElse(null));
+
+    ResultSet rs =
+        session.execute(
+            SimpleStatement.builder("SELECT * FROM system.peers").setNode(localNode).build());
```
dimas-b, line 47: I'd split these test into separate methods for better isolation... is there any reason to keep them together?

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
```diff
@@ -467,33 +497,26 @@ public String rack() {
...
+          index,
+          i -> {
+            try {
+              try (ServerSocket socket = new ServerSocket(0)) {
+                return socket.getLocalPort();
+              }
+            } catch (IOException e) {
+              throw new UncheckedIOException(e);
+            }
+          });
```
dimas-b, line 519: I think this is a bit risky. We immediately release the socket, so there's no strict guarantee that the next call will not chose the same port, I guess :) This may be critical when we start multi-node Stargate clusters all at once.

testing/src/test/java/io/stargate/it/cql/SystemTablesTest.java
```diff
@@ -1,29 +1,94 @@
...
+@StargateSpec(nodes = 2)
 @ExtendWith(CqlSessionExtension.class)
-public class SystemTablesTest extends BaseOsgiIntegrationTest {
+class SystemTablesTest extends BaseOsgiIntegrationTest {
+  @Test
+  @DisplayName("Should expose Stargate addresses in system.local and system.peers")
+  public void querySystemLocalAndPeers(CqlSession session) {
+    Iterator<Node> nodes = session.getMetadata().getNodes().values().iterator();
+
+    Node localNode = nodes.next();
```
dimas-b, line 35: Can we be sure that the first node in the list is the one to which we connected? Can we get the local node address from the session?

dimas-b: I know that the `StargateContainer` extension always uses the first node for injecting connection info objects, but in this case the CQL session extension is also involved, so the association of the node and session is less direct :thinking:

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
jorgebay: I also think preallocating is safer, we can define an upper bound (like 10 nodes) for all tests and work with that.

testing/src/test/java/io/stargate/it/cql/SystemTablesTest.java
mpenick: It's the local node because the test makes explicit: `SimpleStatement.builder("SELECT * FROM system.local").setNode(localNode)`.

testing/src/test/java/io/stargate/it/cql/SystemTablesTest.java
mpenick: They're a bound together because the test explicitly sets the queried node using `setNode(localNode)`. I expect that `localNode` will not appear in the the `peers` table.

testing/src/test/java/io/stargate/it/storage/StargateContainer.java
mpenick: Good catch. I'll pre-allocate.

jorgebay (approved):
Nice!

testing/src/test/java/io/stargate/it/cql/SystemTablesTest.java
dimas-b: Ah... I missed that, TIL :)

testing/src/test/java/io/stargate/it/cql/SystemTablesTest.java
dimas-b: Yep, now that I see the importance of the `.setNode` call, it makes sense :facepalm:

dimas-b (approved):

dougwettlaufer (approved):

dimas-b (approved):

PR #354, opened 2020-10-29
GraphQL: Surface conversion errors to the client (fixes #256)

### LINKED ISSUE #256: Improve error handling when building GraphQL schema

olim7t:
Expected or unexpected errors can happen when we convert the CQL data model into a GraphQL schema. Up to this point, those errors are logged on the server. The user who manages the data model won't necessarily have access to those logs.

The goal of this PR is to surface them in the GraphQL schema. The best way I've found is to add a special `__conversionWarnings` query. I also added a comment on this query with all the warnings, because it might be convenient to have them in `schema.graphql` without having to execute a query (and I think there won't be a lot of warnings in general).

Example scenarios that produce warnings:
* tuple columns (not implemented yet): `create table foo(k int primary key, t tuple<int,text>)`
* name clash between two tables: `create table foo_bar(k int primary key); create table "fooBar"(k int primary key)`
* name clash between two columns within the same table: `create table foo(k int primary key, a int, "A" text)`

How it looks in the schema definition:
```
type Query {
  [...]
  # Warnings encountered during the translation of the backend schema into GraphQL.
  # For convenience they are also listed below:
  # - Can't convert column foo.a because its GraphQL name a would collide with column A
  __conversionWarnings: [String]
}
```
How it looks as a query:
```
query {
  __conversionWarnings
}
```
Result:
```
{
  "data": {
    "__conversionWarnings": [
      "Can't convert column foo.a because its GraphQL name a would collide with column A"
    ]
  }
}
```

### DIFF
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/DmlSchemaBuilder.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/DmlSchemaBuilder.java
@@ -16,8 +16,8 @@
 package io.stargate.graphql.schema;
 
 import static graphql.Scalars.GraphQLString;
+import static graphql.schema.GraphQLList.list;
 
-import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import graphql.Scalars;
 import graphql.introspection.Introspection;
@@ -51,15 +51,13 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 class DmlSchemaBuilder {
-  private static final Logger log = LoggerFactory.getLogger(DmlSchemaBuilder.class);
 
   private final Persistence persistence;
   private final AuthenticationService authenticationService;
   private final Map<Table, GraphQLOutputType> entityResultMap = new HashMap<>();
+  private final List<String> warnings = new ArrayList<>();
   private final FieldInputTypeCache fieldInputTypes;
   private final FieldOutputTypeCache fieldOutputTypes;
   private final FieldFilterInputTypeCache fieldFilterInputTypes;
@@ -72,9 +70,9 @@ class DmlSchemaBuilder {
     this.authenticationService = authenticationService;
     this.keyspace = keyspace;
 
-    this.nameMapping = new NameMapping(keyspace.tables(), keyspace.userDefinedTypes());
-    this.fieldInputTypes = new FieldInputTypeCache(this.nameMapping);
-    this.fieldOutputTypes = new FieldOutputTypeCache(this.nameMapping);
+    this.nameMapping = new NameMapping(keyspace.tables(), keyspace.userDefinedTypes(), warnings);
+    this.fieldInputTypes = new FieldInputTypeCache(this.nameMapping, warnings);
+    this.fieldOutputTypes = new FieldOutputTypeCache(this.nameMapping, warnings);
     this.fieldFilterInputTypes =
         new FieldFilterInputTypeCache(this.fieldInputTypes, this.nameMapping);
   }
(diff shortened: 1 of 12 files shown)


graphqlapi/src/main/java/io/stargate/graphql/schema/DmlSchemaBuilder.java
```diff
@@ -304,16 +310,22 @@ private GraphQLFieldDefinition buildDelete(Table table) {
...
+        try {
+          fields.add(
+              GraphQLInputObjectField.newInputObjectField()
+                  .name(graphqlName)
+                  .type(fieldFilterInputTypes.get(column.type()))
+                  .build());
+        } catch (Exception e) {
+          warnings.add(
+              String.format(
+                  "Could not create filter input type for column %s.%s, skipping (%s)",
```
jorgebay, line 324: NIT: We could add a more clear notation for column+table: `"[...] for column %s in table %s [...]"`

graphqlapi/src/main/java/io/stargate/graphql/schema/DmlSchemaBuilder.java
```diff
@@ -423,21 +443,44 @@ public GraphQLObjectType buildType(Table table) {
...
 
-  private GraphQLFieldDefinition.Builder buildOutputField(Table table, Column columnMetadata) {
-    return new GraphQLFieldDefinition.Builder()
-        .name(nameMapping.getGraphqlName(table, columnMetadata))
-        .type(fieldOutputTypes.get(columnMetadata.type()));
+  private GraphQLFieldDefinition buildWarnings() {
+    StringBuilder description =
+        new StringBuilder(
+            "Warnings encountered during the translation of the backend schema into GraphQL.\n"
+                + "For convenience they are also listed below:");
```
jorgebay, line 470: NIT: Can we change the message when there aren't any issues to avoid scaring the user, something like "No issues found when generating the GraphQL schema."

jorgebay (approved):
I think its a super convenient solution from the user pov 👍  

Minor nits below.

graphqlapi/src/main/java/io/stargate/graphql/schema/FieldInputTypeCache.java
```diff
@@ -50,18 +50,31 @@ protected GraphQLInputType compute(Column.ColumnType columnType) {
   }
 
   private GraphQLInputType computeUdt(UserDefinedType udt) {
+    String graphqlName = nameMapping.getGraphqlName(udt);
+    if (graphqlName == null) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Could find a GraphQL name mapping for UDT %s, "
```
dougwettlaufer, line 57:
```suggestion
              "Could not find a GraphQL name mapping for UDT %s, "
```

graphqlapi/src/main/java/io/stargate/graphql/schema/DmlSchemaBuilder.java
```diff
@@ -369,27 +381,35 @@ private GraphQLType buildQueryOptionsInputType() {
...
-    for (Column columnMetadata : table.columns()) {
-      try {
-        GraphQLInputObjectField field =
-            GraphQLInputObjectField.newInputObjectField()
-                .name(nameMapping.getGraphqlName(table, columnMetadata))
-                .type(fieldInputTypes.get(columnMetadata.type()))
-                .build();
-        input.field(field);
-      } catch (Exception e) {
-        log.error(
```
dougwettlaufer, line 391: We don't at least want a warn on the server side too?

dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/DmlSchemaBuilder.java
olim7t: Right, I'll add them back.

I think error logs should target operators, so I'm always wary about logging too much: for example if the CQL data model causes a a name clash in `NameMapping`, it's a user issue, not something an operator should be paged about.
But the catch blocks in this class cover unexpected errors, so it's reasonable to log at WARN level.

PR #355, opened 2020-10-29
Support Avatica protocol Frames (paging)

dimas-b:


sql/server/src/main/java/io/stargate/api/sql/server/avatica/StargateMeta.java
```diff
@@ -320,8 +321,15 @@ public ExecuteBatchResult executeBatch(
...
 
   @Override
-  public Frame fetch(StatementHandle h, long offset, int fetchMaxRowCount) {
-    throw new UnsupportedOperationException();
+  public Frame fetch(StatementHandle h, long offset, int fetchMaxRowCount)
+      throws NoSuchStatementException, MissingResultsException {
+    StatementHolder statement = statement(h);
+    ExecutionResult result = statement.result();
+    if (result == null) {
+      throw new MissingResultsException(h);
```
dougwettlaufer, line 329: Will `h` provide enough information or should we add some context to this exception?

sql/server/src/main/java/io/stargate/api/sql/server/avatica/StatementHolder.java
```diff
@@ -52,12 +58,25 @@ public static StatementHolder empty(String connectionId, int id, DataStore dataS
...
 
-  public Prepared prepare(String sql) {
-    return prepare(sql, dataStore);
+  private ExecutionResult execute(Prepared prepared, List<TypedValue> params) {
+    List<Object> localParams =
+        params.stream().map(TypedValue::toLocal).collect(Collectors.toList());
+    Iterable<Object> rows = prepared.query().execute(localParams);
+    ExecutionResult r = new ExecutionResult(prepared, rows.iterator());
+    result = r;
+    return r;
```
dougwettlaufer, line 71: Minor but can this be

```suggestion
    result = new ExecutionResult(prepared, rows.iterator());
    return result;
```

sql/server/src/main/java/io/stargate/api/sql/server/avatica/StargateMeta.java
dimas-b: `MissingResultsException` is defined in Avatica, we have to provide `h`.

I'm not sure about an extended message... the context should be clear, I hope. This is kind of an internal failure mode where the client requests data without first executing something.

dougwettlaufer (approved):

Issue #332, opened 2020-10-27
Migrate direct usages of `System.getProperty()` to config store API.


tomekl007:
Currently, we are using `System.getProperty` directly in multiple places. They should be migrated to use the config-store API instead. For example, in CassandraPersistenceActivator:
```
String clusterName = System.getProperty("stargate.cluster_name", "stargate-cassandra");
```
it should be replaced with:
```
String clusterName = Optional.ofNullable(configStore.getConfigForModule("stargate").get("cluster_name")).orElse("stargate-cassandra");
```
to have the same read path (through config-store API) for every setting.

dougwettlaufer:
Are we replacing all of them or are we keeping things like `System.getProperty("os.arch")`?

tomekl007:
> Are we replacing all of them, or are we keeping things like `System.getProperty("os.arch")`?

It is yet to be decided but replacing static initializers that are using properties maybe not feasible.

tatu-at-datastax:
We may want to move away from direct System.getProperty() usage (... although it is rather prevalent by now). But not by using config-store API, rather use Quarkus facilities.

PR #362, opened 2020-11-02
Mechanism for setting kafka specific producer settings

tomekl007:


producer-kafka/src/test/java/io/stargate/producer/kafka/KafkaProducerActivatorTest.java
```diff
@@ -104,4 +157,126 @@ public void shouldRegisterServiceOnlyOnceEvenIfStartIsCalledSecondTime()
...
+
+  @Test
+  public void shouldNotStartIfOnlyMetricsServiceIsRegistered() throws InvalidSyntaxException {
+    // given
+    BundleContext bundleContext = mock(BundleContext.class);
+    KafkaProducerActivator kafkaProducerActivator = new KafkaProducerActivator();
+    kafkaProducerActivator.start(bundleContext);
+    ServiceEvent metricsServiceEvent = mock(ServiceEvent.class);
+    try (MockedStatic<BundleUtils> bundleUtilsMock = mockStatic(BundleUtils.class)) {
+      mockMetricsServiceNotification(bundleContext, metricsServiceEvent, bundleUtilsMock);
```
dimas-b, line 169: If we're starting to use `BundleUtils` as a sort of "context" object for interacting with the OSGi env., wouldn't it be nice to refactor it to be an interface perhaps, and just have a regular mock in tests? Mocking static methods does not look elegant from my POV... just my 2 cents.

producer-kafka/src/main/java/io/stargate/producer/kafka/configuration/DefaultConfigLoader.java
```diff
@@ -29,22 +30,32 @@
...
 public class DefaultConfigLoader implements ConfigLoader {
-
+  public static final String CONFIG_STORE_MODULE_NAME = "cdc.kafka";
   private static final Pattern CDC_KAFKA_PRODUCER_SETTING_PATTERN =
       Pattern.compile(String.format("^(%s)\\.(.*.)", CDC_KAFKA_PRODUCER_SETTING_PREFIX));
 
   @Override
-  public CDCKafkaConfig loadConfig(Map<String, Object> options) {
+  public CDCKafkaConfig loadConfig(ConfigStore configStore) {
+    return loadConfig(configStore.getConfigForModule(CONFIG_STORE_MODULE_NAME).getConfigMap());
```
dimas-b, line 39: Sorry, I'm a bit lost... Do config overrides happen in this case?

producer-kafka/src/main/java/io/stargate/producer/kafka/configuration/DefaultConfigLoader.java
```diff
@@ -29,22 +30,32 @@
...
     String schemaRegistryUrl = getSchemaRegistryUrl(kafkaProducerSettings);
+    validateBootstrapServersPresent(kafkaProducerSettings);
     putProducerSerializersConfig(kafkaProducerSettings);
     return new CDCKafkaConfig(
         topicPrefixName, schemaRegistryUrl, kafkaProducerSettings, metricsConfig);
   }
 
+  private void validateBootstrapServersPresent(Map<String, Object> kafkaProducerSettings) {
+    // this should not throw
+    getStringSettingValue(kafkaProducerSettings, ProducerConfig.BOOTSTRAP_SERVERS_CONFIG);
```
dimas-b, line 56: Would it be possible to provide methods like `getStringSettingValue` in the Config API module?.. either directly in hte config interface of by means of a utility class?

producer-kafka/src/test/java/io/stargate/producer/kafka/KafkaProducerActivatorTest.java
tomekl007: I am also not a fan of this static method. I would be happy to refactor it, but for sure, in a separate PR (issue).
I think that we have a task with refactoring OSGi Activators' logic, but I cannot find it.

dimas-b: I think we talked about OSGi activation, but never files an issue :) +1 to opening one now.

producer-kafka/src/main/java/io/stargate/producer/kafka/configuration/DefaultConfigLoader.java
tomekl007: When I was thinking about it more, I think that currently, we don't need to support them in the CDC producer - it is documented in the `README.md.` The main reason behind this is that customers can provide any Kafka producer setting using `cdc.kafka.producer.` Because of that, we need to operate on the Map directly.
I think that there is no need for supporting overrides here. 
The override mechanism for config-store was created mainly for this:
https://github.com/stargate/stargate/issues/332

dimas-b: I see. Thanks!

producer-kafka/src/main/java/io/stargate/producer/kafka/configuration/DefaultConfigLoader.java
tomekl007: That is actually an excellent idea. Created task:
https://github.com/stargate/stargate/issues/363
It does not block this PR - it can be addressed afterward, and then I will adapt this PR.

dimas-b (approved):

producer-kafka/src/test/java/io/stargate/producer/kafka/KafkaProducerActivatorTest.java
tomekl007: Created: https://github.com/stargate/stargate/issues/364

PR #366, opened 2020-11-02
Adding JWT Authz - Read Path

dougwettlaufer:
This PR will be the first of several changes to keep things small and easily reviewable. Following changes will be

1. JWT authz on write path
2. Integration with restapi


authnz/pom.xml
```diff
@@ -1,13 +1,29 @@
...
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
     <groupId>io.stargate</groupId>
     <artifactId>stargate</artifactId>
     <version>0.0.20-SNAPSHOT</version>
   </parent>
   <groupId>io.stargate.auth</groupId>
-  <artifactId>authentication</artifactId>
+  <artifactId>authnz</artifactId>
```
dougwettlaufer, line 12: Renaming since this now handles authentication (authn) and authorization (authz)

authnz/src/main/java/io/stargate/auth/AuthnzService.java
```diff
@@ -15,11 +15,30 @@
...
+
+  ResultSet executeDataWriteWithAuthorization(
+      Callable<ResultSet> action, String token, List<String> primaryKeyValues, Table tableMetadata)
+      throws Exception;
+
+  ResultSet executeSchemaReadWithAuthorization(
+      Callable<ResultSet> action, String token, String keyspace, String table) throws Exception;
+
+  ResultSet executeSchemaWriteWithAuthorization(
+      Callable<ResultSet> action, String token, String keyspace, String table) throws Exception;
```
dougwettlaufer, line 43: These 2 are subject to change in following PRs

restapi/src/main/java/io/stargate/web/resources/RowResource.java
```diff
@@ -123,14 +123,21 @@ public Response getRows(
...
+                              .select()
+                              .from(keyspaceName, tableName)
+                              .where(
+                                  buildWhereClause(
+                                      localDB, keyspaceName, tableName, request.getRequestURI()))
+                              .consistencyLevel(ConsistencyLevel.LOCAL_QUORUM)
+                              .execute(),
+                      token,
+                      idFromPath(request.getRequestURI()),
+                      db.getTable(localDB, keyspaceName, tableName));
```
dougwettlaufer, line 140: Sample of how this will be used. Further integration will be in later PRs

authnz/src/main/java/io/stargate/auth/AuthnzService.java
```diff
@@ -15,11 +15,30 @@
  */
 package io.stargate.auth;
 
-public interface AuthenticationService {
+import io.stargate.db.datastore.ResultSet;
+import io.stargate.db.schema.Table;
+import java.util.List;
+import java.util.concurrent.Callable;
+
+public interface AuthnzService {
```
dimas-b, line 23: WDYT about adding a separate interface for the "authz" part (even if we implement both in the same class)?

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
```diff
@@ -69,6 +79,148 @@ public StoredCredentials validateToken(String token) throws UnauthorizedExceptio
...
+    List<Column> keys = tableMetadata.primaryKeyColumns();
+
+    if (primaryKeyValues.size() > keys.size()) {
+      throw new IllegalArgumentException("Provided more primary key values than exists");
+    }
+
+    for (int i = 0; i < primaryKeyValues.size(); i++) {
+      // If one of the columns exist as a field in the JWT claims and the values do not match then
+      // the request is not allowed.
+      if (stargateClaims.has(STARGATE_PREFIX + keys.get(i).name())) {
```
dimas-b, line 187: Do we use column names without their containing table name? Is this descriptive enough?

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
```diff
@@ -69,6 +79,148 @@ public StoredCredentials validateToken(String token) throws UnauthorizedExceptio
...
+
+    preCheckDataReadWrite(stargateClaims, primaryKeyValues, tableMetadata);
+
+    // Just return the result. No value in doing a post check since we can't roll back anyway.
+    return action.call();
+  }
+
+  @Override
+  public ResultSet executeSchemaReadWithAuthorization(
+      Callable<ResultSet> action, String token, String keyspace, String table) throws Exception {
```
dimas-b, line 138: Since we verify access to the specified table, would it make sense to pass it as a parameter to the `action` to protect against mistakes of using different tables inside the action?

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthTableBasedService.java
```diff
@@ -258,4 +261,34 @@ public StoredCredentials validateToken(String token) throws UnauthorizedExceptio
 
     return storedCredentials;
   }
+
+  @Override
+  public ResultSet executeDataReadWithAuthorization(
```
dimas-b, line 266: WDYT about `authorizedRead(...)` for brevity?

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
```diff
@@ -69,6 +79,148 @@ public StoredCredentials validateToken(String token) throws UnauthorizedExceptio
...
+  }
+
+  private void postCheckDataRead(JSONObject stargateClaims, ResultSet resultSet)
+      throws JSONException, UnauthorizedException {
+    if (resultSet == null || resultSet.rows() == null) {
+      return;
+    }
+
+    // If scopes/fields line up with the resultset then return
+    for (Row row : resultSet.rows()) {
```
dimas-b, line 210: I'm not sure how well this is going to work for clients. IIRC, the result set may not support re-iteration and if we iterate all the rows here, the higher level client may not 'see' any data... :thinking: It might be necessary to wrap the result set.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
```diff
@@ -69,6 +79,148 @@ public StoredCredentials validateToken(String token) throws UnauthorizedExceptio
...
+    preCheckSchemaWrite(stargateClaims, keyspace, table);
+
+    // Just return the result. No value in doing a post check since we can't roll back anyway.
+    return action.call();
+  }
+
+  private JSONObject extractClaimsFromJWT(String token) throws JSONException {
+    // Grab the custom claims from the JWT. It's safe to work with the JWT as a plain Base64 encoded
+    // json object here since by this point we've already authenticated the request.
+    String[] parts = token.split("\\.");
```
tomekl007, line 161: I think that it is better to parse(compile) this regexp once, using `java.util.regex.Pattern` and reuse it here instead of compiling it every time.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
```diff
@@ -69,6 +79,148 @@ public StoredCredentials validateToken(String token) throws UnauthorizedExceptio
...
+    String decodedPayload = new String(Base64.getUrlDecoder().decode(parts[1]));
```
tomekl007, line 162: I would add a simple check (parts.size() == 2) and if it is not met, throw `IllegalArgumentException`

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
```diff
@@ -69,6 +79,148 @@ public StoredCredentials validateToken(String token) throws UnauthorizedExceptio
...
+
+    // If scopes/fields line up with the resultset then return
+    for (Row row : resultSet.rows()) {
+      for (Column col : row.columns()) {
+        if (stargateClaims.has(STARGATE_PREFIX + col.name())) {
+
+          String stargateClaimValue = stargateClaims.getString(STARGATE_PREFIX + col.name());
+          String columnValue = row.getString(col.name());
+          if (!stargateClaimValue.equals(columnValue)) {
+            throw new UnauthorizedException("Not allowed to access this resource");
```
tomekl007, line 217: 1. Is ordering of iteration important here?
2. If we already fetched the full result set, maybe it would be better to construct an exception with all problematic values and give a more detailed message (instead of `this`). But I am not sure if this is a good solution from a security perspective.

authnz/src/main/java/io/stargate/auth/AuthnzService.java
dougwettlaufer: I thought about having an Authentication.java and Authorization.java but wasn't sure if there would be enough methods to warrant having both.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
dougwettlaufer: Whatever field is being used should ideally be a unique identifier for a user. Meaning whether they're querying `shopping_cart` or `orders` they should still be constrained to only seeing their data.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
dougwettlaufer: Sure, I can look into that in the following PR for schema related actions.

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthTableBasedService.java
dougwettlaufer: Sure, I can change to `authorizedDataRead` to distinguish it from `authorizedSchemaRead`

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
dougwettlaufer: Oh, good point. What about cloning?

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
dougwettlaufer: Right. Forgot to pull that out.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
dougwettlaufer: To be extra safe I can. As part of authentication it'll make sure that the JWT is properly formatted.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
dougwettlaufer: > Is ordering of iteration important here?

Actually I could probably flip this around since there will be fewer claims than rows

> If we already fetched the full result set, maybe it would be better to construct an exception with all problematic values and give a more detailed message (instead of this). But I am not sure if this is a good solution from a security perspective.

Correct, from a security standpoint we do not want to leak too much information. Although I'm still trying to decide if it should be an exception or just filter out the things that don't match.

authnz/src/main/java/io/stargate/auth/AuthnzService.java
dimas-b: In my head, the nature of those "execute with permission" methods is very different from create/validate token, hence my wish to keep them in different classes / interfaces... just my 2 cents.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
dimas-b: Ah, so it's just about finding / checking the user :zap: it was not apparent to me before :)

Are we going to manage JWT for end users of the application that connects to Stargate?

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
dimas-b: What do you mean by cloning? Mere object copies may not be enough at it ultimately links to paging state in C* :thinking:

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
dougwettlaufer: No there will need to be doc around this. The idea is use whatever auth provider (e.g. auth0, keycloak, ...) and as long as your JWTs have the proper claims and you pass stargate the URL for JWK (a standard endpoint where auth providers expose the public key for verifying a JWT) then we're able to validate them.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
dougwettlaufer: Oh so if I copy the resultSet and iterate over the copy it could still mess with the paging state of the original?

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
dimas-b: Thanks for the info, @dougwettlaufer ! This is LGTM. However, given a very specific nature of the key column links to JWT and the "authz" service, I believe this adds weight to having separate interfaces for the old `AuthenticationService` and the new "authz" methods. WDYT?

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
dimas-b: It will not "break", I think, but I believe we cannot be sure we get exactly the same results.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJwtService.java
dougwettlaufer: Sure, I can split into Authentication.java and Authorization.java. I was on the fence about it anyway.

persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedResultSet.java
```diff
@@ -104,7 +106,15 @@ static ResultSet create(
...
-      fetchedRows.addLast(new ArrayListBackedRow(columns, rowValues, driverProtocolVersion));
+      ArrayListBackedRow row = new ArrayListBackedRow(columns, rowValues, driverProtocolVersion);
+
+      if (authzFilter != null) {
+        if (authzFilter.test(row)) {
+          fetchedRows.addLast(row);
+        }
+      } else {
+        fetchedRows.addLast(row);
+      }
```
dimas-b, line 117: nit: how about:
```
if (authzFilter == null || authzFilter.test(row)) {
  fetchedRows.addLast(row);
}
```
As for me, it is easier to read :)

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Server.java
```diff
@@ -212,6 +212,7 @@ private void close() {
   }
 
   public static class Builder {
+
```
dimas-b, line 215: nit: is this change required?

dimas-b (approved):

auth-api/pom.xml
```diff
@@ -34,6 +34,22 @@
...
+    <dependency>
+      <groupId>io.stargate.db</groupId>
+      <artifactId>persistence-api</artifactId>
+      <version>${project.version}</version>
+      <scope>provided</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>com.google.guava</groupId>
+          <artifactId>guava</artifactId>
+        </exclusion>
```
dimas-b, line 46: tangential: is there a way to define the POMs in such as way that dependents will only pull what is exported in OSGi?

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthzJwtService.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.Base64;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.regex.Pattern;
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AuthzJwtService implements AuthorizationService {
```
dimas-b, line 35: Where do we register this service? I could not find the code :)

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Server.java
dougwettlaufer: According to the formatter it is :)

dimas-b: :shrug:

dimas-b (approved):

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthzJwtService.java
dougwettlaufer: Fixed.

persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedResultSet.java
dougwettlaufer: Fixed.

persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedResultSet.java
```diff
@@ -159,7 +161,7 @@ private boolean hasNextRow() {
   private Row nextRow() {
     while (true) {
       Row nextRow = fetchedRows.pollFirst();
-      if (nextRow != null) {
+      if (nextRow != null && (authzFilter == null || authzFilter.test(nextRow))) {
```
dimas-b, line 164: Is there any particular reason to avoid filtering in `processNewPage`?... that would allow us to have the filtering logic only in one place in this class :thinking:

dougwettlaufer: I was going to but `processNewPage` is only called from the constructor which I don't really have a good way to get the authzFilter into.

dimas-b: Since we have filtering in two places now, could you add unit tests for this class to make sure all call paths are covered by the filter?

dougwettlaufer: Sure

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthzJwtService.java
```diff
@@ -0,0 +1,186 @@
...
+            if (stargateClaims.has(STARGATE_PREFIX + col.name())) {
+
+              String stargateClaimValue;
+              try {
+                stargateClaimValue = stargateClaims.getString(STARGATE_PREFIX + col.name());
+              } catch (JSONException e) {
+                log.warn("Unable to get stargate claim for " + STARGATE_PREFIX + col.name());
+                return false;
+              }
+              String columnValue = row.getString(col.name());
```
mpenick, line 83: Sorry, I'm not fully understanding this. Is this the arbitrary values of columns from the DB? Why is it comparing column values in the DB to values in the JWT payload?

dougwettlaufer: Yes. The idea is that this allows you to do RLAC with JWT claims.

So let's say you have a table `shopping_cart` with pk of `userid` and you want to write a JS app that allows users to login and see the contents of their cart. Rather than having a bunch of roles or doing some sort of backend filtering instead you can assign a new claim to the user's JWT with their userid. Then when you JS app makes a request to graphql with the JWT stargate will return only the data that the user is allowed to see. 

By looking through the claims and comparing to columns all of this can be done completely on the fly by adjusting what claims are in the JWTs rather than making config and/or code changes to filter different columns or tables.

mpenick: Thanks for the explanation. Cool idea.

persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedResultSet.java
dougwettlaufer: Added https://github.com/stargate/stargate/pull/386/files#diff-e692aeca1c123715ea4b6cc35cabafe62af29142e07261996a2412566aca36a7L4

dimas-b (approved):

PR #367, opened 2020-11-03
Add methods for extracting setting of an expected type from the config-map using config-store API

tomekl007:


config-store-api/src/main/java/io/stargate/config/store/api/ConfigWithOverrides.java
```diff
@@ -36,6 +39,84 @@ public ConfigWithOverrides(@Nonnull Map<String, Object> configMap, String module
...
+   * match the expectedType, it throws {@code IllegalArgumentException}.
+   *
+   * @param settingName - it will be used as a key of underlying configMap.
+   * @param expectedType - it will be used to check if the type of an actual list element value
+   *     matches.
+   * @return the list of values associated with settingName. Each value matches expectedType.
+   */
+  @Nonnull
+  @SuppressWarnings("unchecked")
+  public <T> List<T> getSettingValueList(String settingName, Class<T> expectedType) {
```
tomekl007, line 56: I think that methods for getting a list and raw types is enough for us at this point.
The YAML format supports only lists and maps as collection types. Currently, we don't have a use case for the map.
We could create an extensive API that uses guava's `TypeToken,` but:
- it's better to not add the guava dependency in the `API` module
- currently, we don't need more types

config-store-api/src/main/java/io/stargate/config/store/api/ConfigWithOverrides.java
```diff
@@ -36,6 +39,84 @@ public ConfigWithOverrides(@Nonnull Map<String, Object> configMap, String module
...
+    List<?> settingValue = getSettingValue(settingName, List.class);
+    return settingValue.stream()
+        .map(
+            v -> {
+              validateType(String.format("%s.list-value", settingName), v, expectedType);
+              return (T) v;
+            })
+        .collect(Collectors.toList());
```
dimas-b, line 64: `Collections.checkedList`?

dimas-b (approved):

config-store-api/src/main/java/io/stargate/config/store/api/ConfigWithOverrides.java
tomekl007: It will look like this:
```
    List settingValue = getSettingValue(settingName, List.class);
	settingValue.forEach(v -> validateType(String.format("%s.list-value", settingName), v, expectedType));
    return Collections.checkedList(settingValue, expectedType);
```
please note that I cannot do `List<?> settingValue = getSettingValue(settingName, List.class);`
because then `checkedList()` will not work.
I am not sure which version looks better. I am ok with both. wdyt?

dimas-b: this is a minor point. Feel free to go ahead with current version :+1:

config-store-api/src/main/java/io/stargate/config/store/api/ConfigWithOverrides.java
```diff
@@ -36,6 +39,84 @@ public ConfigWithOverrides(@Nonnull Map<String, Object> configMap, String module
...
+   * @return the list of values associated with settingName. Each value matches expectedType.
+   */
+  @Nonnull
+  @SuppressWarnings("unchecked")
+  public <T> List<T> getSettingValueList(String settingName, Class<T> expectedType) {
+    List<?> settingValue = getSettingValue(settingName, List.class);
+    return settingValue.stream()
+        .map(
+            v -> {
+              validateType(String.format("%s.list-value", settingName), v, expectedType);
```
dougwettlaufer, line 61: What's the purpose of `String.format("%s.list-value", settingName)`?

config-store-api/src/main/java/io/stargate/config/store/api/ConfigWithOverrides.java
dougwettlaufer: Is this method just a convenience method for `getSettingValue("foo", List.class)` or is this to make sure it's a list of strings rather than a list of integers?

dougwettlaufer (approved):

PR #368, opened 2020-11-03
GraphQL: Support Tuples

### LINKED ISSUE #233: GraphQL: Add support for tuple type
We should add support for CQL tuples in GraphQL.

jorgebay:
Resolves #233 .


graphqlapi/src/main/java/io/stargate/graphql/schema/types/TupleBuilder.java
```diff
@@ -0,0 +1,81 @@
...
+
+  public GraphQLOutputType buildOutputType() {
+    List<GraphQLFieldDefinition> fields = new ArrayList<>(subTypes.size());
+
+    for (int i = 0; i < subTypes.size(); i++) {
+      GraphQLOutputType t = (GraphQLOutputType) subTypes.get(i);
+      fields.add(
+          GraphQLFieldDefinition.newFieldDefinition()
+              .name(String.format("item%d", i))
+              .type(i == 0 ? new GraphQLNonNull(t) : t)
```
olim7t, line 52: I think CQL allows all fields to be null:
```
> create table foo(k int primary key, t tuple<int,int>);
> insert into foo(k,t) values (1,(null,null));
> select * from foo;

 k | t
---+--------------
 1 | (None, None)

```

graphqlapi/src/main/java/io/stargate/graphql/schema/types/TupleBuilder.java
```diff
@@ -0,0 +1,81 @@
...
+              .build());
+    }
+
+    return GraphQLObjectType.newObject()
+        .description(String.format("Represents a Tuple type with %d items", subTypes.size()))
```
olim7t, line 57: If we're going to mention the origin type maybe we should describe it fully, something like `Represents an <int,int> tuple`.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/DataTypeMapping.java
```diff
@@ -129,6 +130,39 @@ private static void formatUdt(
...
+
+      if (i > 0) {
+        if (item == null) {
+          hasANullItem = true;
+          continue;
+        }
+
+        if (hasANullItem) {
+          throw new UnsupportedOperationException(
+              "Tuple can have a null item followed by a non-null item");
```
olim7t, line 155:
```suggestion
              "Tuple can't have a null item followed by a non-null item");
```

olim7t (changes requested):

graphqlapi/src/main/java/io/stargate/graphql/schema/types/TupleBuilder.java
jorgebay: hm... I wanted to avoid showing the CQL types in the description. The description is a meta property of the schema and it can be viewed along side the GraphQL fields, so there shouldn't be a confusion about what those "n times" are.

graphqlapi/src/main/java/io/stargate/graphql/schema/types/TupleBuilder.java
jorgebay: TIL, nice catch!

I'll fix it.

jorgebay:
I've addressed the issues from the review.

graphqlapi/src/main/java/io/stargate/graphql/schema/types/TupleBuilder.java
mpenick: What do you think about names like `_0`, `_1`, etc. instead of `item0`, to further differentiated it from regular objects type.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/DataTypeMapping.java
```diff
@@ -129,6 +130,39 @@ private static void formatUdt(
...
+      Object item = mapValue.get("item" + i);
+
+      if (item == null) {
+        hasANullItem = true;
+        continue;
+      }
+
+      if (hasANullItem) {
+        throw new UnsupportedOperationException(
+            "Tuple can't have a null item followed by a non-null item");
```
mpenick, line 153: Why the restriction? I tried in cql and it works fine.

olim7t: You're right, we don't need that restriction.
It's my fault, I left a completely misleading comment on the original issue: https://github.com/stargate/stargate/issues/233#issuecomment-722698412

graphqlapi/src/main/java/io/stargate/graphql/schema/types/TupleBuilder.java
olim7t: So then do we need the description at all :) It's pretty easy to count the number of fields in the schema.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/DataTypeMapping.java
jorgebay: Oh, my bad, I'll fix it.

graphqlapi/src/main/java/io/stargate/graphql/schema/types/TupleBuilder.java
jorgebay: As a GraphQL type, having `itemN` properties feels more natural to me (probably influenced by C# tuples and javatuples).

graphqlapi/src/main/java/io/stargate/graphql/schema/types/TupleBuilder.java
jorgebay: I think it can be helpful, specially for complex types like `TupleIntTupleTextText`. This description can be viewed quickly as a tooltip on IDEs.

In any case, I don't feel very strongly about it, I can remove this part of the description.

olim7t (approved):

PR #369, opened 2020-11-03
GraphQL: Support now() and uuid() functions

### LINKED ISSUE #132: Support auto generating timeuuid fieldsSupport auto generating timeuuid fields
TimeUuuid columns are sometimes used as primary keys and users pass the now() function in the query to resolve them. Graphql does not support functions on mutations so a different method will need to be used to express this use case. Since a time uuids are difficult to generate, we should supply a method to be able to deterministically produce them.

```
mutation {
  insertCart(value:{
     _now: [userId]
    cartId: "0000-0000-0000-0000"
    quantity: 1 
  }) {
    value {
      quantity
    }
    applied
  }
  
}
```

jorgebay:
Resolves #132 .


dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/CustomScalar.java
```diff
@@ -171,6 +170,34 @@ private static long parseLong(Object v) {
...
+          String.format("Expected string value for Uuid, obtained: %s", v));
+    }
+
+    String stringValue = (String) v;
+    if (stringValue.equals("uuid()")) {
+      return Uuids.random();
+    }
+    return java.util.UUID.fromString(stringValue);
+  }
+
```
olim7t, line 200: When you rebase this will be in `StringCoercing`.

olim7t (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/CustomScalar.java
jorgebay: Rebased and moved :) 👍

PR #370, opened 2020-11-03
Refactor Activators common code and improve its testability

tomekl007:
I've created the `BaseActivator` that encapsulates activator logic.
It solves a use case when the service needs to wait for N other services. 
This pattern is used in the config-store and Kafka-CDC-producer.
I think that many client's modules may follow the same pattern. 
Let me know wdyt.


core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -0,0 +1,169 @@
...
+    logger.info("Starting {} ...", activatorName);
+    this.context = context;
+
+    for (Class<?> dependentService : dependentServices) {
+      ServiceReference<?> serviceReference =
+          context.getServiceReference(dependentService.getName());
+      if (serviceReference == null) {
+        logger.debug(
+            "{} service is null, registering a listener to get notification when it will be ready.",
+            dependentService.getName());
```
dimas-b, line 92: I guess this may be subject to a (low probability) race since we do not know when exactly the dependent service may be registered.

What do you think about using `org.osgi.util.tracker.ServiceTracker` instead?

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -0,0 +1,169 @@
...
+        context.addServiceListener(
+            this, String.format("(objectClass=%s)", dependentService.getName()));
```
dimas-b, line 94: Could we support richer filters? For example, services that depend on `Persistence` use filter like `(Identifier=%s)` instead of the class name.

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -0,0 +1,169 @@
...
+  /**
+   * Clients should override this method to create the Service that will be registered in the OSGi
+   * container. The dependant services will contain all services registered passed to the
+   * constructor of this class as {@code List<Class<?>> dependentServices}. The ordering will be the
+   * same. You can safely cast the objects to the expected types according to the dependentServices.
+   *
+   * @return ServiceAndProperties that has the service for OSGi registration and the properties that
+   *     will be passed.
+   */
+  protected abstract ServiceAndProperties createService(List<Object> dependentServices);
```
dimas-b, line 157: Could we provide type safe references to dependencies? One possible way is to take "service selector" objects in the constructor and populate them when the services become available.

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -0,0 +1,169 @@
...
+   * notification for dependant service contains a non-null value, it will start the target service.
+   * After the notification is handled, it checks if all dependentServices are not null. If they
+   * are, the client's provided {@link this#createService(List)} is called, and the service is
+   * registered. It will not register the service if it was already registered in the {@link
+   * this#start(BundleContext)}.
+   */
+  @Override
+  public synchronized void serviceChanged(ServiceEvent event) {
+    for (Class<?> dependentService : dependentServices) {
+      Object registeredService = BundleUtils.getRegisteredService(context, event, dependentService);
```
dimas-b, line 126: what about dependent service de-registration?

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -0,0 +1,169 @@
...
+
+  private synchronized void startServiceInternal(List<Object> dependentServices) {
+    if (started) {
+      logger.info("The {} is already started. Ignoring the start request.", activatorName);
+      return;
+    }
+    started = true;
+    ServiceAndProperties service = createService(dependentServices);
+    context.registerService(targetServiceClass.getName(), service.service, service.properties);
+    logger.info("Started {}....", activatorName);
```
dimas-b, line 145: nit: `Started` and `...` kind of contradict each other :)

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  private synchronized void startServiceInternal(List<Object> dependentServices) {
+    if (started) {
+      logger.info("The {} is already started. Ignoring the start request.", activatorName);
+      return;
+    }
+    started = true;
+    ServiceAndProperties service = createService(dependentServices);
+    context.registerService(targetServiceClass.getName(), service.service, service.properties);
```
dimas-b, line 144: Is there any particular reason to use the class _name_ for service registration as opposed to the class _object_?

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -0,0 +1,169 @@
...
+      // It will be started once all dependent services are registered
+      return;
+    }
+
+    // all dependant services are present
+    startServiceInternal(new LinkedList<>(registeredServices.values()));
+  }
+
+  @Override
+  public synchronized void stop(BundleContext context) throws Exception {
```
dougwettlaufer, line 111: In many cases this will need to be overwritten.

core/src/main/java/io/stargate/core/activator/BaseActivator.java
dougwettlaufer: Also, what about things that don't need to be registered? For example, a top level service like graphql doesn't need to register itself.

core/src/main/java/io/stargate/core/activator/BaseActivator.java
tomekl007: Very good idea, done.

core/src/main/java/io/stargate/core/activator/BaseActivator.java
tomekl007: I've added the `protected abstract void stopService();`  that should be called by the client

core/src/main/java/io/stargate/core/activator/BaseActivator.java
tomekl007: I don't see a way of making it more compile-time type-safe without reducing the extensibility of this base class. 
I think that at some point we need to operate on `Object`.
Could you elaborate more (give example) how would you make it type-safe?

core/src/main/java/io/stargate/core/activator/BaseActivator.java
tomekl007: I've reworked this example to use `Tracker`.
It already provides the way to unget service:
```
	public void removedService(ServiceReference<S> reference, T service) {
		context.ungetService(reference);
	}
```
So I think that we don't need to handle it explicitly.

core/src/main/java/io/stargate/core/activator/BaseActivator.java
tomekl007: done in: https://github.com/stargate/stargate/pull/370/commits/47f882cccf8a281b5680d272bf5d8f372b86f0ec

tomekl007: >
Is there any particular reason to use the class name for service registration as opposed to the class object?

the `registerService` API is not accepting the `targetServiceClass` of a `Class<?> type` as an argument.

tomekl007:
@dimas-b @dougwettlaufer I've changed the design to use `ServiceTracker` - it looks better now.
I've addressed all comments and added missing features. I also migrated `GrapqlActivator` to a new mechanism. Let me know wdyt.

core/src/main/java/io/stargate/core/activator/BaseActivator.java
dimas-b: Rough sketch:

```

class BaseActivator {
protected abstract Collection<ServicePointer<?>> dependencies();
... 
}

class ServicePointer<T> {

  ServicePointer(Class<T> targetClass, String filter) {...}

  // called by BaseActivator
  setService(T service) {...}

 // called by user (sub-class of BaseActivator)
T getService() {...}
}
```

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -0,0 +1,255 @@
...
+
+    private final Hashtable<String, String> properties;
+
+    public ServiceAndProperties(Object service, Hashtable<String, String> properties) {
+      this.service = service;
+      this.properties = properties;
+    }
+  }
+
+  public static class DependentService {
```
dimas-b, line 219: What do you think about renaming to `ServiceDeprendency` for the sake of clarify? I believe "dependent" refers to the entity that depends on something else, but this class is meant to represent that "something" on which our services depend.

Also, in another comment I suggested a using a class named like `ServicePointer`, enhanced with the ability to pass type-checked references to the dependent service.

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -0,0 +1,255 @@
...
+    private Class<?> className;
+
+    private Optional<String> identifier;
```
dimas-b, line 222: How about allowing any OSGi filter for generalizability?

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -0,0 +1,255 @@
...
+
+    private DependentService(Class<?> className, String identifier) {
+      this.className = className;
+      this.identifier = Optional.ofNullable(identifier);
+    }
+
+    public static DependentService constructDependentService(
```
dimas-b, line 229: `constructDependentService` -> `create(...)` for simplicity?

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -0,0 +1,255 @@
...
+      return new DependentService(
+          className, String.format("%s=%s", identifierKey, identifierValue));
+    }
+
+    public static DependentService constructDependentService(Class<?> className) {
+      return new DependentService(className, null);
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
dimas-b, line 240: Do we really need equality checks beyond the default object instance identity?

core/src/main/java/io/stargate/core/activator/BaseActivator.java
tomekl007: at some point, you will need to invoke `  setService(T service) {...}` that takes T, but the service retrieved here:
`public Object addingService(ServiceReference<Object> ref)`
is of an `Object` type, so it is impossible to call `setService(T)`.

Second issue - the dependencies must be of a `?` type. Because of that, when the sub-class of `BaseActivator` will get it, it is still impossible to have the type-safety. You will still need to cast:
```
@Override
  protected ServiceAndProperties createService(List<ServicePointer<?>> dependentServices) {
	  Object service = dependentServices.get(0).getService(); // need to cast
            ...
```
So I don't see how the ServicePointer will solve this issue.

core/src/main/java/io/stargate/core/activator/BaseActivator.java
tomekl007: I am not sure what you mean. The filter is used at a different level, see:
`new Tracker(
            context, context.createFilter(constructDependenciesFilter()), registeredServices);`
and it is constructed from a `String` (we are doing this for all other activators).

core/src/main/java/io/stargate/core/activator/BaseActivator.java
dimas-b: Yes, I think it's unavoidable in the java type system. However, I think it is preferable to have the base / utility class deal with type casts so that sub-classes (majority) will not have to do type casts. 

I think we can also check the actual type of the service object before injecting it into `ServicePointer` since we that the `Class` objects for its expected type (e.g. use `Class.cast(...)`).

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    /**
+     * It will try to match all dependentServices with a ServiceReference notification. After the
+     * notification is handled, it checks if all dependentServices are not null. If they are not,
+     * the client's provided {@link this#createService(List)} is called, and the service is
+     * registered. It will not register the service if it was already registered.
+     */
+    @Override
+    public Object addingService(ServiceReference<Object> ref) {
```
dimas-b, line 166: Should we also handle service removal? I'd guess if a dependency disappears we should notify the service impl. sub-class. In the simple case I suppose it makes sense to stop the service. When the dependency re-appears the service will be restarted.

Services that need more complex dependency tracking logic will probably have to use custom logic (i.e. not extend `BaseActivator`)... WDYT?

core/src/main/java/io/stargate/core/activator/BaseActivator.java
tomekl007: The check before injecting already exists:
```
  for (Map.Entry<Class<?>, Object> registeredService : registeredServices.entrySet()) {
        if (registeredService.getKey().isAssignableFrom(service.getClass())) {
          logger.debug("{} using service: {}", activatorName, ref.getBundle());
          registeredServices.put(registeredService.getKey(), service);
        }
      }
```

Regarding hiding it in the parent class, of course, it would be better. Unfortunately, even with your approach, sub-classes will need to cast. The reason for this is the wildcard `?`:
```
Collection<ServicePointer<?>> dependencies();
```
as long as you have one service pointer, it will work:
```
ServicePointer<String> stringServicePointer = new ServicePointer<>(String.class);
String service = stringServicePointer.getService(); // works
```
but please keep in mind that it will be a list of different `T` so for example:
```
List<ServicePointer<?>> dependencies = Arrays.asList(
				new ServicePointer<>(String.class),
				new ServicePointer<>(Integer.class)
		);
```
event if every single `ServicePointer` is type-safe the caller will still operate on `Object`:
```
		Object service1 = dependencies.get(0).getService();
		Object service2 = dependencies.get(0).getService();
```
and do the cast. Using `cast` does not solve this issue:
```
Object cast = dependencies.get(0).targetClass.cast(dependencies.get(0).getService());
```
So as long as the client needs to know exact types (and those are all our cases), the cast MUST be in the client code. wdyt?

core/src/main/java/io/stargate/core/activator/BaseActivator.java
tomekl007: From what I see, the use case that you described (restart service) is not implemented in any existing `Activator`.
I am ok with having that, but I think this should be a follow-up ticket. I would like this PR to have only refactoring changes without adding new features.

core/src/main/java/io/stargate/core/activator/BaseActivator.java
dimas-b: the `?` is only at the base class level. Sub-classes will not their types, e.g.

```
class ExampleServiceActivator extends BaseActivator {

  private final ServiceReference<Persistence> persistence = new ...

 @Override
 protected Collection<ServiceReference<?>> dependencies() {
   // this may need to be adjusted... not sure about exact signature, but should work in general, I believe
   return Collections.singletonList(persistence)
 }

 @Override
 void startServioce() { // or whatever 
   MyService = new MyService(persistence.get());
 }
}

class MyService {
 MyService(Persistence p) {...}
}
```

core/src/main/java/io/stargate/core/activator/BaseActivator.java
dimas-b: I mean allowing BaseActivator sub-classes to provide a general-purpose OSGI filter, not just the service identifier.

core/src/main/java/io/stargate/core/activator/BaseActivator.java
dimas-b: sgtm

core/src/main/java/io/stargate/core/activator/BaseActivator.java
tomekl007: I see your point but I don’t see a use case for it. We have only two cases: exact class and identifier. Both are supported by the current solution. If there will be a new use case, we can easily parameterize it or expose filter.

dimas-b: ok

core/src/main/java/io/stargate/core/activator/BaseActivator.java
tomekl007: That was a very good idea! Thank you. I've implemented the type-safe variant of this API using your approach, please take a look.

tomekl007:
@dimas-b I've migrated the API to type-safe according to your suggestion. It looks pretty solid now. Please take a look.
Once we merge this PR, I will create a follow-up ticket to migrate all existing `Activator`s to this new API.
It should be done separately because all other activators have no unit or integration tests, so migrating them will require a substantial amount of manual testing. I've migrated only `HealthCheckerActivator` and `GraphqlActivator` because they were simple to test, and I wanted to have N use-cases for this new `BaseActivator` class in this PR.
Once we merge this PR, all new activators should use this new mechanism (I will adapt the CDC and Kafka CDC producer activators)

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -0,0 +1,227 @@
...
+        Class<T> className, String identifierKey, String identifierValue) {
+      return new ServicePointer<>(
+          className, String.format("%s=%s", identifierKey, identifierValue));
+    }
+
+    public static <T> ServicePointer<T> create(Class<T> className) {
+      return new ServicePointer<>(className, null);
+    }
+
+    public T getService() {
```
dimas-b, line 218: nit: maybe simply `get` since the class is a "service pointer" already?

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -0,0 +1,227 @@
...
+    private final Hashtable<String, String> properties;
+
+    public ServiceAndProperties(Object service, Hashtable<String, String> properties) {
+      this.service = service;
+      this.properties = properties;
+    }
+  }
+
+  public static class ServicePointer<T> {
+    private Class<T> className;
```
dimas-b, line 197: `className`?.. maybe `expectedClass`?

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+    builder.append(")");
+    return builder.toString();
+  }
+
+  @Override
+  public synchronized void stop(BundleContext context) throws Exception {
+    if (started) {
+      logger.info("Stopping {}", activatorName);
+      stopService();
```
dimas-b, line 104: We automatically register the service, but do not automatically unregister it... this looks a bit awkward to me.

How about changing the sub-class lifecycle from `createService/stopService` to `startService/stopService` and delegate registration to sub-classes? This should also maturally suppose use cases where the sub-class registers more than one service object (from the same set of dependencies), which may soon be the case on the `feature-sql` branch.

dimas-b (commented):
LGTM overall :+1:  just one minor comment about create/stop vs. start/stop lifecycle (above).

core/src/main/java/io/stargate/core/activator/BaseActivator.java
tomekl007: The `Tracker` extends `ServiceTracker` and I think it handles the de-registration automatically:
```
	public void removedService(ServiceReference<S> reference, T service) {
		context.ungetService(reference);
	}
```

tomekl007: Re: registering in the child classes. I think that hiding this logic internally is quite convenient because you don't need to worry about it nor test it. For the N services use case, maybe it would be better to just rework the `BaseActivator` logic for this use case?

dimas-b: sgtm.

dimas-b (approved):

core/src/main/java/io/stargate/core/activator/BaseActivator.java
tomekl007: I've added de-registration of target service: https://github.com/stargate/stargate/pull/370/commits/d39e9e92f95d2672bf78635957e1173536fa2767

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -97,15 +100,30 @@ String constructDependenciesFilter() {
...
+      deregisterService();
     }
     tracker.close();
   }
 
+  private void deregisterService() {
+    if (targetServiceRegistration != null) {
+      ServiceReference<?> reference = targetServiceRegistration.getReference();
+      logger.info("Unget service {} from {}", reference.getBundle(), activatorName);
+      context.ungetService(reference);
```
dimas-b, line 123: Should we also unget the dependencies?

dimas-b (approved):

dougwettlaufer (approved):

core/src/main/java/io/stargate/core/activator/BaseActivator.java
tomekl007: No, they should be automatically `unget` by the `ServiceTracker`

PR #374, opened 2020-11-04
make config-store identifier key name unique

tomekl007:


config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreActivator.java
```diff
@@ -79,7 +79,7 @@ private synchronized void startConfigStore(Metrics metrics) {
     started = true;
 
     Hashtable<String, String> props = new Hashtable<>();
-    props.put("Identifier", CONFIG_STORE_YAML_IDENTIFIER);
+    props.put("ConfigStoreIdentifier", CONFIG_STORE_YAML_IDENTIFIER);
```
dimas-b, line 82: Just wondering: What is the reason for this change?

dimas-b (approved):

config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreActivator.java
tomekl007: I see that we have this pattern for authIdentifier (having unique selector name) 
It maybe easier to find usages of it instead of using Identifier name everywhere.

Issue #376, opened 2020-11-04
Stargate doesn't report proper options in the SupportedMessage


aboudreault:
We are running dse-server:6.8.5 with stargate and it seems that the supportedmessage doesn't return the proper options: like the protocol_version, product_types etc. The drivers are not able to configure properly for the server version without them.

Here are my options when connecting to dse directly:
```
<SupportedMessage(cql_versions=['3.4.5'], options={'SERVER_VERSION': ['6.8.5'], 'PRODUCT_TYPE': ['DATASTAX_APOLLO'], 'EMULATE_DBAAS_DEFAULTS': ['true'], 'COMPRESSION': ['snappy', 'lz4'], 'PAGE_UNIT': ['BYTES', 'ROWS'], 'PROTOCOL_VERSIONS': ['3/v3', '4/v4', '5/v5-beta', '65/dse-v1', '66/dse-v2']}, stream_id=0, trace_id=None)>
```

This is what I get when connecting to stargate:
```
<SupportedMessage(cql_versions=['3.4.5'], options={'PROTOCOL_VERSIONS': ['3/v3', '4/v4', '5/v5-beta'], 'COMPRESSION': ['snappy', 'lz4']}, stream_id=0, trace_id=None)>
```

aboudreault:
@mpenick I tested with an astra cluster and I see the proper options. So that might be related to our stargate driver endpoint...

PR #379, opened 2020-11-04
CQL: Fix supported options

mpenick:
Fixes https://github.com/stargate/stargate/issues/376


dimas-b (approved):

testing/src/test/java/io/stargate/it/cql/SupportedOptionsTest.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import org.jetbrains.annotations.NotNull;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+public class SupportedOptionsTest extends BaseOsgiIntegrationTest {
+  /** Custom session builder that records supported options. */
+  private static class RecordingSupportedOptionsSessionBuilder
```
olim7t, line 50: You can access the options using the driver's internal APIs:
```java
import com.datastax.oss.driver.internal.core.channel.DriverChannel;
import com.datastax.oss.driver.internal.core.context.InternalDriverContext;

private static Map<String, List<String>> getSupportedOptions(CqlSession session) {
  DriverChannel controlChannel =
      ((InternalDriverContext) session.getContext()).getControlConnection().channel();
  assertThat(controlChannel).as("Expected control connection to be connected").isNotNull();
  return controlChannel.getOptions();
}
```
This should allow you to write this test with the regular `CqlSessionExtension`.

olim7t (changes requested):
LGTM, but a suggestion to simplify the test.

testing/src/test/java/io/stargate/it/cql/SupportedOptionsTest.java
mpenick: Thanks.

mpenick:
Fixes https://github.com/stargate/stargate/issues/376

olim7t (approved):

Issue #340, opened 2020-10-28
Cache file read by config-store-yaml to improve performance


tomekl007:
Reading the file every time may be problematic from the performance perspective:
```
Map<String, Map<String, Object>> result =
          mapper.readValue(configFilePath.toFile(), yamlConfigType);
```
and we may consider adding a cache atop of it. I've created a simple JMH micro-benchmark comparing loading the file every time this method is called versus caching the file for 30 seconds (exact duration yet to be decided):
```

@Fork(1)
@Warmup(iterations = 5)
@Measurement(iterations = 10)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.MILLISECONDS)
public class ConfigStoreYamlBenchmark {
  private static final int NUMBER_OF_ITERATIONS = 100_000;


  @Benchmark
  public void directLoad(Blackhole blackhole) {
	  // given
	  Path path =
			  Paths.get(
					  Objects.requireNonNull(getClass().getClassLoader().getResource("stargate-config.yaml"))
							  .getPath());
	  ConfigStoreYaml configStoreYaml = new ConfigStoreYaml(path);

	  for(int i = 0; i < NUMBER_OF_ITERATIONS; i++) {


		// when
		ConfigWithOverrides configForModule1 = configStoreYaml.getConfigForModule("extension-1");
		ConfigWithOverrides configForModule2 = configStoreYaml.getConfigForModule("extension-2");

		// then
		blackhole.consume(configForModule1);
		blackhole.consume(configForModule2);
	}
  }

	@Benchmark
	public void loadWithCache(Blackhole blackhole){
		// given
		Path path =
				Paths.get(
						Objects.requireNonNull(getClass().getClassLoader().getResource("stargate-config.yaml"))
								.getPath());
		ConfigStoreYaml configStoreYaml = new ConfigStoreYaml(path);

		for(int i = 0; i < NUMBER_OF_ITERATIONS; i++) {

			// when
			ConfigWithOverrides configForModule1 = configStoreYaml.getConfigForModuleWithCache("extension-1");
			ConfigWithOverrides configForModule2 = configStoreYaml.getConfigForModuleWithCache("extension-2");

			// then
			blackhole.consume(configForModule1);
			blackhole.consume(configForModule2);
		}
	}

}
```
The results show that there is a huge difference between both approaches:
```
Benchmark                               Mode  Cnt      Score     Error  Units
[...]

PR #383, opened 2020-11-06
Support enabling/disabling CDC within stargate

tomekl007:


persistence-api/src/main/java/io/stargate/db/cdc/config/CDCConfigLoader.java
```diff
@@ -0,0 +1,108 @@
...
+      @Override
+      public boolean isTrackedByCDC(Table table) {
+        // we need to support runtime reload, therefore every call needs to invoke the
+        // getConfigForModule() method
+        ConfigWithOverrides configForModule =
+            configStore.getConfigForModule(CDC_CORE_CONFIG_MODULE_NAME);
+        return configForModule.getSettingValueList(ENABLED_TABLES_SETTINGS_NAME, String.class)
+            .stream()
+            .map(CDCConfigLoader::mapToKeyspaceTable)
+            .anyMatch(v -> isKeyspaceTableTracked(v, table));
```
dimas-b, line 47: I suspect this may be too slow for production (it's called for every mutation). IIRC, java streams showed up as bottleneck in graph a few times. Would it be possible to construct a hash map on config load? I believe we should be able to use (immutable) `Table` objects as keys (I assume the keyspace is part of its identity). WDYT?

tomekl007: We need to execute this call:
`configForModule.getSettingValueList(ENABLED_TABLES_SETTINGS_NAME, String.class)
.stream()
.map(CDCConfigLoader::mapToKeyspaceTable)`
on every operation to get the runtime change (we can change it to listener API in future if it will turn out to be problematic.
Regarding:
`.anyMatch(v -> isKeyspaceTableTracked(v, table));`
this check:
```
return cdcEnabledKeyspaceTable.table.equals(table.name())
        && cdcEnabledKeyspaceTable.keyspace.equals(table.keyspace());
```
will be executed N times, where N is equal to the number of elements returned by the:
`configForModule.getSettingValueList(ENABLED_TABLES_SETTINGS_NAME, String.class)`
In the real-live, N will be very low (I suspect that less than 10), so I don't see how introducing a hash map could improve performance.
Regarding using streams, I can convert it into normal for loop.
I can create 3 JMH benchmarks:
- current solution
- current solution but with for loop
- using a hash map where Table is a key.
And see if adding a hash map would give us any benefit here.

tomekl007: I've implemented 3 solutions and run benchmarks for them:
1. The current one (the same as the current code on this branch)
2. Using for-loop instead of streams:
```
	ConfigWithOverrides configForModule =
ConfigWithOverrides configForModule =
		configStore.getConfigForModule(CDC_CORE_CONFIG_MODULE_NAME);
List<String> settingValueList =
		configForModule.getSettingValueList(ENABLED_TABLES_SETTINGS_NAME, String.class);
for (String settingValue : settingValueList) {
	KeyspaceTable keyspaceTable = mapToKeyspaceTable(settingValue);
	boolean keyspaceTableTracked = isKeyspaceTableTracked(keyspaceTable, table);
	if (keyspaceTableTracked) {
		return true;
	}
}
return false;
```
3. Constructing hash-map:
```
ConfigWithOverrides configForModule =
    configStore.getConfigForModule(CDC_CORE_CONFIG_MODULE_NAME);
HashMap<Table, Boolean> enabledPerTable = new HashMap<>();

List<String> settingValueList =
    configForModule.getSettingValueList(ENABLED_TABLES_SETTINGS_NAME, String.class);
for (String settingValue : settingValueList) {
  KeyspaceTable keyspaceTable = mapToKeyspaceTable(settingValue);
  enabledPerTable.put(
      ImmutableTable.builder()
          .keyspace(keyspaceTable.keyspace)
          .name(keyspaceTable.table)
          .build(),
      true);
}
Boolean exists = enabledPerTable.get(table);
if (exists == null) {
  return false;
} else {
  return exists;
}
```
The JMH benchmark:
```
@Fork(1)
@Warmup(iterations = 2)
@Measurement(iterations = 10)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.MILLISECONDS)
public class ConfigLoaderPerformanceBenchmark {
  private static final int ITERATIONS_NUMBER = 1;
  // 10 tables to check
  private static final List<String> TABLES_TO_CHECK =
      Arrays.asList(
          "ks.1",
          "ks1.table_a",
          "non.existing",
          "something.b",
          "aa.bb",
          "ks.ttt",
          "ks.1",
          "ks1.table1",
          "non.existing",
          "something.b");

  @Benchmark
  public void baseline(Blackhole blackhole) {
    // given
    ConfigStoreYaml configStoreYaml = new ConfigStoreYaml(getConfigPath(), new MetricRegistry());
    for (String tableToCheck : TABLES_TO_CHECK) {
      // when
      CDCConfig cdcConfig = CDCConfigLoader.loadConfig(configStoreYaml);

      // then
      String[] split = tableToCheck.split("\\.");
      blackhole.consume(cdcConfig.isTrackedByCDC(mockTableWithKeyspace(split[0], split[1])));
    }
  }

  @Benchmark
  public void forLoop(Blackhole blackhole) {
    ConfigStoreYaml configStoreYaml = new ConfigStoreYaml(getConfigPath(), new MetricRegistry());
    for (String tableToCheck : TABLES_TO_CHECK) {
      // given

      // when
      CDCConfig cdcConfig = CDCConfigLoader.loadConfigForLoop(configStoreYaml);

      // then
      String[] split = tableToCheck.split("\\.");
      blackhole.consume(cdcConfig.isTrackedByCDC(mockTableWithKeyspace(split[0], split[1])));
    }
  }

  @Benchmark
  public void hashMap(Blackhole blackhole) {
    // given
    ConfigStoreYaml configStoreYaml = new ConfigStoreYaml(getConfigPath(), new MetricRegistry());
    for (String tableToCheck : TABLES_TO_CHECK) {
      // when
      CDCConfig cdcConfig = CDCConfigLoader.loadConfigHashMap(configStoreYaml);

      // then
      String[] split = tableToCheck.split("\\.");
      blackhole.consume(cdcConfig.isTrackedByCDC(mockTableWithKeyspace(split[0], split[1])));
    }
  }

  @NotNull
  private Path getConfigPath() {
    return Paths.get(
        Objects.requireNonNull(getClass().getClassLoader().getResource("stargate-config.yaml"))
            .getPath());
  }

  private Table mockTableWithKeyspace(String ks, String table) {
    Table tableMock = mock(Table.class);
    when(tableMock.name()).thenReturn(table);
    when(tableMock.keyspace()).thenReturn(ks);
    return tableMock;
  }
}
```
Results are the following:
```
ConfigLoaderPerformanceBenchmark.baseline  avgt   10  1.112 ± 0.068  ms/op
ConfigLoaderPerformanceBenchmark.forLoop   avgt   10  1.225 ± 0.117  ms/op
ConfigLoaderPerformanceBenchmark.hashMap   avgt   10  0.531 ± 0.054  ms/op
```
The hash-map solution is twice faster than other solutions, but it is a difference of 0.5 milliseconds per operation (very negligible). The `forLoop` is not faster than the baseline. 
I could implement the `hashMap` solution, but unfortunately the `Table` `hashCode` and `equals` used `columns` and `indexes`  as well:
```
private boolean equalTo(ImmutableTable another) {
    if (hashCode != another.hashCode) return false;
    return name.equals(another.name)
        && keyspace.equals(another.keyspace)
        && columns.equals(another.columns)
        && indexes.equals(another.indexes);
  }
```
and we don't have that information when constructing a new `Table` from the CDC config option.
We can create and intermediate object:
```
ConfigWithOverrides configForModule =
    configStore.getConfigForModule(CDC_CORE_CONFIG_MODULE_NAME);
HashMap<KeyspaceTable, Boolean> enabledPerTable = new HashMap<>();

List<String> settingValueList =
    configForModule.getSettingValueList(ENABLED_TABLES_SETTINGS_NAME, String.class);
for (String settingValue : settingValueList) {
  KeyspaceTable keyspaceTable = mapToKeyspaceTable(settingValue);
  enabledPerTable.put(keyspaceTable, true);
}
Boolean exists = enabledPerTable.get(mapToKeyspaceTable(table));
if (exists == null) {
  return false;
} else {
  return exists;
}
```
but it performs a bit slower than the original version:
Benchmark                                 Mode  Cnt  Score   Error  Units
ConfigLoaderPerformanceBenchmark.hashMap  avgt   10  0.656 ± 0.094  ms/

it is still faster than `forLoop` and stream, but we will be creating a hashMap for every query, so we will have higher memory consumption. I am not sure if it is worth the effort, @dimas-b let me know wdyt.

dimas-b: Right, I suppose if the size of config data is small, the impacts of java streams is negligible.

However, what about the call to `configForModule.getSettingValueList`? IIRC, it may take tens of milliseconds on average... do I recall correctly? If so, that might be too much overhead for each mutation :thinking:

tomekl007: No, because we implemented caching:
https://github.com/stargate/stargate/issues/340

jorgebay: Isn't it possible to use a config listener that gets invoked when config changes?

I agree with @dimas-b that memoizing a map will future proof it, we could use immutable objects on volatile fields.

It doesn't have to be implemented now, as we expect few keyspaces and tables but I think it should be implemented that way. Should we file a ticket?

jorgebay (commented):
Late to the party as this was already merged but lgtm.

I think that we should use some kind of memoization to avoid iterating the `enabledTables` each time, maybe that can be tackled in a different ticket.

tomekl007:
> Late to the party as this was already merged but lgtm.
> 
> I think that we should use some kind of memoization to avoid iterating the `enabledTables` each time, maybe that can be tackled in a different ticket.

Instead of memoization, we can improve it more once this: https://github.com/stargate/stargate/issues/303 will be done, and we will have a listener API for subscribing to changes in the config-store.

persistence-api/src/main/java/io/stargate/db/cdc/config/CDCConfigLoader.java
dimas-b: Re: #340, the main comment there states about 30ms per `loadWithCache`... Do I read it correctly?

dimas-b: +1 to filing a follow-up ticket for perf. improvement.

tomekl007: Yes, the method that is used here: `getConfigForModule()` uses cache already underneath.

PR #385, opened 2020-11-06
Adding JWT Authz - Schema 

dougwettlaufer:


authnz/src/main/java/io/stargate/auth/AuthorizationService.java
```diff
@@ -22,17 +22,69 @@
...
+   * @param action A {@link QueryBuilder} object to be executed and authorized against a token.
+   * @param token The authenticated token to use for authorization.
+   * @param keyspace Either the keyspace containing the resource to be read or the actual resource
+   *     being read.
+   * @param table The table within the provided keyspace that is being read.
+   * @return On success will return the result of the query and otherwise will return an exception
+   *     relating to the failure to authorize.
+   * @throws Exception An exception relating to the failure to authorize.
+   */
   ResultSet authorizedSchemaRead(
```
dimas-b, line 72: What's the difference between "data" read and "schema" read?

authnz/src/main/java/io/stargate/auth/AuthorizationService.java
```diff
@@ -22,17 +22,69 @@
...
+   * @param token The authenticated token to use for authorization.
+   * @param primaryKeyValues A list of primary key values that will be used in the query and should
+   *     be authorized against the token.
+   * @param tableMetadata The {@link Table} that will be queried against.
+   * @return On success will return the result of the query and otherwise will return an exception
+   *     relating to the failure to authorize.
+   * @throws Exception An exception relating to the failure to authorize.
+   */
   ResultSet authorizedDataRead(
       Callable<ResultSet> action, String token, List<String> primaryKeyValues, Table tableMetadata)
```
dimas-b, line 39: may be late at this point, but shouldn't `primaryKeyValues` be a `List<Object>`? Do we always encode all types as `String`?

authnz/src/main/java/io/stargate/auth/AuthorizationService.java
dougwettlaufer: A "data" read is going to be a `SELECT ...` while a "schema" read will be like a `DESCRIBE ...`. For now "data" reads can be authorized using claims in the JWT while "schema" reads will be authorized by the role and permissions set up in C*. But this does open up to the possibility for more fine grained IAM using something like [OPA](https://www.openpolicyagent.org/) or Redhat Keycloak.

authnz/src/main/java/io/stargate/auth/AuthorizationService.java
dougwettlaufer: Realized I had to go a completely different direction with that in https://github.com/stargate/stargate/pull/386.

dimas-b: :thinking: do we still want to merge this PR, then?

authnz/src/main/java/io/stargate/auth/TypedKeyValue.java
```diff
@@ -0,0 +1,64 @@
+package io.stargate.auth;
+
+import io.stargate.db.datastore.query.Value;
+import io.stargate.db.datastore.query.WhereCondition;
+
+public class TypedKeyValue {
+
+  String name;
+  String type;
+  Object value;
```
dimas-b, line 10: `private`?

dimas-b: Also, are setters used? Could these fields also be `final`?

dougwettlaufer: Right, forgot to circle back on the `private`, thanks. Probably not so I can make them `final`.

dimas-b (approved):

authnz/src/main/java/io/stargate/auth/AuthorizationService.java
```diff
@@ -54,20 +54,16 @@ ResultSet authorizedDataWrite(
...
-   * @return On success will return the result of the query and otherwise will return an exception
-   *     relating to the failure to authorize.
+   * @param keyspaceNames Either the keyspace(s) containing the resource(s) to be read or the actual
+   *     resource being read.
+   * @param tableNames The table(s) within the provided keyspace(s) that is being read.
    * @throws Exception An exception relating to the failure to authorize.
    */
-  ResultSet authorizedSchemaRead(
-      Callable<ResultSet> action, String token, String keyspace, String table) throws Exception;
+  void authorizedSchemaRead(String token, List<String> keyspaceNames, List<String> tableNames)
```
dimas-b, line 65: nit: Since the read is now going to happen after this method returns, how about renaming `authorizedSchemaRead` -> `authorizeSchemaRead` (drop 'd')? (just to emphasize that we authorizing here, but not actually reading)

dimas-b (approved):

PR #389, opened 2020-11-10
GraphQL: Use enum for filter operators

### LINKED ISSUE #307: GraphQL: extract 'FilterOperator' enum
Filter operators are referenced in various parts of the code. Extract an enum for a cleaner design. See [this comment](https://github.com/stargate/stargate/pull/286#discussion_r506772034).

Tentative API:
```java
enum FilterOperator {
  EQ, NOT_EQ, GT, ...

  // Called from FieldFilterInputTypeCache
  // isMap is necessary because CONTAINS extracts a different subtype for a list or a map. Maybe we can use a
  // new FilterType enum instead.
  GraphQLInputObjectField buildFilterField(GraphQLInputType columnType, boolean isMap) { ... }

  // Called from DmlFetcher
  Relation buildRelation(Column column, Object graphqlValue) { ... }
}
```

jorgebay:
Resolves #307 .


EricBorczuk (approved):

dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/DmlFetcher.java
```diff
@@ -43,36 +43,18 @@ protected DmlFetcher(
...
+      for (Map.Entry<String, Object> conditionMap : clauseEntry.getValue().entrySet()) {
+        FilterOperator operator = FilterOperator.fromFieldName(conditionMap.getKey());
+        Condition condition;
+        if (operator != FilterOperator.IN) {
+          condition =
+              operator.buildCondition(column.name(), toCqlTerm(column, conditionMap.getValue()));
+        } else {
+          condition =
+              operator.buildCondition(
+                  column.name(), buildListLiterals(column, conditionMap.getValue()));
```
olim7t, line 55: Have you considered pushing more logic into the enum? Currently there are two `buildCondition` variant, but each constant implements exactly one of them, it's up to the caller to build the right-hand term and call the right method.

We could have single enum method:
```java
Condition buildCondition(String columnName, Object value)
```
And then each operator parses the value into terms according to what it expects (essentially moving `toCqlTerm` and `buildListLiterals` as internal utility methods in the enum).

This would allow us to write this block without any special cases:
```java
      for (Map.Entry<String, Object> conditionMap : clauseEntry.getValue().entrySet()) {
        FilterOperator operator = FilterOperator.fromFieldName(conditionMap.getKey());
        clause.add(operator.buildCondition(column.name(), conditionMap.getValue()));
      }
```

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/DmlFetcher.java
```diff
@@ -113,31 +98,9 @@ protected DmlFetcher(
...
-            case "lt":
-              relation = relationStart.isLessThan(rightTerm);
-              break;
-            case "lte":
-              relation = relationStart.isLessThanOrEqualTo(rightTerm);
-              break;
-            default:
-              throw new IllegalStateException("Unsupported relation type " + condition.getKey());
-          }
+          relation = operator.buildRelation(column.name(), toCqlTerm(column, condition.getValue()));
```
olim7t, line 103: Same remark as `buildCondition`. It would be nice to have all the details in the enum, and treat every operator the same here (the way it's already done for "simple" ones):
```java
      for (Map.Entry<String, Object> condition : clauseEntry.getValue().entrySet()) {
        FilterOperator operator = FilterOperator.fromFieldName(condition.getKey());
        relations.add(operator.buildRelation(column, condition.getValue()));
      }
```

graphqlapi/src/main/java/io/stargate/graphql/schema/FilterOperator.java
```diff
@@ -0,0 +1,217 @@
...
+    @Override
+    public Relation buildRelation(ColumnRelationBuilder<Relation> relationStart, Term value) {
+      throw new IllegalStateException("CONTAINS ENTRY requires key and value terms");
+    }
+
+    @Override
+    public Relation buildRelation(String columnName, Term key, Term value) {
+      return Relation.mapValue(columnName, key).isEqualTo(value);
+    }
+  };
```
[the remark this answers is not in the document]
jorgebay: Sure, makes sense.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/DmlFetcher.java
jorgebay: Yeah, initially I was going in that direction but passing `nameMapping` reference put me off.

I'll give it another try to see how that goes.

jorgebay: I've moved the logic to the enum, I'm happy with the result.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/FilterOperator.java
```diff
@@ -0,0 +1,268 @@
...
+      ConditionBuilder<Condition> conditionBuilder, Term value);
+
+  public Relation buildRelation(Column column, Object value, NameMapping nameMapping) {
+    return buildRelation(
+        Relation.column(CqlIdentifier.fromInternal(column.name())),
+        DataTypeMapping.toCqlTerm(column.type(), value, nameMapping));
+  }
+
+  protected abstract Relation buildRelation(
+      ColumnRelationBuilder<Relation> relationStart, Term value);
```
olim7t, line 230: Nit: that abstract method doesn't feel right, only some of the children use it, yet the others still have to implement it and throw. It makes the code a bit harder to follow.

I think a static utility method would be a better way to factor the common parts of simple operators:
```java
  EQUAL("eq") {
     ...
    @Override
    public Relation buildRelation(Column column, Object value, NameMapping nameMapping) {
      return buildSimpleRelation(column, value, nameMapping, ArithmeticRelationBuilder::isEqualTo);
    }
  },

  private static Relation buildSimpleRelation(
      Column column,
      Object value,
      NameMapping nameMapping,
      BiFunction<ColumnRelationBuilder<Relation>, Term, Relation> operator) {
    return operator.apply(
        Relation.column(CqlIdentifier.fromInternal(column.name())),
        DataTypeMapping.toCqlTerm(column.type(), value, nameMapping));
  }
```

A similar thing can probably be done for conditions too, since they use the same approach.

olim7t (approved):
LGTM, one last comment but it's more a matter of style so feel free to apply it or not.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/FilterOperator.java
jorgebay: Good point, I've refactored it as suggested.

PR #395, opened 2020-11-11
When supplying a where clause to the "full document" search, the path should be matched exactly

EricBorczuk:
Sourav pointed out a bug that he found - when there were two documents, one that had a nested value and one not, such as:
`{ "some": { "nested": {"value": "a"}}` and `{"value": "a"}`

The search `where={"value": {"$eq": "a"}}}` was matching BOTH of these documents, which was not the intention ("recursive" searching was removed a while ago). Also, even more wrong - searching `where={"some.*.value": {"$eq": "a"}}}` was _also_ matching both documents (definitely a bug)

So This PR fixes both those cases

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
@@ -1183,14 +1183,13 @@ public class DocumentService {
                       .filter(
                           r -> {
                             if (r == null || r.getString("leaf") == null) {
-                              System.out.println("Returning false");
                               return false;
                             }
                             String[] parentPath = getParentPathFromRow(r).split("/");
-                            if (parentPath.length != 2) {
-                              return false;
+                            String rowPath = "";
+                            if (parentPath.length == 2) {
+                              rowPath = parentPath[1];
                             }
-                            String rowPath = parentPath[1];
                             return pathsMatch(rowPath + r.getString("leaf"), filterFieldPath);
                           })
                       .findFirst();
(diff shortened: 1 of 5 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -1102,6 +1102,23 @@ private String getParentPathFromRow(Row row) {
     return s.toString();
   }
 
+  private boolean pathsMatch(String path1, String path2) {
+    String[] parts1 = path1.split("\\.");
```
tomekl007, line 1106: If this is on the hot-path, maybe it will be more efficient to have the compiler Pattern and reused it here?

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -1102,6 +1102,23 @@ private String getParentPathFromRow(Row row) {
     return s.toString();
   }
 
+  private boolean pathsMatch(String path1, String path2) {
```
tomekl007, line 1105: If both paths will be empty it will return true. Is it expected?

dougwettlaufer (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: Yea that's fine; in practice both paths should never really be empty but that's reasonable behavior in that case

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: I made a Pattern for this and actually changed a whole bunch of `split` calls to use it instead :) thanks for pointing this out!

PR #399, opened 2020-11-12
Register producer extension based on config

tomekl007:


producer-kafka/src/test/resources/stargate-config-disabled.yaml
```diff
@@ -0,0 +1,6 @@
+cdc.kafka:
+  enabled: false
+  metrics.enabled: true
+  producer.bootstrap.servers: http://kafka-server
+  producer.schema.registry.url: http://schema-registry-url
+  topic.prefix-name: prefix
```
dougwettlaufer, line 6: Does this work with objects like Spring?

```yaml
  producer:
    bootstrap.servers: http://kafka-server
    schema.registry.url: http://schema-registry-url
  topic.prefix-name: prefix
```

dougwettlaufer (approved):

PR #401, opened 2020-11-13
Validate that the createTableSchemaAsync is invoked only if the table's schema differs.

tomekl007:


persistence-api/src/test/java/io/stargate/db/cdc/SchemaAwareCDCProducerTest.java
```diff
@@ -0,0 +1,179 @@
...
+    // when
+    producer.publish(createMutationEvent(table));
+
+    // then
+    assertThat(producer.createNumberOfCalls).isEqualTo(1);
+    assertThat(producer.sendNumberOfCalls).isEqualTo(1);
+  }
+
+  @Test
+  public void shouldCallCreateTableSchemaOnceForSameTableCalls() {
```
dimas-b, line 51: nit: It looks like this test completely covers `shouldUpdateSchemaWhenPublishNewTable`... do we need both of them?

dimas-b (approved):

dougwettlaufer (approved):

persistence-api/src/test/java/io/stargate/db/cdc/SchemaAwareCDCProducerTest.java
tomekl007: It does but I think those are different functionalities. For the potential reader, it would be beneficial to have an additional test covering only this first happy-path

PR #404, opened 2020-11-13
Add @ManagedAsync to document API endpoints

EricBorczuk:
Currently running benchmarks off of this in Fallout, and it seems like it will speed things up considerably with concurrent loads :)

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
@@ -283,7 +283,7 @@ public class DocumentService {
       String leaf = null;
       for (int i = 0; i < fieldNames.length; i++) {
         String fieldName = fieldNames[i];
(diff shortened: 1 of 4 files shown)


dougwettlaufer (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -283,7 +283,7 @@ private boolean isEmptyArray(Object v) {
       String leaf = null;
       for (int i = 0; i < fieldNames.length; i++) {
         String fieldName = fieldNames[i];
-        boolean isArrayElement = fieldName.startsWith("[");
+        boolean isArrayElement = fieldName.startsWith("[") && fieldName.endsWith("]");
```
EricBorczuk, line 286: This fixes just a bunch of Bad Request-type errors due to randomized input from YCSB (they actually end up slowing down throughput when they fail :( )

PR #405, opened 2020-11-13
Migrate GraphQL web stack to DropWizard

### LINKED ISSUE #131: Upgrade graphql-java-servlet dependency once it supports async
Currently the GraphQL extension executes queries in a blocking manner. This is a limitation of the graphql-java-servlet library.

I started apull requestto add full async support. Once it gets merged, it should just be a matter of upgrading the dependency: our data fetchers are already async, so it's the last missing piece.

### LINKED ISSUE #400: GraphQL: Switch the web stack to DropWizard
Drop the dependency to graphql-java-servlet, and instead implement the `/graphql-schema` and `/graphql/{ks}` URIs as DropWizard REST resources.

Advantages:
* use the same stack as other web bundles
* get metrics out of the box
* this will allow us to implement async, and close #131.

olim7t:
Fixes #131
Fixes #400


graphqlapi/src/main/java/io/stargate/graphql/GraphqlActivator.java
```diff
@@ -42,7 +43,7 @@
   private ServicePointer<Metrics> metrics = ServicePointer.create(Metrics.class);
 
   @GuardedBy("this")
-  private WebImpl web;
+  private Server server;
```
olim7t, line 46: All that was left of `WebImpl` was just an empty shell around `Server`, I removed it.

graphqlapi/src/main/java/io/stargate/graphql/web/HttpAwareContext.java
```diff
@@ -48,70 +36,14 @@
...
   }
 
   public String getAuthToken() {
     return request.getHeader("X-Cassandra-Token");
   }
 
-  public String getUserOrRole() {
-    if (hasJWTToken()) {
-      String authHeader = request.getHeader(HEADER);
-      String jwt = authHeader.substring(authHeader.indexOf(PREFIX) + PREFIX.length());
```
olim7t, line 78: This method was dead code before my changes already.

graphqlapi/src/main/java/io/stargate/graphql/web/HttpAwareContext.java
```diff
@@ -48,70 +36,14 @@
...
-      HttpServletResponse response) {
-    this.dataLoaderRegistry = dataLoaderRegistry;
+  public HttpAwareContext(HttpServletRequest request) {
     this.request = request;
-    this.response = response;
-  }
-
-  // Web socket
-  public HTTPAwareContextImpl(
-      DataLoaderRegistry dataLoaderRegistry, Session session, HandshakeRequest handshakeRequest) {
```
olim7t, line 65: I'm not sure if/how web socket worked, it was definitely not covered by the tests.

It's the only thing we lose from moving away from graphql-servlet, I can look at re-adding it if it's something we want to pursue.

graphqlapi/src/main/resources/config.yaml
```diff
@@ -0,0 +1,7 @@
+server:
+  applicationConnectors:
+    - type: http
+      port: 8080
+  adminConnectors:
+    - type: http
+      port: 8091
```
olim7t, line 7: The convention in other bundles is to use applicationPort + 1, but 8081 is already taken by the auth API.

I wish we would simply remove those admin connectors, they serve no purpose because health-checker is supposed to centralize those features. But if we do it introduces a bug because of this line in `Starter.java`:
```java
System.setProperty("dw.server.adminConnectors[0].bindHost", listenHostStr);
```

testing/src/test/java/io/stargate/it/http/GraphqlTest.java
```diff
@@ -1050,8 +1050,6 @@ public void invalidTypeMappingReturnsErrorResponse() {
   private Map<String, Object> executePost(String path, String query) throws IOException {
     OkHttpClient okHttpClient = getHttpClient();
     String url = String.format("http://%s:8080%s", stargate.seedAddress(), path);
-    HttpUrl.Builder httpBuilder = HttpUrl.parse(url).newBuilder();
-    httpBuilder.addQueryParameter("query", query);
```
olim7t, line 1054: See my comment in `GraphqlResourceBase.postJson`.

graphql-java-servlet was apparently more lenient when the query is both in the URI and the request body, but I think it's better to error out.

graphqlapi/src/main/java/io/stargate/graphql/web/ApplicationConfiguration.java
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.graphql.web;
+
+import io.dropwizard.Configuration;
+
+public class ApplicationConfiguration extends Configuration {}
```
tomekl007, line 20: If this is empty, maybe we can create an Application starer with the default `Configuration` instead of creating this class? Or maybe add an explanation of why this is empty.

graphqlapi/src/main/java/io/stargate/graphql/web/Server.java
```diff
@@ -0,0 +1,130 @@
...
+    bootstrap.setMetricRegistry(metrics.getRegistry("graphqlapi"));
+  }
+
+  private void enableCors(Environment environment) {
+    FilterRegistration.Dynamic filter =
+        environment.servlets().addFilter("cors", CrossOriginFilter.class);
+
+    filter.setInitParameter(
+        CrossOriginFilter.ALLOWED_METHODS_PARAM, "POST,GET,OPTIONS,PUT,DELETE,PATCH");
+    filter.setInitParameter(CrossOriginFilter.ALLOWED_ORIGINS_PARAM, "*");
```
tomekl007, line 115: is it safe to allow all?

graphqlapi/src/main/java/io/stargate/graphql/web/Server.java
```diff
@@ -0,0 +1,130 @@
...
+import io.stargate.graphql.web.resources.PlaygroundResource;
+import java.util.EnumSet;
+import javax.servlet.DispatcherType;
+import javax.servlet.FilterRegistration;
+import org.eclipse.jetty.servlets.CrossOriginFilter;
+import org.glassfish.hk2.utilities.binding.AbstractBinder;
+import org.osgi.framework.Bundle;
+import org.osgi.framework.FrameworkUtil;
+
+public class Server extends Application<ApplicationConfiguration> {
```
tomekl007, line 39: nit: maybe name it for example `DropwizardServer` to prevent name clash between this class and `org.eclipse.jetty.server.Server`?

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlDdlResource.java
```diff
@@ -0,0 +1,76 @@
...
+public class GraphqlDdlResource extends GraphqlResourceBase {
+
+  private final GraphQL graphql;
+
+  @Inject
+  public GraphqlDdlResource(GraphqlCache graphqlCache) {
+    this.graphql = graphqlCache.getDdl();
+  }
+
+  @GET
```
tomekl007, line 45: maybe add: `@Produces(MediaType.APPLICATION_JSON)` or `APPLICATION_GRAPHQL` to have consistent headers with other HTTP methods?

olim7t: I assume you meant `@Consumes`, there's already a `@Produces(MediaType.APPLICATION_JSON)` inherited from the parent class.

I'm not sure if `@Consumes` is relevant for a GET request, since it doesn't have a body.

graphqlapi/src/main/java/io/stargate/graphql/web/ApplicationConfiguration.java
dougwettlaufer: This is empty since we use yaml config. It still needs to be here so we can extend it in io.stargate.graphql.web.Server and set up the bootstrap stuff.

graphqlapi/src/main/java/io/stargate/graphql/web/Server.java
dougwettlaufer: They're different packages, how would they clash? This is also consistent the other modules.

dougwettlaufer: Oh, I see what you mean. The fully qualified down below.

graphqlapi/src/main/java/io/stargate/graphql/web/Server.java
dougwettlaufer: We plan to make this configurable (e.g. provide a single host or list of hosts). I believe there's an issue somewhere for it.

graphqlapi/src/main/java/io/stargate/graphql/web/Server.java
```diff
@@ -0,0 +1,130 @@
...
+              }
+            });
+
+    environment.jersey().register(GraphqlDmlResource.class);
+    environment.jersey().register(GraphqlDdlResource.class);
+    environment.jersey().register(PlaygroundResource.class);
+
+    enableCors(environment);
+
+    environment.lifecycle().addServerLifecycleListener(server -> Server.this.jettyServer = server);
```
dougwettlaufer, line 99: Why aren't we just letting dropwizard handle everything?

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlResourceBase.java
```diff
@@ -0,0 +1,176 @@
...
+      String operationName,
+      String variables,
+      GraphQL graphql,
+      HttpServletRequest httpRequest,
+      AsyncResponse asyncResponse) {
+
+    if (Strings.isNullOrEmpty(query)) {
+      replyWithGraphqlError(
+          400, "You must provide a GraphQL query as a URL parameter", asyncResponse);
+    } else {
```
dougwettlaufer, line 67: If we return right after the `replyWithGraphqlError` above we don't need an else, right?

graphqlapi/src/main/java/io/stargate/graphql/web/Server.java
olim7t: I didn't find any programmatic "stop" method in the DropWizard API (I think because most of the time it's used in a dedicated VM so you just kill it).

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlResourceBase.java
olim7t: Yes, we can remove the else's and have explicit return statements in each branch instead.

I don't really have a preference, so if that's easier to follow I can change it.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlResourceBase.java
```diff
@@ -0,0 +1,176 @@
...
+
+        if (!Strings.isNullOrEmpty(variables)) {
+          @SuppressWarnings("unchecked")
+          Map<String, Object> parsedVariables = OBJECT_MAPPER.readValue(variables, Map.class);
+          input = input.variables(parsedVariables);
+        }
+
+        executeAsync(input.build(), graphql, asyncResponse);
+      } catch (JsonProcessingException e) {
+        replyWithGraphqlError(400, "Could not parse variables: " + e.getMessage(), asyncResponse);
```
dougwettlaufer, line 82: Minor: could you use the enums in javax.ws.rs.core.Response.Status rather than the int HTTP status?

graphqlapi/src/main/java/io/stargate/graphql/web/ApplicationConfiguration.java
olim7t: Yes you're right. I copy-pasted from the REST bundle and didn't put much thought into it, but that class is indeed unnecessary.

graphqlapi/src/main/resources/config.yaml
dougwettlaufer: And we can't get rid of that line in Starter.java?

dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/web/Server.java
dougwettlaufer: Got it.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlResourceBase.java
dougwettlaufer: I picked it up with golang but think it makes sense in other languages too

https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

graphqlapi/src/main/java/io/stargate/graphql/web/Server.java
olim7t: It is indeed very permissive.
It seems hard to come up with a good value though, because it will typically depend on the user's environment. We can't predict where their client applications will be located. Maybe we should make this configurable. @dougwettlaufer any opinion?

dougwettlaufer: Yep, as I mentioned above we'll make all of the CORs settings configurable later in another ticket.

graphqlapi/src/main/java/io/stargate/graphql/web/Server.java
olim7t: Good thing that you mentioned that because I had forgotten to change it.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlResourceBase.java
olim7t: Interesting post. I've simplified a few of them.

tomekl007 (commented):
LGTM, but you may want to decrease the complextiy of 
```
The method 'postJson(GraphqlJsonBody, String, GraphQL, HttpServletRequest, AsyncResponse)' has an NPath complexity of 288, current threshold is 200
```
because it was detected as a potential problem, see:
https://app.codacy.com/gh/stargate/stargate/pullRequest?prid=6507004&bid=21021379

graphqlapi/src/main/resources/config.yaml
olim7t: No, because it still applies to the health-checker module.

PR #407, opened 2020-11-16
Support token authz for CQL

dougwettlaufer:


authnz/src/main/java/io/stargate/auth/AuthenticationService.java
```diff
@@ -15,11 +15,19 @@
...
 public interface AuthenticationService {
 
   String createToken(String key, String secret) throws UnauthorizedException;
 
   String createToken(String key) throws UnauthorizedException;
 
   StoredCredentials validateToken(String token) throws UnauthorizedException;
+
+  SaslNegotiator getSaslNegotiator(
+      AuthenticationService authentication,
```
dimas-b, line 29: Is `authentication` not the same as `this`?

authnz/src/main/java/io/stargate/auth/Credentials.java
```diff
@@ -0,0 +1,20 @@
+package io.stargate.auth;
+
+public class Credentials {
+
+  private final String username;
+  private final String password;
```
dimas-b, line 6: nit: if we want to be strict, it might be worth converting this to `char[]` so that the password could be zeroed-out after use. This is the pattern in the JRE internal security APIs. Doing this is moderately important in Stargate, I guess, since we allow arbitrary OSGi bundles to be plugged in.

authnz/src/main/java/io/stargate/auth/Scope.java
```diff
@@ -0,0 +1,12 @@
+package io.stargate.auth;
+
+public enum Scope {
+  READ,
+  MODIFY, // includes INSERT and UPDATE of data
+  DELETE, // encompasses deleting a row, dropping a table, or deleting a keyspace
```
dimas-b, line 6: Would it be worth splitting "row delete" and "table delete" scopes?

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -64,15 +87,188 @@ public ResultMessage process(
...
+    try {
+      refs = bundleContext.getServiceReferences(AuthorizationService.class.getName(), null);
+    } catch (InvalidSyntaxException e) {
+      logger.error("Failed to get service reference for AuthorizationService", e);
+      return null;
+    }
+
+    if (refs != null) {
+      for (ServiceReference<?> ref : refs) {
+        Object service = bundleContext.getService(ref);
```
dimas-b, line 262: Do we runs this on every CQL execution? I suppose this will case OSGi reference leaks since we never "unget" :thinking:

dimas-b (changes requested):

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/interceptors/AuthorizationQueryInterceptor.java
```diff
@@ -0,0 +1,32 @@
...
+  public void initialize() {}
+
+  @Override
+  public ResultMessage interceptQuery(
+      CQLStatement statement,
+      QueryState state,
+      QueryOptions options,
+      Map<String, ByteBuffer> customPayload,
+      long queryStartNanoTime) {
+    return null;
```
dimas-b, line 27: Is this interceptor used?

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -64,15 +87,188 @@ public ResultMessage process(
...
-        ? QueryProcessor.instance.processStatement(
-            statement, queryState, options, queryStartNanoTime)
-        : result;
+
+    if (result != null) {
+      return result;
+    }
+
+    if (customPayload != null && customPayload.containsKey("token")) {
+      authorizeByToken(customPayload.get("token"), statement);
```
dimas-b, line 102: Could re-parsing the token on every message be expensive and affect performance?

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -64,15 +87,188 @@ public ResultMessage process(
...
+      logger.debug(
+          "authorized statement of type {} on {}.{}",
+          castStatement.getClass().toString(),
+          keyspaceName,
+          tableName);
+    }
+  }
+
+  private AuthorizationService getAuthorizationServiceFromContext() {
+    BundleContext bundleContext = FrameworkUtil.getBundle(this.getClass()).getBundleContext();
```
dimas-b, line 250: Adding OSGi dependencies this deep in a persistence module code looks a bit awkward to me... Is there a way to set the auth service reference at the handler creation time? I suppose it should be available when the handler is created.

At creation site we might be able to pass the auth reference through a thread-local variable, thus removing OSGi dependencies from this class. WDYT?

auth-table-based-service/src/main/java/io/stargate/auth/table/PlainTextTableBasedTokenSaslNegotiator.java
```diff
@@ -0,0 +1,50 @@
...
+      if (credentials.getPassword().length() > tokenMaxLength) {
+        logger.error("Token was too long ({} characters)", credentials.getPassword().length());
+        return false;
+      }
+
+      storedCredentials = authentication.validateToken(credentials.getPassword());
+      if (storedCredentials == null) {
+        logger.error("Null credentials returned from authentication service");
+        return false;
+      }
```
dimas-b, line 42: Do we need to pass the token into `storedCredentials.setPassword()` is this case too?

authnz/src/main/java/io/stargate/auth/Credentials.java
dimas-b: cf. `javax.security.auth.callback.PasswordCallback`

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dimas-b: Even though, aub-classing `AuthenticatedUser` might cause some extra work in the DSE Persistence, but it should allow us to cache pre-parsed auth data (minimizing per-message work)... WDYT?

authnz/src/main/java/io/stargate/auth/Scope.java
dougwettlaufer: Good point. Probably should have `DROP` and `DELETE`. I was thinking you could differentiate between the two when defining resources in IAM but unless you're going row level it won't really make sense.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dougwettlaufer: Added

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/interceptors/AuthorizationQueryInterceptor.java
dougwettlaufer: It was :). Forgot to remove when I decided to go another direction.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dougwettlaufer: I know, I felt dirty when I did it. The problem is if we try to register the service in the io.stargate.db.cassandra.CassandraPersistenceActivator it'll create a circular dependency for auth-table-based-service.

auth-table-based-service/src/main/java/io/stargate/auth/table/PlainTextTableBasedTokenSaslNegotiator.java
dougwettlaufer: No because a table based token doesn't give us the info we need for further authorization. I'll add a comment.

authnz/src/main/java/io/stargate/auth/AuthenticationService.java
dougwettlaufer: True. I'll remove that param.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dimas-b: We could pass around an `AtomicReference<AuthorizationService>` at init time and set it (from the activator) when the service is available, perhaps?

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -102,4 +142,214 @@ public ResultMessage processBatch(
...
+
+    if (refs != null) {
+      for (ServiceReference<?> ref : refs) {
+        Object service = bundleContext.getService(ref);
+        if (service instanceof AuthorizationService
+            && ref.getProperty("AuthIdentifier") != null
+            && ref.getProperty("AuthIdentifier").equals(System.getProperty("stargate.auth_id"))) {
+          authorizationReference = ref;
+          return (AuthorizationService) service;
+        }
```
dimas-b, line 344: I believe we also need to "unget" when the `if` does not match.

authnz/src/main/java/io/stargate/auth/Credentials.java
dougwettlaufer: Done

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dougwettlaufer: done

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dougwettlaufer: Done

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -177,6 +180,7 @@ protected void initializePersistence(Config config) {
 
     interceptor.initialize();
     stargateHandler().register(interceptor);
+    stargateHandler().setAuthorizationService(this.authorizationService);
```
dimas-b, line 183: nit: I just realized that since we have a direct call path into the handler, we do not really have to use `AtomicReference` as the param type, we could use `AuthorizationService` and the atomic thing will only exist in the handler for concurrent access... just looks nicer I guess :)

dimas-b: I mean we'd call `CassandraPersistence.setAuth...` from the activators and it would pass the value to the handler?

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -102,4 +157,184 @@ public ResultMessage processBatch(
...
+  private void authorizeByToken(ByteBuffer token, CQLStatement statement) {
+    String authToken = StandardCharsets.UTF_8.decode(token).toString();
+
+    if (!getAuthorizationService().isPresent()) {
+      throw new RuntimeException(
+          "Failed to find an io.stargate.auth.AuthorizationService to authorize request");
+    }
+
+    AuthorizationService authorization = getAuthorizationService().get();
+    if (statement instanceof SelectStatement) {
```
dimas-b, line 170: Is this common enough across the 3 persistence implementations to be worth a common / shared method or class?

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
dougwettlaufer: Except this is called before the AuthorizationService is populated. This init runs so that we can register the persistence bundle after which the auth-* bundles register.

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dougwettlaufer: That's about the only thing common between them, 1) decode token, 2) check if authService is present, 3) get the authService. Is four lines of code worth a new class?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -188,11 +190,15 @@ protected void initializePersistence(Config config) {
 
     waitForSchema(STARTUP_DELAY_MS);
 
-    if (USE_PROXY_PROTOCOL) interceptor = new ProxyProtocolQueryInterceptor();
-    else interceptor = new DefaultQueryInterceptor();
+    if (USE_PROXY_PROTOCOL) {
+      interceptor = new ProxyProtocolQueryInterceptor();
+    } else {
+      interceptor = new DefaultQueryInterceptor();
+    }
```
dimas-b, line 197: This will likely conflict with recent changes in another PR from @mpenick

dougwettlaufer: Didn't notice. That was actually the formatter's doing.

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dimas-b: Ah, ok, sorry, I did not compare the code carefully :facepalm:

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
dimas-b: We call `cassandraDB.setAuthorizationService(ref)` from the activator(s)... I guess we could call `cassandraDB.setAuthorizationService(service)` when the service is resolved. This way we will not have to have a shared `AtomicReference` object. I suggested sharing `AtomicReference` before because I was not sure we could have a direct method call chain into the query handler.

authnz/src/main/java/io/stargate/auth/AuthenticationService.java
```diff
@@ -15,11 +15,16 @@
...
 public interface AuthenticationService {
 
   String createToken(String key, String secret) throws UnauthorizedException;
 
   String createToken(String key) throws UnauthorizedException;
 
   StoredCredentials validateToken(String token) throws UnauthorizedException;
+
+  SaslNegotiator getSaslNegotiator(
+      SaslNegotiator wrapped, String tokenUsername, int tokenMaxLength);
```
mpenick, line 29: Why are `tokenUsername` and `tokenMaxLength` passed as arguments?  These parameters are implementation details and should be constants in the specific implementing class.

dougwettlaufer: done

dimas-b (approved):

PR #412, opened 2020-11-17
GraphQL: use full URI in playground tabs

olim7t:


graphqlapi/src/main/resources/playground.html
```diff
@@ -558,24 +558,29 @@
         "x-cassandra-token": (tokenValue == null ? "" : tokenValue.trim()) || "populate_me"
       }
 
+      // We expose the GraphQL resources at the same level as the playground:
+      const playgroundUri = window.location.href
```
olim7t, line 562: We could also use `window.location.pathname` to omit the scheme and authority (e.g. `path/to/graphql-schema`), but in the end I prefer having a full URI, it's easier to copy-paste, for example to a CURL command.

dougwettlaufer (approved):

mpenick (approved):

PR #419, opened 2020-11-17
Support JWT authz in Graphql

dougwettlaufer:


graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/DmlFetcher.java
```diff
@@ -88,6 +99,118 @@ protected DmlFetcher(
...
+
+        if (isOrContainsUDT(Objects.requireNonNull(column.type()))) {
+          // Null out the value for now since UDTs are not allowed for use with custom authorization
+          typedKeyValues.add(
+              new TypedKeyValue(
+                  column.cqlName(), Objects.requireNonNull(column.type()).cqlDefinition(), null));
+          continue;
+        }
+
+        if (term instanceof TupleTerm) {
```
olim7t, line 134: When do we hit that branch? The only place where the driver's query builder creates a `TupleTerm` is `QueryBuilder.tuple()`, and it's not referenced anywhere in this project. `DmlFetcher.toCqlTerm` only creates instances of `RawTerm`.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/AllKeyspacesFetcher.java
```diff
@@ -17,21 +17,46 @@
...
+      @SuppressWarnings("unchecked")
+      List<Map<String, Object>> tables = (List<Map<String, Object>>) keyspace.get("tables");
+
+      List<String> tableNames = null;
+      if (tables != null) {
+        tableNames = tables.stream().map(t -> (String) t.get("name")).collect(Collectors.toList());
+      }
+
+      authorizationService.authorizeSchemaRead(
+          token, Collections.singletonList(keyspaceName), tableNames);
```
olim7t, line 58: I think it would be better to push that logic into `KeyspaceFormatter`: check as we go, instead of parsing everything and checking after the fact. It will also allow code reuse between the single-keyspace and multi-keyspace methods.

olim7t, line 58: What is the expected behavior if the user is not authorized for all keyspaces: fail the whole query, or silently omit the restricted ones?

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/CreateTableFetcher.java
```diff
@@ -23,21 +23,35 @@
...
 
-  public String getQuery(DataFetchingEnvironment dataFetchingEnvironment) {
+  public String getQuery(DataFetchingEnvironment dataFetchingEnvironment)
+      throws UnauthorizedException {
+    String keyspaceName = dataFetchingEnvironment.getArgument("keyspaceName");
+    String tableName = dataFetchingEnvironment.getArgument("tableName");
+
+    HttpAwareContext httpAwareContext = dataFetchingEnvironment.getContext();
+    String token = httpAwareContext.getAuthToken();
+    authorizationService.authorizeSchemaWrite(token, keyspaceName, tableName, Scope.CREATE);
```
olim7t, line 50: These 5 lines are the same for all table operations, only the scope changes. We could extract a `TableFetcher` parent class where `getQuery` performs the check and delegates to an abstract `getQuery(DataFetchingEnvironment, String, String)`.

graphqlapi/src/test/java/io/stargate/graphql/schema/GraphQlTestBase.java
```diff
@@ -67,6 +72,11 @@ public void setupEnvironment() {
       String roleName = "mock role name";
       when(authenticationService.validateToken(token)).thenReturn(storedCredentials);
       when(storedCredentials.getRoleName()).thenReturn(roleName);
+      when(authorizationService.authorizedDataRead(actionCaptor.capture(), eq(token), any()))
+          .then(
+              i -> {
+                return actionCaptor.getValue().call();
+              });
```
olim7t, line 79:
```suggestion
      when(authorizationService.authorizedDataRead(actionCaptor.capture(), eq(token), any()))
          .then(i -> actionCaptor.getValue().call());
```

olim7t (commented):
A few comments on the GraphQL part.
Depending on the ETA for #413, it might make sense to wait because it should make it easier to retrieve values from the queries.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/DmlFetcher.java
dougwettlaufer: I ran into this for "IN" queries like

```sql
SELECT l FROM collections."PkListTable" WHERE l IN ([1,2,3],[4,5,6])
```

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/DmlFetcher.java
```diff
@@ -88,6 +99,118 @@ protected DmlFetcher(
...
+          columnId = ((ColumnComponentLeftOperand) leftOperand).getColumnId();
+          queryByEntries = true;
+        } else {
+          columnId = ((ColumnLeftOperand) leftOperand).getColumnId();
+        }
+
+        Column column = getColumn(table, columnId.asInternal());
+        assert term != null;
+
+        if (isOrContainsUDT(Objects.requireNonNull(column.type()))) {
```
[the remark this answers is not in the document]
dougwettlaufer: Good point, I'll update. And agreed, that warning is annoying.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/AllKeyspacesFetcher.java
dougwettlaufer: Good question. I should probably pull this up so we can just filter out the ones the user isn't authorized for.

authnz/src/main/java/io/stargate/auth/AuthorizationService.java
```diff
@@ -66,20 +71,16 @@ void authorizeDataWrite(String token, String keyspace, String table, Scope scope
...
-   *     relating to the failure to authorize.
-   * @throws Exception An exception relating to the failure to authorize.
+   * @param scope The {@link Scope} of the action to be performed.
+   * @throws UnauthorizedException An exception relating to the failure to authorize.
    */
-  ResultSet authorizedDataWrite(
-      Callable<ResultSet> action, String token, List<TypedKeyValue> typedKeyValues)
-      throws Exception;
+  void authorizedDataWrite(String token, List<TypedKeyValue> typedKeyValues, Scope scope)
+      throws UnauthorizedException;
```
dimas-b, line 83: nit: Since we not longer returning the `ResultSet`, shall we rename to `authorizeDataWrite` (drop the `d`) following the pattern with `authorizeSchemaRead`, etc?

dimas-b:
How does this interact with #407? Graphql API generates CQL... aren't wee going to do atuhz twice?

dougwettlaufer:
> How does this interact with #407? Graphql API generates CQL... aren't wee going to do authz twice?

The token will only be in the custom payload for the CQL path
https://github.com/stargate/stargate/pull/407/files#diff-0d01320b95459a7d0d2a6e36a7d5fb6e4cad3a7019e50e995a426ea24c8e7951R117

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/dml/DmlFetcher.java
olim7t: Ah my bad, that's indeed how IN relations work.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/CreateTableFetcher.java
dougwettlaufer: Made the change. Let me know what you think.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/KeyspaceFormatter.java
```diff
@@ -69,13 +96,30 @@
...
       for (ChildT child : allChildrenGetter.get()) {
+        if (childFieldName.equals("table")) {
+          try {
+            authorizationService.authorizeSchemaRead(
+                token,
+                Collections.singletonList(keyspaceName),
+                Collections.singletonList(((ImmutableTable) child).name()));
+          } catch (UnauthorizedException e) {
+            LOG.info(
+                "Not returning table {}.{} due to not being authorized",
```
olim7t, line 117: Nit: this method is used both for tables and UDTs, so
```suggestion
                "Not returning table/UDT {}.{} due to not being authorized",
```
Although this is just a log, so maybe it doesn't matter.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/KeyspaceFormatter.java
```diff
@@ -18,30 +18,51 @@
...
+      Set<Keyspace> keyspaces,
+      DataFetchingEnvironment environment,
+      AuthorizationService authorizationService,
+      String token) {
     List<Map<String, Object>> list = new ArrayList<>();
     for (Keyspace keyspace : keyspaces) {
-      list.add(formatResult(keyspace, environment));
+      try {
+        authorizationService.authorizeSchemaRead(
+            token, Collections.singletonList(keyspace.name()), null);
```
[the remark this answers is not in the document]
dougwettlaufer: SingleKeyspaceFetcher will still call this method, right?

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/KeyspaceFormatter.java
dougwettlaufer: I'm only checking tables right now with the `if(childFieldName.equals("table"))` up above

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/KeyspaceFormatter.java
```diff
@@ -69,13 +96,30 @@
...
       List<Map<String, Object>> formattedChildren = new ArrayList<>();
       for (ChildT child : allChildrenGetter.get()) {
+        if (childFieldName.equals("table")) {
+          try {
+            authorizationService.authorizeSchemaRead(
+                token,
+                Collections.singletonList(keyspaceName),
+                Collections.singletonList(((ImmutableTable) child).name()));
+          } catch (UnauthorizedException e) {
+            LOG.info(
```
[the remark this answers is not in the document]
dougwettlaufer: Yeah I guess I can change to debug. Since it's just filtering them out I felt like there needed to be some sort of signal.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/KeyspaceFormatter.java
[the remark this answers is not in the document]
dougwettlaufer: Oh, never mind. It goes to the other formatResult.

dimas-b (approved):
LGTM. One naming comment in `AuthorizationService`

authnz/src/main/java/io/stargate/auth/AuthorizationService.java
dougwettlaufer: Sure, will change after I merge into the base branch to avoid waiting on CI.

PR #422, opened 2020-11-18
Call the CDCProducer#init() to initialize the CDC Producer	

tomekl007:


producer-kafka/src/main/java/io/stargate/producer/kafka/KafkaCDCProducer.java
```diff
@@ -32,15 +31,21 @@
...
 import org.apache.cassandra.stargate.db.MutationEvent;
 import org.apache.cassandra.stargate.db.RowUpdateEvent;
 import org.apache.kafka.clients.producer.ProducerRecord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class KafkaCDCProducer extends SchemaAwareCDCProducer {
-
+  private static final Logger LOGGER = LoggerFactory.getLogger(KafkaCDCProducer.class);
+  private static final Executor KAFKA_INIT_EXECUTOR = Executors.newSingleThreadExecutor();
```
dimas-b, line 47: Since this is a one-use thread pool, is it worth keeping it open after `init`? What do you think about using a simple thread that would exit after running the `init` method?

producer-kafka/src/main/java/io/stargate/producer/kafka/KafkaProducerActivator.java
```diff
@@ -52,6 +54,15 @@ protected ServiceAndProperties createService() {
       CDCProducer producer =
           new KafkaCDCProducer(
               metrics.get().getRegistry(KAFKA_CDC_METRICS_PREFIX), configStore.get());
+      try {
+        producer.init().get();
```
dimas-b, line 58: Why do we need to return a `CompletableFuture` from `init()` if we wait for it to complete anyway?

tomekl007: Yes, it would be good enough for this producer, but other producers that implement the interface may want to provide an async init and let the caller decide.

producer-kafka/src/main/java/io/stargate/producer/kafka/KafkaCDCProducer.java
tomekl007: good point, done

dougwettlaufer (approved):

producer-kafka/src/main/java/io/stargate/producer/kafka/KafkaProducerActivator.java
dimas-b: I'm wondering, should we `init` it at activations time or let the CDC service call `init`?

producer-kafka/src/main/java/io/stargate/producer/kafka/KafkaCDCProducer.java
```diff
@@ -61,25 +64,42 @@ public KafkaCDCProducer(MetricRegistry registry, ConfigStore configStore) {
...
+                // However, in our OSGi framework,
+                // the default context class loader is the system class loader,
+                // which does not contain module jars.
+                currentThread.setContextClassLoader(contextClassLoader);
+                return new CompletableKafkaProducer<>(cdcKafkaConfig.getKafkaProducerSettings());
+              } finally {
+                currentThread.setContextClassLoader(ldr);
+              }
+            },
+            Executors.newSingleThreadExecutor());
```
dimas-b, line 102: The executor's thread will keep running after the task is done, will it not?

tomekl007: You are right, it will not stop the executor and I think we shouldn't stop it. The reason for this is that we don't want to block here and let the caller decide.

tomekl007: I've did add shutdown() in the close().

producer-kafka/src/main/java/io/stargate/producer/kafka/KafkaCDCProducer.java
```diff
@@ -149,10 +168,12 @@ private void registerMetrics(MetricRegistry registry) {
...
-          producer.flush();
-          producer.close();
-        });
+    return kafkaProducer
+        .thenAccept(
+            producer -> {
+              producer.flush();
+              producer.close();
+            })
+        .thenAccept(v -> KAFKA_INIT_EXECUTOR.shutdown());
```
dimas-b, line 177: Sorry, for being picky on this, but this thread pool will still apparently keep an idle thread after the `init()` is done, until the whole service is shut down, I guess :thinking: 

Since we always completely initialize the producer before registering it in this case, would it make sense to initialize it on the activator thread (`main`) and return a completed future from `init()`?

tomekl007: As I pointed out in the previous comment:
https://github.com/stargate/stargate/pull/422#discussion_r526184287
We may decide to not block on init, so to fulfill the current API (returning the `CompletableFuture`) we should not block in the init() method. Other CDC producers may choose a different approach for initializing.

producer-kafka/src/main/java/io/stargate/producer/kafka/KafkaProducerActivator.java
tomekl007: That was my initial idea but I moved it to the extension itself because of the classpath problem.

PR #424, opened 2020-11-19
Add ErrorProne to the build

olim7t:
This is based on the configuration we use in the Java driver. It's a bit outdated and we won't be able to use the latest version, but the method described on the ErrorProne website does not seem to work with a Java 8 SDK.

### DIFF
--- a/auth-api/src/main/java/io/stargate/auth/model/Credentials.java
+++ b/auth-api/src/main/java/io/stargate/auth/model/Credentials.java
@@ -23,7 +23,7 @@ import java.util.Objects;
 
 /** Credentials object used to remain compliant with existiing API contract */
 @JsonIgnoreProperties(ignoreUnknown = true)
-public class Credentials {
+public final class Credentials {
   String username;
   String password;
 
--- a/auth-api/src/main/java/io/stargate/auth/model/Credentials.java
+++ b/auth-api/src/main/java/io/stargate/auth/model/Credentials.java
@@ -59,7 +59,7 @@ public class Credentials {
     if (this == o) {
       return true;
     }
-    if (o == null || getClass() != o.getClass()) {
+    if (!(o instanceof Credentials)) {
       return false;
     }
     Credentials that = (Credentials) o;
--- a/auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJWTServiceActivator.java
+++ b/auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJWTServiceActivator.java
@@ -40,6 +40,7 @@ public class AuthJWTServiceActivator implements BundleActivator {
 
   public static final String AUTH_JWT_IDENTIFIER = "AuthJwtService";
 
+  @SuppressWarnings("JdkObsolete")
   private static final Hashtable<String, String> props = new Hashtable<>();
 
   static {
--- a/auth-jwt-service/src/test/java/io/stargate/auth/jwt/AuthJWTServiceActivatorTest.java
+++ b/auth-jwt-service/src/test/java/io/stargate/auth/jwt/AuthJWTServiceActivatorTest.java
@@ -41,6 +41,7 @@ public class AuthJWTServiceActivatorTest {
 
     activator.start(bundleContext);
 
+    @SuppressWarnings("JdkObsolete")
     Hashtable<String, String> expectedProps = new Hashtable<>();
     expectedProps.put("AuthIdentifier", AuthJWTServiceActivator.AUTH_JWT_IDENTIFIER);
 
@@ -65,6 +66,7 @@ public class AuthJWTServiceActivatorTest {
 
     activator.start(bundleContext);
 
+    @SuppressWarnings("JdkObsolete")
     Hashtable<String, String> expectedProps = new Hashtable<>();
     expectedProps.put("AuthIdentifier", AuthJWTServiceActivator.AUTH_JWT_IDENTIFIER);
 
@@ -91,6 +93,7 @@ public class AuthJWTServiceActivatorTest {
         assertThrows(RuntimeException.class, () -> activator.start(bundleContext));
     assertThat(ex).hasMessage("Property `stargate.auth.jwt_provider_url` must be set");
 
+    @SuppressWarnings("JdkObsolete")
     Hashtable<String, String> expectedProps = new Hashtable<>();
     expectedProps.put("AuthIdentifier", AuthJWTServiceActivator.AUTH_JWT_IDENTIFIER);
 
@@ -117,6 +120,7 @@ public class AuthJWTServiceActivatorTest {
         assertThrows(RuntimeException.class, () -> activator.start(bundleContext));
     assertThat(ex).hasMessage("Property `stargate.auth.jwt_provider_url` must be set");
 
+    @SuppressWarnings("JdkObsolete")
     Hashtable<String, String> expectedProps = new Hashtable<>();
     expectedProps.put("AuthIdentifier", AuthJWTServiceActivator.AUTH_JWT_IDENTIFIER);
 
@@ -143,6 +147,7 @@ public class AuthJWTServiceActivatorTest {
         assertThrows(RuntimeException.class, () -> activator.start(bundleContext));
     assertThat(ex).hasMessage("Failed to create JwtValidator: no protocol: foo");
 
+    @SuppressWarnings("JdkObsolete")
     Hashtable<String, String> expectedProps = new Hashtable<>();
     expectedProps.put("AuthIdentifier", AuthJWTServiceActivator.AUTH_JWT_IDENTIFIER);
 
--- a/auth-jwt-service/src/test/java/io/stargate/auth/jwt/AuthnJwtServiceTest.java
+++ b/auth-jwt-service/src/test/java/io/stargate/auth/jwt/AuthnJwtServiceTest.java
@@ -28,7 +28,6 @@ import com.nimbusds.jose.proc.BadJOSEException;
 import com.nimbusds.jose.proc.SecurityContext;
 import com.nimbusds.jose.proc.SimpleSecurityContext;
 import com.nimbusds.jwt.JWTClaimsSet;
-import com.nimbusds.jwt.JWTClaimsSet.Builder;
 import com.nimbusds.jwt.SignedJWT;
 import com.nimbusds.jwt.proc.ConfigurableJWTProcessor;
 import com.nimbusds.jwt.proc.DefaultJWTProcessor;
@@ -92,7 +91,8 @@ public class AuthnJwtServiceTest {
       throws UnauthorizedException, ParseException, JOSEException, BadJOSEException {
     Map<String, Object> stargate_claims = new HashMap<>();
     stargate_claims.put("x-stargate-role", "user");
-    JWTClaimsSet jwtClaimsSet = new Builder().claim("stargate_claims", stargate_claims).build();
+    JWTClaimsSet jwtClaimsSet =
+        new JWTClaimsSet.Builder().claim("stargate_claims", stargate_claims).build();
     when(jwtProcessorMocked.process("token", null)).thenReturn(jwtClaimsSet);
 
     StoredCredentials storedCredentials = mockAuthnJwtService.validateToken("token");
@@ -105,7 +105,7 @@ public class AuthnJwtServiceTest {
   public void validateTokenMissingClaims() throws ParseException, JOSEException, BadJOSEException {
     Map<String, Object> stargate_claims = new HashMap<>();
     stargate_claims.put("x-stargate-role", "user");
-    JWTClaimsSet jwtClaimsSet = new Builder().claim("claims", stargate_claims).build();
+    JWTClaimsSet jwtClaimsSet = new JWTClaimsSet.Builder().claim("claims", stargate_claims).build();
     when(jwtProcessorMocked.process("token", null)).thenReturn(jwtClaimsSet);
 
     UnauthorizedException ex =
@@ -117,7 +117,8 @@ public class AuthnJwtServiceTest {
   public void validateTokenMissingRole() throws ParseException, JOSEException, BadJOSEException {
     Map<String, Object> stargate_claims = new HashMap<>();
     stargate_claims.put("x-stargate-foo", "user");
-    JWTClaimsSet jwtClaimsSet = new Builder().claim("stargate_claims", stargate_claims).build();
+    JWTClaimsSet jwtClaimsSet =
+        new JWTClaimsSet.Builder().claim("stargate_claims", stargate_claims).build();
     when(jwtProcessorMocked.process("token", null)).thenReturn(jwtClaimsSet);
 
     UnauthorizedException ex =
@@ -129,7 +130,8 @@ public class AuthnJwtServiceTest {
   public void validateTokenEmptyRole() throws ParseException, JOSEException, BadJOSEException {
     Map<String, Object> stargate_claims = new HashMap<>();
     stargate_claims.put("x-stargate-role", "");
-    JWTClaimsSet jwtClaimsSet = new Builder().claim("stargate_claims", stargate_claims).build();
+    JWTClaimsSet jwtClaimsSet =
+        new JWTClaimsSet.Builder().claim("stargate_claims", stargate_claims).build();
     when(jwtProcessorMocked.process("token", null)).thenReturn(jwtClaimsSet);
 
     UnauthorizedException ex =
@@ -141,7 +143,8 @@ public class AuthnJwtServiceTest {
   public void validateTokenRoleWrongType() throws ParseException, JOSEException, BadJOSEException {
     Map<String, Object> stargate_claims = new HashMap<>();
     stargate_claims.put("x-stargate-role", 1);
-    JWTClaimsSet jwtClaimsSet = new Builder().claim("stargate_claims", stargate_claims).build();
+    JWTClaimsSet jwtClaimsSet =
+        new JWTClaimsSet.Builder().claim("stargate_claims", stargate_claims).build();
     when(jwtProcessorMocked.process("token", null)).thenReturn(jwtClaimsSet);
 
     UnauthorizedException ex =
--- a/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Frame.java
+++ b/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Frame.java
@@ -191,10 +191,10 @@ public class Frame {
         throw ErrorMessage.wrap(e, streamId);
       }
 
-      long bodyLength = buffer.getUnsignedInt(idx);
+      int bodyLength = buffer.getInt(idx);
       idx += Header.BODY_LENGTH_SIZE;
 
-      long frameLength = bodyLength + Header.LENGTH;
+      long frameLength = (long) bodyLength + Header.LENGTH;
       if (frameLength > MAX_FRAME_LENGTH) {
         // Enter the discard mode and discard everything received so far.
         discardingTooLongFrame = true;
@@ -211,7 +211,7 @@ public class Frame {
       ClientRequestSizeMetrics.bytesRecievedPerFrame.update(frameLength);
 
       // extract body
-      ByteBuf body = buffer.slice(idx, (int) bodyLength);
+      ByteBuf body = buffer.slice(idx, bodyLength);
       body.retain();
 
       idx += bodyLength;
@@ -267,6 +267,7 @@ public class Frame {
 
   @ChannelHandler.Sharable
   public static class Encoder extends MessageToMessageEncoder<Frame> {
+    @Override
     public void encode(ChannelHandlerContext ctx, Frame frame, List<Object> results)
         throws IOException {
       ByteBuf header = CBUtil.allocator.buffer(Header.LENGTH);
@@ -296,6 +297,7 @@ public class Frame {
 
   @ChannelHandler.Sharable
   public static class InboundBodyTransformer extends MessageToMessageDecoder<Frame> {
+    @Override
     public void decode(ChannelHandlerContext ctx, Frame frame, List<Object> results)
         throws IOException {
       Connection connection = ctx.channel().attr(Connection.attributeKey).get();
@@ -324,6 +326,7 @@ public class Frame {
 
   @ChannelHandler.Sharable
   public static class OutboundBodyTransformer extends MessageToMessageEncoder<Frame> {
+    @Override
     public void encode(ChannelHandlerContext ctx, Frame frame, List<Object> results)
         throws IOException {
       Connection connection = ctx.channel().attr(Connection.attributeKey).get();
--- a/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Server.java
+++ b/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Server.java
@@ -282,7 +282,6 @@ public class Server implements CassandraDaemon.Server {
       groups.get(type).add(ch);
     }
 
-    @SuppressWarnings("FutureReturnValueIgnored")
     public void send(Event event) {
       groups.get(event.type).writeAndFlush(new EventMessage(event));
     }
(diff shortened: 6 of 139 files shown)


pom.xml
```diff
@@ -336,9 +336,31 @@
           <artifactId>maven-compiler-plugin</artifactId>
           <version>3.8.1</version>
           <configuration>
+            <compilerId>javac-with-errorprone</compilerId>
+            <forceJavacCompilerUse>true</forceJavacCompilerUse>
             <source>1.8</source>
             <target>1.8</target>
+            <compilerArgs combine.children="override">
+              <compilerArg>-XepAllErrorsAsWarnings</compilerArg>
```
olim7t, line 344: This is temporary to avoid breaking the build, but eventually we should remove it.

pom.xml
```diff
@@ -336,9 +336,31 @@
...
+              <!-- This would require adding ErrorProne annotations in our code: -->
```
pcmanus, line 345: Fwiw, I'd be ok with that. Getting format validation is nice-to-have, and I don't see too many downsides to using the annotations. Totally fine with leaving that to later though.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/PlaygroundResource.java
```diff
@@ -15,6 +15,7 @@
  */
 package io.stargate.graphql.web.resources;
 
+import com.datastax.oss.driver.shaded.guava.common.base.Charsets;
```
pcmanus, line 18: The graphQL modules imports it's own guava, we should probably import that one.

pcmanus (approved):
+1 modulo a minor import remark. Thanks for looking at it.

Maybe worth opening a ticket to track getting rid of `-XepAllErrorsAsWarnings`.

graphqlapi/src/main/java/io/stargate/graphql/web/GraphqlCache.java
```diff
@@ -128,7 +128,7 @@ private static String findDefaultKeyspace(DataStore dataStore) {
   }
 
   private void addOrReplaceDmlGraphql(
-      String keyspaceName, String reason, String... reasonArguments) {
+      String keyspaceName, String reason, Object... reasonArguments) {
```
tomekl007, line 131: Narrowing the type does not seem like a good direction. Why error-prone enforce it?

graphqlapi/src/main/java/io/stargate/graphql/util/ByteBufferUtils.java
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.graphql.util;
+
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+import java.util.Base64;
+
+public class ByteBufferUtils {
```
tomekl007, line 22: I think that we should add a unit test for it.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Frame.java
```diff
@@ -191,10 +191,10 @@ Frame decodeFrame(ByteBuf buffer) throws Exception {
         throw ErrorMessage.wrap(e, streamId);
       }
 
-      long bodyLength = buffer.getUnsignedInt(idx);
+      int bodyLength = buffer.getInt(idx);
```
olim7t, line 194: The protocol spec is not 100% clear on this, but it says "a 4 byte integer", in any other place if integers are unsigned it's said explicitly. And the rest of the decoder wouldn't work anyway if the length was bigger than an int.

olim7t, line 194: The protocol spec is not 100% clear on this, but it says "a 4 byte integer", in any other place if integers are unsigned it's said explicitly. And the rest of the decoder wouldn't work anyway if the length was bigger than an int.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Server.java
```diff
@@ -590,12 +593,20 @@ private void send(InetAddressAndPort endpoint, Event.NodeEvent event) {
       // then don't send the notification. This covers the case of rpc_address set to "localhost",
       // which is not useful to any driver and in fact may cauase serious problems to some drivers,
       // see CASSANDRA-10052
-      if (!endpoint.equals(FBUtilities.getBroadcastAddressAndPort())
+      if (!areEqual(endpoint, FBUtilities.getBroadcastAddressAndPort())
```
olim7t, line 596: That looks like a legitimate bug, the two objects can't be equal because the types are not the same.

pom.xml
```diff
@@ -338,9 +338,30 @@
           <artifactId>maven-compiler-plugin</artifactId>
           <version>3.8.1</version>
           <configuration>
+            <compilerId>javac-with-errorprone</compilerId>
+            <forceJavacCompilerUse>true</forceJavacCompilerUse>
             <source>1.8</source>
             <target>1.8</target>
+            <compilerArgs combine.children="override">
+              <compilerArg>-Xep:MixedMutabilityReturnType:OFF</compilerArg>
```
olim7t, line 346: https://errorprone.info/bugpattern/MixedMutabilityReturnType

It pops up for any method that returns an ArrayList in the general case, but Collections.emptyList() in the empty case. I think we're good, the callers don't try to change the list.

pom.xml
```diff
@@ -338,9 +338,30 @@
...
+              <compilerArg>-Xep:ImmutableEnumChecker:OFF</compilerArg>
```
olim7t, line 347: https://errorprone.info/bugpattern/ImmutableEnumChecker

Alternatively we could annotate every type that's referenced from an enum with `@Immutable`.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Server.java
```diff
@@ -284,6 +282,7 @@ public void register(Event.Type type, Channel ch) {
       groups.get(type).add(ch);
     }
 
+    @SuppressWarnings("FutureReturnValueIgnored")
     public void send(Event event) {
       groups.get(event.type).writeAndFlush(new EventMessage(event));
```
olim7t, line 287: https://errorprone.info/bugpattern/FutureReturnValueIgnored

I'm a bit annoyed with that one. It's a perfectly valid issue, but it often happens in cases like this where the API was not designed to propagate an async error. The method would have to return a future.

olim7t (commented):
Rebased on current master, and I'm going to address all the warnings. I'm committing a first batch of modules now because I want to add review comments as I go.

auth-api/src/main/java/io/stargate/auth/model/Credentials.java
```diff
@@ -59,7 +59,7 @@ public boolean equals(Object o) {
     if (this == o) {
       return true;
     }
-    if (o == null || getClass() != o.getClass()) {
+    if (!(o instanceof Credentials)) {
```
dimas-b, line 62: The new logic admits sub-classes, while the old logic required the instance to be exactly of type `Credentials`

dimas-b: Does ErrorProne require this change? What's the rationale?

olim7t: https://errorprone.info/bugpattern/EqualsGetClass

The bottom line is that there's not really a good way to make equals work with subclasses. But it's kind of a moot point here anyway since none of these classes have subclasses.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/interceptors/DefaultQueryInterceptor.java
```diff
@@ -252,9 +253,12 @@ private void applyState(
       case RPC_READY:
         notifyRpcChange(endpoint, epState.isRpcReady());
         break;
+      default:
+        // ignore other states
```
olim7t, line 257: Can someone confirm that this is the intended behavior?

olim7t (commented):
I'm seeing a few instances of this in the persistence-dse module:
```
[WARNING] /persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java:[342,26]
[RxReturnValueIgnored] Returned Rx objects must be checked.
Ignoring a returned Rx value means it is never scheduled for execution
    (see https://errorprone.info/bugpattern/RxReturnValueIgnored)
```
I assume that if we were scheduling stuff that doesn't get executed we would have noticed. Are these false positives or is there something we should do with the return value to avoid them?

dimas-b:
> Ignoring a returned Rx value means it is never scheduled for execution

This particular way of stating the issue looks like an exaggeration to me. Execution may not be strictly depended on the presence of an observer, specifically when we return `Single` objects during login (cf. `DsePersistence.java:[342,26]`).

However, I believe if we treat this as a warning it would make sense. In many cases subscribing is indeed required for the actions to occur.

olim7t:
> This particular way of stating the issue looks like an exaggeration to me. Execution may not be strictly dependent on the presence of an observer, specifically when we return `Single` objects during login (cf. `DsePersistence.java:[342,26]`).

I suspected it was something like that. I'll suppress those 3 warnings individually, and we can handle it case by case if it pops up again.

olim7t:
I'd like to have all warnings as errors, but unfortunately ErrorProne doesn't have a flag for that, the only solution would be to list them all one by one.

The next option would be to set `<failOnWarning>true</failOnWarning>` in the POM, but that will also catch warnings unrelated to ErrorProne. In particular this one which doesn't look solvable:
```
[WARNING] Supported source version 'RELEASE_7' from annotation processor 'com.github.misberner.duzzt.processor.DuzztMainProcessor' less than -source '1.8'
```

So I'm afraid it's going to be on us to monitor warnings in the Maven logs.

dimas-b:
> So I'm afraid it's going to be on us to monitor warnings in the Maven logs.

Would it be possible to store the warning somehow during the compilation phase and re-parse during test execution? This way we should be able to convert them to test failures.

pom.xml
pcmanus: Do we have many violations?

In general, we do should be careful to only use effectively immutable values in enum, so I don't dislike the idea of the checker at least reminding us to double-check ourselves. So, unless this triggers all the time (but that would feel surprising), I'd have a slight preference for using `@SuppressWarnings("ImmutableEnumChecker")` with a short sentence to remind why it's ok. 

Though in general, I kind of like the use of `@Immutable` (though it does often require use of ``@SuppressWarnings("Immutable")` because the checker is only so good and we have to be ok with that; this mean it maybe require its own discussion and documentation if we want to go there), so I'm up for that as well.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/interceptors/DefaultQueryInterceptor.java
pcmanus: Yes, pretty sure that's ok (Gossip has quite a few states, not all of which we care for what the interceptor does).

auth-jwt-service/src/test/java/io/stargate/auth/jwt/AuthJWTServiceActivatorTest.java
```diff
@@ -41,6 +41,7 @@ public void shouldRegisterIfSelectedAuthProvider() {
 
     activator.start(bundleContext);
 
+    @SuppressWarnings("JdkObsolete")
```
pcmanus, line 44: Why not replace by a `HashMap`? It doesn't seem we need thread safety.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthJWTServiceActivator.java
```diff
@@ -40,6 +40,7 @@
 
   public static final String AUTH_JWT_IDENTIFIER = "AuthJwtService";
 
+  @SuppressWarnings("JdkObsolete")
```
pcmanus, line 43: I'd replace by a `HashMap` (or `ConcurrentHashMap` if we're freaked out by concurrency, though I don't think it's an issue here).

graphqlapi/src/test/java/io/stargate/graphql/schema/fetchers/dml/ScalarsDmlTest.java
```diff
@@ -192,6 +193,7 @@ public void uuidTypeShouldSupportUuidFunction() {
         .isEqualTo(4);
   }
 
+  @SuppressWarnings("unused")
```
pcmanus, line 196: How comes those are unused, yet useful? Out of curiosity.

persistence-cassandra-4.0/src/main/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
```diff
@@ -39,7 +41,7 @@ public Counter counter(MetricName name) {
     return actualRegistry.counter(name.getMetricName());
   }
 
-  public Counter counter(MetricName name, MetricName alias) {
+  public Counter counter(MetricName name, MetricName alians) {
```
pcmanus, line 44: Is that a typo?

pcmanus (approved):
Lgtm in principles though I'd suggest:
- removing the uses of `HashTable`. I really think there is no good reason to use this.
- I'd consider silencing "FutureReturnValueIgnored" globally. Without completely discarding the usefulness of the check, I feel we end up requiring suppression often enough that it's losing value and become noises.

Neither of which I feel super strongly about though.

auth-jwt-service/src/test/java/io/stargate/auth/jwt/AuthJWTServiceActivatorTest.java
olim7t: The OSGi method where this is used declares `Dictionary`, `HashMap` doesn't implement it.

persistence-cassandra-4.0/src/main/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
olim7t: Yes 😳

graphqlapi/src/test/java/io/stargate/graphql/schema/fetchers/dml/ScalarsDmlTest.java
olim7t: They provide the parameters to the test methods, see the `@MethodSource` annotations. I will add an explanatory comment.

auth-jwt-service/src/test/java/io/stargate/auth/jwt/AuthJWTServiceActivatorTest.java
dimas-b: I wonder if using `Dictionary` as the type of parameters is going to please ErrorProne :thinking:

auth-jwt-service/src/test/java/io/stargate/auth/jwt/AuthnJwtServiceTest.java
```diff
@@ -28,7 +28,6 @@
 import com.nimbusds.jose.proc.SecurityContext;
 import com.nimbusds.jose.proc.SimpleSecurityContext;
 import com.nimbusds.jwt.JWTClaimsSet;
-import com.nimbusds.jwt.JWTClaimsSet.Builder;
```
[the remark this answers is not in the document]
olim7t: It's just a warning, the rationale is that importing common names like "Builder" can make the code harder to read:

> **BadImport**
> Importing nested classes/static methods/static fields with commonly-used names can make code harder to read, because it may not be clear from the context exactly which type is being referred to. Qualifying the name with that of the containing class can make the code clearer.

I tend to agree. In this class it's still pretty clear because the builder is used close to its parent class:
```java
JWTClaimsSet jwtClaimsSet = new Builder().claim("stargate_claims", stargate_claims).build();
```
But in the general case it can become unclear, for example if we store the builder in a field or pass it as a method parameter.

persistence-api/src/main/java/io/stargate/db/query/SchemaKey.java
```diff
@@ -67,7 +67,7 @@ public boolean equals(Object o) {
     if (this == o) {
       return true;
     }
-    if (o == null || getClass() != o.getClass()) {
+    if (!(o instanceof SchemaKey)) {
```
dimas-b, line 70: In this particular case this change looks questionable to me. We have two sub-classes: `PartitionKey` and `PrimaryKey`, assuming the table does not have clustring columns those objects may be mixed up... Do we want to allow equality between a `PartitionKey` instance and a `PrimaryKey` instance in principle? @pcmanus WDYT?

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
```diff
@@ -1438,7 +1440,7 @@ protected BuiltSelect selectQuery() {
         valueCodec,
         executor,
         builder,
-        new HashSet<>(selectedColumns),
+        allSelected,
```
dimas-b, line 1443: Does this change semantics? `allSelected` seems to be modified after it is created from `selectedColumns` :thinking:

persistence-api/src/main/java/io/stargate/db/schema/SchemaBuilderImpl.java
```diff
@@ -345,14 +341,10 @@ public void indexCustom() {
   }
 
   @DSLAction
-  public void from(String fromVertex) {
-    this.fromVertex = fromVertex;
-  }
+  public void from(String fromVertex) {}
```
dimas-b, line 344: This is probably out of scope of this PR but we should remove Graph methods completely. They are apparently no longer part of the DSL language (top-level class annotations).

persistence-cassandra-4.0/src/main/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
dimas-b, line 44: `alians`? typo?

persistence-dse-6.8/src/test/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptorTest.java
```diff
@@ -70,7 +69,7 @@
     }
   }
 
-  private static final Map<String, AbstractType<?>> TYPES =
+  private static final ImmutableMap<String, AbstractType<?>> TYPES =
```
dimas-b, line 72: Why do we have to change the declared type to `ImmutableMap`?

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
```diff
@@ -196,15 +196,15 @@ public void testKeyspace() throws ExecutionException, InterruptedException {
...
-    ResultSet createTable =
-        dataStore
-            .queryBuilder()
-            .create()
-            .table(keyspace, table)
-            .column("created", Boolean, PartitionKey)
-            .build()
-            .execute()
-            .join();
+    //    ResultSet createTable =
```
dimas-b, line 199: nit: Remove this commented-out variable?

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
```diff
@@ -219,15 +219,15 @@ public void testAlterAndDrop() throws ExecutionException, InterruptedException {
...
-    ResultSet alterTable =
-        dataStore
-            .queryBuilder()
-            .alter()
-            .table(keyspace, table)
-            .addColumn("added", Boolean)
-            .build()
-            .execute()
-            .join();
+    //    ResultSet alterTable =
```
dimas-b, line 222: nit: remove?

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
```diff
@@ -241,8 +241,8 @@ public void testAlterAndDrop() throws ExecutionException, InterruptedException {
                 .keyspace(keyspace)
                 .table(table));
 
-    ResultSet dropTable =
-        dataStore.queryBuilder().drop().table(keyspace, table).build().execute().join();
+    //    ResultSet dropTable =
```
dimas-b, line 244: nit: remove?

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
```diff
@@ -287,7 +287,8 @@ public void testColumnKinds() throws ExecutionException, InterruptedException {
   @Disabled("Disabling for now since it currently just hangs")
   @Test
   public void testInsertingAndReadingDifferentTypes() throws Exception {
-    Keyspace ks = createKeyspace();
+    //    Keyspace ks =
```
dimas-b, line 290: nit: remove?

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
olim7t: It does, @pcmanus confirmed that it was a legitimate bug.

persistence-cassandra-4.0/src/main/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
olim7t: 👍  will fix

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
olim7t: The reason I left it (and other occurrences in this class) is that there's other already commented code that references it:
```java
    // TODO: [doug] 2020-07-10, Fri, 16:50 More schemaAgreement to revisit
    // assertThat(createTable.waitedForSchemaAgreement()).isTrue();
```
So if someone were to uncomment that code in the future, they might not know what `createTable` is supposed to be.

The larger debate is whether we should have commented code at all, but this felt a bit off topic for this PR.

persistence-dse-6.8/src/test/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptorTest.java
olim7t: The rationale is that you want to signal the immutability in the declared type, supposedly to prevent someone from trying to modify it. Maybe that's a bit excessive, especially for what's clearly a constant.

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
dimas-b: Ah, good point... I missed that :facepalm:

dimas-b (commented):
LGTM overall +1. My only concern is with `.equals()` on sub-classes of `SchemaKey` (I added s more specific comment on the class itself).

persistence-api/src/main/java/io/stargate/db/query/SchemaKey.java
olim7t: My bad, I didn't notice there were subclasses here.

My guess is that it doesn't matter because the two implementations are probably never compared with each other. But I can revert the changes to keep it safe.

pcmanus: I do would prefer reverting the change and using a suppression (and a short comment). For all that I tend to agree with the argument that `instanceof` is a better "default", I don't think it's an absolute, and that's imo an example.

And yes, pretty sure we never compare `PrimaryKey` with `PartitionKey` now, but that could create confusion some day, and the goal of Error Prone is to help us get code that is less prone to errors, not more.

olim7t:
> * I'd consider silencing "FutureReturnValueIgnored" globally. Without completely discarding the usefulness of the check, I feel we end up requiring suppression often enough that it's losing value and become noises.

Yes I agree, I checked our config for the Java driver and actually we had disabled it. It generates a lot of false positives.

dimas-b (approved):

dougwettlaufer (approved):

PR #429, opened 2020-11-23
Use single-port Dropwizard servers for all web resources

olim7t:
The idea is to centralize all admin features in health-checker: we already do it for metrics, and we should do the same for health checks.

Proof that the `--bind-to-listen-address` option still works as expected:
```
$ lsof -i TCP:8085

# without option:
NODE NAME
 TCP *:8085 (LISTEN)

# with option:
NODE NAME
 TCP 127.0.0.2:8085 (LISTEN)
```


dimas-b (approved):

dimas-b:
Just wondering: why do we need admin connectors at all?

olim7t:
> Just wondering: why do we need admin connectors at all?

Good point. We could expose metrics and health checks as regular resources in health-checker.

health-checker/src/main/java/io/stargate/health/Server.java
```diff
@@ -70,6 +75,21 @@ protected void configure() {
...
+      DefaultServerFactory server =
+          (DefaultServerFactory) applicationConfiguration.getServerFactory();
+      ConnectorFactory connector = server.getAdminConnectors().get(0);
+      if (connector instanceof HttpConnectorFactory) {
+        ((HttpConnectorFactory) connector)
+            .setBindHost(System.getProperty("stargate.listen_address"));
+      } else {
+        throw new AssertionError("Unexpected connector class: " + connector.getClass().getName());
+      }
+    }
```
tomekl007, line 90: maybe add `log.debug` in the else branch to not "fail" silently?

dougwettlaufer (approved):

health-checker/src/main/java/io/stargate/health/Server.java
olim7t: The exception should bubble up and appear in the Stargate logs I think.

olim7t:
Following Dmitri's suggestion, I switched all modules to the "simple" Dropwizard server type, which uses only one port. The health-checker module exposes admin resources (metrics, health checks...) on that port. The other modules expose their application resources unchanged.

dimas-b (approved):

dimas-b (approved):

auth-api/src/main/java/io/stargate/auth/api/impl/AuthApiServerFactory.java
```diff
@@ -0,0 +1,72 @@
...
+import org.eclipse.jetty.server.Handler;
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.util.thread.ThreadPool;
+
+/**
+ * @see "META-INF/services/io.dropwizard.server.ServerFactory"
+ * @see "config.yaml"
+ */
+@JsonTypeName("auth-api")
+public class AuthApiServerFactory extends SimpleServerFactory {
```
olim7t, line 34: The admin resources were still exposed under `/admin` (default path), this completely removes them.

It's the exact same class for all 3 modules. I would have liked to factor it in `core`, but this introduces more issues with OSGi 😞

dimas-b: What kind of issues? :)

We could create a new module (e.g. `web-common`) and produce a _plain_ jar from it that would then be _embedded_ into the various OSGi bundles (unlike `core`, which is a bundle in itself). I believe this way would give us ability to reuse code without introducing dependency problems.

olim7t: I've created a follow-up ticket: #439.

PR #431, opened 2020-11-24
Publish health checks from each bundle

olim7t:
@tomekl007 this is an attempt at introducing a mechanism where the health-checker module would act as a centralized registry where each bundle could register its checks.

http://127.0.0.2:8085/healthcheck?pretty=true


graphqlapi/src/main/java/io/stargate/graphql/GraphqlActivator.java
```diff
@@ -87,6 +96,19 @@ private synchronized void maybeStopService() {
...
+
+  private static class GraphqlHealthCheck extends HealthCheck {
+
+    private volatile boolean healthy = false;
+
+    @Override
+    protected Result check() throws Exception {
+      return healthy
+          ? Result.healthy("Ready to process requests")
+          : Result.unhealthy("Server not started");
```
olim7t, line 111: Not sure if it's the kind of health check we'll want eventually, this is more of a proof of concept at this point.

graphqlapi/src/main/java/io/stargate/graphql/GraphqlActivator.java
```diff
@@ -87,6 +96,19 @@ private synchronized void maybeStopService() {
...
+      graphqlHealthCheck.healthy = false;
+    }
+  }
+
+  private static class GraphqlHealthCheck extends HealthCheck {
+
+    private volatile boolean healthy = false;
+
+    @Override
+    protected Result check() throws Exception {
```
tomekl007, line 108: This check looks good.
For the Kafka CDC producer I would like to expose a list of checks:
`ZookeperHealthCheck`, `KafkaHealthCheck` and `SchemaRegistryHealthCheck`.
With the current solution, I can construct one result:
```
Result.builder().
              withMessage("Kafka cdc Producer running")
              .withDetail("kafka", "UP")
              .withDetail("zookeeper", "UP")
              .withDetail("schema-registry", "UP")
              .build()
```
However, I wonder that maybe it will be better to register N health-checks for every system:
```
Result kafkaHealthCheck = Result.healthy("kafka UP");
Result zookeeperHealthCheck = Result.healthy("zookeeper UP");
healthCheckRegistry.get().register("kafkaCdcProducer.kafka", kafkaHealthCheck);
healthCheckRegistry.get().register("kafkaCdcProducer.zookeeper", zookeeperHealthCheck);
```
What is important, both solutions seem feasible using the approach implemented here so I am +1 on it.

graphqlapi/src/main/java/io/stargate/graphql/GraphqlActivator.java
```diff
@@ -52,6 +58,7 @@ public GraphqlActivator() {
   @Override
   @Nullable
   protected ServiceAndProperties createService() {
+    healthCheckRegistry.get().register("graphql", graphqlHealthCheck);
```
tomekl007, line 61: will it be evaluated every time the `/health` endpoint is called?

graphqlapi/src/main/java/io/stargate/graphql/GraphqlActivator.java
olim7t: Yeah the keys are simple strings so there are no constraints on the client bundles. Registering multiple checks for a given API sounds like a perfectly valid approach.

graphqlapi/src/main/java/io/stargate/graphql/GraphqlActivator.java
olim7t: Yes, the `HealthCheck.check()` method is invoked every time the URI is accessed.

health-checker/src/main/java/io/stargate/health/Server.java
```diff
@@ -75,6 +79,7 @@ public void initialize(final Bootstrap<ApplicationConfiguration> bootstrap) {
     super.initialize(bootstrap);
     bootstrap.setConfigurationSourceProvider(new ResourceConfigurationSourceProvider());
     bootstrap.setMetricRegistry(metrics.getRegistry("health-checker"));
+    bootstrap.setHealthCheckRegistry(healthCheckRegistry);
```
tomekl007, line 82: can we add an integration test that validates if health-check was exposed on the dropwizard `/helalth` endpoint?

olim7t: 👍  done

tomekl007 (approved):

PR #445, opened 2020-12-01
Extend BaseActivator with lazy init support and migrate all `CassandraActivators`

tomekl007:


persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/CassandraPersistenceActivator.java
```diff
@@ -135,78 +133,39 @@ public static Config makeConfig(File baseDir) throws IOException {
...
+    // TODO copy metrics if this gets invoked more than once?
+    CassandraMetricsRegistry.actualRegistry =
+        metrics.get().getRegistry("persistence-cassandra-3.11");
 
     try {
       // Throw away data directory since stargate is ephemeral anyway
       File baseDir = Files.createTempDirectory("stargate-cassandra-3.11").toFile();
 
-      cassandraDB.setAuthorizationService(authorizationService);
+      cassandraDB.setAuthorizationService(authorizationService.get());
```
dimas-b, line 148: There is a direct call path from `CassandraPersistenceActivator` to `StargateQueryHandler`, which uses the `AuthorizationService`. So, it should be possible to avoid using an `AtomicReference` in this case and pass the service object to the `StargateQueryHandler` when the service is available. `StargateQueryHandler` may still use `AtomicReference` for storing the reference, but in terms of method parameters all of them will declare `AuthorizationService` explicitly. I believe that would be nicer from the API definition perspective. 

`BaseActivator` could provide a callback when the "lazy" service becomes available. WDYT?

tomekl007: I wanted to keep the existing behavior. If it can be avoided, why it was introduced in the first place?
I think that If It will return the object not wrapped in `AtomicReference` we won't be able to fulfill it later. 
The `CassandraPersistenceActivator`  needs to register `Persistence` BEFORE the `AuthorizationService ` is available.
If we will not return `AtomicReference<T>` it will just return `null`.
Regarding callback, yes it may be useful but the question is where. I don't see a use case for it. We can create a callback mechanism instead of this `AtomicReference`, but I am not sure if this is simpler.

dimas-b: From my POV the `AtomicReference` is a remnant of earlier implementation approaches. I believe it is not necessary even in current code, but since we're refactoring the basic lazy init logic, I think it makes sense to fix it now (if people agree).

tomekl007: We need some form of postponing the init. It blocks when there is no lazy init mechanism (I've tested that). The reason for this is that `AuthorizationService` is injected AFTER the CassandraActivator creates service. We need some form of this mechanism.

dimas-b: @dougwettlaufer : WDYT?

mpenick: The `AtomicReference` in `CassandraPersistence` is unnecessary.

```java
diff --git a/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java b/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
index e50d6b36..4342c62a 100644
--- a/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
+++ b/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
@@ -104,7 +104,6 @@ public class CassandraPersistence
 
   // C* listener that ensures that our Stargate schema remains up-to-date with the internal C* one.
   private SchemaChangeListener schemaChangeListener;
-  private AtomicReference<AuthorizationService> authorizationService;
 
   public CassandraPersistence() {
     super("Apache Cassandra");
@@ -180,7 +179,6 @@ public class CassandraPersistence
 
     interceptor.initialize();
     stargateHandler().register(interceptor);
-    stargateHandler().setAuthorizationService(this.authorizationService);
   }
 
   @Override
@@ -307,7 +305,7 @@ public class CassandraPersistence
   }
 
   public void setAuthorizationService(AtomicReference<AuthorizationService> authorizationService) {
-    this.authorizationService = authorizationService;
+    stargateHandler().setAuthorizationService(authorizationService);
   }
 
   private class CassandraConnection extends AbstractConnection {
```

tomekl007: I don't understand this comment. 
You are still using it in the 
```  
public void setAuthorizationService(AtomicReference<AuthorizationService> authorizationService) 
```

mpenick: Why do you need to have this intermediate reference in `CasasndraPersistence`? 
`private AtomicReference<AuthorizationService> authorizationService;`

`setAuthoriazationService()` is called before initialization.

```java
      cassandraDB.setAuthorizationService(authorizationService.get());
      cassandraDB.initialize(makeConfig(baseDir));
```

 Why not just set the atomic reference in `stargateHandler` directly?

dimas-b: What I'm proposing is to call `cassandraDB.setAuthorizationService(authorizationService)` (no `AtomicReference`) from a (new) callback invoked from the `BaseActivator` code when the service becomes available.

tomekl007: yes, the callback will be also a feasible solution. I can try to implement it and then we will compare it with the `AtomicReference` solution.

mpenick (approved):
LGTM. I'm fine with using the `AtomicReference` vs a callback. Either way it requires a volatile because of the cross thread access. One benefit of staying with `AtomicReference` is it results in less code churn.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/CassandraPersistenceActivator.java
dougwettlaufer: Really just as long as we can prevent AuthorizationService from blocking activation but it being there by the time we get to StargateQueryHandler whatever works. Which we know AuthorizationService has to be registered by the time a request comes in since it blocks the activation of all other user facing modules.

PR #451, opened 2020-12-02
Fix cassandra authz (fixes #449)

### LINKED ISSUE #449: GraphQl should prevent schema modification on tables with only SELECT role.
When the SELECT role is granted on the specific `ks.table`, users are still able to modify this table using graphQL API (`http://localhost:8080/graphql-schema` endpoint)
Steps to reproduce:
1. Enable auth on C* node:
modify the `cassandra.yaml` with
```
authenticator: PasswordAuthenticator
authorizer: CassandraAuthorizer
```

2. Start stargate with enable auth:
```
./starctl --cluster-name stargate --cluster-seed 127.0.0.1 --cluster-version 3.11 --listen 127.0.0.10 --simple-snitch --cql-port 9043 --enable-auth
```

3. Create keyspace:
```
ccm node1 cqlsh -p cassandra -u cassandra
CREATE KEYSPACE IF NOT EXISTS dataprovider WITH REPLICATION = {'class':'SimpleStrategy', 'replication_factor':'1'}
```

4. Create a table using graphQL API (`http://localhost:8080/graphql-schema`)
```
# Write your query or mutation here
mutation {
    communication: createTable(
    keyspaceName: "dataprovider"
    tableName: "communication"
    partitionKeys: [{ name: "id", type: { basic: VARINT } }]
    values: [
      # The values associated with the keys
      { name: "channel", type: { basic: TEXT } }
      { name: "cret", type: { basic: INT } }

    ]
  )
 }
```

5. Create a new role and grant only SELECT on this table.
```
ccm node1 cqlsh -p cassandra -u cassandra
CREATE ROLE IF NOT EXISTS 'read_only_user' WITH PASSWORD = 'read_only_user' AND LOGIN = TRUE;
GRANT SELECT ON TABLE dataprovider.communication TO read_only_user;
```

6. Execute the CREATE TABLE request from graphQL API again (using Cassandra username and token). I don't see an auth failure.  
It seems that this is a missing feature (or bug) in Stargate

dougwettlaufer:
fixes #449


auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthzJwtService.java
```diff
@@ -117,7 +117,8 @@ public void authorizeDataWrite(String token, String keyspace, String table, Scop
 
   /** {@inheritdoc} */
   @Override
-  public void authorizeDataWrite(String token, List<TypedKeyValue> typedKeyValues, Scope scope)
+  public void authorizeDataWrite(
+      String token, String keyspace, String table, List<TypedKeyValue> typedKeyValues, Scope scope)
```
dimas-b, line 121: I'm a bit lost... the newly added `keyspace` and `table` parameters do not seem to be used... did I miss something?

dougwettlaufer: For a JWT we're assuming that a role+permissions have been setup meaning that the keyspace and table are not used. But for extensibility to add a more advanced auth provided like say, OPA, you'll need to know the keyspace and table that are being accessed.

dimas-b (approved):

testing/src/test/java/io/stargate/it/http/RestApiTableTokenAuthTest.java
```diff
@@ -0,0 +1,213 @@
...
+
+    RestUtils.post(
+        authToken,
+        String.format("%s:8082/v2/schemas/keyspaces/%s/tables", host, keyspace),
+        objectMapper.writeValueAsString(tableAdd),
+        HttpStatus.SC_CREATED);
+
+    try {
+      // Need to sleep to allow propagation of permissions
+      Thread.sleep(5000L);
```
tomekl007, line 125: can we use `Awaiatality` (https://github.com/awaitility/awaitility) with `await.until()`?

dougwettlaufer: Actually this is no longer needed. Forgot to remove after debugging.

PR #455, opened 2020-12-03
REST: Allow non-string JSON types for incoming column data (fixes #49)

### LINKED ISSUE #49: Rest Api V2: not able to handle integer in post request
```
curl --request POST \
  --url 'localhost:8082/v2/keyspaces/a_manu/widgets' \
 --header 'x-cassandra-token: ef15d2dc-0741-4d58-a499-3ada37788d51'  --header 'accepts: application/json' \
--header 'content-type: application/json' \
--data '{"hour_created":"2020-01-24T11:00:00.00Z","machine_code":"131", "serial_number":"afadfaf","times_inspected":1}' 
{"description":"Server error: java.lang.Integer cannot be cast to java.lang.String","code":500}%
```

start of the stack

```
CREATE TABLE a_manu.widgets (
    hour_created timestamp,
    machine_code text,
    serial_number text,
    created_at timestamp,
    est_unit_cost decimal,
    est_unit_cost_updated timestamp,
    factory_code text,
    inspection_notes text,
    last_inspected_at timestamp,
    part_name text,
    part_number text,
    times_inspected int,
    PRIMARY KEY ((hour_created, machine_code), serial_number)
) WITH CLUSTERING ORDER BY (serial_number ASC)
```

```
ERROR [2020-09-14 14:15:39,471] io.stargate.web.resources.RequestHandler: Error when executing request
! java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
! at io.stargate.web.resources.Converters.colToValue(Converters.java:249)
! at io.stargate.web.resources.v2.RowsResource.lambda$null$2(RowsResource.java:150)
! at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
! at java.util.Iterator.forEachRemaining(Iterator.java:116)
! at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
! at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
! at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
! at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
! at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
! at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:566)
! at io.stargate.web.resources.v2.RowsResource.lambda$add$3(RowsResource.java:151)
! at io.stargate.web.resources.RequestHandler.handle(RequestHandler.java:24)
! at io.stargate.web.resources.v2.RowsResource.add(RowsResource.java:141)
! at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
! at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
! at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
! at java.lang.reflect.Method.invoke(Method.java:498)
```

There is a hack workaround for this, you can just send the integer as a string and it will work, but this will trip a lot of folks up.

olim7t:
(Note that this is based on #413 which is not merged yet.)

Revisit REST API v2 so that it accepts all JSON types when posting row data, for example JSON number for CQL int, etc.

I reused some code from the GraphQL module and the Java driver, but with some adaptations so I don't think it's worth trying to factor it at this point.

The v1 API is unchanged: only strings are allowed. However I fixed a few things that were likely bugs:
- the blob implementation was using the string's bytes. I switched to base64, as we've been doing in other places.
- the collection implementations did not handle some corner cases: strings elements that contain quotes or commas.
- the tuple implementation always returned an empty tuple.
- the UDT implementation was using the string's bytes and assuming that it contained the binary representation of the tuple.

TODO:
- [ ] cover error cases in unit test
- [x] check integration tests
- [x] revisit responses as well. Currently we just take the objects returned by the persistence API and pass them directly to the Jackson object mapper. This does not work for complex types: for example maps are serialized as JSON objects (wrong because it always turns the keys into strings), UDTs reveal all the internal fields of the `UdtValue` object, etc.


testing/src/test/java/io/stargate/it/http/RestApiTest.java
```diff
@@ -689,7 +689,7 @@ public void addRowWithList() throws IOException {
 
     ColumnModel emailColumn = new ColumnModel();
     emailColumn.setName("email");
-    emailColumn.setValue("[foo@example.com,bar@example.com]");
+    emailColumn.setValue("['foo@example.com','bar@example.com']");
```
olim7t, line 692: Leaving the strings unquoted works on simple examples (see `ConvertersTest`), but here the parser gets stuck on the `@`.

I guess that could be considered a regression, but on the other hand if we want to handle strings containing quotes or commas we need to be more strict than a simple `String.split`.

restapi/src/test/java/io/stargate/web/resources/ConvertersTest.java
```diff
@@ -0,0 +1,201 @@
...
+  private static Arguments[] toJsonSamples() throws Exception {
+    return new Arguments[] {
+      arguments(
+          UUID.fromString("59e86020-3502-11eb-bc1b-53a4f383ccdf"),
+          "\"59e86020-3502-11eb-bc1b-53a4f383ccdf\""),
+      arguments("abc", "\"abc\""),
+      arguments(CqlDuration.newInstance(0, 1, 0), "\"1d\""),
+      arguments(1L, "\"1\""),
+      arguments(BigInteger.ONE, "\"1\""),
+      arguments(BigDecimal.ONE, "\"1\""),
```
dougwettlaufer, line 71: I'm confused, why are these a string "1" rather than number 1?

dougwettlaufer: Is it the same for a boolean?

restapi/src/main/java/io/stargate/web/resources/Converters.java
```diff
@@ -78,336 +82,762 @@
...
+    Column.ColumnType keyType = type.parameters().get(0);
+    Column.ColumnType valueType = type.parameters().get(1);
+    LinkedHashMap<Object, Object> result = new LinkedHashMap<>();
+    int index = 0;
+    for (Object jsonValue : jsonEntries) {
+      if (!(jsonValue instanceof Map)
+          || !((Map) jsonValue).containsKey("key")
+          || !((Map) jsonValue).containsKey("value")) {
+        throw new IllegalArgumentException(
+            "Invalid Map value: inner elements must be objects with two fields 'key' and 'value'");
```
dougwettlaufer, line 381: Why must the fields be 'key' and 'value'?

restapi/src/main/java/io/stargate/web/resources/Converters.java
```diff
@@ -78,336 +82,762 @@
     final Map<String, Object> map = new HashMap<>(defs.size());
     for (final Column column : defs) {
 
-      map.put(column.name(), transformObjectToJavaObject(row.getObject(column.name())));
+      map.put(column.name(), toJsonValue(row.getObject(column.name())));
```
dougwettlaufer, line 85: Being paranoid since this is also called in v1, nothing changes about the response. Right?

restapi/src/main/java/io/stargate/web/resources/ParseUtils.java
```diff
@@ -0,0 +1,141 @@
...
+
+    int cbrackets = 0;
+    int sbrackets = 0;
+    int parens = 0;
+    boolean inString = false;
+
+    do {
+      char c = toParse.charAt(idx);
+      if (inString) {
+        if (c == '\'') {
```
dougwettlaufer, line 54: What about `\"`? Is that valid?

testing/src/test/java/io/stargate/it/http/RestApiTest.java
dougwettlaufer: Odd that it gets stuck on `@`. Shouldn't that be treated like any other char?

restapi/src/main/java/io/stargate/web/resources/ParseUtils.java
```diff
@@ -0,0 +1,141 @@
...
+   * Assuming that idx points to the beginning of a CQL identifier in toParse, returns the index of
+   * the first character after this identifier.
+   *
+   * @param toParse the string to skip an identifier from.
+   * @param idx the index to start parsing an identifier from.
+   * @return the index ending the CQL identifier starting at {@code idx}.
+   * @throws IllegalArgumentException if idx doesn't point to the start of a valid CQL identifier.
+   */
+  static int skipCQLId(String toParse, int idx) {
+    if (idx >= toParse.length()) throw new IllegalArgumentException();
```
dougwettlaufer, line 106: Should the `throw new IllegalArgumentException();` in this method have messages or was this intentional?

dougwettlaufer (approved):
A few small comments but looks good otherwise. Looks like this will also fix #72

restapi/src/test/java/io/stargate/web/resources/ConvertersTest.java
mpenick: Maybe it's because the ranges of the types long (64-bit integer), bigint and bigdecimal cannot be represented in a json number type?

olim7t: Yes, it's for interoperability issues. JSON itself does not impose limitations on numbers; but REST APIs are often consumed from Javascript clients, which don't properly handle integers above 2^53. This can be observed in your browser's Javascript console: 
```
> JSON.parse("9223372036854775807")
9223372036854776000
```

This only affects Long, BigInteger and BigDecimal. Boolean and the smaller int types are returned as their direct JSON counterparts. I forgot to add test samples for them, will do.

restapi/src/main/java/io/stargate/web/resources/Converters.java
olim7t: This is what a serialized map looks like:
```
[{"key":1,"value":"a"},{"key":2,"value":"b"}]
```
The entries are objects with two fields, these were the most logical names for them.

Alternatively, we could use two-element arrays, or allow arbitrary field names, but I don't like either of them because it's less clear and requires them to be in a given order.

restapi/src/main/java/io/stargate/web/resources/Converters.java
olim7t: It doesn't impact the structure of the response, but changes how some of the column values are represented. I can keep v1 exactly the way it was, but that includes the bugs about large numbers, map keys, and severely broken UDTs and tuples. For example:

* before:
    ```
    {"count":1,"data":[{"a":{"type":{"keyspace":{},"name":{},"frozen":false,"fieldNames":[{}],"fieldType
    s":[{"protocolCode":9,"detached":false}],"attachmentPoint":{"protocolVersion":"V4","codecRegistry":{
    }},"detached":false,"protocolCode":48},"formattedContents":"{i:2}"},"k":1}]}
    ```

* after:
    ```
    {"count":1,"data":[{"a":{"i":2},"k":1}]}
    ```

restapi/src/main/java/io/stargate/web/resources/ParseUtils.java
olim7t: Inside strings, yes.

restapi/src/main/java/io/stargate/web/resources/Converters.java
olim7t: Based on offline conversation, reverting v1 to previous behavior.

restapi/src/test/java/io/stargate/web/resources/ConvertersTest.java
olim7t: Also forgot to mention that for BigDecimal it's not only a matter of range, but also precision. The same reason you don't use `float` or `double` to represent exact values in pure Java.

restapi/src/main/java/io/stargate/web/resources/ParseUtils.java
olim7t: This is only called from `Converters.toCqlUdt`, which catches the exception and rethrows a different one.

In fact we could use a special return value instead of throwing an exception, I'll make the change.

olim7t:
I reverted the V1 API to its previous behavior. To sum up the differences:

V1 request:
```
curl -G -X GET http://127.0.0.2:8082/v1/keyspaces/test/tables/foo/rows/1 \
  -H "X-Cassandra-Token: $TOKEN" -H 'Content-Type: application/json'
```
V2 request:
```
curl -G -X GET http://127.0.0.2:8082/v2/keyspaces/test/foo \
  -H "X-Cassandra-Token: $TOKEN" -H 'Content-Type: application/json' \
  --data-urlencode 'where={"k":{"$eq":1}}'
```

#### Decimal
```
create table foo(k int primary key, v decimal);
insert into foo(k,v) values (1, 0.2);
```
V1 response:
```
{"count":1,"rows":[{"v":0.2,"k":1}]}
```
Looks correct, but would cause issues when parsed by a Javascript client:
```
d = JSON.parse("0.2")
s = 0
for (i = 0; i < 100; i++) s += d
// 19.99999999999996
```
Similar issues with Long and Bigint.

V2 response:
```
{"count":1,"data":[{"v":"0.2","k":1}]}
```

#### Maps
```
create table foo(k int primary key, v map<blob,int>);
insert into foo(k,v) values (1, {0x01:1, 0x02:2});
```
V1 response:
```
{"count":1,"rows":[{"v":{"java.nio.HeapByteBuffer[pos=0 lim=1 cap=22]":1,"java.nio.HeapByteBuffer[pos=0 lim=1 cap=9]":2},"k":1}]}
```
V2 response:
```
{"count":1,"data":[{"v":[{"key":"AQ==","value":1},{"key":"Ag==","value":2}],"k":1}]}
```

#### Tuples (and UDTs)
```
create table foo(k int primary key, v tuple<int,int>);
insert into foo (k,v) values (1, (2,3));
```
V1 reponse:
```
{"code":400,"message":"Unable to process JSON"}
```
V2 reponse:
```
{"count":1,"data":[{"v":[2,3],"k":1}]}
```

PR #493, opened 2020-12-03
Add cpu metrics

dougwettlaufer:


health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
```diff
@@ -0,0 +1,97 @@
...
+      // Enough time has elapsed.
+      updateSystemTicks();
+    }
+
+    long user = curTicks[TickType.USER.getIndex()] - prevTicks[TickType.USER.getIndex()];
+    long nice = curTicks[TickType.NICE.getIndex()] - prevTicks[TickType.NICE.getIndex()];
+    long sys = curTicks[TickType.SYSTEM.getIndex()] - prevTicks[TickType.SYSTEM.getIndex()];
+    long idle = curTicks[TickType.IDLE.getIndex()] - prevTicks[TickType.IDLE.getIndex()];
+    long iowait = curTicks[TickType.IOWAIT.getIndex()] - prevTicks[TickType.IOWAIT.getIndex()];
+    long totalCPU = user + nice + sys + idle + iowait;
```
dimas-b, line 62: I do not really know how `oshi.hardware` reports data, but is it guaranteed that these numbers properly add up to the "tocal CPU"? I'd guess total should be a deterministic value based on the CPU frequency and time interval... is there a method in this tool to get the total without adding up components?

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
```diff
@@ -0,0 +1,97 @@
...
+    // Skip update if ticks is all zero.
+    // Iterate to find a nonzero tick value and return; this should quickly
+    // find a nonzero value if one exists and be fast in checking 0's
+    // through branch prediction if it doesn't
+    for (long tick : ticks) {
+      if (tick != 0) {
+        // We have a nonzero tick array, update and return!
+        this.tickTime = System.currentTimeMillis();
+        // Copy to previous
+        System.arraycopy(this.curTicks, 0, this.prevTicks, 0, this.curTicks.length);
```
dimas-b, line 91: If we're concerned about performance (looking at the comment above), it should be possible to avoid this array copy by alternating between "current" and "previous" arrays (e.g. by using `long[0][TickType.values().length]`). WDYT?

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
```diff
@@ -0,0 +1,97 @@
...
+
+    return processor.getSystemCpuLoadBetweenTicks(prevTicks);
+  }
+
+  private Double getTick(String tickType) {
+    long now = System.currentTimeMillis();
+    if (now - this.tickTime > 950) {
+      // Enough time has elapsed.
+      updateSystemTicks();
+    }
```
dimas-b, line 55: Time threshold can be overrun in-between `getTick` calls for different metrics, thus their values will be based on different tick samples... not a huge deal, but might lead to oddities like percentages not adding up to `100`, I guess.

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
dougwettlaufer: > is there a method in this tool to get the total without adding up components

No, unfortunately you just have and array of "ticks"

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
dougwettlaufer: Honestly for alerting and such I would want to deal with thresholds and trends. Answering things like "is there a spike in CPU usage when X event happens" or "are we using more than 80% of the CPU". Also, I'm on the fence of whether I want to expose all of these or not (like "nice" for example).

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
dougwettlaufer: > by using long[2][TickType.values().length]
Not sure I follow

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
dougwettlaufer: For most use cases I'd imagine cpuload would be enough but I can see where iowait would be useful. There's actually more ticks available that I'm omitting
https://github.com/goodyinxin/eladmin/blob/4adea4ebe2c4ca6f1c691f64036cc96a4a10737b/eladmin-monitor/src/test/java/me/zhengjie/SystemInfoTest.java#L194-L202

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
```diff
@@ -0,0 +1,97 @@
...
+import com.codahale.metrics.Metric;
+import com.codahale.metrics.MetricSet;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import oshi.SystemInfo;
+import oshi.hardware.CentralProcessor;
+import oshi.hardware.CentralProcessor.TickType;
+
+public class CPUGaugeMetricSet implements MetricSet {
```
tomekl007, line 13: Tests unit OR/AND IT are missing. The logic is not trivial, please add tests. We can mock the CPU system provider, and test our logic.

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
```diff
@@ -0,0 +1,97 @@
...
+      // Enough time has elapsed.
+      updateSystemTicks();
+    }
+
+    return processor.getSystemCpuLoadBetweenTicks(prevTicks);
+  }
+
+  private Double getTick(String tickType) {
+    long now = System.currentTimeMillis();
+    if (now - this.tickTime > 950) {
```
tomekl007, line 52: why `950`? Please extract it to constant.

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
```diff
@@ -0,0 +1,97 @@
...
+
+    long user = curTicks[TickType.USER.getIndex()] - prevTicks[TickType.USER.getIndex()];
+    long nice = curTicks[TickType.NICE.getIndex()] - prevTicks[TickType.NICE.getIndex()];
+    long sys = curTicks[TickType.SYSTEM.getIndex()] - prevTicks[TickType.SYSTEM.getIndex()];
+    long idle = curTicks[TickType.IDLE.getIndex()] - prevTicks[TickType.IDLE.getIndex()];
+    long iowait = curTicks[TickType.IOWAIT.getIndex()] - prevTicks[TickType.IOWAIT.getIndex()];
+    long totalCPU = user + nice + sys + idle + iowait;
+
+    switch (tickType) {
+      case "user":
```
tomekl007, line 65: Maybe we can create an enum of supported ticks and use it everywhere, instead of relying on Strings?

tomekl007 (changes requested):

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
dimas-b: `long[0]` starts as the "previous array, `long[1]` as "current". At the next refresh `long[0]` becomes "current", `long[1]` "previous"... We could keep indexes `prev = 0`, `current = 1` initially and then do `prev = 1 - prev`, `current = 1 - current` during refresh.

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
dimas-b: From that link it looks like we're missing some data in the calculation of `totalCPU` :thinking:

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
dimas-b: (also mention is another comment, but relevant here)
This example uses more ticks for calculating `totalCPU`:
https://github.com/goodyinxin/eladmin/blob/4adea4ebe2c4ca6f1c691f64036cc96a4a10737b/eladmin-monitor/src/test/java/me/zhengjie/SystemInfoTest.java#L194-L202

If the tool reports CPU usage as a collection of ticks, I suppose we ought to use them all for `totalCPU` even if we do not expose some of them as metrics.

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
dougwettlaufer: Only used 2 places in the same class but sure.

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
dougwettlaufer: Yeah I left those out since they're not being exposed. Was thinking we'd just calculate percent of CPU based on what's being exposed. I can put them back if you'd prefer.

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
dougwettlaufer: I'm just grabbing an array and adding some numbers together. Seems pretty straightforward to me.

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
dougwettlaufer: Because it's what OSHI used? 
https://github.com/oshi/oshi/blob/3.13.X/oshi-core/src/main/java/oshi/hardware/common/AbstractCentralProcessor.java#L381

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
dimas-b: I believe the percentage calculation can be incorrect if we leave out those other numbers. I assume each CPU tick goes into one of those buckets, so we have to account for all of them in the total.

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
dimas-b: FWTW, we could use a unit test to make sure input tick numbers are wired to the right metric name.

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
dougwettlaufer: Added all of them

health-checker/src/main/java/io/stargate/health/CPUGaugeMetricSet.java
dougwettlaufer: added

dimas-b (approved):

health-checker/src/test/java/io/stargate/health/CPUGaugeMetricSetTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+    CentralProcessor processor = mock(CentralProcessor.class);
+    CPUGaugeMetricSet cpuGaugeMetricSet = new CPUGaugeMetricSet(processor, curTicks, prevTicks);
+
+    when(processor.getSystemCpuLoadTicks()).thenReturn(new long[0]);
+    when(processor.getSystemCpuLoadBetweenTicks(any())).thenReturn(0.1);
+
+    Map<String, Metric> metrics = cpuGaugeMetricSet.getMetrics();
+
+    assertThat(((Gauge<Double>) metrics.get("load")).getValue()).isEqualTo(0.1);
```
tomekl007, line 44: why those numbers don't sum up to `1.0`?

PR #498, opened 2020-12-04
Add new authz to docsapi

dougwettlaufer:


restapi/src/main/java/io/stargate/web/docsapi/resources/NamespacesResource.java
```diff
@@ -211,6 +223,9 @@ public Response createNamespace(
             replication = Replication.simpleStrategy((int) requestBody.getOrDefault("replicas", 1));
           }
 
+          db.getAuthorizationService()
+              .authorizeSchemaWrite(token, keyspaceName, null, Scope.CREATE);
```
dimas-b, line 227: Is this that same as lines 206-207?

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -576,13 +600,18 @@ private void validateOpAndValue(String op, JsonNode value, String fieldName) {
...
     for (PathSegment pathSegment : path) {
       String pathStr = pathSegment.getPath();
       convertedPath.add(convertArrayPath(pathStr));
     }
     Long now = ChronoUnit.MICROS.between(Instant.EPOCH, Instant.now());
+
+    db.getAuthorizationService()
+        .authorizeDataWrite(db.getAuthToken(), keyspace, collection, Scope.DELETE);
+
     db.delete(keyspace, collection, id, convertedPath, now);
```
dimas-b, line 615: Would it make sense to move the `authorizeDataWrite` call inside `db.delete` to have to apply to all `delete` cases (essentially reducing the number of places where `authorizeDataWrite` is invoked)?

mpenick: I wonder if it makes sense to do this in general? That is, to push all the authz into the db layer. It has access to the authz service and auth token already and it's aware of the type of modification being made. It might be cleaner and more potentially less error prone.

dougwettlaufer: Fixed.

restapi/src/main/java/io/stargate/web/docsapi/resources/NamespacesResource.java
dougwettlaufer: Oops, copied instead of cut.

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -887,12 +912,15 @@ private void updateExistenceForMap(
         ImmutableList.of(BuiltCondition.of("key", Predicate.IN, new ArrayList<>(docNames)));
 
     db = dbFactory.getDocDataStoreForToken(authToken);
+
     List<Row> rows = db.executeSelect(keyspace, collection, predicate).rows();
     Map<String, List<Row>> rowsByDoc = new HashMap<>();
     for (Row row : rows) {
       String key = row.getString("key");
       List<Row> rowsAtKey = rowsByDoc.getOrDefault(key, new ArrayList<>());
-      if (fields.isEmpty() || fields.contains(row.getString("p0"))) rowsAtKey.add(row);
```
EricBorczuk, line 895: Did our style rules change? pretty sure `mvnw xml:format format` caused this `if` to end up this way, weird

EricBorczuk (approved):
looks good to me, pending other suggestions!

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
dougwettlaufer: Maybe? Personally I'm all for it. I know java supports one line if without curly braces but I'm not a fan.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -49,6 +57,8 @@
       new String[] {"leaf", "text_value", "dbl_value", "bool_value"};
 
   final DataStore dataStore;
+  AuthorizationService authorizationService;
+  String authToken;
```
dimas-b, line 61: nit: `final`?

dimas-b (approved):

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
dougwettlaufer: Fiiiiine :)

dimas-b:
I'm just wondering why we use row-level filtering in `RowsResource.getRows` but most (if not all) of the checks added in this PR only do authorization at the table level? :thinking:

dougwettlaufer:
> I'm just wondering why we use row-level filtering in `RowsResource.getRows` but most (if not all) of the checks added in this PR only do authorization at the table level? 🤔

Because the auth model is going to be slightly different here. RLAC for a table will be different than RLAC (FLAC?) for a document.

PR #501, opened 2020-12-04
Use a tinyint for backends where secondary indexes are not supported

EricBorczuk:


restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -1150,9 +1157,17 @@ private boolean pathsMatch(String path1, String path2) {
         .collect(Collectors.toList());
   }
 
+  private Boolean getBooleanFromRow(Row row, String colName) {
+    if (DocumentDB.USE_CNDB) {
+      byte value = row.getByte(colName);
+      return value != 0;
```
dimas-b, line 1163: Will this work with `NULL`?

dimas-b, line 1163: Will this work with `NULL`?

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -34,6 +34,7 @@
   private static final List<String> allPathColumnNames;
   private static final List<Column.ColumnType> allPathColumnTypes;
   public static final Integer MAX_DEPTH = Integer.getInteger("stargate.document_max_depth", 64);
+  public static final Boolean USE_CNDB = Boolean.getBoolean("stargate.use_cndb");
```
dimas-b, line 37: suggestion: `stargate.docsapi.use_tinyint_for_bool`

persistence-api/src/main/java/io/stargate/db/datastore/DataStore.java
```diff
@@ -151,6 +151,11 @@ default QueryBuilder queryBuilder() {
    */
   Schema schema();
 
+  /** Returns true if the persistence backend supports secondary indexes */
+  default boolean supportsSecondaryIndex() {
+    return true;
```
dimas-b, line 156: Having two defaults for the same flag seems a bit error-prone to me :thinking: how about we keep the default on `Persistence` and `DataStore` impl. will redirect there?

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -1144,15 +1166,25 @@ private boolean pathsMatch(String path1, String path2) {
...
+                  && allFiltersMatch(fieldRow.get(), inMemoryFilters, numericBooleans);
             })
         .flatMap(x -> x.stream())
         .collect(Collectors.toList());
   }
 
-  private boolean allFiltersMatch(Row row, List<FilterCondition> filters) {
+  private Boolean getBooleanFromRow(Row row, String colName, boolean numericBooleans) {
+    if (row.isNull("bool_value")) return null;
+    if (numericBooleans) {
```
dimas-b, line 1178: Looks like this flag is use in many places... Would it make sense to convert it to an field on the `DocumentService` object?

persistence-api/src/main/java/io/stargate/db/datastore/DataStore.java
EricBorczuk: I was kinda trying to do that - but since according to this hierarchy it's possible that a DataStore does not have a Persistence (i.e. according to this type structure not all DataStores are PersistenceBackedDataStores), I needed a value for `supportsSecondaryIndex` when the DataStore is theoretically not backed by a persistence.

But is there a real situation where a DataStore isn't backed by Persistence? That kind of confused me

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: Yea I could definitely do that - before I had `!db.supportsSecondaryIndex()` inlined everywhere and this was a step cleaner than that; what you suggested is even better xD

persistence-api/src/main/java/io/stargate/db/datastore/DataStore.java
dimas-b: In tests I did have `DataStore` without a `Persistence`, but in real implementations is it conceptually associated with `Persistence` I think.

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: There is one issue with this - technically a DataStore is only in scope after an auth token is processed during a request, but the DocumentService is a singleton that is created on service startup. So in practice, there's actually a new DataStore object on every single request, and that's why I didn't make it a field on the service.

There's definitely a way to not repeat myself so much here...

dimas-b (approved):

PR #502, opened 2020-12-06
Add support for AuthenticationStatement and AuthorizationStatement in new auth models

dougwettlaufer:


persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -244,69 +258,228 @@ private void authorizeByToken(ByteBuffer token, CQLStatement statement) {
           castStatement.keyspace(),
           castStatement.name());
     } else if (statement instanceof SchemaTransformation) {
-      SchemaTransformation castStatement = (SchemaTransformation) statement;
```
mpenick, line 247: On line 246:

```java
        authorization.authorizeSchemaWrite(
            authToken, castStatement.keyspace(), castStatement.name(), Scope.TRUNCATE);
```

Should truncate be considered at data write: `authorizeDataWrite()` like a delete?

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -260,57 +275,216 @@ private void authorizeByToken(ByteBuffer token, CQLStatement statement) {
...
-      } else if (statement instanceof AlterKeyspaceStatement) {
-        scope = Scope.ALTER;
-        keyspaceName = castStatement.keyspace();
-        tableName = null;
+      authorizeSchemaAlteringStatement(statement, authToken, authorization);
+    } else if (statement instanceof AuthorizationStatement) {
+      authorizeAuthorizationStatement(statement, authToken, authorization);
+    } else if (statement instanceof AuthenticationStatement) {
+      authorizeAuthenticationStatement(statement, authToken, authorization);
+    }
```
mpenick, line 283: This should have a final `else` and throw for unhandled statement types.

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dougwettlaufer: Good point, fixing

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dougwettlaufer: Probably a good idea, adding

dougwettlaufer: `java.lang.UnsupportedOperationException` sound right?

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -244,69 +258,228 @@ private void authorizeByToken(ByteBuffer token, CQLStatement statement) {
...
+      RoleResource roleResource = (RoleResource) f.get(stmt);
+
+      return roleResource != null ? roleResource.getName() : null;
+    } catch (Exception e) {
+      logger.error("Unable to get " + fieldName, e);
+      throw new RuntimeException("Unable to get private field", e);
     }
   }
 
+  private String getResourceFromStatement(PermissionsManagementStatement stmt) {
```
mpenick, line 406: It'd be nice if these reflection base methods were be added to `ReflectionUtils.java`. I'm wondering if we should statically initialize the `Field` too so it doesn't have to be done for every request.

dougwettlaufer: I'm not sure I follow about statically initializing `Field`

mpenick:
It looks like we're not authorizing/authenticating for `StargateQueryHandler.processBatch()` and I couldn't find an outstanding issue to fix it. It looks like `BatchStatement` contains a list of `ModificationStatement` statements that we could loop through.

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
mpenick: Never mind on this. We're getting the fields of types that aren't know at compile time.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
mpenick: yes, that seems fine.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthzJwtService.java
```diff
@@ -152,6 +152,64 @@ public void authorizeSchemaWrite(String token, String keyspace, String table, Sc
...
+  /**
+   * Authorization for role management is not provided by JWTs so all authorization will be deferred
+   * to the underlying permissions assigned to the role the JWT maps to.
+   *
+   * <p>{@inheritdoc}
+   */
+  @Override
+  public void authorizeRoleManagement(String token, String role, Scope scope)
+      throws UnauthorizedException {
+    // Cannot perform authorization with a JWT token so just return
```
tomekl007, line 164: maybe add `log.debug()`? It maybe useful for debugging auth

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java
```diff
@@ -72,7 +72,11 @@ public AuthResponse(byte[] token) {
       byte[] challenge = negotiator.evaluateResponse(token);
       if (negotiator.isComplete()) {
         AuthenticatedUser authenticatedUser = negotiator.getAuthenticatedUser();
-        persistenceConnection().login(authenticatedUser);
+        if (!(authenticatedUser.usesTransitionalAuth()
+            && Boolean.parseBoolean(
+                System.getProperty("stargate.cql_use_transitional_auth", "false")))) {
```
tomekl007, line 77: we should add some documentaiton around: `"stargate.cql_use_transitional_auth`

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java
```diff
@@ -72,7 +72,11 @@ public AuthResponse(byte[] token) {
       byte[] challenge = negotiator.evaluateResponse(token);
       if (negotiator.isComplete()) {
         AuthenticatedUser authenticatedUser = negotiator.getAuthenticatedUser();
-        persistenceConnection().login(authenticatedUser);
+        if (!(authenticatedUser.usesTransitionalAuth()
```
tomekl007, line 75: Why do we have such logic?
If the user is not `usesTransitionalAuth` and `cql_use_transitional_auth` set to `true`?
It looks a bit convoluted to me. Could you elaborate more?

persistence-api/src/main/java/io/stargate/db/AuthenticatedUser.java
```diff
@@ -26,11 +26,25 @@
   @Nullable
   String token();
 
+  boolean usesTransitionalAuth();
+
   static AuthenticatedUser of(String userName) {
-    return ImmutableAuthenticatedUser.builder().name(userName).build();
+    return ImmutableAuthenticatedUser.builder().name(userName).usesTransitionalAuth(false).build();
```
tomekl007, line 32: We are setting it to `false` for backward compatibility. Maybe extract it to a constant with `DEFAULT_USES_TRANSTIONAL_AUTH`?

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -207,14 +227,9 @@ private void authorizeByToken(ByteBuffer token, CQLStatement statement) {
...
-      Scope scope;
-      if (statement instanceof DeleteStatement) {
-        scope = Scope.DELETE;
-      } else {
-        scope = Scope.MODIFY;
-      }
-
+      authorizeModificationStatement(statement, authToken, authorization);
+    } else if (statement instanceof TruncateStatement) {
+      TruncateStatement castStatement = (TruncateStatement) statement;
```
tomekl007, line 232: Why do we need this special case? Previously, it was working for all scopes, and now we have a dedicated branch for `TRUNCATE`. Shouldn't both use the generic mechanism?

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -236,81 +251,275 @@ private void authorizeByToken(ByteBuffer token, CQLStatement statement) {
...
+      return roleResource != null ? roleResource.getName() : null;
+    } catch (Exception e) {
+      logger.error("Unable to get " + fieldName, e);
+      throw new RuntimeException("Unable to get private field", e);
+    }
+  }
+
+  private String getResourceFromStatement(PermissionsManagementStatement stmt) {
+    try {
+      Field f = stmt.getClass().getSuperclass().getDeclaredField("resource");
```
tomekl007, line 504: the `resource` is protected. Maybe instead of accessing it using reflection, we can create a class that extracts it?
We could create an `org.apache.cassandra.cql3.statements;` package in our project and extract it from there?

tomekl007 (dismissed):
This change is quite involved and captures requests on the hot-path.
Are there integration tests that are capturing those new cases? I don't see any tests in this PR.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -236,81 +251,275 @@ private void authorizeByToken(ByteBuffer token, CQLStatement statement) {
...
+  }
+
+  private void authorizeSchemaAlteringStatement(
+      CQLStatement statement, String authToken, AuthorizationService authorization) {
+    SchemaAlteringStatement castStatement = (SchemaAlteringStatement) statement;
+    Scope scope = null;
+    String keyspaceName = null;
+    String tableName = null;
+
+    if (statement instanceof CreateTableStatement) {
```
pcmanus, line 317: Afaict, this doesn't include all schema altering statements as it doesn't include any of the statements for 2ndary indexes, materialized views, UDT, functions/aggregate or triggers. For those, the `keyspaceName, `tableName` and `scope` will remain `null`. Assuming this won't just NPE (the javadoc of `authorizeSchemaWrite` doesn't make it explicit this has to be handled), are we sure this doesn't authorize the missed operations for everyone?

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthzJwtService.java
dimas-b: Should we deny access in this case?

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthzJwtService.java
```diff
@@ -152,6 +152,64 @@ public void authorizeSchemaWrite(String token, String keyspace, String table, Sc
     // Cannot perform authorization with a JWT token so just return
   }
 
+  /**
+   * Authorization for role management is not provided by JWTs so all authorization will be deferred
+   * to the underlying permissions assigned to the role the JWT maps to.
```
dimas-b, line 157: Does this hold for case when the `AuthenticatedUser` has `usesTransitionalAuth() == false`?

dougwettlaufer: Yes, it's when they're true that a different authz mechanism needs to be available.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthzJwtService.java
dougwettlaufer: Access here would depend on the permissions of the underlying role associated with the JWT.

dougwettlaufer: It's provided in the interface in the event you want to implement your own auth-provider that does handle authz like C* is doing.

dougwettlaufer:
> This change is quite involved and captures requests on the hot-path.
> Are there integration tests that are capturing those new cases? I don't see any tests in this PR.

Tests can be found here
https://github.com/stargate/stargate/blob/master/testing/src/test/java/io/stargate/it/cql/JwtAuthTest.java

and here
https://github.com/stargate/stargate/blob/master/testing/src/test/java/io/stargate/it/http/RestApiJWTAuthTest.java

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java
dougwettlaufer: `usesTransitionalAuth` will be set to true in the event that authentication is happening with something other than the traditional user/pass found in system_auth.roles but someone still might want to rely on C* permissions for authz. In the event that you're transitioning from purely C* authz to another system you can set `cql_use_transitional_auth` to true to say "if the authentication was with a different system AND transitional auth is supported then skip authenticating against C*"

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java
dougwettlaufer: We need to add consolidated documentation for all of these flags

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dougwettlaufer: There was always a dedicated branch for truncate, this diff just looks weird after I extracted a method for ModificationStatement

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dougwettlaufer: Seems like a lot of copy pasted code for a single field since we would have to copy pretty much every class that implements CQLStatement across C* 3.11, C* 4.0, and DSE 6.8. We can revisit later if this becomes and issue.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java
dimas-b: Could we also add an integration test for `stargate.cql_use_transitional_auth` == `true`?

authnz/src/main/java/io/stargate/auth/PlainTextTokenSaslNegotiator.java
```diff
@@ -66,7 +66,8 @@ public boolean isComplete() {
   @Override
   public AuthenticatedUser getAuthenticatedUser() throws AuthenticationException {
     if (storedCredentials != null) {
-      return AuthenticatedUser.of(storedCredentials.getRoleName(), storedCredentials.getPassword());
+      return AuthenticatedUser.of(
+          storedCredentials.getRoleName(), storedCredentials.getPassword(), true);
```
dimas-b, line 70: Do we ever use `AuthenticatedUser.of(..., ..., false)`? Would it make sense to move the reading of the `stargate.cql_use_transitional_auth` system property into this class (as a static constant, perhaps)?

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java
olim7t: I think what's confusing is that we use very similar names for two different things (see also [Dmitri's comment](https://github.com/stargate/stargate/pull/502/files#r538533657) above).

Maybe renaming the Java method to something like `authenticatedUser.isFromExternalAuth()` would help.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java
```diff
@@ -72,7 +72,11 @@ public AuthResponse(byte[] token) {
       byte[] challenge = negotiator.evaluateResponse(token);
       if (negotiator.isComplete()) {
         AuthenticatedUser authenticatedUser = negotiator.getAuthenticatedUser();
-        persistenceConnection().login(authenticatedUser);
+        if (!(authenticatedUser.isFromExternalAuth()
+            && Boolean.parseBoolean(
+                System.getProperty("stargate.cql_use_transitional_auth", "false")))) {
```
dimas-b, line 77: Do we still need to check `stargate.cql_use_transitional_auth` now that we have it in `PlainTextTokenSaslNegotiator`

dimas-b (approved):

mpenick (approved):

PR #503, opened 2020-12-07
Also support bearer token in swagger and graphql

dougwettlaufer:


mpenick (approved):
There's a bit of duplicated logic, but maybe it's not worth consolidating it.

PR #510, opened 2020-12-07
Fix search for tinyint bools

EricBorczuk:


dimas-b (approved):

persistence-api/src/main/java/io/stargate/db/Persistence.java
```diff
@@ -87,7 +87,7 @@
   boolean isInSchemaAgreement();
 
   default boolean supportsSecondaryIndex() {
-    return true;
+    return Boolean.parseBoolean(System.getProperty("persistence.support_2i", "true"));
```
dimas-b, line 90: `persistence.support_2i` -> `stargate.persistence.2i.support.default`?

testing/src/test/java/io/stargate/it/http/docsapi/DocumentApiV2TestDisable2i.java
```diff
@@ -0,0 +1,2808 @@
...
+import okhttp3.Response;
+import okhttp3.ResponseBody;
+import org.apache.http.HttpStatus;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+@NotThreadSafe
+@StargateSpec(parametersCustomizer = "disable2i")
+public class DocumentApiV2TestDisable2i extends BaseOsgiIntegrationTest {
```
dimas-b, line 48: We could keep actual test code in an abstract base class and have two concrete sub-classes one for default 2i support and the other using `parametersCustomizer`

PR #514, opened 2020-12-08
Fix: Use same ClientState when using external auth

mpenick:


persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -351,7 +352,13 @@ public Persistence persistence() {
     @Override
     protected void loginInternally(io.stargate.db.AuthenticatedUser user) {
       try {
-        clientState.login(new AuthenticatedUser(user.name()));
+        if (user.isFromExternalAuth()
+            || Boolean.parseBoolean(
+                System.getProperty("stargate.cql_use_transitional_auth", "false"))) {
+          clientState.login(AuthenticatedUser.ANONYMOUS_USER);
```
mpenick, line 358: I'm not `00% sure if this is the right move for the Cassandra backends.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -371,6 +377,8 @@ protected void loginInternally(io.stargate.db.AuthenticatedUser user) {
             new QueryState(
                 ClientState.forExternalCalls(AuthenticatedUser.ANONYMOUS_USER),
                 UserRolesAndPermissions.SYSTEM));
+      } else if (clientState.getUser().isAnonymous()) {
+        return Single.just(new QueryState(clientState, UserRolesAndPermissions.SYSTEM));
```
mpenick, line 381: This is a bit scary, but I'm not sure a way around it.

dougwettlaufer (approved):

PR #516, opened 2020-12-09
Make RestApiJWTAuthTest runnable with Java 11

dimas-b:
If the tests are running under Java 11, Instant objects
will uses nanosecond precision.

Unfortunately nanos are lost in the server side, so
this commit restricts the tests to only use millisecond
precision.

### DIFF
--- a/testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
+++ b/testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
@@ -22,6 +22,7 @@ import io.stargate.it.storage.StargateSpec;
 import io.stargate.it.storage.UseStargateContainer;
 import java.io.File;
 import java.net.URL;
+import java.time.Instant;
 import org.junit.jupiter.api.BeforeEach;
 
 /** This class manages starting Stargate OSGi containers. */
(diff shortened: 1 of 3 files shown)


testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
```diff
@@ -53,4 +54,9 @@ public boolean isCassandra4() {
     return !backend.isDse()
         && Version.parse(backend.clusterVersion()).nextStable().compareTo(Version.V4_0_0) >= 0;
   }
+
+  public static Instant now() {
+    // Avoid using Instants with nanosecond precision as nanos may be lost on the server side
```
tomekl007, line 59: In the past during Jepsen tests, there were problems causing by trimming the nano-seconds.
(https://aphyr.com/posts/294-jepsen-cassandra)
Are you sure that the nanoseconds are trimmed down?

dimas-b: This is not about C* internal time tracking. This is purely for user-level data.

Yes, I observed the precision loss when the test is executed by Java 11 but Stargate run in Java 8.

Please note, though, that we currently run tests using Java 8 in CI, in which case test data uses `Instant` objects with millisecond precision, AFAIK.

We can certainly open a follow-up issue for this.

tomekl007 (approved):

testing/src/test/java/io/stargate/it/BaseOsgiIntegrationTest.java
tomekl007: Ok. let's create a follow-up ticket.

dimas-b: #521

PR #517, opened 2020-12-09
REST: Revisit error messages when parsing values

olim7t:


dimas-b (approved):

dougwettlaufer:
Looks like a test is failing

```
Expecting throwable message:
  <"Invalid Inet value a: Name or service not known">
to contain:
  <"Invalid Inet value a: nodename nor servname provided, or not known">
but did not.

Throwable that failed the check:

java.lang.IllegalArgumentException: Invalid Inet value a: Name or service not known
	at io.stargate.web.resources.Converters.toCqlValue(Converters.java:593)
	at io.stargate.web.resources.Converters.toCqlValue(Converters.java:198)
```

restapi/src/main/java/io/stargate/web/resources/Converters.java
```diff
@@ -489,63 +512,105 @@ public static Object toCqlValue(Column.ColumnType type, String value) {
...
+        return new BigInteger(value);
+      }
+      if (rawType == Column.Type.Decimal) {
+        return new BigDecimal(value);
+      }
+    } catch (IllegalArgumentException e) {
+      // The exception's message doesn't really add information in any of those cases, use a generic
+      // description:
+      throw new IllegalArgumentException(
+          String.format("Invalid %s value '%s': cannot parse", rawType, value));
```
olim7t, line 586: In the original PR I avoided reflecting back the value because I was worried that would open the door to XSS attacks. But actually this is safe, we wrap the error in a JSON response in `RequestHandler`, so the message will be properly sanitized.

dougwettlaufer (approved):

PR #526, opened 2020-12-10
Refactor authn/z to pass around object rather than plain token string

dougwettlaufer:


authnz/src/main/java/io/stargate/auth/AuthenticationPrincipal.java
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.auth;
+
+public class AuthenticationPrincipal {
```
dimas-b, line 18: nit: looks like this is a simple value holder... might be worth converting to a `@Value.Immutable` interface to have all the setters and builders generated.

authnz/src/main/java/io/stargate/auth/AuthenticationPrincipal.java
```diff
@@ -0,0 +1,59 @@
...
+
+  private String token;
```
dimas-b, line 20: Since we're keeping credentials here, this object looks more like a "Subject" to me, cf. `javax.security.auth.Subject`. How about renaming to `AuthenticationSubject` or perhaps just to `Subject` since clashes with JRE are not likely in this context, I guess.

dimas-b: Also, `javax.security.auth.login.LoginContext` produces a `Subjet` on successful login.

authnz/src/main/java/io/stargate/auth/PlainTextTokenSaslNegotiator.java
```diff
@@ -35,7 +35,7 @@
 
   protected final AuthenticationService authentication;
   private final Authenticator.SaslNegotiator wrapped;
-  protected StoredCredentials storedCredentials;
+  protected AuthenticationPrincipal authenticationPrincipal;
   protected final String tokenUsername;
   protected final int tokenMaxLength;
   private static final boolean USE_TRANSITIONAL_AUTH =
```
dimas-b, line 41: `USE_TRANSITIONAL_AUTH` does not seem to be used any more.

dougwettlaufer: removed

authnz/src/main/java/io/stargate/auth/AuthenticationPrincipal.java
dougwettlaufer: Yeah that makes sense. Will rename to AuthenticationSubject

authnz/src/main/java/io/stargate/auth/AuthenticationSubject.java
```diff
@@ -0,0 +1,40 @@
...
+  static AuthenticationSubject of(String token, String roleName, boolean fromExternalAuth) {
+    return ImmutableAuthenticationSubject.builder()
+        .token(token)
+        .roleName(roleName)
+        .isFromExternalAuth(fromExternalAuth)
+        .build();
+  }
+
+  static AuthenticationSubject of(String token, String roleName) {
+    return ImmutableAuthenticationSubject.builder().token(token).roleName(roleName).build();
```
dimas-b, line 38: Should we set `isFromExternalAuth` too?... It does not appear to have a default values, so I'd guess the `build()` call will error out :thinking:

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Message.java
```diff
@@ -376,6 +380,26 @@ public void decode(ChannelHandlerContext ctx, Frame frame, List results) {
...
+      objectOutputStream.writeObject(authenticatedUser);
+      objectOutputStream.flush();
+      bytes = outputStream.toByteArray();
+    }
+
+    InputStream inputStream = new ByteArrayInputStream(bytes);
+    ByteBuffer byteBuffer = ByteBuffer.allocate(bytes.length);
+    while (inputStream.available() > 0) {
+      byteBuffer.put((byte) inputStream.read());
+    }
```
dimas-b, line 397: `ByteBuffer.wrap(bytes)`?

authnz/src/main/java/io/stargate/auth/AuthenticationSubject.java
dougwettlaufer: done

persistence-api/src/main/java/io/stargate/db/datastore/DataStoreFactory.java
```diff
@@ -35,6 +35,9 @@
    */
   DataStore create(@Nullable String userName, @Nonnull DataStoreOptions options);
 
+  DataStore create(
+      @Nullable String userName, boolean isFromExternalAuth, @Nonnull DataStoreOptions options);
```
dimas-b, line 39: Why not use `AuthenticationSubject` as the parameter? It looks like we need most of its properties anyway?

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -362,7 +362,7 @@ public Persistence persistence() {
     protected void loginInternally(io.stargate.db.AuthenticatedUser user) {
       try {
         if (user.isFromExternalAuth()
-            || Boolean.parseBoolean(
+            && Boolean.parseBoolean(
                 System.getProperty("stargate.cql_use_transitional_auth", "false"))) {
```
dimas-b, line 366: `Boolean.getBoolean("stargate.cql_use_transitional_auth")`?

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Message.java
dougwettlaufer: done

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -293,8 +306,42 @@ private void authorizeByToken(ByteBuffer token, CQLStatement statement) {
...
+    ObjectInput in = null;
+    try {
+      if (token.position() == token.limit()) {
+        token.flip();
+      }
+      byte[] bytes = new byte[token.remaining()];
+      token.get(bytes);
+      ByteArrayInputStream bis = new ByteArrayInputStream(bytes);
+      in = new ObjectInputStream(bis);
+      AuthenticatedUser authenticatedUser = (AuthenticatedUser) in.readObject();
```
dimas-b, line 320: Java serialization can be slow (or used to be?).. since this in on the critical CQL query path, would it be worth using a custom serialization for the properties of `AuthenticatedUser`?... a bit more work, but hopefully more performant... WDYT?

persistence-api/src/main/java/io/stargate/db/datastore/DataStoreFactory.java
dougwettlaufer: Trying to avoid a circular dependency between authnz and persistence-api

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Message.java
```diff
@@ -340,20 +342,20 @@ public void decode(ChannelHandlerContext ctx, Frame frame, List results) {
...
+              && ((ServerConnection) connection).clientInfo().getAuthenticatedUser() != null) {
             if (customPayload != null) {
               message
                   .getCustomPayload()
                   .put(
                       "token",
-                      StandardCharsets.UTF_8.encode(
-                          ((ServerConnection) connection).clientInfo().getAuthToken()));
+                      authenticatedUserToByteBuffer(
+                          ((ServerConnection) connection).clientInfo().getAuthenticatedUser()));
```
dimas-b, line 352: Could we cache this in `ClientInfo` to avoid re-serialization on every message?

dougwettlaufer: Oh, good call.

dimas-b (commented):
LGTM overall :+1:  comments above.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dougwettlaufer: Not sure I follow regarding the custom serialization.

persistence-api/src/main/java/io/stargate/db/datastore/DataStoreFactory.java
dimas-b: ah, makes sense

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dimas-b: I mean constructing a custom byte array from the properties of `AuthenticatedUser` instead of using `ObjectOutputStream` / `ObjectInputStream`.

dimas-b: here, specifically, we'd do something like
```
int size = token.getInt()
byte[] chars = new byte[size];
token.get(chars);
String roleName = new String(chars, UTF8);
```

dimas-b: for that matter it might be more convenience to put individual `AuthenticatedUser` as separate `ByteBuffers` into the custom payload... makes the deserialization logic easier, I guess.

persistence-api/src/main/java/io/stargate/db/ClientInfo.java
```diff
@@ -41,12 +45,34 @@ public void registerDriverInfo(DriverInfo info) {
...
+    this.token =
+        authenticatedUser.token() != null
+            ? ByteBuffer.wrap(authenticatedUser.token().getBytes(StandardCharsets.UTF_8))
+            : null;
+    this.roleName = ByteBuffer.wrap(authenticatedUser.name().getBytes(StandardCharsets.UTF_8));
+    this.isFromExternalAuth =
+        ByteBuffer.wrap(
+            (authenticatedUser.isFromExternalAuth()
+                ? new byte[] {(byte) 1}
+                : new byte[] {(byte) 0}));
```
dimas-b, line 63: nit: Since we're are the only consumer of these buffers, we could probably use `null` to indicate `false` and a static constant buffer to indicate `true`.

dimas-b (approved):

persistence-api/src/main/java/io/stargate/db/ClientInfo.java
dougwettlaufer: True, that cleans it up a little

dimas-b (approved):

PR #528, opened 2020-12-11
Declare support for SAI in Persistence

dimas-b:
Avoid relying on the Persistence name to
determine whether SAI is supported.

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
@@ -95,6 +95,10 @@ public class DocumentDB {
     this.dataStore = dataStore;
     this.authToken = authToken;
     this.authorizationService = authorizationService;
+
+    if (!dataStore.supportsSAI() && !dataStore.supportsSecondaryIndex()) {
+      throw new IllegalStateException("Backend does not support any known index types.");
+    }
   }
 
   public AuthorizationService getAuthorizationService() {
@@ -194,9 +198,9 @@ public class DocumentDB {
     }
   }
 
-  public boolean maybeCreateTableIndexes(String keyspaceName, String tableName, boolean isDse) {
+  public boolean maybeCreateTableIndexes(String keyspaceName, String tableName) {
     try {
-      if (isDse) {
+      if (dataStore.supportsSAI()) {
         createSAIIndexes(keyspaceName, tableName);
       } else {
         createDefaultIndexes(keyspaceName, tableName);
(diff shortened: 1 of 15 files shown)


restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -95,6 +95,10 @@ public DocumentDB(
     this.dataStore = dataStore;
     this.authToken = authToken;
     this.authorizationService = authorizationService;
+
+    if (!dataStore.supportsSAI() && !dataStore.supportsSecondaryIndex()) {
+      throw new IllegalStateException("Backend does not support any known index types.");
```
tomekl007, line 100: It seems that in some places we just log that SAI is not supported (https://github.com/stargate/stargate/pull/528/files#diff-d36bfbe0c5f35c2d92be42df6186c221c2a80c49022c728aae5ccd217443be3aR226) but here we are throwing an error. Is it must have to have one of those indexes support?

restapi/src/test/java/io/stargate/web/docsapi/dao/DocumentDBTest.java
```diff
@@ -336,6 +336,11 @@ public boolean supportsSecondaryIndex() {
       return true;
     }
 
+    @Override
+    public boolean supportsSAI() {
+      return true;
```
tomekl007, line 341: can we create a test that validates both SAI `true` and `false`?

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
dimas-b: I believe so. When `DocumentDB` [creates indexes](https://github.com/stargate/stargate/pull/528/files#diff-d36bfbe0c5f35c2d92be42df6186c221c2a80c49022c728aae5ccd217443be3aR203-R206) it has only two options: SAI and "default". The "default" in this contex may sound a bit confusion, but really it mean "secondary" IIRC.

@EricBorczuk : WDYT?

restapi/src/test/java/io/stargate/web/docsapi/dao/DocumentDBTest.java
dimas-b: SAI `true` -> DSE integration tests, SAI 'false` -> C* 3.11 and 4.0 integration tests.

dimas-b: I actually updated `DocumentApiV2TestDisable2i` (and force-pushed) because it failed :)

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
dougwettlaufer: Right, it has to support at least one of them for docsapi. If SAI isn't supported would be good to just fallback to secondary.

dimas-b: That's what Docs API does (fallback). This check is just to ensure we have at least one of the index kinds available to us.

dimas-b:
> something failing.

Socket timeouts :face_with_head_bandage:

dougwettlaufer (approved):

Issue #529, opened 2020-12-11
Put `AuthResponse` validation on a separate thread pool


mpenick:
In C* (and DSE) the request logic normally runs on a separate threads from the event loop threads. However, the logic for `AuthResponse` is run directly on event loop threads which is unlike other requests that run on `Persistence` backend threads. This was done so that custom authentication logic could be added for Stargate. The problem with running this logic on event loop threads is `PasswordAuthenticator` runs `BCrypt.checkpw(password, hash)` which is quite CPU intensive (intentionally) and causes significant delays to other requests. 

To avoid delays on the event loop threads we should move processing `AuthResponse` requests to another thread pool.

PR #530, opened 2020-12-11
Fix: Move CQL AuthResponse request handling to request threads

mpenick:
This moves AuthResponse request handling off of the event loop threads
to the appropriate request thread pool.

In C* (and DSE) the request logic normally runs on a separate threads
from the event loop threads. However, in Stargate, the logic for
AuthResponse is run directly on event loop threads which is unlike other
requests that run on Persistence backend threads. This was done so that
custom authentication logic could be added for Stargate. The problem
with running this logic on event loop threads is PasswordAuthenticator
runs BCrypt.checkpw(password, hash) which is quite CPU intensive
(intentionally) and causes significant delays to other requests.

### DIFF
--- a/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java
+++ b/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java
@@ -91,9 +91,11 @@ public class AuthResponse extends Message.Request {
                 } else {
                   future.complete(new AuthChallenge(challenge));
                 }
-              } catch (AuthenticationException e) {
+              } catch (AuthenticationException ae) {
                 ClientMetrics.instance.markAuthFailure();
-                future.complete(ErrorMessage.fromException(e));
+                future.complete(ErrorMessage.fromException(ae));
+              } catch (Exception e) {
+                future.completeExceptionally(e);
               }
             });
     return future;
--- a/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java
+++ b/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java
@@ -67,25 +67,35 @@ public class AuthResponse extends Message.Request {
 
   @Override
   protected CompletableFuture<? extends Response> execute(long queryStartNanoTime) {
-    try {
-      Authenticator.SaslNegotiator negotiator = ((ServerConnection) connection).getSaslNegotiator();
-      byte[] challenge = negotiator.evaluateResponse(token);
-      if (negotiator.isComplete()) {
-        AuthenticatedUser authenticatedUser = negotiator.getAuthenticatedUser();
-        persistenceConnection().login(authenticatedUser);
-        if (authenticatedUser.token() != null) {
-          ((ServerConnection) connection).clientInfo().setAuthToken(authenticatedUser.token());
-        }
+    CompletableFuture<Response> future = new CompletableFuture<>();
+    persistence()
+        .executeAuthResponse(
+            () -> {
+              try {
+                Authenticator.SaslNegotiator negotiator =
+                    ((ServerConnection) connection).getSaslNegotiator();
+                byte[] challenge = negotiator.evaluateResponse(token);
+                if (negotiator.isComplete()) {
+                  AuthenticatedUser authenticatedUser = negotiator.getAuthenticatedUser();
+                  persistenceConnection().login(authenticatedUser);
+                  if (authenticatedUser.token() != null) {
+                    ((ServerConnection) connection)
+                        .clientInfo()
+                        .setAuthToken(authenticatedUser.token());
+                  }
 
-        ClientMetrics.instance.markAuthSuccess();
-        // authentication is complete, send a ready message to the client
-        return CompletableFuture.completedFuture(new AuthSuccess(challenge));
-      } else {
-        return CompletableFuture.completedFuture(new AuthChallenge(challenge));
-      }
-    } catch (AuthenticationException e) {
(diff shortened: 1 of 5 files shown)


mpenick:
Fixes https://github.com/stargate/stargate/issues/529

dimas-b (approved):

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -281,6 +281,11 @@ public boolean isInSchemaAgreement() {
         .build();
   }
 
+  @Override
+  public void executeAuthResponse(Runnable handler) {
+    executor.execute(handler);
```
tomekl007, line 286: It looks like you are using this executor:
```
   executor =
        SHARED.newExecutor(
            DatabaseDescriptor.getNativeTransportMaxThreads(),
            Integer.MAX_VALUE,
            "transport",
            "Native-Transport-Requests");
```
responsible for transport. Is it a good one?
Shouldn't we consider creating a separate executor service responsible for processing auth requests and use It here?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -279,6 +282,12 @@ public boolean isInSchemaAgreement() {
         .build();
   }
 
+  @Override
+  public void executeAuthResponse(Runnable handler) {
```
tomekl007, line 286: I think that we could achieve a cleaner and more caller friendly API by returning:
`CompletableFuture<Void>` from this method. If we do that, the calling code doesn't need to create a `CompletableFuture` itself and a callback that fulfills the CompletableFuture. wdyt?

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java
```diff
@@ -67,25 +67,35 @@ public AuthResponse(byte[] token) {
...
-      }
-    } catch (AuthenticationException e) {
-      ClientMetrics.instance.markAuthFailure();
-      return CompletableFuture.completedFuture(ErrorMessage.fromException(e));
-    }
+                  future.complete(new AuthSuccess(challenge));
+                } else {
+                  future.complete(new AuthChallenge(challenge));
+                }
+              } catch (AuthenticationException e) {
```
tomekl007, line 94: should the returned `CompletableFuture` wrap all possible exceptions? If there will be a different kind of exception, it will be propagated to the caller now. I think that when returning the `CompletableFuture` it should encapsulate all kinds of exceptions, so I suggest to change this code to:
```
  } catch (Exception e) {
  // ...
  }
```

tomekl007 (changes requested):

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java
dimas-b: Good point. I missed that :facepalm:

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
mpenick: This is the same thread pool used by C* implementations.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java
mpenick: This is already handle at the `.execute()` call site: https://github.com/stargate/stargate/blob/v1.0.0/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Message.java#L715-L717

The `} catch (AuthenticationException e) {` is to specifically handle and mark failed authentication attempts.

dimas-b: I'm not sure... if an unexpected exception occurs within the task submitted to `persistence().executeAuthResponse(...)` the future retuned from this method will never complete, will it?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
mpenick: I might be misunderstanding what you're suggesting.

`AuthResponse.execute()` expects a return type of `CompletableFuture<? extends Response>` so we couldn't return the `CompletableFuture<Void>` directly so it'd have to call `thenApply()` or `thenApplyAsync()`. The problem with `thenApply()` is it could be called synchronously on the thread calling `thenApply()` if the future has already been completed. `thenApplyAsync()` can't be called because we con't have access to the request executor.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java
mpenick: yeah, you're right.

mpenick: Added fix: https://github.com/stargate/stargate/pull/530/commits/bab6c86b912184c046ff6580dd306207a02a9dae

mpenick: Good catch. Thanks.

tomekl007 (approved):

PR #531, opened 2020-12-11
Re-include Docs API unit tests (fixes #527)

### LINKED ISSUE #527: DocumentDBTest does not run in CI
`DocumentDBTest` uses JUnit _4_ annotations, therefore it is not executed during CI.

`DocumentDBTest` should be migrated to JUnit 5

dimas-b:
* Convert Docs API tests to JUnit 5

* Remove dependency on PowerMock as it does not appear to be usable under JUnit 5

* Exclude transitive dependencies on JUnit4 to make sure tests do not use it accidentally

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
@@ -3,6 +3,7 @@ package io.stargate.web.docsapi.resources;
 import com.datastax.oss.driver.api.core.NoNodeAvailableException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.annotations.VisibleForTesting;
 import io.stargate.auth.UnauthorizedException;
 import io.stargate.web.docsapi.dao.DocumentDB;
 import io.stargate.web.docsapi.examples.WriteDocResponse;
@@ -59,8 +60,18 @@ public class DocumentResourceV2 {
   @Inject private Db dbFactory;
   private static final Logger logger = LoggerFactory.getLogger(DocumentResourceV2.class);
   private static final ObjectMapper mapper = new ObjectMapper();
-  private final DocumentService documentService = new DocumentService();
+  private final DocumentService documentService;
   private final int DEFAULT_PAGE_SIZE = 100;
+
+  public DocumentResourceV2() {
+    documentService = new DocumentService();
+  }
+
+  @VisibleForTesting
+  DocumentResourceV2(Db dbFactory, DocumentService documentService) {
+    this.dbFactory = dbFactory;
+    this.documentService = documentService;
+  }
 
   @POST
   @ManagedAsync
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
@@ -9,6 +9,7 @@ import com.fasterxml.jackson.databind.node.LongNode;
 import com.fasterxml.jackson.databind.node.NullNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.node.TextNode;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.gson.JsonArray;
@@ -745,7 +746,8 @@ public class DocumentService {
     return ImmutablePair.of(docsResult, pageState);
   }
 
-  private void addRowsToMap(Map<String, List<Row>> rowsByDoc, List<Row> rows) {
+  @VisibleForTesting
+  void addRowsToMap(Map<String, List<Row>> rowsByDoc, List<Row> rows) {
     for (Row row : rows) {
       String key = row.getString("key");
       List<Row> rowsAtKey = rowsByDoc.getOrDefault(key, new ArrayList<>());
@@ -969,7 +971,8 @@ public class DocumentService {
    * @throws ExecutionException
    * @throws InterruptedException
    */
-  private ImmutablePair<List<Row>, ByteBuffer> searchRows(
+  @VisibleForTesting
+  ImmutablePair<List<Row>, ByteBuffer> searchRows(
       String keyspace,
       String collection,
       DocumentDB db,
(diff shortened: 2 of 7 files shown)


restapi/src/test/java/io/stargate/web/docsapi/dao/DocumentDBTest.java
```diff
@@ -68,12 +92,22 @@ public void containsIllegalChars() {
...
+    int idx = 0;
+    values[idx++] = "key";
+    for (int i = 0; i < DocumentDB.MAX_DEPTH; i++) {
+      values[idx++] = "value" + i;
+    }
+    values[idx++] = "leaf";
+    values[idx++] = "text";
+    values[idx++] = 1.2d; // dbl_value
+    values[idx++] = true; // bool_value
+    assertThat(idx).isEqualTo(values.length);
```
tomekl007, line 105: This assertion is weird to me. It seems that we are testing the test init code, is it?
I that it will be easier if we would create a `Map<String, Object>` and use the names as keys:
```
map.put("key", "key");
// ...
map.put("dbl_value", 1.2d);
// ...
```
wdyt?

restapi/src/test/java/io/stargate/web/docsapi/dao/DocumentDBTest.java
```diff
@@ -173,24 +209,26 @@ public void deleteThenInsertBatch() throws UnauthorizedException {
 
     assertThat(generatedQueries.get(0).queryString())
         .isEqualTo(
-            "DELETE FROM \"keyspace\".\"table\" USING TIMESTAMP ? WHERE key = :key AND p0 = :p0 AND p1 = :p1 AND p2 = :p2");
+            "DELETE FROM \"keyspace\".\"table\" USING TIMESTAMP ? WHERE key = ? AND p0 = ? AND p1 = ? AND p2 = ?");
```
tomekl007, line 212: bind parameters(`:smth`) are more explicit than wildcard markers(`?`). Why we are changing this?

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -59,9 +60,19 @@
...
   private static final ObjectMapper mapper = new ObjectMapper();
-  private final DocumentService documentService = new DocumentService();
+  private final DocumentService documentService;
   private final int DEFAULT_PAGE_SIZE = 100;
 
+  public DocumentResourceV2() {
+    documentService = new DocumentService();
+  }
+
+  @VisibleForTesting
```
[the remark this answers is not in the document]
dimas-b: I meant this annotation to indicate that the constructor should not be used except by test code, but I can remove if it's a distraction.

restapi/src/test/java/io/stargate/web/docsapi/dao/DocumentDBTest.java
```diff
@@ -32,17 +36,37 @@
...
 public class DocumentDBTest {
+  private static final Schema schema = buildSchema();
+
   private DocumentDB documentDB;
   private TestDataStore ds;
   private static final ObjectMapper mapper = new ObjectMapper();
 
-  @Before
+  private static Schema buildSchema() {
+    SchemaBuilder__5 schemaBuilder =
```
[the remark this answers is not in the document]
dimas-b: This is an unfortunate side effect of the way [duzzt](https://github.com/misberner/duzzt) generates code. We use it to generate the schema builder. Normally these class names do not surface, but here we use it in a loop, as I wanted to avoid copying the same line 64 times :)

restapi/src/test/java/io/stargate/web/docsapi/dao/DocumentDBTest.java
dimas-b: because the way CQL text is generated has changed. We do not directly control the CQL, it is produced by the query build and it now uses `?` for parameters.

restapi/src/test/java/io/stargate/web/docsapi/dao/DocumentDBTest.java
dimas-b: We are asserting that we have values for all all columns (`DocumentDB.allColumns()`). We make some values in a loop and some by hand-written lines, so this assertion is to protect against human errors when editing this code.

`documentDB.getInsertStatement` requires an array of parameters. So I make the array. We can certainly use a map, but to me it seems that the extra changes to handle map in this context may be an overkill.

dougwettlaufer (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -969,7 +971,8 @@ private void updateExistenceForMap(
    * @throws ExecutionException
    * @throws InterruptedException
    */
-  private ImmutablePair<List<Row>, ByteBuffer> searchRows(
+  @VisibleForTesting
```
tomekl007, line 974: this change made the codancy fail with the error:
`Avoid really long methods.`

dimas-b: I did not change the method body, just made it more accessible so that tests could use simple mocks for this method (previously PowerMock was required, but it does not work well with JUnit 5).

Perhaps Codacity flags this error because the method declaration changed... :thinking: Is it worth fixing at this point? The purpose of this PR was mainly to re-enable the unit tests that we not being executed. I wanted to minimize changes to production code in this PR. WDYT?

dimas-b: This problem is already reported on `master` so this PR do not make the situation worse :) https://app.codacy.com/gh/stargate/stargate/file/51191973112/issues/source?bid=21416080&fileBranchId=21416080

I had a go at factoring some of the code out, but it is not trivial as there pieces of the code inside this method are interdependent logically. So in effect this method is indeed complex and some simplification would be good, but at the same time I'd like to avoid scope creep on this PR. I'm ok with opening an follow-up issue for this though. WDYT?

tomekl007: good idea with the follow-up ticket.

tomekl007 (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
dimas-b: #537

PR #534, opened 2020-12-15
Docs API changes to improve performance

EricBorczuk:
A few changes that improve performance for the docs API.

Summary of changes:
1) validating columns for document API queries gets skipped (since we control the schema, the change is [here](https://github.com/stargate/stargate/pull/534/files#diff-b8c255d9e805201cb8ed7ab124b0de34256e996dbc41ac155ac99dab4e6f7ae5R1391)
2) allColumns() doesn't have to be calculated each time the function is called, fix is [here](https://github.com/stargate/stargate/pull/534/files#diff-d36bfbe0c5f35c2d92be42df6186c221c2a80c49022c728aae5ccd217443be3aR46)
3) Caching of the _entire_ data store is now done.  This was achieved by setting pageSize/pageState values on each individual query, rather than on the data store itself, to allow proper reuse. See [here](https://github.com/stargate/stargate/pull/534/files#diff-d36bfbe0c5f35c2d92be42df6186c221c2a80c49022c728aae5ccd217443be3aR336) for an example of that querying change.

These changes all together cause latency on a setup with 2 stargate/2 storage nodes, 4000 ops/sec to drop from 40ms per read to 20ms per read. Not bad!

### DIFF
--- a/persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
+++ b/persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
@@ -1089,11 +1089,7 @@ public class QueryBuilderImpl {
   }
 
   private static List<Column> convertToColumns(Table table, Collection<String> columnNames) {
-    List<Column> columns = new ArrayList<>(columnNames.size());
-    for (String name : columnNames) {
-      columns.add(table.column(name));
-    }
-    return columns;
+    return columnNames.stream().map(table::column).collect(toList());
   }
 
   private Set<String> names(Collection<Column> columns) {
--- a/persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
+++ b/persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
@@ -1088,6 +1088,14 @@ public class QueryBuilderImpl {
     return columnNames.stream().map(table::existingColumn).collect(toList());
   }
 
+  private static List<Column> convertToColumns(Table table, Collection<String> columnNames) {
+    List<Column> columns = new ArrayList<>(columnNames.size());
+    for (String name : columnNames) {
+      columns.add(table.column(name));
+    }
+    return columns;
+  }
+
   private Set<String> names(Collection<Column> columns) {
     return columns.stream().map(SchemaEntity::name).collect(toSet());
   }
@@ -1380,13 +1388,13 @@ public class QueryBuilderImpl {
   protected BuiltSelect selectQuery() {
     Table table = schemaTable();
     QueryStringBuilder builder = new QueryStringBuilder(markerIndex);
-    List<Column> selectedColumns = validateColumns(table, selection);
+    List<Column> selectedColumns = convertToColumns(table, selection);
     // Using a linked set for the minor convenience of get back the columns in the order they were
     // passed to the builder "in general".
     Set<Column> allSelected = new LinkedHashSet<>(selectedColumns);
     Column wtColumn = null;
     if (writeTimeColumn != null) {
-      wtColumn = table.existingColumn(writeTimeColumn);
+      wtColumn = table.column(writeTimeColumn);
       allSelected.add(wtColumn);
     }
     builder.append("SELECT");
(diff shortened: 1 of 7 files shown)


persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
```diff
@@ -1088,6 +1088,14 @@ private boolean indexExists(Keyspace keyspace, String indexName) {
     return columnNames.stream().map(table::existingColumn).collect(toList());
   }
 
+  private static List<Column> convertToColumns(Table table, Collection<String> columnNames) {
+    List<Column> columns = new ArrayList<>(columnNames.size());
+    for (String name : columnNames) {
+      columns.add(table.column(name));
+    }
+    return columns;
```
eribeiro, line 1096:
```suggestion
    return columnNames.stream().map(table::column).collect(toList());
```

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -64,26 +69,33 @@
...
     for (int i = 0; i < MAX_DEPTH; i++) {
       allPathColumnNames.add("p" + i);
       allPathColumnTypes.add(Type.Text);
+      allColumns.add(Column.create("p" + i, Type.Text));
     }
     allColumnNames.addAll(allPathColumnNames);
     allColumnTypes.addAll(allPathColumnTypes);
     allColumnNames.add("leaf");
     allColumnTypes.add(Type.Text);
+    allColumns.add(Column.create("leaf", Type.Text));
```
tomekl007, line 89: you may change the ordering of opeartions to:
```
Column leafColumn = Column.create("leaf", Type.Text);
allColumns.add(leafColumn);
allColumnNames.add(leafColumn.name());
allColumnTypes.add(leafColumn.type());
```
to reduce the duplication a bit.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -322,6 +333,29 @@ public ResultSet executeSelect(
...
   }
 
+  public ResultSet executeSelect(
+      String keyspace,
+      String collection,
+      List<BuiltCondition> predicates,
+      int pageSize,
+      ByteBuffer pageState)
+      throws UnauthorizedException {
+    // Run generic authorizeDataRead for now
```
tomekl007, line 343: could you elaborate more on that comment?

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -322,6 +333,29 @@ public ResultSet executeSelect(
...
+    getAuthorizationService()
+        .authorizeDataRead(getAuthenticationSubject(), keyspace, collection, SourceAPI.REST);
+    UnaryOperator<Parameters> parametersModifier =
+        p -> ImmutableParameters.builder().pageSize(pageSize).pagingState(pageState).build();
+    return this.builder()
+        .select()
+        .column(DocumentDB.allColumns())
+        .writeTimeColumn("leaf")
```
tomekl007, line 351: As far as I can see, the leaf column type is text. Why the `writeTime` is text?

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
dougwettlaufer: For now we'll just use authz that tells us whether the user has access to the table but in the future we should add a more granular authz method that's able to work with documents.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
dougwettlaufer: `writeTime` is a long. This query is pulling the writeTime of the leaf column in order to return the most recent version I believe.

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -975,6 +999,8 @@ private void updateExistenceForMap(
    * @param path the path in the document that is being searched on
    * @param recurse legacy boolean, only used for v1 of the API
    * @param documentKey filter down to only one document's results
+   * @param pageSize Cassandra page size
+   * @param pageState Cassandra page state
```
dougwettlaufer, line 1003: Suggestion

```suggestion
   * @param pageSize number of rows to return 
   * @param pageState current state of database paging
```

dougwettlaufer (approved):

PR #538, opened 2020-12-16
Support service access permissions

tomekl007:


dougwettlaufer (approved):

authnz/src/main/java/io/stargate/auth/AuthorizationService.java
```diff
@@ -51,9 +53,11 @@ ResultSet authorizedDataRead(
    * @param token The authenticated token to use for authorization.
    * @param keyspace The keyspace containing the table with data to be read.
    * @param table The table within the provided keyspace containing the data to be read.
+   * @param sourceAPI The source api which calls this method.
    * @throws UnauthorizedException An exception relating to the failure to authorize.
    */
-  void authorizeDataRead(String token, String keyspace, String table) throws UnauthorizedException;
+  void authorizeDataRead(String token, String keyspace, String table, SourceAPI sourceAPI)
+      throws UnauthorizedException;
```
dimas-b, line 60: Sorry for the late comment, but having a fixed enum for `sourceAPI` may be inconvenient to API extensions that are not part of this repo... :thinking:

tomekl007: What do you mean? The c2 is importing the stargate code, and this enum will be visible. The same pattern is implemented for `Scope` - this is an enum as well.

dimas-b: I mean if somebody implements a different "API" on top of Stargate in another repository, they will not be able to construct the `sourceAPI` parameter for their API without first making changes to Stargate core, which would cause a dependency cycle.

tomekl007: Good point, but don't you think that having type-safe API instead of string API is worth that price? 
They can firstly add the new API to the `SourceAPI` enum without using it. Next, they can submit a new module that leverages the new `SourceAPI` value.

dimas-b: We could use something other than a fixed `enum` to identify API types. We could use the OSGi bundle ID for that, as an example, but there are many options, I guess. If we want to be able to list all possible values, we could have a service where API ID would have to be registered before they are used for authorization purposes.

tomekl007: I think that the use case that we discussing now is hypothetical. Let's get back to it if it turns out that we have such a use-case and it is problematic, ok?

PR #504, opened 2020-12-07
[ON HOLD] Capture BoundDMLQuery in the DataStore calls 

tomekl007:
it solves:
- Capture BoundDMLQuery in the DataStore calls https://github.com/stargate/stargate/issues/491 
- Create cdc_events table https://github.com/stargate/stargate/issues/458 
- Serialization of CDC event https://github.com/stargate/stargate/issues/460

PR #540, opened 2020-12-16
Data store as a service

tomekl007:


dougwettlaufer (commented):
What's the reason for this change?

persistence-api/src/main/java/io/stargate/db/DbActivator.java
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  @Override
+  protected void stopService() {
+    // no-op
+  }
+
+  @Override
+  protected List<ServicePointer<?>> dependencies() {
+    return Collections.singletonList(persistence);
```
dimas-b, line 28: Why does this activator have to depend on `Persistence` if it does not use the reference to the service apparently?

persistence-api/src/main/java/io/stargate/db/DbActivator.java
```diff
@@ -0,0 +1,30 @@
...
+  private final ServicePointer<Persistence> persistence =
+      BaseActivator.ServicePointer.create(Persistence.class);
+
+  public DbActivator() {
+    super("DB services");
+  }
+
+  @Override
+  protected ServiceAndProperties createService() {
+    return new ServiceAndProperties(new DataStoreFactory(), DataStoreFactory.class);
```
dimas-b, line 18: What is the rationale for registering `DataStoreFactory` as a service? The `DataStoreFactory` class appears to be self-sufficient. It is a concrete class without any configuration.... Bundles that use it should be able to create it whenever necessary without depending on other bundles :thinking:

tomekl007: The intention is to have one common place where it is created - a singleton. Then, this one instance is injected wherever it is needed instead of using the static method and static init.

persistence-api/src/main/java/io/stargate/db/DbActivator.java
tomekl007: That was a left-over. Good catch removed

persistence-api/src/main/java/io/stargate/db/DbActivator.java
dimas-b: Sorry, but I do not see any benefits of this. `DataStores` can be created from `Persistence` even now and without any static init AFAIK. This PR does not simplify the creation process, it seems - users still have to provide the same amount of information as before.

tomekl007:
> What's the reason for this change?

I implemented this for CDC write path where two `DataStores` types can be crated (and injected):
https://github.com/stargate/stargate/pull/504
the work was already done and I think that this is a good direction to remove static initializers and have an injection of service instead. It improves the testability of our code-base. I'm open to discussion whether you think it is needed. 

The second argument for this change is that this refactor touches multiple places, and there is a high risk that the CDC PR will diverge a lot from master: https://github.com/stargate/stargate/pull/504 because the CDC work is paused now. 
It would be better to have this in the main branch, when(if) the CDC work will be resumed.

persistence-api/src/main/java/io/stargate/db/DbActivator.java
tomekl007: > DataStores can be created from Persistence even now and without any static init AFAIK.
What do you mean? All places where DataStore is created are using `create()` method that is static.

dimas-b: `create()` is indeed static, but in my mind it does not "initialize" anything, it's just a convenience method for creating a `DataStore`.

I did not see that other comment when I posted my comment avove :)

Having multiple `DataStore` implementations was not apparent from the PR :)

If we want to centralize `DataStore` creation, I'd propose defining an `interface` for that service rather than using a concrete class for that purpose.

Also since Stargate kind of assumes that only one `Persistence` implementation is active at any time. I believe it would be worth abstracting the dependency on `Persistence` behind the factory interface, so that callers could depend just on the facotory service. WDYT?

tomekl007: Do you mean injecting the `Persistence` into this `DataStoreFactory` and remove the `Persistence` from all of it's create methods? That could be indeed a good improvement. However, from what I see, the `Persistence` is also used directly - no via the `DataStore` interface. Because of that, the `Persistence` cannot be hidden entirely from the callers. In the context of `DataStore` it can be hidden -  f that is enough for you, I may try to do it.

dimas-b: Yes, that's what I mean.

I suppose not all `DataStore` users need to access `Persistence` directly... e.g. in GraphQL API it looks like in many places where `Persistence` is used, it is used only to create `DataStore`... I did not check all of the usages, but there's certainly a pattern there.

tomekl007: I've moved the `Persistence` into `DataStore`. It looks better now. Please take a look.

persistence-api/src/main/java/io/stargate/db/datastore/DataStoreFactory.java
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * Creates a new DataStore using the provided connection for querying and with the provided
+   * default parameters.
+   *
+   * @param connection the persistence connection to use for querying.
+   * @param options the options for the create data store.
+   * @return the created store.
+   */
+  public DataStore create(Persistence.Connection connection, @Nonnull DataStoreOptions options) {
```
dimas-b, line 47: Now that `Persistence` is encapsulated in `DataStoreFactory`, exposing a `public` method that takes a `Persistence.Connection` parameter seems contradictory to me.

persistence-api/src/main/java/io/stargate/db/datastore/DataStoreFactory.java
```diff
@@ -0,0 +1,126 @@
...
+import com.datastax.oss.driver.shaded.guava.common.base.Strings;
+import io.stargate.db.AuthenticatedUser;
+import io.stargate.db.ClientInfo;
+import io.stargate.db.Persistence;
+import java.net.InetSocketAddress;
+import java.util.Objects;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class DataStoreFactory {
```
dimas-b, line 27: I'm still not sure about exposing concrete classes as OSGi services. What do you think about adding an interface so that caller from other bundles would not 'see' the impl. class?

persistence-api/src/main/java/io/stargate/db/datastore/DataStoreFactory.java
tomekl007: Good point, I've made it private.

persistence-api/src/main/java/io/stargate/db/datastore/DataStoreFactory.java
tomekl007: good idea, done.

persistence-api/src/main/java/io/stargate/db/datastore/DefaultDataStoreFactory.java
```diff
@@ -0,0 +1,95 @@
...
+   * @return the created store.
+   */
+  private DataStore create(
+      @Nullable String userName,
+      @Nonnull DataStoreOptions options,
+      @Nullable ClientInfo clientInfo) {
+    Persistence.Connection connection = persistence.newConnection();
+    if (clientInfo != null) {
+      connection = persistence.newConnection(clientInfo);
+    }
```
dimas-b, line 66: Did not notice before: If `clientInfo` is not `null` the first call to `newConnection()` is useless, right?

persistence-api/src/main/java/io/stargate/db/datastore/DefaultDataStoreFactory.java
```diff
@@ -0,0 +1,95 @@
...
+import com.datastax.oss.driver.shaded.guava.common.base.Strings;
+import io.stargate.db.AuthenticatedUser;
+import io.stargate.db.ClientInfo;
+import io.stargate.db.Persistence;
+import java.net.InetSocketAddress;
+import java.util.Objects;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class DefaultDataStoreFactory implements DataStoreFactory {
```
dimas-b, line 27: nit: suggestion: `PersistenceBackedDataStoreFactory`?

persistence-api/src/main/java/io/stargate/db/datastore/DefaultDataStoreFactory.java
tomekl007: yes, it seems so. We can create a follow-up ticket. I don't want to change functionality in the refactoring PR.

persistence-api/src/main/java/io/stargate/db/datastore/DefaultDataStoreFactory.java
tomekl007: done

persistence-api/src/main/java/io/stargate/db/datastore/DefaultDataStoreFactory.java
dimas-b: ok, but it is not a greater functional change as introducing `DataStoreFactory` IMHO :)

tomekl007: let me change this and see if CI passes.

dimas-b (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/CassandraFetcher.java
```diff
@@ -82,8 +82,9 @@ public final ResultT get(DataFetchingEnvironment environment) throws Exception {
 
     DataStoreOptions dataStoreOptions =
         DataStoreOptions.builder().defaultParameters(parameters).alwaysPrepareQueries(true).build();
+    // todo here
```
dougwettlaufer, line 85: Leftover comment?

dougwettlaufer (approved):

dimas-b (approved):

PR #544, opened 2020-12-16
GraphQL: Make all query fetchers async (fixes #421)

### LINKED ISSUE #421: GraphQL: Make all data fetchers async
With #400 the GraphQL web layer can now handle client requests asynchronously. We can extend that all the way to the persistence layer:
* change all of our data fetchers to return a `CompletionStage<...>` of whatever they were returning previously
* never block on the futures returned by the persistence layer. Instead, chain the transformations via future callbacks (`thenApply`, `whenComplete`, etc) (I attempted that previously but had to revert it in 9b6df32b).

We just have to be careful not to cross result page boundaries, because that would trigger a synchronous fetch. This can be achieved by using `currentPageRows()`, the way it's already done in `QueryFetcher`. This will only return one page of results, but that's what we want anyway.

This should only be done after #413 to avoid unnecessary conflicts.

olim7t:


auth-table-based-service/src/main/java/io/stargate/auth/table/AuthzTableBasedService.java
```diff
@@ -40,6 +42,29 @@ public ResultSet authorizedDataRead(
...
+      List<TypedKeyValue> typedKeyValues,
+      SourceAPI sourceAPI) {
+    // Cannot perform authorization with a table based token so just wrap and return
+    try {
+      return action.call();
+    } catch (Exception e) {
+      CompletableFuture<ResultSet> failedFuture = new CompletableFuture<>();
+      failedFuture.completeExceptionally(e);
+      return failedFuture;
+    }
```
dimas-b, line 66: Suggestion: If we used an functional interface without checked exception for `action` (e.g. `Supplier`), we could write the body as:
```
    return CompletableFuture
        .completedFuture(null)
        .thenCompose(__ -> action.get());
```

The benefit is that it's more async-native and handles edge cases (like `Error`) gracefully.

WDYT?

dimas-b (approved):

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthzJwtService.java
```diff
@@ -60,6 +66,32 @@ public ResultSet authorizedDataRead(
...
+      Callable<CompletionStage<ResultSet>> action,
+      String token,
+      String keyspace,
+      String table,
+      List<TypedKeyValue> typedKeyValues,
+      SourceAPI sourceAPI) {
+    JSONObject stargateClaims;
+    CompletionStage<ResultSet> resultSetFuture;
+    try {
+      stargateClaims = extractClaimsFromJWT(token);
```
tomekl007, line 83: from what I see, those two actions will be blocking, right?
Can't we make them async as well and chain using `CompletableFuture API`?
i.e.
```
CompletableFuture.supplyAsync(extractClaimsFromJWT).thenApply(preCheckDataReadWrite).thenCompose(action.call())`
```

dimas-b: `extractClaimsFromJWT` should not do any I/O, AFAIK, it is supposed to just parse the token.

tomekl007: still, even. if this is the only CPU bound the async method should be fully async, and not block

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthzTableBasedService.java
olim7t: Yeah my choice would have been `Supplier` if I wrote it from scratch, but the sync version already uses `Callable`. I'll take a look at where it's used and see if we can change it.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthzJwtService.java
olim7t: Yeah parsing the token is CPU-bound and shouldn't be too intensive, so I don't see a strong incentive to do it asynchronously, it would just move the same work to a different executor.

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthzTableBasedService.java
dimas-b: For context: We already have non-backward compatible changes in service interfaces.

dougwettlaufer: I don't see a problem with using `Supplier` here if that makes more sense in an async context. We can leave the other as is or swap `Callable` with `Supplier` in another PR if you want.

dougwettlaufer (approved):

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthzTableBasedService.java
olim7t: I switched the async method only. For the sync one, the lambdas passed as arguments throw checked exceptions, so `Callable` makes sense.

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthzTableBasedService.java
```diff
@@ -40,6 +47,22 @@ public ResultSet authorizedDataRead(
...
+   */
+  @Override
+  public CompletionStage<ResultSet> authorizedAsyncDataRead(
+      Supplier<CompletionStage<ResultSet>> action,
+      String token,
+      String keyspace,
+      String table,
+      List<TypedKeyValue> typedKeyValues,
+      SourceAPI sourceAPI) {
+    return CompletableFuture.completedFuture(null).thenCompose(__ -> action.get());
```
dimas-b, line 64: What do you think about using this pattern in `AuthzJwtService` too? My point that this would unify all exception handling under the `CompletionStage` framework. Currently we have possibility of some exceptions being thrown directly (in `AuthzJwtService`) and some being wrapped in the `CompletionStage`. Granted, some direct throwables are possible with his approach too, but they are mostly limited to severe linkage errors, OOM and stuff... execution exceptions would certainly go inside the `CompletionStage`. WDYT?

olim7t: I agree we should consistently returns exceptions as a failed future.

`CompletionStage` transformations do so for any type of error:

> if a stage's computation terminates abruptly with an (unchecked) exception or error, then all dependent stages requiring its completion complete exceptionally as well, with a CompletionException holding the exception as its cause.

I think we can achieve the same in `AuthzJwtService` by just catching `Throwable` instead of `Exception` in the synchronous part.

dimas-b: > I think we can achieve the same in AuthzJwtService by just catching Throwable instead of Exception in the synchronous part.

Yes, but it adds complexity to our code... just my 2 cents :)

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthnTableBasedService.java
```diff
@@ -220,60 +222,75 @@ protected static boolean checkpw(String password, String hash) {
...
+  @Override
+  public CompletionStage<StoredCredentials> validateTokenAsync(String token) {
+    return CompletableFuture.completedFuture(token)
+        .thenApply(this::parseToken)
+        .thenCompose(this::queryAuthTable)
+        .thenApply(this::extractCredentialsRow)
+        .thenCompose(
+            row ->
+                updateAuthTableTtl(row)
+                    .thenApply(__ -> new StoredCredentials().roleName(row.getString("username"))));
```
olim7t, line 247: We wait for the TTL update to complete before returning the response. I did it that way to mimic the behavior of the sync version, but I think we could fire the update asynchronously and not wait for the response, it doesn't really impact the validation.

I can make that change in both methods (sync and async).

authnz/src/main/java/io/stargate/auth/UnauthorizedException.java
```diff
@@ -15,7 +15,7 @@
  */
 package io.stargate.auth;
 
-public class UnauthorizedException extends Exception {
+public class UnauthorizedException extends RuntimeException {
```
olim7t, line 18: This is way more convenient for the async code.

The downside is that clients aren't forced to handle the error explicitly anymore. But from a quick scan of the code, it looks like in most places this was handled in a generic way anyway (`catch Exception`, Jersey exception handler, etc).

olim7t (commented):
One more set of changes to handle authentication, I realized that the auth-table based variant does perform blocking I/O as well.

dimas-b (approved):

olim7t:
The integration tests are hitting read timeouts, they're also reproducible manually. I've traced the error to the login phase in `DataStoreFactory.create()`, and in particular that section in `DsePersistence.loginInternally`:

```java
          // For now, we're blocking as the login() API is synchronous. If this is a problem, we may
          // have to make said API asynchronous, but it makes things a tad more complex.
          ClientState unused =
              this.clientState.login(new AuthenticatedUser(user.name())).blockingGet();
```

That is indeed the problem. If we want to allow asynchronous use of the persistence API, we can't have blocking methods on the execution path.

dimas-b:
So we'd have to make `DataStoreFactory.create()` async, right, @olim7t ?

I suppose it'd be best done in a separate PR to avoid tangential changes.

olim7t:
> So we'd have to make `DataStoreFactory.create()` async, right, @olim7t ?

Yes.

> I suppose it'd be best done in a separate PR to avoid tangential changes.

Sounds good, I'll do that.

olim7t:
I'm going to close this. Unfortunately I came to the realization that it's going to be very hard to achieve, given that we piggyback on the internal Cassandra code which is not designed to be async (at least the OSS variants, I think DSE would have been fine).

The latest hurdle I ran into was that `DataStoreFactory.create()` performs authorization checks as part of opening a connection. This depends on code deep in the Cassandra codebase that executes statements synchronously (e.g. `CassandraRoleManager.select()` calls `statement.execute()`).

We could introduce auxiliary thread pools to schedule such blocking tasks, but from my experience that approach is very hard to maintain, because you have to guarantee that those tasks won't accidentally trigger something that ends up being scheduled on that same pool (introducing the risk of a deadlock).

PR #554, opened 2020-12-18
API for authorization processors

dimas-b:
In reviewable state now. See commit log for specific change comments.


authnz/src/main/java/io/stargate/auth/entity/AuthEntityWrapper.java
```diff
@@ -0,0 +1,34 @@
...
+import org.immutables.value.Value;
+
+/**
+ * This is a base class of authentication and authorization data exposed by a particular {@link
+ * Persistence} implementation.
+ *
+ * <p>The wrapped object is intended to be recoverable by code that is aware of the the persistence
+ * implementation details. Sub-interfaces also define some generic property accessors.
+ *
+ * @param <Wrapped> The persistence-specific entity type.
```
tomekl007, line 28: I don't see a `Wrapped` class in this PR. What am I missing?

authnz/src/main/java/io/stargate/auth/entity/ResourceKind.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.auth.entity;
+
+public enum ResourceKind {
+  TABLE,
+  FUNCTION,
```
tomekl007, line 20: why do we need this distinction. Could you give me an example of a function?

authnz/src/main/java/io/stargate/auth/entity/AuthorizedResource.java
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * Identifies the scope of the resource according to the resource {@link #kind}.
+   *
+   * <p>For roles the scope should be {@code null}.
+   *
+   * <p>For tables and functions the scope is the keyspace name. If {@code null}, it means "all
+   * keyspaces" (in which case {@link #element()} should be {@code null} too).
+   */
+  @Value.Default
+  default String scope() {
```
tomekl007, line 37: I wonder if we could create an Enum that presents all possible scopes? This way we would make the API more type-safe) or reuse because I think there is already an Enum that carries the info about `Scope`)

authnz/src/main/java/io/stargate/auth/entity/AuthorizedResource.java
```diff
@@ -0,0 +1,50 @@
...
+    return null;
+  }
+
+  /**
+   * Identifies the name of the resource.
+   *
+   * <p>In the name is {@code null}, it means "all resources" within the scope.
+   */
+  @Value.Default
+  default String element() {
```
tomekl007, line 47: Maybe we could create a simple abstract class `Element`. Then it will have two child classes:
1.  has only one field: `elementName`. 
2. It has no elements but it denotes this special case: `all resources`.

authnz/src/main/java/io/stargate/auth/PermissionKind.java
```diff
@@ -0,0 +1,38 @@
...
+   */
+  ACCESS,
+
+  /**
+   * Indicates the related permission controls the authority to grant/revoke access the resource to
+   * other users.
+   *
+   * <p>This flag generally corresponds to {@code GRANT/REVOKE AUTHORIZE FOR} CQL statements.
+   */
+  AUTHORITY,
```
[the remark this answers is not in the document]
dimas-b: It indicates that the target use gets the permission to _grant_ the listed permissions (from the `addPermissions` method) to someone else, but it cannot use the listed permissions to _access_ the resource. 

This is to support DSE's "separation of duties"

https://docs.datastax.com/en/security/6.8/security/grantAuthPer.html

Please note that the Stargate API is a bit wider than what DSE supports. I felt it was justified to make `PermissionKind` and `AuthorizationOutcome` orthogonal.

dougwettlaufer (approved):

authnz/src/main/java/io/stargate/auth/AuthorizationProcessor.java
```diff
@@ -0,0 +1,75 @@
...
+   * {@code grantee}. Note that revoking a {@link AuthorizationOutcome#DENY negative} permission is
+   * essentially widening access to the resource.
+   *
+   * @param performer the user under whose authority the revocation operation is performed
+   * @param outcome whether operations under the specified permissions are allowed or denied
+   * @param kind whether permission is granted to access the resource or control access to it
+   * @param permissions the list of permissions to be revoked
+   * @param resource the resource whose access permissions are revoked
+   * @param grantee the role whose permissions are revoked
+   * @return the {@link CompletionStage} tracking the progress of the grant operation.
```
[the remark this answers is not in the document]
dimas-b: by exception, I assume.

authnz/src/main/java/io/stargate/auth/AuthorizationProcessor.java
```diff
@@ -0,0 +1,75 @@
...
+   *
+   * @param performer the user under whose authority the grant operation is performed
+   * @param outcome whether operations under the specified permissions are allowed or denied
+   * @param kind whether permission is granted to access the resource or control access to it
+   * @param permissions the list of permissions to be granted
+   * @param resource the resource whose access permissions are granted
+   * @param grantee the role receiving the permissions
+   * @return the {@link CompletionStage} tracking the progress of the grant operation.
+   */
+  CompletionStage<Void> addPermissions(
```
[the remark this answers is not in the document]
dimas-b: `AuthorizationOutcome` is meant to indicate the outcome of future permission checks when these permissions are in force.

authnz/src/main/java/io/stargate/auth/entity/AccessPermission.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Represents a permission from a particular {@link Persistence} implementation.
+ *
+ * <p>From Stargate perspective only the name of the permission matters.
+ */
+@Value.Immutable
+public interface AccessPermission<W> extends AuthEntityWrapper<W> {
+
+  @Value.Parameter
+  String name();
```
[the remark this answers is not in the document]
dimas-b: `SELECT`

authnz/src/main/java/io/stargate/auth/entity/AuthEntityWrapper.java
dimas-b: It's a type parameter... I'm thinking to remove the wrappers, actually.

authnz/src/main/java/io/stargate/auth/entity/ResourceKind.java
dimas-b: `avg`... per-function access may or may have to not be enforced in Stargate, but since grants can be made to tables and functions separately, I added this enum for to have parity with CQL capabilities.

authnz/src/main/java/io/stargate/auth/entity/AuthorizedResource.java
dimas-b: `scope` is either `null` of the keyspace name... I'll rename the method to `keyspace()` for clarity.

authnz/src/main/java/io/stargate/auth/entity/AuthorizedResource.java
dimas-b: Yes, that's doable. I did not want to create too many classes, but if you find it useful, I'll add that.

dimas-b: Added `EntitySelector` for this purpose.

dimas-b: I also replaced old usages of `AuthorizedRole` with the new `EntitySelector`

authnz/pom.xml
```diff
@@ -52,7 +52,7 @@
             </Bundle-Description>
             <Bundle-SymbolicName>io.stargate.auth</Bundle-SymbolicName>
             <Import-Package>org.osgi.framework</Import-Package>
-            <Export-Package>io.stargate.auth</Export-Package>
+            <Export-Package>io.stargate.auth,io.stargate.auth.entity</Export-Package>
```
tomekl007, line 55: do we need to specify the inner package? 
Is the outer package `io.stargate.auth` not enough?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/DsePersistenceActivator.java
```diff
@@ -31,12 +35,20 @@
...
   private final LazyServicePointer<AuthorizationService> authorizationService =
       LazyServicePointer.create(
           AuthorizationService.class,
           "AuthIdentifier",
           System.getProperty("stargate.auth_id", "AuthTableBasedService"));
+  private final ServicePointer<AuthorizationProcessor> authorizationProcessor =
+      ServicePointer.create(
+          AuthorizationProcessor.class,
+          "AuthProcessorId",
+          AUTHZ_PROCESSOR_ID == null ? "NONE" : AUTHZ_PROCESSOR_ID);
```
tomekl007, line 51: Why do we need this special if and return `NONE`? Is this an `OSGi` filter meaning that it won't wait for this dependent service?
I see that in the dependencies section we fiter out if null:
```
  if (AUTHZ_PROCESSOR_ID != null) {
      dependencies.add(authorizationProcessor);
    }
```
so this condition: `AUTHZ_PROCESSOR_ID == null ? "NONE" : AUTHZ_PROCESSOR_ID` shouldn't be needed.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DelegatingAuthorizer.java
```diff
@@ -0,0 +1,222 @@
...
+              permissions(permissions),
+              resource(resource),
+              role(grantee));
+
+      stages.add(stage.toCompletableFuture());
+    }
+
+    get(CompletableFuture.allOf(stages.toArray(new CompletableFuture[0])));
+
+    // assume all permissions were granted if we did not an exception
```
tomekl007, line 87: typo/missing word: `// assume all permissions were granted if we did not an exception` -> ` if we did not get an exception`

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DelegatingAuthorizer.java
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  @Override
+  public Set<Permission> revoke(
+      AuthenticatedUser performer,
+      Set<Permission> permissions,
+      IResource resource,
+      RoleResource revokee,
+      GrantMode... grantModes) {
+    AuthorizationProcessor processor = authProcessor;
```
tomekl007, line 98: why do you need this reassignment, instead of using `authProcessor` directly?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DelegatingAuthorizer.java
```diff
@@ -0,0 +1,222 @@
...
+    get(CompletableFuture.allOf(stages.toArray(new CompletableFuture[0])));
+
+    // assume all permissions were revoked if we did not an exception
+    return permissions;
+  }
+
+  private static void get(CompletionStage<Void> stage) {
+    try {
+      // wait for completion since the calling API is synchronous
+      stage.toCompletableFuture().get();
```
tomekl007, line 128: We are risking indefinitely blocking here. We should use get with some sane timeout to prevent that.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DelegatingAuthorizer.java
```diff
@@ -0,0 +1,222 @@
...
+    } catch (ExecutionException e) {
+      Throwable cause = e.getCause();
+      if (cause instanceof RequestValidationException) {
+        throw (RequestValidationException) cause;
```
tomekl007, line 132: what value do we get by throwing the `RequestValidationException` in that case instead of coalescing all catch blocks and throw:
`throw new AuthRequestExecutionException(SERVER_ERROR, e.getMessage(), e);` always?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DelegatingAuthorizer.java
```diff
@@ -0,0 +1,222 @@
...
+  private static AuthorizedResource resource(IResource resource) {
+    String resourcePath = resource.getName();
+    ImmutableAuthorizedResource authResource;
+
+    if (resource instanceof FunctionResource) {
+      authResource = ImmutableAuthorizedResource.of(ResourceKind.FUNCTION);
+
+      FunctionResource fr = (FunctionResource) resource;
+
+      boolean hasKeyspace = resourcePath.indexOf("/") > 0;
```
tomekl007, line 174: I don't understand why `resourcePath.indexOf("/") > 0` means that it has a keyspace. Could you elaborate more?

testing-services/src/main/java/io/stargate/testing/TestingServicesActivator.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Collections;
+import java.util.Hashtable;
+import java.util.List;
+
+public class TestingServicesActivator extends BaseActivator {
+
+  private static final String TEST_AUTHZ_PROCESSOR_ID = "TestAuthzProcessor";
+
+  public TestingServicesActivator() {
+    super("testing-services");
```
tomekl007, line 30: nit: change `testing-services` to something more descriptive, i.e.:
`testing-authZ-services`

testing-services/src/main/java/io/stargate/testing/TestingServicesActivator.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @Override
+  protected List<ServiceAndProperties> createServices() {
+    if (TEST_AUTHZ_PROCESSOR_ID.equals(System.getProperty("stargate.authorization.processor.id"))) {
```
tomekl007, line 35: I am not sure if having the test logic based on the global system property is a good way to test our code.
We are risking that the `createServices` will not create any service and this class client's logic will pass silently without examining the actual service. wdyt?

testing/src/test/java/io/stargate/it/storage/LogCollector.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public class LogCollector implements OutputListener, AutoCloseable {
```
tomekl007, line 28: very nice addition to the existing Stargate testing infra.

testing/src/test/java/io/stargate/it/cql/AuthorizationProcessorTest.java
```diff
@@ -0,0 +1,56 @@
...
+  @SuppressWarnings("unused") // referenced in @StargateSpec
+  public static void buildParameters(StargateParameters.Builder builder) throws IOException {
+    builder.enableAuth(true);
+    builder.putSystemProperties("stargate.authorization.processor.id", "TestAuthzProcessor");
+  }
+
+  @Test
+  public void grantSelect(LogCollector log, CqlSession session) {
+    session.execute("GRANT SELECT on auth_keyspace2.table3 TO 'auth_user1'");
+    assertThat(log.filter(0, Pattern.compile(".+testing: addPermissions: (.+)"), 1))
```
tomekl007, line 53: I think that verifying logs is sometimes useful, but for the auth functionality maybe not enough?
Shouldn't we execute the query and validate if the `auth_user1` can indeed SELECT on `auth_keyspace2.table3`?
And the other user doesn't?

authnz/pom.xml
dimas-b: we could add `.*`, but the inner package is exported because it is/will be used in the authz processor implementations.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/DsePersistenceActivator.java
dimas-b: perhaps we could use `null` in the filter as well. It just did not feel right to me :)

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DelegatingAuthorizer.java
dimas-b: it's a remnant of previous experiment when the field was an `AtomicReference`... I'll clean up.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DelegatingAuthorizer.java
dimas-b: good point.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DelegatingAuthorizer.java
dimas-b: It's based on the way `getName` output is formatted. Unfortunately, `FunctionResource` does not provide a getter to checking whether it has a keyspace and will throw exceptions if we try to get the keyspace if it does not have one.

testing-services/src/main/java/io/stargate/testing/TestingServicesActivator.java
dimas-b: My thinking was that perhaps we may need to add other testing services later.

testing-services/src/main/java/io/stargate/testing/TestingServicesActivator.java
dimas-b: If `stargate.authorization.processor.id` is set, but the implementation is not registered, the persistence will not activate and tests will fail for sure.

dimas-b: In general, this seems to be our pattern for swapping service implementations in runtime... we have pretty much the same logic for other "auth" services.

testing/src/test/java/io/stargate/it/cql/AuthorizationProcessorTest.java
dimas-b: In this particular case that test is not meaningful because we do not have pluggable RBAC. When we intercept grants, DSE will not update its permission data, and we cannot (yet) redirect permission checks to Stargate.

dimas-b: So the test essentially validates that we can property intercept various CQL statements (more tests to come).

testing-services/src/main/java/io/stargate/testing/TestingServicesActivator.java
tomekl007: good point.

tomekl007:
There are test failures on this branch:
```
Step #7 - "dse-6.8": [INFO] Results:
Step #7 - "dse-6.8": [INFO] 
Step #7 - "dse-6.8": [ERROR] Errors: 
Step #7 - "dse-6.8": [ERROR]   BatchStatementTest ? IllegalState java.util.concurrent.ExecutionException: org...
Step #7 - "dse-6.8": [ERROR]   JwtAuthTest.createAlterDropTable:324 ? ReadTimeout Cassandra timeout during re...
Step #7 - "dse-6.8": [ERROR]   JwtAuthTest.setup:74 ? AllNodesFailed All 1 node(s) tried for the query failed...
Step #7 - "dse-6.8": [ERROR]   JwtAuthTest.setup:74 ? AllNodesFailed All 1 node(s) tried for the query failed...
Step #7 - "dse-6.8": [ERROR]   JwtAuthTest.setup:74 ? AllNodesFailed All 1 node(s) tried for the query failed...
Step #7 - "dse-6.8": [ERROR]   SchemaChangesTest ? IllegalState java.util.concurrent.ExecutionException: org....
Step #7 - "dse-6.8": [ERROR]   SnappyCompressionTest ? IllegalState java.util.concurrent.ExecutionException: ...
Step #7 - "dse-6.8": [ERROR]   SupportedOptionsTest ? IllegalState java.util.concurrent.ExecutionException: o...
```

persistence-dse-6.8/src/main/java/io/stargate/db/dse/DsePersistenceActivator.java
dimas-b: I reworked this code in more recent commits.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DelegatingAuthorizer.java
dimas-b: fixed

dimas-b:
Rebased to the latest `master`

dimas-b:
Closing for now.

TDB: intercept GRANT / REVOKE operations in C* 3.11 and 4.0

dimas-b:
Since last review: Added coverage for C* 3.11 and 4.0, rebased and minor tweaks to the API and DSE impl .

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/DelegatingAuthorizer.java
```diff
@@ -0,0 +1,183 @@
...
+import org.apache.cassandra.auth.IResource;
+import org.apache.cassandra.auth.Permission;
+import org.apache.cassandra.auth.RoleResource;
+import org.apache.cassandra.exceptions.RequestExecutionException;
+import org.apache.cassandra.exceptions.RequestValidationException;
+import org.apache.cassandra.stargate.exceptions.AuthenticationException;
+
+public class DelegatingAuthorizer extends CassandraAuthorizer {
+
+  private static final long PROCESSING_TIMEOUT_MILLIS =
```
tomekl007, line 47: please use `Duratiion.ofMinutes(5).getMillis()` for providing the default. It will be more readable.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/DelegatingAuthorizer.java
```diff
@@ -0,0 +1,183 @@
...
+    get(stage);
+  }
+
+  private static void get(CompletionStage<Void> stage) {
+    try {
+      // wait for completion since the calling API is synchronous
+      stage.toCompletableFuture().get(PROCESSING_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
+    } catch (ExecutionException e) {
+      Throwable cause = e.getCause();
+      if (cause instanceof org.apache.cassandra.stargate.exceptions.RequestValidationException) {
```
tomekl007, line 110: is it possible that `cause == null`?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DelegatingAuthorizer.java
```diff
@@ -0,0 +1,240 @@
...
+import org.apache.cassandra.auth.Permission;
+import org.apache.cassandra.auth.RoleResource;
+import org.apache.cassandra.auth.permission.CorePermission;
+import org.apache.cassandra.stargate.exceptions.RequestValidationException;
+import org.javatuples.Pair;
+
+public class DelegatingAuthorizer extends CassandraAuthorizer {
+
+  private static final long PROCESSING_TIMEOUT_MILLIS =
+      Long.getLong("stargate.authorization.processing.timeout.millis", 5 * 60 * 1000);
```
tomekl007, line 54: same comment about using `Duration`

authnz/src/main/java/io/stargate/auth/AuthorizationProcessor.java
```diff
@@ -0,0 +1,75 @@
...
+      Actor performer,
+      AuthorizationOutcome outcome,
```
mpenick, line 49: This is sort of confusing. So you can add permissions for a DENY outcome? Seems weird.

dimas-b: DSE has `RESTRICT` CQL statements, which sort of add "negative" permissions, which is this parameter is meant to represent.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/DelegatingAuthorizer.java
dimas-b: yes, and the `instanceof` result will be `false` in that case.

authnz/src/main/java/io/stargate/auth/entity/EntitySelector.java
```diff
@@ -0,0 +1,51 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.auth.entity;
+
+/**
+ * Identities one of a ground of entities.
```
dougwettlaufer, line 19: Sorry, not sure I understand this comment.

testing-services/pom.xml
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>io.stargate</groupId>
+    <artifactId>stargate</artifactId>
+    <version>1.0.5-SNAPSHOT</version>
```
dougwettlaufer, line 7:
```suggestion
    <version>1.0.7-SNAPSHOT</version>
```

testing-services/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+      <artifactId>core</artifactId>
+      <version>${project.version}</version>
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
```
dougwettlaufer, line 41: Does this need to be an OSGI bundle if its for testing?

testing-services/src/main/java/io/stargate/testing/TestingServicesActivator.java
```diff
@@ -0,0 +1,58 @@
...
+package io.stargate.testing;
+
+import io.stargate.auth.AuthorizationProcessor;
+import io.stargate.core.activator.BaseActivator;
+import io.stargate.testing.auth.LoggingAuthorizationProcessorImpl;
+import java.util.Collections;
+import java.util.Hashtable;
+import java.util.List;
+
+public class TestingServicesActivator extends BaseActivator {
```
dougwettlaufer, line 25: Maybe I missed it but what prevents this from being pulled into the actual deployment?

dougwettlaufer (approved):

authnz/src/main/java/io/stargate/auth/entity/EntitySelector.java
dimas-b: typo: `group`. WIll fix.

testing-services/pom.xml
dimas-b: Integration test code runs in a separate JVM. We cannot really provide "test" implementations for service interfaces unless we deploy them in OSGi.

We could probably use a simple jar for that, but I though a full bundle would allow for better test coverage with respect to service impl. discovery and dependency resolution (from core code).

testing-services/src/main/java/io/stargate/testing/TestingServicesActivator.java
dimas-b: The `testing-services` bundle is not put into `stargate-lib` during the main build sequence. It's copied there only during integration tests: https://github.com/stargate/stargate/pull/554/files#diff-6eb247058d9d4d2db49e65e071f2595396f06165f3ee3eb78477c403242b3353R366

testing-services/pom.xml
dimas-b: fixed and rebased

authnz/src/main/java/io/stargate/auth/entity/EntitySelector.java
dimas-b: fixed

PR #556, opened 2020-12-18
Make builds serial for now

dougwettlaufer:
While GCP support is working on what's going on with the builds let's try running serially so at least we can get things to pass without docker OOM.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -52,7 +52,7 @@ steps:
       - 'PERSISTENCE_BACKEND=cassandra-3.11'
       - 'COMMIT_ID=$COMMIT_SHA'
   - id: 'cassandra-4.0'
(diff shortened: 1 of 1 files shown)


dimas-b (approved):

tomekl007 (approved):
Worth pointing out: it increases the total build time to ~ 2h 30 minutes

dougwettlaufer:
> Worth pointing out: it increases the total build time to ~ 2h 30 minutes

Yes, this is unfortunate and completely expected. We can hopefully revert this once GCP support comes back with more info on what's going on with our existing builds.

dimas-b:
Might be worth increasing the build timeout from 3 hr to 4+ hr, given that the normal run time is pretty close to the threshold already

PR #559, opened 2020-12-22
Fix #545 - NPE is thrown if TableOptions is null

### LINKED ISSUE #545: NPE is thrown if tableOptions are not specified in v1 create table
Request is causing a 500 due to an NPE

```
curl -L -X POST 'http://localhost:8082/v1/keyspaces/ks1/tables' \
-H 'X-Cassandra-Token: TOKEN' \
--data-raw '{
    "name": "cyclist_name",
    "columnDefinitions": [
        {
            "name": "id",
            "typeDefinition": "uuid"
        },
        {
            "name": "lastname",
            "typeDefinition": "text"
        },
        {
            "name": "firstname",
            "typeDefinition": "text"
        }
    ],
    "primaryKey": {
        "partitionKey": [
            "id"
        ]
    }
}'
```

eribeiro:
A NPE is thrown whenever TableOptions is set to null in TableResources.

### DIFF
--- a/restapi/src/main/java/io/stargate/web/resources/TableResource.java
+++ b/restapi/src/main/java/io/stargate/web/resources/TableResource.java
@@ -200,13 +200,15 @@ public class TableResource {
           db.getAuthorizationService()
               .authorizeSchemaWrite(token, keyspaceName, tableName, Scope.CREATE, SourceAPI.REST);
 
+          int ttl = (options == null) ? 0 : options.getDefaultTimeToLive();
+
           localDB
               .queryBuilder()
               .create()
               .table(keyspaceName, tableName)
               .ifNotExists(tableAdd.getIfNotExists())
               .column(columns)
-              .withDefaultTTL(options.getDefaultTimeToLive())
+              .withDefaultTTL(ttl)
               .build()
               .execute(ConsistencyLevel.LOCAL_QUORUM)
               .get();
--- a/restapi/src/main/java/io/stargate/web/resources/TableResource.java
+++ b/restapi/src/main/java/io/stargate/web/resources/TableResource.java
@@ -200,7 +200,10 @@ public class TableResource {
           db.getAuthorizationService()
               .authorizeSchemaWrite(token, keyspaceName, tableName, Scope.CREATE, SourceAPI.REST);
 
-          int ttl = (options == null) ? 0 : options.getDefaultTimeToLive();
+          int ttl = 0;
+          if (options != null && options.getDefaultTimeToLive() != null) {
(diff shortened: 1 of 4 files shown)


restapi/src/main/java/io/stargate/web/resources/TableResource.java
```diff
@@ -200,13 +200,15 @@ public Response addTable(
           db.getAuthorizationService()
               .authorizeSchemaWrite(token, keyspaceName, tableName, Scope.CREATE, SourceAPI.REST);
 
+          int ttl = (options == null) ? 0 : options.getDefaultTimeToLive();
```
dimas-b, line 203: `getDefaultTimeToLive()` returns an object `Integer`... Would it make sense to cover the case when it is `null`? It seems to be `null` by default.

eribeiro: Good point. IMHO, we could 
1) Set a default value
```
public class TableOptions {

  private Integer defaultTimeToLive = 0;
```  
as both the constructor and the setter  (shown below) currently accepts the primitive type only, so they cannot set `defaultTimeToLive` to null. 

```
  public TableOptions(
    @JsonProperty("defaultTimeToLive") final int defaultTimeToLive,
    ...
    public void setDefaultTimeToLive(int defaultTimeToLive) {
      this.defaultTimeToLive = defaultTimeToLive;
     }
    ...
  ```
  
  2) Make `defaultTimeToLive` field a primitive type for the same reasons as above;
  
  I am more inclined to the second option. Wdyt?

dougwettlaufer: It's defined as an Integer so that we can validate during update without accidentally setting TTL to 0 when it's not intended
https://github.com/stargate/stargate/blob/master/restapi/src/main/java/io/stargate/web/resources/v2/schemas/TablesResource.java#L326-L332

Since it's not marked as required for createTable I think that if `defaultTimeToLive == null` then we just set it as 0 and create the table.

dimas-b:
Also, @eribeiro , could you rebase to the latest `master`? It looks like you're missing some CI changes that make it less flaky (but longer, of course :shrug: )

restapi/src/main/java/io/stargate/web/resources/TableResource.java
eribeiro: Sure, makes sense. I have just done the changes as required to treat `defaultTimeToLive` as `null`. :)

dougwettlaufer (approved):

Issue #80, opened 2020-09-15
Surface Indexes


blerghh:
Indexes should be treated as directives for schema evolution and migration purposes.

```
type Tweet 
@Index(name: "MyCustomIndex", type: StorageAttachedIndex, on: [ username ])
@PrimaryKey(field: [tweet])
{
  tweet: Text
  username: Text
}
```

olim7t:
Same comment as #139.

PR #562, opened 2020-12-24
Add index mutation for GraphQL

### LINKED ISSUE #543: Expose ability to manage indexes in GraphQL
We should provide a new mutation that allows for managing the full index lifecycle via graphql.

eribeiro:
Fixes #543 - Expose ability to manage indexes in GraphQL

### DIFF
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/DdlSchemaBuilder.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/DdlSchemaBuilder.java
@@ -36,9 +36,11 @@ import io.stargate.db.datastore.DataStoreFactory;
 import io.stargate.graphql.schema.fetchers.ddl.AllKeyspacesFetcher;
 import io.stargate.graphql.schema.fetchers.ddl.AlterTableAddFetcher;
 import io.stargate.graphql.schema.fetchers.ddl.AlterTableDropFetcher;
+import io.stargate.graphql.schema.fetchers.ddl.CreateIndexFetcher;
 import io.stargate.graphql.schema.fetchers.ddl.CreateKeyspaceFetcher;
 import io.stargate.graphql.schema.fetchers.ddl.CreateTableFetcher;
 import io.stargate.graphql.schema.fetchers.ddl.CreateTypeFetcher;
+import io.stargate.graphql.schema.fetchers.ddl.DropIndexFetcher;
 import io.stargate.graphql.schema.fetchers.ddl.DropKeyspaceFetcher;
 import io.stargate.graphql.schema.fetchers.ddl.DropTableFetcher;
 import io.stargate.graphql.schema.fetchers.ddl.DropTypeFetcher;
@@ -72,6 +74,8 @@ class DdlSchemaBuilder {
                 buildDropTable(),
                 buildCreateType(),
                 buildDropType(),
+                buildCreateIndex(),
+                buildDropIndex(),
                 buildCreateKeyspace(),
                 buildDropKeyspace()))
         .query(buildQuery(buildKeyspaceByName(), buildKeyspaces()))
@@ -450,6 +454,41 @@ class DdlSchemaBuilder {
         .type(Scalars.GraphQLBoolean)
         .dataFetcher(
             new CreateTableFetcher(authenticationService, authorizationService, dataStoreFactory))
+        .build();
+  }
+
+  private GraphQLFieldDefinition buildCreateIndex() {
+    return GraphQLFieldDefinition.newFieldDefinition()
+        .name("createIndex")
+        .argument(GraphQLArgument.newArgument().name("keyspaceName").type(Scalars.GraphQLString))
+        .argument(
+            GraphQLArgument.newArgument().name("tableName").type(nonNull(Scalars.GraphQLString)))
+        .argument(
+            GraphQLArgument.newArgument().name("columnName").type(nonNull(Scalars.GraphQLString)))
+        .argument(GraphQLArgument.newArgument().name("indexName").type(Scalars.GraphQLString))
+        .argument(
+            GraphQLArgument.newArgument().name("customIndexClass").type(Scalars.GraphQLString))
+        .argument(GraphQLArgument.newArgument().name("ifNotExists").type(Scalars.GraphQLBoolean))
+        .argument(GraphQLArgument.newArgument().name("indexEntries").type(Scalars.GraphQLBoolean))
+        .argument(GraphQLArgument.newArgument().name("indexKeys").type(Scalars.GraphQLBoolean))
+        .argument(GraphQLArgument.newArgument().name("indexValues").type(Scalars.GraphQLBoolean))
+        .argument(GraphQLArgument.newArgument().name("indexFull").type(Scalars.GraphQLBoolean))
+        .type(Scalars.GraphQLBoolean)
+        .dataFetcher(
+            new CreateIndexFetcher(authenticationService, authorizationService, dataStoreFactory))
+        .build();
+  }
+
+  private GraphQLFieldDefinition buildDropIndex() {
+    return GraphQLFieldDefinition.newFieldDefinition()
+        .name("dropIndex")
+        .argument(GraphQLArgument.newArgument().name("keyspaceName").type(Scalars.GraphQLString))
+        .argument(
+            GraphQLArgument.newArgument().name("indexName").type(nonNull(Scalars.GraphQLString)))
+        .argument(GraphQLArgument.newArgument().name("ifExists").type(Scalars.GraphQLBoolean))
+        .type(Scalars.GraphQLBoolean)
+        .dataFetcher(
+            new DropIndexFetcher(authenticationService, authorizationService, dataStoreFactory))
         .build();
   }
 
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/DdlSchemaBuilder.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/DdlSchemaBuilder.java
@@ -467,12 +467,33 @@ class DdlSchemaBuilder {
             GraphQLArgument.newArgument().name("columnName").type(nonNull(Scalars.GraphQLString)))
         .argument(GraphQLArgument.newArgument().name("indexName").type(Scalars.GraphQLString))
         .argument(
-            GraphQLArgument.newArgument().name("customIndexClass").type(Scalars.GraphQLString))
+            GraphQLArgument.newArgument()
+                .name("indexType")
+                .description(
+                    "Adds a custom index type that can be identified by name (e.g., StorageAttachedIndex), "
+                        + "or class name (e.g., org.apache.cassandra.index.sasi.SASIIndex) ")
+                .type(Scalars.GraphQLString))
         .argument(GraphQLArgument.newArgument().name("ifNotExists").type(Scalars.GraphQLBoolean))
-        .argument(GraphQLArgument.newArgument().name("indexEntries").type(Scalars.GraphQLBoolean))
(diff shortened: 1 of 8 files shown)


eribeiro:
@dougwettlaufer Oops, sorry Doug, I forced pushed a commit to this branch soon after you run /gcbrun. Do you need to run it again?

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/CreateIndexFetcher.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    CollectionIndexingType indexingType =
+        ImmutableCollectionIndexingType.builder()
+            .indexEntries(indexEntries)
+            .indexKeys(indexKeys)
+            .indexValues(indexValues)
+            .indexFull(indexFull)
+            .build();
+
```
dougwettlaufer, line 89: Will also need to call authorizationService here, similar to

https://github.com/stargate/stargate/blob/master/graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/TableFetcher.java

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/DropIndexFetcher.java
```diff
@@ -0,0 +1,55 @@
...
+
+    if (keyspaceName == null) {
+      List<String> parts = Splitter.on('.').splitToList(indexName);
+      if (parts.size() < 2) {
+        throw new IllegalArgumentException("Missing field argument keyspaceName @ 'dropIndex'");
+      }
+      keyspaceName = parts.get(0);
+      indexName = parts.get(1);
+    }
+
```
dougwettlaufer, line 52: Same comment regarding auth as above but will use scope DROP instead of CREATE.

dougwettlaufer (commented):
Small comments about authz

eribeiro:
Ouch, looks like the building is failing but I don't have access to the full report. Maybe it's failing because of the changes to `QueryBuilderImpl`? 

I have identified some errors when it was building the CREATE INDEX command and an index name was supplied. Also when using custom indexes (CUSTOM).

PS: I have just submitted the changes as reviewed. :)

graphqlapi/src/main/java/io/stargate/graphql/schema/DdlSchemaBuilder.java
```diff
@@ -453,6 +457,41 @@ private GraphQLFieldDefinition buildCreateTable() {
...
+    return GraphQLFieldDefinition.newFieldDefinition()
+        .name("createIndex")
+        .argument(GraphQLArgument.newArgument().name("keyspaceName").type(Scalars.GraphQLString))
+        .argument(
+            GraphQLArgument.newArgument().name("tableName").type(nonNull(Scalars.GraphQLString)))
+        .argument(
+            GraphQLArgument.newArgument().name("columnName").type(nonNull(Scalars.GraphQLString)))
+        .argument(GraphQLArgument.newArgument().name("indexName").type(Scalars.GraphQLString))
+        .argument(
+            GraphQLArgument.newArgument().name("customIndexClass").type(Scalars.GraphQLString))
```
dougwettlaufer, line 470: I know under the hood in QueryBuilder it's `customIndexClass` but is that how we want to expose it? For example, an SAI index would be `StorageAttachedIndex` rather than a class like `org.apache.cassandra.index.sasi.SASIIndex`. 

Maybe something like 'indexType` and then add a description explaining it can be the class name or type of index?

graphqlapi/src/main/java/io/stargate/graphql/schema/DdlSchemaBuilder.java
```diff
@@ -453,6 +457,41 @@ private GraphQLFieldDefinition buildCreateTable() {
...
+        .argument(GraphQLArgument.newArgument().name("ifNotExists").type(Scalars.GraphQLBoolean))
+        .argument(GraphQLArgument.newArgument().name("indexEntries").type(Scalars.GraphQLBoolean))
+        .argument(GraphQLArgument.newArgument().name("indexKeys").type(Scalars.GraphQLBoolean))
+        .argument(GraphQLArgument.newArgument().name("indexValues").type(Scalars.GraphQLBoolean))
+        .argument(GraphQLArgument.newArgument().name("indexFull").type(Scalars.GraphQLBoolean))
```
dougwettlaufer, line 475: Might be good to surface these as descriptions

https://github.com/stargate/stargate/blob/master/persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java#L1049-L1062

dougwettlaufer:
> Ouch, looks like the building is failing but I don't have access to the full report. Maybe it's failing because of the changes to `QueryBuilderImpl`?
> 
> I have identified some errors when it was building the CREATE INDEX command and an index name was supplied. Also when using custom indexes (CUSTOM).
> 
> PS: I have just submitted the changes as reviewed. :)

Looks like the error is

```
[ERROR] Failed to execute goal com.coveo:fmt-maven-plugin:2.9:check (default) on project graphqlapi: Found 1 non-complying files, failing build -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
```

Quick fix is to run `mvn xml-format:xml-format fmt:format`

dougwettlaufer (commented):
Looks good overall, just a few small comments. Would like to get a second look from @olim7t when he gets back next week before merging.

eribeiro:
I have addressed the review comments. Please, let me know (and excuse me in advance) if I understood some comment incorrectly. :) I've also run the format command.

dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/DdlSchemaBuilder.java
```diff
@@ -453,6 +457,62 @@ private GraphQLFieldDefinition buildCreateTable() {
...
+                .description(
+                    "Adds a custom index type that can be identified by name (e.g., StorageAttachedIndex), "
+                        + "or class name (e.g., org.apache.cassandra.index.sasi.SASIIndex) ")
+                .type(Scalars.GraphQLString))
+        .argument(GraphQLArgument.newArgument().name("ifNotExists").type(Scalars.GraphQLBoolean))
+        .argument(
+            GraphQLArgument.newArgument()
+                .name("indexKeys")
+                .description("Index keys of a map")
+                .type(Scalars.GraphQLBoolean))
```
olim7t, line 481: A GraphQL enum might be a bit more user-friendly:
```
enum IndexKind {
  KEYS
  VALUES
  ENTRIES
  FULL
}
```
For the default kind we can either keep the field nullable, or introduce a `SIMPLE` constant (that's how it's called in the Cassandra codebase).

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/IndexFetcher.java
```diff
@@ -0,0 +1,50 @@
...
+import io.stargate.auth.*;
+import io.stargate.auth.UnauthorizedException;
+import io.stargate.db.datastore.DataStoreFactory;
+import io.stargate.db.query.Query;
+import io.stargate.db.query.builder.QueryBuilder;
+import io.stargate.graphql.web.HttpAwareContext;
+
+public abstract class IndexFetcher extends DdlQueryFetcher {
+
+  protected Scope scope;
```
olim7t, line 15: Nit: if you're going to keep `getScope()`, you can turn this into a local variable.
Although I would personally remove `getScope()` and set it via the constructor instead.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/CreateIndexFetcher.java
```diff
@@ -0,0 +1,104 @@
...
+      String tableName) {
+    String columnName = dataFetchingEnvironment.getArgument("columnName");
+
+    if (keyspaceName == null) {
+      List<String> parts = Splitter.on('.').splitToList(tableName);
+      if (parts.size() < 2) {
+        throw new IllegalArgumentException("Missing field argument keyspaceName @ 'createIndex'");
+      }
+      keyspaceName = parts.get(0);
+      tableName = parts.get(1);
```
olim7t, line 58: Why not, but if we do this here, we should revisit all the table and type fetchers for uniformity. I'm not sure if it's worth it, seems simple enough to pass the keyspace separately.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/CreateIndexFetcher.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    String indexName =
+        dataFetchingEnvironment.getArgumentOrDefault(
+            "indexName", tableName + "_" + columnName + "_idx");
```
olim7t, line 63: Don't generate a name if it's not provided, just omit the `builder.index()` call (passing `null` is also valid I think). It's optional, Cassandra will generate one.

Otherwise there is a subtle bug here: special characters are allowed in column names, but not in index names. For example:
```
mutation {
  createTable(
    keyspaceName: "test"
    tableName: "foo"
    partitionKeys: [{ name: "k", type: { basic: INT } }]
    values: [{ name: "the value", type: { basic: INT } }]
  )
  createIndex(keyspaceName: "test", tableName: "foo", columnName: "the value")
}
```
Produces the error:
```
Index name must not be empty, more than 2147483647 characters long, or contain non-alphanumeric-underscore characters (got \"foo_the value_idx\"
```

olim7t (changes requested):

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/CreateIndexFetcher.java
eribeiro: @olim7t I have tried to address your review comments, but I am starting to learn GraphQL (and Stargate codebase), so please let me know if I misunderstood something. Particularly, the GraphQL enum declaration and usage. Thanks!

eribeiro:
@olim7t I have tried to address your review comments, but I am starting to learn GraphQL (and Stargate codebase), so please let me know if I misunderstood something. Particularly, the GraphQL enum declaration and usage. Thanks!

graphqlapi/src/main/java/io/stargate/graphql/schema/DdlSchemaBuilder.java
eribeiro: Please, let me know if I have addressed this review comment accordingly. I am still getting acquainted with GraphQL. :)

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/IndexFetcher.java
eribeiro: Done. 👍

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/CreateIndexFetcher.java
eribeiro: I have removed the optional keyspaceName. 👍

olim7t:
Looks good, the GraphQL enum is what I had in mind 👍

eribeiro:
@olim7t Awesome! 😄 Btw, sincere question: this PR would suffice to PR #80, too?

PS: had to remove a last minute left over snippet.

dougwettlaufer:
FYI @polandll you can use this in your graphql examples now instead of dropping down to cql.

PR #565, opened 2021-01-05
Pass headers to the AuthN and AuthZ

tomekl007:


cql/src/main/java/org/apache/cassandra/stargate/transport/internal/ServerConnection.java
```diff
@@ -129,8 +130,11 @@ void applyStateTransition(Message.Type requestType, Message.Type responseType) {
               .getAuthenticator()
               .newSaslNegotiator(clientInfo.remoteAddress().getAddress(), certificates());
 
+      // todo how to pass headers here
```
tomekl007, line 133: @dougwettlaufer 
The `SaslNegotiator` is problematic here because c2 is initializing the negotiator with `orgId` (that needs to be retrieved using `tenant_id`).
Because of that, we need to pass the headers to the `AuthenticationService#getSaslNegotiator` method.
From what I see, the `ServerConnection` is created in the `Server#initChannel`. Is the channel created for every user request? 
- If yes, then maybe we can propagate and extract this information from `ClientInfo clientInfo`?
- If no and the `ServerConnection` is shared between multiple requests, it means that we need to carry the `headers` in the `clientResponse`:
```
 interface SaslNegotiator {
    byte[] evaluateResponse(byte[] clientResponse)
}
```
or modify this method to take headers.
to be able to extract headers for every request.
wdyt?

tomekl007 (commented):
@dougwettlaufer 
I've refactored the code to propagate headers from `HttpServletRequest` to `AuthN` service.
I think that we should not propagate `HttpServletRequest` directly because not all modules that are using AuthZ may need to depend on the `javax.servlet.api`.
This refactor allows the implementations of `AuthenticationService` to extract the `tenant_id` and use it.
Please let me know if this is a good direction. If it is, I will refactor `AuthorizationService` in a similar way.

I am still not sure:
- [ ] how to tackle the `SaslNegotiator` init problem
- [ ] extract headrs (tenant_id) for the CQL (should we use `HAProxyProtocolDetectingDecoder`)?

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/ServerConnection.java
dougwettlaufer: You'll want to sync up with @tjake regarding his change here https://github.com/stargate/stargate/pull/564/files

tomekl007:
Closing in favor of https://github.com/stargate/stargate/pull/581

PR #573, opened 2021-01-06
Add Starter options to allow running NodeTool

dimas-b:
Also add a helper nodetool script.

Expose JMX port in StargateConnectionInfo for tests.

### DIFF
--- a/stargate-starter/src/main/java/io/stargate/starter/Starter.java
+++ b/stargate-starter/src/main/java/io/stargate/starter/Starter.java
@@ -19,17 +19,24 @@ import com.github.rvesse.airline.HelpOption;
 import com.github.rvesse.airline.SingleCommand;
 import com.github.rvesse.airline.annotations.Command;
 import com.github.rvesse.airline.annotations.Option;
+import com.github.rvesse.airline.annotations.Parser;
 import com.github.rvesse.airline.annotations.help.License;
 import com.github.rvesse.airline.annotations.restrictions.Port;
 import com.github.rvesse.airline.annotations.restrictions.Required;
 import com.github.rvesse.airline.help.cli.CliCommandUsageGenerator;
+import com.github.rvesse.airline.model.OptionMetadata;
 import com.github.rvesse.airline.parser.ParseResult;
+import com.github.rvesse.airline.parser.ParseState;
 import com.github.rvesse.airline.parser.errors.ParseException;
+import com.github.rvesse.airline.parser.options.AbstractOptionParser;
+import io.stargate.starter.Starter.NodeToolOptionParser;
 import java.io.File;
 import java.io.IOError;
 import java.io.IOException;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.nio.file.ClosedWatchServiceException;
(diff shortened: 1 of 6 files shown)


stargate-starter/src/main/java/io/stargate/starter/Starter.java
```diff
@@ -373,6 +392,28 @@ public void start() throws BundleException {
       Bundle b = context.installBundle(jar.toURI().toString());
       bundleList.add(b);
     }
+
+    if (nodetool) {
+      for (Bundle bundle : bundleList) {
+        try {
+          Class<?> tool = bundle.loadClass("org.apache.cassandra.tools.NodeTool");
```
dougwettlaufer, line 399: Do we just try to run it on every bundle until we find it? Would it be possible to short circuit and go direct to the bundle rather than looping?

dimas-b: yes. I had that initially, but for some reason went back to the loop. I'll update to use the persistence bundle directly.

dougwettlaufer: Oh I see down below where you put it as the first one

nodetool
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+set -e
+
+# If STARGATE_HOME wasn't specified in the environment, then search for one...
+if [ "x$STARGATE_HOME" == "x" ]; then
+    STARGATE_HOME=`dirname "$0"`
```
dougwettlaufer, line 6: Doesn't this need the directory like starctl?

```
STARGATE_HOME="`dirname "$0"`/stargate-lib"
```

nodetool
```diff
@@ -0,0 +1,39 @@
...
+      exit 2
+    fi
+fi
+
+declare -a CMD=("${STARGATE_HOME}/starctl" "--nodetool")
+
+# If the Stargate LISTEN address is set use it for the --host nodetool option by default
+if [[ ! -z "$LISTEN" ]]; then
+    CMD+=("-T" "--host" "-T" "$LISTEN")
+fi
```
dougwettlaufer, line 23: Would this make sense to do in Starter.java instead since its typically not set in docker?

```
    if [[ ! -z "$LISTEN" ]]; then
        STARGATE_ARGS+=("--listen" "$LISTEN")
    else
        LISTEN="$(_ip_address)"
        STARGATE_ARGS+=("--listen" "$LISTEN")
    fi
```

dimas-b: Probably... I'll give it a try.

nodetool
```diff
@@ -0,0 +1,39 @@
...
+done
+
+# Use environment variables to determine persistence type and version
+if [[ ! -z "$DSE" ]]; then
+    CMD+=("--dse")
+fi
+
+if [[ ! -z "$CLUSTER_VERSION" ]]; then
+    CMD+=("--cluster-version" "$CLUSTER_VERSION")
+fi
```
dougwettlaufer, line 37: Is this the only config needed? If so does it make sense to reuse starctl and Starter.java or would it be cleaner to introduce a new Nodetool.java?

dougwettlaufer (approved):
Few small comments but looks good otherwise

nodetool
dimas-b: We only need those options that are required for selecting the persistence bundle (I guess it makes sense to use the same `NodeTool` version as the backend).

Adding a new command class seemed too invasive to me, mainly because `rvesse.airline` does not have an easy way to group common options (at least I could not find it). All options have to be declared on the class hierarchry, which makes extending harder - we'd have to duplicate all command classes wherever they are extended.

dimas-b:
@dougwettlaufer : I've reworked this PR a bit - removed the `nodetool` script in favour of a custom option parser in `Starter`. I hope it's more user-friendly and does not require extra shell scripts.

dougwettlaufer (approved):

Issue #239, opened 2020-10-09
Create an http-utility module


dougwettlaufer:
As was discussed in https://github.com/stargate/stargate/pull/235/files#r502590739 it would be good if we created a module to consolidate common dependencies and code used by our http modules.

ivansenic:
include `HttpMetricsTagProvider` as agreed in the https://github.com/stargate/stargate/pull/836#discussion_r613440525

tatu-at-datastax:
Obsolete due to work for Stargate V2.

PR #581, opened 2021-01-07
Pass the tenant

tomekl007:


tomekl007:
It's based on the https://github.com/stargate/stargate/pull/581

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -192,7 +194,8 @@ public ResultMessage processPrepared(
     }
 
     if (customPayload != null && customPayload.containsKey("token")) {
-      authorizeByToken(customPayload, statement);
+      // todo extract tenant_id and map to header
+      authorizeByToken(customPayload, statement, Collections.emptyMap());
```
tomekl007, line 198: There is a problem with accessing tenant_id here (same for `StargateQueryHandler` for C* 4 and DSE).
In all places that I marked with ` // todo extract tenant_id and map to header` we need extract the tenant_id somehow.
In this context, we have access to `  QueryState queryState,` which has the `clientState`.
Unfortunately, the clientState does not carry the same info as `ClientInfo`. It has only `remoteAddress` and no public address(`sourceAddress`) (`queryState.getClientState().getRemoteAddress()`). 
This PR: https://github.com/stargate/stargate/pull/564/files propagates the tenant_id using `sourceAddress`.
@tjake Is there a way to inject that information to `queryState` as well?

graphqlapi/src/main/java/io/stargate/graphql/web/RequestToHeadersMapper.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.graphql.web;
+
+import java.util.*;
+import javax.servlet.http.HttpServletRequest;
+
+public class RequestToHeadersMapper {
+  public static Map<String, String> getAllHeaders(HttpServletRequest request) {
```
tomekl007, line 22: This logic needs to be duplicated because of this:

In the stargate we are using two dependencies that are providing the `HttpServletRequest` class:
- `jakarta.servlet-api`
- `javax.servlet:javax.servlet-api`

I want to add logic to the core module that uses this class.
Unfortunately, it does not work with OSGIi; it fails with:
```INFO  [Exec Stream Pumper] 2021-01-08 11:12:04,337 StargateContainer.java:435 - sg0-0> ! java.lang.LinkageError: loader constraint violation: loader (instance of org/apache/felix/framework/BundleWiringImpl$BundleClassLoader) previously initiated loading for a different type with name "javax/servlet/http/HttpServletRequest"```
because the `HttpServletRequest` class is sometimes (for some modules) javax and for some others it is Jakarta. Does someone observe this problem?
Shouldn’t we use the same implementation for all modules? I need to use the same logic that uses `HttpServletApi` in `restapi`, `grapqhlapi`, and `auth-api`. I solved this problem by duplicating small part of the logic that uses `HttpServletRequest` but it is not ideal.

File: cql/src/main/java/org/apache/cassandra/stargate/transport/internal/RequestToHeadersMapper.java [hunk 3]
```diff
@@ -0,0 +1,34 @@
...
+
+public class RequestToHeadersMapper {
+  public static final String TENANT_ID_HEADER_NAME = "tenant_id";
+
+  @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
+  public static Map<String, String> toHeaders(
+      Optional<InetSocketAddress> sourceAddressWithEncodedTenantId) {
+    Map<String, String> tenantIdHeaders = new HashMap<>();
+    sourceAddressWithEncodedTenantId
+        .map(tenantIdAddress -> UUID.nameUUIDFromBytes(tenantIdAddress.getAddress().getAddress()))
```
dimas-b, line 29: I'm not sure this conversion is correct.

The proxy protocol [handler](https://github.com/stargate/stargate/blob/b04b68424634ca50c1f1fcae4b9aa966f6827c7c/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/HAProxyProtocolDetectingDecoder.java#L73) creates the address as `new InetSocketAddress(msg.sourceAddress(), msg.sourcePort())`, so I believe the tenant ID is the host _name_ portion of this address. The binary address may be different depending on how load balancing is set up.

I'd assume we ought to use UUID.fromString(tenantIsAddress.getHostName())`

@tjake WDYT?

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/RequestToHeadersMapper.java
```diff
@@ -0,0 +1,34 @@
...
+        .ifPresent(
+            tenantIdUUID -> tenantIdHeaders.put(TENANT_ID_HEADER_NAME, tenantIdUUID.toString()));
```
dimas-b, line 31: I'm not sure this conversion is correct.

The proxy protocol [handler](https://github.com/stargate/stargate/blob/b04b68424634ca50c1f1fcae4b9aa966f6827c7c/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/HAProxyProtocolDetectingDecoder.java#L73) creates the address as `new InetSocketAddress(msg.sourceAddress(), msg.sourcePort())`, so I believe the tenant ID is the host _name_ portion of this address. The binary address may be different depending on how load balancing is set up.

I'd assume we ought to use UUID.fromString(tenantIsAddress.getHostName())`

@tjake WDYT?

dimas-b (changes requested):

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/RequestToHeadersMapper.java
dimas-b: In this case we do not even need to parse the UUID string, I think.

authnz/src/main/java/io/stargate/auth/AuthenticationService.java
```diff
@@ -16,14 +16,17 @@
 package io.stargate.auth;
 
 import io.stargate.db.Authenticator.SaslNegotiator;
+import java.util.Map;
 
 public interface AuthenticationService {
 
-  String createToken(String key, String secret) throws UnauthorizedException;
+  String createToken(String key, String secret, Map<String, String> headers)
```
dimas-b, line 23: I'd rename `heads` to `properties`, for the sake of clarify since they do not always come from HTTP header, nor are they required to be HTTP headers, AFAIK.

dimas-b: Suggestion: it might be worth defining a separate holder class for "headers" and also make it encapsulate the `key` and the `secret` WDYT?

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/RequestToHeadersMapper.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.net.InetSocketAddress;
+import java.util.*;
+
+public class RequestToHeadersMapper {
+  public static final String TENANT_ID_HEADER_NAME = "tenant_id";
+
+  @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
+  public static Map<String, String> toHeaders(
+      Optional<InetSocketAddress> sourceAddressWithEncodedTenantId) {
```
dimas-b, line 26: From my POV it would be more intuitive to use `ClientInfo` as a parameter (or the extracted ID `String`) since the way we determine tenant ID is very specific to how `ClientInfo` is constructed. Any `InetSocketAddress` would be be suitable.

authnz/src/main/java/io/stargate/auth/AuthorizationService.java
```diff
@@ -42,7 +44,8 @@ ResultSet authorizedDataRead(
       String keyspace,
       String table,
       List<TypedKeyValue> typedKeyValues,
-      SourceAPI sourceAPI)
+      SourceAPI sourceAPI,
+      Map<String, String> headers)
```
dimas-b, line 48: I'm not sure the `headers` is justified in this context. I believe all the information required for authorization should be contained in the `AuthenticationSubject` object, which is already defined in this method.

dimas-b: Basically, I think we ought to add tenant ID (plus whatever else is needed) to `AuthenticationSubject`. We already reference tenant ID in `RequestToHeadersMapper` so it's not completely transparent and it would be nice to make tenant ID handling explicit.

dimas-b: If the list of extra properties is not know, we could add a `properties` object of type `Map<String, String>` to `AuthenticationSubject`, but I think it is important to keep authorization inputs together. That is the pattern in the JRE auth APIs, cf. `javax.security.auth.Subject`. WDYT?

authnz/src/main/java/io/stargate/auth/AuthenticationService.java
dimas-b: Another point of having a custom "parameters" or perhaps "context" object: Currently there are two distinct use cases: 1) for HTTP-based APIs 2) for CQL. The latter uses the "source host name" from `ClientInfo` for defining the tenant. So, rather than converting everything to the headers map, we could have both pieces of data defined independently in the "context" object. The auth implementation will use one of them (or both) as appropriate. The benefit would be clearer separation of concerns, e.g. "tenant ID" would not have be an common entity. Only those auth implementations that have that notion will have code dealing with it. Other implementation will have to be aware of that concept.

dimas-b: Similarly, `AuthenticationSubject` would only contain a tenant "Principal" when tenants are in use. Other authorization implementations will not need to be aware of tenants. At java API level we'd have a generic collection of "Principal" objects.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/TableFetcher.java
```diff
@@ -39,7 +42,7 @@ protected TableFetcher(
     }
 
     authorizationService.authorizeSchemaWrite(
-        authenticationSubject, keyspaceName, tableName, scope, SourceAPI.GRAPHQL);
+        authenticationSubject, keyspaceName, tableName, scope, SourceAPI.GRAPHQL, EMPTY_HEADERS);
```
dimas-b, line 45: I'm not sure using `EMPTY_HEADERS` here is correct. In case authentication was tenant-based, do we not have to provide tenant info in this context too?

By the way, if we keep tenant ID in the `authenticationSubject` this class will not have to change at all, I believe.

graphqlapi/src/main/java/io/stargate/graphql/web/RequestToHeadersMapper.java
dimas-b: We could use an embedded jar (cf. `persistence-common`) as opposed to an OSGi bundle in this case. The classes in the embedded jar will be resolved against the servlet API version specific to each containing bundle. That should be successful as long at the servlet API version is compatible.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -238,7 +240,8 @@ protected void authorizeByToken(Map<String, ByteBuffer> customPayload, CQLStatem
             authenticationSubject,
             castStatement.keyspace(),
             castStatement.columnFamily(),
-            SourceAPI.CQL);
+            SourceAPI.CQL,
+            EMPTY_HEADERS);
```
dimas-b, line 244: Same concern as in https://github.com/stargate/stargate/pull/581/files#r559749974

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -217,7 +219,8 @@ protected void authorizeByToken(Map<String, ByteBuffer> customPayload, CQLStatem
             authenticationSubject,
             castStatement.keyspace(),
             castStatement.columnFamily(),
-            SourceAPI.CQL);
+            SourceAPI.CQL,
+            EMPTY_HEADERS);
```
dimas-b, line 223: Same concern as in https://github.com/stargate/stargate/pull/581/files#r559749974

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateQueryHandler.java
```diff
@@ -191,7 +193,11 @@ protected void authorizeByToken(Map<String, ByteBuffer> customPayload, CQLStatem
 
       try {
         authorization.authorizeDataRead(
-            authenticationSubject, castStatement.keyspace(), castStatement.table(), SourceAPI.CQL);
+            authenticationSubject,
+            castStatement.keyspace(),
+            castStatement.table(),
+            SourceAPI.CQL,
+            EMPTY_HEADERS);
```
dimas-b, line 200: Same concern as in https://github.com/stargate/stargate/pull/581/files#r559749974

restapi/src/main/java/io/stargate/web/resources/Db.java
```diff
@@ -36,7 +38,7 @@
   private final DataStore dataStore;
   private final AuthenticationService authenticationService;
   private final AuthorizationService authorizationService;
-  private final LoadingCache<String, AuthenticationSubject> docsTokensToRoles =
+  private final LoadingCache<TokenAndHeaders, AuthenticationSubject> docsTokensToRoles =
```
dimas-b, line 41: If we use _all_ headers at the key, is it going to be different for every request? I suspect some of the HTTP headers may be different in each request... or at least likely to be different, e.g. `Content-Length`... WDYT?

dimas-b: Again, if we store auth data inside `AuthenticationSubject` we should be able to continue using only the token as the key here, I believe.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
tjake: Each backend needs to have a ClientStateWithTenant subclass that is setup when the connection is created.

tomekl007: yes, it is using the `ClientState` now. This comment got outdated, resolving now.

graphqlapi/src/main/java/io/stargate/graphql/web/RequestToHeadersMapper.java
tomekl007: We should postpone it to a separate ticket since this one is a high priority, and this code duplication is not blocking.There is already ticket to capture that:
https://github.com/stargate/stargate/issues/239

authnz/src/main/java/io/stargate/auth/AuthenticationService.java
tomekl007: We were discussing that with @dougwettlaufer and it was suggested to pass the map of headers because this mechanism is generic:
> I'd rename headers to properties, for the sake of clarify since they do not always come from HTTP header, nor are they required to be HTTP headers, AFAIK.

The same comment, we decided to stick to the headers. The CQL use case is a special case and we can make it generic by using the generic map I think. 

>Suggestion: it might be worth defining a separate holder class for "headers" and also make it encapsulate the key and the secret WDYT?

both `key` and `secret` are not related to changes in this PR. I would like to reduce the number of refactoring changes here; we can address is separate.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/RequestToHeadersMapper.java
tomekl007: done

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
tomekl007: For stargate query handlers we should pass `EMPTY_HEADERS`. Only the CNDB `StargasteQueryHandler` will extract the tenant id from `ClientStateWithTenant` and pass it to the components. It was discussed on slack.

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
tomekl007: For stargate query handlers we should pass EMPTY_HEADERS. Only the CNDB StargasteQueryHandler will extract the tenant id from ClientStateWithTenant and pass it to the components. It was discussed on slack.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateQueryHandler.java
tomekl007: For stargate query handlers we should pass EMPTY_HEADERS. Only the CNDB StargasteQueryHandler will extract the tenant id from ClientStateWithTenant and pass it to the components. It was discussed on slack.

authnz/src/main/java/io/stargate/auth/AuthorizationService.java
tomekl007: Please note that `headers` are just passing through. They are not used in the `Stargate` implementations at all. Also, the `tenant_id` is needed not only for IAM but also for example billing. By including that information in the `AuthenticationSubject` we will hide it from other callers. I think that this is one of the main intentions of why we want to pass it separately.

graphqlapi/src/main/java/io/stargate/graphql/schema/fetchers/ddl/TableFetcher.java
tomekl007: I missed that one, thx for catching it. Fixed.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/RequestToHeadersMapper.java
tomekl007: I think that we need, for example, the IPv6 will look like this: `3ffe:0:0:1:200:f8ff:fe75:50df` whereas the tenant_id should be UUID (with `-`)

restapi/src/main/java/io/stargate/web/resources/Db.java
tomekl007: Good point, we should construct key based only on token and tenant_id and Host headers (if they are present)

auth-jwt-service/src/main/java/io/stargate/auth/jwt/PlainTextJwtTokenSaslNegotiator.java
```diff
@@ -21,20 +21,24 @@
...
       LoggerFactory.getLogger(PlainTextJwtTokenSaslNegotiator.class);
+  private final Map<String, String> headers;
 
   public PlainTextJwtTokenSaslNegotiator(
       AuthenticationService authentication,
       SaslNegotiator wrapped,
       String tokenUsername,
-      int tokenMaxLength) {
+      int tokenMaxLength,
+      Map<String, String> headers) {
```
mpenick, line 39: Not a big deal, but this is weird passing  "HTTP headers" around in the CQL code. Because of this I lean toward calling it "properties" instead of "headers".

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/RequestToHeadersMapper.java
mpenick: Is the tenant UUID encoded as a IPv6 address here? If so, then using `UUID.nameUUIDFromBytes()` is not correct.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/PlainTextJwtTokenSaslNegotiator.java
tomekl007: Nothing in the name imposes that those are `http headers`. There are just `headers`. I can change it to properties if you prefer. But in that case, the name should be changed only in the `cql` related code, right? The non-CQL code should still name this parameter as `headers`?

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/RequestToHeadersMapper.java
mpenick: We should add a comment that states that the raw bytes of the IPv6 address are the UUID because it's out of the ordinary and will catch unfamiliar developers off guard.

tomekl007: @mpenick what do you mean by `UUID.nameUUIDFromBytes` is not correct.
Based on our discussion on slack, the Ipv6 bytes are UUID. To implement that:
1. We need to extract those bytes, using: `tenantIdAddress.getAddress().getAddress()`
2. Transform bytes to UUID, and this method `UUID.nameUUIDFromBytes` seems to be designed for it.
Could you elaborate more?

dimas-b: +1 to documenting and I assume we'd use `new UUID(long mostSigBits, long leastSigBits)` in that case?

authnz/src/main/java/io/stargate/auth/AuthenticationService.java
dimas-b: Ultimately, headers are required for authentication, right? The result of a successful authentication is an `AuthenticationSubject`, so why not record the headers inside `AuthenticationSubject` for later use during authorization? I believe this would be consistent with our auth model and might also require less code changes.

Regarding "principals" that is a secondary point. I'd be ok with using "headers", but I'm not sure it's worth putting them as separate parameters into authorization API methods :thinking:

mpenick (approved):

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/RequestToHeadersMapper.java
tomekl007: I've implemented it according to the slack discussion, I've also added a comment describing it. Please take a look.

dimas-b (approved):

authnz/src/main/java/io/stargate/auth/AuthenticationService.java
tomekl007: I am +1 on trying that path and compare to the current design. However, I feel that because of the timeline we should do it after the merge and initial testing. Feel free to create a follow-up ticket.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthzJwtService.java
```diff
@@ -58,7 +59,8 @@ public ResultSet authorizedDataRead(
       String table,
       List<TypedKeyValue> typedKeyValues,
       SourceAPI
-          sourceAPI) // this isn’t supported but if you want to use it you’ll need something other
+          sourceAPI, // this isn’t supported but if you want to use it you’ll need something other
+      Map<String, String> headers)
       // than a JWT
```
dougwettlaufer, line 64: This comment needs fixed.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/PlainTextJwtTokenSaslNegotiator.java
dougwettlaufer: +1 for anything cql related being referred to as properties.

auth-jwt-service/src/test/java/io/stargate/auth/jwt/AuthzJwtServiceTest.java
```diff
@@ -98,7 +99,8 @@ public void executeDataReadWithAuthorization() throws Exception {
             "keyspace",
             "table",
             typedKeyValues,
-            SourceAPI.CQL);
+            SourceAPI.CQL,
+            EMPTY_HEADERS);
```
dougwettlaufer, line 103: Still going through but do we have tests anywhere that pass something other than an empty map?

tomekl007: it is verified in C2 tests by passing tenant_id and verifying the full flow. I think the c2 is a better place for those tests because it actually uses the passed headers.

authnz/src/main/java/io/stargate/auth/AuthorizationService.java
```diff
@@ -32,6 +33,7 @@
    * @param typedKeyValues A list of {@link TypedKeyValue} that will be used in the query and should
    *     be authorized against the token.
    * @param sourceAPI The source api which calls this method.
+   * @param headers The map of headers propagated from the request (HTTP, GraphQl, ...)
```
dougwettlaufer, line 36: Should probably clarify that this pertains to headers for the HTTP services and properties for CQL. This comment implies these are only necessary for rest/graphql.

authnz/src/main/java/io/stargate/auth/AuthorizationService.java
dougwettlaufer: I agree with @dimas-b that a more generic term would be better but perhaps we should create an issue in order to come back and refactor?

authnz/src/main/java/io/stargate/auth/AuthorizationService.java
tomekl007: done

auth-jwt-service/src/main/java/io/stargate/auth/jwt/PlainTextJwtTokenSaslNegotiator.java
tomekl007: I am not sure about boundaries where they should be called `headers` and where `properties`.
For example, `PlainTextJwtTokenSaslNegotiator` is not in the `CQL` module (or package) so I am not sure if this can be used only in the cql context.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/RequestToHeadersMapper.java
dougwettlaufer: This is very implementation specific logic. Can we add the raw header to the map and deal with this conversion farther down where it's actually used?

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dougwettlaufer: This should pass all the cql properties. This logic needs to be generic and apply to all backends.

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -94,6 +95,7 @@
   private static final Logger logger = LoggerFactory.getLogger(StargateQueryHandler.class);
   private final List<QueryInterceptor> interceptors = new CopyOnWriteArrayList<>();
   private AtomicReference<AuthorizationService> authorizationService;
+  private static final Map<String, String> EMPTY_HEADERS = Collections.emptyMap();
```
dougwettlaufer, line 98: Same comment as 3.11. We should be passing the properties/headers in all cases.

tomekl007: I don't understand this comment. What do you mean? Do you mean that we should rename it?

restapi/src/main/java/io/stargate/web/resources/Db.java
```diff
@@ -136,4 +145,42 @@ public DocumentDB getDocDataStoreForToken(String token, int pageSize, ByteBuffer
...
+    private static final String TENANT_ID_HEADER = "tenant_id";
+    private static final String HOST_HEADER = "Host";
+    private final String token;
+    private final Map<String, String> headers;
+
+    static TokenAndHeaders create(String token, Map<String, String> headers) {
+      return new TokenAndHeaders(token, filterHeaders(headers));
+    }
+
+    private static Map<String, String> filterHeaders(Map<String, String> headers) {
```
dougwettlaufer, line 159: Why are we filtering headers? We need to make this work for everyone.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
tomekl007: We were discussing this on the #stargate-team. We don't have access to headers nor `ClientState` here. The conclusion was that we need to extract it from the `ClientState` and pass further only for the CNDB's `StargateQueryHandler`

restapi/src/main/java/io/stargate/web/resources/Db.java
tomekl007: this is for building the cache key. We want the cache to be per `token`, `Host`, and `tenant_id`. 
It was discussed here:
https://github.com/stargate/stargate/pull/581#discussion_r559756983

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
tomekl007: Implemented in:
https://github.com/stargate/stargate/pull/581/commits/05d379881dcf18262d3d07eaf21ccb33dad3960c

restapi/src/main/java/io/stargate/web/resources/Db.java
tomekl007: Implemented in:
https://github.com/stargate/stargate/pull/581/commits/05d379881dcf18262d3d07eaf21ccb33dad3960c

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/RequestToHeadersMapper.java
tomekl007: Our API assumes that we need to pass the generic `Map<String, String>`. Because of that reason, I think we need to transform the address to a `String` format to not complicate the API more. Otherwise, we would need to propagate it as a `byte[]` and we don't want to do it.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/PlainTextJwtTokenSaslNegotiator.java
dougwettlaufer: Yes, that is only used for CQL

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dougwettlaufer: https://github.com/stargate/stargate/pull/581#discussion_r560285601

authnz/src/main/java/io/stargate/auth/AuthorizationService.java
```diff
@@ -54,13 +58,16 @@ ResultSet authorizedDataRead(
    * @param keyspace The keyspace containing the table with data to be read.
    * @param table The table within the provided keyspace containing the data to be read.
    * @param sourceAPI The source api which calls this method.
-   * @throws UnauthorizedException An exception relating to the failure to authorize.
+   * @param headers The map of headers propagated from the request (HTTP, GraphQl, ...) or
+   *     properties propagated from CQL.* @throws UnauthorizedException An exception relating to the
+   *     failure to authorize.
```
dougwettlaufer, line 63:
```suggestion
   * @param headers The map of headers propagated from the request (HTTP, GraphQl, ...) or
   *     properties propagated from CQL.* 
   * @throws UnauthorizedException An exception relating to the failure to authorize.
```

authnz/src/main/java/io/stargate/auth/AuthorizationService.java
```diff
@@ -74,14 +81,17 @@ void authorizeDataRead(
    * @param table The table within the provided keyspace containing the data to be modified.
    * @param scope The table within the provided keyspace that is being modified.
    * @param sourceAPI The source api which calls this method.
-   * @throws UnauthorizedException An exception relating to the failure to authorize.
+   * @param headers The map of headers propagated from the request (HTTP, GraphQl, ...) or
+   *     properties propagated from CQL.* @throws UnauthorizedException An exception relating to the
+   *     failure to authorize.
```
dougwettlaufer, line 86:
```suggestion
   * @param headers The map of headers propagated from the request (HTTP, GraphQl, ...) or
   *     properties propagated from CQL.
   * @throws UnauthorizedException An exception relating to the failure to authorize.
```

authnz/src/main/java/io/stargate/auth/AuthorizationService.java
dougwettlaufer: Actually just feel free to fix all these in a follow up.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/RequestToHeadersMapper.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package io.stargate.db.cassandra.impl;
+
+import java.nio.ByteBuffer;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+public class RequestToHeadersMapper {
+  public static final String TENANT_ID_HEADER_NAME = "tenant_id";
```
dougwettlaufer, line 24: The `tenant_id` doesn't apply in this context. Instead these should be something like `host` or `publicAddress`.

dougwettlaufer: Same for 4.0 and DSE.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/RequestToHeadersMapper.java
dougwettlaufer: But can we just pass around the host? Should really be working with publicAddresses rather than tenantIDs.

restapi/src/main/java/io/stargate/web/resources/Db.java
dougwettlaufer: The host contains the tenantID so we don't need both, only the host header.

tomekl007: For `cql` it will contain tenant_id. So it can be one or another.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/RequestToHeadersMapper.java
tomekl007: the `tenant_id` is an actual piece of information that we want to propagate. The caller doesn't care nor should care how it was provided. For now, it is extracted from the `publicAddress` but it may change in the future. We should treat the `tenant_id` header name as the API contract exposed by our code. If we would call it `publicAddress` we would tight couple those concepts. Because it is `tenant_id` we are abstracting away how it is provided. Also, the c2 logic is based on the `tenant_id` header name. wdyt?

restapi/src/main/java/io/stargate/web/resources/Db.java
dougwettlaufer: This is only for the restapi module so it should be dealing with just the host header. We should avoid referencing the tenantID as much as possible to avoid confusion for anyone trying to modify this code.

tomekl007: good point, done

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/RequestToHeadersMapper.java
dougwettlaufer: But if we can deal with just the host that would be much better for users. No one else will use nor understand `tenant_id`

dimas-b: `host` would also be misleading in this context, because it would not be a real host name

dimas-b: I do not insist, but I do believe that if we want to have cleaner separation of concerns it might be preferable to pass extra auth data via the `AuthenticatedSubject`. We could add an opaque set of properties for simplicity at this point.

tomekl007: Feel free to implement that OR change the header name. For both cases, the C2 code needs to be adapted.

tomekl007:
Closing in favor of https://github.com/stargate/stargate/pull/604

PR #585, opened 2021-01-09
Use proxy path in Swagger examples (fixes #563)

### LINKED ISSUE #563: Swagger should use proxy path in examples
In https://github.com/stargate/stargate/pull/441 we updated swagger to load when running behind a proxy but that proxy path is not reflected in the examples. This means whenever you try one of the endpoints a 404 is returned. 

To fix this we need to apply the same base URI for the examples as we already are using when loading swagger.json. We should be able to do something with the basePath field.

olim7t:
If Stargate is behind a proxy that adds a prefix to the URL path, ensure that "Try it out" requests use the correct URLs.

### DIFF
--- a/restapi/src/main/resources/swagger-ui-cust/index.html
+++ b/restapi/src/main/resources/swagger-ui-cust/index.html
@@ -34,12 +34,9 @@
 <script src="./swagger-ui-standalone-preset.js" charset="UTF-8"></script>
 <script>
     window.onload = function() {
-      const swaggerUri = window.location.href
-      const baseUri = swaggerUri.substring(0, swaggerUri.lastIndexOf('/swagger-ui'))
 
-      // Begin Swagger UI call region
-      const ui = SwaggerUIBundle({
-        url: baseUri + "/swagger.json",
+      const swaggerOptions = {
+        url: "/swagger.json",
         dom_id: '#swagger-ui',
         deepLinking: true,
         presets: [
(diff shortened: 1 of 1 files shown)


olim7t:
Tested locally with a small js proxy:
```js
var http = require('http'),
  httpProxy = require('http-proxy');

var proxy = httpProxy.createProxyServer({});

var server = http.createServer(function(req, res) {
  var url = req.url;
  if (url.startsWith("/stargate")) {
    req.url = url.slice(9)
    proxy.web(req, res, { target: 'http://127.0.0.2:8082' });
  } else {
    res.writeHead(404, {"Content-Type": "text/plain"});
    res.write("Not found: " + url);
    res.end();
  }
});

console.log("listening on port 10000")
server.listen(10000);
```

restapi/src/main/resources/swagger-ui-cust/index.html
```diff
@@ -56,7 +53,26 @@
             return tokenValue == null ? "" : tokenValue.trim();
           }
         }
-      })
+      };
+
+      const pathname = window.location.pathname;
+      if (pathname !== '/swagger-ui') {
```
dimas-b, line 59: Could we support proxied URLs without an `if` statement? Could we make the logic to work with empty prefixes?

I believe it would be more robust to avoid forks in the processing logic. WDYT?

olim7t: We can remove that check, but then we would register a request interceptor that does nothing.

dimas-b: that may still happen if line 68 is executed without a proxy? :thinking:

olim7t: If there is no proxy, `/swagger-ui` should be exposed at the root, and then we don't enter the if block and don't install the request interceptor.

dimas-b: Ah, TIL :)

dimas-b (approved):

mpenick (approved):

PR #598, opened 2021-01-15
Change default TPC core count for Stargate

tjake:


jimdickinson (approved):
👍

dimas-b (approved):

persistence-dse-6.8/src/main/java/io/stargate/db/dse/DsePersistenceActivator.java
```diff
@@ -112,6 +113,7 @@ public static Config makeConfig(File baseDir) throws IOException {
     c.storage_port = listenPort;
     c.listen_address = listenAddress;
     c.broadcast_address = listenAddress;
+    c.tpc_cores = FBUtilities.getAvailableProcessors();
```
dimas-b, line 116: Why not `FBUtilities.getAvailableProcessors() - 1`?

mpenick: Isn't this already done if tpc_cores is not set? Looking for the source code line. I've also tested it by running it in docker with different numbers for `--cpus`.

mpenick (approved):

persistence-dse-6.8/src/main/java/io/stargate/db/dse/DsePersistenceActivator.java
tjake: Since SG is coordinator only it can use all cores.

PR #600, opened 2021-01-16
Run DSE tests first in CI

dimas-b:


tomekl007 (approved):
Very good idea to run those most unstable tests firstly.

dougwettlaufer (approved):

PR #601, opened 2021-01-16
Exposed root auth failure messages

dimas-b:


dougwettlaufer (commented):
Looks good but would suggest we do it on the other two as well for consistency

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateQueryHandler.java
```diff
@@ -332,8 +332,9 @@ private void authorizeAuthenticationStatement(
         authorization.authorizeRoleManagement(
             authenticationSubject, role, grantee, scope, SourceAPI.CQL);
       } catch (io.stargate.auth.UnauthorizedException e) {
+        logger.debug("Unauthorized statement: " + scope, e);
         throw new UnauthorizedException(
-            String.format("Missing correct permission on role %s", role));
+            String.format("Missing correct permission on role %s: %s", role, e.getMessage()));
```
tomekl007, line 337: Isn't there any IT/unit-test that validates the exception message? I don't see any test changes in the stargate.

tomekl007 (approved):

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateQueryHandler.java
dimas-b: unit tests added

dimas-b:
Added similar changes to C* 3.11 and 4.0 persistence modules, unit tests and rebased to the latest `master`

PR #604, opened 2021-01-19
Encapsulate auth data extras

dimas-b:


persistence-api/src/main/java/io/stargate/db/AuthenticatedUser.java
```diff
@@ -29,23 +36,92 @@
...
+      if (token == null || roleName == null) {
+        throw new IllegalStateException("token and roleName must be provided");
+      }
+
+      Builder<String, String> map = ImmutableMap.builder();
+      for (Entry<String, ByteBuffer> e : customPayload.entrySet()) {
+        String key = e.getKey();
+        if (key.startsWith(CUSTOM_PAYLOAD_NAME_PREFIX)) {
+          String name = key.substring(CUSTOM_PAYLOAD_NAME_PREFIX.length());
+          String value = StandardCharsets.UTF_8.decode(token).toString();
```
dougwettlaufer: Shouldn't this be either token or role?

dougwettlaufer (commented):
Looks like the right direction

persistence-api/src/main/java/io/stargate/db/AuthenticatedUser.java
dimas-b: fixed

restapi/src/main/java/io/stargate/web/resources/RowResource.java
```diff
@@ -546,7 +558,8 @@ public Response updateRow(
       final RowUpdate changeSet) {
     return RequestHandler.handle(
         () -> {
-          AuthenticatedDB authenticatedDB = db.getDataStoreForToken(token);
+          Map<String, String> allHeaders = getAllHeaders(request);
+          AuthenticatedDB authenticatedDB = db.getDataStoreForToken(token, allHeaders);
```
tomekl007, line 562: This is problematic. It calls the `authenticationService.validateToken(token, headers);` that in turn call the:
```
default AuthenticationSubject validateToken(String token, Map<String, String> headers)
      throws UnauthorizedException {
    return validateToken(token);
  }
```
and headers are ignored, and not propagated. Similar for other calls

tomekl007 (approved):

restapi/src/main/java/io/stargate/web/resources/RowResource.java
dimas-b: Authentication implementations that need to examine headers are expected to override `validateToken(String token, Map<String, String> headers)`

dimas-b: the `default` method is a convenience for "simple" authenticators

dougwettlaufer (approved):

mpenick (approved):

PR #605, opened 2021-01-19
Adds cql event filters so persistence can decide if events should be sent to a given connection

tjake:


cql/src/main/java/org/apache/cassandra/stargate/transport/internal/ProxyInfo.java
```diff
@@ -13,4 +18,12 @@ public ProxyInfo(InetSocketAddress publicAddress, InetSocketAddress sourceAddres
     this.publicAddress = publicAddress;
     this.sourceAddress = sourceAddress;
   }
+
+  public Map<String, String> toHeaders() {
+    return ImmutableMap.of(
```
mpenick, line 23: Should we allocate this in the constructor and cache in a member variable so it doesn't need to be allocated on every event?

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/ProxyInfo.java
```diff
@@ -13,4 +18,12 @@ public ProxyInfo(InetSocketAddress publicAddress, InetSocketAddress sourceAddres
     this.publicAddress = publicAddress;
     this.sourceAddress = sourceAddress;
   }
+
+  public Map<String, String> toHeaders() {
```
mpenick, line 22: Using `Map<String, Object>` here and in the predicate will allow this us to keep this as an `InetSocketAddress` which might require less intermediate objects to convert in downstream code.

persistence-api/src/main/java/io/stargate/db/EventListenerWithChannelFilter.java
```diff
@@ -0,0 +1,221 @@
...
+/**
+ * Specialized version of EventListener which accepts a filter of which channels to send the event
+ * to.
+ */
+public interface EventListenerWithChannelFilter extends EventListener {
+  void onCreateKeyspace(String keyspace, Predicate<Map<String, String>> headerFilter);
+
+  @Override
+  default void onCreateKeyspace(String keyspace) {
+    onCreateKeyspace(keyspace, c -> true);
```
mpenick, line 32: Using `null` instead of `c -> true` will allow the channel matcher to early exit without calling `toHeaders()` and the predicate for the default implementation.

mpenick (approved):
Looks good. Couple minor things that can likely be ignored.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Event.java
```diff
@@ -106,31 +111,35 @@ private NodeEvent(Type type, InetSocketAddress node) {
...
-          Change.MOVED_NODE, new InetSocketAddress(address.address, address.port));
+          Change.MOVED_NODE, new InetSocketAddress(address.address, address.port), headerFilter);
     }
 
     // Assumes the type has already been deserialized
     private static TopologyChange deserializeEvent(ByteBuf cb) {
       Change change = CBUtil.readEnumValue(Change.class, cb);
       InetSocketAddress node = CBUtil.readInet(cb);
-      return new TopologyChange(change, node);
+      return new TopologyChange(change, node, c -> true);
```
tomekl007, line 142: a small thing, we may use:
```
Predicates.alwaysTrue();
```
instead of:
```
 c -> true
```
also, could you elaborate more on which use cases we will pass different type of predicate (other than always true)?

dougwettlaufer (approved):

PR #609, opened 2021-01-21
Pass custom properties to persistence backend

tjake:


restapi/src/main/java/io/stargate/web/resources/Db.java
```diff
@@ -100,7 +103,11 @@ private DataStore getDataStoreInternal(
             .build();
 
     DataStoreOptions options =
-        DataStoreOptions.builder().defaultParameters(parameters).alwaysPrepareQueries(true).build();
+        DataStoreOptions.builder()
+            .defaultParameters(parameters)
+            .alwaysPrepareQueries(true)
+            .customProperties(authenticationSubject.customProperties())
```
dimas-b, line 109: This looks a bit inconsistent with the other places in restapi. In other cases we use full HTTP headers, but here only the subject's headers, which is not guaranteed to be the same. One level up the call chain to this method full headers are available... perhaps we could just pass them here as one extra parameter?

As for `AuthenticationSubject`, would it make sense to _combine_ subject properties (may be set by auth services) with the raw headers? The subject's properties are different from raw HTTP headers, but will reflect the authenticators interpretation of the request, which might be worth examining by persistence implementations. WDYT?

tjake: Ideally these would be the raw http headers, I didn't see where those were.

tjake: Would you mind pushing that change?

dimas-b: sure

dimas-b: implemented that in c16c71458ee7989a06e0222898978775f6dac415

dimas-b (approved):

restapi/src/main/java/io/stargate/web/resources/Db.java
```diff
@@ -88,19 +92,28 @@ public AuthenticatedDB getDataStoreForToken(
...
         authenticationService.validateToken(token, headers);
     return new AuthenticatedDB(
-        getDataStoreInternal(authenticationSubject, pageSize, pagingState), authenticationSubject);
+        getDataStoreInternal(authenticationSubject, pageSize, pagingState, headers),
+        authenticationSubject);
   }
 
   private DataStore getDataStoreInternal(
-      AuthenticationSubject authenticationSubject, int pageSize, ByteBuffer pagingState) {
+      AuthenticationSubject authenticationSubject,
```
dougwettlaufer, line 100: There won't be an authenticationSubject for internal. An "internal" datastore is for things like validating a table based token.

dougwettlaufer: Never mind, was thinking of `createInternal`.

dougwettlaufer (approved):

tomekl007 (approved):

mpenick (approved):

PR #610, opened 2021-01-21
Fix key check for customPayload in StargateQueryHandler

dougwettlaufer:


mpenick (approved):
Maybe create a constant for "stargate.auth.subject.token", but I don't feel strongly about it.

dimas-b (approved):
Nice catch! Thanks!

dimas-b:
There are constants for these keys in `AuthenticatedUser`, they just need to be made public.

PR #611, opened 2021-01-21
Do not pull schema from non-token ring members

mpenick:
In retrospect, it is a bug to pull schema from non-ring members and schema should only be pulled from the storage nodes. Pulling schema from ring members is handled in the`StorageService#onChange()` method.

### DIFF
--- a/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/interceptors/DefaultQueryInterceptor.java
+++ b/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/interceptors/DefaultQueryInterceptor.java
@@ -27,7 +27,6 @@ import org.apache.cassandra.gms.Gossiper;
 import org.apache.cassandra.gms.IEndpointStateChangeSubscriber;
 import org.apache.cassandra.gms.VersionedValue;
 import org.apache.cassandra.locator.InetAddressAndPort;
-import org.apache.cassandra.schema.MigrationManager;
 import org.apache.cassandra.schema.Schema;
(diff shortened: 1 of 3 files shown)


persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/interceptors/DefaultQueryInterceptor.java
```diff
@@ -253,10 +252,6 @@ private void applyState(
         // for DDL queries to reach agreement before returning.
         StargateSystemKeyspace.updatePeerInfo(
             endpoint, "schema_version", StargateSystemKeyspace.SCHEMA_VERSION);
-
-        // This fix schedules a schema pull for the non-member node and is required because
-        // `StorageService.onChange()` doesn't do this for non-member nodes.
-        MigrationManager.instance.scheduleSchemaPull(endpoint, epState);
```
tomekl007, line 259: is it feasible to write a test (probably integration test) that checks that the schema from the non-member node is not pulled?

dougwettlaufer (approved):

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/interceptors/DefaultQueryInterceptor.java
mpenick: I really don't think so, but I did prove it manually and by analyzing the C*/DSE code.

dimas-b (approved):
Only deleted lines :) :+1:

Issue #572, opened 2021-01-05
Support multiple `where` clauses over multiple `fields` (still using only AND)


EricBorczuk:
Currently, a user can do multiple `where` clauses on a single field, such as:
`where={"a.b": {"$gt": 0, "$lt": 3}}`

With SAI in the mix, it might be possible to add where clauses across multiple fields, `AND`'d together, such as:
`where={"a.b": {"$gt": 0, "$lt": 3}, "a.z": {"$eq": true}}`

Though I seem to remember that I had tried to support this and it involved using `OR` which is not supported in C*...

PR #614, opened 2021-01-22
Add ability to search across collections with multiple `where` clauses

EricBorczuk:
Resolves issue https://github.com/stargate/stargate/issues/572


eribeiro (commented):
I'd like to suggest two possible optimizations:

1) Change `existsByDoc` to be a `LinkedHashSet`. It would need adjustments in the code, but it preserves order and uses a `Map<String, Boolean>` behind the scenes. 

2) Lines 921-935 could be written as:
```
    if (docNames.size() > limit) {
      docNames = docNames.stream().limit(limit).collect(Collectors.toCollection(LinkedHashSet::new));

      final Set<String> finalDocNames = docNames;
      int totalSize = countsByDoc.entrySet().stream().filter( e -> finalDocNames.contains(e.getKey()) ).map( e -> e.getValue() ).reduce(0, (x, y) -> x + y);
```
Does it make it more readable or efficient? I don't have a final opinion on that one, tbqh.

dougwettlaufer (approved):

PR #619, opened 2021-01-25
Perform authz before query interception

dougwettlaufer:
Nothing complex is currently happening in the query interceptors but in the event something is added in the future lets do authz before intercepting queries.


persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
```diff
@@ -145,17 +145,17 @@ public ResultMessage process(
       QueryProcessor.metrics.regularStatementsExecuted.inc();
     }
 
+    if (customPayload != null && customPayload.containsKey("stargate.auth.subject.token")) {
+      authorizeByToken(customPayload, statement);
+    }
```
dimas-b, line 150: Since we're tightening security in this PR, I guess we ought to check whether the token is _expected_ to be in the payload. Right now we allow the statement to proceed if the token is _not_ present... perhaps it is not present when if _should_ be present? Can we detect that (and fail in that case)? WDYT?

Tangential: It might be worth moving the `if` into `authorizeByToken` in case the decision is going to involve other conditions.

dimas-b (approved):
Changes LGTM as they are, but perhaps we can improve permission enforcement a bit more (as noted in another comment).

mpenick (approved):

mpenick (approved):

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java
dougwettlaufer: Well, you can have a customPayload without using token auth. Plus even if we set `-Dstargate.auth_api_enable_username_token=true -Dstargate.cql_use_auth_service=true` this could still be a login with a username/password so there really isn't a good way to know if it _should_ be there unless I'm missing something.

ivansenic:
@mpenick I rebased this on the latest master.. Is this something we can merge? Was approved, but for some reason not merged.. Not sure why?

kathirsvn:
@jeffreyscarpenter @ivansenic I have rebased and tested this branch for cql with cassandra 311 backend, looks good. Now, we need to merge this, deploy into dev and check both serverless and classic instances to see if everything is good. Can you please review and approve?
And I believe, we can merge this change into V2 (main) once v1 is tested and looks good.

jeffreyscarpenter:
@kathirsvn for this PR we only need to ensure that the integration tests pass on this repo for C* 3.11, C* 4.0, DSE 6.8.

tatu-at-datastax:
@kathirsvn thank you for merging this. Please also make sure to merge into `main` so we keep `main` up-to-date with changes.

kathirsvn:
@tatu-at-datastax Done, this change is merged to main as well.

PR #602, opened 2021-01-18
Add a startctl option for cassandra.yaml

marksurnin:
This PR allows passing cassandra.yaml to Stargate. Here's the relevant [issue](https://github.com/stargate/stargate/issues/48) and the [Discord thread](https://discord.com/channels/745831180213158001/766324539063533569/799684757008941106).

I will make changes to CassandraPersistenceActivator in cassandra-4.0 and write tests, but I wanted to hear your opinion on this so far.

PR #622, opened 2021-01-27
Increase DSE requests timeouts on the Stargate side in tests

dimas-b:
... due to slow CI

### DIFF
--- a/persistence-dse-6.8/src/main/java/io/stargate/db/dse/DsePersistenceActivator.java
+++ b/persistence-dse-6.8/src/main/java/io/stargate/db/dse/DsePersistenceActivator.java
@@ -125,6 +125,19 @@ public class DsePersistenceActivator extends BaseActivator {
       c.guardrails = guardrailsConfig;
     }
 
+    long timeout;
+    if ((timeout = Long.getLong("stargate.request_timeout_in_ms", -1)) > 0) {
+      c.request_timeout_in_ms = timeout;
+    }
+
+    if ((timeout = Long.getLong("stargate.write_request_timeout_in_ms", -1)) > 0) {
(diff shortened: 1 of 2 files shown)


persistence-dse-6.8/src/main/java/io/stargate/db/dse/DsePersistenceActivator.java
```diff
@@ -125,6 +125,19 @@ public static Config makeConfig(File baseDir) throws IOException {
...
+    long timeout;
+    if ((timeout = Long.getLong("stargate.request_timeout_in_ms", -1)) > 0) {
+      c.request_timeout_in_ms = timeout;
+    }
+
+    if ((timeout = Long.getLong("stargate.write_request_timeout_in_ms", -1)) > 0) {
+      c.write_request_timeout_in_ms = timeout;
+    }
+
+    if ((timeout = Long.getLong("stargate.read_request_timeout_in_ms", -1)) > 0) {
```
tomekl007, line 137: do we want the default to be `-1`? Will it mean that it waits infinitely?

dimas-b: this is just for the `if` it will not be used. I'm also going to refactor this to use config files when #602 merges.

dougwettlaufer (approved):

persistence-dse-6.8/src/main/java/io/stargate/db/dse/DsePersistenceActivator.java
dimas-b: I'll follow up with a more generic solution in another PR.

PR #623, opened 2021-01-27
Add decorate keyspace method for graphql cache

tjake:


graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlDmlResource.java
```diff
@@ -143,15 +144,17 @@ private GraphQL getDefaultGraphql(AsyncResponse asyncResponse) {
...
     if (!KEYSPACE_NAME_PATTERN.matcher(keyspaceName).matches()) {
       LOG.warn("Invalid keyspace in URI, this could be an XSS attack: {}", keyspaceName);
       // Do not reflect back the value
       replyWithGraphqlError(Status.BAD_REQUEST, "Invalid keyspace name", asyncResponse);
       return null;
     }
 
-    GraphQL graphql = graphqlCache.getDml(keyspaceName);
+    GraphQL graphql =
+        graphqlCache.getDml(keyspaceName, RequestToHeadersMapper.getAllHeaders(request));
```
tomekl007, line 157: You may need to do a similar in the:`GraphqlDdlResource`
It is using the graphQl cache:
```
this.graphql = graphqlCache.getDdl();
```
and passes it to `GraphqlResourceBase`

graphqlapi/src/main/java/io/stargate/graphql/web/GraphqlCache.java
```diff
@@ -85,17 +89,20 @@ public GraphQL getDdl() {
     return ddlGraphql;
   }
 
-  public GraphQL getDml(String keyspace) {
-    DmlGraphqlReference ref = dmlGraphqls.get(keyspace);
+  public GraphQL getDml(String keyspace, Map<String, String> headers) {
+    String decoratedName = persistence.decorateKeyspaceName(keyspace, headers);
```
dimas-b, line 93: Would it be possible to put the decoration method on `DataStore` to consolidate all user-level access to data in the `DataStore` interface? In the majority of cases rich APIs only need to deal with the `DataStore`. Making `Persistence` a required dependency where keyspaces are manipulated seems a bit inconvenient. Also `DataStore` instances already have relevant request properties. WDYT?

dimas-b: If the DataStore-based approach looks reasonable, decorating keyspaces may not be required. We could have something like `DataStore.getNamespace()` and use it here as part of the cache key :thinking:

tjake: The Graphql cache is global so not sure how to dynamically pass headers many times?

dimas-b: Yeah, it might require some refactoring, but I guess we have all the data available at the start of the requests and it is used a bit later to construct the  `DataStore`. @olim7t WDYT?

olim7t: Indeed the cache is global, it serves `GraphQL` instances for all the keyspaces known to this Stargate instance. When an HTTP request comes in, it uses the "external" keyspace such as `ks1`, but the cache knows that keyspace by its "internal" name such as `someNamespace_ks1`.
We need that translation, otherwise we'd either have to: a) have one cache per user, which complicates things unnecessarily, or b) remove the cache and recompute the `GraphQL` for every request, which is impractical.

> use it here as part of the cache key

We already do. The cache key is the decorated keyspace name `someNamespace_ks1`.

> put the decoration method on `DataStore`

It can be done. We'd have to keep a reference to a `DataStore` instance in the cache. For example the "internal" one that we use in the constructor. Line 79:
```java
this.dataStore = dataStoreFactory.createInternal();
```
Line 93:
```java
String decoratedName = this.dataStore.decorateKeyspaceName(keyspace, headers);
```
But it feels wrong to me because it's only a coincidence that we already have that instance available. If we didn't we'd have to create it just for the sake of calling `decorateKeyspace`, which is wasteful. If anything, I think `DataStoreFactory` would be a better place to put it.

dimas-b: ok. thanks! let's go with current impl. :+1:

dimas-b (approved):

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlDmlResource.java
olim7t: We don't need to. There are two steps where decorating keyspaces might come into play: building the GraphQL schema, and executing operations.

If we are building the **DML** schema for a keyspace `ks1` that contains a table `foo`, the GraphQL schema will contain a query `foo`, so at construction time we need to inspect the tables. So if keyspace decoration is required, it needs to happen at that time.

On the other hand, the DDL schema only contains generic operations such as `createTable(keyspaceName, tableName)`. Those operations will be the same whichever user keyspaces are present. If you look at `DdlSchemaBuilder`,  it's considerably simpler than `DmlSchemaBuilder`, and at no point does the CQL schema affect the GraphQL schema that we generate.

Execution time is another story: when we process a `createTable(keyspaceName, tableName)`, we need to decorate the keyspace name. This happens in `CassandraFetcher`, we pass the headers when we create the datastore so it will be properly scoped:
```java
    DataStoreOptions dataStoreOptions =
        DataStoreOptions.builder()
            .putAllCustomProperties(httpAwareContext.getAllHeaders())  // <---------------------
            .putAllCustomProperties(authenticationSubject.customProperties())
            .defaultParameters(parameters)
            .alwaysPrepareQueries(true)
            .build();
```

olim7t (approved):
LGTM 👍 

Pending the decision of where we put `decorateKeyspaceName`, but I'm fine with either solution.

PR #626, opened 2021-01-28
Include health checks in the readiness check

### LINKED ISSUE #620: Readiness check should wait for all bundles to start
As reported in #617 `GET /checker/readiness` may report 200 (ok) before all the bundles are started.

This is problematic because some services may still be unavailable at that time and may still be consuming system resources for their own bootstrap activities.

The readiness check should probably report OK only when all services are up (or at least all known services).

Note: we may be able to use OSGi bundle meta-data to flag "important" services.

dimas-b:
Add health checks for REST API and CQL modules

Closes #620

### DIFF
--- a/core/src/main/java/io/stargate/core/activator/BaseActivator.java
+++ b/core/src/main/java/io/stargate/core/activator/BaseActivator.java
@@ -15,11 +15,14 @@
  */
 package io.stargate.core.activator;
 
+import com.codahale.metrics.health.HealthCheck;
+import com.codahale.metrics.health.HealthCheckRegistry;
 import java.util.*;
 import java.util.Hashtable;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import org.osgi.framework.BundleActivator;
 import org.osgi.framework.BundleContext;
@@ -34,7 +37,13 @@ import org.slf4j.LoggerFactory;
 public abstract class BaseActivator implements BundleActivator {
   private static final Logger logger = LoggerFactory.getLogger(BaseActivator.class);
 
+  private final ServicePointer<HealthCheckRegistry> healthCheckRegistry =
+      ServicePointer.create(HealthCheckRegistry.class);
+
   private final String activatorName;
+  private final AtomicBoolean available = new AtomicBoolean();
+  private final HealthCheck healthCheck;
+  private List<ServicePointer<?>> dependencies;
 
   protected BundleContext context;
 
@@ -46,7 +55,26 @@ public abstract class BaseActivator implements BundleActivator {
 
   /** @param activatorName - The name used when logging the progress of registration. */
   public BaseActivator(String activatorName) {
+    this(activatorName, false);
+  }
+
+  /**
+   * @param activatorName - The name used when logging the progress of registration and as the name
+   *     of the health checker (in lower case).
+   */
+  public BaseActivator(String activatorName, boolean registerHealthCheck) {
     this.activatorName = activatorName;
+    this.healthCheck =
+        registerHealthCheck
+            ? new HealthCheck() {
+              @Override
+              protected Result check() {
+                return available.get()
+                    ? Result.healthy("Available")
+                    : Result.unhealthy("Not Available");
+              }
+            }
+            : null;
   }
 
   /**
@@ -63,19 +91,29 @@ public abstract class BaseActivator implements BundleActivator {
   public synchronized void start(BundleContext context) throws InvalidSyntaxException {
     logger.info("Starting {} ...", activatorName);
     this.context = context;
-    if (dependencies().isEmpty() && lazyDependencies().isEmpty()) {
+
+    String filter = constructDependenciesFilter();
+
+    if (dependencies.isEmpty() && lazyDependencies().isEmpty()) {
       startServiceInternal();
     } else {
-      tracker = new Tracker(context, context.createFilter(constructDependenciesFilter()));
+      tracker = new Tracker(context, context.createFilter(filter));
       tracker.open();
     }
   }
 
   String constructDependenciesFilter() {
+    List<ServicePointer<?>> dep = dependencies();
+    if (healthCheck != null) {
+      dep = new ArrayList<>(dep);
+      dep.add(healthCheckRegistry);
+    }
+    dependencies = dep;
+
     StringBuilder builder = new StringBuilder("(|");
 
     List<Service<?>> allDependencies = new ArrayList<>();
-    allDependencies.addAll(dependencies());
+    allDependencies.addAll(dependencies);
     allDependencies.addAll(lazyDependencies());
 
     for (Service<?> servicePointer : allDependencies) {
@@ -96,6 +134,8 @@ public abstract class BaseActivator implements BundleActivator {
    */
   @Override
   public synchronized void stop(BundleContext context) throws Exception {
+    available.set(false);
+
     if (started) {
       logger.info("Stopping {}", activatorName);
       stopService();
@@ -121,6 +161,11 @@ public abstract class BaseActivator implements BundleActivator {
     }
     started = true;
 
+    if (healthCheck != null) {
+      String healthCheckName = activatorName.toLowerCase();
+      healthCheckRegistry.get().register(healthCheckName, healthCheck);
+    }
+
     try {
       List<ServiceAndProperties> services = createServices();
       for (ServiceAndProperties service : services) {
@@ -131,6 +176,7 @@ public abstract class BaseActivator implements BundleActivator {
                   service.targetServiceClass.getName(), service.service, service.properties));
         }
       }
+      available.set(true);
       logger.info("Started {}", activatorName);
     } catch (Exception e) {
       throw new ServiceStartException("Unable to start " + activatorName, e);
@@ -161,7 +207,7 @@ public abstract class BaseActivator implements BundleActivator {
       if (service == null) {
         return;
       }
-      for (ServicePointer<?> servicePointer : dependencies()) {
+      for (ServicePointer<?> servicePointer : dependencies) {
         if (servicePointer.expectedClass.isAssignableFrom(service.getClass())) {
           logger.debug("{} using service: {}", activatorName, ref.getBundle());
           servicePointer.set(service);
@@ -175,7 +221,7 @@ public abstract class BaseActivator implements BundleActivator {
         }
       }
 
-      if (dependencies().stream().map(v -> v.service).allMatch(Objects::nonNull)) {
+      if (dependencies.stream().map(v -> v.service).allMatch(Objects::nonNull)) {
         startServiceInternal();
       }
     }
--- a/cql/src/main/java/io/stargate/cql/CqlActivator.java
+++ b/cql/src/main/java/io/stargate/cql/CqlActivator.java
@@ -15,6 +15,8 @@
  */
 package io.stargate.cql;
 
+import com.codahale.metrics.health.HealthCheck;
+import com.codahale.metrics.health.HealthCheckRegistry;
 import io.stargate.auth.AuthenticationService;
 import io.stargate.core.activator.BaseActivator;
 import io.stargate.core.metrics.api.Metrics;
@@ -24,12 +26,18 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.Arrays;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.cassandra.config.Config;
 import org.jetbrains.annotations.Nullable;
 
 public class CqlActivator extends BaseActivator {
+
+  public static final String HEALTH_CHECK_NAME = "cql";
+
   private CqlImpl cql;
   private final ServicePointer<Metrics> metrics = ServicePointer.create(Metrics.class);
+  private final ServicePointer<HealthCheckRegistry> healthCheckRegistry =
+      ServicePointer.create(HealthCheckRegistry.class);
   private final ServicePointer<AuthenticationService> authentication =
       ServicePointer.create(
           AuthenticationService.class,
@@ -44,6 +52,15 @@ public class CqlActivator extends BaseActivator {
   private static final boolean USE_AUTH_SERVICE =
       Boolean.parseBoolean(System.getProperty("stargate.cql_use_auth_service", "false"));
 
+  private final AtomicBoolean started = new AtomicBoolean();
+  private final HealthCheck healthCheck =
+      new HealthCheck() {
+        @Override
+        protected Result check() throws Exception {
+          return started.get() ? Result.healthy("Started") : Result.unhealthy("Not started");
+        }
+      };
+
   public CqlActivator() {
     super("CQL");
   }
@@ -54,8 +71,12 @@ public class CqlActivator extends BaseActivator {
     if (cql != null) { // Already started
       return null;
     }
+
+    healthCheckRegistry.get().register(HEALTH_CHECK_NAME, healthCheck);
+
     cql = new CqlImpl(makeConfig(), persistence.get(), metrics.get(), authentication.get());
     cql.start();
+    started.set(true);
 
     return null;
   }
@@ -65,16 +86,18 @@ public class CqlActivator extends BaseActivator {
     if (cql == null) { // Not started
       return;
     }
+    started.set(false);
     cql.stop();
+    healthCheckRegistry.get().unregister(HEALTH_CHECK_NAME);
     cql = null;
   }
 
   @Override
   protected List<ServicePointer<?>> dependencies() {
     if (USE_AUTH_SERVICE) {
-      return Arrays.asList(metrics, persistence, authentication);
+      return Arrays.asList(metrics, healthCheckRegistry, persistence, authentication);
     } else {
-      return Arrays.asList(metrics, persistence);
+      return Arrays.asList(metrics, healthCheckRegistry, persistence);
     }
   }
 
--- a/cql/src/main/java/io/stargate/cql/CqlActivator.java
+++ b/cql/src/main/java/io/stargate/cql/CqlActivator.java
@@ -15,8 +15,6 @@
  */
 package io.stargate.cql;
 
-import com.codahale.metrics.health.HealthCheck;
-import com.codahale.metrics.health.HealthCheckRegistry;
 import io.stargate.auth.AuthenticationService;
 import io.stargate.core.activator.BaseActivator;
 import io.stargate.core.metrics.api.Metrics;
@@ -26,18 +24,12 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.Arrays;
 import java.util.List;
-import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.cassandra.config.Config;
 import org.jetbrains.annotations.Nullable;
 
 public class CqlActivator extends BaseActivator {
-
-  public static final String HEALTH_CHECK_NAME = "cql";
-
   private CqlImpl cql;
   private final ServicePointer<Metrics> metrics = ServicePointer.create(Metrics.class);
-  private final ServicePointer<HealthCheckRegistry> healthCheckRegistry =
-      ServicePointer.create(HealthCheckRegistry.class);
   private final ServicePointer<AuthenticationService> authentication =
       ServicePointer.create(
           AuthenticationService.class,
@@ -52,17 +44,8 @@ public class CqlActivator extends BaseActivator {
   private static final boolean USE_AUTH_SERVICE =
       Boolean.parseBoolean(System.getProperty("stargate.cql_use_auth_service", "false"));
 
-  private final AtomicBoolean started = new AtomicBoolean();
-  private final HealthCheck healthCheck =
-      new HealthCheck() {
-        @Override
-        protected Result check() throws Exception {
-          return started.get() ? Result.healthy("Started") : Result.unhealthy("Not started");
-        }
-      };
-
   public CqlActivator() {
-    super("CQL");
+    super("CQL", true);
   }
 
   @Nullable
@@ -71,12 +54,8 @@ public class CqlActivator extends BaseActivator {
     if (cql != null) { // Already started
       return null;
     }
-
-    healthCheckRegistry.get().register(HEALTH_CHECK_NAME, healthCheck);
-
     cql = new CqlImpl(makeConfig(), persistence.get(), metrics.get(), authentication.get());
     cql.start();
-    started.set(true);
 
     return null;
   }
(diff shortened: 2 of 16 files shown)


testing/src/test/java/io/stargate/it/http/HealthCheckerTest.java
```diff
@@ -36,14 +38,28 @@ public void liveness() throws IOException {
...
+  public void missingReadinessCheck() throws IOException {
+    String body =
+        RestUtils.get(
+            "",
+            String.format("%s:8084/checker/readiness?check=testUnknown", host),
+            HttpStatus.SC_SERVICE_UNAVAILABLE);
+
+    assertThat(body).isEqualTo("NOT READY");
+  }
+
```
tomekl007, line 62: missing test case for a known health check that reports unhealthy.

health-checker/src/main/java/io/stargate/health/CheckerResource.java
```diff
@@ -33,12 +45,40 @@ public Response checkLiveness() {
...
+    boolean ready = started && dataStoreAvailable;
+
+    for (Entry<String, Result> e : status.entrySet()) {
+      String name = e.getKey();
+      requiredChecks.remove(name);
+      boolean healthy = e.getValue().isHealthy();
+
+      if (!healthy) {
+        logger.warn("Failed health check: {}", name);
+        ready = false;
```
tomekl007, line 69: what will happen if the first health check is not ready, but the second one is? 
You are doing:
```
requiredChecks.remove(name);
```
before checking if it is `healthy`.
As far as I see, this logic will not work as expected if it is not healthy but known.

health-checker/src/main/java/io/stargate/health/CheckerResource.java
```diff
@@ -33,12 +45,40 @@ public Response checkLiveness() {
...
+
+    if (!requiredChecks.isEmpty()) {
+      logger.warn("Missing required health checks: {}", requiredChecks);
+      ready = false;
+    }
+
+    if (ready) {
       return Response.status(Response.Status.OK).entity("READY").build();
+    } else {
+      return Response.status(Response.Status.SERVICE_UNAVAILABLE).entity("NOT READY").build();
```
tomekl007, line 81: I think that we should include a list of unavailable health checks in the entity to give the caller more info. In case the checks lists will have N value, the caller will not know (without looking at the logs) which health checked failed. The UX of this is non-ideal.

tomekl007 (changes requested):

cql/src/main/java/io/stargate/cql/CqlActivator.java
```diff
@@ -44,6 +52,15 @@
   private static final boolean USE_AUTH_SERVICE =
       Boolean.parseBoolean(System.getProperty("stargate.cql_use_auth_service", "false"));
 
+  private final AtomicBoolean started = new AtomicBoolean();
+  private final HealthCheck healthCheck =
+      new HealthCheck() {
+        @Override
+        protected Result check() throws Exception {
+          return started.get() ? Result.healthy("Started") : Result.unhealthy("Not started");
```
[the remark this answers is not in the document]
dimas-b: what other information do you mean?

health-checker/src/main/java/io/stargate/health/BundleService.java
```diff
@@ -32,14 +37,12 @@ public boolean checkBundleStates() {
     return true;
   }
 
-  public boolean checkIsReady() {
-    ServiceReference dataStoreFactoryReference =
-        context.getServiceReference(DataStoreFactory.class.getName());
+  public boolean checkDataStoreAvailable() {
```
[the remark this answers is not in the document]
dimas-b: Periodic, yes.

As for caching the factory reference, is the get/unget operation too expensive? I hope not, and going through OSGi add extra validation that the service is operational since OSGi is expected to refuse the `get` operation if the bundle is no longer valid (e.g. undeployed).

health-checker/src/main/java/io/stargate/health/CheckerResource.java
dimas-b: We remove the name from `requiredChecks` to indicate that we processes that checker. If anything is unhealthy then `ready` will be `false` and the HTTP response will be `SERVICE_UNAVAILABLE`

health-checker/src/main/java/io/stargate/health/CheckerResource.java
dimas-b: I thought about that but did not want to alter the API behaviour from the client's perspective.

We also have the `/healthcheck` resource that reports full checker details including the message and timestamp. The only validations this method performs that will be missing from `/healthcheck` output are the bundle states and `DataStore` access. If exposing them to the client is worthwhile, I'd rather add them to `/healthcheck`. WDYT?

health-checker/src/main/java/io/stargate/health/CheckerResource.java
```diff
@@ -33,12 +45,40 @@ public Response checkLiveness() {
 
   @GET
   @Path("/readiness")
-  public Response checkReadiness() {
-    // Readiness requires a more heavy weight check to decide if the app is able to successfully
-    // respond to traffic
-    if (bundleService.checkIsReady()) {
+  public Response checkReadiness(@QueryParam("check") Set<String> checks) {
```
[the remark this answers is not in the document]
dimas-b: `Optional` is not supported by our Web framework: ` [[FATAL] No injection source found for a parameter of type public javax.ws.rs.core.Response io.stargate.health.CheckerResource.checkReadiness(java.util.Optional)`

health-checker/src/main/java/io/stargate/health/CheckerResource.java
tomekl007: How does it alter the behavior? We will return proper status, only the entity will contain more data.
I think that ideally, for example for `check=a&check=b` we could return
```
NOT READY
a = HEALTHY
b = UNHELATHY
``` 
That way our API does not change, but we are giving more info to our customers, wdyt?

health-checker/src/main/java/io/stargate/health/BundleService.java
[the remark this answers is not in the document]
tomekl007: Regarding being too expensive, I am not sure. Good point about additional validation.

cql/src/main/java/io/stargate/cql/CqlActivator.java
```diff
@@ -44,6 +52,15 @@
   private static final boolean USE_AUTH_SERVICE =
       Boolean.parseBoolean(System.getProperty("stargate.cql_use_auth_service", "false"));
 
+  private final AtomicBoolean started = new AtomicBoolean();
```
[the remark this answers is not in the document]
dimas-b: `BaseActivator.started` is set inside `synchronized` blocks. So to be sure we always have up-to-date values we'd have to synchronize the get operation too, which I think is not desirable for health checkers. I'm not really familiar with the health checker framework, but I would not want to block though checks.

Also, `BaseActivator.started` is unset _after_ the service stopped completely, but I suppose we ought to report "unhealthy" right after it only started shutting down.

I can probably refactor ``BaseActivator` to provide the required information without blocking. Do you think it'd be worth doing this refactoring?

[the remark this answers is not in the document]
tomekl007: I am not sure. If all activators are creating the health-check in the same way (via `started` flag) then maybe it is worth doing?

health-checker/src/main/java/io/stargate/health/CheckerResource.java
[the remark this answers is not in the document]
tomekl007: Then maybe wrap the value in the `Optional.ofNullable`? I am not sure if it will improve readability or not 🤔

cql/src/main/java/io/stargate/cql/CqlActivator.java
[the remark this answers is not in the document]
dougwettlaufer: +1 for not blocking in checks. Depending on type (liveness vs readiness) they should run pretty quick.

cql/src/main/java/io/stargate/cql/CqlActivator.java
```diff
@@ -54,8 +71,12 @@ protected ServiceAndProperties createService() {
     if (cql != null) { // Already started
       return null;
     }
+
+    healthCheckRegistry.get().register(HEALTH_CHECK_NAME, healthCheck);
+
     cql = new CqlImpl(makeConfig(), persistence.get(), metrics.get(), authentication.get());
     cql.start();
+    started.set(true);
```
dougwettlaufer, line 79: Do we want to define "started" vs "running"? Like after calling `start()` it'll take a little bit before it's ready to start handling connections. 

_You might have this and I just haven't gotten to it yet_

health-checker/pom.xml
```diff
@@ -198,6 +198,9 @@
         <configuration>
           <archive>
             <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+            <manifestEntries>
+              <x-Stargate-Health-Checkers>deadlocks</x-Stargate-Health-Checkers>
```
dougwettlaufer, line 202: "deadlocks"?

graphqlapi/pom.xml
```diff
@@ -189,6 +189,9 @@
         <configuration>
           <archive>
             <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+            <manifestEntries>
+              <x-Stargate-Health-Checkers>graphql</x-Stargate-Health-Checkers>
```
dougwettlaufer, line 193: Should probably be `graphqlapi` to match what you have for restapi.

health-checker/src/main/java/io/stargate/health/CheckerResource.java
dougwettlaufer: If you're in k8s all you care about is 200 or non-200 responses the response body is mostly ignored/swallowed. So if checks are failing you'd check the logs anyway.

dougwettlaufer (approved):

graphqlapi/pom.xml
dimas-b: I was going by the service name (from the corresponding activator) in lower case.

health-checker/pom.xml
dimas-b: comes from the framework, I guess (did not chase it down). I thought it would be useful to enforce by default.

dimas-b: it checks for java thread deadlocks.

cql/src/main/java/io/stargate/cql/CqlActivator.java
dimas-b: I'll double check. At first glance it looks like the TCP stuff should be ready by the time `cql.start()` returns :thinking:

dimas-b: `CqlImpl` calls `persistence.setRpcReady(true)` before returning from `start()` I assume this means we're ready for client requests at this point.

graphqlapi/pom.xml
dimas-b: In the latest commit, the reuse of the activator name (lower case) is explicit (cf. `BaseActivator`).

testing/src/test/java/io/stargate/it/http/HealthCheckerTest.java
dimas-b: added unit tests

health-checker/src/main/java/io/stargate/health/CheckerResource.java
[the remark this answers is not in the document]
dimas-b: Not sure wrapping helps that much because we also need to check `isEmpty()`

cql/src/main/java/io/stargate/cql/CqlActivator.java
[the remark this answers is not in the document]
dimas-b: Refactored `BaseActivator` to provide "available" flags that are slightly different from its `started` flag - pls see updated source code.

cql/src/main/java/io/stargate/cql/CqlActivator.java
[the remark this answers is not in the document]
dimas-b: (Please note that the checker code has been refactored and moved to `BaseActivator`)

health-checker/pom.xml
dougwettlaufer: Oh, interesting.

cql/src/main/java/io/stargate/cql/CqlActivator.java
dougwettlaufer: CQL was just an example. Probably a bigger issue for rest and graphql since they take a little longer to start up.

dimas-b: GraphQL preloads its cache before returning from "start" (verified in debugger). As for REST it appears to be pretty stateless.

We will certainly have some period of slow operation before JIT kicks in, but warming it up is probably too much for this PR's scope WDYT?

health-checker/src/main/java/io/stargate/health/CheckerResource.java
tomekl007: Will it be ALWAYS run in the k8s env? If yes, then I agree. However, if there will be some users that will run stargate themself, the additional monitoring information may be useful, right?

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -63,19 +91,29 @@ public BaseActivator(String activatorName) {
...
-      tracker = new Tracker(context, context.createFilter(constructDependenciesFilter()));
+      tracker = new Tracker(context, context.createFilter(filter));
       tracker.open();
     }
   }
 
   String constructDependenciesFilter() {
+    List<ServicePointer<?>> dep = dependencies();
+    if (healthCheck != null) {
+      dep = new ArrayList<>(dep);
```
tomekl007, line 108: is this copy needed because you are taking into account the fact that `dependencies()` may return an immutable list?

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -121,6 +161,11 @@ private synchronized void startServiceInternal() {
     }
     started = true;
 
+    if (healthCheck != null) {
+      String healthCheckName = activatorName.toLowerCase();
```
tomekl007, line 165: is this `toLowerCase` required? Is there a problem to have health-checks not lowercased?

health-checker/src/test/java/io/stargate/health/CheckerResourceTest.java
```diff
@@ -0,0 +1,179 @@
...
+    assertThat(checker.checkReadiness(null).getStatus())
+        .isEqualTo(SERVICE_UNAVAILABLE.getStatusCode());
+    assertThat(checker.checkReadiness(Collections.emptySet()).getStatus())
+        .isEqualTo(SERVICE_UNAVAILABLE.getStatusCode());
+    assertThat(checker.checkReadiness(Collections.singleton("test-unhealthy")).getStatus())
+        .isEqualTo(SERVICE_UNAVAILABLE.getStatusCode());
+    assertThat(checker.checkReadiness(ImmutableSet.of("test-ok", "test-unhealthy")).getStatus())
+        .isEqualTo(SERVICE_UNAVAILABLE.getStatusCode());
+    assertThat(checker.checkReadiness(Collections.singleton("test-ok")).getStatus())
+        .isEqualTo(OK.getStatusCode());
```
tomekl007, line 177: Nit: seems like a candidate for splitting into two tests

health-checker/src/main/java/io/stargate/health/BundleService.java
```diff
@@ -32,36 +37,52 @@ public boolean checkBundleStates() {
...
+      logger.warn("DataStoreFactory service is not available");
+    } catch (Exception e) {
+      logger.warn("checkIsReady failed with", e);
+      return false;
     }
 
     return false;
   }
+
+  public Set<String> defaultHeathCheckNames() {
```
tub, line 77: Typo - HeathCheck -> HealthCheck :)

health-checker/src/main/java/io/stargate/health/CheckerResource.java
dimas-b: but we have `/healthcheck` that shows that info. Why would we want to expose the same data in two places?

We can add extra info to `/healthcheck` since it's more of an investigation tool, while `readiness` is supposed to be an on/off quick check. WDYT?

tomekl007: sounds good.

core/src/main/java/io/stargate/core/activator/BaseActivator.java
dimas-b: yes, some services return immutable dependencies lists.

core/src/main/java/io/stargate/core/activator/BaseActivator.java
dimas-b: not really required, but I thought it would be nice for consistency since they are used as keys in URLs. I can remove the conversion to lower case if this is a concern.

health-checker/src/main/java/io/stargate/health/BundleService.java
dimas-b: Thanks - fixed in 5c2e082378c386d5b0d95d44738506fa72e9522c

health-checker/src/test/java/io/stargate/health/CheckerResourceTest.java
dimas-b: Converted to `@ParameterizedTest` in 41cc8a73883bbd8583bf41d19b6762769bc17ebf

tomekl007 (approved):

health-checker/src/main/java/io/stargate/health/CheckerResource.java
dimas-b: Implemented in f90f105f241098f9bc651ec256d4bd89cdcbd292.

Sample `/healthcheck` output:

```
{
  "bundles": {
    "healthy": false,
    "message": "Inactive bundles: [io.stargate.db, io.stargate.web]",
    "duration": 0,
    "timestamp": "2021-01-29T12:43:11.298-05:00"
  },
  "cql": {
    "healthy": true,
    "message": "Available",
    "duration": 0,
    "timestamp": "2021-01-29T12:43:11.299-05:00"
  },
  "datastore": {
    "healthy": true,
    "message": "DataStore is operational",
    "duration": 27,
    "timestamp": "2021-01-29T12:43:11.298-05:00"
  },
  "deadlocks": {
    "healthy": true,
    "duration": 0,
    "timestamp": "2021-01-29T12:43:11.299-05:00"
  }
}
```

dimas-b:
I did some refactoring in the latest commit (as discussed in comments). Please review again.

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
```diff
@@ -0,0 +1,85 @@
...
+  protected Result check() {
+    try {
+      ServiceReference<DataStoreFactory> dataStoreFactoryReference =
+          context.getServiceReference(DataStoreFactory.class);
+      if (dataStoreFactoryReference != null) {
+        DataStoreFactory dataStoreFactory = context.getService(dataStoreFactoryReference);
+        try {
+          DataStore dataStore = dataStoreFactory.createInternal();
+
+          Future<ResultSet> rs =
```
tomekl007, line 49: What consistency level this query uses execution? Is it `QUORUM`? If it is `ONE` it may be too low for health-check.

health-checker/src/main/java/io/stargate/health/HealthCheckerActivator.java
```diff
@@ -22,13 +26,22 @@ public HealthCheckerActivator() {
...
   @Nullable
   @Override
   protected ServiceAndProperties createService() {
     log.info("Starting healthchecker....");
     try {
+      BundleStateChecker bundleStateChecker = new BundleStateChecker(context);
+      healthCheckRegistry.get().register(BUNDLES_CHECK_NAME, bundleStateChecker);
+
+      DataStoreHealthChecker dataStoreHealthChecker = new DataStoreHealthChecker(context);
+      healthCheckRegistry.get().register(DATASTORE_CHECK_NAME, dataStoreHealthChecker);
```
tomekl007, line 43: why don't treat the `DATASTORE_CHECK_NAME` as a normal `BUNDLES_CHECK_NAME`? If we would do that, we won't need this special case. It would be handled by the `BundleStateChecker`

health-checker/src/main/java/io/stargate/health/CheckerResource.java
```diff
@@ -47,7 +50,8 @@ public CheckerResource() {
   public Response checkLiveness() {
     // Sufficient to just return a 200 OK since liveness just means "app is up and doesn't need
     // restarted"
-    if (bundleService.checkBundleStates()) {
+    HealthCheck bundleHealthCheck = healthCheckRegistry.getHealthCheck(BUNDLES_CHECK_NAME);
+    if (bundleHealthCheck != null && bundleHealthCheck.execute().isHealthy()) {
```
tomekl007, line 54: in what situation the `bundleHealthCheck == null`? Isn't it abnormal? Right now, if this happens to be null, we will return the status `UP`. It may be misleading.

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
dimas-b: Good point. `system.local` is handled by Stargate itself, so running this query does not guarantee that we can fetch data from storage nodes. On the other hand changing this check to hit actual storage nodes may be too much for this PR, which only deals with exposing more of exisisting checks to HTTP clients. I'll file a follow-up issues for this.

health-checker/src/main/java/io/stargate/health/HealthCheckerActivator.java
dimas-b: I assumed we wanted to delineate checks by their function to make trouble-shooting easier. Different names will ensure that we expose information about these checks in separate JSON elements under `/healthcheck`.

health-checker/src/main/java/io/stargate/health/CheckerResource.java
dimas-b: Not sure I follow, sorry. If `bundleHealthCheck` is `null` the execution will proceed to line 58 and return `SERVICE_UNAVAILABLE` :thinking: 

The check for `null` is just a basic precaution if case the code in other classes is changed and the check is not registered for some reason.

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
dimas-b: filed #631

health-checker/src/main/java/io/stargate/health/CheckerResource.java
tomekl007: You are right, sorry I missed that 👍

PR #628, opened 2021-01-28
Upgrade graphql-java to 16.1

### LINKED ISSUE #627: GraphQL: fix aliased sub-selections in queries
We use `DataFetchingSelectionSet.getField()` in a couple of places to customize our queries based on what data the user requested.

That method has been removed in `graphql-java` 16.1, for a good reason: it is deceptive because the same field can appear multiple times, under different aliases.

For example in our DDL schema:

```graphql
{
  keyspace(name: "test") {
    users: table(name: "users") { columns { name } }
    products: table(name: "products") { columns { name } }
  }
}
```

And in our DML schema (more convoluted, but technically still a valid query):

```graphql
{
  users(value: { id: 1 }) {
    values {
      id
    }
    values2: values {
      first_name
      last_name
    }
  }
}
```

Both queries are broken currently (returning `null` for some of the fields). We need to use getField***s***() instead of getField(), iterate the resulting list and put each result under the correct alias.

olim7t:
Fixes #627


dougwettlaufer:
Test failure related?

olim7t:
> Test failure related?

No, it seems unrelated. I'm relaunching the build.

dougwettlaufer (approved):

PR #629, opened 2021-01-29
Initial rate limiting capability proposal for Stargate

pcmanus:
This commit adds a "pluggable" rate limiting system. Specifically, a new
`RateLimitingPersistence` wraps the underlying "DB" persistence and
allow to optionally rate limit (or even completly reject) queries. The
actual decision of what queries are rate limited and by how much is
delegated to a new `RateLimitingManager` interface.

The commit also provide an example `RateLimitingManager` that rate limit
_all_ queries to a certain rate (in QPS).


persistence-api/src/main/java/io/stargate/db/DbActivator.java
```diff
@@ -3,28 +3,71 @@
...
     super("DB services");
   }
 
   @Override
-  protected ServiceAndProperties createService() {
-    return new ServiceAndProperties(
-        new PersistenceDataStoreFactory(persistence.get()), DataStoreFactory.class);
+  protected List<ServiceAndProperties> createServices() {
+    Persistence persistence = this.dbPersistence.get();
+    if (!RATE_LIMITING_IDENTIFIER.equalsIgnoreCase("<none>")) {
```
dimas-b, line 48: I'm not sure how well this is going to work in OSGi. If a specific `RATE_LIMITING_IDENTIFIER` is not configured, the dependency will never be satisfied and `createService` will not be invoked.

It might be preferable not to add the dependency if we know it is not configured. For example: https://github.com/stargate/stargate/pull/554/files#diff-d00f29611c415af2e85b87fe3e7559fdc9125b9d0957be87f750fce77d7c8908R180-R187

persistence-api/src/main/java/io/stargate/db/DbActivator.java
```diff
@@ -3,28 +3,71 @@
...
+  private static Hashtable<String, String> stargatePersistenceProperties() {
+    Hashtable<String, String> props = new Hashtable<>();
+    props.put("Identifier", PERSISTENCE_IDENTIFIER);
+    return props;
   }
 
   @Override
   protected List<ServicePointer<?>> dependencies() {
-    return Collections.singletonList(persistence);
+    return Arrays.asList(dbPersistence, rateLimitingManager);
```
dimas-b, line 71: I'm not sure how well this is going to work in OSGi. If a specific `RATE_LIMITING_IDENTIFIER` is not configured, the dependency will never be satisfied and `createService` will not be invoked.

It might be preferable not to add the dependency if we know it is not configured. For example: https://github.com/stargate/stargate/pull/554/files#diff-d00f29611c415af2e85b87fe3e7559fdc9125b9d0957be87f750fce77d7c8908R180-R187

rate-limiting-global/src/main/java/io/stargate/db/limiter/global/impl/GlobalRateLimitingManager.java
```diff
@@ -0,0 +1,143 @@
...
+    //   reads that drivers do on connection (basically, for testing to be reliable, we want to set
+    //   a  very aggressive rate limiting, but if driver queries hit it, the test will take
+    // forever).
+    //   So we explicitly target those driver system table reads. It's super fragile because 1) it
+    //   assume the driver don't add extra spaces compared to what we check and 2) it doesn't use
+    //   prepared queries (if it did, we wouldn't rate limit the prepare, but we would the
+    //   execution).
+    //   Tl;dr: we'd need Stargate to stop passing raw strings into the persistence to be able to
+    //   do this properly.
+    return RATE_LIMIT_SYSTEM_TABLES && query.startsWith("SELECT") && query.contains("FROM System");
```
dimas-b, line 119: Should the condition on `RATE_LIMIT_SYSTEM_TABLES` be inverted, perhaps?

stargate-starter/src/main/java/io/stargate/starter/Starter.java
```diff
@@ -488,6 +488,17 @@ public void start() throws BundleException {
         }
 
         if (name.endsWith(".jar") && !name.startsWith("stargate-starter-")) {
+          // Any rate-limiting jar needs to be installed before the persistence-api is activated
+          // (because it needs to find the RateLimitingManager service), so we put it in front.
+          // But as we except the actual persistence implementation jar to be the very first, we
+          // add it as 2nd.
```
dimas-b, line 494: I hope this should not be required when `DbActivator` does not declare a dependency on a `RateLimitingManager` in case it is not configured. OSGi plus `BaseActivator` should be able to resolve dependencies regardless of the order of bundle jars.

For reference: the persistence jar is put first in the list only to simplify the finding of the NodeTool class (as some other bundles may be able to load it too).

rate-limiting-global/src/main/java/io/stargate/db/limiter/global/impl/GlobalRateLimitingManager.java
```diff
@@ -0,0 +1,143 @@
...
+    if (statement instanceof SimpleStatement
+        && shouldExclude(((SimpleStatement) statement).queryString())) {
+      return RateLimitingDecision.unlimited();
+    }
+    return RateLimitingDecision.limit(limiter, 1);
+  }
+
+  @Override
+  public RateLimitingDecision forBatch(Batch batch, Parameters parameters) {
+    return RateLimitingDecision.limit(limiter, 1);
```
dimas-b, line 141: Why do we acquire 1 permit here but not `batch.size()`? I assume a permit corresponds to one "basic" query, so would it make sense to make batch cost progressively higher according to the number of queries it contains?

rate-limiting-global/src/main/java/io/stargate/db/limiter/global/impl/GlobalRateLimitingManager.java
```diff
@@ -0,0 +1,143 @@
...
+    if (rateStr == null || rateStr.isEmpty()) {
+      throw new IllegalArgumentException(
+          format(
+              "Global rate limiting is enabled but missing (or empty) value for property '%s'",
+              RATE_PROPERTY));
+    }
+
+    try {
+      return new AsyncRateLimiter(
+          executor, Long.parseLong(rateStr), TimeUnit.SECONDS, 1, TimeUnit.MINUTES);
```
tomekl007, line 75: any reason why we use SECOND and MINUTES? Wouldn't it be more consistent to use the same time-unit for both? It may be more important if both values will be provided via an external setting.

testing/src/test/java/io/stargate/it/limiter/GlobalRateLimitingTest.java
```diff
@@ -0,0 +1,46 @@
...
+  public static void buildParameters(StargateParameters.Builder builder) {
+    builder.putSystemProperties(DbActivator.RATE_LIMITING_ID_PROPERTY, "GlobalRateLimiting");
+    builder.putSystemProperties("stargate.limiter.global.rate_qps", "1");
+  }
+
+  @Test
+  @DisplayName("Should rate limit queries")
+  public void testRateLimited(CqlSession session) {
+    int QUERIES = 5;
+    // We're executing a point query on an empty table 5 times, so this should be really fast, even
```
tomekl007, line 34: Shouldn't we provide a unit test for rate limiter? I think that unit test allows to test this behavior better (with isolation and without external factors)

persistence-api/src/main/java/io/stargate/db/DbActivator.java
pcmanus: Skipped the dependency when rate limiting wasn't set.

rate-limiting-global/src/main/java/io/stargate/db/limiter/global/impl/GlobalRateLimitingManager.java
pcmanus: Yes. Thanks for catching that.

stargate-starter/src/main/java/io/stargate/starter/Starter.java
pcmanus: Indeed. I added in earlier iterations when it wasn't working and I was figuring it out. But I removed it now and it does seem that it works fine.

rate-limiting-global/src/main/java/io/stargate/db/limiter/global/impl/GlobalRateLimitingManager.java
pcmanus: Maybe. I could argue that the rate limiting is not really based on the queries "size" in that a full-table scan will be rate limited the same than a 1 row query, so it's not that clear that making a special case for batch is decidedly better or worst. But I changed it, because both are meaningful options and I don't really care. As noted in the comments in the code, I added that `GlobalRateLimitingManager` to have a real example and to have a semi-realistic test, but I don't really believe this is for using in production in this state so ....

persistence-api/src/main/java/io/stargate/db/DbActivator.java
```diff
@@ -3,28 +3,71 @@
...
-    return new ServiceAndProperties(
-        new PersistenceDataStoreFactory(persistence.get()), DataStoreFactory.class);
+  protected List<ServiceAndProperties> createServices() {
+    Persistence persistence = this.dbPersistence.get();
+    if (!RATE_LIMITING_IDENTIFIER.equalsIgnoreCase("<none>")) {
+      RateLimitingManager rateLimiter = rateLimitingManager.get();
+      if (rateLimiter == null) {
+        throw new RuntimeException(
+            String.format(
+                "Could find rate limiter service with id '%s'", RATE_LIMITING_IDENTIFIER));
```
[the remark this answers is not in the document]
pcmanus: Fixed.

persistence-api/src/main/java/io/stargate/db/RateLimitingPersistence.java
```diff
@@ -0,0 +1,168 @@
...
+
+    @Override
+    public Persistence persistence() {
+      return RateLimitingPersistence.this;
+    }
+
+    @Override
+    public void login(AuthenticatedUser user) throws AuthenticationException {
+      connection.login(user);
+      rateLimiter.onUserLogged(user);
```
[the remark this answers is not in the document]
pcmanus: Not 100% sure to follow the question. The idea is that `RateLimitingManager` is an API that could see various implementations, so some may want to track at the user granularity level. And while the one implementation in the PR does not do so, I have a another use case in mind that will use it.

persistence-api/src/main/java/io/stargate/db/limiter/AsyncRateLimiter.java
```diff
@@ -0,0 +1,218 @@
...
+
+/**
+ * An asynchronous rate limiter.
+ *
+ * <p>Conceptually, this provide functionality similar to <a
+ * href="https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/util/concurrent/RateLimiter.html">Guava
+ * RateLimiter</a>, but can be used when asynchronicity is desired and blocking is not an option.
+ * Instead of blocking to obtain "permits", if permits cannot be obtained immediately (if execution
+ * needs to be delayed to enforce the rate limit), then the task is scheduled for later execution on
+ * a predefined executor.
```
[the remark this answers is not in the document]
pcmanus: That's for the user of the `AsyncRateLimiter` to decide and that class doesn't take a stand, on purpose.

The one implementation included, `GlobalRateLimitingManager`, does use an executor with an unbounded queue but I don't even think it's create a new problem. We're putting user queries in that queue, so either client connections are limited by some other way and the number of in-flight queries is limited and this queue cannot grow indefinitely. Or client connections are not limited, and we can OOM with or without this.

Don't get me wrong: I'm not saying proper back-pressure that ensures the server never OOM isn't important. Just that this PR doesn't really makes things better or worst on that aspect.

persistence-api/src/main/java/io/stargate/db/limiter/AsyncRateLimiter.java
```diff
@@ -0,0 +1,218 @@
...
+      long rate,
+      TimeUnit rateUnit,
+      long reserveWindow,
+      TimeUnit reserveWindowUnit,
+      long schedulingThreshold,
+      TimeUnit schedulingThresholdUnit) {
+    this.executor = executor;
+    this.nanosPerPermit = rateUnit.toNanos(1) * 1.0 / rate;
+    this.reserveWindowNanos = reserveWindowUnit.toNanos(reserveWindow);
+    this.consumedToTime = new AtomicLong(System.nanoTime());
```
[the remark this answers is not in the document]
pcmanus: I've left it out for now because the unit test don't use it, and if we really want to allow some time injection, we'd also need to take care of the `System.nanoTime` in `acquireAndExecute` and I'm not sure about adding that complexity if it's unused. This can obviously be changed later.

rate-limiting-global/src/main/java/io/stargate/db/limiter/global/impl/GlobalRateLimitingManager.java
pcmanus: Not reason, but I don't see how this is in any way confusing and what "consistency" would add. And this can be changed later easily enough if the context changes.

persistence-api/src/test/java/io/stargate/db/limiter/AsyncRateLimiterTest.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Note: The test is made to pass when the infrastructure does not give it the right timings. This means that any
+ * failure, however rare or flaky, must be taken seriously as it signifies a real problem with the code.
+ */
+public class AsyncRateLimiterTest
```
dimas-b, line 18: The code fomatter does not like this file :shrug:

persistence-api/src/main/java/io/stargate/db/DbActivator.java
```diff
@@ -3,28 +3,81 @@
...
-        new PersistenceDataStoreFactory(persistence.get()), DataStoreFactory.class);
+  protected List<ServiceAndProperties> createServices() {
+    Persistence persistence = this.dbPersistence.get();
+    if (hasRateLimitingEnabled()) {
+      RateLimitingManager rateLimiter = rateLimitingManager.get();
+      if (rateLimiter == null) {
+        throw new RuntimeException(
+            String.format(
+                "Could not find rate limiter service with id '%s'", RATE_LIMITING_IDENTIFIER));
+      }
```
dimas-b, line 59: Semi-related: We should probably make `ServiceReference` do this check (for all services) and return a non-null value. If the dependency is not satisfied the `createService` method should not be invoked to begin with.

rate-limiting-global/pom.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>stargate</artifactId>
+    <groupId>io.stargate</groupId>
+    <version>1.0.5-SNAPSHOT</version>
```
dimas-b, line 6: `master` is at `1.0.7-SNAPSHOT` now

persistence-api/src/main/java/io/stargate/db/limiter/AsyncRateLimiter.java
```diff
@@ -0,0 +1,218 @@
...
+   */
+  public <T> CompletableFuture<T> acquireAndExecute(
+      long permits, Supplier<CompletableFuture<T>> task) {
+    long currentTime = System.nanoTime();
+    long scheduleTime = acquire(permits, currentTime);
+    long delay = scheduleTime - currentTime;
+
+    if (delay < schedulingThresholdNanos) {
+      // Time is in the past, or very close in the future. Execute immediately.
+      return task.get();
```
dimas-b, line 197: nit: To me personally `.get()` looks like a method that should generally be idempotent. I know that the `Supplier` contract does not impose any requirements on the implementation, still it feels a bit odd to me. What do you think about using `Callable` for the `task` parameter?

dimas-b (approved):

persistence-api/src/test/java/io/stargate/db/limiter/AsyncRateLimiterTest.java
pcmanus: My bad, forget to run it. Push-fixed.

rate-limiting-global/pom.xml
pcmanus: Fixed, thanks for spotting it.

persistence-api/src/main/java/io/stargate/db/limiter/AsyncRateLimiter.java
pcmanus: Honestly, `Callable` is more annoying because of it throwing `Exception`. Here, I can't propagate the exception up (because the `Persistence` methods don't throw `Exception` in particular) so we'd end up wrapping it in a `RuntimeException` or equivalent. To me, `Supplier` exists exactly to avoid that annoyance of `Callable` coming from a per-lambda, not to add some implicit idempotency contract, and I've see tons of use of it not idempotent so I'd be careful with running with that assumption.

So I haven't changed it, but that said, not trying to be difficult: if it bugs you enough that you'd prefer using `Callable` despite the need to sometimes wrap the resulting exception, then let me know and I'll change.

dimas-b: No worries. I too thought exceptions would be annoying, but have not tried to actually refactor :) Current code is good from my POV.

PR #634, opened 2021-02-02
Add GraphQL schema-first API (fixes #123)

### LINKED ISSUE #123: Graphql schema first workflow
For getting started guides and new developer workflows, a workflow could be to post a cassandra decorated graphql schema file and have it create the database and APIs. This will ease the burden of having to set up all the tables and APIs manually.

This differs from the migration endpoint as it will assume failure if there are any conflicts.

Full graphql example:

```
# Entities
@PartitionKey(fields: ["user_id"])
@ClusteringKey(fields: [{"cart_id": DESC}])
type CartEntry {
  userId: Uuid!
  cartId: TimeUuid!
  product: Product
}

type Product @embedded {
 name: String
 price: BigDecimal
}

#Mutations and Queries Roots
type Mutation {
  deleteCart(
    value: CartInput!
    ifNotExists: Boolean
    options: MutationOptions
  ): CartMutationResult @cql(type: DELETE, type: CartEntry)
  updateCart(
    ifCondition: CartFilterInput
    options: MutationOptions
    value: CartInput!
    ifExists: Boolean
  ): CartMutationResult @cql(type: UPDATE, type: CartEntry)
  insertCart(
    value: CartInput!
    ifNotExists: Boolean
    options: MutationOptions
  ): CartMutationResult @cql(type: INSERT, type: CartEntry)
}

type Query {
  cartByUser(
    value: CartInput
    orderBy: [CartOrder]
    options: QueryOptions
  ): Cart
}

## Api Definitions

@api(path: "v1/cartByUser", type: GET)
query getCartByUser($pageState: String) {
  cartByUser(options: { 
      pageState: $pageState
  }) {
    value {
      product {
         name
         price
      }
      quantity
      cart_count: _count(cart_id)
    }
    pageState
    hasNextPage
  }
}

@api(path: "v1/insertCartEntry", type: POST)
mutation InsertCartEntry($cartId: TimeUuid, $quantity: Int) {
  insertCart(value: { cartId: $cartId, quantity: $quantity, _jwt_claim: {userId: "user_id"} }) {
   value {
    cartId
    quantity
   }
   applied
  }
}

@api(path: "v1/updateCartQuantity", type: PUT)
mutation updateCartEntry($cartId: TimeUuid, $quantity: Int) {
  updateCart(value: { cartId: $cartId, quantity: $quantity, _jwt_claim: {userId: "$user_id"} }) {
   value {
    cartId
    quantity
   }
   applied
  }
}

@api(path: "v1/deleteCart", type: POST)
mutation deleteCartEntry($cartId: TimeUuid, $quantity: Int) {
  deleteCart(value: { cartId: $cartId, _jwt_claim: {userId: "$user_id"} }) {
   applied
  }
}
```

olim7t:
Sharing a draft early because this is a big ticket. Nothing is set in stone yet, the code is still moving a lot. I might still force-push at any time.

TODO:
- [x] more flexible rules for queries: right now only full primary key queries are supported, we should also support partial PK (returns multiple entities), arbitrary predicates on any field, etc. See [this](https://docs.datastax.com/en/developer/java-driver/4.10/manual/mapper/daos/select/) for inspiration.
- [x] map GraphQL mutations to inserts, updates and deletes.
- [x] handle conversions for all fields types: see `GeneratedFetcher.toCqlValue/toGraphqlValue` (use similar rules as `DataTypeMapping` in the legacy API). Maybe reintroduce the custom scalars of the legacy API.
- [x] "transient" types: at some point there will probably be object types that are not directly mapped to CQL, but just used as containers for query results (much like the `XxxResult` types we have in the legacy API). They will have to be marked as such, either by annotating them, or annotating the non-transient types. Also the query processing logic will have to detect that kind of situation (the entity type is not directly the result, but a field in the result type).
- [x] add a REST endpoint in the admin API to download schema contents as a file. The URI will be returned in the response of the `schema` query. See the todo in `AdminSchemaBuilder.SCHEMA_TYPE`.
- `deploySchema` mutation:
  - [x] figure out schema migrations. Currently we always drop and recreate all tables. This should be configurable, with multiple strategies: reuse existing schema (fail if it doesn't match), drop and recreate only the types that don't match, try to alter the tables... (others?)
  - [x] figure out a robust concurrency model: the mutation inserts a new entry in the `graphql_schema` table, migrates the CQL schema (in the future), and installs the new GraphQL schema. We can't do this in a fully atomic manner, but at the very least two concurrent deploys should never race with each other, and a deploy should have a minimal impact on runtime queries (I don't think we can avoid a short window where the GraphQL and CQL schemas will be out of sync, but it should be as small as possible).
  - [x] add a "dry run" mode
- [x] add a purging mechanism for the `graphql-schema` table (only keep X latest entries)
- [x] fix bug when `deploySchema` specifies `expectedVersion`, but the `graphql_schema` table does not exist yet.


graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/SchemaFetcher.java
```diff
@@ -0,0 +1,78 @@
...
+        SourceAPI.GRAPHQL);
+
+    ImmutableMap.Builder<String, Object> result = ImmutableMap.builder();
+    result.put("namespace", namespace);
+    SchemaSource source = new SchemaSourceDao(dataStore).getLatest(namespace);
+    if (source != null) {
+      UUID version = source.getVersion();
+      ZonedDateTime deployDate = source.getDeployDate();
+      String contents =
+          String.format("# Schema for '%s', version %s (%s)\n\n", namespace, version, deployDate)
```
aimiebell, line 69: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [FS: Format string should use %n rather than n (VA_FORMAT_STRING_USES_NEWLINE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#fs-format-string-should-use-n-rather-than-n-va-format-string-uses-newline).
This format string includes a newline character (\n). In format strings, it is generally  preferable to use %n, which will produce the platform-specific line separator.

aimiebell, line 69: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [FS: Format string should use %n rather than n (VA_FORMAT_STRING_USES_NEWLINE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#fs-format-string-should-use-n-rather-than-n-va-format-string-uses-newline).
This format string includes a newline character (\n). In format strings, it is generally  preferable to use %n, which will produce the platform-specific line separator.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
```diff
@@ -0,0 +1,155 @@
...
+          Collections.singletonList(namespace),
+          Collections.singletonList(SchemaSourceDao.TABLE_NAME),
+          SourceAPI.GRAPHQL);
+      return true;
+    } catch (UnauthorizedException e) {
+      return false;
+    }
+  }
+
+  private static Map<String, String> getAllHeaders(HttpServletRequest request) {
```
tomekl007, line 143: We don't need this method. We should use
``` 
Map<String, String> headers = RequestToHeadersMapper.getAllHeaders(httpRequest);
```
instead.

olim7t:
Force-pushed to rebase on latest master.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -0,0 +1,311 @@
...
+
+  private void ensureTableExists() throws Exception {
+    dataStore
+        .execute(
+            dataStore
+                .queryBuilder()
+                .create()
+                .keyspace(KEYSPACE_NAME)
+                .ifNotExists()
+                .withReplication(Replication.simpleStrategy(1))
```
dimas-b, line 204: Could this be flaky is realistic replicated environments? With `RF==1` if one of the storage nodes is taken down for maintenance, this keyspace will become unavailable to a portion of clients :thinking:

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/NamespaceResource.java
```diff
@@ -0,0 +1,141 @@
...
+import javax.ws.rs.container.Suspended;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Serves the GraphQL schema deployed on each namespace. */
+@Singleton
+@Path(ResourcePaths.NAMESPACES)
```
dougwettlaufer, line 45: Would it make sense to add swagger to the REST APIs in here? I can create an issue if we think it'd be beneficial

dougwettlaufer: Now that I look FilesResource is the only rest in here?

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/NamespaceResource.java
```diff
@@ -0,0 +1,141 @@
...
+      if (graphql == null) {
+        replyWithGraphqlError(
+            Response.Status.NOT_FOUND,
+            String.format(
+                "Could not find a GraphQL schema for '%s', either the namespace does not exist, "
+                    + "or no schema was deployed to it yet.",
+                namespace),
+            asyncResponse);
+        return null;
+      } else if (!isAuthorized(request, namespace)) {
```
dougwettlaufer, line 123: Why not authorize before checking? Checking first could leak whether something does or doesn't exist.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
dougwettlaufer: Yeah I think we're going to want to make this configurable.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlResourceBase.java
```diff
@@ -137,16 +154,164 @@ protected static void postJson(
     executeAsync(input.build(), graphql, asyncResponse);
   }
 
+  /**
+   * Handles a GraphQL POST request that uses the {@link MediaType#MULTIPART_FORM_DATA} content
```
dougwettlaufer, line 158: Nice javadoc!

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/FieldTypeCache.java
```diff
@@ -91,51 +91,20 @@ private ColumnType normalize(ColumnType type) {
 
   protected GraphQLScalarType getScalar(Type type) {
```
dougwettlaufer, line 92: Why are all the types being removed here? Isn't cql-first staying the same?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaFetcherBase.java
```diff
@@ -0,0 +1,119 @@
...
+    Keyspace keyspace = dataStore.schema().keyspace(namespace);
+    if (keyspace == null) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Namespace '%s' does not exist. "
+                  + "Use the 'createNamespace' mutation to create it first.",
+              namespace));
+    }
+
+    authorizationService.authorizeSchemaWrite(
```
dougwettlaufer, line 63: This has changed. Might want to sync up with master.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
dougwettlaufer: Plus wrap it in a flag for whether or not SG should be allowed to create the keyspace and table on the fly.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -0,0 +1,320 @@
...
+        allSchemasForNamespace.size() - NUMBER_OF_RETAINED_SCHEMA_VERSIONS;
+    if (numberOfEntriesToRemove > 0) {
+      LOGGER.info("Removing {} old schema entries.", numberOfEntriesToRemove);
+
+      // remove N oldest entries
+      SchemaSource mostRecentToRemove =
+          allSchemasForNamespace.get(NUMBER_OF_RETAINED_SCHEMA_VERSIONS);
+
+      BoundQuery deleteSchemaQuery =
+          dataStore
```
dougwettlaufer, line 309: All of the references to datastore in here are an "internal" datastore aren't they? These queries all need to be in the context of an authenticatedUser so that the correct keyspace is used.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -0,0 +1,320 @@
...
+    List<Row> row = dataStore.execute(schemaQuery(namespace)).get().rows();
+    if (row == null) {
+      return Collections.emptyList();
+    }
+    return row.stream().map(r -> toSchemaSource(namespace, r)).collect(Collectors.toList());
+  }
+
+  public SchemaSource getByVersion(
+      String namespace,
+      @SuppressWarnings("OptionalUsedAsFieldOrParameterType") Optional<UUID> maybeVersion)
```
dougwettlaufer, line 97: Nit: It feels odd that in a method named `getByVersion` the version is optional. Should this be named `getSchema`?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/CreateNamespaceFetcher.java
```diff
@@ -0,0 +1,131 @@
...
+        // Unexpected technical error: rethrow directly
+        return errorResult(e, environment).build();
+      }
+    }
+
+    ImmutableMap.Builder<String, Object> data = ImmutableMap.builder();
+    data.put("query", environment.getGraphQLSchema().getQueryType());
+
+    Keyspace keyspace = dataStore.schema().keyspace(name);
+    if (isReadAuthorized(keyspace, authenticationSubject)) {
```
dougwettlaufer, line 102: Is this necessary since we're essentially returning what was sent in?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaResponseDto.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public UUID getVersion() {
+    return version;
+  }
+
+  public void setLogs(List<ProcessingMessage<ProcessingLogType>> logs) {
+    this.logs = logs;
+  }
+
+  public List<Map<String, Object>> getLogs(DataFetchingEnvironment environment) {
```
dougwettlaufer, line 34: It doesn't look like this or `getCqlChanges` below are used. Is that right or is Intellij confused?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/SchemaFetcher.java
```diff
@@ -0,0 +1,52 @@
...
+
+public abstract class SchemaFetcher<ResultT> extends CassandraFetcher<ResultT> {
+  public SchemaFetcher(
+      AuthorizationService authorizationService, DataStoreFactory dataStoreFactory) {
+    super(authorizationService, dataStoreFactory);
+  }
+
+  protected void authorize(AuthenticationSubject authenticationSubject)
+      throws UnauthorizedException {
+    authorizationService.authorizeSchemaRead(
```
dougwettlaufer, line 37: Would it be possible someone would have access to the `schema_source` table but not the actual table they're getting the schema for? Just wondering if we should be authorizing the `stargate_graphql` keyspace and the namespace they're working with?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/SingleNamespaceFetcher.java
```diff
@@ -0,0 +1,48 @@
...
+  protected Map<String, Object> get(
+      DataFetchingEnvironment environment,
+      DataStore dataStore,
+      AuthenticationSubject authenticationSubject) {
+
+    String name = environment.getArgument("name");
+    Keyspace keyspace = dataStore.schema().keyspace(name);
+
+    if (keyspace == null || !isReadAuthorized(keyspace, authenticationSubject)) {
+      throw new IllegalArgumentException(String.format("Unknown keyspace '%s'", name));
```
dougwettlaufer, line 43: Does this error get returned to the end user? If it does should the message be "Unknown namespace"?

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
olim7t: Yes, this is the default if the keyspace gets created by the first deployment, and we have to use a common denominator because we don't know the cluster topology. Ideally it should be created ahead of time by an administrator with a higher RF, I'll make sure to mention it in the docs.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/NamespaceResource.java
olim7t: I'm not sure if that adds a lot of value. These resources are a direct implementation of the standard way of serving GraphQL over HTTP, as specified [here|https://graphql.org/learn/serving-over-http/]. Users will typically use a client or framework that abstracts those details, so I don't think they'll look for Swagger docs.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/NamespaceResource.java
olim7t: IIRC we had that discussion for the CQL-first API. If we don't want to reveal whether it's missing vs forbidden, then we would have to pick a single error code for both conditions.

Side note: the user is already authentified at this point.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/FieldTypeCache.java
olim7t: Nothing changed, the logic was just moved to `CqlScalar.fromCqlType(type)` (which is possible because `CqlScalar` was turned into an enum).

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaResponseDto.java
olim7t: graphql-java calls them at runtime, it matches the GraphQL field to the name of the property. We have to use POJOs for fields that can be parameterized, that doesn't work with the map approach.

And in fact I wished we had relied on POJOs more from the start, it's a cleaner approach.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/SchemaFetcher.java
olim7t: Right, it would make sense to also check the keyspace here.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/SingleNamespaceFetcher.java
olim7t: Yes.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaFetcherBase.java
olim7t: I see. Rebasing right now.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
olim7t: Hrmm I need to check. Things should be fine in the fetchers, but there might be some issues in `SchemaFirstCache`.  We're planning to merge the two caches anyway, so it will be a good moment to revisit this.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
olim7t: Yes, it's more like "get a single version". I'll clean up all method names.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/CreateNamespaceFetcher.java
olim7t: I was thinking of the case where `ifNotExists == true`, and the keyspace already existed.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/util/DirectedGraph.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  /**
+   * this assumes that {@code from} and {@code to} were part of the vertices passed to the
+   * constructor
+   */
+  public void addEdge(VertexT from, VertexT to) {
+    Preconditions.checkArgument(vertices.containsKey(from) && vertices.containsKey(to));
+    adjacencyList.put(from, to);
+    vertices.put(to, vertices.get(to) + 1);
```
dougwettlaufer, line 64: For the layman can you please add a comment about what the increment means here?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/util/DirectedGraph.java
```diff
@@ -0,0 +1,104 @@
...
+      for (VertexT successor : adjacencyList.get(vertex)) {
+        if (decrementAndGetCount(successor) == 0) {
+          queue.add(successor);
+        }
+      }
+    }
+
+    if (result.size() != vertices.size()) {
+      throw new IllegalArgumentException("failed to perform topological sort, graph has a cycle");
+    }
```
dougwettlaufer, line 93: Soooo, my graph theory is a little rusty... Can you please add some comments in here for people like me? :)

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/FieldTypeCache.java
dougwettlaufer: Ah, gotcha.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/NamespaceResource.java
dougwettlaufer: Sure but isn't FilesResource.java actual REST? Or am I misunderstanding and it's also graphql over http?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/AdminSchemaBuilder.java
```diff
@@ -0,0 +1,622 @@
...
+      newObject()
+          .name("Query")
+          .field(NAMESPACES_QUERY)
+          .field(NAMESPACE_QUERY)
+          .field(SCHEMA_QUERY)
+          .field(SCHEMA_HISTORY_PER_NAMESPACE_QUERY)
+          .build();
+
+  // See https://graphql-rules.com/rules/mutation-payload-query
+  private static GraphQLFieldDefinition QUERY_FIELD =
```
dougwettlaufer, line 243: Should this one be final too?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/AdminSchemaBuilder.java
```diff
@@ -0,0 +1,622 @@
...
+
+  private static final GraphQLScalarType UPLOAD_SCALAR =
+      newScalar().name("Upload").coercing(new InputStreamCoercing()).build();
+
+  private static final GraphQLFieldDefinition DEPLOY_SCHEMA_FILE_MUTATION =
+      deploySchemaStart()
+          .name("deploySchemaFile")
+          .description(
+              "Deploys a GraphQL schema to a namespace via a file upload.\n"
+                  + "This mutation must be executed with a [multipart request](https://github.com/jaydenseric/graphql-multipart-request-spec) "
```
dougwettlaufer, line 532: Does this actually render markdown?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/AdminSchemaBuilder.java
```diff
@@ -0,0 +1,622 @@
...
+  private static final GraphQLObjectType MUTATION =
+      newObject()
+          .name("Mutation")
+          .field(CREATE_NAMESPACE_MUTATION)
+          .field(DROP_NAMESPACE_MUTATION)
+          .field(DEPLOY_SCHEMA_MUTATION)
+          .field(DEPLOY_SCHEMA_FILE_MUTATION)
+          .build();
+
+  private final AuthenticationService authenticationService;
```
dougwettlaufer, line 551: Does this need to be here? Doesn't seem to be used.

dougwettlaufer (commented):
Looks good!

ivansenic:
@olim7t I noticed following things when I was creating the nosqlbench's.. I am pasting these here directly, I am not sure how many of those are relevant. I am also gonna go through your PR today :factory_worker:

* Dropping namespace does not drop the schema created in this namespace.. Not sure if this is indented behavior, but if I re-create the namespace with the same name as the one dropped, schema will be there and deploy schema will ask for the version. Seems odd for me..
* I did not manage to drop the schema completely, again not sure if this is indented
* I hit the query where as GQL float I specify `-Infinity`, so as I understood this is not supported by the GQL, but it seems that it returns `200` back.. The warning is printed, but imo this query should result in 400, here's the log in question:

```
127.0.0.1 - - [29/Mar/2021:14:20:42 +0000] "POST /graphqlv2/namespace/test1 HTTP/1.1" 200 1284 "-" "Java-http-client/15.0.1" 18
WARN  [2021-03-29 14:20:42,498] notprivacysafe.graphql.GraphQL: Query failed to parse : 'mutation {
  insertIot(iot: {machine_id: "000b99e5-31d9-421d-90ab-aa41f3b19a08", sensor_name: "clarity", time: "1970-01-01T01:00:00Z", sensor_value: -Infinity, station_id: "28df63b7-cc57-43cb-9752-fae69d1653da", data: "+Lorem+ipsum+dolor+sit+amet%2C+consectetur+adipiscing+elit.+Curabitur+non+aliquam+massa.+Nulla+sagittis+mattis+lorem%2C+in+dignissim+turpis+blandit+sit+amet.%0ASuspendisse+eu+placerat+sem.+Maecenas+rhoncus%2C+enim+sed+pharetra+pharetra%2C+turpis+justo+imperdiet+justo%2C+eget+aliquet+elit+enim+eget+orci.+Proin+rhoncus+ut+justo+eget+egestas.+Donec+quis+mattis+tellus%2C+vel+tempus+nibh.+Vestibulum+tincidunt%2C+ligula+et+posuere+ultrices%2C+dolor+nibh+pellentesque+ante%2C+in+pharetra+nibh+risus+ac+nisi.+Nam+ornare+nibh+est.+Curabitur+aliquet+sed+erat+eu+venenatis.%0AProin+iaculis+libero+at+tellus+lacinia%2C+vel+scelerisque+tellus+lobortis.+Sed+quis+massa+nec+arcu+euismod+lacinia.+Suspendisse+consequat+purus+sed+cursus+ultricies.+Nunc+faucibus+sem+risus%2C+sit+amet+euismod+ipsum+pretium+eget.+Fusce+blandit+turpis+ac+leo+scelerisque+tempor.+Suspendisse+pretium+dolor+rutrum%2C+dignissim+urna+vitae%2C+consequat+sem.+Maecenas+commodo+laoreet+arcu+et+pretium.+Sed+a+enim+vel+lectus+semper+semper.+Nulla+ac+consectetur+tortor.+Maecenas+a+enim+lacus.+Vivamus+consectetur%2C+magna+eu+placerat+tempus"}) {
    machine_id
    sensor_name
    time
    sensor_value
    station_id
    data
  }
}
'
127.0.0.1 - - [29/Mar/2021:14:20:42 +0000] "POST /graphqlv2/namespace/test1 HTTP/1.1" 200 199 "-" "Java-http-client/15.0.1" 11
``` 
* What would be the correct input for the `pagingState: String @cql_pagingState` and how can I state the limit of the returned results. I am also not aware of how this aligns with this `@cql_select(pageSize: 10)`, can you please maybe explain me this example below.. In my tests `pagingState` is always `null`:
```
type SelectFooResult @cql_payload {
  data: [Foo]
  pagingState: String
}
type Query {
  foos(
    pk1: Int!,
    pk2: Int!,
    pagingState: String @cql_pagingState
  ): SelectFooResult @cql_select(pageSize: 10)
}
```

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/util/DirectedGraph.java
olim7t: It's [Kahn's algorithm](https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm), I'll add more explanations to the code.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/AdminSchemaBuilder.java
olim7t: Yes, per the GraphQL specification descriptions support Markdown.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/AdminSchemaBuilder.java
olim7t: Good catch, it's a leftover.

olim7t:
> Dropping namespace does not drop the schema created in this namespace

Makes sense, I hadn't thought about it.

> I did not manage to drop the schema completely, again not sure if this is indented

Yes, we need an `undeploy` operation, I plan to add that soon.

>  it seems that it returns 200 back.. The warning is printed, but imo this query should result in 400

So I looked into that a bit, and it looks like the commonly accepted answer is that HTTP errors codes should be reserved for HTTP-level failures. For GraphQL-level errors, HTTP 200 is returned, and clients will rely on the `errors` field in the JSON response.

> What would be the correct input for the pagingState: String @cql_pagingState and how can I state the limit of the returned results.

From a user POV, the paging state is an opaque string, you can only use what was returned by a previous query, to resume the iteration. See `SelectTest.selectFullPartitionWithPagination` for an example.
In practice it's a binary string encoded in base64.

> I am also not aware of how this aligns with this `@cql_select(pageSize: 10)`

If you use page size 10, then the query should return at most 10 results, with a paging state that allows you to start at the 11th element if you pass it back with the next query. Did you have more than 10 elements in the partition in your test?

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/NamespaceResource.java
ivansenic: I think it's supporting stuff, where you actually expose the schema and directives, thus I would agree with @olim7t and not add this to Swagger as this should be used by a GQL client..

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
```diff
@@ -0,0 +1,164 @@
...
+        .header("Content-Disposition", "inline; filename=\"cql_directives.graphql\"")
+        .build();
+  }
+
+  @GET
+  @Path("/namespace/{namespaceName}.graphql")
+  @Produces(MediaType.TEXT_PLAIN)
+  public Response getSchema(
+      @HeaderParam("X-Cassandra-Token") String token,
+      @PathParam("namespaceName") String namespace,
```
ivansenic, line 86: so in the Namespace resource you test this against a pattern in order to prevent XSS attack, but here you are not doing this.. Anything different here?

graphqlapi/src/main/java/io/stargate/graphql/web/models/GraphqlJsonBody.java
```diff
@@ -32,4 +32,18 @@ public void setOperationName(String operationName) {
   public void setVariables(Map<String, Object> variables) {
     this.variables = variables;
   }
+
+  @Override
+  public String toString() {
```
ivansenic, line 37: so this class does not exists in the GraphQL API itself? Strange that we had to build it..

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/util/Uuids.java
```diff
@@ -0,0 +1,225 @@
+package io.stargate.graphql.schema.schemafirst.util;
```
ivansenic, line 1: This class misses the Stargate authors notice.. Also a link where did you copy this from would be nice (hope you were not writing this from scratch)

ivansenic (commented):
Pufff, this is so huge, there is no chance that I can read all of it and understand it.. But I guess I got the generic idea, however going into each fetcher and each processor is really challenging.. We should never again have such huge PRs, for example this cloud have started only with the admin endpoint imo.. But anyway this is insane amount of work :clap:, this can be a product of it's own in fact.. Well done Olivier!

Here are some general things:

* More Java docs would be helpful, also more inline comments, splitting of huge methods in smaller ones, the code is not really reader friendly.. This stands especially true for the Schema processing and all the models & builders..
* since the queries/mutations and fetchers are so closely related (like expecting the names of the arguments to match, etc) I would really consider joining these.. Meaning that fetchers exposes the `GraphQLFieldDefinition` operating on.. I am not really sure if this is doable at all, but at least you would be able then to also hard code stuff and actually connect on what operation type is fetcher working on..
* I would strongly consider that all fetchers that operate on the `dataStore` to be in the async mode, this has to be supported as the `AsyncDataFetcher` exists and thus returning `CompletableFuture<T>` should be doable.. Of course this would be very time consuming, but could in the long run improve the performance a bit..

Here are some usability questions:
* We do support arrays right? So if I have a field in the type of `[String]`
 this will just be mapped to Casandra list. What happens for the same in case of the User Defined Type (`@cql_entity(target: UDT)`)

And some things I did not do yet:
* was not checking tests
* was not checking the migrations at all (need more context here)
* in fact I only checked 52/117 files for now

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/NamespaceResource.java
olim7t: I see what you mean now, it's done differently in the CQL-first resource. This will get taken care of when I merge the two resources.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
olim7t: Just to confirm, I'm working on the merge now and this will indeed get fixed in the process.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -0,0 +1,319 @@
...
+    dataStore.execute(updateDeploymentToNotInProgress).get();
+  }
+
+  public void purgeOldVersions(String namespace) throws Exception {
+    List<SchemaSource> allSchemasForNamespace = getAllVersions(namespace);
+
+    int numberOfEntriesToRemove =
+        allSchemasForNamespace.size() - NUMBER_OF_RETAINED_SCHEMA_VERSIONS;
+    if (numberOfEntriesToRemove > 0) {
+      LOGGER.info("Removing {} old schema entries.", numberOfEntriesToRemove);
```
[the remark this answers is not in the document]
olim7t: It's maybe a bit deceptive because, since we purge on each insert, `numberOfEntriesToRemove` will always be 1 in practice. The only way to observe something higher would be to insert entries manually.

> which btw is not even the last one

Are you sure? The table is clustered by `version DESC`, and `version` is a version-1 UUID, so I think we're good. But if you can think of a failing scenario we can add a test in `SchemaDeploymentTest`.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -0,0 +1,319 @@
...
+                  CONTENTS_COLUMN_NAME, Column.Kind.Regular, Column.Type.Varchar),
+              ImmutableColumn.create(
+                  LATEST_VERSION_COLUMN_NAME, Column.Kind.Static, Column.Type.Timeuuid),
+              ImmutableColumn.create(
+                  DEPLOYMENT_IN_PROGRESS_COLUMN_NAME, Column.Kind.Static, Column.Type.Boolean))
+          .build();
+
+  private final DataStore dataStore;
+
+  public SchemaSourceDao(DataStore dataStore) {
```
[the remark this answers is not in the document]
olim7t: In theory I wholeheartedly agree. It was my plan to make everything async and I tried in [this PR](https://github.com/stargate/stargate/pull/544#issuecomment-758786043). Unfortunately I hit a wall because our internal API is not fully async, due to the fact that the Cassandra code we rely on was not designed that way either.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DropNamespaceFetcher.java
```diff
@@ -0,0 +1,65 @@
...
+      DataStore dataStore,
+      AuthenticationSubject authenticationSubject)
+      throws Exception {
+
+    String name = environment.getArgument("name");
+    boolean ifExists = environment.getArgumentOrDefault("ifExists", Boolean.FALSE);
+
+    authorizationService.authorizeSchemaWrite(
+        authenticationSubject, name, null, Scope.DROP, SourceAPI.GRAPHQL, ResourceKind.KEYSPACE);
+
```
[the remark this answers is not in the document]
olim7t: 👍 
I'm creating a ticket.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/processor/MappingModel.java
```diff
@@ -0,0 +1,193 @@
...
+    // Operation types can have custom names if the schema is explicitly declared, e.g:
+    //   schema { query: MyQueryRootType }
+
+    String typeName =
+        registry
+            .schemaDefinition()
+            .flatMap(
+                schema -> {
+                  for (OperationTypeDefinition operation : schema.getOperationTypeDefinitions()) {
+                    if (operation.getName().equals(fieldName)) {
```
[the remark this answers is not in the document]
olim7t: I generally use the null-safe variant when the arguments are actually nullable. Here `operation.getName()` comes from graphql-java, and there's no way a null name can be produced from a valid schema. That would be a serious bug in the library, so I'd rather crash hard with an NPE than potentially missing it.

graphqlapi/src/main/java/io/stargate/graphql/web/models/GraphqlJsonBody.java
olim7t: graphql-java focuses on the core features, it's agnostic to the transport, I think that's why. The closest concept is `graphql.ExecutionInput`, but it's already in object form.

And for context, there is also a higher-level GraphQL servlet library. We used to use it, but decided to drop it for various reasons, see stargate/stargate#405.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
olim7t: Nope, that's an oversight. I'll add the check here too.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/SchemaFirstCache.java
```diff
@@ -0,0 +1,181 @@
...
+        result.put(namespace, holder);
+      }
+    }
+    return result;
+  }
+
+  class GraphqlHolder {
+    private final SchemaSource source;
+    private final Keyspace keyspace;
+    // This future is used to guarantee that the cached value is built only twice, even if two
```
olim7t, line 141: > built only twice

🤦 

I've fixed it directly in #812.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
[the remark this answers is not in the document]
ivansenic: I saw later that you are purging on every insert.. But still the method does not do what's supposed imo.. Imagine there is exception between insert and purge, you will have extra schema hanging forever as soon as this was insert that created the 11th version.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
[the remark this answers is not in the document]
ivansenic: Yea exactly, I think there should be a way to overcome this problem. Let's keep this commit for later. You can always wrap sync code into `CompletableFuture` (there should be no difference if the thread running it is HTTP one or the one from the executor, as you can have dedicated executors for this, that Cassandra code has no access to) or exec synchronously and then switch to async.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/processor/MappingModel.java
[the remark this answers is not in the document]
ivansenic: Fine.

graphqlapi/src/main/java/io/stargate/graphql/web/models/GraphqlJsonBody.java
ivansenic: Good to know, resolving.

dimas-b (approved):

mpenick (approved):
+1 I don't understand the full extent of all changes, but I have been looking in periodically.

PR #648, opened 2021-02-08
Test whether schema argreement is achievable during the liveness check.

### LINKED ISSUE #636: Include schema agreement into the liveness check
Stargate nodes act as fat clients from the perspective of storage nodes. This naturally requires Stargate to sync schema on every restart.

In resource-constrained environments there is a probability for schema pull requests to time out.

If Stargate ndoe is not able to sync schema for a prolonged period of time, its usability is greatly affected. Therefore, this issue proposed to include schema agreement in the `/liveness` checks.

If Stargate is not able to sync schema and its local schema version is not changing over time, it should fail [liveness](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) checks so that the infrastructure layer can act on that information and, for example, force restart the SG node.

dimas-b:
Add a separate health check for verifying whether schema agreement is achievable.

Fixes #636

### DIFF
--- a/health-checker/src/main/java/io/stargate/health/SchemaAgreementChecker.java
+++ b/health-checker/src/main/java/io/stargate/health/SchemaAgreementChecker.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.health;
+
+import com.codahale.metrics.health.HealthCheck;
+import io.stargate.db.Persistence;
+import java.util.Collection;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceReference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SchemaAgreementChecker extends HealthCheck {
+  private static final Logger logger = LoggerFactory.getLogger(SchemaAgreementChecker.class);
+
+  private static final String DB_PERSISTENCE_IDENTIFIER =
+      System.getProperty("stargate.persistence_id", "CassandraPersistence");
+  private static final String DB_PERSISTENCE_FILTER =
+      String.format("(Identifier=%s)", DB_PERSISTENCE_IDENTIFIER);
+
+  private final BundleContext context;
+
+  public SchemaAgreementChecker(BundleContext context) {
+    this.context = context;
+  }
+
+  @Override
+  protected Result check() {
+    try {
+      Collection<ServiceReference<Persistence>> services =
+          context.getServiceReferences(Persistence.class, DB_PERSISTENCE_FILTER);
+
+      // at most one persistence implementation is expected to be selected
+      for (ServiceReference<Persistence> service : services) {
+        Persistence persistence = context.getService(service);
+        try {
+          if (persistence.isInSchemaAgreement()) {
+            return Result.healthy("All schemas agree");
+          }
+
+          if (persistence.isSchemaAgreementAchievable()) {
+            return Result.healthy("Waiting for schema agreement");
+          }
+
+          return Result.unhealthy("Schema agreement is not achievable");
+        } finally {
+          context.ungetService(service);
+        }
+      }
+
+      return Result.healthy("Persistence is not initialized yet");
+
+    } catch (Exception e) {
+      logger.warn("Schema agreement check failed with {}", e.getMessage(), e);
+      return Result.unhealthy("Unable to check schema agreement: " + e);
+    }
+  }
+}
--- a/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
+++ b/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
@@ -308,9 +308,13 @@ public class CassandraPersistence
         <= 1;
   }
 
+  /**
+   * This method indicates whether storage nodes (i.e. excluding Stargate) agree on the
+   * schema version among themselves.
+   */
   private boolean isStorageInSchemaAgreement() {
     // See comment in isInSchemaAgreement()
-    // Here we also exclude Stargate nodes (isGossipOnlyMember)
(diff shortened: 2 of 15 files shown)


health-checker/src/main/java/io/stargate/health/SchemaAgreementChecker.java
```diff
@@ -0,0 +1,71 @@
...
+
+import com.codahale.metrics.health.HealthCheck;
+import io.stargate.db.Persistence;
+import java.util.Collection;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceReference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SchemaAgreementChecker extends HealthCheck {
```
tomekl007, line 26: why don't extend `BaseActivator` for getting services?

health-checker/src/main/java/io/stargate/health/SchemaAgreementChecker.java
```diff
@@ -0,0 +1,71 @@
...
+  private final BundleContext context;
+
+  public SchemaAgreementChecker(BundleContext context) {
+    this.context = context;
+  }
+
+  @Override
+  protected Result check() {
+    try {
+      Collection<ServiceReference<Persistence>> services =
```
tomekl007, line 43: in what cases it can return > 1 `Persistence`? I thought that the stargate user can only pick one of the supported `Persistence` implementations

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -110,6 +113,24 @@
   private static final int STARTUP_DELAY_MS =
       Integer.getInteger("stargate.startup_delay_ms", 3 * MigrationManager.MIGRATION_DELAY_IN_MS);
 
+  // SCHEMA_SYNC_GRACE_PERIOD should be longer than MigrationManager.MIGRATION_DELAY_IN_MS to allow
+  // the schema pull tasks to be initiated, plus some time for executing the pull requests plus
+  // some time to merge the responses. By default the pull task timeout is equal to
+  // DatabaseDescriptor.getRpcTimeout() (10 sec) and there are no retries. We assume that the merge
+  // operation should complete within the default MIGRATION_DELAY_IN_MS.
```
tomekl007, line 120: the described logic seems very complex to test and hard to tweak. Do we need to expose the possibility to set it via public `schema_sync_grace_period_ms`? In what cases we may need to change it?

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -285,6 +308,28 @@ public boolean isInSchemaAgreement() {
         <= 1;
   }
 
+  private boolean isStorageInSchemaAgreement() {
+    // See comment in isInSchemaAgreement()
```
tomekl007, line 312: why not create a standard java-doc for this method(Instead of a simple comment)?
I was looking at the comment in the `isInSchemaAgreement()` and there is no mention of the `Stargate` node directly, only dead nodes: `and waiting for failed nodes is not a great idea.`
It also does not execute the `isGossipOnlyMember` method. Could you elaborate more?

persistence-common/src/main/java/io/stargate/db/datastore/common/util/HealthCheckWithGracePeriod.java
```diff
@@ -0,0 +1,64 @@
...
+    this.timeSource = timeSource;
+  }
+
+  protected abstract boolean isHealthy();
+
+  protected void reset() {
+    failureTimestamp.set(-1);
+  }
+
+  public boolean check() {
```
tomekl007, line 38: nit: add simple java-doc explaining the behavior and parameters

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -531,4 +587,31 @@ private Object queryOrId(Statement statement) {
...
+  private static class PullRequestGetter {
+    private static final Field schedulerField;
+    private static final Method nonCompletedPullRequestsMethod;
+    private static final Object scheduler; // PullRequestScheduler
+
+    static {
+      try {
+        schedulerField = MigrationManager.class.getDeclaredField("scheduler");
+        schedulerField.setAccessible(true);
+        scheduler = schedulerField.get(MigrationManager.instance);
```
tomekl007, line 600: why not cast to `(MigrationManager`) and change the `scheduler` type from `Object` to `MigrationManager`?

health-checker/src/main/java/io/stargate/health/SchemaAgreementChecker.java
dimas-b: This checker is created and registered by `HealthCheckerActivator` among other checkers. AFAIK we can only have one activator per bundle.

health-checker/src/main/java/io/stargate/health/SchemaAgreementChecker.java
dimas-b: Currently we can have at most one `Persistence`, but hypothetically OSGi allows many.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
dimas-b: This will only need to be changed in the defaul RPC timeouts and `MIGRATION_DELAY_IN_MS` are changed or if schema merges are very slow for some reason... so not likely.

However, we cannot directly derive the value for this setting from other settings because we do not know how long the merge operation is going to take, which is why I made it configurable.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
dimas-b: `scheduler` is of type `PullRequestScheduler`, which is not `public`

persistence-common/src/main/java/io/stargate/db/datastore/common/util/HealthCheckWithGracePeriod.java
dimas-b: javadoc added

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
dimas-b: I've added javadoc and updated the _next_ comment line. I hope it clarified what is happening in this method, but please let me know if you think more info would be beneficial.

`isGossipOnlyMember` effectively means the endpoint is a Stargate node.

dougwettlaufer (approved):

tomekl007 (approved):

PR #649, opened 2021-02-09
Update the high level arch guide per Paras

polandll:


polandll:
Actually, I just realized that this diagram ONLY has DSE, and not Cassandra, so I'm waiting for a reply from Paras.

olim7t (approved):

denisekgosnell:
@polandll and @dougwettlaufer It is great to show new backends that we support. However, I am -1 on switching the diagram to the one in the PR because it implies that Stargate only supports DSE.

A diagram that shows that we support DSE and different versions of Cassandra is okay.

dougwettlaufer (changes requested):
Marking needs work to avoid accidentally merging before we get the new diagram.

msmygit:
Also, what happened to the `gRPC*` and `Events*` based extensions? Shouldn't we include them back?

dougwettlaufer:
Closing per discussion with @polandll

PR #650, opened 2021-02-09
Add an admin query to list archived schema versions

tomekl007:


graphqlapi/src/test/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/SchemasPerNamespaceFetcherTest.java
```diff
@@ -0,0 +1,91 @@
...
+        .contains(
+            new AbstractMap.SimpleEntry<>("namespace", namespace),
+            new AbstractMap.SimpleEntry<>("version", schemaSource1.getVersion()),
+            new AbstractMap.SimpleEntry<>("deployDate", schemaSource1.getDeployDate()));
+    Map<String, Object> secondSchema = result.get(1);
+    assertThat(secondSchema)
+        .contains(
+            new AbstractMap.SimpleEntry<>("namespace", namespace),
+            new AbstractMap.SimpleEntry<>("version", schemaSource2.getVersion()),
+            new AbstractMap.SimpleEntry<>("deployDate", schemaSource2.getDeployDate()));
```
olim7t, line 75: Checking the raw result like this is a bit tedious. It would look better if these tests executed GraphQL queries.
I think we have all the pieces already, `GraphQlTestBase` has the infrastructure to load a schema and run it against a mock datastore. `ScalarsDmlTest#shouldEncodeScalarValues` shows how to mock a result set.

If you don't want to sidetrack the feature we can create a follow-up ticket.

olim7t (changes requested):
Looks good, a few minor comments inline but they should be quick to fix.

graphqlapi/src/test/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/SchemasPerNamespaceFetcherTest.java
tomekl007: I agree that `GraphQlTestBase` test adds using the graphql schema, but it also mocks everything underneath.
I wonder that maybe we should take a more fully IT approach when testing graph first: 
-use the underlying persistence via ccm
- execute queries via graphql rest API
- validate the data in the actual persistence
It may give us a lot more confidence. Regardless of the approach we take, I think we should implement (and discuss) that in the follow-up ticket to cover more graphql-first code. I am happy to work on it.

olim7t (approved):

PR #658, opened 2021-02-10
Add more CQL migration strategies

olim7t:
TODO:
- [x] topologically sort migration queries: we can't drop a UDT before any UDT/table that references it, etc.

### DIFF
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaFetcherBase.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaFetcherBase.java
@@ -15,6 +15,7 @@
  */
 package io.stargate.graphql.schema.schemafirst.fetchers.admin;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import graphql.GraphqlErrorHelper;
 import graphql.schema.DataFetchingEnvironment;
@@ -29,6 +30,9 @@ import io.stargate.db.schema.Keyspace;
 import io.stargate.graphql.persistence.schemafirst.SchemaSource;
 import io.stargate.graphql.persistence.schemafirst.SchemaSourceDao;
 import io.stargate.graphql.schema.CassandraFetcher;
+import io.stargate.graphql.schema.schemafirst.migration.CassandraMigrator;
+import io.stargate.graphql.schema.schemafirst.migration.MigrationQuery;
+import io.stargate.graphql.schema.schemafirst.migration.MigrationStrategy;
 import io.stargate.graphql.schema.schemafirst.processor.ProcessedSchema;
 import io.stargate.graphql.schema.schemafirst.processor.ProcessingMessage;
 import io.stargate.graphql.schema.schemafirst.processor.SchemaProcessor;
@@ -69,25 +73,43 @@ abstract class DeploySchemaFetcherBase extends CassandraFetcher<Map<String, Obje
 
     String input = getSchemaContents(environment);
     UUID expectedVersion = getExpectedVersion(environment);
+    MigrationStrategy migrationStrategy = environment.getArgument("migrationStrategy");
+    boolean dryRun = environment.getArgument("dryRun");
 
+    ImmutableMap.Builder<String, Object> response = ImmutableMap.builder();
     ProcessedSchema processedSchema =
         new SchemaProcessor(authenticationService, authorizationService, dataStoreFactory)
             .process(input, keyspace);
-
-    // TODO this is racy -- figure out a way to handle concurrency better
-    SchemaSource newSource =
-        new SchemaSourceDao(dataStore).insert(namespace, expectedVersion, input);
-    processedSchema.dropAndRecreateSchema(dataStore);
-
-    // TODO update SchemaFirstCache from here instead of letting it reload from the DB
-
-    return ImmutableMap.of(
-        "version",
-        newSource.getVersion(),
+    response.put(
         "messages",
         processedSchema.getMessages().stream()
             .map(this::formatMessage)
             .collect(Collectors.toList()));
+
+    // TODO the rest of the method is racy -- we should at least ensure that two deploy mutations
+    // can't compete with each other (maybe add a "migration in progress" column in the database and
+    // do an additional LWT on it)
+    List<MigrationQuery> queries =
+        new CassandraMigrator(dataStore, processedSchema.getMappingModel(), migrationStrategy)
+            .compute();
+
+    response.put(
+        "cqlChanges",
+        queries.isEmpty()
+            ? ImmutableList.of("No changes, the CQL schema is up to date")
+            : queries.stream().map(MigrationQuery::getDescription).collect(Collectors.toList()));
+
+    if (!dryRun) {
+      for (MigrationQuery query : queries) {
+        dataStore.execute(query.build(dataStore)).get();
+      }
+      SchemaSource newSource =
+          new SchemaSourceDao(dataStore).insert(namespace, expectedVersion, input);
+      response.put("version", newSource.getVersion());
+      // TODO update SchemaFirstCache from here instead of letting it reload from the DB
+    }
+
+    return response.build();
   }
 
   protected abstract String getSchemaContents(DataFetchingEnvironment environment)
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraMigrator.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraMigrator.java
@@ -1,0 +1,213 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.graphql.schema.schemafirst.migration;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
+import graphql.GraphqlErrorException;
+import io.stargate.db.datastore.DataStore;
+import io.stargate.db.schema.Column;
+import io.stargate.db.schema.Schema;
+import io.stargate.db.schema.SchemaEntity;
+import io.stargate.db.schema.Table;
+import io.stargate.db.schema.UserDefinedType;
+import io.stargate.graphql.schema.schemafirst.migration.CassandraSchemaHelper.Difference;
+import io.stargate.graphql.schema.schemafirst.processor.EntityMappingModel;
+import io.stargate.graphql.schema.schemafirst.processor.MappingModel;
+import io.stargate.graphql.schema.schemafirst.util.DirectedGraph;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class CassandraMigrator {
+
+  private final DataStore dataStore;
+  private final MappingModel mappingModel;
+  private final MigrationStrategy strategy;
+
+  public CassandraMigrator(
+      DataStore dataStore, MappingModel mappingModel, MigrationStrategy strategy) {
+    this.dataStore = dataStore;
+    this.mappingModel = mappingModel;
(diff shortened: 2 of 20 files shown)


graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaFetcherBase.java
```diff
@@ -75,25 +79,41 @@
...
+    // do an additional LWT on it)
+    List<MigrationQuery> queries =
+        new CassandraMigrator(dataStore, processedSchema.getMappingModel(), migrationStrategy)
+            .compute();
+
+    response.put(
+        "cqlChanges",
+        queries.isEmpty()
+            ? ImmutableList.of("No changes, the CQL schema is up to date")
+            : queries.stream().map(MigrationQuery::getDescription).collect(Collectors.toList()));
```
tomekl007, line 106: how the value for `cqlChange` wilt be formatted? How does it handle the list of `String`?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaFetcherBase.java
```diff
@@ -75,25 +79,41 @@
...
+
+    if (!dryRun) {
+      queries.forEach(q -> dataStore.execute(q.getQuery()));
```
tomekl007, line 109: should we use map and do something with the outcome of executed query? I.e:
```
List<ResultSet> result = List<queries.map(q -> dataStore.execute(q.getQuery());
validateIfAllApplied(result);
```

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraMigrator.java
```diff
@@ -0,0 +1,174 @@
...
+          throw new AssertionError("Unexpected target " + entity.getTarget());
+      }
+    }
+    if (!errors.isEmpty()) {
+      throw GraphqlErrorException.newErrorException()
+          .message(
+              String.format(
+                  "The schema you provided can't be mapped to the current CQL data model."
+                      + "Consider using a different migration strategy (current: %s). "
+                      + "See details in `extensions.migrationErrors` below.",
```
tomekl007, line 99: I don't see what does it mean below in this context
`See details in `extensions.migrationErrors` below.`
Will those details be printed out? If yes, where?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraMigrator.java
```diff
@@ -0,0 +1,174 @@
...
+          case ADD_MISSING_TABLES:
+            errors.addAll(
+                differences.stream().map(Difference::toString).collect(Collectors.toList()));
+            break;
+          case ADD_MISSING_TABLES_AND_COLUMNS:
+            List<Difference> blockers =
+                differences.stream().filter(d -> !isAddableColumn(d)).collect(Collectors.toList());
+            if (blockers.isEmpty()) {
+              for (Difference difference : differences) {
+                assert isAddableColumn(difference);
```
tomekl007, line 150: why assert? Should we check and throw an exception if it does not match?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraMigrator.java
```diff
@@ -0,0 +1,174 @@
...
+                queries.add(addColumnBuilder.build(expected, difference.getColumn(), dataStore));
+              }
+            } else {
+              errors.addAll(
+                  blockers.stream().map(Difference::toString).collect(Collectors.toList()));
+            }
+            break;
+          case DROP_AND_RECREATE_IF_MISMATCH:
```
tomekl007, line 158: Can we refactor it to simplify Ifs and switch? Right now it is hard to read because of enclosed conditionals.
Also, how this `DROP_AND_RECREATE_IF_MISMATCH` is different from `DROP_AND_RECREATE_ALL`? I see that they execute the same logic. Can we merge those together?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraSchemaHelper.java
```diff
@@ -0,0 +1,222 @@
...
+              String.format("expected %s, found %s", expectedColumn.kind(), actualColumn.kind())));
+    } else if (expectedColumn.kind() == Column.Kind.Clustering
+        && expectedColumn.order() != actualColumn.order()) {
+      differences.add(
+          new Difference(
+              expectedColumn,
+              DifferenceType.WRONG_CLUSTERING_ORDER,
+              String.format(
+                  "expected %s, found %s", expectedColumn.order(), actualColumn.order())));
+    }
```
tomekl007, line 85: is this logic handling static, non-static?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraSchemaHelper.java
```diff
@@ -0,0 +1,222 @@
...
+
+    public DifferenceType getType() {
+      return type;
+    }
+
+    @Override
+    public String toString() {
+
+      return String.format(
```
tomekl007, line 206: This `toString` seems important because we are using it in the logic. Should we instead create a custom method that returns the representation of this class? (I don't strongly feel about it)

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/MigrationQuery.java
```diff
@@ -0,0 +1,56 @@
...
+package io.stargate.graphql.schema.schemafirst.migration;
+
+import io.stargate.db.query.builder.AbstractBound;
+
+/** A DDL query to be executed as part of a migration. */
+public class MigrationQuery implements Comparable<MigrationQuery> {
+
+  private final AbstractBound<?> query;
+  private final String description;
+  private final boolean isAddColumn;
```
tomekl007, line 25: we don't allow removing columns from the schema, right?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaFetcherBase.java
olim7t: `MigrationQuery::getDescription` returns a string.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaFetcherBase.java
olim7t: All the CQL queries have to succeed, otherwise the deployment fails with a GraphQL error. So if you reach the point where you are reading `cqlChanges` from a `deploySchema` response, the only possible outcome is success.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraMigrator.java
olim7t: This should be clearer on an example: create an empty keyspace named `accounts`, then go to the GraphQL playground (`http://$STARGATE_HOST:8080/playground`), enter `http://$STARGATE_HOST:8080/graphqlv2/admin` in the URL bar and execute the following mutation:
```graphql
mutation {
  deploySchema(
    namespace: "accounts"
    migrationStrategy: USE_EXISTING
    schema: """
    type User @key(fields: "id") @cql_input {
      id: ID!
      name: String
      username: String
    }
    type Query {
      getUser(id: ID!): User
    }
    """
  ) {
    version
  }
}
```
Notice the `USE_EXISTING` strategy, which means this will fail because the table does not exist yet.

Here's what the error response will look like:
```graphql
{
  "errors": [
    {
      "message": "Exception while fetching data (/deploySchema) : The schema you provided can't be mapped to the current CQL data model.Consider using a different migration strategy (current: USE_EXISTING). See details in `extensions.migrationErrors` below.",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "deploySchema"
      ],
      "extensions": {
        "migrationErrors": [
          "Missing table User"
        ],
        "classification": "DataFetchingException"
      }
    }
  ],
  "data": {
    "deploySchema": null
  }
}
```

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraSchemaHelper.java
olim7t: Yes, static is a `Kind`.

By the way we'll be able to reuse this logic in `SchemaSourceDao`.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraSchemaHelper.java
olim7t: Good point, will do.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraMigrator.java
olim7t: Because I know at this point that this is true, otherwise `blockers` would not be empty.

The only way this could fail is a programming error. This is more about documenting it, and protecting against future refactorings (e.g. if we change the filter condition in `blockers` but forget to revisit this). I feel that's a good fit for assertions.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/MigrationQuery.java
olim7t: No. We could, but so far I've avoided it. My rationale is that someone might reuse an existing schema that's shared with other non-GraphQL applications, and not map every column. It would be a shame if GraphQL accidentally dropped a bunch of columns that were actually needed by something else.

With that being said, the `DROP_AND_RECREATE*` strategies _will_ drop columns in such a scenario, but I think their names are a clear enough indication that they are destructive.

The strategies are not final yet, we can revisit them. I think there might always be corner cases that are better handled with manual CQL queries. If it's just cleaning up obsolete columns that should be fine.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraMigrator.java
olim7t: Yeah the logic is a bit convoluted, I'll see how I can improve it.

`DROP_AND_RECREATE_ALL` drops and recreates all tables no matter what.

`DROP_AND_RECREATE_IF_MISMATCH` only recreates the tables that have changed. For example, let's say you have an existing schema with a `User` type and some associated data. You're working on a v2 that adds an `Address` table. This strategy allows you to iterate on `Address` with drop-and-recreate (which is the most convenient approach for completely new types), without losing your existing data for `User`.

olim7t: I've changed to a single switch statement, it should be easier to follow now.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraMigrator.java
```diff
@@ -0,0 +1,193 @@
...
+    this.mappingModel = mappingModel;
+    this.strategy = strategy;
+  }
+
+  /** @throws GraphqlErrorException if the CQL data model can't be migrated */
+  public List<MigrationQuery> compute() {
+
+    Schema schema = dataStore.schema();
+
+    List<MigrationQuery> queries = new ArrayList<>();
```
olim7t, line 51: I've temporarily disabled the sorting because my comparator was broken. I need to revisit this with the topological sort anyway.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaFetcherBase.java
```diff
@@ -75,25 +79,43 @@
...
+
+    response.put(
+        "cqlChanges",
+        queries.isEmpty()
+            ? ImmutableList.of("No changes, the CQL schema is up to date")
+            : queries.stream().map(MigrationQuery::getDescription).collect(Collectors.toList()));
+
+    if (!dryRun) {
+      for (MigrationQuery query : queries) {
+        dataStore.execute(query.build(dataStore)).get();
```
olim7t, line 110: Queries are now built at the last moment. This avoids the issue I was mentioning on #664.

olim7t:
Merging this, it's been a week and I have other branches that depend on it. The target branch is still a PR so if there are follow-ups we can address them there.

PR #659, opened 2021-02-10
Revisit processing messages

olim7t:
- always mention the element name (even though we include a location,
  having the name is convenient).
- emit INFO messages for some default behaviors.

### DIFF
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/processor/EntityMappingModel.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/processor/EntityMappingModel.java
@@ -154,7 +154,7 @@ public class EntityMappingModel {
     List<FieldMappingModel> regularColumns = new ArrayList<>();
 
     for (FieldDefinition fieldDefinition : type.getFieldDefinitions()) {
-      FieldMappingModel.build(fieldDefinition, context, target)
+      FieldMappingModel.build(fieldDefinition, context, graphqlName, target)
           .ifPresent(
               fieldMapping -> {
                 if (fieldMapping.isPartitionKey()) {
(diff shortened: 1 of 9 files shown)


graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/processor/EntityMappingModel.java
```diff
@@ -173,14 +173,21 @@ public boolean isFederated() {
         if (partitionKey.isEmpty()) {
           FieldMappingModel firstField = regularColumns.get(0);
           if (TypeHelper.isGraphqlId(firstField.getGraphqlType())) {
+            context.addInfo(
```
tomekl007, line 176: When this is executed? Is it only once during the init phase? Or for every request (same question for other `info`)

olim7t: `MappingModel` is rebuilt every time we need to refresh the `GraphQL` instance (`SchemaProcessor` is the high-level entry point to do that).

This happens:
- whenever a `deploySchema` mutation is executed (`DeploySchemaFetcherBase`)
- whenever we find out that a new row was inserted in the `graphql_schema` table (`SchemaFirstCache`). We check for every request, but we only rebuild if the version has changed.

tomekl007: Ok, I wonder how often in practice this will happen. Won't the `log.info` span the end-user too much?

tomekl007 (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/processor/EntityMappingModel.java
olim7t: I plan to add a filter to `DeploySchemaResponse.messages`. That would allow you to select warning messages only.

dougwettlaufer (approved):

PR #661, opened 2021-02-10
Add a 'version' argument to the 'schema' admin query

tomekl007:


graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/AdminSchemaBuilder.java
```diff
@@ -242,6 +242,12 @@ public GraphQLSchema build() {
                   .description("The name of the namespace")
                   .type(nonNull(GraphQLString))
                   .build())
+          .argument(
+              newArgument()
+                  .name("version")
+                  .description(
+                      "The version of the schema to get. If not specified, it will return the latest version.")
+                  .type(GraphQLString))
```
tomekl007, line 250: is it a proper way for defining an optional argument? (not using `nonNull`)?
Also, I wonder if the `GraphQLString` is the correct type. From what I see, there is no UUID scalar in grapqh-java. However, there is an open PR for it, see https://github.com/graphql-java/graphql-java-extended-scalars/pull/18/files. Should we wait for it? Then, we would be able to do:
```
public static GraphQLScalarType UUID = new UUIDScalar();
```

olim7t: Yes, that is the proper way to make it optional.

I think `String` is good enough: the version is an opaque value as far as users are concerned, they only read it and provide it back when they deploy a new one. No user should ever need to parse it as a UUID.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -70,7 +70,10 @@ public SchemaSourceDao(DataStore dataStore) {
     return row.stream().map(r -> toSchemaSource(namespace, r)).collect(Collectors.toList());
   }
 
-  public SchemaSource getLatest(String namespace) throws Exception {
+  public SchemaSource getLatest(
```
olim7t, line 73: Since we provide the version this won't necessary be the latest. I think we can just call this method `get`.

olim7t (approved):
👍  LGTM pending minor naming remark

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
tomekl007: I've changed to `getByVersion`

dougwettlaufer (approved):

PR #665, opened 2021-02-10
Expose metrics mbeans

dougwettlaufer:


persistence-cassandra-3.11/src/main/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
```diff
@@ -20,51 +25,75 @@
...
+  private final MBeanWrapper mBeanServer = MBeanWrapper.instance;
+
   private CassandraMetricsRegistry() {
     super();
   }
 
   public Counter counter(MetricName name) {
-    return actualRegistry.counter(name.getMetricName());
+    Counter counter = actualRegistry.counter(name.getMetricName());
+    registerMBean(counter, name.getMBeanName());
```
dimas-b, line 40: Since this class replaces the class with the same name coming from C* jars, the risk remains that future C* changes introduce behaviours that we do not support in Stargate.

What do you think about adding a check for the exact C* version so that when we upgrade our C* dependency it would will the build and remind us to review this file?

I'm just wondering what is our diff with respect to C* sources for this class? Could we apply the change that we need during the build (automatically)?

Also, we should do a similar fix for C* 4.0.

dougwettlaufer: After removing formatting differences this is the diff

```diff
--- theirs.java	2021-02-10 16:02:29.000000000 -0600
+++ ours.java	2021-02-10 16:02:41.000000000 -0600
@@ -1,30 +1,18 @@
 package org.apache.cassandra.metrics;
 
+import com.codahale.metrics.Counter;
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.Histogram;
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.Metered;
+import com.codahale.metrics.Metric;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Timer;
 import java.lang.reflect.Method;
 import java.util.Locale;
 import java.util.concurrent.TimeUnit;
-
-import com.codahale.metrics.*;
 import javax.management.MalformedObjectNameException;
 import javax.management.ObjectName;
-
 import org.apache.cassandra.utils.MBeanWrapper;
 
 /**
@@ -28,231 +16,194 @@
 import org.apache.cassandra.utils.MBeanWrapper;
 

-public class CassandraMetricsRegistry extends MetricRegistry
-{
+public class CassandraMetricsRegistry extends MetricRegistry {
+
+  public static volatile MetricRegistry actualRegistry = new MetricRegistry();
     public static final CassandraMetricsRegistry Metrics = new CassandraMetricsRegistry();
 
     private final MBeanWrapper mBeanServer = MBeanWrapper.instance;
 
 
-    public Counter counter(MetricName name)
-    {
-        Counter counter = counter(name.getMetricName());
+  public Counter counter(MetricName name) {
+    Counter counter = actualRegistry.counter(name.getMetricName());
         registerMBean(counter, name.getMBeanName());
 
         return counter;
     }
  
-    public Meter meter(MetricName name)
-    {
-        Meter meter = meter(name.getMetricName());
+  public Meter meter(MetricName name) {
+    Meter meter = actualRegistry.meter(name.getMetricName());
         registerMBean(meter, name.getMBeanName());
 
         return meter;
     }
  
-    public Histogram histogram(MetricName name, boolean considerZeroes)
-    {
-        Histogram histogram = register(name, new ClearableHistogram(new DecayingEstimatedHistogramReservoir(considerZeroes)));
+  public Histogram histogram(MetricName name, boolean considerZeroes) {
+    Histogram histogram =
+        register(
+            name, new ClearableHistogram(new DecayingEstimatedHistogramReservoir(considerZeroes)));
         registerMBean(histogram, name.getMBeanName());
 
         return histogram;
     }
  
-    public <T extends Metric> T register(MetricName name, T metric)
-    {
-        try
-        {
-            register(name.getMetricName(), metric);
+  public <T extends Metric> T register(MetricName name, T metric) {
+    try {
+      T register = actualRegistry.register(name.getMetricName(), metric);
             registerMBean(metric, name.getMBeanName());
-            return metric;
-        }
-        catch (IllegalArgumentException e)
-        {
-            Metric existing = Metrics.getMetrics().get(name.getMetricName());
-            return (T)existing;
+      return register;
+    } catch (IllegalArgumentException e) {
+      @SuppressWarnings("unchecked")
+      T existing = (T) actualRegistry.getMetrics().get(name.getMetricName());
+      return existing;
         }
     }
 
-    public boolean remove(MetricName name)
-    {
-        boolean removed = remove(name.getMetricName());
+  public boolean remove(MetricName name) {
+    boolean removed = actualRegistry.remove(name.getMetricName());
 
-        try
-        {
+    try {
             mBeanServer.unregisterMBean(name.getMBeanName());
-        } catch (Exception ignore) {}
+    } catch (Exception ignore) {
+    }
 
         return removed;
     }
 
-    private void registerAlias(MetricName existingName, MetricName aliasName)
-    {
-        Metric existing = Metrics.getMetrics().get(existingName.getMetricName());
+  private void registerAlias(MetricName existingName, MetricName aliasName) {
+    Metric existing = actualRegistry.getMetrics().get(existingName.getMetricName());
         assert existing != null : existingName + " not registered";
 
         registerMBean(existing, aliasName.getMBeanName());
     }
```

dougwettlaufer: Regarding the check, do you mean something like this?

```diff
diff --git a/persistence-cassandra-3.11/src/main/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java b/persistence-cassandra-3.11/src/main/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
index e1130299..9238d90a 100644
--- a/persistence-cassandra-3.11/src/main/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
+++ b/persistence-cassandra-3.11/src/main/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
@@ -13,6 +13,7 @@ import java.util.Locale;
 import java.util.concurrent.TimeUnit;
 import javax.management.MalformedObjectNameException;
 import javax.management.ObjectName;
+import org.apache.cassandra.utils.FBUtilities;
 import org.apache.cassandra.utils.MBeanWrapper;

 /**
@@ -31,6 +32,17 @@ public class CassandraMetricsRegistry extends MetricRegistry {

   private final MBeanWrapper mBeanServer = MBeanWrapper.instance;

+  static {
+    String supportedVersion = "3.11.6";
+    String currentVersion = FBUtilities.getReleaseVersionString();
+    if (!supportedVersion.equals(currentVersion)) {
+      throw new IllegalStateException(
+          String.format(
+              "Cassandra version [%s] does not match supported version [%s]",
+              currentVersion, supportedVersion));
+    }
+  }
+
   private CassandraMetricsRegistry() {
     super();
   }
```

dimas-b: I thought we'd do it in maven, but this works too :)

dimas-b: I wonder if we could replace the `ConcurrentMap<String, Metric> metrics` inside the C* `CassandraMetricsRegistry` object and intercept metrics this way... java reflection should be able to overcome the `final` declaration of that field, but this approach still may be messy in the end. Just wondering what you think about it in principle. If it works then we won't have to modify C* code.

persistence-cassandra-3.11/src/main/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
```diff
@@ -20,51 +25,75 @@
  * <p>This class must mimic the exact public API of the original one.
  */
 public class CassandraMetricsRegistry extends MetricRegistry {
```
tomekl007, line 27: This class is copied from the original Cassandra source code, right?
Is there also a test class for it? (`CassandraMetricsRegistryTest`). If yes, then I think we should port it to the Stargate database, (and adapt to the new JMX changes)

dimas-b: `CassandraMetricsRegistryTest` indeed exists, but at first glace it does not look like it covers all the registration paths, just some of them. Plus if the code evolves on the C* side and the test is updated we are still not going to get the updates automatically, so the risk of behaviour divergence remains.

tomekl007: Yes, It does not solve the problem of `behaviour divergence`, only lack of testing

dimas-b:
I think we need to do the same for C* 4.0 persistence.

dimas-b:
LGTM overall. Just wondering if we are going to implement version change checks (discussed in other comments) so that we get to review possible C* OSS changes in `CassandraMetricsRegistry` and incorporate them into Stargate?

dougwettlaufer:
> I think we need to do the same for C* 4.0 persistence.

Yes. Just wanted to get this out sooner to address the immediate issue.

dougwettlaufer:
> LGTM overall. Just wondering if we are going to implement version change checks (discussed in other comments) so that we get to review possible C* OSS changes in `CassandraMetricsRegistry` and incorporate them into Stargate?

Version check added to the test here
https://github.com/stargate/stargate/pull/665/files#diff-524a1fb2e40dd851849ecd32aa82dc8dec95ecf951d394b35a31512cb5f71845R52-R60

dimas-b:
Ah, yes - sorry, I missed that check in the test code :facepalm:

dimas-b (approved):

dougwettlaufer:
FYI @marksurnin

PR #667, opened 2021-02-11
Add a REST endpoint to download the definition of the Stargate CQL directives

tomekl007:


dimas-b (approved):

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
```diff
@@ -0,0 +1,45 @@
...
+@Path(ResourcePaths.FILES)
+@Produces(MediaType.APPLICATION_JSON)
+public class FilesResource extends GraphqlResourceBase {
+
+  @GET
+  @Path("/cql_directives.graphql")
+  @Produces(MediaType.APPLICATION_OCTET_STREAM)
+  public Response getCqlDirectives() {
+    return Response.ok(loadCqlDirectivesFile()).build();
+  }
```
olim7t, line 38: Since the file is text, I think the best experience would be to open it in the browser and let the user "save as..." as they wish:

```suggestion
  @Produces(MediaType.TEXT_PLAIN)
  public Response getCqlDirectives() {
    return Response.ok(loadCqlDirectivesFile())
        .header("Content-Disposition", "inline; filename=\"cql_directives.graphql\"")
        .build();
  }
```
This works well in Firefox. Unfortunately both Chrome and Safari seem to ignore the `Content-Disposition` header and add an extension after `.graphql` when you save. It would be nice to fix that, but if we don't find a way I think it's minor.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
```diff
@@ -0,0 +1,45 @@
...
+import javax.inject.Singleton;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+@Singleton
+@Path(ResourcePaths.FILES)
+@Produces(MediaType.APPLICATION_JSON)
```
olim7t, line 30: Not needed.
Or you can declare it once here and omit it on the methods, I think all the ones we plan to add will have the same type.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
tomekl007: when I change the content type to `application/graphql` it does not add `txt` to the attachment part, but unable for the user to open it in a browser and save-as. It just downloads the file right away. Should I leave it as it is?

tomekl007:
@olim7t I've merged it because it blocks my work on `Expose a REST endpoint to download schemas`, I can address any new comment afterward.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
olim7t: Yes, I believe `application/graphql` is for actual GraphQL queries.

I think we should go with `TEXT_PLAIN`, the extension issue is a minor thing and opening in the browser first is better for the user: it lets them choose where to save, or even if they want to save at all (they might just need to take a look at the contents). As a user if it downloads immediately I would find it annoying.

tomekl007: done

PR #668, opened 2021-02-11
fix some codacy warnings

tomekl007:
@olim7t This is fixing the majority of Codancy errors on the `graphql-first` branch.
There are 3 problems left pending, please take a look:
- [ ] `Avoid really long methods.
` on `static Optional<EntityMappingModel> build(ObjectTypeDefinition type, ProcessingContext context) {
`. Please take a look and split if possible
- [x] `Avoid really long methods.` on `protected static void postMultipartJson`
- [x] `Avoid unused private methods such as 'parseSchema(String)'.` on
`private static TypeDefinitionRegistry parseSchema(String inputText) throws GraphqlErrorException`. This method is not used anywhere. Is it safe to remove?

### DIFF
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/AdminSchemaBuilder.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/AdminSchemaBuilder.java
@@ -50,61 +50,6 @@ import java.io.InputStream;
 
 public class AdminSchemaBuilder {
 
-  private final AuthenticationService authenticationService;
-  private final AuthorizationService authorizationService;
-  private final DataStoreFactory dataStoreFactory;
-
-  public AdminSchemaBuilder(
-      AuthenticationService authenticationService,
-      AuthorizationService authorizationService,
-      DataStoreFactory dataStoreFactory) {
-    this.authenticationService = authenticationService;
-    this.authorizationService = authorizationService;
-    this.dataStoreFactory = dataStoreFactory;
-  }
-
-  public GraphQLSchema build() {
-    return newSchema()
-        .query(QUERY)
-        .mutation(MUTATION)
-        .codeRegistry(
-            newCodeRegistry()
-                .dataFetcher(
-                    coordinates(QUERY, NAMESPACES_QUERY),
-                    new AllNamespacesFetcher(
-                        authenticationService, authorizationService, dataStoreFactory))
-                .dataFetcher(
-                    coordinates(QUERY, NAMESPACE_QUERY),
-                    new SingleNamespaceFetcher(
-                        authenticationService, authorizationService, dataStoreFactory))
-                .dataFetcher(
-                    coordinates(QUERY, SCHEMA_QUERY),
-                    new SingleSchemaFetcher(
-                        authenticationService, authorizationService, dataStoreFactory))
-                .dataFetcher(
-                    coordinates(QUERY, SCHEMA_HISTORY_PER_NAMESPACE_QUERY),
-                    new AllSchemasFetcher(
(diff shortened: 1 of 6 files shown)


dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaFetcherBase.java
```diff
@@ -122,6 +122,7 @@ private UUID getExpectedVersion(DataFetchingEnvironment environment) {
             .collect(Collectors.toList()));
   }
 
+  // todo why it is not used?
   private static TypeDefinitionRegistry parseSchema(String inputText) throws GraphqlErrorException {
```
olim7t, line 126: Yes, this is a leftover, I'll remove it.

mpenick (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/AdminSchemaBuilder.java
```diff
@@ -50,61 +50,6 @@
...
-                    new DeploySchemaFetcher(
-                        authenticationService, authorizationService, dataStoreFactory))
-                .dataFetcher(
-                    coordinates(MUTATION, DEPLOY_SCHEMA_FILE_MUTATION),
-                    new DeploySchemaFileFetcher(
-                        authenticationService, authorizationService, dataStoreFactory))
-                .build())
-        .build();
-  }
-
```
olim7t, line 107: This is going to be subjective, but I don't think constants should necessarily come first.

When I open a source file I'm not familiar with, what I need is a quick idea of what's going on. Comparing the two versions here:
* if the first thing I see is the `build()` method, it tells me everything I need to know: that what the class builds is a `GraphQLSchema`, and its overall structure. If I'm interested in a particular part, I can use the "jump to definition" feature in my IDE to progressively zoom into it.
* if the first thing I see is a wall of constants, it doesn't tell me how they are being used. I have to sift through some ~400 lines before I get to the meat of the class.

It's not that important so if we want to stick to the convention so be it. But IMHO readability is more important than making Codacy happy.

Same for other constant-related changes in this PR, but they're less significant because they have only a few constants.

olim7t:
I've fixed two of the remaining warnings.
Let's leave `EntityMappingModel` as-is for now, it's not high priority and the class will probably change a lot in future PRs, we can either try to extract methods as we go, or do it at the end when it's more stable.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/AdminSchemaBuilder.java
tomekl007: I also think that we should favor readability over making a tool happy. However, without those changes, we will always get a red build now. We could make exclusions, but I am not sure if we should do it (and how to do it) because the project is OS, and maybe the convention is more important. I can rollback or merge as it is up to you.

dougwettlaufer: I prefer we stick to the conventions. Readability will be higher if all the source code looks the same and you always know where to find something (ie constants at the top). Personally the original version of this file feels really weird to me since this format of defining things at the top of the file is very well established.

olim7t: OK, as I said it's not that important. Let's go with the constants first.

olim7t (approved):

PR #674, opened 2021-02-12
Check that env var DEVELOPER_MODE is set and 'true' before enabling developer mode

### LINKED ISSUE #673: DEVELOPER_MODE starctl option is enabled when value other than 'true' is used
When running stargate via k8 I added the following environment variable into the deployment descriptor:

* env:
  * name: DEVELOPER_MODE
    value: "false"

This still enabled developer mode when it shouldn't according to the documentation. The only way to disable developer mode was to remove the DEVELOPER_MODE environment variable from the deployment.

dougwettlaufer:
resolves #673

### DIFF
--- a/starctl
+++ b/starctl
@@ -132,7 +132,7 @@ if [[ ${#STARGATE_ARGS[@]} -eq 0 ]]; then
         STARGATE_ARGS+=("--proxy-port" "$PROXY_PORT")
     fi
 
-    if [[ ! -z "$DEVELOPER_MODE" ]] && [ "$DEVELOPER_MODE" = true ]; then
+    if [[ "$DEVELOPER_MODE" = true ]] || [[ "$DEVELOPER_MODE" = 1 ]]; then
         STARGATE_ARGS+=("--developer-mode")
     fi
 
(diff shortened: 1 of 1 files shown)


starctl
```diff
@@ -132,7 +132,7 @@ if [[ ${#STARGATE_ARGS[@]} -eq 0 ]]; then
         STARGATE_ARGS+=("--proxy-port" "$PROXY_PORT")
     fi
 
-    if [[ ! -z "$DEVELOPER_MODE" ]]; then
+    if [[ ! -z "$DEVELOPER_MODE" ]] && [ "$DEVELOPER_MODE" = true ]; then
```
mpenick, line 135: `DEVELOPER_MODE=1` should probably work too.

dimas-b: Agree

dimas-b: I think `[[ ! -z "$DEVELOPER_MODE" ]]` is not needed now that we have  `&&` and a test for a non-empty value. 

Also the second check should probably use double brackets for consistency with other `if` statements.

dougwettlaufer: Double bracket and single bracket tests in bash always trip me up.

dimas-b (approved):

Issue #270, opened 2020-10-14
Consider simpler alternative to Apollo library for GraphQL ITs


olim7t:
Adding new GraphQL integration tests involves a lot of ceremony:
* update the relevant `schema.json`. The simplest way to do that is to start a standalone Stargate instance and backend, create the schema manually, and run an introspection query.
* create new `*.graphql` files to add queries.
* run the build to generate the Java code from those files.
* the Apollo query API is convoluted: execute the query to get an "observable", create a future, enqueue the future on the observable, wait for the future to complete, then cancel the observable.

(see also comments in `GraphQlTest`)

I wouldn't mind writing my queries directly into Java strings (as it's done in unit tests such as `QueryFetcherTest`). Then we should be able to use any HTTP library to wrap that into a REST query. I'm less sure about inspecting the results, but maybe we can find an easy way to parse the response into maps of maps.

dougwettlaufer:
Agreed, it's not ideal but I went with the Apollo client for a few reasons

1. It seems to be the de facto client for graphql in java
2. Wanted to provide examples of queries and how to work with graphqlapi in a more realistic manner
3. Graphql over plain http is kind of ugly, it looks like json but its not plus you have to figure out how to serialize results into objects for assertions
4. Similar to #2 wanted to also validate that our graphql api actually works with a real client.

Not against making things easier but just some thoughts.

olim7t:
Thanks for the explanations, makes sense.

Maybe we can have one test with Apollo, and the rest with a more lightweight alternative.

olim7t:
I used the "lightweight" approach in `GraphqlJWTAuthTest`: GraphQL queries are passed as plain strings, and the test uses `OkHttpClient` to wrap them into a REST request. This can be used as a template for future tests.

PR #675, opened 2021-02-12
Expose a REST endpoint to download schemas

tomekl007:
WIP

TODO:
- [x] revisit media type (per https://github.com/stargate/stargate/pull/667#discussion_r574805538)


graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/NamespaceResource.java
```diff
@@ -59,7 +51,7 @@ public void get(
       @Context HttpServletRequest httpRequest,
       @Suspended AsyncResponse asyncResponse) {
 
-    GraphQL graphql = getGraphl(namespace, httpRequest, asyncResponse);
+    GraphQL graphql = getGraphql(namespace, httpRequest, asyncResponse, graphqlCache);
```
tomekl007, line 54: It is not clear to me how it should work (and how it works now).
1. Is the `NamespaceResource` serving the schema based on version? If yes, why it needs to accept the query, variables, and other query parameters? Shouldn't it just return the schema for a given version?
A similar pattern is for `AdminResource` get method. Shouldn't we just return the `SchemaSource` from the new endpoint?

testing/src/test/java/io/stargate/it/http/graphfirst/AdminResourcesTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @BeforeAll
+  public static void crateTestKeyspace() {
+    KEYSPACE = UUID.randomUUID().toString();
+    session.execute(
+        "CREATE KEYSPACE IF NOT EXISTS ks_1 WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1 };");
+  }
+
+  @Test
+  public void createTable() throws ExecutionException, InterruptedException {
```
tomekl007, line 29: I am trying to write an IT test for admin resources to understand how schema retrieval works. I am trying to base it on the exiting cql-first `GraphqlTest`. 
From what I see, this test is pre-creating all schema via cql query from the `betterbotz.cql`. 
Regarding the new `graph-first` API my impression is that we should allow end-user to create everything via `graph-ql` queries only. Is that true?
If yes, can the new test use the same types, i.e: ` GetTableQuery.Table` or do we need to define new ones?

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/SchemaFirstCache.java
```diff
@@ -74,12 +76,13 @@ public GraphQL getAdminGraphql() {
...
-    return getGraphql(persistence.decorateKeyspaceName(namespace, headers));
+  public GraphQL getGraphql(String namespace, Map<String, String> headers, Optional<UUID> version)
+      throws Exception {
+    return getGraphql(persistence.decorateKeyspaceName(namespace, headers), version);
   }
 
-  private GraphQL getGraphql(String namespace) throws Exception {
-    SchemaSource latestSource = schemaSourceDao.getLatest(namespace);
+  private GraphQL getGraphql(String namespace, Optional<UUID> version) throws Exception {
+    SchemaSource latestSource = schemaSourceDao.getByVersion(namespace, version);
```
tomekl007, line 85: One solution for exposing the schema to the end-user would be to rework this method to return the `GraphqlHolder`. The caller would be able to access the `SchemaSrouce` but also `GraphQL`. 
However, in the `SchemaSource` java-doc I see that:
```
 * <p>Usually this schema is partial, it will be completed (with default/generated types, directive
 * definitions...) to produce the actual one used at runtime.
```
so I am not sure if this is proper information that should be returned to a client.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/SchemaFirstCache.java
```diff
@@ -74,12 +76,13 @@ public GraphQL getAdminGraphql() {
...
-  public GraphQL getGraphql(String namespace, Map<String, String> headers) throws Exception {
-    return getGraphql(persistence.decorateKeyspaceName(namespace, headers));
+  public GraphQL getGraphql(String namespace, Map<String, String> headers, Optional<UUID> version)
+      throws Exception {
+    return getGraphql(persistence.decorateKeyspaceName(namespace, headers), version);
   }
 
-  private GraphQL getGraphql(String namespace) throws Exception {
-    SchemaSource latestSource = schemaSourceDao.getLatest(namespace);
+  private GraphQL getGraphql(String namespace, Optional<UUID> version) throws Exception {
```
tomekl007, line 84: I am not sure if we should modify this method to get the schema for a specific version or create a new method.
The reason for this concern is that this method uses the `graphqlHolders` map, which keeps only the last holder per namespace.
If customers will often request schema with non-lastest version, this "cache" will have a lot of miss (because elements will be constantly replaced), wdyt?

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/NamespaceResource.java
olim7t: There are two aspects of the schema:
* the "raw" GraphQL string that the user passed to `deploySchema`. This is what we store in the `graphql_schema` table.
* the "runtime" `GraphQL` object that we use to execute queries. We build it by processing the raw schema, adding any generated types, binding the fetchers, etc. (see `SchemaProcessor`).

In this ticket, the goal is to expose the raw schema (so that users can download a previous version of their schema, modify it locally and re-deploy it).

> Shouldn't we just return the `SchemaSource` from the new endpoint?

Correct. `NamespaceResource` handles user queries on an already deployed schema, so it deals with the "runtime" schema. It's not the right place to look, instead the new REST resource should be more similar to `SingleSchemaFetcher`: call `SchemaSourceDao.get()` for the provided namespace and version, and then serve the `contents` field. You'll have to register a DAO instance in the Jersey environment in order to inject it in the resource, see `DropwizardServer`.
`SchemaFirstCache` builds its own DAO instance, so we might want to refactor to reuse a single one instead.

I haven't done a great job with naming. Maybe the `schema*` admin queries should have some notion of "raw" or "source" in the name, I guess that might be confusing for users too. Also I don't have a good name for the "non-admin" GraphQL schema, i.e. what gets deployed to a namespace (your question shows that `NamespaceResource` was a poor choice). I thought of "runtime", but that's not the best description because admin queries execute at runtime too. If you have suggestions I'd like to hear them.

testing/src/test/java/io/stargate/it/http/graphfirst/AdminResourcesTest.java
olim7t: `GraphqlTest` uses the Apollo client, this is good to have an example with a "real-life" GraphQL library, but adding new tests requires a lot of boilerplate. See the discussion on #270.
For our schema-first tests I think it will be faster to use the approach in `GraphqlJWTAuthTest`.

> Regarding the new `graph-first` API my impression is that we should allow end-user to create everything via `graph-ql` queries only. Is that true?

Yes. I think most of the time we won't need to run any CQL queries, the test will start with a `deploySchema` operation, and proceed from there. Except of course to cover specific database states (e.g. migration of an existing schema, `graphql_schema` exists but with the wrong columns, etc).

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/NamespaceResource.java
tomekl007: The `SchemaSourceDao` accepts `DataStore` as an argument. We could remove it from the constructor, and add pass it as an argument to every method that needs it. However, as I see the `SchemaSourceDao` is stateless (besides the usage of the `dataStore`) so I think we can create a new instance of it wherever it's needed. Also, there are some code paths that do not have `DataStore` instance (i.e when `SchemaSourceDao` is called from *Resource),
wdyt?
Regarding naming, maybe:
- `RawSchema` and `ProcessedSchema`?

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
```diff
@@ -38,6 +50,27 @@ public Response getCqlDirectives() {
...
+        schemaSourceDao.getByVersion(namespace, Optional.ofNullable(version).map(UUID::fromString));
+
+    return Response.ok(schemaSource.getContents())
+        .header("Content-Disposition", "inline; filename=" + createFileName(schemaSource))
+        .build();
+  }
+
+  private String createFileName(SchemaSource schemaSource) {
+    return String.format(
+        "\"%s-%s.graphql\"", schemaSource.getNamespace(), schemaSource.getVersion());
```
tomekl007, line 71: I think that it may be UX friendly if we will include both namespace and version in the schema name. Wdyt?

olim7t: Yes, that's what I had in mind 👍

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
```diff
@@ -38,6 +50,27 @@ public Response getCqlDirectives() {
...
 
+  @GET
+  @Path("/namespace/{namespaceName}.graphql")
+  @Produces(MediaType.TEXT_PLAIN)
+  public Response getSchema(
+      @PathParam("namespaceName") String namespace,
+      @QueryParam("version") String version,
+      @Context HttpServletRequest httpRequest)
+      throws Exception {
+    SchemaSource schemaSource =
```
dougwettlaufer, line 61: What's this going to return if the namespace doesn't exist?

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
```diff
@@ -38,6 +50,27 @@ public Response getCqlDirectives() {
         .build();
   }
 
+  @GET
+  @Path("/namespace/{namespaceName}.graphql")
```
dougwettlaufer, line 54: I know I'm late to the party but do we think "namespace" used in graphql will be confused with "namespace" in the docsapi?

testing/src/test/java/io/stargate/it/http/graphfirst/FilesResourceTest.java
```diff
@@ -49,4 +67,85 @@ public void shouldGetGraphQLDirectivesFile() throws IOException {
...
+        .contains((String.format("%s-%s.graphql", keyspaceId.toString(), DEPLOYED_SCHEMA_VERSION)));
+
+    String responseBody = getSchemaResponse.body().string();
+    assertThat(responseBody).isNotEmpty();
+    assertThat(responseBody).contains("type User");
+    assertThat(responseBody).contains("type Query");
+  }
+
+  @Test
+  public void shouldGetFileWithSchemaUsingVersionArgument(@TestKeyspace CqlIdentifier keyspaceId)
```
dougwettlaufer, line 96: Would be good to add a test for namespace that doesn't exist. What happens if I provide a valid keyspace that hasn't been graphqlified like `system`?

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
```diff
@@ -38,6 +50,27 @@ public Response getCqlDirectives() {
...
+  @Produces(MediaType.TEXT_PLAIN)
+  public Response getSchema(
```
dougwettlaufer, line 56: All of the endpoints in here should require the usual auth headers as well.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
olim7t: I picked it because it's essentially the same thing: a thin abstraction over a CQL keyspace. But I don't mind changing it if you think we should use something different.

Btw you're not late, there's still time to improve naming and in fact I planned to do so before we merge the main PR (starting with the URIs).

dougwettlaufer: Cool, I think that makes sense. We probably just need to be cognizant of the double meaning when documenting or discussing. `namespace` is probably a more familiar term for those using graphql as well.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
olim7t: 👍 , this method should do the same check as `SchemaFetcher`.

`getCqlDirectives()` returns static contents that will be the same for every Stargate installation, so I think it  can skip auth.

testing/src/test/java/io/stargate/it/http/graphfirst/FilesResourceTest.java
olim7t: +1 to add a test, as well as for existing namespace but wrong version.

From a quick look at the code it looks like the DAO returns `null` and that gets encoded as the response. I think a better behavior would be to return a 404.

testing/src/test/java/io/stargate/it/http/graphfirst/FilesResourceTest.java
```diff
@@ -49,4 +67,85 @@ public void shouldGetGraphQLDirectivesFile() throws IOException {
     assertThat(responseBody).isNotEmpty();
     assertThat(responseBody).contains("directive @cql_input");
   }
+
+  @Test
+  public void shouldGetFileWithSchema(@TestKeyspace CqlIdentifier keyspaceId) throws IOException {
+    // when get schema file
+    String url = String.format("%s:8080/graphqlv2/files/namespace/%s.graphql", host, keyspaceId);
```
olim7t, line 74: Can you use `ResourcePaths` instead of hard-coding `/graphqlv2/files`?
Paths are likely to change in the near future.

testing/src/test/java/io/stargate/it/http/graphfirst/FilesResourceTest.java
```diff
@@ -17,15 +17,33 @@
...
 import javax.ws.rs.core.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
 
+@ExtendWith(CqlSessionExtension.class)
 public class FilesResourceTest extends GraphqlITBase {
```
olim7t, line 35: Nit: could you name the parent class `GraphqlTestBase` instead?
The existing integration tests all use "Test", I think it's better to stay consistent.

olim7t (changes requested):
The integration tests are getting hard to follow. Suggestions:
* rename `GraphqlTest` to `GraphqlApolloTest` to emphasize that it's the only one using the Apollo library
* follow the production package names
* I think `GraphqlJWTAuthTest` can now also extend `GraphqlITBase`

That would leave us with something like:
```
graphql/
  GraphqlTestBase
  TupleHelper
  cqlfirst/
    GraphqlApolloTest
    GraphqlJWTAuthTest
  schemafirst/
    FilesResourceTest
```

testing/src/test/java/io/stargate/it/http/graphfirst/FilesResourceTest.java
tomekl007: the `testing` module does not have a dependency on the `graphql` module, therefore the `ResourcePaths` is not available.

tomekl007:
> The integration tests are getting hard to follow. Suggestions:
> 
> * rename `GraphqlTest` to `GraphqlApolloTest` to emphasize that it's the only one using the Apollo library
> * follow the production package names
> * I think `GraphqlJWTAuthTest` can now also extend `GraphqlITBase`
> 
> That would leave us with something like:
> 
> ```
> graphql/
>   GraphqlTestBase
>   TupleHelper
>   cqlfirst/
>     GraphqlApolloTest
>     GraphqlJWTAuthTest
>   schemafirst/
>     FilesResourceTest
> ```

Done, regarding:
>  I think `GraphqlJWTAuthTest` can now also extend `GraphqlITBase`
I don't think we should do it. The `GraphqlTestBase` uses the v2 API, whereas the `GraphqlJWTAuthTest` relies on the v1 API. Also, the  `GraphqlJWTAuthTest` uses a different strategy for generating `authTokens` and authorization.
The only common code is the `getHttpClient`. We can use it from this class since it is static in the `GraphqlTestBase`. Also, it is hard for me to refactor the `GraphqlJWTAuthTest` because it is not running on my local, getting:
```
org.testcontainers.containers.ContainerLaunchException: Timed out waiting for URL to be accessible (http://localhost:55007/auth/realms/master should return HTTP 200)
```

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlResourceBase.java
```diff
@@ -338,4 +341,41 @@ protected static void replyWithGraphqlError(
...
+      } else {
+        return graphql;
+      }
+    } catch (Exception e) {
+      LOG.error("Unexpected error while accessing namespace {}", namespace, e);
+      replyWithGraphqlError(
+          Response.Status.NOT_FOUND, "Unexpected error while accessing namespace", asyncResponse);
+      return null;
+    }
+  }
```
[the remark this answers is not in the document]
tomekl007: Do you want a new empty class only for the purpose of being explicit? I've created a new `SchemaFirstResourceBase`

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
tomekl007: Are you sure that we can do the same check as `SchemaFetcher#authorize`?
I see a problem with it: we don't have the `AuthenticationSubject` in the `FilesResource`, but it's needed for the `authorize` check. Should we construct the `ImmutableAuthenticationSubject` somehow? Should the `FilesResource` create the `AuthenticationSubject` from the `X-Cassandra-Token` header?

2nd question: 
I don't see an `authZ` in any other `grapqhl-first` resources: (`NamespaceResource`, `AdminResource`).
Are we missing the `authZ` there as well?

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/NamespaceResource.java
olim7t: > However, as I see the `SchemaSourceDao` is stateless (besides the usage of the `dataStore`) so I think we can create a new instance of it wherever it's needed.

Thew way you did it in `DropwizardServer` looks good to me.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlResourceBase.java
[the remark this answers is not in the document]
olim7t: No, it will contain the namespace-related stuff, which is specific to schema-first.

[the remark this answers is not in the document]
olim7t: Hrmm, in fact I see that `getGraphql()` is only used by `NamespaceResource` now, probably a leftover from your initial prototype. We don't need a parent class anymore, I'll revert the changes.

testing/src/test/java/io/stargate/it/http/graphfirst/FilesResourceTest.java
olim7t: Good point. Maybe we can extract the paths in the parent test class then. They'll still be duplicated, but only in one place.

olim7t:
> The `GraphqlTestBase` uses the v2 API, whereas the `GraphqlJWTAuthTest` relies on the v1 API

In that case `GraphqlApolloTest` should not extend `GraphqlTestBase` either, it also uses the v1 API.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
olim7t: We do authentication and authorization in the REST API, it should be the same here.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
```diff
@@ -38,9 +70,73 @@ public Response getCqlDirectives() {
...
+        isAuthorized(token, namespace, httpRequest)
+            ? schemaSourceDao.getByVersion(
+                namespace, Optional.ofNullable(version).map(UUID::fromString))
+            : null;
+    if (schemaSource == null) {
+      return Response.status(
+              Response.Status.NOT_FOUND.getStatusCode(),
+              String.format(
+                  "The schema for namespace %s and version %s does not exists, "
+                      + "or you're not authorized to read it.",
```
dougwettlaufer, line 93: Not authorized should be a 401

testing/src/test/java/io/stargate/it/http/graphql/graphfirst/FilesResourceTest.java
```diff
@@ -0,0 +1,191 @@
...
+        new Request.Builder()
+            .get()
+            .addHeader("content-type", MediaType.TEXT_PLAIN)
+            .url(url)
+            .build();
+
+    Response getSchemaResponse = getHttpClient("wrong auth token").newCall(getRequest).execute();
+
+    // then
+    assertThat(getSchemaResponse.code()).isEqualTo(404);
```
dougwettlaufer, line 159: Should be a 401

testing/src/test/java/io/stargate/it/http/graphql/graphfirst/FilesResourceTest.java
```diff
@@ -0,0 +1,191 @@
...
+    String responseBody = getSchemaResponse.body().string();
+    assertThat(responseBody).isNotEmpty();
+    assertThat(responseBody).contains("type User");
+    assertThat(responseBody).contains("type Query");
+  }
+
+  @Test
+  public void shouldReturnNotFoundIfSchemaDoesNotExists() throws IOException {
+    // when get schema file
+    String url = String.format("%s:8080/graphqlv2/files/namespace/%s.graphql", host, "unknown");
```
dougwettlaufer, line 127: What about correct namespace and wrong version? Is that a 404 or a 400?

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
```diff
@@ -38,9 +70,73 @@ public Response getCqlDirectives() {
         .build();
   }
 
+  @GET
+  @Path("/namespace/{namespaceName}.graphql")
+  @Produces(MediaType.TEXT_PLAIN)
+  public Response getSchema(
+      @HeaderParam("X-Cassandra-Token") String token,
+      @PathParam("namespaceName") String namespace,
+      @QueryParam("version") String version,
```
dougwettlaufer, line 79: Are versions user defined or an incremental number? Just wondering if this should be an int or if there should be some sort of other validation.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
dougwettlaufer: Is there an issue or TODO somewhere to also make sure this is added to `getCqlDirectives`?

testing/src/test/java/io/stargate/it/http/graphql/graphfirst/FilesResourceTest.java
```diff
@@ -0,0 +1,191 @@
...
+            .build();
+
+    // when
+    Response response = httpClient.newCall(getRequest).execute();
+
+    // then
+    assertThat(response.code()).isEqualTo(200);
+    assertThat(response.body()).isNotNull();
+    String responseBody = response.body().string();
+    assertThat(responseBody).isNotEmpty();
```
dougwettlaufer, line 68: RestUtils does this already
https://github.com/stargate/stargate/blob/master/testing/src/test/java/io/stargate/it/http/RestUtils.java#L45-L64

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
dougwettlaufer: Oh, looks like a UUID? In that case should probably validate it and return an appropriate error message rather than bubbling up the 500.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
olim7t: Do we want to make that distinction though? If we return a 401, we're leaking the fact that the namespace and version exist.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
olim7t: Versions are timeuuids.

testing/src/test/java/io/stargate/it/http/graphql/graphfirst/FilesResourceTest.java
olim7t: It will be a 404 as well. I'll add a test.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
olim7t: > In that case should probably validate it and return an appropriate error message rather than bubbling up the 500

Indeed. I returned a 404 as well because versions are opaque strings as far as users are concerned.

olim7t:
@tomekl007 I refactored the integration tests, I hope you don't mind, I want to merge this PR because I need the changes in #679.
I ended up ditching parent classes, in favor of utility classes like `RestUtils` (but with some state).

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/FilesResource.java
olim7t: `getCqlDirectives()` serves generic content, from a file that is publicly available in this github repo.

olim7t (approved):

PR #676, opened 2021-02-12
STAR-157 parent pom build fix

jtgrabowski:


pom.xml
```diff
@@ -165,6 +165,7 @@
         <module>config-store-yaml</module>
         <module>rate-limiting-global</module>
         <!-- testing last -->
+        <module>testing-services</module>
```
dougwettlaufer, line 168: Good catch!

DEV_GUIDE.md
```diff
@@ -77,7 +77,7 @@ docker pull stargateio/stargate-3_11:v0.0.2
 ```
 
 ```sh
-docker run --name stargate -d stargateio/stargate-3_11:v0.0.2 --cluster-name stargate --cluster-seed 127.0.0.1 --cluster-version 3.11 --listen 127.0.0.2 --simple-snitch
+docker run --name stargate -d stargateio/stargate-3_11:v0.0.2 --cluster-name stargate --cluster-seed 127.0.0.1 --cluster-version 3.11 --listen 127.0.0.2 --bind-to-listen-address --simple-snitch
```
dougwettlaufer, line 80: Were you getting a port collision on 9042 without this?

dougwettlaufer (approved):

mpenick (approved):

pom.xml
dimas-b: yes, thanks for the fix!

DEV_GUIDE.md
jtgrabowski: Yes. Without bind-to-listen-address CQL native transport defaults to 0.0.0.0:9042 which conflicts with 127.0.0.1:9042 (used by C* started with ccm).

dimas-b: Could be caused by the old version of the container, perhaps? Does the latest container also get port conflicts?

jtgrabowski: @dimas-b I've just realized that I made a mistake. The flag was supposed to be added to starctl cmd, not to `docker run`. Sorry for the confusion. Here is a PR that moves the flag to the right place: #705.

PR #681, opened 2021-02-16
Do what filters we can in cassandra

EricBorczuk:
When doing the "full collection" search with a supported cassandra filter, execute it in cassandra rather than defaulting to in-memory. The in-memory approach will still have to be used for multiple filters and/or filters that are not supported by C* (in, nin, neq for instance)

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
@@ -921,23 +921,14 @@ public class DocumentService {
             .collect(Collectors.toList());
 
     Set<String> inCassandraPaths =
-        inCassandraFilters.stream().map(f -> f.getPathString()).collect(Collectors.toSet());
-    if (inCassandraPaths.size() > 1) {
-      // If the request involves more than one filter that could be handled by cassandra,
-      // it would require a CQL `OR` (which doesn't yet exist) - and so we have to default to using
-      // in-memory operations
-      // except for the first path found, since that can be supported by CQL
-      String firstPath = inCassandraPaths.stream().sorted().collect(Collectors.toList()).get(0);
-      List<FilterCondition> cassandraFiltersToUse =
-          inCassandraFilters.stream()
-              .filter(f -> f.getPath().equals(firstPath))
-              .collect(Collectors.toList());
-      List<FilterCondition> theRest =
-          inCassandraFilters.stream()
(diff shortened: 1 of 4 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -911,18 +911,40 @@ void addRowsToMap(Map<String, List<Row>> rowsByDoc, List<Row> rows) {
...
+            .filter(f -> FilterOp.LIMITED_SUPPORT_FILTERS.contains(f.getFilterOp()))
+            .collect(Collectors.toList());
+
+    Set<String> inCassandraPaths =
+        inCassandraFilters.stream().map(f -> f.getFullFieldPath()).collect(Collectors.toSet());
+    if (inCassandraPaths.size() > 1 || inMemoryFilters.size() > 0) {
+      // If the request involves more than one filter that could be handled by cassandra, or if
+      // there is a combination of supported and unsupported filters, then we have to default to the
+      // I/O intensive
+      // in-memory filter.
```
dimas-b, line 929: Why is that? Could you clarify the comment, please?

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -911,18 +911,40 @@ void addRowsToMap(Map<String, List<Row>> rowsByDoc, List<Row> rows) {
...
+      inCassandraFilters = new ArrayList<>();
```
dimas-b, line 930: nit: `Collections.emptyList()`?

dimas-b (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: Yea, I'll change the comment - pretty much it comes down to what CQL is capable of, if you have more than one C* supported filter you'll have to use OR inadvertently (which isn't supported by CQL), and obviously if you use a filter that is not supported in C* you can't use CQL (IN, NOT IN, etc)

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: Yea thanks, I always forget about that :P

dougwettlaufer (approved):

PR #682, opened 2021-02-17
Protect against concurrent deployments

tomekl007:


graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -249,4 +250,63 @@ static boolean hasExpectedSchema(Table table) {
...
+  private void transitionToNoSchemaDeploymentInProgress(String namespace)
+      throws ExecutionException, InterruptedException {
+    BuiltCondition schemaDeploymentInProgressTrue =
+        BuiltCondition.of(DEPLOYMENT_IN_PROGRESS, Predicate.EQ, true);
+
+    BoundQuery updateDeploymentToNotInProgress =
+        dataStore
+            .queryBuilder()
+            .update(namespace, TABLE_NAME)
+            .value(DEPLOYMENT_IN_PROGRESS, null)
```
tomekl007, line 303: When this CAS is executed for the first time, the `DEPLOYMENT_IN_PROGRESS==null`.
Once it is finished, we could update to `false`, but I was unable to create a CAS that checks if `DEPLOYMENT_IN_PROGRESS == null OR false`. Simplifying it to `null/true` works, wdyt?

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -249,4 +250,63 @@ static boolean hasExpectedSchema(Table table) {
...
+            .where(KEY_CONDITION)
+            .ifs(schemaDeploymentInProgressTrue)
+            .build()
+            .bind();
+    dataStore.execute(updateDeploymentToNotInProgress).get();
+    // todo I think it is safe to ignore whether it was applied or not.
```
tomekl007, line 309: I think it is safe to ignore whether it was applied or not.
If it was not applied, it means that the DEPLOYMENT_IN_PROGRESS was already set to false. do you agree?

dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -249,4 +250,63 @@ static boolean hasExpectedSchema(Table table) {
...
+    BuiltCondition schemaDeploymentInProgressNull =
+        BuiltCondition.of(DEPLOYMENT_IN_PROGRESS, Predicate.EQ, null);
+
+    BoundQuery updateDeploymentToInProgress =
+        dataStore
+            .queryBuilder()
+            .update(namespace, TABLE_NAME)
+            .value(DEPLOYMENT_IN_PROGRESS, true)
+            .where(KEY_CONDITION)
+            .ifs(schemaDeploymentInProgressNull)
```
olim7t, line 285: We also need to check that the latest version matches:
```suggestion
            .ifs(schemaDeploymentInProgressNull)
            .ifs(LATEST_VERSION_COLUMN_NAME, Predicate.EQ, expectedLatestVersion)
```
(`expectedLatestVersion` must be added as a new method parameter)

This will cover the scenario:
1. user1 reads the schema at version v2
2. user2 reads the schema at v2, then deploys, which creates v3
3. user1 tries to deploy on top of v2

Even though the boolean is back to false, we must not allow this because user1 would override user2's changes.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -249,4 +250,63 @@ static boolean hasExpectedSchema(Table table) {
...
+  /**
+   * Returns true if there was no schema deployment in progress. It means that that is safe to
+   * deploy a new schema. If it returns false, it means that there is already schema deployment in
+   * progress.
+   */
+  private boolean transitionToSchemaDeploymentInProgress(String namespace)
+      throws ExecutionException, InterruptedException {
+
+    BuiltCondition schemaDeploymentInProgressNull =
+        BuiltCondition.of(DEPLOYMENT_IN_PROGRESS, Predicate.EQ, null);
```
olim7t, line 277: And to answer your question further down: you can use
```java
DEPLOYMENT_IN_PROGRESS, Predicate.NEQ, true
```
to deal with the fact that it will be null the first time.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -249,4 +250,63 @@ static boolean hasExpectedSchema(Table table) {
...
+    BuiltCondition schemaDeploymentInProgressTrue =
+        BuiltCondition.of(DEPLOYMENT_IN_PROGRESS, Predicate.EQ, true);
+
+    BoundQuery updateDeploymentToNotInProgress =
+        dataStore
+            .queryBuilder()
+            .update(namespace, TABLE_NAME)
+            .value(DEPLOYMENT_IN_PROGRESS, null)
+            .where(KEY_CONDITION)
+            .ifs(schemaDeploymentInProgressTrue)
```
olim7t, line 305: We don't need an LWT here: if only one user is allowed to set the flag, then it's guaranteed that no more than one user will try to unset it.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -249,4 +250,63 @@ static boolean hasExpectedSchema(Table table) {
...
+
+  public Map<String, Object> tryDeployNewSchema(
+      String namespace, Callable<Map<String, Object>> newSchemaCreator) throws Exception {
+    ensureTableExists(namespace);
+    if (transitionToSchemaDeploymentInProgress(namespace)) {
+      try {
+        return newSchemaCreator.call();
+      } finally {
+        transitionToNoSchemaDeploymentInProgress(namespace);
+      }
```
olim7t, line 262: Error handling is going to be a bit tricky: if something fails during the schema validation phase, it's safe to put the flag back to `false`. But if one of the migration queries failed, I think we should leave the flag to `true`: the database is in an inconsistent state, and allowing further migrations on top of it is asking for more trouble. At this point it's probably better if an administrator checks the database and resets the flag manually.

So instead of passing a callable here, I think we should expose the primitives (start deployment, finish deployment, abort deployment), and handle the fine details in the calling code. I'm going to add another comment in `DeploySchemaFetcherBase` to explain how I envision it.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaFetcherBase.java
```diff
@@ -77,40 +78,42 @@
...
+        () -> {
+          ImmutableMap.Builder<String, Object> response = ImmutableMap.builder();
+          ProcessedSchema processedSchema =
+              new SchemaProcessor(authenticationService, authorizationService, dataStoreFactory)
+                  .process(input, keyspace);
+          response.put(
+              "messages",
+              processedSchema.getMessages().stream()
+                  .map(this::formatMessage)
+                  .collect(Collectors.toList()));
```
olim7t, line 92: Per my other comment int the DAO, I think we should handle errors here instead of passing a callable. Another reason is that if `dryRun` is set we don't want to write anything to the database:

```java
    List<MigrationQuery> queries;
    SchemaSourceDao schemaSourceDao = new SchemaSourceDao(dataStore);
    if (!dryRun) {
      schemaSourceDao.startDeployment(); // sets the flag with LWT
    }
    try {
      ProcessedSchema processedSchema = ...
      response.put( "messages", ...
      queries = ...
      response.put("cqlChanges", ...
    } catch (Exception e) {
      if (!dryRun) {
        schemaSourceDao.abortDeployment(); // unsets the flag (no need for LWT)
      }
      throw e;
    }
    if (!dryRun) {
      for (MigrationQuery query : queries) { ... }
      // inserts the new version **and** unsets the flag (no need for LWT)
      SchemaSource newSource = schemaSourceDao.insert(namespace, expectedVersion, input);
      response.put("version", newSource.getVersion());
    }
```

Note that I'm also suggesting that `insert` should unset `deployment_in_progress`. It can all be done in one query now, we don't need an LWT or a CQL batch anymore:
```
INSERT INTO graphql_schema(key, version, latest_version, contents, deployment_in_progress)
VALUES ('key', <new_version>, <new_version>, 'the contents...', false)
```

olim7t (changes requested):

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
olim7t: see https://github.com/stargate/stargate/pull/682#discussion_r577917056

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
olim7t: I think we don't need an LWT, see https://github.com/stargate/stargate/pull/682#discussion_r577917845

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaFetcherBase.java
tomekl007: Thank you for this comment, very good ideas!.
I've addressed the feedback. Two additional things that I've added:
1. We need to check if `schemaSourceDao.startDeployment()` was successful, and throw if not:
```
if (!schemaSourceDao.startDeployment(namespace, expectedVersion)) {
        throw new IllegalStateException(
            String.format(
                "It looks like someone else is deploying a new schema for namespace: %s and version: %s. "
                    + "Please retry later.",
                namespace, expectedVersion));
      }
``` 
otherwise, the deployment will progress regardless of the fact if there was `deployment_in_progress` or not
2. We don't need to pass the lastest_schema_version to insert anymore:
```
 SchemaSource newSource = schemaSourceDao.insert(namespace, input);
```
because it inserts a new version for both `version` and `latest_version` columns. 
Do you agree with those changes?

testing/src/test/java/io/stargate/it/http/graphql/graphfirst/SchemaDeploymentTest.java
```diff
@@ -0,0 +1,61 @@
...
+  @DisplayName("Should deploy schema and set the deployment_in_progress column to null")
+  public void deploySchemaAndSetDeploymentInProgressToNull(
+      @TestKeyspace CqlIdentifier keyspaceId, CqlSession session) {
+    // when
+    CLIENT.deploySchema(keyspaceId.asInternal(), SCHEMA_CONTENTS);
+
+    // then
+    Row row = session.execute("select * from graphql_schema").one();
+    assertThat(row).isNotNull();
+    assertThat(row.isNull("deployment_in_progress")).isFalse();
```
tomekl007, line 58: I need this pre-check, because `getBoolean` returns `boolean`. If the `deployment_in_progress == null`, the `getBoolean` returns false

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -134,54 +132,33 @@ BoundQuery schemaQuery(String namespace) {
...
-            String.format(
-                "Could not deploy schema: you indicated your changes applied to version '%s',"
-                    + " but there is a more recent version '%s'.",
-                expectedLatestVersion, actualLatestVersion);
-      }
-      throw new IllegalStateException(message);
+    try {
+      dataStore.execute(insertNewSchema).get();
+    } catch (Exception e) {
+      throw new IllegalStateException(
```
tomekl007, line 156: I am not sure what type of exception we should throw here.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaFetcherBase.java
olim7t: Yes 👍

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
olim7t: GraphQL does not surface the Java type, any runtime exception thrown from a fetcher will be classified as `DataFetchingException`. So it's the message that really matters.

Indeed it's not really an illegal state, but I don't think it's worth creating a custom type either. So maybe simply `RuntimeException`.

olim7t (approved):

PR #685, opened 2021-02-17
Always login users into Persistence Connections

### LINKED ISSUE #683: PersistenceDataStoreFactory does not call login for external users
When a user is authenticated by an external auth provider, CQL connections do call the `Persistence.Connection.login()` method with the corresponding `AuthenticatedUser` object. However, `PersistenceDataStoreFactory` does not. 

This causes different behaviour depending on the API used: CQL or REST / GraphQL.

dimas-b:
Even external users need to go through the Persistence.Connection.login() call.
Persistence implementations are prepared to deal with them. Moreover, external
users are already logged in when used in CQL connections.

Note: this commit makes considerable changes to the DataStoreFactory API,
however the removed methods were not used and conceptually should not be
used because DataStore clients should obtain user information from the
authentication services.

Also, it is worth noting that the removed methods had nullable user name
parameters, but actually using null user names would lead to runtime
exceptions because an AuthenticatedUser (created further down the call
path) cannot have a null name.

Fixes #683

### DIFF
--- a/auth-table-based-service/src/test/java/io/stargate/auth/AuthenticationSubjectTest.java
+++ b/auth-table-based-service/src/test/java/io/stargate/auth/AuthenticationSubjectTest.java
@@ -93,4 +93,14 @@ class AuthenticationSubjectTest {
     assertThat(subject.isFromExternalAuth()).isEqualTo(true);
     assertThat(subject.customProperties()).isEmpty();
   }
+
+  @Test
+  public void convertToAuthenticatedUser() {
+    AuthenticatedUser user =
+        AuthenticationSubject.of("token1", "user2", true, ImmutableMap.of("p1", "v1")).asUser();
+    assertThat(user.token()).isEqualTo("token1");
+    assertThat(user.name()).isEqualTo("user2");
+    assertThat(user.isFromExternalAuth()).isEqualTo(true);
+    assertThat(user.customProperties()).isEqualTo(ImmutableMap.of("p1", "v1"));
+  }
 }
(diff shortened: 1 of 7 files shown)


auth-table-based-service/src/test/java/io/stargate/auth/AuthenticationSubjectTest.java
```diff
@@ -93,4 +93,14 @@ public void fullInputsExternalWithNoCustomProperties() {
     assertThat(subject.isFromExternalAuth()).isEqualTo(true);
     assertThat(subject.customProperties()).isEmpty();
   }
+
+  @Test
+  public void asUser() {
```
tomekl007, line 98: can we change the name of this test to be more meaningful?

persistence-api/src/main/java/io/stargate/db/datastore/PersistenceDataStoreFactory.java
```diff
@@ -69,8 +55,15 @@ private DataStore create(
...
     }
 
-    if (!isFromExternalAuth) {
-      connection.login(AuthenticatedUser.of(userName));
+    connection.login(user);
+
+    // Note: the user's custom properties (if provided) override any matching properties previously
+    // set in DataStoreOptions. This is intentional to give the AuthenticatedUser data more
+    // authority.
+    Map<String, String> customProperties = user.customProperties();
```
tomekl007, line 63: Why do we need this override behavior?

dougwettlaufer (approved):

persistence-api/src/main/java/io/stargate/db/datastore/PersistenceDataStoreFactory.java
dimas-b: As explained in the comment - so that any properties provided by auth services through the `AuthenticationSubject` and `AuthenticatedUser` objects always take precedence over other values under the same key.

auth-table-based-service/src/test/java/io/stargate/auth/AuthenticationSubjectTest.java
dimas-b: Renamed to `convertToAuthenticatedUser`, hope it's more readable :)

PR #689, opened 2021-02-18
Improve error reporting when we load a schema from the database

olim7t:


graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/SchemaFirstCache.java
```diff
@@ -114,11 +116,19 @@ private GraphQL getGraphql(String namespace) throws Exception {
...
+  private ConcurrentMap<String, GraphqlHolder> initSchemas() {
     ConcurrentHashMap<String, GraphqlHolder> result = new ConcurrentHashMap<>();
     for (Keyspace keyspace : persistence.schema().keyspaces()) {
       String namespace = keyspace.name();
-      SchemaSource source = schemaSourceDao.getLatest(namespace);
+      SchemaSource source;
+      try {
+        source = schemaSourceDao.getLatest(namespace);
+      } catch (Exception e) {
+        // The only way this can happen is if the graphql_schema table exists but with the wrong
```
tomekl007, line 127: maybe at least `log.debug`?

graphqlapi/src/main/java/io/stargate/graphql/web/resources/schemafirst/SchemaFirstCache.java
```diff
@@ -147,11 +157,11 @@ void init() {
     }
 
     private GraphQL buildGraphql() {
-      // TODO handle errors better - things could still fail e.g. if we're loading a namespace that
-      // was not created through this Stargate instance, and the CQL schema doesn't match anymore.
       ProcessedSchema processedSchema =
-          new SchemaProcessor(authenticationService, authorizationService, dataStoreFactory)
+          new SchemaProcessor(authenticationService, authorizationService, dataStoreFactory, true)
```
tomekl007, line 161: do we need to have the `isPersisted` information in two places?
I can see that we are keeping it in the `SchemaProcessor` but also in the `CassandraMigrator` when calling `forPersisted()`. Isn't it enough to keep it in one place?

olim7t: `SchemaProcessor` and `CassandraMigrator` are independent classes, so we can't pass the information from one to the other.

PR #690, opened 2021-02-18
Add a way to filter messages in the deploySchema response

olim7t:


graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaResponseDto.java
```diff
@@ -0,0 +1,64 @@
...
+    return version;
+  }
+
+  public void setMessages(List<ProcessingMessage> messages) {
+    this.messages = messages;
+  }
+
+  public List<Map<String, Object>> getMessages(DataFetchingEnvironment environment) {
+    Stream<ProcessingMessage> stream = messages.stream();
+    String category = environment.getArgument("category");
```
tomekl007, line 36: maybe we should create an enum with all possible categories?

olim7t: Actually there is already one: `ProcessingMessageType`. But it also contains error statuses which are not used in this context. It would probably be cleaner to split it into two different enums, I'll give that a try.

tomekl007 (approved):

PR #694, opened 2021-02-19
Reuse CassandraSchemaHelper.compare in SchemaSourceDao

olim7t:


graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -129,15 +149,13 @@ BoundQuery schemaQuery(String namespace) {
...
+        .where(KEY_COLUMN_NAME, Predicate.EQ, UNIQUE_KEY)
         .build()
         .bind();
   }
 
   /** @return the new version */
-  public SchemaSource insert(String namespace, String newContents) throws Exception {
-
-    ensureTableExists(namespace);
+  public SchemaSource insert(String namespace, String newContents) {
```
olim7t, line 158: I removed the `ensureTableExists` call here because `insert` is not the first method we call anymore. If we get here, we can reasonably assume that the table is what we expect. Of course someone could always modify the table at any time, but that is true with any application.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -45,10 +47,28 @@
   @VisibleForTesting static final String APPLIED_COLUMN_NAME = "[applied]";
   private static final String DEPLOYMENT_IN_PROGRESS_COLUMN_NAME = "deployment_in_progress";
 
+  private static final Table EXPECTED_TABLE =
+      ImmutableTable.builder()
+          .keyspace("ks") // mock keyspace name, it won't be used
```
tomekl007, line 52: why keyspace is not used?

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -116,7 +136,7 @@ BoundQuery schemaQueryWithSpecificVersion(String namespace, UUID uuid) {
         .select()
         .column(VERSION_COLUMN_NAME, CONTENTS_COLUMN_NAME)
         .from(namespace, TABLE_NAME)
-        .where(KEY_CONDITION)
+        .where(KEY_COLUMN_NAME, Predicate.EQ, UNIQUE_KEY)
```
tomekl007, line 139: Are you removing the static for consistency?

tomekl007 (approved):

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
olim7t: Yes. I made you inline the other one on your last PR, so it didn't make sense to keep this one either.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
olim7t: The actual keyspace will be different every time.

`CassandraSchemaHelper.compare`ignores the keyspace, and `ensureTableExists` uses its own.

PR #695, opened 2021-02-19
Refactor EntityMappingModel.build() into a builder class

olim7t:
This is a purely technical refactoring to split the method into manageable chunks. There are no functional changes.


graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/processor/EntityMappingModelBuilder.java
```diff
@@ -0,0 +1,296 @@
...
+    }
+    return false;
+  }
+
+  /**
+   * Check that if the @key directive is present, it matches the CQL primary key.
+   *
+   * @return empty if the directive is invalid (an error has already been logged).
+   */
+  private Optional<Boolean> isFederated(
```
olim7t, line 184: This signature is... not great. What we would really need here is some sort of `Try` type, not `Optional`. But let's not go there.

But really it makes me consider replacing the whole `Optional` thing by a checked `SkipException`.

tomekl007:
Is there any new code here or only refactoring?

olim7t:
> Is there any new code here or only refactoring?

Only refactoring.

dimas-b (approved):

PR #696, opened 2021-02-19
Purge the graphql-schema table periodically

tomekl007:


graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaFetcherBase.java
```diff
@@ -114,6 +114,7 @@
         dataStore.execute(query.build(dataStore)).get();
       }
       SchemaSource newSource = schemaSourceDao.insert(namespace, input);
+      schemaSourceDao.purgeOldSchemaEntries(namespace);
```
tomekl007, line 117: This will be executed outside of the LWT transition (after the deployement_in_progress is set to `false`) but I don't think it is a problem. If another process (P2) will deploy a new schema before old entries are removed by P1 there will be a short period of time when there will be `NUMBER_OF_RETAINED_SCHEMA_VERSIONS+1` entries, but they will be purge by the P2 `purgeOldSchemaEntries` call.

There could be a situation that one process removes an entry that another tries to remove, but removing a non-existing entry is not problematic, right?

testing/src/test/java/io/stargate/it/http/graphql/graphfirst/SchemaDeploymentTest.java
```diff
@@ -36,6 +39,8 @@
 @ExtendWith(CqlSessionExtension.class)
 public class SchemaDeploymentTest extends BaseOsgiIntegrationTest {
 
+  private static final int NUMBER_OF_RETAINED_SCHEMA_VERSIONS = 10;
```
tomekl007, line 42: we don't have access to `SchemaSourceDao#NUMBER_OF_RETAINED_SCHEMA_VERSIONS` so I duplicated it here. 
I am not sure how can we handle it without the needed duplication :/

olim7t: Yeah I think it's acceptable to have a bit of duplication in ITs, we've done it in a few places already.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -282,4 +286,33 @@ public void abortDeployment(String namespace) throws ExecutionException, Interru
...
+                .queryBuilder()
+                .delete()
+                .from(namespace, TABLE_NAME)
+                .where(KEY_CONDITION)
+                .where(VERSION_COLUMN_NAME, Predicate.EQ, schemaSource.getVersion())
+                .build()
+                .bind();
+
+        dataStore.execute(deleteSchemaQuery).get();
+      }
```
olim7t, line 315: This works, but since timeuuids are comparable it can be simplified to a single query:
```
delete from graphql_schema where key = 'key' and version < :oldest_version_to_keep;
```

testing/src/test/java/io/stargate/it/http/graphql/graphfirst/GraphqlFirstClient.java
```diff
@@ -51,8 +51,16 @@ public GraphqlFirstClient(String host, String authToken) {
    * @return the resulting version.
    */
   public UUID deploySchema(String namespace, String contents) {
+    return deploySchema(namespace, null, contents);
+  }
+
+  public UUID deploySchema(String namespace, UUID expectedVersion, String contents) {
```
olim7t, line 57: Nit: I've generally stuck with plain strings for method arguments in this class, because some of the tests call them with malformed input, e.g. a string that is not a UUID.
That's not the case for this method (yet?), but I think it would be nice to keep things uniform.

graphqlapi/src/main/java/io/stargate/graphql/persistence/schemafirst/SchemaSourceDao.java
```diff
@@ -33,17 +33,20 @@
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 // TODO purge old entries
 public class SchemaSourceDao {
-
+  private static final Logger logger = LoggerFactory.getLogger(SchemaSourceDao.class);
```
olim7t, line 41: Nit: could you upper-case the name?

olim7t (changes requested):

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/DeploySchemaFetcherBase.java
olim7t: Right, the delete queries target specific versions so they are idempotent.
A race condition is unlikely, and even then having an extra version for a short period of time would not be a problem.

olim7t (approved):

PR #699, opened 2021-02-19
Fix #536 - Replace String padding by library call

### LINKED ISSUE #536: Replace String padding code by library call
A minor refactoring: maybe we could replace the code here: https://github.com/stargate/stargate/blob/master/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java#L91-L97 by calls to either `org.apache.commons.lang.StringUtils.leftPad` or `com.google.common.base.Strings.padStart`.

eribeiro:
Replace String padding by library call and minor refactoring.

### DIFF
--- a/restapi/src/test/java/io/stargate/web/docsapi/service/DocumentServiceTest.java
+++ b/restapi/src/test/java/io/stargate/web/docsapi/service/DocumentServiceTest.java
@@ -71,6 +71,7 @@ import org.mockito.Mockito;
 public class DocumentServiceTest {
   private DocumentService service;
   private Method convertToBracketedPath;
+  private Method leftPadTo6;
   private Method convertArrayPath;
   private Method isEmptyObject;
   private Method isEmptyArray;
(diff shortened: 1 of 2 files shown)


dougwettlaufer (approved):

restapi/src/test/java/io/stargate/web/docsapi/service/DocumentServiceTest.java
```diff
@@ -219,30 +216,6 @@ public void convertToBracketedPath() throws IllegalAccessException, InvocationTa
     assertThat(res).isEqualTo("$['']['a']");
   }
 
-  @Test
-  public void leftPadTo6() throws InvocationTargetException, IllegalAccessException {
-    String result = (String) leftPadTo6.invoke(service, "");
-    assertThat(result).isEqualTo("000000");
-
-    result = (String) leftPadTo6.invoke(service, "00");
-    assertThat(result).isEqualTo("000000");
```
dimas-b, line 228: I think it's worth keeping this test. Even though we use a lib to do the work, the fact that `0` is used for padding it something affected by our code... but of course we do not need as many test cases now.

dimas-b (approved):

PR #701, opened 2021-02-19
Use superuser Query/Client state for external users.

dimas-b:
This is to allow forwarding DDL to storage nodes.

Note: access control for external users is performed
independently of their C* role.

### DIFF
--- a/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
+++ b/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
@@ -42,7 +42,9 @@ import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import org.apache.cassandra.auth.AuthenticatedUser;
+import org.apache.cassandra.auth.CassandraRoleManager;
 import org.apache.cassandra.auth.IAuthContext;
+import org.apache.cassandra.auth.RoleResource;
 import org.apache.cassandra.auth.user.UserRolesAndPermissions;
 import org.apache.cassandra.concurrent.TPCTaskType;
 import org.apache.cassandra.config.Config;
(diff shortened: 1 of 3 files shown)


persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -428,15 +437,18 @@ public Persistence persistence() {
...
-          @SuppressWarnings("unused")
-          ClientState unused =
-              this.clientState.login(new AuthenticatedUser(user.name())).blockingGet();
+          loginSingle = this.clientState.login(new AuthenticatedUser(user.name()));
         }
+
+        // For now, we're blocking as the login() API is synchronous. If this is a problem, we may
+        // have to make said API asynchronous, but it makes things a tad more complex.
+        @SuppressWarnings("unused")
+        ClientState unused = loginSingle.blockingGet();
```
dimas-b, line 450: This change is critical even if the "external user" change does not go in for some reason.

tomekl007: Could you explain how it works from the client (`cqlsh`) perspective?

dimas-b: This particular block of changes ensures that we always `get` the login result, which in turn ensures that the login code is executed and the user is saved in `ClientState`.

This change is irrelevant to the "superuser" changes in this PR. Without this change the user information may not always be saved in `ClientState`, which causes auth failures.

dimas-b: As for the end-to-end process:
* `cqlsh` logs in with user = `token`, password = `<the-token-text>`
* `AuthenticationService` creates an `AuthenticatedUser` using the "superuser" role
* the `AuthenticatedUser` is saved in `ClientState`
* subsequent statements are executed under the `AuthenticatedUser`
** If the statement is DML or SELECT, Stargate will use the `AuthorizationService` to check permissions using the token from the `AuthenticatedUser` object.
** If the statement is DDL, Stagate will first use the `AuthorizationService` to check permissions using the token from the `AuthenticatedUser` object. If running with DSE and schema change forwarding is enabled, Stargate will forward the schema change to storage nodes to be executed there under the "superuser" role.

tomekl007: Ok, I see. Thank for the explanation. I wonder how hard will it be to create an integration test for this case.

tomekl007 (approved):

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
dimas-b: Not too hard, I hope but it does require changes to our `ExternalStorage` class, I think because we need to start storage with auth enabled and by default it is disabled. I'll file a separate issue for that.

dougwettlaufer (approved):

PR #703, opened 2021-02-20
Add ability to update counters in restapi

### LINKED ISSUE #688: Provide counter cql column update/write feature
Today with REST API, we could do reads of `counter` cql column, but writes/upserts is not possible and hence this feature request

dougwettlaufer:
resolves #688

### DIFF
--- a/persistence-api/src/main/java/io/stargate/db/query/builder/ValueModifier.java
+++ b/persistence-api/src/main/java/io/stargate/db/query/builder/ValueModifier.java
@@ -32,9 +32,18 @@ public abstract class ValueModifier {
         .build();
   }
 
+  public static ValueModifier of(String columnName, Value<?> value, Operation operation) {
+    return ImmutableValueModifier.builder()
+        .target(Target.of(columnName))
+        .operation(operation)
+        .value(value)
+        .build();
+  }
+
   @Immutable
   @Style(visibility = ImplementationVisibility.PACKAGE)
   abstract static class Target {
+
     abstract String columnName();
 
     abstract @Nullable String fieldName(); // only set for UDT field access
--- a/restapi/src/test/java/io/stargate/web/resources/ConvertersTest.java
+++ b/restapi/src/test/java/io/stargate/web/resources/ConvertersTest.java
@@ -2,6 +2,7 @@ package io.stargate.web.resources;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.params.provider.Arguments.arguments;
 
 import com.datastax.oss.driver.api.core.data.CqlDuration;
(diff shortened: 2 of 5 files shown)


restapi/src/test/java/io/stargate/web/resources/ConvertersTest.java
```diff
@@ -376,4 +404,37 @@ private static ObjectMapper testObjectMapper() {
     Server.configureObjectMapper(mapper);
     return mapper;
   }
+
+  @Test
+  public void colToValuePositiveCounter() {
+    ValueModifier modifier = Converters.colToValue("popularity", "+1", COUNTER_TABLE);
```
dimas-b, line 410: Is the `+` required in value? If not, could you add a test case for an increment without an explicit `+` sign?

dimas-b (approved):

olim7t (approved):
LGTM, only a couple of trivial remarks.

restapi/src/test/java/io/stargate/web/resources/ConvertersTest.java
dougwettlaufer: Being a counter, technically yes but since we default to increment you could get away with only `1`.

restapi/src/test/java/io/stargate/web/resources/ConvertersTest.java
```diff
@@ -376,4 +404,37 @@ private static ObjectMapper testObjectMapper() {
...
+    ValueModifier expected = ValueModifier.of("popularity", Value.of(1L), Operation.INCREMENT);
+
+    assertThat(modifier).usingRecursiveComparison().isEqualTo(expected);
```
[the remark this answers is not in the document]
dougwettlaufer: Weird. I must have had something else at play because it was initially failing without it.

restapi/src/test/java/io/stargate/web/resources/ConvertersTest.java
dimas-b: From the parsing code it looks like we simply parse it an a numeric value without enforcing `+` or `-` :thinking:

persistence-api/src/main/java/io/stargate/db/query/builder/ValueModifier.java
```diff
@@ -32,9 +32,18 @@ public static ValueModifier of(String columnName, Value<?> value) {
         .build();
   }
 
+  public static ValueModifier of(String columnName, Value<?> value, Operation operation) {
```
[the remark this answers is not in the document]
dougwettlaufer: Postfix notation :). It made sense to me to put it at the end since the method above is `public static ValueModifier of(String columnName, Value<?> value) {...}`

restapi/src/test/java/io/stargate/web/resources/ConvertersTest.java
dougwettlaufer: Right we don't need to enforce anything. Input of `+1` or `1` will be an increment and `-1` will decrement.

ds-steven-matison:
Ty @dougwettlaufer for all the hard work and attention to this as a priority!

Request:

```
curl --location --request PUT 'https://[clusterId]-us-east1.apps.astra.datastax.com/api/rest/v2/keyspaces/memegen/memes_likes/161865971' \
--header 'X-Cassandra-Token: [authToken]' \
--data-raw '{
    "likes": "+1"
}'

```
Response: (200)
```
{
    "data": {
        "likes": "+1"
    }
}
```

PR #709, opened 2021-02-23
allow partial PKs in queries (CKs)

tomekl007:


graphqlapi/src/test/java/io/stargate/graphql/schema/schemafirst/processor/MappingModelTest.java
```diff
@@ -0,0 +1,55 @@
...
+import org.junit.jupiter.api.Test;
+
+class MappingModelTest {
+
+  @Test
+  public void shouldBuildAMappingModelWithOnePrimaryKey() {
+    TypeDefinitionRegistry typeDefinitionRegistry =
+        new SchemaParser()
+            .parse(
+                "type User { id: ID! name: String username: String } "
```
tomekl007, line 33: my understanding is that we need to add a way to pass the clustering key in the entity and query, is that right?
As far as I can see the `!` syntax is for Primary keys. @olim7t what syntax should we use for adding other primary keys?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/processor/QueryMappingModel.java
```diff
@@ -72,6 +72,15 @@ public EntityMappingModel getEntity() {
...
       ProcessingContext context) {
+    System.out.println(
+        "QueryMappingModel.build, query: "
+            + query
+            + " , parentTypeName: "
+            + parentTypeName
+            + " , entities: "
+            + entities
+            + "context:"
+            + context);
```
tomekl007, line 84: As far as I can see, the QueryMappingModel is already taking all PKs (including Cks) into account by using this:
```
List<FieldMappingModel> primaryKey = entity.getPrimaryKey();
```
Are the only modifications that we need to do as a part of this ticket is:
- adding a way to pass CK in query and entity
- add a possibility to return a list from `QueryMappingModel` 
Is that true?

graphqlapi/src/test/java/io/stargate/graphql/schema/schemafirst/processor/MappingModelTest.java
olim7t: `!` means non-null in GraphQL, it's not directly related to primary keys (although it makes sense to mark PK fields as such, since that's the case in Cassandra).

We already have syntax to add partition and clustering columns. The explicit way is via the `cql_column` directive:
```graphql
type Foo {
  pk1: Int! @cql_column(partitionKey: true)
  pk2: Int! @cql_column(partitionKey: true)
  cc1: Int! @cql_column(clusteringOrder: ASC)
  cc2: Int! @cql_column(clusteringOrder: DESC)
}
# produces CQL `PRIMARY KEY ((pk1,pk2), cc1, cc2)`
```
For convenience, we also have a convention that if the partition key is a single `ID`, `Uuid` or `TimeUuid` field, then you can omit the annotation if the field is the first in the object. That's what happens in your `User` type above:
```graphql
type User {
  id: ID! # implicit: @cql_column(partitionKey: true)
  name: String
  username: String
}
```
This would also work:
```graphql
type Foo {
  pk: Int!
  cc1: Int! @cql_column(clusteringOrder: ASC)
  cc2: Int! @cql_column(clusteringOrder: DESC)
}
# produces CQL `PRIMARY KEY (pk, cc1, cc2)`
```

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/processor/QueryMappingModel.java
olim7t: Yes. Given the model:
```graphql
type Foo {
  pk1: Int! @cql_column(partitionKey: true)
  pk2: Int! @cql_column(partitionKey: true)
  cc1: Int! @cql_column(clusteringOrder: ASC)
  cc2: Int! @cql_column(clusteringOrder: DESC)
}
```

The only query that can be mapped right now is:
```graphql
foo(pk1: Int!, pk2: Int!, cc1: Int!, cc2: Int!): Foo
```
The goal is to add:
```graphql
foos(pk1: Int!, pk2: Int!, cc1: Int!): [Foo]
foos(pk1: Int!, pk2: Int!): [Foo]
```
Any prefix of the primary key should be allowed, as long as it's in order, and the full partition key is present (e.g. `foos(pk1: Int!)` is invalid).

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/processor/QueryMappingModel.java
```diff
@@ -72,33 +69,52 @@ public EntityMappingModel getEntity() {
...
 
     Type<?> returnType = query.getType();
-    String entityName = (returnType instanceof TypeName) ? ((TypeName) returnType).getName() : null;
+    boolean isListType = returnType instanceof ListType;
+    String entityName = null;
```
tomekl007, line 84: I am not sure if this is the correct approach. 
Let's assume that we have a given graphql schema and query:
```
"type Foo {\n"
                    + "  pk1: Int! @cql_column(partitionKey: true)\n"
                    + "  pk2: Int! @cql_column(partitionKey: true)\n"
                    + "  cc1: Int! @cql_column(clusteringOrder: ASC)\n"
                    + "  cc2: Int! @cql_column(clusteringOrder: DESC)\n"
                    + "}"
                    + "type Query { foo1(pk1: Int!, pk2: Int!, cc1: Int!): [Foo] }"
```
The map of entities will contain only `Foo` entity definition. I wonder If we should add a list of type when building a map of entities?
In the `MappingModel` we have this logic that builds the `EntityMappingModel` for `Foo` :
```
    for (ObjectTypeDefinition type : registry.getTypes(ObjectTypeDefinition.class)) {
      //...
      new EntityMappingModelBuilder(type, context)
          .build()
          .ifPresent(entity -> entitiesBuilder.put(type.getName(), entity));
    }
```
maybe for all non `query` `ObjectTypeDefinition type` we should build a normal Foo type but also a list of foo (since we will support that now)?
If yes, then we need some convention of how those list of foo will be stored in the entities map (the key is a string).
Maybe we should create `list_of_Foo`?
So the logic will be:
```

    for (ObjectTypeDefinition type : registry.getTypes(ObjectTypeDefinition.class)) {
      if (typesToIgnore.contains(type)) {
        continue;
      }
      new EntityMappingModelBuilder(type, context)
          .build()
          .ifPresent(entity -> entitiesBuilder.put(type.getName(), entity));
   // for every entity build a corresponding list[entity]
   new EntityMappingModelBuilder(type, context) // new mapping? or pass the info isList to an existing Builder?
          .build()
          .ifPresent(entity -> entitiesBuilder.put("list_of_" + type.getName(), entity));
    }
```
However, in that case, we would need to create a variant of `EntityMappingModel` that contains a `List<EntityMappingModel>`. 
Or maybe this `entity -> list[entity]` mapping should be handled elsewhere and we don't need an additional entry in the entities map - @olim7t wdyt?

olim7t: I think the simplest approach is to encode this information in a separate field: `QueryMappingModel.returnsList()`. Then in `QueryFetcher` it will be pretty easy to process the result set accordingly: if `returnsList()`, read all the rows and return a list of maps.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/dynamic/QueryFetcher.java
```diff
@@ -43,18 +45,35 @@ public QueryFetcher(
...
-    return querySingleEntity(
-        model.getEntity(),
-        environment.getArguments(),
-        Optional.of(model.getInputNames()),
-        dataStore,
-        keyspace,
-        authenticationSubject);
+
+    if (model.isReturnsList()) {
+      List<Map<String, Object>> listOfEntities =
```
tomekl007, line 56: @olim7t I followed your suggestion from https://github.com/stargate/stargate/pull/709#discussion_r582431729 but there is one caveat here. Since in the case of a list is returned by the `queryListOfEntities`, we need to change this fetcher to return list (`List<Map<String, Object>>` cannot be the case to `Map<String, Object>`).
However, in that case, if the query returns one value, we need to wrap it in the `Collections.singletonList`.
I am not sure if this is a good solution for us. wdyt?

testing/src/test/java/io/stargate/it/http/graphql/graphfirst/SelectTest.java
```diff
@@ -0,0 +1,119 @@
...
+    CLIENT.deploySchema(
+        NAMESPACE,
+        "type Foo @cql_input {\n"
+            + "  pk1: Int! @cql_column(partitionKey: true)\n"
+            + "  pk2: Int! @cql_column(partitionKey: true)\n"
+            + "  cc1: Int! @cql_column(clusteringOrder: ASC)\n"
+            + "  cc2: Int! @cql_column(clusteringOrder: DESC)\n"
+            + "}\n"
+            + "\n"
+            + "type Query { fooPartial(pk1: Int!, pk2: Int!, cc1: Int!): [Foo]"
```
tomekl007, line 51: because of the `QueryFatcher` change discussed here: https://github.com/stargate/stargate/pull/709#discussion_r582779608
both queries need to return `[Foo]` now.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/dynamic/DynamicFetcher.java
```diff
@@ -231,11 +257,51 @@ private Object toGraphqlUdtValue(UdtValue udtValue) {
...
-      FieldMappingModel field = entity.getPrimaryKey().get(i);
+
+    List<FieldMappingModel> primaryKey = entity.getPrimaryKey();
+    if (valueNamesOptional.isPresent()) {
+      List<String> valueNames = valueNamesOptional.get();
+      // use only PKs that are provided in the valueNames
+      // todo should we use pk.getGraphqlName() or pk.getCqlName()?
+      primaryKey =
+          entity.getPrimaryKey().stream()
+              .filter(pk -> valueNames.contains(pk.getCqlName()))
```
tomekl007, line 297: I am not sure if we should use `pk.getGraphqlName()` or `pk.getCqlName()`?
I think that `pk.getCqlName()` because `entity.getPrimaryKey()` returns cql names.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/dynamic/QueryFetcher.java
olim7t: The simplest solution is to have the fetcher return `Object`.

If that makes the code cleaner, we can indeed have two subclasses and have `QueryMappingModel.getDataFetcher` pick the right one depending on the value of `returnsList`.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/dynamic/DynamicFetcher.java
olim7t: `entity.getPrimaryKey()` returns a list of `FieldMappingModel`, that have both a GraphQL and a CQL name.

This is because we allow users to specify a custom CQL name if they don't like the default mapping. For example given this definition:
```graphql
type User {
  id: ID
  userName: String @cql_column(name: "user_name")
}
```
The first field has the same names. But for the second field graphqlName=userName and cqlName=user_name.

In this method we're trying to bind the parameters of the mutation to the primary key fields in the query. They must have the same types but we allow the names to be different, for example:
```graphql
type Query {
  findUser(id_to_find: ID): User
}
```
`id_to_find` is the first argument, it matches with the first PK field `id` (`entity.getPrimaryKey.get(0)`). So we take the value of that field, convert it to a CQL value, and bind it to the CQL name of that entity field.

I don't follow what you are doing in this loop, I think the code should be very similar to the single-entity variant, except that you stop the iteration earlier because there are less parameters to bind. Maybe you can remember how many parameters the GraphQL query has, and use that as the limit:
```java
    for (int i = 0; i < argumentCount /* instead of entity.getPrimaryKey().size() */; i++) {
      FieldMappingModel field = entity.getPrimaryKey().get(i);
      int finalI = i;
      String inputName = valueNames.map(l -> l.get(finalI)).orElse(field.getGraphqlName());
      Object graphqlValue = values.get(inputName);
      whereConditions.add(
          BuiltCondition.of(
              field.getCqlName(),
              Predicate.EQ,
              toCqlValue(graphqlValue, field.getCqlType(), keyspace)));
    }
```

olim7t (approved):

PR #717, opened 2021-02-26
Unlogged batches property

EricBorczuk:
This exposes a parameter that can be used by a persistence layer to not use logged batches in stargate.

This affects the behavior of the documents API and graphql, so please take a close look at graphql (as I am unfamiliar)


dougwettlaufer (approved):

persistence-api/src/main/java/io/stargate/db/Persistence.java
```diff
@@ -94,6 +94,11 @@ default boolean supportsSecondaryIndex() {
   /** Returns true if the persistence backend supports Storage Attached Indexes. */
   boolean supportsSAI();
 
+  default boolean useLoggedBatches() {
```
dimas-b, line 97: I'd call it `supportsLoggedBatches`

persistence-api/src/main/java/io/stargate/db/Persistence.java
```diff
@@ -94,6 +94,11 @@ default boolean supportsSecondaryIndex() {
...
+    return Boolean.parseBoolean(
+        System.getProperty("stargate.persistence.loggedbatches.support.default", "true"));
```
dimas-b, line 99: If we delegate this decision to `Persistence` why use a system property here? I guess it should be done in `Persistence` implementations so that the returned value were provided thoughtfully.

dimas-b: Conversely, if we want to control Docs API behaviour independently of the persistence impl, shouldn't the property be defined somewhere in Doc API classes instead?

EricBorczuk: I guess if it's not always explicitly defined in the Persistence implementation, then sometimes you would want to configure it by setting properties at runtime? To be honest I just did it in an analogous way to how supports2i was done and didn't think much of it

olim7t (approved):

persistence-api/src/main/java/io/stargate/db/Persistence.java
dimas-b: Yeah... I know. I guess is those earlier days we wanted to keep internal interface backward-compatibility as much as possible :thinking: now, though, I think we have a lot of change going on at the persistence level, that keeping strict compatibility is no longer practical.

Making the new API mandatory, will draw attention from `Persistence` implementors, which is a good thing in this situation, I think.

@dougwettlaufer : what is you take on this?

EricBorczuk: I think this behavior is pretty closely tied to the persistence though - maybe it shouldn't be exposed as a property at all? (exposing the property does make it a bit more straightforward to test though)

dimas-b: For testing, would it be possible to allow overrides in Docs API code? As in: if override is not set, use `Persistence.supportsLoggedBatches` and let tests override?

EricBorczuk: Yea that sounds more right - I'll change to that

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -740,7 +753,11 @@ void deleteDeadLeaves(
     }
 
     // Fire this off in a future
-    dataStore.batch(queries, ConsistencyLevel.LOCAL_QUORUM);
+    if (useLoggedBatches) {
+      dataStore.batch(queries, ConsistencyLevel.LOCAL_QUORUM).join();
+    } else {
+      dataStore.unloggedBatch(queries, ConsistencyLevel.LOCAL_QUORUM).join();
+    }
```
dimas-b, line 760: nit: Extract this logic into a method? It is repeated 3 times in this class :shrug:

dimas-b (approved):

persistence-api/src/main/java/io/stargate/db/Persistence.java
dougwettlaufer: @dimas-b I can see it both ways. I get that keeping consistent with the 2i makes sense but personally I'd lean to deferring this to the services to decide. As in maybe docsapi wants to use logged batches but graphql is fine with unlogged. But really I'm good either way.

dimas-b: yeah, I think we have it this way in the latest commit. `Persistence` declares whether logged batches are supported and Docs API has a `-D` that controls whether they are actually used or not (defaults to the persistence-level flag).

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
EricBorczuk: I want to, but I really can't have CI run again 😬

EricBorczuk: If it fails for any reason I will make this change

PR #722, opened 2021-02-27
Infer field list for Federation key directive

olim7t:
This also fixes a couple of conversion bugs when processing the `_entities` query.

### DIFF
--- a/testing/src/test/java/io/stargate/it/http/graphql/graphfirst/FederationTest.java
+++ b/testing/src/test/java/io/stargate/it/http/graphql/graphfirst/FederationTest.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


testing/src/test/java/io/stargate/it/http/graphql/graphfirst/FederationTest.java
```diff
@@ -0,0 +1,137 @@
...
+          + "  k1: Int! @cql_column(partitionKey: true)\n"
+          + "  k2: Int! @cql_column(partitionKey: true)\n"
+          + "  cc1: Int! @cql_column(clusteringOrder: ASC)\n"
+          + "  cc2: Int! @cql_column(clusteringOrder: ASC)\n"
+          + "}\n"
+          + "type Key @cql_entity(target: UDT) { k: Int }\n"
+          + "type Entity4 @key { k: Key! @cql_column(partitionKey: true) }\n"
+          + "type Query {\n"
+          + "  entity1(k: ID!): Entity1\n"
+          + "}\n";
```
tomekl007, line 49: Isn't the federation's goal is allowing clients to fetch aggregated entities(i.e., from N micro-services) via one query?
If yes, is it possible to do it now?

olim7t: > Isn't the federation's goal is allowing clients to fetch aggregated entities(i.e., from N micro-services) via one query?

Yes. In order to be one of those micro-services, Stargate needs to provide an `_entities` query. This is what we are testing in this class.
The `_entities` query does not need to be defined explicitly here, it gets added automatically when we process the schema with the `federation-jvm` library (see `SchemaProcessor.buildGraphql`).

However I need at least one explicit query here, otherwise the schema will fail to validate (the `Query` type must exist). That's why we have this `entity1()`, but really it's not used anywhere in the test. Maybe we should not require an explicit `Query` type if there are federated entities, I'll create a ticket.

PR #725, opened 2021-03-02
set authenticator via system property

tomekl007:


dimas-b (commented):
I'm not sure this is needed by any existing use cases.

PR #726, opened 2021-03-02
Improved error message when schema agreement fails, as currently it always logs the check took zero milliseconds

maxtomassi:
At the moment, when schema agreement is not reached after the maximum amount of retries, it throws an exception with message:
```
"Failed to reach schema agreement after 0 milliseconds."
```

That's because `count` is used in the error message, but it's value is zero at that point, as we decrement it every time we perform a new retry.


dougwettlaufer (approved):

dougwettlaufer:
Hi @maxtomassi looks like the files need reformatted. Can you please run `mvn xml-format:xml-format fmt:format -Pdse`?

maxtomassi:
@dougwettlaufer sorry I missed the formatting check. I've just ran that command and re-pushed. Thanks!

dimas-b (approved):

PR #731, opened 2021-03-03
Require authorization to read DML schemas

olim7t:
Motivation:

Unauthorized users were correctly prevented from executing data queries,
but they could still introspect the GraphQL schemas, which reveals
details about the CQL data model.

Modifications:

Perform an authorization check before we select the GraphQL schema to
use.

Result:

Unauthorized users will get a 401 for any GraphQL query, whether it's an
introspection query or one that goes to the backend.

Note that the DDL schema is not protected, it's a generic schema that is
the same for any Stargate installation, so exposing it doesn't reveal
any sensitive information.

### DIFF
--- a/graphqlapi/src/main/java/io/stargate/graphql/web/resources/cqlfirst/GraphqlDmlResource.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/web/resources/cqlfirst/GraphqlDmlResource.java
@@ -16,15 +16,23 @@
 package io.stargate.graphql.web.resources.cqlfirst;
 
 import graphql.GraphQL;
+import io.stargate.auth.AuthenticationService;
+import io.stargate.auth.AuthenticationSubject;
+import io.stargate.auth.AuthorizationService;
+import io.stargate.auth.SourceAPI;
+import io.stargate.auth.UnauthorizedException;
 import io.stargate.graphql.web.RequestToHeadersMapper;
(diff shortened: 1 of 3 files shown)


graphqlapi/src/main/java/io/stargate/graphql/web/resources/cqlfirst/GraphqlDmlResource.java
```diff
@@ -163,4 +188,19 @@ private GraphQL getGraphl(
       return graphql;
     }
   }
+
+  private boolean isAuthorized(String token, String keyspaceName, Map<String, String> headers) {
+    try {
+      AuthenticationSubject authenticationSubject =
+          authenticationService.validateToken(token, headers);
```
dimas-b, line 195: Could we pass the `AuthenticationSubject` into subsequent validation calls on the same request (to avoid re-authentication)?

This is a nice to have, not required for this PR, I think.

dimas-b (approved):

graphqlapi/src/main/java/io/stargate/graphql/web/resources/cqlfirst/GraphqlDmlResource.java
olim7t: Yes, there is a bit of duplication, we can definitely clean that up in a follow-up PR. I'll create at ticket.

dougwettlaufer (approved):

Issue #142, opened 2020-09-24
Relay nested pagination


blerghh:
Nested pagination is very difficult from an api perspective because there could be a large number of pageState parameters being passed in. The other option is to utilize connections which will allow new pages to be streamed in without having to recall the entire api.

For example:

```
fragment Feed_user on User {
  # This connection, under a specific User, will be uniquely identified by
  # the key "Feed_feed" and the value of `$orderBy` (given that no `filters` were provided)
  feed(
    first: $count
    after: $cursor
    orderby: $orderBy
  ) @connection(key: "Feed_feed") {
    edges {
      node {
        id
      }
  }
}
```

https://relay.dev/docs/en/pagination-container

olim7t:
The Relay example is a client-side feature:

> Relay uses directives to add additional information to GraphQL documents, which are used by the Relay Compiler to generate the appropriate runtime artifacts. These directives only appear in your application code and are removed from requests sent to your GraphQL server.

(https://relay.dev/docs/en/graphql-in-relay#directives)

We don't control which client is used to access our API, so we should limit ourselves to standard GraphQL as defined by the spec. If we want streaming, we should look at subscriptions and maybe continuous paging on the server side (but that is only available with DSE).

olim7t:
After reading up a bit more on Relay, it turns out I didn't have the full picture there ☝️ 

Relay directives are indeed compiled at build time, but Relay's pagination feature assumes a particular schema on the server. So we could support that structure (either in the CQL-first or the GraphQL-first API) for Relay users.

PR #735, opened 2021-03-04
Implement paging for SELECT queries

olim7t:


testing/src/test/java/io/stargate/it/http/graphql/schemafirst/SelectTest.java
```diff
@@ -47,73 +47,189 @@ public static void setup(StargateConnectionInfo cluster, @TestKeyspace CqlIdenti
...
+            + "  fooByPkAndCc1(pk1: Int!, pk2: Int!, cc1: Int!): [Foo]\n"
+            + "  fooByPk(pk1: Int!, pk2: Int!): [Foo]\n"
+            + "  fooByPkLimit(\n"
+            + "    pk1: Int!,\n"
+            + "    pk2: Int!\n"
+            + "  ): [Foo] @cql_select(limit: 5)\n"
+            + "  fooByPkPaginated(\n"
+            + "    pk1: Int!,\n"
+            + "    pk2: Int!,\n"
+            + "    pagingState: String @cql_pagingState\n"
```
tomekl007, line 65: Shouldn't we consider implementing paging according to the `Relay` specification? 
i.e as discussed in the `production-ready grapqhl book`:
```
type PageInfo {
  endCursor: String
  hasNextPage: Boolean!
  hasPreviousPage: Boolean!
  startCursor: String
}

```
It may be simpler to integrate for clients that are already familiar with the `Relay`. I am not sure how widely it is adopted. wdyt?

olim7t: Yes, this is only a first version, but we could have other pagination options:

Relay pagination is on the roadmap for CQL-first (#142), and I agree that we should have it in GraphQL-first too.

We could also add a more traditional offset-based variant, like [OffsetPager](https://docs.datastax.com/en/drivers/java/4.10/com/datastax/oss/driver/api/core/paging/OffsetPager.html) in the Java driver.

PR #738, opened 2021-03-05
Clean up GraphQL authentication (fixes #732)

### LINKED ISSUE #732: Clean up GraphQL authentication
Following #731 we now perform authentication early in `GraphqlDmlResource`, so we should hold on to the `authenticationSubject` and pass it along to the data fetchers, instead of retrieving it again in `CassandraFetcher.get`.

olim7t:
Centralize auth in AuthenticationFilter, that will be applied before
every GraphQL REST resource annotated with `@Authenticated`.

Once we have the authentication subject, store it as a request
attribute, and later in the GraphQL context. We don't need to pass
the AuthenticationService into every fetcher anymore.


EricBorczuk (approved):
Seems straightforward overall :+1:

Issue #739, opened 2021-03-05
Add additional system properties to support other keyspace settings for tablebased token keyspace


dougwettlaufer:
Need to add two new system properties so that the table based token keyspace can be created with replication strategies other than simple. One property will be a list of DC names while the other will be RF. From these we can then build a NetworkTopologyStrategy if desired.

dougwettlaufer:
SimpleStrategy is the ideal default for someone that's getting started with Stargate. Although I do agree that it's not the ideal setting for a production use case which is why we have https://github.com/stargate/stargate/issues/801 to log warnings about such settings to help avoid potential issues.

jsanda:
In k8ssandra, we want to have the operator configure the replication since it knows the desired cluster topology before deploying Stargate. Is it possible to tell Stargate to altogether skip creating the auth keyspace?

adutra:
> In k8ssandra, we want to have the operator configure the replication since it knows the desired cluster topology before deploying Stargate. Is it possible to tell Stargate to altogether skip creating the auth keyspace?

The keyspace is already being created with `IF NOT EXISTS` so I guess that's already the current Stargate behavior.

dougwettlaufer:
> Is it possible to tell Stargate to altogether skip creating the auth keyspace?

Yep, just set `-Dstargate.auth_tablebased_init=false`

PR #742, opened 2021-03-08
Validate CQL access to storage nodes during readiness and liveness checks

tomekl007:


health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
```diff
@@ -82,4 +102,45 @@ protected Result check() {
       return Result.unhealthy("Unable to access DataStore: " + e);
     }
   }
+
+  private void ensureTableExists(DataStore dataStore)
```
tomekl007, line 106: I think that we should have a dedicated table that is used only for this health-check. Other tables (from `system.*`)are mocked by the Stargate.

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
```diff
@@ -20,55 +20,75 @@
...
 import org.slf4j.LoggerFactory;
 
 public class DataStoreHealthChecker extends HealthCheck {
-  private static final Logger logger = LoggerFactory.getLogger(DataStoreHealthChecker.class);
-
+  private static final Logger logger = LoggerFactory.getLogger(StargateNodesHealthChecker.class);
+  private static final String KEYSPACE_NAME = "data_store_health_check";
+  private static final String TABLE_NAME = "dummy_table";
+  public static final String PK_COLUMN_VALUE = "dummy_pk";
+  public static final String VALUE_COLUMN_VALUE = "dummy_value";
```
dougwettlaufer, line 40: These all need to be configurable.

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
```diff
@@ -82,4 +102,45 @@ protected Result check() {
...
+        .execute(
+            dataStore
+                .queryBuilder()
+                .create()
+                .table(KEYSPACE_NAME, TABLE_NAME)
+                .ifNotExists()
+                .column(EXPECTED_TABLE.columns())
+                .build()
+                .bind())
+        .get();
```
dougwettlaufer, line 132: These should be in a static block to only run once. If at some point the keyspace and/or table disappear then that's a problem.

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
```diff
@@ -82,4 +102,45 @@ protected Result check() {
...
+
+    // insert record
+    dataStore
```
dougwettlaufer, line 135: Can we put this as part of the check so it's, insert record and then read it back?

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
```diff
@@ -20,55 +20,75 @@
...
+                  .column(VALUE_COLUMN_NAME)
+                  .from(KEYSPACE_NAME, TABLE_NAME)
                   .build()
                   .execute();
 
           Row row = rs.get().one();
-          String clusterName = row.getString("cluster_name");
-          UUID schemaVersion = row.getUuid("schema_version");
+          String pk = row.getString(PK_COLUMN_NAME);
+          String value = row.getString(VALUE_COLUMN_NAME);
```
dougwettlaufer, line 81: As commented below, can we check the value we just updated?

health-checker/src/main/java/io/stargate/health/StargateNodesHealthChecker.java
```diff
@@ -0,0 +1,85 @@
...
+import io.stargate.db.datastore.ResultSet;
+import io.stargate.db.datastore.Row;
+import java.util.UUID;
+import java.util.concurrent.Future;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceReference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class StargateNodesHealthChecker extends HealthCheck {
```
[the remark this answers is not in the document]
dimas-b: `system.local` and `system.peers` are relevant to CQL drivers. IIRC, we had some cases when those tables disappeared due to `nodetool resetlocalschema`.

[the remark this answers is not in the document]
dimas-b: If we decide to keep it, I'd prefer to add `system.peers` to `DataStoreHealthChecker` and put storage-level queries in a new class to reduce code disturbance since `DataStoreHealthChecker` already deals with `system.local`.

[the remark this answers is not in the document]
tomekl007: @dimas-b 
`"If we decide to keep it, I'd prefer to add system.peers to DataStoreHealthChecker and put storage-level queries in a new class to reduce code disturbance since DataStoreHealthChecker already deals with system.local."`
I've already done it this way, isn't it?

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
dougwettlaufer: Right, constructor. If it fails on init then we want that to fail startup because something is not in a healthy state.

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
tomekl007: keyspace and table name? Sure, I can load it from `System.property` with those defaults. However, I think that `PK_COLUMN_VALUE` and `VALUE_COLUMN_VALUE` don't need to be configured.

health-checker/src/main/java/io/stargate/health/StargateNodesHealthChecker.java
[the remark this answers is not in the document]
dimas-b: Ah... might be GH missing the rename... the diff looks like a big change in the old `DataStoreHealthChecker` plus old code showing as "new" in `StargateNodesHealthChecker` :shrug:

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
```diff
@@ -50,65 +49,55 @@
...
+        return Result.unhealthy("DataStore did not return the proper data.");
       }
 
-      logger.warn("DataStoreFactory service is not available");
-      return Result.unhealthy("DataStoreFactory service is not available");
-
+      if (pk.equals(PK_COLUMN_VALUE) && value.equals(VALUE_COLUMN_VALUE)) {
+        return Result.healthy("DataStore is operational");
+      } else {
+        return Result.healthy("DataStore did not return the proper data.");
```
dougwettlaufer, line 99: Shouldn't this be `unhealthy()`?

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
```diff
@@ -50,65 +49,55 @@
...
-          context.ungetService(dataStoreFactoryReference);
-        }
+      if (pk == null || value == null) {
+        return Result.unhealthy("DataStore did not return the proper data.");
       }
 
-      logger.warn("DataStoreFactory service is not available");
-      return Result.unhealthy("DataStoreFactory service is not available");
-
+      if (pk.equals(PK_COLUMN_VALUE) && value.equals(VALUE_COLUMN_VALUE)) {
```
dougwettlaufer, line 96: If you flip this around you can get rid of the if block above
```suggestion
      if (PK_COLUMN_VALUE.equals(pk) && VALUE_COLUMN_VALUE.equals(value)) {
```

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
```diff
@@ -50,65 +49,55 @@
...
+          .execute(
               dataStore
                   .queryBuilder()
-                  .select()
-                  .column(PK_COLUMN_NAME)
-                  .column(VALUE_COLUMN_NAME)
-                  .from(KEYSPACE_NAME, TABLE_NAME)
+                  .insertInto(KEYSPACE_NAME, TABLE_NAME)
+                  .value(PK_COLUMN_NAME, PK_COLUMN_VALUE)
+                  .value(VALUE_COLUMN_NAME, VALUE_COLUMN_VALUE)
```
dougwettlaufer, line 73: This value we're inserting should probably change from check to check by adding a UUID or timestamp. The pk value should probably be unique to a stargate instance too otherwise we could have N > 1 instances running all updating and reading a value for the same pk.

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
```diff
@@ -20,66 +20,116 @@
...
+          .build();
+  private final DataStoreFactory dataStoreFactory;
 
-  public DataStoreHealthChecker(BundleContext context) {
-    this.context = context;
+  public DataStoreHealthChecker(DataStoreFactory dataStoreFactory)
+      throws ExecutionException, InterruptedException {
+
+    this.dataStoreFactory = dataStoreFactory;
+    ensureTableExists(dataStoreFactory.createInternal());
```
dimas-b, line 58: What about making this optional as discussed offline?

Naturally, if this checker is disabled, the `check()` method should probably return perpetual `healthy` state with the appropriate message. WDYT?

health-checker/src/main/java/io/stargate/health/HealthCheckerActivator.java
```diff
@@ -16,9 +17,12 @@
 
   public static final String BUNDLES_CHECK_NAME = "bundles";
   public static final String DATASTORE_CHECK_NAME = "datastore";
+  public static final String STARGATE_NODES_CHECK_NAME = "stargate-nodes";
```
dimas-b, line 20: Sorry, but I find the change in checker names confusing. The `datatore` checker used to validate that the `DataStore` in operational internally, i.e. queries can be made against data managed by current Stargate node itself. Now this is done by the `stargate-nodes` checks, but that name is too generic. Staragte nodes have many services by the checker validates only the internal code path of `DataStore` and `Persistence`... How about keeping the old `datastore` name for the checker that queries `system.local`?

As for the new checker that talks to storage nodes, how about naming it `storage`? WDYT?

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
tomekl007: I was thinking that if the health-check is disabled, we should not register it at all. 
So it can be handled at the `HealthCheckerActivator` level:
```
if(Boolean.parseBoolean(System.getProperty("stargate.health_check.data_store.enabled", "true"))){
         healthCheckRegistry
          .get()
          .register(DATASTORE_CHECK_NAME, new DataStoreHealthChecker(dataStoreFactory.get()));
}
```
wdyt?

Regarding making the `ensureTableExists` optional. Do you want to do it because of the potential auth problem? 
Done in: 
https://github.com/stargate/stargate/pull/742/commits/b5d2072db5042cf95fa430ba09f9867814e91f06
But it means that we have yet another system property parameter.

health-checker/src/main/java/io/stargate/health/HealthCheckerActivator.java
tomekl007: good idea, done

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
tomekl007: I don't see a clear way of getting the id of the stargate instance. Do we generate a Stargate instance unique id?
Otherwise, we might try to use one of the addresses as pk, i.e.:
```
stargate.listen_address
```
but this setting is optional:
```
String listenAddress =
        System.getProperty("stargate.listen_address", InetAddress.getLocalHost().getHostAddress());
```
Otherwise, we could make a PK `UUID` and insert records with some low TTL. It seems that this solves both problems.
wdyt? 
implemented in:
https://github.com/stargate/stargate/pull/742/commits/a27e362d22b173473d155406cb2b73f92915f9df

dimas-b: We could use the host ID from `system.local`, I guess

dimas-b: Actually, if we access `system.local` for the host ID, we could combine the checks for `DataStore` availability (step 1) and storage availability (step 2) in the same checker, I guess. Step 2 could be optional (configurable). WDYT?

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
dimas-b: In some cases it may be preferable to create those tables manually. And yes, if DDL forwarding is enabled (assuming DSE backend) the "internal" `DataStore` will not be able to execute DDL (because it does not have a real user, which is required for forwarding DDL).

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
```diff
@@ -0,0 +1,142 @@
...
+public class StorageHealthChecker extends HealthCheck {
+  private static final Logger logger = LoggerFactory.getLogger(DataStoreHealthChecker.class);
+  private static final boolean SHOULD_CREATE_KS_AND_TABLE =
+      Boolean.parseBoolean(
+          System.getProperty("stargate.health_check.data_store.create_ks_and_table", "true"));
+  private static final String KEYSPACE_NAME =
+      System.getProperty("stargate.health_check.keyspace_name", "data_store_health_check");
+  private static final String TABLE_NAME =
+      System.getProperty("stargate.health_check.table_name", "health_table");
+  private static final int INSERT_TTL_SECONDS = 60;
```
dimas-b, line 45: Might be too short for under-resourced environments. If we only need it to automatically expire stale data, why not go with 600 (10 min)?

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
tomekl007: I think that having two separate health checks is cleaner. Also, the `StorageHelathCheck` cab be disabled, whereas `DataStore` doesn't. So for me, both of those offer different functionality. The solution with UUID as PK solves those problems.

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
```diff
@@ -0,0 +1,142 @@
...
+      ensureTableExists(dataStoreFactory.createInternal());
+    }
+  }
+
+  @Override
+  protected Result check() throws Exception {
+    try {
+      DataStore dataStore = dataStoreFactory.createInternal();
+
+      UUID pkValue = Uuids.timeBased();
```
dimas-b, line 74: I might be nit-picking, but I see a potential for unbounded data growth here if the health check is executed too often (some whatever reason). I know my suggestion to increase the TTL makes it worse :facepalm: 

What do you think about using logic like this:
* Have a static UUID PK value per Stargate node (changes on restart)
* Use current timestamp at the "test" value
* Insert with a user-provided write timestamp
* Read the value back
* If the value is what we write - healthy
* If the value is different - it might be a concurrent health check operation from the same node - check the the row timestamp, if it's greater that ours - healthy

WDYT?

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
dimas-b: Hmmm... I do not see how `StorageHelathCheck` can be disabled... I only see that DDL can be disabled, but the checker always runs... Did I miss something? :thinking:

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
tomekl007:
```
I might be nit-picking, but I see a potential for unbounded data growth here if the health check is executed too often 
```
Could you elaborate on how it may grow? Every health-check is inserting a new row with TTL = 10 minutes (after change) 
Assuming we will execute 1 health-check per second, this is `600` row maximum,. After this time they will be removed. 
Assuming that we have 10 Stargate nodes it will be `600 * 10`. So where's the problem?

dimas-b: It might be a problem if the health check URL is hit with a high rate... this would be an abnormal use case, but I am worries about us opening backdoors for malicious use cases.

tomekl007:
```
If the value is different - it might be a concurrent health check operation from the same node - check the row timestamp, if it's greater that ours - healthy
```
I think that this is not possible, because the WHERE clause fetches exactly the same record that was inserted.
We may consider using the same CL for both write and read queries, maybe `LOCAL_QUORUM`?

tomekl007: > It might be a problem if the health check URL is hit with a high rate... this would be an abnormal use case, but I am worried about us opening backdoors for malicious use cases.

If that the case, wouldn't it be a problem for all health-checks that we have? If this is a real-risk we should consider implementing a mechanism top of all health-checks to solve that problem. Not a work-around for it in this specific health-check (because other health-checks will have the same problem). wdyt?

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
tomekl007: I will add a `"stargate.health_check.data_store.enabled"` for disabling that. 
I've proposed a solution here:
https://github.com/stargate/stargate/pull/742#discussion_r592987436
still waiting for feedback at which level do the enable/disable switch.

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
dimas-b: If we do not register the health check sometimes we'd also have to fix `BundleService.defaultHealthCheckNames()` to return the appropriate list, because if a checker is in that list, but not registered, if we cause a failure to be reported by `CheckerResource.checkReadiness()`.

IMHO, it would be cleaner to always register the checker, but not execute the CQL if it is disabled and just return something like `healthy("Storage check disabled")`.

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
dimas-b: _(added a comment in the other linked thread)_

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
tomekl007: done in: https://github.com/stargate/stargate/pull/742/commits/7bab4def792b1660f746f8662dee176c64a449f7

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
dimas-b: Other health checkers do not generate data, so they do not have this problem, I think. That's why I'm proposing to use a fixed PK value. This way the amount of data does not grow with high usage.

dimas-b: By the way, this is the only remaining concern from my POV. Otherwise, the PR LGMT :+1: 

@dougwettlaufer : what's your take?

dougwettlaufer: +1 for fixed PK per stargate deployment. The default check for k8s is every 10 seconds so if you're running multiple stargate pods (as recommended) with each creating a new record every 10 seconds then I could definitely see data blowing up.

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
```diff
@@ -22,61 +22,46 @@
...
-  public DataStoreHealthChecker(BundleContext context) {
-    this.context = context;
+  public DataStoreHealthChecker(DataStoreFactory dataStoreFactory) {
+    this.dataStoreFactory = dataStoreFactory;
   }
 
   @Override
   protected Result check() {
     try {
-      ServiceReference<DataStoreFactory> dataStoreFactoryReference =
```
dougwettlaufer, line 42: Why are we removing this? The idea is to check that we can properly pull the bundle from the context and then exercise it.

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
```diff
@@ -0,0 +1,147 @@
...
+
+    this.dataStoreFactory = dataStoreFactory;
+    if (SHOULD_CREATE_KS_AND_TABLE) {
+      ensureTableExists(dataStoreFactory.createInternal());
+    }
+  }
+
+  @Override
+  protected Result check() throws Exception {
+    if (!STORAGE_CHECK_ENABLED) {
```
dougwettlaufer, line 73: Why not put this in the activator and if it isn't enable then just don't register it? Otherwise we're going to either create the keyspace/table even when it's not enabled or require setting 2 different flags to disable this check.

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
```diff
@@ -0,0 +1,147 @@
...
+public class StorageHealthChecker extends HealthCheck {
+  private static final Logger logger = LoggerFactory.getLogger(DataStoreHealthChecker.class);
+  private static final boolean SHOULD_CREATE_KS_AND_TABLE =
+      Boolean.parseBoolean(
+          System.getProperty("stargate.health_check.data_store.create_ks_and_table", "true"));
+  private static final String KEYSPACE_NAME =
+      System.getProperty("stargate.health_check.keyspace_name", "data_store_health_check");
+  private static final String TABLE_NAME =
+      System.getProperty("stargate.health_check.table_name", "health_table");
+  private static final int INSERT_TTL_SECONDS = 600;
```
dougwettlaufer, line 45: This seems really high considering most health checks will run several times a minute.

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
dimas-b: If we do not register a standard checker, we also have to make changes to `BundleService.defaultHealthCheckNames()` to not return its name. I'd be fine with that, but it looks like a cross-cutting concern... As for me, registering, but not running the check is cleaner.

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
dimas-b: I suggested bumping it up based on slow schema sync observations in smaller environments... how about 180 (3 min)?

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
tomekl007: @dougwettlaufer there is a write TTL. The problem that we discussing is with a malicious user that can call this endpoint N times per second. For your example of query per 10 second and ttl = 10 minutes even if we assume 20 nodes it will be `120 000` records. If we change the TTL to a lower value (i.e. 3 minutes) it's `36 000`. I don't such an amount of data is blowing up. I'll rework it.

health-checker/src/main/java/io/stargate/health/DataStoreHealthChecker.java
tomekl007: I've refactored it to make an activator "checking" that:
```
  private ServicePointer<DataStoreFactory> dataStoreFactory =
      ServicePointer.create(DataStoreFactory.class);
```
If the data store won't be available, it will detect that sooner. Extraction of `DataStore` is validated in the `*Activator` instead of health-checks. Isn't it enough?

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
tomekl007: I think that after reworking it to one row per stargate node (https://github.com/riptano/floodgate/compare/main...stargate_number_of_connections?expand=1) it's totally safe to have the TTL for 10 minutes. 
When storage is operational, the record for the stargate node will be overridden. In such a scenario, we should have `number of rows == number of Stargate nodes`
The ttl would be needed only when the stargate node will be recreated. If one stargate node is re-created, we will have N + 1 rows in the table (`N == number of stargate nodes`) for TTL time. Assuming that this is not a very common situation, the higher TTL should be ok.

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
```diff
@@ -0,0 +1,158 @@
...
+      throws ExecutionException, InterruptedException {
+    // create a dedicated health-check keyspace to not risk conflicts with users keyspace
+    dataStore
+        .execute(
+            dataStore
+                .queryBuilder()
+                .create()
+                .keyspace(KEYSPACE_NAME)
+                .ifNotExists()
+                .withReplication(Replication.simpleStrategy(1))
```
dimas-b, line 140: Sorry, I did not notice this before. In a replicated env. with `RF==1` shutting down any storage node will make _all_ Stargate nodes "unhealthy"... Would it make sense to make this parameter configurable so that people would not have to always use custom health checker DDL in realistic deployments?

tomekl007: As far as I can see, the same RF is used for `auth` and `grapqhQL` keyspaces. 
If we want to make it configurable, shouldn't we do the same for other keyspaces?

dimas-b: auth has to be custom-created in prod. I was just thinking that health-checking can avoid that, but perhaps is has to use custom DDL too :thinking: 

What is  the `graphQL` keyspace that you refer to? Where is it created? I do not know about that off the top of my head.

tomekl007: GraphQL:
https://github.com/stargate/stargate/pull/634/files#diff-e181be50d5f73de330179580432cbe71c8b32630dcbb05a81af75bbd7ab88fa4R196

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
```diff
@@ -0,0 +1,158 @@
...
+import io.stargate.db.schema.ImmutableColumn;
+import io.stargate.db.schema.ImmutableTable;
+import io.stargate.db.schema.Table;
+import java.time.Instant;
+import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class StorageHealthChecker extends HealthCheck {
```
mpenick, line 35: Is `StorageHealthChecker#check()` called every time a health check request is made to HTTP endpoint?

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
dimas-b: Well, I think GraphQL has a similar issue in this case - added a comment on that PR.

tomekl007: I can make it configurable, I am only afraid that this Health-Check already has a lot of config settings.

dimas-b: True, but I believe we have use cases for all of them :)

For RF, in particular, it really depends on the deployment, and `1` is not a viable option in production... I do not really see how we can have a default for that that would work in all cases. 

Since the checker is on by default, this PR might introduce unexpected behaviour changes into environments where people already use Stargate and have replicated storage.

The alternative, I guess, would be to avoid running DDL at all, and have this checker disabled by default. Users would have to define the checker's table as part of their workflows and then configure it to use the custom table.

dougwettlaufer: Auth needs to be more configurable

https://github.com/stargate/stargate/issues/739

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
```diff
@@ -0,0 +1,158 @@
...
+      logger.warn("checkIsReady failed with {}", e.getMessage(), e);
+      return Result.unhealthy("Unable to access Storage: " + e);
+    }
+  }
+
+  private boolean isGreaterThanOrEqual(Instant timestampRead, Instant expectedTimestamp) {
+    if (timestampRead == null) {
+      return false;
+    }
+    // in case there was a concurrent update, validate that the timestampRead is greater than
```
dougwettlaufer, line 125: When would this happen? There should only be one node modifying a partition at a time. Right?

dimas-b: Hypothetically the health check URL might be accessed concurrently and be executed concurrently.

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
tomekl007: I am yielding towards disabling it by default and enabling in classic deployment (Non-CNDB). It will be easier to use for our clients (CNDB,...) if it runs with just defaults. For other cases, we may enable it explicitly.
Implemented the RF setting in https://github.com/stargate/stargate/pull/742/commits/c968175c6d6736dad30851bd2503ac0770f1c38b

dimas-b (approved):

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
dimas-b: SGTM. @dougwettlaufer : WDYT?

dougwettlaufer: Yeah I think that makes sense. It's a heavier weight check so "opt-in" is probably a better solution.

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.concurrent.ExecutionException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class StorageHealthChecker extends HealthCheck {
+  private static final UUID STARGATE_NODE_ID = UUID.randomUUID();
+  private static final Logger logger = LoggerFactory.getLogger(DataStoreHealthChecker.class);
+  private static final boolean SHOULD_CREATE_KS_AND_TABLE =
+      Boolean.parseBoolean(
+          System.getProperty("stargate.health_check.data_store.create_ks_and_table", "true"));
```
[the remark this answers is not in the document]
dimas-b: How about combining this setting with `REPLICATION_FACTOR`? If `REPLICATION_FACTOR` is set we will run DDL, if not assume the user manages table creation?.. just one less -D param to manage.

PR #745, opened 2021-03-09
Add more features for delete mutations

olim7t:


graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/dynamic/DeleteFetcher.java
```diff
@@ -54,36 +62,36 @@ public DeleteFetcher(
...
+    Function<String, Object> fieldGetter;
+    if (model.getEntityArgumentName().isPresent()) {
+      Map<String, Object> entity = environment.getArgument(model.getEntityArgumentName().get());
+      fieldGetter = entity::get;
+    } else {
+      fieldGetter = environment::getArgument;
     }
 
+    Collection<BuiltCondition> conditions = new ArrayList<>();
+    bindPartitionKey(fieldGetter, keyspace, conditions);
```
tomekl007, line 85: those two methods (`bindPartitionKey` and `bindClusteringColumns`) seem like a good candidate for reusing in other Fetchers. Is it possible?

graphqlapi/src/main/resources/schemafirst/cql_directives.graphql
```diff
@@ -121,7 +121,23 @@ directive @cql_update on FIELD_DEFINITION
...
+directive @cql_delete(
+    """
+    The name of the type to delete. This is only needed if the mutation takes individual key fields
+    as arguments (as opposed to an instance of the type).
+    This must be a type that maps to a table.
+    """
+    targetEntity: String
+
+    """
+    What to do if the entity does not exist. If `false`, the operation will fail. If `true`, it
```
tomekl007, line 133: what does it mean that `the operation will fail.`? What will be returned to the caller in this case?

olim7t: Actually that description is wrong, the operation will always succeed even if the row does not exist. I'll fix it.

mpenick:
I had to upgrade to get this built locally:

testing/src/test/java/io/stargate/it/http/graphql/schemafirst/DeleteTest.java
```diff
@@ -0,0 +1,176 @@
...
+            + "type Query {\n"
+            + "  foo(pk: Int!, cc1: Int!, cc2: Int!): Foo\n"
+            + "}\n"
+            + "type DeleteFooResponse @cql_payload {\n"
+            + "  applied: Boolean"
+            + "}\n"
+            + "type Mutation {\n"
+            + "  deleteFoo(foo: FooInput!): Boolean \n"
+            + "  deleteFoo2(pk: Int, cc1: Int, cc2: Int): Boolean\n"
+            + "    @cql_delete(targetEntity: \"Foo\")\n"
```
mpenick, line 67: Because this has the `@cql_delete` directive should maybe this should be called `foo2` instead of `deleteFoo2`?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/dynamic/DeleteFetcher.java
```diff
@@ -52,36 +60,36 @@ public DeleteFetcher(
...
+    bindClusteringColumns(fieldGetter, keyspace, conditions);
```
mpenick, line 84: Should this do any validation to see if there's at least one condition that contains the partition key? Or is this not necessary because the field is not optional and validated by graphql?

graphqlapi/src/main/resources/schemafirst/cql_directives.graphql
```diff
@@ -121,7 +125,32 @@ directive @cql_update on FIELD_DEFINITION
...
+
+    * If `ifExists` is `false`, the mutation will always return `true`, whether it actually deleted
+      something or not.
+    * If `ifExists` is `true`, the mutation will return `true` if it deleted something, and `false`
+      otherwise.
+
+    Note that setting this flag to `true` might increase the latency of the operation. It should
+    not be used casually.
+
+    By convention, this flag will be set automatically if the mutation name ends with `IfExists`.
```
mpenick, line 150: That's pretty cool.

mpenick (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/dynamic/DeleteFetcher.java
olim7t: Yes, I plan to revisit and unify condition handling, I'll clean that up.

testing/src/test/java/io/stargate/it/http/graphql/schemafirst/DeleteTest.java
olim7t: The operation names are completely free-form. There is just the rule that if it starts with certain prefixes then you can skip the directive, but that's just a convenience.

Here we have to use the annotation because we want to provide the `targetEntity` argument.

olim7t:
> I had to upgrade to get this built locally (from 0.6.1 because it doesn't look to be available anymore):
> 
> ```
>     <dependency>
>       <groupId>com.apollographql.federation</groupId>
>       <artifactId>federation-graphql-java-support</artifactId>
>       <version>0.6.3</version>
>     </dependency>
> ```

Indeed, it's not in Maven central. My local build got it from our internal mirror, which seems to have pulled it from JCenter.

I'll update to the central version on the main branch.

PR #749, opened 2021-03-10
Allow "hidden" keyspaces in Schema

olim7t:


olim7t:
This will probably be closed in favor of #761.

PR #753, opened 2021-03-11
Add request rate metric to CQL

mpenick:


dougwettlaufer (approved):

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -54,6 +54,7 @@
 
   private AtomicInteger pausedConnections;
   private Meter requestDiscarded;
+  private Meter requestRate;
```
dimas-b, line 57: nit: this `Meter` produces multiple outputs, so there's no single "rate" in the strict sense I guess. By analogy with `requestDiscarded`, would it be more readable to call the new `Meter` something like `requestsReceived`?

dimas-b (approved):

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
mpenick: Good suggestions. Thanks. wdyt about `requestsProcessed`? (it's in the current commit)

dimas-b (approved):

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
dimas-b: sgtm :+1:

PR #756, opened 2021-03-11
Add metric to cql for auth errors

dougwettlaufer:


cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/AuthResponse.java
```diff
@@ -98,6 +98,7 @@ public AuthResponse(byte[] token) {
                 ClientMetrics.instance.markAuthFailure();
                 future.complete(ErrorMessage.fromException(ae));
               } catch (Exception e) {
+                ClientMetrics.instance.markAuthError();
                 future.completeExceptionally(e);
```
dimas-b, line 102: tangential to this PR: IIRC in case of unexpected exceptions during auth, the CQL client often does not get any clear error message and often just times out... might be worth having a deeper look.

dimas-b (approved):

mpenick (approved):

PR #757, opened 2021-03-12
Fixes #752 - NPE for bad column name in Where clause

### LINKED ISSUE #752: io.stargate.web.service.WhereParser#parseWhere throws NPE for bad column name
If a bad `where` is passed to the restapi then an NPE is thrown if the provided column name doesn't exist. Relevant code block

```java
Object val = value.asText();
Column.ColumnType columnType = tableData.column(fieldName).type(); // <-- This is the NPE
if (columnType != null) {
    Column.ColumnType valueType;
    if (op == FilterOp.$CONTAINS) {
    if (columnType.isCollection()) {
        valueType =
            columnType.isMap()
                ? columnType.parameters().get(1)
                : columnType.parameters().get(0);
    } else {
        throw new RuntimeException(
            String.format(
                "Field %s: operation %s is only supported for collection types",
                fieldName, rawOp));
    }
```

We need to check if the column exists BEFORE grabbing the type.

eribeiro:
Fixes invalid field name in RestApiv2's where clause.

### DIFF
--- a/restapi/src/main/java/io/stargate/web/resources/RowResource.java
+++ b/restapi/src/main/java/io/stargate/web/resources/RowResource.java
@@ -647,7 +647,7 @@ public class RowResource {
   private static Object filterToValue(Object val, String name, Table tableData) {
     Column column = tableData.column(name);
     if (column == null) {
-      throw new IllegalArgumentException(String.format("unknown field name '%s'", name));
+      throw new IllegalArgumentException(String.format("Unknown field name '%s'.", name));
     }
     Object value = val;
     Column.ColumnType type = column.type();
--- a/restapi/src/main/java/io/stargate/web/resources/RowResource.java
+++ b/restapi/src/main/java/io/stargate/web/resources/RowResource.java
@@ -644,10 +644,13 @@ public class RowResource {
     }
   }
 
-  private static Object filterToValue(Object val, String column, Table tableData) {
-    Column.ColumnType type = tableData.column(column).type();
+  private static Object filterToValue(Object val, String name, Table tableData) {
+    Column column = tableData.column(name);
+    if (column == null) {
+      throw new IllegalArgumentException(String.format("unknown field name '%s'", name));
+    }
     Object value = val;
-
+    Column.ColumnType type = column.type();
     if (type != null) {
       value = Converters.toCqlValue(type, (String) val);
     }
--- a/restapi/src/main/java/io/stargate/web/service/WhereParser.java
+++ b/restapi/src/main/java/io/stargate/web/service/WhereParser.java
@@ -97,7 +97,12 @@ public class WhereParser {
                         + "with two fields 'key' and 'value'.",
                     fieldName, rawOp));
           }
-          Column.ColumnType mapType = tableData.column(fieldName).type();
+          Column column = tableData.column(fieldName);
+          if (column == null) {
+            throw new IllegalArgumentException(
+                String.format("Unknown column name '%s'.", fieldName));
+          }
+          Column.ColumnType mapType = column.type();
           if (mapType == null || !mapType.isMap()) {
             throw new RuntimeException(
                 String.format(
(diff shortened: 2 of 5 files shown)


testing/src/test/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -563,6 +563,35 @@ public void getRowsWithNotFound() throws IOException {
...
+        objectMapper.writeValueAsString(row),
+        HttpStatus.SC_CREATED);
+
+    String whereClause = "{\"invalid_field\":{\"$eq\":\"test\"}}";
+    String body =
+        RestUtils.get(
+            authToken,
+            String.format(
+                "%s:8082/v2/keyspaces/%s/%s?where=%s", host, keyspaceName, tableName, whereClause),
+            HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
dimas-b, line 587: Specifying an invalid field name in query params looks like a user error to me... Do we want to report it as  `SC_INTERNAL_SERVER_ERROR`? :thinking:

eribeiro: You right. I am trying to figure out why it's coming as 500 instead of 400.

restapi/src/main/java/io/stargate/web/service/WhereParser.java
```diff
@@ -126,7 +126,12 @@
             conditions.add(conditionToWhere(fieldName, op, true));
           } else {
             Object val = value.asText();
-            Column.ColumnType columnType = tableData.column(fieldName).type();
+            Column column = tableData.column(fieldName);
+            if (column == null) {
+              throw new RuntimeException(
```
dougwettlaufer, line 131: Might want to throw an IllegalArgument and then catch it above so we can return a 400.

testing/src/test/java/io/stargate/it/http/RestApiv2Test.java
dougwettlaufer: Looks like you caught it down here. See my comment above about throwing something other than a RuntimeException so we can catch and return 400 rather than 500.

restapi/src/main/java/io/stargate/web/service/WhereParser.java
eribeiro: Yes, I pinpointed and changed that. 🙂 Btw, there are some other RuntimeExceptions that could be casted to IllegalArgumentException in WhereParser#parseWhere. Wdyt?

restapi/src/main/java/io/stargate/web/service/WhereParser.java
```diff
@@ -91,15 +92,20 @@
...
                 String.format(
                     "Value entry for field %s, operation %s must be an object "
                         + "with two fields 'key' and 'value'.",
                     fieldName, rawOp));
           }
-          Column.ColumnType mapType = tableData.column(fieldName).type();
+          Column column = tableData.column(fieldName);
+          if (column == null) {
+            throw new IllegalArgumentException(
+                String.format("Unknown column name '%s'.", fieldName));
```
dimas-b, line 104: `Unknown field name` like in other similar places?

dougwettlaufer (approved):

restapi/src/main/java/io/stargate/web/resources/RowResource.java
```diff
@@ -647,7 +647,7 @@ private Predicate getOp(Filter.Operator operator) {
   private static Object filterToValue(Object val, String name, Table tableData) {
     Column column = tableData.column(name);
     if (column == null) {
-      throw new IllegalArgumentException(String.format("unknown field name '%s'", name));
+      throw new IllegalArgumentException(String.format("unknown field name '%s'.", name));
```
dimas-b, line 650: nit: in other places we have upper case first letter :shrug:

dimas-b (approved):

PR #767, opened 2021-03-16
Add Cache-Control header to auth-api responses - Fixes #762

### LINKED ISSUE #762: Return Cache-Control and Expires headers on successful REST authorization call
When making a successful authorization call to Stargate via REST, we should see the following two headers to indicate how long the authToken can be cached for.

These headers should not be present or indicate no caching in the event of a failed authorization call (whether it's internal failure or bad request).

Header docs:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires

dougwettlaufer:


auth-api/src/main/java/io/stargate/auth/api/resources/AuthResource.java
```diff
@@ -232,16 +250,50 @@ public Response createTokenFromUsername(
...
+    return cacheControl;
+  }
+
+  /**
+   * Used to create a Cache-Control header that will indicate the response should not be cached.
+   * This is intended to be used in the event of a failure to authenticate.
+   *
+   * @return A {@link CacheControl} object denoting the response should not be cached.
+   */
+  private CacheControl cacheControlNoCache() {
```
michael-baraboo-datastax, line 291: Should probably be called cacheControlNoStore to avoid being confused that this is setting the `no-cache` header that has no guarantee to prevent caching. 🙂

auth-api/src/main/java/io/stargate/auth/api/resources/AuthResource.java
```diff
@@ -232,16 +250,50 @@ public Response createTokenFromUsername(
...
+    CacheControl cacheControl = new CacheControl();
+    cacheControl.setMaxAge(0);
```
eribeiro, line 293: Instead of creating this object each time, couldn't we create as a `public static final field` and reuse it?

auth-api/src/main/java/io/stargate/auth/api/resources/AuthResource.java
eribeiro: `clearCacheControl` or `emptyCacheControl` could be alternative names?

auth-api/src/main/java/io/stargate/auth/api/resources/AuthResource.java
```diff
@@ -232,16 +252,53 @@ public Response createTokenFromUsername(
...
+   * s-maxage as defined by {@code stargate.auth_tokenttl} minus some time to avoid expiration. If
+   * the property is not set then the default value of 1790 will be used.
+   *
+   * @return A {@link CacheControl} object which is converted into a Cache-Control header with
+   *     proper directives.
+   */
+  private CacheControl cacheControlWithAge() {
+    // For a large enough TTL it's safe to knock off 10 seconds but for a smaller number reducing by
+    // that much will negatively impact caching so instead subtract only 1 second.
+    int maxAge = tokenMaxAge <= 100 ? tokenMaxAge - 1 : tokenMaxAge - 10;
```
dimas-b, line 282: nit: if `tokenMaxAge` is `1` or `0` the result is probably going to be not what we want... WDYT?

auth-api/src/main/java/io/stargate/auth/api/resources/AuthResource.java
dimas-b: the same rationale applies to `cacheControlWithAge`, I guess.

auth-api/src/main/java/io/stargate/auth/api/resources/AuthResource.java
```diff
@@ -232,16 +252,53 @@ public Response createTokenFromUsername(
...
+   * Used to create a Cache-Control header containing the proper values for the token's max-age and
+   * s-maxage as defined by {@code stargate.auth_tokenttl} minus some time to avoid expiration. If
+   * the property is not set then the default value of 1790 will be used.
+   *
+   * @return A {@link CacheControl} object which is converted into a Cache-Control header with
+   *     proper directives.
+   */
+  private CacheControl cacheControlWithAge() {
+    // For a large enough TTL it's safe to knock off 10 seconds but for a smaller number reducing by
+    // that much will negatively impact caching so instead subtract only 1 second.
```
tomekl007, line 281: maybe instead of this logic, we should validate that `tokenMaxAge` is set to some sane value? And use `- 10` always?

auth-api/src/main/java/io/stargate/auth/api/resources/AuthResource.java
```diff
@@ -232,16 +252,53 @@ public Response createTokenFromUsername(
...
+    int maxAge = tokenMaxAge <= 100 ? tokenMaxAge - 1 : tokenMaxAge - 10;
+    CacheControl cacheControl = new CacheControl();
+    cacheControl.setMaxAge(maxAge);
+    cacheControl.setSMaxAge(maxAge);
```
dimas-b, line 285: In general, caching may be tricky in this context. If the client gets a value from the cache, Stargate will _not_ update the row, so the data will not be refreshed and the TTL will apply to the original insert (assuming no on-going validation calls). Consequently, if the client attempts to validates the cached value after a period close to the TTL, the token row may already be expired, which will violate the TTL contract from the client's perspective.

I understand we need to _disable_ the caching error response. However, It is really necessary to _enable_ the caching of positive responses?

auth-api/src/main/java/io/stargate/auth/api/resources/AuthResource.java
dougwettlaufer: In either case the maxAge set in the header will be `0` and `-1` respectively which essentially mean the same thing, don't cache. If you have a token TTL of 1 second I think telling it not to cache is valid. As for `0` that's not a valid value and since we're using the same system property on the table based  side you'll have other issues if you set ttl to 0 :)

auth-api/src/main/java/io/stargate/auth/api/resources/AuthResource.java
dougwettlaufer: Since the token is on a sliding window, in all actually it will be valid _longer_ than the max age. And I believe this will be the flow on the client side


10:00: createToken call, max age = 1790
10:20: Make request requiring token, check cache if token is still valid or a new one is needed. Still valid so use cached value
10:30: token row expires (according to client's caching logic)
10:40: Client makes another request and checking the cache sees the token is "expired" and calls createToken again

Although the logic will still hold with your scenario since at 10:20 when they call createToken again they will get a fresh token from stargate with a new TTL and corresponding max-age header.

auth-api/src/main/java/io/stargate/auth/api/resources/AuthResource.java
dougwettlaufer: Well `cacheControlWithAge` is being created in the constructor so we're still ending up with a single object, right?

auth-api/src/main/java/io/stargate/auth/api/resources/AuthResource.java
dougwettlaufer: If you're using a TTL of 5 seconds for creating tokens then we have to respect that here as well but we still need to allow some padding for users to be able to generate an new token before the old expires

auth-api/src/main/java/io/stargate/auth/api/resources/AuthResource.java
dimas-b: > Although the logic will still hold with your scenario since at 10:20 when they call createToken again they will get a fresh token from stargate with a new TTL and corresponding max-age header.

Not sure I understand. Why does it have to be a new token if cache settings allow reusing the old value?

auth-api/src/main/java/io/stargate/auth/api/resources/AuthResource.java
dimas-b: Yes, I missed that it was created in the contructor :facepalm: Current code LGTM :+1:

auth-api/src/main/java/io/stargate/auth/api/resources/AuthResource.java
dimas-b: Also, my example is not about how the flow proceeds normally, but how it can go wrong and produce unexpected results.

dougwettlaufer: > Not sure I understand. Why does it have to be a new token if cache settings allow reusing the old value?

Do you mean "call" createToken but actually get the value out of the cache? If that's the case it's going to depend on how the client implements caching logic, either setting TTL at the cache level or thing-calling-cache-level. If it's at the cache level then should still be okay since I'd expect the cache to respect the max-age value. If it's at the thing-calling-cache-level then I'd hope it's recomputing the TTL otherwise that sounds like a bug waiting happen since there's a cache TTL disconnect between cache and the caller of the cache.

> Also, my example is not about how the flow proceeds normally, but how it can go wrong and produce unexpected results.

Sure but I think the best we can do is follow conventions of the Cache-Control header and hope they implement caching properly on their side. And honestly these tokens will last longer than they think due to the sliding window.

dimas-b: I do believe even a properly configured cache can lead to unexpected behaviour here, because the HTTP call to create a token may return previous values within the "maxage" timeframe without hitting the backend, hence without updating the row timestamp in storage. Validation calls come though a different API and the HTTP cache is not aware of that.

What do you think about using "no-store" for _both_ successes and failures in token-related Web APIs?

dougwettlaufer: > because the HTTP call to create a token may return previous values within the "maxage" timeframe without hitting the backend

Sure but I don't see how that's a problem. We return the max-age in order to tell the cache to evict a token once it hits that age. If there's multiple tokens within 0 and 30 minutes old that's fine because they're still valid. It's up to the cache implementation to evict tokens based on `max-age` directive before they expire. If someone wants to set up a mutli-layer cache then I'd say the onus is on them to ensure that each of their layers is following caching standards as we are here by returning the Cache-Control header. 


> Validation calls come though a different API and the HTTP cache is not aware of that.

Sure and they shouldn't care. The job of the cache is to continue serving up the token until it needs to be evicted based on the prescribed TTL.


> What do you think about using "no-store" for both successes and failures in token-related Web APIs?

Then we'd be telling users to never cache a token. By default tokens have a TTL of 30 minutes which should be a perfectly acceptable time period to cache.

dimas-b: Well, caching is different from TTL. We're telling caches not to cache, but end user know the token is valid for as long as the TTL. However, authentication-related interactions should normally not be cached. My reading of [RFC-7234](https://tools.ietf.org/html/rfc7234#section-3.2) is that it is the default standard behaviour, however, we explicitly allow caching in this case by providing cache-control headers.

dougwettlaufer: > A shared cache MUST NOT use a cached response to a request with an Authorization header field

That's referring to caching the response of an authenticated request. So for example, if someone were to call Stargate and get a row that technically shouldn't be cached because it's an authenticated request (granted we don't use the `Authorization` header but same concept). 

I'm pretty sure that section of the RFC is there to keep people honest. If I revoke your access to a resource you shouldn't have access to it anymore but you will if you've cached it. This doesn't really apply to our tokens since they're revalidated on every request which means if I revoke your token even if you have it cached it still won't do you any good since it's being revalidated.

auth-api/src/main/java/io/stargate/auth/api/resources/AuthResource.java
```diff
@@ -52,8 +53,12 @@
 
   private static final Logger logger = LoggerFactory.getLogger(AuthResource.class);
 
+  private static final int tokenMaxAge =
```
EricBorczuk, line 56: lil nit: I think this is in seconds, could you put a comment to that effect? (not a big deal)

EricBorczuk (approved):
This seems fine to me, if the ongoing discussions are settled!

dimas-b (approved):

PR #777, opened 2021-03-18
Execute DSE requests on TPC threads

dimas-b:
This is mainly for CQL tracing to work properly.

In old code queries start on epoll or I/O threads this is where
the initial tracing thread-local state is set. Requests later jump
to TPC threads and the thread-local state is transferred to the
TPC thread. However, the tracing state on the initial thread remains.

There is risk of some other request reusing the tracing state from
the initial thread and braking that tracing session.

This commit forces all processing after establishing the query state
to be done on a TPC thread, including setting up tracing sessions.

Note: TPCRunnable automatically resets ExecutorLocals data (including
tracing) before running each task.

Note: this commit also enables assertions for integration tests to
validate proper thread-local handling.


dougwettlaufer (approved):

dimas-b:
Alternative fix: #786

sbtourist (changes requested):
> Alternative fix: #786

I find the alternative fix better, as it avoids the extra thread jump and directly conveys the intent behind the fix.

dimas-b:
Closing in favour of #786

Issue #439, opened 2020-11-30
Create utility web module


olim7t:
This would be a plain JAR that factors elements common to all web modules:

* DropWizard dependencies
* the "no admin" server factory that's currently duplicated in a few modules: `GraphqlApiServerFactory`, `AuthApiServerFactory`, `RestApiServerFactory` (see [#429](https://github.com/stargate/stargate/pull/429#discussion_r530723538))
* create a utility to unify the creation of the Response with the `Error` entity so that status codes match in both, and thus fix all the places where `Error` is created without a status code.

jeffreyscarpenter:
This is rendered moot by v2 architecture.

PR #785, opened 2021-03-22
fixes #775: better status codes for missing tables in document api

### LINKED ISSUE #755: Geospacial types cleanup
This is a follow-up ticket for issues raised during the review of https://github.com/stargate/stargate/pull/750

### LINKED ISSUE #775: Better status codes for table not exist/table already exists errors
Currently, if you create a table with the same name as an existing table, a 500 is returned. In a similar vein, if you try to read from a table that doesn't exist, a 500 is returned.  These should both be 4XX errors; 400 for the former and 404 for the latter

ivansenic:
Here is the quick overview:

* So the collection creation already was returning `409` when trying to create same, I added a test and adapted the response to `Error` as per API docs
* for the document get, I had to go for schema check (one `TODO` needs to be resolved there, please advise)..
* extended the API docs as well

If there's anything I missed, please let me know..

p.s. imports got somehow automatically adapted by IntelliJ..


dimas-b:
> imports got somehow automatically adapted by IntelliJ..

I use the `GoogleStyle` plugin in IntelliJ... it does not produce exactly the same results as the formatter enforced by our Maven config, but it's close enough for day-to-day work.

restapi/src/main/java/io/stargate/web/docsapi/exception/XNotFoundException.java
```diff
@@ -0,0 +1,12 @@
+package io.stargate.web.docsapi.exception;
+
+/** Thrown when API can not find a requested resource. */
+public class XNotFoundException extends RuntimeException {
```
dimas-b, line 4: Why `X`? :thinking: What do you think about `ResourceNotFoundException`?

restapi/src/main/java/io/stargate/web/docsapi/exception/XNotFoundException.java
```diff
@@ -0,0 +1,12 @@
+package io.stargate.web.docsapi.exception;
```
dimas-b, line 1: Please add the standard copyright header, e.g. copy from `DocCollection`

In fact these headers appear to be missing in a lot of Docs API node... I'm not sure why... @dougwettlaufer the standard header is appropriate for all of the Docs API code, right?

restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
```diff
@@ -151,7 +139,10 @@ public Response createCollection(
           } else {
             return Response.status(Response.Status.CONFLICT)
                 .entity(
-                    String.format("Create failed: collection %s already exists", info.getName()))
+                    new Error(
+                        String.format(
+                            "Create failed: collection %s already exists.", info.getName()),
+                        Response.Status.CONFLICT.getStatusCode()))
```
dimas-b, line 145: Would it make sense to create a utility class to avoid accidentally using different code in the `Response` (line 140) and `Error` (line 145) objects?

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -633,6 +639,19 @@ public Response getDocPath(
             }
           }
 
+          // check first that namespace and table exist
+          AuthenticatedDB authenticatedDB = dbFactory.getDataStoreForToken(authToken, allHeaders);
+          // TODO does this require authenticateSchemaRead?
+          Keyspace keyspace = authenticatedDB.getKeyspace(namespace);
```
dimas-b, line 645: I think not because here we just validating the name. I assume an `authorizeDataRead` check is executed further down the call path. `authorizeSchemaRead` is required when we are exposing information about the schema, like listing objects of "describing", which here it is just a data read. @EricBorczuk , @dougwettlaufer please correct me if I'm wrong.

testing/src/test/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -259,6 +257,26 @@ public void testInvalidKeyspaceAndTable() throws IOException {
...
   }
 
+  @Test
+  public void testDocGetOnNotExistingNamespace() throws IOException {
+    String resp =
+        RestUtils.get(
+            authToken, hostWithPort + "/v2/namespaces/unknown/collections/collection/1", 404);
+    assertThat(resp)
+        .isEqualTo(
+            "{\"description\":\"Not found: Namespace unknown does not exist.\",\"code\":404}");
```
eribeiro, line 267:
```suggestion
    Error response = objectMapper.readValue(resp, Error.class);
    assertThat(response.getCode()).isEqualTo(HttpStatus.SC_NOT_FOUND);
    assertThat(response.getDescription()).isEqualTo("Not found: Namespace unknown does not exist.");
```

testing/src/test/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -2840,6 +2858,42 @@ public void deleteNamespace() throws IOException {
...
   }
 
+  @Test
+  public void createCollection() throws IOException {
+    String tableName = "tb_createTable_" + System.currentTimeMillis();
+
+    RestUtils.post(
+        authToken,
+        hostWithPort + "/v2/namespaces/" + keyspace + "/collections",
+        "{\"name\" : \"" + tableName + "\"}",
```
eribeiro, line 2868:
```suggestion
        String.format("%s/v2/namespaces/%s/collections", hostWithPort, keyspace),
        objectMapper.writeValueAsString(payload),
```

testing/src/test/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -2840,6 +2858,42 @@ public void deleteNamespace() throws IOException {
         HttpStatus.SC_NOT_FOUND);
   }
 
+  @Test
+  public void createCollection() throws IOException {
+    String tableName = "tb_createTable_" + System.currentTimeMillis();
+
```
eribeiro, line 2864: Map<String, String> payload = new HashMap<String, String>();
payload.put("name", tableName);

eribeiro:
@ivansenic 
> p.s. imports got somehow automatically adapted by IntelliJ..

Intellij "optimize" imports automatically. You can change this behaviour in Editor -> Code Style -> Java -> Import Tab. I've setup the "Class count to use import *" to a large value (20) but there should be a smarter way of avoiding wildcards import in this screen.

restapi/src/main/java/io/stargate/web/docsapi/exception/XNotFoundException.java
ivansenic: Done.

restapi/src/main/java/io/stargate/web/docsapi/exception/XNotFoundException.java
ivansenic: Done. But just for this new class, we should have a proper clean up task to handle other missing class and enforce a IntelliJ rule that can be shared in the repo, so that anybody working can get it directly without much hassle.

restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
ivansenic: It would, but I think this can be done in separate ticket/PR if we want, would be strange that I add it now and use it in only this place. In addition I see that in lot of places error is created without the status code, so this could be fixed as well.

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
ivansenic: OK, removing TODO for now.

ivansenic:
> @ivansenic
> 
> > p.s. imports got somehow automatically adapted by IntelliJ..
> 
> Intellij "optimize" imports automatically. You can change this behaviour in Editor -> Code Style -> Java -> Import Tab. I've setup the "Class count to use import *" to a large value (20) but there should be a smarter way of avoiding wildcards import in this screen.

This is a nice solution. However, this setting must be shared with the team. Otherwise it makes no sense, as one will have it this way and other another. Then we will constantly have a ping-pong in git changes, which is really bad imo. Any reason why default IntelliJ settings are not good enough?

testing/src/test/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
ivansenic: I would rather not do it this way and this also stands for other comments/suggestions you provided, let me explain why. So these are integration tests. From the outside there is no object mapper, the request body and responses are strings. It's very easy to have some really nice bugs hidden if you use your internal object mapper to actually produce request bodies (trust me I have been there). 

Furthermore, wrt responses I would rather keep this as string, imagine a situation where you add a new field to the `Error` class. The changes above will keep the test green, although you actually changed the response sent back. In such situation, I would like a test to fail, so that we can be informed and we can say yep this was intentional, we extended the error class and let me fix the test.

Does this make sense?

ivansenic:
Rebased on the latest master and forced pushed.

testing/src/test/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
eribeiro: >It's very easy to have some really nice bugs hidden if you use your internal object mapper to actually produce request bodies (trust me I have been there).

You have a point. I have seen this myself. 

As for the the use of Error.class this is the way we have been doing (see here: https://github.com/stargate/stargate/blob/master/testing/src/test/java/io/stargate/it/http/RestApiv2Test.java#L1494). So, for the sake of consistency it would be nice to adhere to that or refactor those other uses. If we change the Error class with a new field the tests will break if this new field is required. OTOH, if it's optional so it should not mess with other tests besides the tests that actually need this new field, ideally.

ivansenic: So you mean refactor all uses in the `BaseDocumentApiV2Test.java ` to this what you liked above? Cause in this test class it has been direct string assertion. Actually if I am up to refactor stuff, seems that some kind of json path assert is way better imo, what do you think about that?

restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
dimas-b: sgtm :+1:

dimas-b:
>  Any reason why default IntelliJ settings are not good enough?

It was decided to enforce the coding style of [com.coveo:fmt-maven-plugin](https://github.com/coveooss/fmt-maven-plugin), which does not match IntelliJ defaults :shrug: 

You can run `mvn com.coveo:fmt-maven-plugin:format` to automatically reformat all code.

restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
dougwettlaufer: We can add that to https://github.com/stargate/stargate/issues/439

ivansenic: OK will resolve this, and a note to #439.

testing/src/test/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
dimas-b: Refactoring to use JSON asserts is a good idea, I think. As far as this PR goes, I think string asserts are ok for the new functionality. I'd be +1 to do the refactoring in a dedicated (separate) PR.

dimas-b (approved):

testing/src/test/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
ivansenic: I would agree with @dimas-b, let's then merge this I guess..

testing/src/test/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
ivansenic: Agreed not to produce request bodies with the object mapper.

testing/src/test/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
ivansenic: Same here..

PR #786, opened 2021-03-22
Reset ExecutorLocals in DSE persistence requests

dimas-b:
This is mainly for CQL tracing to work properly.

In old code queries start on epoll or I/O threads and this is where
the initial tracing thread-local state is set. Requests later jump
to TPC threads and the thread-local state is transferred to the
TPC thread. However, the tracing state on the initial thread remains.

There is risk of some other request reusing the tracing state from
the initial thread and breaking that tracing session.

This commit forces DSE ExecutorLocals to be reset (including resetting
tracing state) before invoking DSE request processing code.

Note: TPCRunnable automatically resets ExecutorLocals data (including
tracing) before running each task.

Note: this commit also enables assertions for integration tests to
validate proper thread-local handling.

### DIFF
--- a/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
+++ b/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
@@ -46,6 +46,7 @@ import org.apache.cassandra.auth.CassandraRoleManager;
 import org.apache.cassandra.auth.IAuthContext;
 import org.apache.cassandra.auth.RoleResource;
 import org.apache.cassandra.auth.user.UserRolesAndPermissions;
+import org.apache.cassandra.concurrent.ExecutorLocals;
 import org.apache.cassandra.concurrent.TPCTaskType;
 import org.apache.cassandra.config.Config;
 import org.apache.cassandra.config.DatabaseDescriptor;
@@ -481,10 +482,8 @@ public class DsePersistence
         Parameters parameters, long queryStartNanoTime, Supplier<Request> requestSupplier) {
 
       try {
-        if (parameters.protocolVersion().isGreaterOrEqualTo(ProtocolVersion.V4))
-          ClientWarn.instance.captureWarnings();
-
         Single<QueryState> queryState = newQueryState();
+
         Request request = requestSupplier.get();
         if (parameters.tracingRequested()) {
           request.setTracingRequested();
@@ -517,6 +516,19 @@ public class DsePersistence
                         return result;
                       } finally {
                         ClientWarn.instance.resetWarnings();
+                      }
+                    })
+                .doOnSubscribe(
+                    disposable -> {
+                      // When running inside DSE, query tasks clear ExecutorLocals before
+                      // running, which is handled by its Message.channelRead0. In Stargate
+                      // requests may come from Epoll threads of the CQL module, from HTTP
+                      // threads or any other client. So here we ensure that DSE code starts
+                      // processing the request with a clean thread local state.
+                      ExecutorLocals.set(null);
+
+                      if (parameters.protocolVersion().isGreaterOrEqualTo(ProtocolVersion.V4)) {
+                        ClientWarn.instance.captureWarnings();
                       }
                     })
                 .subscribe(
--- a/testing/src/test/java/io/stargate/it/storage/StargateContainer.java
+++ b/testing/src/test/java/io/stargate/it/storage/StargateContainer.java
@@ -533,6 +533,7 @@ public class StargateContainer extends ExternalResource<StargateSpec, StargateCo
     @Override
     public Collection<String> commandArguments(ClusterConnectionInfo backend) {
       Collection<String> args = new ArrayList<>();
+      args.add("-ea");
       args.add("-jar");
       args.add(starterJar().getAbsolutePath());
       args.add("--cluster-seed");


persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -484,7 +485,18 @@ protected void loginInternally(io.stargate.db.AuthenticatedUser user) {
         if (parameters.protocolVersion().isGreaterOrEqualTo(ProtocolVersion.V4))
           ClientWarn.instance.captureWarnings();
 
-        Single<QueryState> queryState = newQueryState();
+        Single<QueryState> queryState =
+            newQueryState()
+                .doOnEvent(
```
sbtourist, line 490: I find calling this here very confusing, what about using `doOnSubscribe()` when firing up the request chain [here](https://github.com/stargate/stargate/pull/786/files#diff-0411e9be5edd1fdbfc193db9d75fc9c254d6030806bc90a27a19c9d45d0fb1e0R510)? That should work the same, and provide a more readable, closer to the "source", solution.

dimas-b: Yes, I was thinking about `doOnSubscribe()`... will change.

The `here` link above does not seem to work for me, sorry. Could you just name the file and line number? @sbtourist ?

dimas-b: Switched to `doOnSubscribe` and fixed the handling of warnings.

sbtourist: > The here link above does not seem to work for me, sorry. Could you just name the file and line number?

I mean where `request.execute()` is called just below.

dimas-b: Updated.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -496,7 +495,21 @@ protected void loginInternally(io.stargate.db.AuthenticatedUser user) {
         @SuppressWarnings("unused")
         Disposable unused =
             request
-                .execute(queryState, queryStartNanoTime)
+                .execute(
+                    queryState.doOnSubscribe(
```
sbtourist, line 499: This is sort of a nit, but is there any reason why you chain to `queryState`, rather than doing: 
`request.execute().map().doOnSubscribe().subscribe()`?

dimas-b: We need to reset thread locals on the _initial_ thread before DSE code starts processing the request. Since the inner RxJava steps (added by DSE) can potentially jump threads, I do not think using the last `doOnSubscribe` for thread locals is guaranteed to be always effective.

This is hypothetical, but since at Stargate level we do not control DSE steps, I thought it would be safer to reset thread locals on the `QueryState` `Single` that is the basic of further processing step.

Hypothetical scenario where this will become critical: `Request.execute()` adds .subscribeOn(TPC.ioScheduler()` to the `QueryState` `Single` (AFAIK something like this may happen in case of Auth cache expiration). In this case the `doOnSubscribe` in front of the final `.subscribe()` call will be invoked on the thread the submits the requests, but code in `QueryMessage.execute()` that deals with tracing will be invoked on the I/O thread.

dimas-b: Having thought more on this, I think going back to `doOnEvent` is preferable because it is strongly associated with the flow of the value object, while `onSubscribe` callbacks may be delivered on another thread depending on whether `subscribeOn` steps are inserted before or after it in the pipeline.

dimas-b: Ah, I think I finally saw your point @sbtourist . Yes, we can reset thread locals in `doOnSubscribe` just before `.subscribe()`. DSE code (TPCRunnable) should be able to propagate `ExecutorLocals` to other threads as needed... sorry about the confusion. I'll make this change.

sbtourist: > We need to reset thread locals on the initial thread before DSE code starts processing the request. Since the inner RxJava steps (added by DSE) can potentially jump threads, I do not think using the last doOnSubscribe for thread locals is guaranteed to be always effective.

AFAIK the `doOnSubscribe()` callback is invoked on the "source" thread, before eventually jumping to the `subscribeOn` thread; I don't have time to double check this right now, I'll get back about it later.

sbtourist (approved):
Looks good!

dougwettlaufer (approved):

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -519,6 +518,19 @@ protected void loginInternally(io.stargate.db.AuthenticatedUser user) {
...
                       }
                     })
+                .doOnSubscribe(
+                    disposable -> {
+                      // When running inside DSE, query tasks clear ExecutorLocals before
+                      // running, which is handled by its Message.channelRead0. In Stargate
+                      // requests may come from Epoll threads of the CQL module, from HTTP
+                      // threads or any other client. So here we ensure that DSE code starts
+                      // processing the request with a clean thread local state.
+                      ExecutorLocals.set(null);
```
ivansenic, line 528: OK, I don't know where and how is `ExecutorLocals` being actually used.. But I understand the problem, you actually wanna clear state one a thread that picks up the request execution I guess.. 

First problem is that there is absolutely no-guarantee that thread that does `doOnSubscribe` also does the stuff where the ThreadLocal is used. Seems that the execution is actually outside of this class, thus in the creation of the reactive sequence you never know which schedulers are used. But if it works, it could be a quick fix.

The reactive stuff and thread locals are not really friends. This is why usually you wanna handle these sort of things with reactive context. Reactive context is passed along and you always have the guarantee that you actually have correct context at the time of the execution, no matter which thread it is ending in. But of course, there has to be a "contract" between two sides and yea this old style thread local needs to go away. I know this is not so simple :)

Anyway, I did wrote a quick blog on similar matter (https://www.novatec-gmbh.de/en/blog/how-can-the-mdc-context-be-used-in-the-reactive-spring-applications/), yes it was Kotlin and different stuff, but a similar idea could be applied here as well.. Although a way nicer things, as we anyway are in control of the `ExecutorLocals`, would be to remove it completely and migrate to context. In the blog, we needed to do it, as we were depending on the 3rd party impl.

ivansenic (commented):
Maybe somebody else should provide the approval.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
dimas-b: Yes, this is tricky indeed. The propagation context data in RxJava in this case is actually handled by `TPCRunnable`, which is part of DSE code. It picks up `ExecutorLocals` from current thread and moves them to another thread in case Rx tasks switch threads.

sbtourist: @ivansenic,

> First problem is that there is absolutely no-guarantee that thread that does doOnSubscribe also does the stuff where the ThreadLocal is used.

That is correct, more specifically `doOnSubscribe()` is guaranteed to be called by the thread setting up the reactive chain, regardless of where the subscription ends up being handled, but that's what we care about in this ticket: the rest, that is the actual propagation, as mentioned by @dimas-b, is handled by the TPC framework.

PR #787, opened 2021-03-22
Quick graphql readme cleanup

dougwettlaufer:


graphqlapi/README.md
```diff
@@ -198,136 +192,45 @@ type Query {
...
+* `books()`: Query book values by equality. If no `value` argument is provided then the first 100 (
+  default pagesize) values are returned.
 
 * `booksFilter`: (**Deprecated**) Query book values by filtering the result with additional
-  operators e.g. `gt` (greater than), `lt` (less than), `in` (in a list of
-  values) etc. The `books()` equality style query is preferable if your queries
-  don't require the use non-equality operators.
+  operators e.g. `gt` (greater than), `lt` (less than), `in` (in a list of values) etc.
+  The `books()` equality style query is preferable if your queries don't require the use
+  non-equality operators.
```
olim7t, line 209:
```suggestion
  of non-equality operators.
```

graphqlapi/README.md
```diff
@@ -15,46 +16,39 @@ To build the GraphQL plugin individually, run the command:
...
 
 ### Using the Playground
 
-The easiest way to get started is to use the built-in GraphQL playground. GraphQL
-playground can be downloaded as a standalone app [here](https://github.com/prisma-labs/graphql-playground)
-and can then be accessed by going to http://localhost:8080/graphql/{keyspace}. 
-Once in the playground, you can create new schema and interact with the GraphQL APIs.
+The easiest way to get started is to use the built-in GraphQL playground and can then be accessed by
+going to <http://localhost:8080/graphql/{keyspace}>. Once in the playground, you can create new
+schema and interact with the GraphQL APIs.
```
olim7t, line 26:
```suggestion
The easiest way to get started is to use the built-in GraphQL playground at http://localhost:8080/playground.
You can then interact with the various GraphQL schemas by entering their URL in the playground's address bar, for example http://localhost:8080/graphql/{keyspace}.
```

graphqlapi/README.md
```diff
@@ -15,46 +16,39 @@ To build the GraphQL plugin individually, run the command:
...
+  fields.
 
-For each keyspace created in your Cassandra schema, a new path is created under
-the root `graphql-path` (default is: `/graphql`). For example, a path
-`/graphql/library` is created for the `library` keyspace when it is added to
-the Cassandra schema.
+#### Keyspaces
 
+For each keyspace created in your Cassandra schema, a new path is created under the
+root `graphql-path` (default is: `/graphql`). For example, a path
```
olim7t, line 41: This seems to imply that the `/graphql` prefix is configurable; currently it isn't.

olim7t (approved):
A few minor remarks.

But frankly I'm wondering if it makes sense to keep that documentation here, it is redundant with stargate.io.

dougwettlaufer:
> But frankly I'm wondering if it makes sense to keep that documentation at all, it is redundant with stargate.io.

@olim7t I had the same thought as well. I initially started this PR with the intent of removing the majority of this but kept it since this seems to go more in depth. This readme seems to be more of a technical "what" and the docs are more of a "how".


@polandll what are your thoughts?

polandll:
I can go through this and incorporate anything that the docs are missing, and then this doc can be removed. Seems maybe the best way to do it. That way, you all don't have to keep maintaining.

PR #789, opened 2021-03-23
change CL of StorageHealthChecker queries to LOCAL_QUORUM

tomekl007:


health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
```diff
@@ -92,7 +93,8 @@ protected Result check() throws Exception {
                   .ttl(INSERT_TTL_SECONDS)
                   .timestamp(writeTimestamp.toEpochMilli())
                   .build()
-                  .bind())
+                  .bind(),
+              ConsistencyLevel.LOCAL_QUORUM)
```
dougwettlaufer, line 97: I thought we wanted to write at `LOCAL_ONE` and read at `LOCAL_QUORUM`

tomekl007: why not read and write at `LOCAL_QUORUM`?

dimas-b: Yes, I made a mistake in my message above :facepalm: `LOCAL_QUORUM` is the right CL for both reads and writes.

dimas-b (approved):

dougwettlaufer (approved):

Issue #579, opened 2021-01-06
Add "CREATE TYPE" to REST API for creating UDTs


polandll:
Currently, UDTs must be created in CQL, then can be queried in REST.

eribeiro:
Copying and pasting @clun's description.

> To amend the schema definition with REST API, it would be required to create a UserDefinedType resource
> With the following URL:
> `/v2/schemas/keyspaces/{keyspaceName}/udts`
> With the following operations (update is not allowed on a UDT)
```
- getAllUdts()
- getOneUdt(String udtName)
- createUdt(UdtAdd u)
- deleteUdt(String udtName)
```

clun:
Update for @dougwettlaufer it would be more relevant to use `types`in the resources URL and method names.

dougwettlaufer:
Yeah, @clun and I were chatting about this and I suggest `/types` as its a little more generic and hopefully for those with less familiarity with Cassandra it's easier to easier to understand then `/udts`

eribeiro:
I have just opened a PR to support UDT operations via REST. Following the discussion here I updated the endpoint to be /types. The PR is an early cut, so there will be some round of reviews, more test cases, design changes, bug fixes, etc. 

As of now, it's working as below:

`GET /v2/schemas/keyspaces/{keyspaceName}/types`  - get all UDTs
`GET /v2/schemas/keyspaces/{keyspaceName}/types/{typeName}` - get a given UDT
`POST /v2/schemas/keyspaces/{keyspaceName}/types/{typeName}` + JSON - creates an UDT
`DELETE /v2/schemas/keyspaces/{keyspaceName}/types/{typeName}` - delete an UDT

The info shown for each UDT when GETting is crappy, so I need more guidance on what to present in the JSON returned, besides other design opinions.

polandll:
Any movement on this issue? I'm interested.

dougwettlaufer:
@polandll there's a PR up here https://github.com/stargate/stargate/pull/799

PR #798, opened 2021-03-25
Add UDT creation/drop/retrieval via REST API

eribeiro:
Fixes https://github.com/stargate/stargate/issues/579 by adding support to UDT creation/drop/retrieval.


eribeiro:
This is an early cut for the support of UDT support in REST endpoint. Still need more tests, and some questions lingering, but hopefully enough to allow a initial review.

authnz/src/main/java/io/stargate/auth/entity/ResourceKind.java
```diff
@@ -23,5 +23,6 @@
   TRIGGER,
   AGGREGATE,
   VIEW,
-  INDEX
+  INDEX,
+  UDT
```
dimas-b, line 27: We have `TYPE` on line 22... How is `UDT` different from the auth perspective?

dimas-b: I mean that we do not currently distinguish UDTs in CQL auth... If we want to make it a distinct auth resource kinds, we'd probably have to fix CQL auth too (`StargateQueryHandler`).

dougwettlaufer: +1 for using `TYPE`

restapi/src/main/java/io/stargate/web/resources/Converters.java
```diff
@@ -1031,4 +1036,129 @@ public static String maybeQuote(String text) {
...
+        return Column.Type.Float;
+      case LIST:
+        if (info == null) {
+          throw new IllegalArgumentException(
+              "List cqlType should contain an 'info' field specifying the sub cqlType");
+        }
+        if (subTypes == null || subTypes.size() != 1) {
+          throw new IllegalArgumentException("List sub types should contain 1 item");
+        }
+        return Column.Type.List.of(decodeType(subTypes.get(0))).frozen(frozen);
```
dimas-b, line 1121: Technically, in `List.of(Int)` for example, `Int` is not a sub-type. JRE calls them "Type Parameter".

I personally find the term "sub-type" confusing in this context. I do not insist on "type parameter" but could related code be refactored to avoid "sub-type" in this context? thx!

eribeiro:
Both listAllUdts and getUdt return a Map as below:

``{"your_type":{"birthdate":"Column{name=birthdate, type=Date, kind=Regular, keyspace=null, table=null, order=null}"},"my_type":{"firstName":"Column{name=firstName, type=Varchar, kind=Regular, keyspace=null, table=null, order=null}"}}``

I am not happy with this output so I'd like to hear from you what could be a good output format and tips on how to accomplish that, if possible. 🙂

clun:
I think I told the team I was working on this. Will do a PR as well let's see

restapi/src/main/java/io/stargate/web/models/UdtAdd.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.validation.constraints.NotNull;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class UdtAdd {
+  private boolean ifNotExists;
+  private @NotNull List<UdtType> fields;
+
+  @ApiModelProperty(
+      value =
+          "Determines whether to create a new udt if an udt with the same name exists. Attempting to create an existing udt returns an error unless this option is true.")
```
dougwettlaufer, line 32: Might get confusing in the swagger if we talk about `udt` when the path is `/types`
```suggestion
          "Determines whether to create a new type if a type with the same name exists. Attempting to create an existing type returns an error unless this option is true.")
```

restapi/src/main/java/io/stargate/web/models/UdtAdd.java
```diff
@@ -0,0 +1,58 @@
...
+  public boolean getIfNotExists() {
+    return ifNotExists;
+  }
+
+  public void setIfNotExists(boolean ifNotExists) {
+    this.ifNotExists = ifNotExists;
+  }
+
+  @ApiModelProperty(value = "The fields defined for this type.")
```
dougwettlaufer, line 41: Should this be required?

restapi/src/main/java/io/stargate/web/models/udt/UdtInfo.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.List;
+import javax.validation.constraints.NotNull;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class UdtInfo {
+  private String name;
+  private boolean frozen;
+  private @NotNull List<UdtType> typeParams;
+
+  @ApiModelProperty(value = "return the name")
```
dougwettlaufer, line 31:
```suggestion
  @ApiModelProperty(value = "The name of the type")
```

restapi/src/main/java/io/stargate/web/models/udt/UdtInfo.java
```diff
@@ -0,0 +1,68 @@
...
+  public String getName() {
+    return name;
+  }
+
+  public void setName(String name) {
+    this.name = name;
+  }
+
+  @ApiModelProperty(value = "return true if is frozen")
```
dougwettlaufer, line 40: For this I'd suggest something like
```java
@ApiModelProperty(value = "Denotes whether or not the type should be defined as frozen. Using frozen means...")
```

restapi/src/main/java/io/stargate/web/models/udt/UdtInfo.java
```diff
@@ -0,0 +1,68 @@
...
+  public boolean isFrozen() {
+    return frozen;
+  }
+
+  public void setFrozen(boolean frozen) {
+    this.frozen = frozen;
+  }
+
+  @JsonProperty("typeParams")
+  @ApiModelProperty(value = "return type parameters")
```
dougwettlaufer, line 50:
```java
@ApiModelProperty(value = "The CQL type of the parameters being used (e.g. text, int, date)")
```

dougwettlaufer: Or something to that effect.

restapi/src/main/java/io/stargate/web/models/udt/UdtType.java
```diff
@@ -0,0 +1,68 @@
...
+import io.swagger.annotations.ApiModelProperty;
+import javax.validation.constraints.NotNull;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class UdtType {
+  private String name;
+  private @NotNull CQLType basic;
+  private UdtInfo info;
+
+  @ApiModelProperty(value = "return the field name")
```
dougwettlaufer, line 30: Same comments for these since they'll show up in the swagger endpoint.

restapi/src/main/java/io/stargate/web/models/UdtAdd.java
```diff
@@ -0,0 +1,58 @@
...
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.google.common.base.MoreObjects;
+import io.stargate.web.models.udt.UdtType;
+import io.swagger.annotations.ApiModelProperty;
+import java.util.List;
+import javax.validation.constraints.NotNull;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class UdtAdd {
```
dougwettlaufer, line 26: Should we call this `TypeAdd` for consistency?

restapi/src/main/java/io/stargate/web/resources/v2/schemas/UdtsResource.java
```diff
@@ -0,0 +1,380 @@
...
+@Produces(MediaType.APPLICATION_JSON)
+public class UdtsResource {
+
+  @Inject private Db db;
+
+  @Timed
+  @GET
+  @ApiOperation(
+      value = "Get all UDTs",
+      notes = "Get all UDTs from a given keyspace",
```
dougwettlaufer, line 76:
```suggestion
      value = "Get all Types",
      notes = "Get all Types from a given keyspace",
```

restapi/src/main/java/io/stargate/web/resources/v2/schemas/UdtsResource.java
```diff
@@ -0,0 +1,380 @@
...
+      response = GetResponseWrapper.class,
+      responseContainer = "List")
+  @ApiResponses(
+      value = {
+        @ApiResponse(code = 200, message = "OK", response = GetResponseWrapper.class),
+        @ApiResponse(code = 400, message = "Bad Request", response = Error.class),
```
dougwettlaufer, line 82: Looks like this can also return a 404

restapi/src/main/java/io/stargate/web/resources/Converters.java
eribeiro: Please, see if the changes I've done suffice. 🙂

restapi/src/main/java/io/stargate/web/models/UdtAdd.java
eribeiro: Definitely. 🙂

dougwettlaufer: Got it, then you'll want to add `required = true` to the annotation so swagger marks it as such.

restapi/src/main/java/io/stargate/web/resources/Converters.java
```diff
@@ -1031,4 +1036,126 @@ public static String maybeQuote(String text) {
       return text;
     return '"' + PATTERN_DOUBLE_QUOTE.matcher(text).replaceAll(ESCAPED_DOUBLE_QUOTE) + '"';
   }
+
+  public static Map<String, String> convertColumnMap(UserDefinedType udt) {
+    return udt.columnMap().entrySet().stream()
```
dougwettlaufer, line 1041: Is this safe or do we need a null check on `udt` beforehand?

eribeiro: Good catch.

restapi/src/main/java/io/stargate/web/resources/v2/schemas/UdtsResource.java
```diff
@@ -0,0 +1,380 @@
...
+            code = 201,
+            message = "resource created",
+            response = Map.class,
+            responseContainer = "Map"),
+        @ApiResponse(code = 400, message = "Bad Request", response = Error.class),
+        @ApiResponse(code = 401, message = "Unauthorized", response = Error.class),
+        @ApiResponse(code = 409, message = "Conflict", response = Error.class),
+        @ApiResponse(code = 500, message = "Internal server error", response = Error.class)
+      })
+  @Path("/{typeName}")
```
dougwettlaufer, line 302: Since this is a POST it would make more sense to have the name in the body. This is implying that a resource already exists at `/typeName`

restapi/src/main/java/io/stargate/web/models/udt/UdtInfo.java
dougwettlaufer: I'm guessing this also needs `required = true`?

restapi/src/main/java/io/stargate/web/models/udt/UdtInfo.java
```diff
@@ -0,0 +1,68 @@
...
+import com.google.common.base.MoreObjects;
+import io.swagger.annotations.ApiModelProperty;
+import java.util.List;
+import javax.validation.constraints.NotNull;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class UdtInfo {
+  private String name;
+  private boolean frozen;
+  private @NotNull List<UdtType> typeParams;
```
dougwettlaufer, line 29: I'm a little confused. UdtAdd has a list of UdtTypes called `fields` and each UdtType has an `info` of UdtInfo. Then each UdtInfo also has a list of UdtTypes called `typeParams`?

dougwettlaufer: Okay, just read through `decodeType()` and I think I understand now. It might be easier to follow if we call this `subType` or something like that.

restapi/src/main/java/io/stargate/web/models/udt/UdtInfo.java
dougwettlaufer: I'm amending my original suggestion after reading further. So this here is actually the sub-type which we should definitely call out in the swagger. 

```java
@ApiModelProperty(value = "The CQL sub-type of the parent type. This is to be used when the parent type is a collection (list, tuple, map, or set) or a UDT.")
```

restapi/src/main/java/io/stargate/web/models/udt/UdtType.java
```diff
@@ -0,0 +1,68 @@
...
+  public String getName() {
+    return name;
+  }
+
+  public void setName(String name) {
+    this.name = name;
+  }
+
+  @JsonProperty("basic")
+  @ApiModelProperty(value = "return the field type.")
```
dougwettlaufer, line 40:
```suggestion
  @ApiModelProperty(value = "The basic CQL type to be used for the field. If this type is a collection or User Defined Type then an info will also need to be provided for its type. An example of this would be a list of ints.")
```

restapi/src/main/java/io/stargate/web/resources/Converters.java
```diff
@@ -1031,4 +1036,126 @@ public static String maybeQuote(String text) {
...
+          throw new IllegalArgumentException(
+              "UDT cqlType should contain an 'info' field specifying the UDT name");
+        }
+        return UserDefinedType.reference(udtType.getName()).frozen(frozen);
+      case TUPLE:
+        if (info == null) {
+          throw new IllegalArgumentException(
+              "TUPLE cqlType should contain an 'info' field specifying the sub types");
+        }
+        if (typeParams == null || typeParams.isEmpty()) {
```
dougwettlaufer, line 1149: Why is this checking `isEmpty` but above it's `typeParams.size() != 1`?

restapi/src/main/java/io/stargate/web/resources/v2/schemas/UdtsResource.java
```diff
@@ -0,0 +1,380 @@
...
+  @GET
+  @ApiOperation(
+      value = "Get an UDT by name",
+      notes = "Get an UDT by name from a given keyspace",
+      response = GetResponseWrapper.class,
+      responseContainer = "Map")
+  @ApiResponses(
+      value = {
+        @ApiResponse(code = 200, message = "OK", response = GetResponseWrapper.class),
+        @ApiResponse(code = 400, message = "Bad Request", response = Error.class),
```
dougwettlaufer, line 141: 404 as well?

restapi/src/main/java/io/stargate/web/resources/Converters.java
```diff
@@ -1031,4 +1036,126 @@ public static String maybeQuote(String text) {
...
+          throw new IllegalArgumentException("TUPLE cqlType should have at least one sub cqlType");
+        }
+        Column.ColumnType[] decodedTypeParams =
+            typeParams.stream()
+                .map(Converters::decodeType)
+                .collect(Collectors.toList())
+                .toArray(new Column.ColumnType[0]);
+        return Column.Type.Tuple.of(decodedTypeParams);
+    }
+    throw new RuntimeException(String.format("Data cqlType %s is not supported", basic));
```
dougwettlaufer, line 1159: IllegalArgumentException as well here? This should return a 400 rather than a 500.

restapi/src/main/java/io/stargate/web/models/udt/UdtInfo.java
eribeiro: I have renamed to subType following the review comment by @dimas-b :)

restapi/src/main/java/io/stargate/web/resources/Converters.java
eribeiro: Because tuple may have many sub types, but list, for example, can only have one subtype.

restapi/src/main/java/io/stargate/web/models/udt/UdtInfo.java
eribeiro: > I'm guessing this also needs required = true?
IIRC, in the case of the Tuple it doesn't need a sub-type.

restapi/src/main/java/io/stargate/web/resources/v2/schemas/UdtsResource.java
```diff
@@ -0,0 +1,380 @@
...
+  public Response listAllUdts(
+      @ApiParam(
+              value =
+                  "The token returned from the authorization endpoint. Use this token in each request.",
+              required = true)
+          @HeaderParam("X-Cassandra-Token")
+          String token,
+      @ApiParam(value = "Name of the keyspace to use for the request.", required = true)
+          @PathParam("keyspaceName")
+          final String keyspaceName,
```
dougwettlaufer, line 95: Can we also add the `raw` param to all of these to be consistent with the other endpoints?

dougwettlaufer:
@clun can you and @eribeiro please work together on which of your PRs are going in? If you'd like I can create a branch with some common code for you two to both work off.

eribeiro:
Hey, my bad @dougwettlaufer @clun. I should have requested the assignment of this ticket to me before working on it.  

Well, I can review @clun's PR or work on a common branch. Either approach is fine to me, whatever is easier. 🙂

PR #802, opened 2021-03-25
Do not fail health check when schema is in agreement with storage nodes

dimas-b:


persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -309,6 +310,25 @@ public boolean isInSchemaAgreement() {
         <= 1;
   }
 
+  @Override
+  public boolean isInSchemaAgreementWithStorage() {
```
ivansenic, line 314: No extra tests for these new `isInSchemaAgreementWithStorage` implementations?

ivansenic (approved):
Seems like exactly what we need..

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -309,6 +310,25 @@ public boolean isInSchemaAgreement() {
...
+    InetAddress localAddress = FBUtilities.getBroadcastAddress();
+    return Gossiper.instance.getLiveMembers().stream()
+            .filter(
+                ep -> {
+                  EndpointState epState = Gossiper.instance.getEndpointStateForEndpoint(ep);
+                  return epState != null
+                      && !Gossiper.instance.isDeadState(epState)
+                      && (!Gossiper.instance.isGossipOnlyMember(ep) || localAddress.equals(ep));
+                })
+            .map(Gossiper.instance::getSchemaVersion)
```
tomekl007, line 326: Why we need this map stage? As far as I can see, the returned `schemaVersion` value is not checked anywhere.
There is only a check on `size <= 1;`.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
tomekl007: +1 on adding tests

tomekl007: +1 on adding test(s)

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
dimas-b: We collect schema UUIDs in a set in the following step to remove duplicates. The size check ensures at most one UUID remains.

dimas-b: We could refactor this to use some other technique to achieve the same result. I just followed the pattern of other similar methods that check schema versions.

persistence-api/src/main/java/io/stargate/db/Persistence.java
```diff
@@ -86,6 +86,11 @@
 
   boolean isInSchemaAgreement();
 
+  /** Returns <code>true</code>> if the local schema agrees with all storage nodes. */
```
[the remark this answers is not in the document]
dimas-b: You're right. I'll remove the default impl.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
tomekl007: >We collect schema UUIDs in a set in the following step to remove duplicates. The size check ensures at most one UUID remains.
It seems implicit, I am +1 on refactoring that if that's not a big effort.

dimas-b: refactored

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
dimas-b: tests added

dougwettlaufer (approved):

PR #806, opened 2021-03-29
Remove dynamic OSGi imports

### LINKED ISSUE #730: Consider removing dynamic OSGi imports
Stargate uses `<DynamicImport-Package>*</DynamicImport-Package>` in many OSGi bundle manifests.

In my local testing I hit some cases where classes are loaded from other bundles without an explicit declaration, e.g. in second level (and deeper) library dependencies. This could lead to unpredictable behaviour changes when dependency versions change in one of the bundles, but not in others.

It might be preferable to avoid dynamic OSGi imports and explicitly declare all packages are expected to be imported.

dimas-b:
Also clean-up OSGi import/export declarations a bit

Fixes #730

### DIFF
--- a/auth-jwt-service/pom.xml
+++ b/auth-jwt-service/pom.xml
@@ -83,9 +83,17 @@
             <Bundle-Description>AuthJWTService component for project</Bundle-Description>
             <Bundle-SymbolicName>io.stargate.auth.jwt</Bundle-SymbolicName>
             <Bundle-Activator>io.stargate.auth.jwt.AuthJWTServiceActivator</Bundle-Activator>
-            <Import-Package>io.stargate.auth, org.osgi.framework</Import-Package>
+            <Import-Package><![CDATA[
+              org.osgi.framework,
+              io.stargate.core.*,
+              io.stargate.auth,
+              io.stargate.db.*,
+              org.apache.cassandra.stargate,
+              org.apache.cassandra.stargate.*,
+              org.javatuples,
+              com.datastax.oss.driver.shaded.guava.*,
+            ]]></Import-Package>
             <Export-Package>!*</Export-Package>
-            <DynamicImport-Package>*</DynamicImport-Package>
             <Embed-Dependency>*;scope=compile|runtime</Embed-Dependency>
           </instructions>
         </configuration>
--- a/persistence-cassandra-3.11/pom.xml
+++ b/persistence-cassandra-3.11/pom.xml
@@ -202,20 +202,22 @@
           <instructions>
             <Bundle-Name>Cassandra-Persistence</Bundle-Name>
             <Bundle-Description>Cassandra DB component for project</Bundle-Description>
-            <Bundle-SymbolicName>io.stargate.db.cassandra</Bundle-SymbolicName>
+            <Bundle-SymbolicName>io.stargate.db.cassandra_3_11</Bundle-SymbolicName>
             <Bundle-Activator>io.stargate.db.cassandra.CassandraPersistenceActivator</Bundle-Activator>
             <Import-Package><![CDATA[
                             org.osgi.framework,
+                            io.stargate.core.*,
+                            io.stargate.auth.*,
                             io.stargate.db,
-                            io.stargate.db.datastore,
-                            io.stargate.db.datastore.*,
+                            io.stargate.db.*,
                             org.apache.cassandra.stargate,
+                            org.apache.cassandra.stargate.*,
+                            org.javatuples,
                             !com.codahale.metrics.graphite.*,
-                            io.stargate.core,
                             com.codahale.metrics,
+                            com.datastax.oss.driver.shaded.guava.*,
                         ]]></Import-Package>
-            <Export-Package>io.stargate.db.cassandra</Export-Package>
-            <DynamicImport-Package>*</DynamicImport-Package>
+            <Export-Package>!*</Export-Package>
             <Bundle-ClassPath>.,sjk-core-0.14.jar,{maven-dependencies}</Bundle-ClassPath>
             <Embed-Dependency>*;scope=compile|runtime</Embed-Dependency>
             <Embed-Transitive>true</Embed-Transitive>
--- a/rate-limiting-global/pom.xml
+++ b/rate-limiting-global/pom.xml
@@ -36,10 +36,11 @@
             <Bundle-Activator>io.stargate.db.limiter.global.GlobalRateLimitingActivator</Bundle-Activator>
             <Import-Package><![CDATA[
               org.osgi.framework,
-              io.stargate.db
+              io.stargate.core.*,
+              io.stargate.db,
+              io.stargate.db.*,
             ]]></Import-Package>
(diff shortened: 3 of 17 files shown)


auth-jwt-service/pom.xml
```diff
@@ -83,9 +83,17 @@
...
-            <Import-Package>io.stargate.auth, org.osgi.framework</Import-Package>
+            <Import-Package><![CDATA[
+              org.osgi.framework,
+              io.stargate.core.*,
+              io.stargate.auth,
+              io.stargate.db.*,
+              org.apache.cassandra.stargate,
+              org.apache.cassandra.stargate.*,
+              org.javatuples,
+              com.datastax.oss.driver.shaded.guava.*,
```
tomekl007, line 94: importing a shaded dependency seems a bit odd. Can't we add an explicit dependency in the core module and import it here (instead of the shaded version)?

persistence-cassandra-3.11/pom.xml
```diff
@@ -202,20 +202,22 @@
           <instructions>
             <Bundle-Name>Cassandra-Persistence</Bundle-Name>
             <Bundle-Description>Cassandra DB component for project</Bundle-Description>
-            <Bundle-SymbolicName>io.stargate.db.cassandra</Bundle-SymbolicName>
+            <Bundle-SymbolicName>io.stargate.db.cassandra_3_11</Bundle-SymbolicName>
```
tomekl007, line 205: doesn't it impact our clients? (i.e. CNDB)

stargate-starter/src/main/java/io/stargate/starter/Starter.java
```diff
@@ -449,7 +449,19 @@ public void start() throws BundleException {
         System.getProperty("felix.log.level", String.valueOf(Logger.LOG_WARNING)));
     configMap.put(
         FelixConstants.FRAMEWORK_SYSTEMPACKAGES_EXTRA,
-        "sun.misc,sun.nio.ch,com.sun.management,sun.rmi.registry");
+        String.join(
+            ",", "sun.misc,sun.nio.ch", "com.sun.management", "sun.rmi.registry", "org.osgi.*"));
+    configMap.put(
+        FelixConstants.FRAMEWORK_BOOTDELEGATION,
```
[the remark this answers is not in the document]
ivansenic: Seems it s telling to look for these classes in the boot class loader, is this really needed? I mean class loading works in the parent-first, so these will anyway first be looked in the boot.. What's the reason to explicitly specify this?

auth-jwt-service/pom.xml
```diff
@@ -83,9 +83,17 @@
...
+            ]]></Import-Package>
             <Export-Package>!*</Export-Package>
```
ivansenic, line 96: should the export for all auth packages be as for the `authnz`?

`<Export-Package>io.stargate.auth,io.stargate.auth.entity</Export-Package>`

What this means, export all right?

ivansenic (approved):
I managed to build and run this fine and for me this is good enough knowing how complicated playing with OSGi is.. I guess that at the point when requests start flying in we load all the classes and that we would see some issues if these chnages would not be OK..

auth-jwt-service/pom.xml
mpenick: Maybe we're exposing types from that shaded guava in shared APIs?

mpenick:
Did you see any improvement in startup times?

stargate-starter/src/main/java/io/stargate/starter/Starter.java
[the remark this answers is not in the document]
mpenick: +1 on the reason for this.

auth-jwt-service/pom.xml
dimas-b: The basic problem is that the shaded guava classes are actually used. The import just makes it apparent. Before this PR, these classes were imported implicitly.

dimas-b: I'm +1 to using original guava artifacts, but it's probably beyond the scope of this PR.

persistence-cassandra-3.11/pom.xml
dimas-b: AFAIK, this ID is internal to OSGi. It does not affect maven dependencies.

I changed it to make it easier to distinguish 3.11 and 4.0 persistence in logs.

stargate-starter/src/main/java/io/stargate/starter/Starter.java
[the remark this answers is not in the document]
dimas-b: AFAIK, OSGi does not automatically delegate to the boot class loader (i.e. JRE). When we had dynamic imports it was happening because of that, as far as I understand, but now that we do not use dynamic imports, we have to declare what we want to be available from the JRE.

dimas-b:
> Did you see any improvement in startup times?

I did not measure :) The main purpose of this change is to improve functional isolation of classes. Any perf. improvement is a bonus, but was not a goal here :)

auth-jwt-service/pom.xml
ivansenic: OK export nothing is even better,

persistence-cassandra-3.11/pom.xml
```diff
@@ -202,20 +202,22 @@
...
+                            io.stargate.db.*,
                             org.apache.cassandra.stargate,
+                            org.apache.cassandra.stargate.*,
+                            org.javatuples,
                             !com.codahale.metrics.graphite.*,
-                            io.stargate.core,
                             com.codahale.metrics,
+                            com.datastax.oss.driver.shaded.guava.*,
                         ]]></Import-Package>
-            <Export-Package>io.stargate.db.cassandra</Export-Package>
```
dougwettlaufer, line 217: We no longer need the export?

rate-limiting-global/pom.xml
```diff
@@ -36,10 +36,11 @@
             <Bundle-Activator>io.stargate.db.limiter.global.GlobalRateLimitingActivator</Bundle-Activator>
             <Import-Package><![CDATA[
               org.osgi.framework,
-              io.stargate.db
+              io.stargate.core.*,
+              io.stargate.db,
+              io.stargate.db.*,
             ]]></Import-Package>
-            <Export-Package>io.stargate.db.limiter.global</Export-Package>
```
dougwettlaufer, line 41: Doesn't this need to be exported in the event someone wants to use it?

dougwettlaufer (approved):

rate-limiting-global/pom.xml
dimas-b: The linkage should be done via the proper API, I think. That is `RateLimitingManager` exported by `persistence-api`.

Linking to an implementation class in a bundle is probably not a good idea. If we want to share implementaiton code, I believe it would be preferable to publish it as a plain jar and have it _embedded_ in the bundle that needs to use it.

auth-jwt-service/pom.xml
dimas-b: opened #840

PR #811, opened 2021-03-29
Revisit PK conditions handling

olim7t:
* require operation arguments to have the same name as their
  corresponding fields (we won't be able to rely on order when we later
  add regular columns).
* use common code for all operations.


olim7t:
Merged to #634 locally.

PR #812, opened 2021-03-30
Merge /graphqlv2 into /graphql

olim7t:
Both CQL-first and GraphQL-first are now exposed under the same path:
* if the user never deployed a custom schema for a keyspace, the legacy CQL-first schema is served
* otherwise, it's the GraphQL-first schema.

Switching to GraphQL-first is an "all-or-nothing" decision: once a custom schema is installed, the legacy schema is not accessible anymore. Currently there is no way to switch back, but I'll add an `undeploy` operation in a follow-up PR.

The admin services have moved to `/graphql-admin` and `/graphql-files`.


graphqlapi/src/main/java/io/stargate/graphql/web/resources/DmlResource.java
```diff
@@ -169,17 +164,29 @@ private GraphQL getGraphl(
...
-    if (graphql == null) {
+    try {
+      GraphQL graphql =
+          graphqlCache.getDml(
+              keyspaceName,
+              (AuthenticationSubject) httpRequest.getAttribute(AuthenticationFilter.SUBJECT_KEY),
+              RequestToHeadersMapper.getAllHeaders(httpRequest));
+      if (graphql == null) {
+        replyWithGraphqlError(
+            Status.NOT_FOUND, String.format("Unknown keyspace '%s'", keyspaceName), asyncResponse);
```
olim7t, line 175: Now that the resources are merged, and the term "keyspace" appears in a GraphQL-first context, I'm tempted to drop the whole "namespace" concept altogether. Users would create their keyspaces via the DDL service, and we can just use "keyspace" everywhere.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/DmlResource.java
```diff
@@ -59,7 +68,7 @@ public void get(
       @Context HttpServletRequest httpRequest,
       @Suspended AsyncResponse asyncResponse) {
 
-    GraphQL graphql = getDefaultGraphql(httpRequest, asyncResponse);
+    GraphQL graphql = getGraphql(graphqlCache.getDefaultKeyspaceName(), httpRequest, asyncResponse);
```
tomekl007, line 71: what will `getDefaultKeyspaceName()` return in this context?

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
```diff
@@ -80,33 +90,54 @@ public GraphQL getDdl() {
...
         DataStoreOptions.builder().putAllCustomProperties(headers).build();
-    DataStore dataStore =
-        subject == null
-            ? dataStoreFactory.createInternal(dataStoreOptions)
-            : dataStoreFactory.create(subject.asUser(), dataStoreOptions);
-    Keyspace keyspace = dataStore.schema().keyspace(keyspaceName);
-    return (keyspace == null)
-        ? null
-        : newGraphql(SchemaFactory.newDmlSchema(authorizationService, keyspace, dataStoreFactory));
+    return subject == null
```
tomekl007, line 138: when the `subject == null` can be `null`?

tomekl007:
Is this change backward compatible? (re: `cql-first` API)

graphqlapi/src/main/java/io/stargate/graphql/web/resources/DmlResource.java
ivansenic: Yes this is now very confusing imo.. Especially that the code switches back and forth between keyspace and namespace (like `SchemaSourceDao` uses namespace everywhere)..

ivansenic (commented):
I did not found anything suspicious here, although I did not test it locally..

graphqlapi/src/main/java/io/stargate/graphql/web/resources/DmlResource.java
olim7t: It will be `null` if the default keyspace feature is disabled. So I need to check and return a 404 like in the previous version.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
olim7t: Right, that was when there was a `getDefaultDml`. I'll  remove this.

olim7t:
> Is this change backward compatible? (re: `cql-first` API)

The client has to explicitly deploy a GraphQL schema. Until then, the CQL-first schema will be exposed. So we are not unexpectedly breaking legacy applications.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/DmlResource.java
olim7t: I'll remove namespace operations and unify naming in a follow-up PR.

PR #816, opened 2021-04-02
Use logback for logging in Dropwizard. 

tomekl007:


auth-api/src/main/java/io/stargate/auth/api/impl/AuthApiServerFactory.java
```diff
@@ -60,7 +63,11 @@ public Server build(Environment environment) {
         Collections.singletonMap(getApplicationContextPath(), applicationHandler);
     final ContextRoutingHandler routingHandler = new ContextRoutingHandler(handlers);
     final Handler gzipHandler = buildGzipHandler(routingHandler);
-    server.setHandler(addStatsHandler(addRequestLog(server, gzipHandler, environment.getName())));
+    if (REQUEST_LOGS_ENABLED) {
+      server.setHandler(addStatsHandler(addRequestLog(server, gzipHandler, environment.getName())));
```
ivansenic, line 67: So this would disable the logging completely, even for the `500`s, maybe this is fine if we have the full stack trace but it's often quite helpful to have those in the logs imo.

auth-api/src/main/java/io/stargate/auth/api/impl/AuthApiServerFactory.java
```diff
@@ -33,6 +33,9 @@
 @JsonTypeName("auth-api")
 public class AuthApiServerFactory extends SimpleServerFactory {
 
+  private static final boolean REQUEST_LOGS_ENABLED =
+      Boolean.parseBoolean(System.getProperty("stargate.requests_logging.enabled", "true"));
```
ivansenic, line 37: Wouldn't it be better to go with some drop-wizard built-in properties, I am not sure if this is possible, but we could look into it: https://www.dropwizard.io/en/latest/manual/configuration.html#logging

ivansenic (commented):
You missed the `HealthCheckerServerFactory` this is especially noisy. Have the look on the other comments as well.

auth-api/src/main/java/io/stargate/auth/api/impl/AuthApiServerFactory.java
tomekl007: Please look at the ticket description for more details. I've tried all approaches.

auth-api/src/main/java/io/stargate/auth/api/impl/AuthApiServerFactory.java
tomekl007: I think that this is good enough for the client's feature request.

tomekl007:
> You missed the `HealthCheckerServerFactory` this is especially noisy. Have the look at the other comments as well.

The requests for `/health` are not logged. When I do:
```
cat stargate.log| grep checker
```
or
```
 cat stargate.log| grep health
```
there is almost no logs.

ivansenic:
> @ivansenic
> 
> > You missed the `HealthCheckerServerFactory` this is especially noisy. Have the look at the other comments as well.
> 
> The requests for `/health` are not logged. When I do:
> 
> ```
> cat stargate.log| grep checker
> ```
> 
> or
> 
> ```
>  cat stargate.log| grep health
> ```
> 
> there is almost no logs.

I am seeing this everywhere in the cluster logs:

```
172.18.136.181 - - [01/Apr/2021:12:35:12 +0000] "GET /checker/liveness HTTP/1.1" 200 2 "-" "kube-probe/1.15" 4
172.18.136.181 - - [01/Apr/2021:12:35:13 +0000] "GET /checker/readiness HTTP/1.1" 200 5 "-" "kube-probe/1.15" 5
```

tomekl007:
> checker

Is this generated at the Stargate side or K8s operator when it executes the request?

tomekl007:
> I am seeing this everywhere in the cluster logs:

Is this generated at the Stargate side or K8s operator when it executes the request?

ivansenic:
> > I am seeing this everywhere in the cluster logs:
> 
> Is this generated at the Stargate side or K8s operator when it executes the request?

Stargate side.. Just start with your fix and the the `checker/..` endpoints to double check. If there are not logs then fine/

auth-api/src/main/java/io/stargate/auth/api/impl/AuthApiServerFactory.java
dimas-b: Yes, I guess it's ok to disable the request log completely. We should really have a separate log message with the exception for all 500's, so having a line for the HTTP status is not critical, I think. Besides, 500's can spam the log just as well as 200's in high load tests. This new setting is intended to be used a measure against logs growing too long.

dimas-b:
+1 to adding the new option to all Web modules

auth-api/src/main/java/io/stargate/auth/api/impl/AuthApiServerFactory.java
ivansenic: @tomekl007 Have a look here for a solution: https://stackoverflow.com/questions/45484804/pass-an-empty-array-to-dropwizard-via-command-line

`-Ddw.server.requestLog.appenders=,`

tomekl007: Unfortunately, the same problem for both:
```
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize instance of `java.util.ArrayList<io.dropwizard.logging.AppenderFactory<ch.qos.logback.access.spi.IAccessEvent>>` out of VALUE_STRING token
```

tomekl007:
I've changed our DW to use log back instead of the console logger(https://www.dropwizard.io/en/latest/manual/core.html#logginguse)
It works very well. Logs are now logged with the standard levels, i.e:
```
INFO  [dw-125] 2021-04-02 17:36:05,119 Slf4jRequestLogWriter.java:62 - 0:0:0:0:0:0:0:1 - - [02/Apr/2021:15:36:05 +0000] "GET /checker/readiness HTTP/1.1" 200 5 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 11_2_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.90 Safari/537.36" 113
INFO  [dw-131] 2021-04-02 17:36:15,813 Slf4jRequestLogWriter.java:62 - 0:0:0:0:0:0:0:1 - - [02/Apr/2021:15:36:15 +0000] "GET /v2/keyspaces/k1/t HTTP/1.1" 404 421 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 11_2_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.90 Safari/537.36" 1
INFO  [dw-127] 2021-04-02 17:36:29,212 Slf4jRequestLogWriter.java:62 - 0:0:0:0:0:0:0:1 - - [02/Apr/2021:15:36:28 +0000] "GET /metrics HTTP/1.1" 200 155965 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 11_2_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.90 Safari/537.36" 585
```
and they can be suppressed by standard logback config, i.e. changing:
```
  <root level="${stargate.logging.level.root:-INFO}">
```
to
```
  <root level="${stargate.logging.level.root:-WARN}">
```
or at a lower granularity level.

dimas-b:
> I've changed our DropWizard to use log back instead of the console logge

Nice :) Could you also add a default log level spec to our `logback.xml` and a `-D` property for `Slf4jRequestLogWriter` so that it would be configure globally on Stargate start? (similar to what we have for `org.apache.cassandra`)?

EricBorczuk (approved):

graphqlapi/src/main/java/io/stargate/graphql/web/GraphqlApiServerFactory.java
```diff
@@ -61,7 +60,6 @@ public Server build(Environment environment) {
     final ContextRoutingHandler routingHandler = new ContextRoutingHandler(handlers);
     final Handler gzipHandler = buildGzipHandler(routingHandler);
     server.setHandler(addStatsHandler(addRequestLog(server, gzipHandler, environment.getName())));
-
     return server;
```
olim7t, line 63: Nit: unnecessary changes (probably a side-effect of something that was reverted).

olim7t (approved):
> I've changed our DropWizard to use log back instead of the console logger

Just to clarify (because it took me a while to understand what's going on), DropWizard uses SLF4J and Logback no matter what. The only thing the default logging factory does is hijack the configuration so that appenders can be defined in the Dropwizard config file. And indeed it sends `INFO` to the `STDOUT` appender by default.

`Slf4jRequestLogWriter` is a Jetty class. It's not clear to me why some non-Dropwizard classes' logging would be controlled by Dropwizard, while others (like are own code) aren't. But anyway by switching to `external` unify everything, the Dropwizard code still calls SLF4J, but it's now handled by our own configuration.

I think logging every request at `INFO` is a wrong, but there's no way to change it to `DEBUG`. So we should disable this logger by default in `logback.xml`, and maybe consider switching to a custom logger.

dougwettlaufer:
> I think logging every request at `INFO` is a wrong, but there's no way to change it to `DEBUG`. So we should disable this logger by default in `logback.xml`, and maybe consider switching to a custom logger.

I think what's here now is the correct approach. Logging requests at info level is fairly common practice. If someone wants to disable the logs we should provide that ability but I believe we should have the logging enabled by default.

dougwettlaufer (approved):

tomekl007:
> Nice :) Could you also add a default logger to our logback.xml and a -D property for Slf4jRequestLogWriter so that it would be configure globally on Stargate start? (similar to what we have for org.apache.cassandra)?

@dimas-b 

As far as I can see, it is already there:
```
  <logger name="io.stargate.web" level="${stargate.logging.level.web:-INFO}"/>
  <logger name="io.stargate.db" level="${stargate.logging.level.persistence:-INFO}"/>
  <logger name="io.stargate.auth" level="${stargate.logging.level.auth:-INFO}"/>
```
do you want a logger at a higher level?
```
  <logger name="io.stargate" level="${stargate.logging.level:-INFO}"/
```

dimas-b:
I mean:
```
<logger name="org.eclipse.jetty.server.Slf4jRequestLogWriter" level="${stargate.logging.level.requestlog:-INFO}"/>
```

tomekl007:
> I mean:
> 
> ```
> <logger name="org.eclipse.jetty.server.Slf4jRequestLogWriter" level="${stargate.logging.level.requestlog:-INFO}"/>
> ```

Setting the logger in this way doesn’t seem to work. I’ve tried also without package: `Slf4jRequestLogWriter` but the same outcome.

changing
```
<root level="${stargate.logging.level.root:-WARN}"> 
```
works.
It seems that the `Slf4jRequestLogWriter` inherits the logger config.

PR #824, opened 2021-04-05
Add undeploySchema operation

olim7t:


graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/UndeploySchemaFetcher.java
```diff
@@ -0,0 +1,75 @@
...
+        keyspaceName,
+        null,
+        Scope.MODIFY,
+        SourceAPI.GRAPHQL,
+        ResourceKind.KEYSPACE);
+
+    UUID expectedVersion = getExpectedVersion(environment);
+    boolean force = environment.getArgument("force");
+
+    new SchemaSourceDao(dataStore).undeploy(keyspaceName, expectedVersion, force);
```
ivansenic, line 63: So undeploy is not purging the existing schemas, is this intentional? And would this have an effect to a possible re-deploy (I guess not)?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/UndeploySchemaFetcher.java
```diff
@@ -0,0 +1,75 @@
...
+    return true;
+  }
+
+  private UUID getExpectedVersion(DataFetchingEnvironment environment) {
```
ivansenic, line 67: This also exists in the deploy schema fetcher. Does it make sense to try to have less duplicated code by creating maybe interfaces with default methods that fetchers can implement to share such things? I guess there is certain level of duplication in other fetchers as well.. So you could have there, `ISchemaFetcher` with default impl of the `getExpectedVersion`..

ivansenic (approved):
Looks good..

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/UndeploySchemaFetcher.java
olim7t: Yes, the old versions stay there, but they are all inactive, so they won't block a new deployment from scratch (in other words, `expectedVersion` is not required for the next `deploySchema`).

I considered removing them completely, but I don't really see a strong case for it, and I think it could be useful to preserve the history.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/fetchers/admin/UndeploySchemaFetcher.java
olim7t: Yes but for deployment it's slightly different, since the argument can be absent. That's why I didn't try to factor it.

PR #748, opened 2021-03-10
Add optimization for multiple filters in `where` for documents API

EricBorczuk:
This changes the format of the paging state for the documents API, implementing a "last seen" document ID that is used during search requests.

This allows us to optimize searches with multiple AND's a bit, whose implementation is in this PR.

PR #825, opened 2021-04-05
Change the encoding of Pagination for Docs API

EricBorczuk:
Because we are doing internal pagination (over cassandra) in the Docs API, but then are creating documents and making an extra layer of paging over that, this PR adds extra information into a DocsAPI-specific Paging State object. Essentially that object is just a Base64-encoded JSON blob with

```
{
 "documentId": XXX,
 "internalPageState": YYY
}
```

The documentId is the last document that was seen and returned to the user in the last request, and the internalPageState represents the _cassandra_ page that the document was on. So when you search using a Docs page state, it will start at the relevant C* page, and then fast-forward to just past the documentId that was last seen.

This should actually make searches better in two ways:
1) Pages will be definitely consistent/correct, even with mutliple nodes (which is a bit hard to test locally)
2) Search is expected to be a bit faster (there used to be a query with a particular page size that was done to determine the page state when the final document was mid-page, that no longer occurs)

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentSearchPageState.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentSearchPageState.java
@@ -1,0 +1,41 @@
+package io.stargate.web.docsapi.dao;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.nio.ByteBuffer;
+import java.util.Base64;
+
+public class DocumentSearchPageState {
+  @JsonProperty("documentId")
+  String documentId;
+
+  @JsonProperty("internalPageState")
+  String internalPageState;
+
+  @JsonCreator
+  public DocumentSearchPageState(
+      @JsonProperty("documentId") final String documentId,
+      @JsonProperty("internalPageState") final String internalPageState) {
+    this.documentId = documentId;
+    this.internalPageState = internalPageState;
+  }
+
+  public DocumentSearchPageState(final String documentId, final ByteBuffer internalPageStateBuf) {
+    this.documentId = documentId;
+    this.internalPageState = Base64.getEncoder().encodeToString(internalPageStateBuf.array());
+  }
+
+  @JsonIgnore
+  public ByteBuffer getPageState() {
+    if (internalPageState.isEmpty()) {
+      return null;
+    }
+    return ByteBuffer.wrap(Base64.getDecoder().decode(internalPageState));
+  }
+
+  @JsonIgnore
+  public String getLastSeenDocId() {
+    return documentId;
+  }
+}
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
@@ -25,6 +25,7 @@ import io.stargate.db.query.Predicate;
 import io.stargate.db.query.builder.BuiltCondition;
 import io.stargate.db.schema.Column;
 import io.stargate.web.docsapi.dao.DocumentDB;
+import io.stargate.web.docsapi.dao.DocumentSearchPageState;
 import io.stargate.web.docsapi.exception.DocumentAPIErrorHandlingStrategy;
 import io.stargate.web.docsapi.exception.DocumentAPIRequestException;
 import io.stargate.web.docsapi.service.filter.FilterCondition;
@@ -766,7 +767,6 @@ public class DocumentService {
 
   private List<Row> updateExistenceForMap(
       Set<String> existsByDoc,
-      Map<String, Integer> rowCountsByDoc,
       List<Row> rows,
       List<FilterCondition> filters,
       boolean booleansStoredAsTinyint,
@@ -790,11 +790,6 @@ public class DocumentService {
       if (!filteredRows.isEmpty()) {
         existsByDoc.add(key);
       }
-
-      if (!chunk.isEmpty()) {
-        int value = rowCountsByDoc.getOrDefault(key, 0);
-        rowCountsByDoc.put(key, value + chunk.size());
-      }
     }
 
     if (!chunksList.isEmpty() && !endOfResults) {
@@ -804,17 +799,48 @@ public class DocumentService {
   }
 
   /**
+   * If the DocumentSearchPageState has a defined last seen `documentId`, advance the result set to
+   * be just after that ID.
+   */
+  private List<Row> skipSeenRows(DocumentSearchPageState initialPagingState, List<Row> rows) {
+    if (initialPagingState == null) {
+      return rows;
+    }
+
+    String lastSeenId = initialPagingState.getLastSeenDocId();
+    if (lastSeenId == null) {
+      return rows;
+    }
+
+    boolean idFound = false;
+    boolean doneSkipping = false;
+    while (!doneSkipping && rows.size() > 0) {
+      String docId = rows.get(0).getString("key");
+      if (docId.equals(lastSeenId)) {
+        idFound = true;
+        rows.remove(0);
+      } else if (idFound) {
+        doneSkipping = true;
+      } else {
+        rows.remove(0);
+      }
+    }
+
+    return rows;
+  }
+
+  /**
    * This method gets all the rows for @param limit documents, by fetching result sets sequentially
    * and stringing them together. This is NOT expected to perform well for large documents.
    */
-  public ImmutablePair<JsonNode, ByteBuffer> getFullDocuments(
+  public ImmutablePair<JsonNode, DocumentSearchPageState> getFullDocuments(
       Db dbFactory,
       DocumentDB db,
       String authToken,
       String keyspace,
       String collection,
       List<String> fields,
-      ByteBuffer initialPagingState,
+      DocumentSearchPageState initialPagingState,
       int pageSize,
       int limit,
       Map<String, String> headers)
@@ -822,9 +848,20 @@ public class DocumentService {
     ObjectNode docsResult = mapper.createObjectNode();
     LinkedHashMap<String, List<Row>> rowsByDoc = new LinkedHashMap<>();
 
-    ImmutablePair<List<Row>, ByteBuffer> page;
-    ByteBuffer currentPageState = initialPagingState;
+    ImmutablePair<List<Row>, ByteBuffer> page = null;
+    ByteBuffer currentPageState;
+    boolean firstRequest = true;
     do {
+      if (firstRequest) {
+        if (initialPagingState == null) {
+          currentPageState = null;
+        } else {
+          currentPageState = initialPagingState.getPageState();
+        }
+      } else {
+        currentPageState = page.right;
+      }
+
       page =
           searchRows(
               keyspace,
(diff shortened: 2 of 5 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -837,39 +874,40 @@ void addRowsToMap(Map<String, List<Row>> rowsByDoc, List<Row> rows) {
...
       for (int i = 0; i < limit; i++) {
         Map.Entry<String, List<Row>> e = iter.next();
-        totalSize += e.getValue().size();
         List<Row> rows = new ArrayList<>();
         for (Row row : e.getValue()) {
-          if (fields.isEmpty() || fields.contains(row.getString("p0"))) rows.add(row);
+          lastIdSeen = row.getString("key");
+          if (fields.isEmpty() || fields.contains(row.getString("p0"))) {
+            rows.add(row);
+          }
```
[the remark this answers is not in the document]
eribeiro: No need the lines 893, 894, and 899 with the suggestion above.

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -803,28 +798,70 @@ void addRowsToMap(Map<String, List<Row>> rowsByDoc, List<Row> rows) {
...
+        idFound = true;
+        rows.remove(0);
+      } else if (idFound) {
+        doneSkipping = true;
+      } else {
+        rows.remove(0);
+      }
+    }
+
+    return rows;
```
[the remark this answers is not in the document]
eribeiro: oops, looks like `dropWhile` is not available java8. :(

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -950,48 +998,38 @@ void addRowsToMap(Map<String, List<Row>> rowsByDoc, List<Row> rows) {
...
+      Iterator<String> iter = existsByDoc.iterator();
+      String lastSeenId = null;
+      for (int i = 0; i < limit; i++) {
+        lastSeenId = iter.next();
+        docNames.add(lastSeenId);
+      }
+      if (currentPageState != null) {
+        finalPagingState = new DocumentSearchPageState(lastSeenId, currentPageState);
+      } else {
+        finalPagingState = new DocumentSearchPageState(lastSeenId, "");
```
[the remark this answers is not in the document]
eribeiro: Couldn't we pass `currentPageState`. If it's `null` we can anticipate this possibility in `DocumentSearchPageState`

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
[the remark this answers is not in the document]
eribeiro: Other suggestion could be to create a new List and pass it:
```
boolean idFound = false;
List<Row> newRows = new ArrayList<>();
for (Row row : rows) {
     if (docId.equals(lastSeenId)) {
         found = true;
         continue;
     }
     if (found) {
         newRows.add(row);
     }
}
return newRows;
```

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentSearchPageState.java
```diff
@@ -0,0 +1,41 @@
...
+
+  @JsonProperty("internalPageState")
+  String internalPageState;
+
+  @JsonCreator
+  public DocumentSearchPageState(
+      @JsonProperty("documentId") final String documentId,
+      @JsonProperty("internalPageState") final String internalPageState) {
+    this.documentId = documentId;
+    this.internalPageState = internalPageState;
```
eribeiro, line 21:
```suggestion
    this.documentId = Strings.nullToEmpty(documentId);
    this.internalPageState = Strings.nullToEmpty(internalPageState);
```

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentSearchPageState.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  public DocumentSearchPageState(final String documentId, final ByteBuffer internalPageStateBuf) {
+    this.documentId = documentId;
+    this.internalPageState = Base64.getEncoder().encodeToString(internalPageStateBuf.array());
```
eribeiro, line 26:
```suggestion
    this.documentId = Strings.nullToEmpty(documentId);
    if (internalPageStateBuf != null) {
    this.internalPageState = Base64.getEncoder().encodeToString(internalPageStateBuf.array());
    }
    else {
         this.internalPageState = ""
    }
```

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -928,14 +965,25 @@ void addRowsToMap(Map<String, List<Row>> rowsByDoc, List<Row> rows) {
...
         inCassandraFilters.isEmpty()
             ? Collections.emptyList()
             : inCassandraFilters.get(0).getPath();
     do {
+      if (firstRequest) {
+        if (initialPagingState == null) {
+          currentPageState = null;
+        } else {
+          currentPageState = initialPagingState.getPageState();
+        }
```
eribeiro, line 982:
```suggestion
          currentPageState = (initialPagingState == null) ? null :  initialPagingState.getPageState();
```

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -950,48 +998,38 @@ void addRowsToMap(Map<String, List<Row>> rowsByDoc, List<Row> rows) {
...
-        if (existsByDoc.contains(e.getKey())) {
-          docNames.add(e.getKey());
-          i++;
-        }
+      Iterator<String> iter = existsByDoc.iterator();
+      String lastSeenId = null;
+      for (int i = 0; i < limit; i++) {
+        lastSeenId = iter.next();
+        docNames.add(lastSeenId);
+      }
```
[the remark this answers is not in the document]
eribeiro: Change suggestion from line 1099.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentSearchPageState.java
dimas-b: JSON should be able to encode `null` values, so why complicate the processing logic?

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentSearchPageState.java
dimas-b: If `internalPageStateBuf` is `null`, it's probably a processing mistake on the server side. I'd rather throw an `IllegalArgumentException` in that case. WDYT?

EricBorczuk: I think that's probably best, yes - I'm going to be making all of these changes on the branch I'm currently working through btw, since it's based off this branch: https://github.com/stargate/stargate/pull/748

dimas-b, line 26: Why not use the `Encoder.encode(ByteBuffer buffer)` method?

dimas-b: Although, it's a minor point.

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -803,28 +798,70 @@ void addRowsToMap(Map<String, List<Row>> rowsByDoc, List<Row> rows) {
...
+      return rows;
+    }
+
+    boolean idFound = false;
+    boolean doneSkipping = false;
+    while (!doneSkipping && rows.size() > 0) {
+      String docId = rows.get(0).getString("key");
+      if (docId.equals(lastSeenId)) {
+        idFound = true;
+        rows.remove(0);
```
dimas-b, line 821: Removing from iterators may not be efficient and may not even be supported if the source is not a plain list.

Would we use filtering instead?

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -803,28 +798,70 @@ void addRowsToMap(Map<String, List<Row>> rowsByDoc, List<Row> rows) {
...
+      } else if (idFound) {
+        doneSkipping = true;
+      } else {
+        rows.remove(0);
```
dimas-b, line 825: I believe we also have to support the case when the reference ID is not found at all. Could this happen when data is deleted or inserted asynchronously?

dimas-b (changes requested):

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
dimas-b: We may need to use something like `DecoratedKey` and `IPartitioner` to be able to detect when we've advanced past the previous ID.

EricBorczuk: Ah that is an interesting edge-case...If the "last seen" ID was deleted, then some weird behavior will be observed and it will actually skip over the entire first page of results...And if data is inserted such that the "last seen" ID is pushed to the following page, the same will happen - so it will show the last seen ID's document again.

We discussed a bit about how inserts/deletions mid-stream would affect the final result, but yes let's talk through these edge cases (and how those two `DecoratedKey` and `IPartitioner` might help)

eribeiro:
@EricBorczuk @dimas-b I some ideas for a refactoring around pagination. I branched from https://github.com/stargate/stargate/pull/748 and code it. My idea is: if you think it's an idea worth trying then Eric could maybe pull those changes into his working PR. 🙂 So, no PR of mine for this one. Wdyt? What would be the best way of discussing this suggestions?

dimas-b:
@eribeiro : we value your input for sure. Do you have a public version of your refactoring for us to have a look?

In this particular PR, I believe we have a fundamental problem when dealing with the second and following pages (as I mentioned in my [comment](https://github.com/stargate/stargate/pull/825#discussion_r608197343) above). I think we have to solve it on a conceptual level first, then we can start refactoring. I believe it would be easier to review functional change in one PR and then refactoring/clean-up in a follow-up PR. Would this work?

eribeiro:
@dimas-b 

> Do you have a public version of your refactoring for us to have a look?

I can push a branch on my fork of SG. I am not leaning towards not opening a PR, because it will not be merged and it can pollute SG PR. But you think it's okay to open a "WIP-DON'T MERGE" PR, I am okay with that. 

>  fundamental problem when dealing with the second and following pages (as I mentioned in my comment above). I think we have to solve it on a conceptual level first

+1. Couldn't agree more.

> functional change in one PR and then refactoring/clean-up in a follow-up PR

The change I would like to propose sort falls in between a fundamental change and refactoring. There's not a clear division. 

> Would this work?

Sure. :) Gonna push the branch to my repo and let you know asap.

Let's discuss this fundamental problem here, right?

dimas-b:
@EricBorczuk and I talked about the paging problem offline a bit. We'd like to merge this PR more or less "as is" (with review feedback) and then work on persistence-specific improvements to allow resuming paging mid-page in a separate PR. There's potential for users observing oddities with pages Docs API queries in the presence of concurrent deletes, until we resolve the deeper problem. We'll file a GH issue with details on that.

@eribeiro : FYI. I suppose we can consider refactoring current code after merging this PR, but it really depends on what exactly is changing :) hopefully it will make resolving the paging problem easier, but let's review when PR(s) are available.

eribeiro:
> We'd like to merge this PR more or less "as is" (with review feedback) and then work on persistence-specific improvements to allow resuming paging mid-page in a separate PR.

Sounds great! Go ahead. 🙂 

> There's potential for users observing oddities with pages Docs API queries in the presence of concurrent deletes, until we resolve the deeper problem. We'll file a GH issue with details on that.

Cool.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentSearchPageState.java
EricBorczuk: I thought I had explained this in a comment somewhere on this PR< but to write it here as well:
JSON can encode null values, sure - but if Jackson is taking data such as
`{"documentId": "1", "internalPageState": null}`
And tries to turn that into a DocumentSearchPageState object, there is no way for it to know which constructor to use, as the `null` could be referring to a String or a ByteBuffer (there are two constructors in the class)

So I decided to represent the "missing" internal page state in JSON as the empty string - this technically does not have to be done for the `documentId`

dimas-b: Thanks for the clarification - it does make sense.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentSearchPageState.java
EricBorczuk: Oh actually it is possible for a valid DocumentSearchPageState to have an internalPageStateBuf that is `null`, since `null` means "the very start of the results"

Say you have 5 small docs that fit in a single C* page; if a user requests just one doc, then the paging state that is given back to them could be represented by `{"documentId": 1, "internalPageState": ""}` (essentially: I last saw document 1 on the first page of results) - and then when that's. sent back over the wire `internalPageStateBuf` is null, and that null is sent down to C*

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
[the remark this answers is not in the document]
eribeiro: @EricBorczuk Instead of skipping by changing the list in place wouldn't be simpler to create a copy? Just in case you didn't see this snippet above. :)

PR #830, opened 2021-04-07
DRAFT - Refactoring to Docs API

eribeiro:
@dimas-b @EricBorczuk This is a rough idea of what an alternative design/refactoring of the pagination portion. Basically, there's an (stateful) object (err... paginator) that groups all the relevant paging information. It's initially setup by the REST endpoint, but then DocumentService is in charge of changing its status. Some methods like `skipSeenRows` naturally migrated to this object. I also removed the need to return a pair of (Result, DocumentSearchPageState) on many methods. You have the result, if you wanna DocumentSearchPageState just ask the Paginator for the latest version of it. If you like the idea and think it's worth trying, we could incorporate this code into one of Eric's ongoing work, maybe. I am biased, but I think the code became more clear and succint. :) 

**I understand the priority is on solving the corner cases with pagination, but I am opening this PR just for your consideration in the future. This branch was based off Eric's `multiple-filter-performance`.**

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
@@ -1,0 +1,117 @@
+package io.stargate.web.docsapi.dao;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.stargate.db.datastore.Row;
+import io.stargate.web.docsapi.exception.DocumentAPIRequestException;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.List;
+
+/**
+ * This class is in charge of keeping the data of the Docs API pagination process.
+ * It's not only a stateful component, but a state machine with a initial state.
+ * It's created on the REST layer and passed through to the DocumentService that
+ * is in charge of its changes. After being populated its `documentPageState`
+ * can be passed by the REST layer to the client as a JSON string.
+ */
+public class Paginator {
+
+  private final ObjectMapper mapper = new ObjectMapper();
+
+  public final int dbPageSize;
+  public final int docPageSize;
+
+  private boolean initialState = true;
+  private ByteBuffer currentDbPageState; // keeps track of the DB page state while querying the DB
+  private DocumentSearchPageState documentPageState = null;
+
+  public Paginator(String pageStateParam, int pageSizeParam, int dbPageSize) throws IOException {
+    docPageSize = Math.max(1, pageSizeParam);
+    if (docPageSize > DocumentDB.MAX_PAGE_SIZE) {
+      throw new DocumentAPIRequestException("The parameter `page-size` is limited to 20.");
+    }
+
+    this.dbPageSize = dbPageSize;
+
+    if (pageStateParam != null) {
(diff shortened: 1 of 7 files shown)


restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
```diff
@@ -0,0 +1,117 @@
...
+  private ByteBuffer currentDbPageState; // keeps track of the DB page state while querying the DB
+  private DocumentSearchPageState documentPageState = null;
+
+  public Paginator(String pageStateParam, int pageSizeParam, int dbPageSize) throws IOException {
+    docPageSize = Math.max(1, pageSizeParam);
+    if (docPageSize > DocumentDB.MAX_PAGE_SIZE) {
+      throw new DocumentAPIRequestException("The parameter `page-size` is limited to 20.");
+    }
+
+    this.dbPageSize = dbPageSize;
```
dimas-b, line 37: Is this parameter used? I could not find places where it is requested.

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
```diff
@@ -0,0 +1,117 @@
...
+  public ByteBuffer getCurrentDbPageState() {
+    return currentDbPageState;
+  }
+
+  private void leaveInitialState() {
+    initialState = false;
+  }
+
+  private boolean isInitialState() {
+    return initialState == true;
```
dimas-b, line 80: `return initialState`?

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  public List<Row> maybeSkipRows(List<Row> rows) {
+    List<Row> filteredRows = rows;
+    if (isInitialState()) {
+      filteredRows = skipSeenRows(documentPageState, rows);
+      leaveInitialState();
```
dimas-b, line 87: `docPagingState.getLastSeenDocId()` may not be found on the fist DB page, but on the seconds, third, etc. I do not think it is guaranteed to be on the fist page after re-fetching the data from the old paging state.

dimas-b (commented):
I think this PR makes a lot of sense. Yet, @EricBorczuk  and I have been discussing some more fundamental changes in persistence API to allow resuming paging min-page. Those may make this refactoring obsolete. So, even though it is very relevant in current code, I'd prefer to delay this PR a bit until we have a change to develop our ideas about persistence API improvements further.

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
eribeiro: Ouch! Indeed. 🤦

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
eribeiro: Agree. This code tries to mimic the same behavior of Eric's code, but it would to redesigned.

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
eribeiro: This is odd. I have used in some places. Maybe It was left behind and not committed

PR #832, opened 2021-04-08
gRPC boilerplate

mpenick:
Protobuf files live in a separate repository: https://github.com/stargate/stargate-grpc-proto so that they can be shared with clients.

Clients repositories:
https://github.com/stargate/stargate-grpc-java-client
https://github.com/stargate/stargate-grpc-go-client


.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "grpc/src/main/proto"]
+	path = grpc/src/main/proto
+	url = https://github.com/stargate/stargate-grpc-proto.git
+[submodule "testing/src/main/proto"]
```
mpenick, line 4: Is there a better way to share the same submodule in multiple directories?

testing/src/main/java/io/stargate/it/grpc/SimpleQueryTest.java
```diff
@@ -0,0 +1,31 @@
...
+import io.grpc.ManagedChannelBuilder;
+import io.stargate.it.BaseOsgiIntegrationTest;
+import io.stargate.it.storage.SkipIfBundleNotAvailable;
+import io.stargate.it.storage.StargateEnvironmentInfo;
+import io.stargate.proto.QueryOuterClass.Query;
+import io.stargate.proto.QueryOuterClass.Result;
+import io.stargate.proto.StargateGrpc;
+import org.junit.jupiter.api.Test;
+
+@SkipIfBundleNotAvailable(bundleName = "grpc")
```
mpenick, line 15: Use `./mvnw clean package -DskipTests -P default,grpc` to build locally. If `grpc` is not build then this annotation will skip the test automatically.

mpenick:
Codacy is failing because of a few unused variables that will be used eventually. This is ready for review.

.gitmodules
dimas-b: Why bother with an external repo and sub-modules? Why not include gRPC in the main `stargate` repo?

dimas-b: If we want to keep gRPC code in a separate repo, it might be worth running a periodic CI job to check "latest against latest" compatibility. The gRPC repo will contain only its own CI, and local testing should be faster there.

grpc/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <configuration>
+          <unpackBundle>true</unpackBundle>
+          <instructions>
+            <Bundle-Name>RestAPI</Bundle-Name>
+            <Bundle-Description>gRPC service for Stargate</Bundle-Description>
+            <Bundle-SymbolicName>io.stargate.grpc</Bundle-SymbolicName>
+            <Bundle-Activator>io.stargate.grpc.GrpcActivator</Bundle-Activator>
+            <!-- TODO [doug] do this the right way https://baptiste-wicht.com/posts/2010/03/bundle-non-osgi-dependencies-maven.html -->
```
dimas-b, line 112: Remove this comment?

grpc/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <configuration>
+          <unpackBundle>true</unpackBundle>
+          <instructions>
+            <Bundle-Name>RestAPI</Bundle-Name>
```
dimas-b, line 108: `gRpcAPI`?

testing/src/main/java/io/stargate/it/grpc/SimpleQueryTest.java
```diff
@@ -0,0 +1,31 @@
...
+    ManagedChannel channel =
+        ManagedChannelBuilder.forAddress(stargate.nodes().get(0).seedAddress(), 8090)
+            .usePlaintext()
+            .build();
+    StargateGrpc.StargateBlockingStub stub = StargateGrpc.newBlockingStub(channel);
+
+    // TODO: Get and check a "real" result
+    Result result = stub.execute(Query.newBuilder().setCql("SELECT * FROM system.local").build());
+    assertThat(result.hasError()).isFalse();
+    assertThat(result.hasEmpty()).isTrue();
```
dimas-b, line 29: nit: `hasEmpty`?.. it just looks odd :) Is there an `isEmpty()` method in `Result`?

testing/src/main/java/io/stargate/it/storage/SkipIfBundleNotAvailable.java
```diff
@@ -0,0 +1,22 @@
...
+ * <p>Uses {@link BundleAvailableCondition}.
+ */
+@Target({ElementType.TYPE, ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(BundleAvailableCondition.class)
+public @interface SkipIfBundleNotAvailable {
+
+  /** The name of the bundle to check for. */
+  String bundleName();
+}
```
dimas-b, line 22: nit: rename to `AssumeBundleAvailable { String name(); } `? 

This should allow usage as `@AssumeBundleAvailable("grpc")`, which, I guess, might be more intuitive in the JUnit context..

dimas-b: or perhaps `@IfBundleAvailable("grpc")`

testing/src/main/java/io/stargate/it/grpc/SimpleQueryTest.java
mpenick: Generate code for a type called `Empty`. https://github.com/stargate/stargate-grpc-proto/blob/master/query.proto#L174

.gitmodules
mpenick: > Why bother with an external repo and sub-modules? Why not include gRPC in the main stargate repo?

testing/src/main/java/io/stargate/it/grpc/SimpleQueryTest.java
dimas-b: TIL :+1:

.gitmodules
dimas-b: Could we publish `.proto` files as Maven artifacts?

mpenick: > Why bother with an external repo and sub-modules? Why not include gRPC in the main stargate repo?

Most of the code for gRPC is contained in the `stargate` repo. The external repo contains only proto (IDL) files describing the types and calls provided by the Stargate gRPC API. This allows:
* Client implementations to interface only with that proto repo to build new clients
* It allows the API to be versioned separately using tags

dimas-b: Good point about versions :+1: 

At the same time we go `master` against `master` now (if I understand correctly)... Should we be using tags for the IDL from the beginning?

mpenick: > Could we publish .proto files as Maven artifacts?

mpenick: > Should we be using tags for the IDL from the beginning?

I'm iterating heavily on the API. I'll switch to tags when things stabilize.

dougwettlaufer (approved):

.gitmodules
mpenick: Created follow-up issue: https://github.com/stargate/stargate/issues/842

dimas-b: sgtm. thx!

dimas-b (approved):

tomekl007 (approved):

PR #834, opened 2021-04-09
Allow conditions with predicates other than EQ

olim7t:


graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/processor/EntityModel.java
```diff
@@ -149,6 +153,76 @@ public boolean isFederated() {
...
+    }
+
+    // When a clustering field is NOT restricted by an equality relation, then no other clustering
+    // field after it can be restricted by any condition.
+    String firstNonEqField = null;
+    for (FieldModel field : this.getClusteringColumns()) {
+      Optional<WhereConditionModel> maybeCondition =
+          conditions.stream().filter(c -> c.getField().equals(field)).findFirst();
+      if (!maybeCondition.isPresent()) {
+        if (firstNonEqField == null) {
```
ivansenic, line 195: how do you know this is non `EQ` field, you are not checking the `Predicate` in the WhereConditionModel? Actually I am confused, you basically set firstNonEqField in case of a missing clustering column in the where, can you elaborate on this?

What if you do find a field, that has the `EQ` predicate?

dimas-b (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/processor/EntityModel.java
olim7t: If we enter this branch, there is no condition at all. So it passes the test "is not restricted by EQ or IN".

In other words, given:
```
type Foo @cql_input {
  pk: Int! @cql_column(partitionKey: true
  cc1: Int! @cql_column(clusteringOrder: ASC)
  cc2: Int! @cql_column(clusteringOrder: ASC)
}
type Query {
  query1(
    pk: Int
    cc1: Int @cql_where(predicate: LT)
    cc2: Int
  ): [Foo]

  query2(
    pk: Int
    cc2: Int
  ): [Foo]
}
```
Both queries are invalid because they have a condition on `cc2` without a sufficient restriction of `cc1`.

dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/processor/EntityModel.java
ivansenic: Understood. Will resolve this.

PR #836, opened 2021-04-09
Added server http requests micrometer metric with custom tags

ivansenic:
Added a micrometer `http.server.requests` metric with possibility to hook extra tags for each request.

So about some impl details:
* I used `micrometer-jersey2` lib, which provides all the needed lifting. We need only to specify the tags provider, metric name and registry there.. The counting is done in the lib itself..
* The new `HttpMetricsTagProvider` service can define extra tags to be extracted from the headers. If no other module registeres this service, core will register the default impl that returns nothing
* I added the non-osgi `metrics-jersey` module that wraps some general way to use the complete solution in all http modules
* I re-used some of the provided Jersey tags, which comply with Micrometer tag naming conventions (so we would be able to use all Micrometer Grafana dashboards for the http requests right away)
* Unit and integration test provided

Open questions:
* should we maybe not measure http metrics on the `health-checker`
* do we need percentiles, and even more histogram buckets


core/src/main/java/io/stargate/core/metrics/StargateTenantExtractor.java
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * Extracts tenant id from the host.
+   *
+   * @param host Host
+   * @return Optional tenant id.
+   */
+  private static Optional<String> tenantFromHost(String host) {
+    int firstDotIndex = host.indexOf('.');
+    if (firstDotIndex > 0) {
```
ivansenic, line 79: here actually prove that it's at least 36 chars, a UUID length, no need to go to the pattern check if not..

core/src/main/java/io/stargate/core/metrics/StargateTenantExtractor.java
```diff
@@ -0,0 +1,88 @@
...
+
+  private static final String[] HOST_HEADERS = new String[] {"x-forwarded-host", "x-host"};
+
+  /**
+   * Returns tenant tag from the headers. If not found returns <code>null</code>.
+   *
+   * @param headers Header map
+   * @return Tag or null
+   */
+  public static String tenantFromHeaders(Map<String, List<String>> headers) {
```
dimas-b, line 39: Could this be refactored into an OSGi service? Perhaps something like `TagsExtractor` returning a `Map` of `tag_name` -> `value`?

core/src/main/java/io/stargate/core/metrics/api/Metrics.java
```diff
@@ -36,4 +37,21 @@
...
    */
   MeterRegistry getMeterRegistry();
+
+  /**
+   * Returns set of default micrometer {@link Tags} for the given module.
+   *
+   * @param module module name
+   * @return Tags
+   */
+  Tags tagsWithoutTenant(String module);
```
dimas-b, line 47: Rename to `defaultTagsForModule`?

core/src/main/java/io/stargate/core/metrics/api/Metrics.java
```diff
@@ -36,4 +37,21 @@
...
+
+  /**
+   * Returns set of default micrometer {@link Tags} for the given module and tenant.
+   *
+   * @param module module name
+   * @param tenant tenant id
+   * @return Tags
+   */
+  Tags tagsWithTenant(String module, String tenant);
```
dimas-b, line 56: If we have a tag extracting service, I guess this method may not be needed.

core/src/main/java/io/stargate/core/metrics/StargateTenantExtractor.java
dimas-b: We could define an interface for extracting tags from HTTP headers and _no_ default impl.

In environments where this is relevant, the impl. will have to be registered in OSGi. Our metrics handling code will look it up, and use if available. 

We could use a naming convention for service identifiers similar to what we have for `Persistence` to allow different implementations to co-exist without conflicts. The "active" impl. could be selected by a java system property.

health-checker/src/main/java/io/stargate/health/metrics/HealthCheckerTagsProvider.java
```diff
@@ -0,0 +1,74 @@
...
+  private final String module;
+
+  public HealthCheckerTagsProvider(Metrics metrics, String module) {
+    this.metrics = metrics;
+    this.module = module;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public Iterable<Tag> httpRequestTags(RequestEvent event) {
```
dimas-b, line 47: Sorry, I'm not familiar with `micrometer` or `jersey`... Could you explain briefly what the call path to this method looks like? 

Is it used for our Web endpoints only?

Would it make sense to introduce a base module for our HTTP interfaces and keep the common setup/processing code there? It does not have to be an OSGi bundle. A shared plain jar (embedded in each API bundle) might work too.

core/src/main/java/io/stargate/core/metrics/StargateTenantExtractor.java
dimas-b: This seems to be conceptually similar to `AuthorizationProcessor`

core/src/main/java/io/stargate/core/metrics/StargateTenantExtractor.java
```diff
@@ -0,0 +1,88 @@
...
+package io.stargate.core.metrics;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class StargateTenantExtractor {
```
tomekl007, line 26: I see that this is covering extraction of the tenant from headers, but not `ClientState` (for `CQL`) 
Do we plan to support it as well?

ivansenic: Yes this will be supported as the next step, but will have this in mind when designing the interfaces/services..

health-checker/src/main/java/io/stargate/health/metrics/HealthCheckerTagsProvider.java
ivansenic: OK so it's app. like this:

- Jersey provides a `ApplicationEventListener` that has a single method `public RequestEventListener onRequest(RequestEvent requestEvent);`
- You can implement `ApplicationEventListener` and `RequestEventListener` to effectively have listeners on each HTTP requests (this is actually how the dropwizard metrics are also collected) 
- `micrometer-jersey2` provides implementation of both for the micrometer
- when creating the `micrometer-jersey2` event listener you need to provide:
   1. `MetricRegistry` to record metrics (this would be our newly created Prometheus registry)
   2. name of the metric you want to save this to, currently defined by us as `server.http.requests`
   3. `TagsProvider`, the implementation that extract tags that should be used when metric is recorded. The `micrometer-jersey2` provides default impl of this, with some common tags extracting like http method, status code, etc. But we here can plug-in our own stuff and change tags as we like.

core/src/main/java/io/stargate/core/metrics/api/Metrics.java
ivansenic: I was focusing that service only provides a tenant id, that it does nothing about the tags.. I would not mix these.. But yes we could have Tags service and move this stuff there..

core/src/main/java/io/stargate/core/metrics/api/Metrics.java
ivansenic: Done.

core/src/main/java/io/stargate/core/metrics/api/Metrics.java
```diff
@@ -36,4 +37,12 @@
...
    */
   MeterRegistry getMeterRegistry();
+
+  /**
+   * Returns set of default micrometer {@link Tags} for the given module.
+   *
+   * @param module module name
+   * @return Tags
+   */
+  Tags defaultTagsForModule(String module);
```
dimas-b, line 47: nit: rename to `tagsForModule`? Just shorter and it's obvious that one can add more tags later. WDYT?

core/src/main/java/io/stargate/core/metrics/StargateTenantExtractor.java
ivansenic: Done as we agreed.

core/src/main/java/io/stargate/core/metrics/api/Metrics.java
ivansenic: Removed, will resolve.

core/src/main/java/io/stargate/core/CoreActivator.java
```diff
@@ -2,26 +2,45 @@
...
-        new ServiceAndProperties(metricsImpl, MetricsScraper.class, null),
+
+    List<ServiceAndProperties> services = new ArrayList<>();
+    services.add(new ServiceAndProperties(metricsImpl, Metrics.class, null));
+    services.add(new ServiceAndProperties(metricsImpl, MetricsScraper.class, null));
+    services.add(
         new ServiceAndProperties(new HealthCheckRegistry(), HealthCheckRegistry.class, null));
+
+    // register default http tag provider if we are not using any special one
+    if (null == HTTP_TAG_PROVIDER_ID) {
```
tomekl007, line 37: instead of this if, you may consider adding a default for ` System.getProperty("stargate.metrics.http_tag_provider.id");`:
```
 System.getProperty("stargate.metrics.http_tag_provider.id", "default_id");
```

core/src/main/java/io/stargate/core/metrics/api/HttpMetricsTagProvider.java
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Returns tags for a HTTP request, usually extracted from the given headers.
+   *
+   * <p>Nota that the implementation must return constant amount of tags for any input.
+   *
+   * @param headers HTTP Headers
+   * @return Tags
+   */
+  default Tags getRequestTags(Map<String, List<String>> headers) {
+    return Tags.empty();
```
tomekl007, line 43: do we have some implementations that not implement `getRequestTags` method? If not, then maybe default impl is not needed?

auth-api/src/main/java/io/stargate/auth/api/impl/Server.java
```diff
@@ -42,12 +44,19 @@
 
 public class Server extends Application<ApplicationConfiguration> {
 
+  private static final String MODULE_NAME = "authapi";
```
tomekl007, line 47: nice extraction, but maybe the better place for it is an activator class?

testing-services/src/main/java/io/stargate/testing/metrics/TagMeHttpMetricsTagProvider.java
```diff
@@ -0,0 +1,42 @@
...
+import io.micrometer.core.instrument.Tags;
+import io.stargate.core.metrics.api.HttpMetricsTagProvider;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Simple {@link HttpMetricsTagProvider} for testing that extracts {@value TAG_ME_HEADER} header and
+ * converts to tag.
+ */
+public class TagMeHttpMetricsTagProvider implements HttpMetricsTagProvider {
```
tomekl007, line 28: do we really need this special test implementation instead of implementing the production code and using it in tests? We can use the proper header (that carries `tenant_id`) name instead of mock (`host` header). wdyt?
Right now this tenant id extraction is not implemented (and not tested)

ivansenic: This is because from the OSS perspective there are no extra tags.. And this tests confirms that anything can be plug-ed in. I am creating a test impl of the `HttpMetricsTagProvider` and registering it, so that I can confirm that the metrics will include the extra tag(s) returned, what ever they are. The tests you are mentioning are for the `c2` repo..

auth-api/src/main/java/io/stargate/auth/api/impl/Server.java
dimas-b: +1 to move module name constant to the activator.

core/src/main/java/io/stargate/core/metrics/StargateMetricConstants.java
```diff
@@ -0,0 +1,32 @@
...
+public class StargateMetricConstants {
+
+  // metric names
+  public static final String METRIC_HTTP_SERVER_REQUESTS = "http.server.requests";
+
+  // tag keys
+  public static final String MODULE_KEY = "module";
+
+  // unknown tag instances
+  public static final Tag TAG_MODULE_UNKNOWN = Tag.of(MODULE_KEY, "UNKNOWN");
```
dimas-b, line 31: nit: lower case for `UNKNOWN`?.. just to be consistent with "known" module names.

metrics-jersey/src/main/java/io/stargate/metrics/jersey/ResourceTagsProvider.java
```diff
@@ -0,0 +1,73 @@
...
+    // adds method, uri, default and request tags
+    ContainerRequest containerRequest = event.getContainerRequest();
+
+    Tags requestTags =
+        httpMetricsTagProvider.getRequestTags(event.getContainerRequest().getHeaders());
+
+    return defaultTags
+        .and(requestTags)
+        .and(JerseyTags.method(containerRequest), JerseyTags.uri(event));
+  }
```
dimas-b, line 72: just curious: why do we have fewer tags in the "long" version?

restapi/src/main/java/io/stargate/web/RestApiActivator.java
```diff
@@ -36,6 +37,8 @@
           "AuthIdentifier",
           System.getProperty("stargate.auth_id", "AuthTableBasedService"));
   private final ServicePointer<Metrics> metrics = ServicePointer.create(Metrics.class);
+  private final ServicePointer<HttpMetricsTagProvider> httpTagProvider =
+      ServicePointer.create(HttpMetricsTagProvider.class);
```
dimas-b, line 41: nit: This seems to be a common dependency. Would it make sense to introduce a base class for web service and hadle it there?

core/src/main/java/io/stargate/core/metrics/api/HttpMetricsTagProvider.java
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * Returns tags for a HTTP request, usually extracted from the given headers.
+   *
+   * <p>Nota that the implementation must return constant amount of tags for any input.
+   *
+   * @param headers HTTP Headers
+   * @return Tags
+   */
+  default Tags getRequestTags(Map<String, List<String>> headers) {
```
dimas-b, line 42: Why are headers a simple list and not a `Map<String, String>` like we have in `AuthenticationService.validateToken(...)`?

If we have simple strings here, we offload the parsing of HTTP headers to the implementation, which could be avoided, I guess. WDYT?

dimas-b (approved):

restapi/src/main/java/io/stargate/web/RestApiActivator.java
dougwettlaufer: Could be done in https://github.com/stargate/stargate/issues/239?

dougwettlaufer (approved):

restapi/src/main/java/io/stargate/web/RestApiActivator.java
dimas-b: sure

metrics-jersey/src/main/java/io/stargate/metrics/jersey/ResourceTagsProvider.java
ivansenic: Yea this actually enables you to meter the long running task before they are done.. Because of this, you can not have the status that comes from the response included.. Effectively we are not using this atm, but you could add a `@Timed(longTask = true)` to any resource method and this would kick in.. Of course this really has to be a long task, like a huge batch job..

The whole idea behind this is alerting.. if you can only measure something when it's done, then it could be already too late to alert.. This way you can set alerting and get notified once the limit is broken and "job" still runs..

core/src/main/java/io/stargate/core/metrics/api/Metrics.java
ivansenic: done, resolving..

core/src/main/java/io/stargate/core/CoreActivator.java
ivansenic: @tomekl007 The idea here is that if the property is not set, then core will initialize the default.. If we go for the default id, then we need to always specify this option? Or what exactly would be the idea here, can you help plz?

core/src/main/java/io/stargate/core/metrics/api/HttpMetricsTagProvider.java
ivansenic: I use default impl in the `HttpMetricsTagProvider#DEFAULT` field.. @tomekl007 fine to stay like this?

auth-api/src/main/java/io/stargate/auth/api/impl/Server.java
ivansenic: done, resolving.

core/src/main/java/io/stargate/core/metrics/StargateMetricConstants.java
ivansenic: done, resolving.

core/src/main/java/io/stargate/core/CoreActivator.java
tomekl007: Ok, so what will happen if `HTTP_TAG_PROVIDER_ID != null`? Where it will be initialized?

core/src/main/java/io/stargate/core/metrics/StargateTenantExtractor.java
ivansenic: @tomekl007  Can this be resolved?

core/src/main/java/io/stargate/core/metrics/api/HttpMetricsTagProvider.java
tomekl007: ok

core/src/main/java/io/stargate/core/metrics/StargateTenantExtractor.java
tomekl007: Where is this implemented? Will it be in `c2`?

ivansenic: Yes, my next task..

core/src/main/java/io/stargate/core/CoreActivator.java
ivansenic: OK, but this is the same as if I would ask if `HTTP_TAG_PROVIDER_ID == someValueMadeUp` who would set it then? Nobody and start would fail.. 

The whole idea is to set a value that is valid and then the specific module can react on that known/valid value by registering a special impl. The core would back-off then. Again I think my solution is simpler as you don't have to specify anywhere this option by default and core would registered a default one.

tomekl007: Ok, so if any other module provides its own `HttpMetricsTagProvider`, it means that it should set the `stargate.metrics.http_tag_provider.id`, right?

tomekl007 (approved):

core/src/main/java/io/stargate/core/CoreActivator.java
ivansenic: True, or this can be set from the outside I hope.. I am quite not into the modules and how these props are set internally and shared..

PR #837, opened 2021-04-09
Fix #819 - contains and containsEntry with AND

### LINKED ISSUE #639: REST v2: Allow complex JSON types as $eq arguments in where clause
Since #455 the v2 REST API allows JSON array and objects to represent CQL complex types (collections, tuples and UDTs). However this doesn't work in WHERE clauses, for example:

```
curl -s -G -L "http://localhost:8082/v2/keyspaces/users_keyspace/users" \
>    -H  "X-Cassandra-Token: $AUTH_TOKEN" \
>    -H  "Content-Type: application/json" \
>    --data-urlencode 'where={
>      "firstname": {"$eq": "Janesha"},
>      "lastname": {"$eq": "Doesha"},
>      "current_country": { "$eq":  [ "France", "2016-01-01", "2020-02-02" ] }
>    }'
{“description”:“Server error: Value entry for field current_country, operation $eq was expecting a value, but found an object or array.“,”code”:500}
```

The culprit is this check in `WhereParser` at  line 114:

```java
          // Remaining operators: the value is a simple node
          if (!value.isValueNode()) {
            throw new RuntimeException(
                String.format(
                    "Value entry for field %s, operation %s was expecting a value, but found an object or array.",
                    fieldName, rawOp));
          }
```

This was true in v1 where the only possible representation was a string, but it doesn't hold anymore.

Once the check is removed I think `Converters.toCqlValue` should handle the complex JSON value correctly.

### LINKED ISSUE #819: `contains` and `containsEntry` don't seem to work when conjuncted in the REST API (NOT documents)
In this example:
`{"text_map": {"$contains": "a"}, "text_map": {"$contains": "b"}}`

Only the latter filter gets honored and used. In this example:
`{"text_map": {"$contains": "a", "$contains": "b"}}`

Again, only the latter is honored.

Is this a problem with only contains? Or is it a problem with all conjuncted filters?

eribeiro:
As discussed in #819, if we have expressions like `{{"text_map": {"$contains": "a"}, "text_map": {"$contains": "b"}}` then the second entry will override the first one and only `"text_map": {"$contains": "b"}` will effectively be sent to C*. 

This PR
1) defines a syntax like `{"text_map": {"$contains": ["a", "b"]}}` (also valid to $containsEntry, `$containsKey`, and even `$eq`) that will accept both a single object as well as an array as parameter to the operators cited. 

2) Fix a bug in $eq that didn't check for column existence during parsing of the expression;

3) Add more tests to the `WhereParser`;

4) Does a extensive refactoring of the `WhereParser`

5) Fixes #639 in the process.

### DIFF
--- a/restapi/src/main/java/io/stargate/web/service/WhereParser.java
+++ b/restapi/src/main/java/io/stargate/web/service/WhereParser.java
@@ -96,6 +96,7 @@ public class WhereParser {
           default: // GT, GTE, LT, LTE, EQ, NE
             {
               addToCondition(conditions, context);
+              break;
             }
         }
       }
--- a/restapi/src/main/java/io/stargate/web/service/WhereParser.java
+++ b/restapi/src/main/java/io/stargate/web/service/WhereParser.java
@@ -79,7 +79,7 @@ public class WhereParser {
 
         switch (operator) {
           case $IN:
-            evaluateIN(conditions, context);
+            evaluateIn(conditions, context);
             break;
           case $EXISTS:
             evaluateExists(conditions, context);
@@ -233,7 +233,7 @@ public class WhereParser {
             mapValue));
   }
 
-  private static void evaluateIN(List<BuiltCondition> conditions, QueryContext context)
+  private static void evaluateIn(List<BuiltCondition> conditions, QueryContext context)
       throws IOException {
     if (!context.value.isArray()) {
       throw new IllegalArgumentException(
--- a/restapi/src/main/java/io/stargate/web/service/WhereParser.java
+++ b/restapi/src/main/java/io/stargate/web/service/WhereParser.java
@@ -15,12 +15,12 @@
  */
 package io.stargate.web.service;
 
+import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
-import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.stargate.db.query.builder.BuiltCondition;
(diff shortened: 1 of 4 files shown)


eribeiro:
Hi @dougwettlaufer, would you mind to run the tests again? Thanks! 🙂

restapi/src/main/java/io/stargate/web/service/WhereParser.java
```diff
@@ -15,168 +15,242 @@
...
+        QueryContext context = new QueryContext();
+        context.fieldName = fieldName;
+        context.rawOp = rawOp;
+        context.operator = operator;
+        context.value = value;
+        context.type = column.type();
+
+        switch (operator) {
+          case $IN:
+            evaluateIN(conditions, context);
```
EricBorczuk, line 82: nit: `evaluateIn`

restapi/src/test/java/io/stargate/web/resources/WhereParserTest.java
```diff
@@ -168,6 +168,69 @@ public void testParse() throws IOException {
...
+    ImmutableTable table =
+        ImmutableTable.builder()
+            .name("table")
+            .keyspace("keyspace")
+            .addColumns(ImmutableColumn.create("text", Column.Type.Text))
+            .build();
+
+    assertThatThrownBy(() -> WhereParser.parseWhere(whereParam, table))
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessageContaining("Duplicate field '$eq'");
```
EricBorczuk, line 183: :100:

EricBorczuk (approved):
HI @eribeiro 

Thanks for this contribution, it looks really good - I made one small comment about a function name, but would love to get this merged by end-of-week if the test runs permit!

restapi/src/main/java/io/stargate/web/service/WhereParser.java
eribeiro: Thanks @EricBorczuk for your time to review and your comments ! 🙂

PR #838, opened 2021-04-09
Add perPartitionLimit() to QueryBuilder DSL

### LINKED ISSUE #835: Support per partition limits in QueryBuilder
Stargate's `QueryBuilder` does not support `PER PARTITION LIMIT`, even though it is available in CQL.

https://docs.datastax.com/en/cql-oss/3.x/cql/cql_using/useQueryColumnsSort.html

https://docs.datastax.com/en/dse/6.8/cql/cql/cql_using/useQueryColumnsSort.html

This could be useful to Stargate modules in general, but in particular for #828 .

dimas-b:
Closes #835

### DIFF
--- a/persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
+++ b/persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
@@ -85,7 +85,7 @@ import org.javatuples.Pair;
       @SubExpr(
           name = "select",
           definedAs =
-              "select star? column* writeTimeColumn? from (where* limit? orderBy*) allowFiltering?"),
+              "select star? column* writeTimeColumn? from (where* perPartitionLimit? limit? orderBy*) allowFiltering?"),
       @SubExpr(
           name = "index",
           definedAs =
@@ -145,6 +145,7 @@ public class QueryBuilderImpl {
   private final List<BuiltCondition> ifs = new ArrayList<>();
 
   private @Nullable Value<Integer> limit;
+  private @Nullable Value<Integer> limitPerPartition;
   private List<ColumnOrder> orders = new ArrayList<>();
 
   private Replication replication;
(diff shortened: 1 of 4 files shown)


persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
```diff
@@ -665,6 +666,19 @@ public void limit(Integer limit) {
     }
   }
 
+  @DSLAction
+  public void limitPerPartition() {
```
olim7t, line 670: We should stick to the CQL terminology and name this `perPartitionLimit()`.

Regardless of which name is better, we can't change CQL and I think it's more important to be consistent across the board (for example when people look for the option with autocomplete, etc).

dimas-b: updated

dougwettlaufer (approved):
Looks like tests need to be re-ran since codacy upload failed.

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
dimas-b: I renamed field names to follow the `perPartitionLimit` pattern in the latest commit, which also fixes a bug in parameter passing to `Bound`.

mpenick (approved):

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
```diff
@@ -85,7 +85,7 @@
       @SubExpr(
           name = "select",
           definedAs =
-              "select star? column* writeTimeColumn? from (where* limit? orderBy*) allowFiltering?"),
+              "select star? column* writeTimeColumn? from (where* perPartitionLimit? limit? orderBy*) allowFiltering?"),
```
EricBorczuk, line 88: It's legal to have both a per partition and "overall" limit, right?

EricBorczuk (approved):
Looks good to me!

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
dimas-b: yes. `PersisteceTest` has a test case for that too.

PR #847, opened 2021-04-13
Add basic authentication to gRPC

mpenick:
Add basic authentication to the gRPC API using call credentials. Other stuff in the API is going to rely on authentication and the credentials (role) so that's why this is comes first.


mpenick:
It seem the GCP submodule step is not reliably pulling the submodules so I've added this fix: https://github.com/stargate/stargate/pull/847/commits/b2a79080c2c02f52d379baaea972d1f07f1bd7ab

grpc/src/main/java/io/stargate/grpc/server/interceptors/AuthenticationInterceptor.java
```diff
@@ -0,0 +1,43 @@
...
+  public <ReqT, RespT> Listener<ReqT> interceptCall(
+      ServerCall<ReqT, RespT> call, Metadata headers, ServerCallHandler<ReqT, RespT> next) {
+    try {
+      String token = headers.get(TOKEN_KEY);
+      if (token == null) {
+        call.close(Status.UNAUTHENTICATED.withDescription("No token provided"), new Metadata());
+        return null;
+      }
+      Context context = Context.current();
+      context = context.withValue(Server.AUTHENTICATION_KEY, authentication.validateToken(token));
```
dimas-b, line 35: Maybe too early for this, but we're going to have to provide something like `ClientInfo` in this context for parity with authentication over CQL.

Also, CQL authentication goes through `SaslNegotiator`, which does not seem to be relevant here. So, we may need to add a new method to `AuthenticationService` for properly handling non-HTTP connections without SASL.

dimas-b: Having thought more about this, I believe it is preferable to call `validateToken(token, ClientInfo))`. The single parameter method is not guaranteed to be working in all environments as its input data is not always sufficient for authentication.

This is probably ok for now, but we should fix before we consider gRPC for use in the cloud.

dimas-b (approved):

grpc/src/main/java/io/stargate/grpc/server/interceptors/AuthenticationInterceptor.java
mpenick: Good call. I'm not sure how to handle that yet... I'm not sure the way we're handling that with CQL is going to work for gRPC. I'll add a JIRA to fix that for cloud.

PR #848, opened 2021-04-13
Added in Paginator object, use it everywhere

EricBorczuk:


restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
```diff
@@ -0,0 +1,106 @@
...
+
+  public ByteBuffer getCurrentDbPageState() {
+    return currentDbPageState;
+  }
+
+  public List<Row> maybeSkipRows(List<Row> rows) {
+    return skipSeenRows(documentPageState, rows);
+  }
+
+  private List<Row> skipSeenRows(DocumentSearchPageState docPagingState, List<Row> rows) {
```
dimas-b, line 78: Is this method the main purpose of `Paginator`? It may become obsolete if my (yet unpublished) changes for resuming fetching mid-page work out... just FYI.

EricBorczuk: That would be fine - this paginator should be in charge of getting that mid-page paging state from lower down, after your changes (instead of skipSeenRows)

dimas-b (approved):

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
```diff
@@ -0,0 +1,106 @@
...
+  private final ObjectMapper mapper = new ObjectMapper();
+
+  public final int dbPageSize;
+  public final int docPageSize;
+
+  private boolean initialState = true;
+  private ByteBuffer currentDbPageState; // keeps track of the DB page state while querying the DB
+  private DocumentSearchPageState documentPageState = null;
+
+  public Paginator(String pageStateParam, int pageSizeParam, int dbPageSize) throws IOException {
```
tomekl007, line 30: nit, why do you end arguments with the `Param` suffix? I think this is redundant

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
```diff
@@ -0,0 +1,106 @@
...
+
+    if (pageStateParam != null) {
+      byte[] decodedBytes = Base64.getDecoder().decode(pageStateParam);
+      documentPageState = mapper.readValue(decodedBytes, DocumentSearchPageState.class);
+    }
+
+    this.currentDbPageState = (documentPageState != null) ? documentPageState.getPageState() : null;
+  }
+
+  public String getDocumentPageStateAsString() throws JsonProcessingException {
```
tomekl007, line 46: is it possible that this method is called N times? If yes, then maybe we should save the result of (pre-compute):
```
 if (documentPageState != null) {
      byte[] pagingJson = mapper.writeValueAsBytes(documentPageState);
      return Base64.getEncoder().encodeToString(pagingJson);
    }
```
in the field and just return it here?

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
```diff
@@ -0,0 +1,106 @@
...
+    if (docPagingState == null) {
+      return rows;
+    }
+
+    String lastSeenId = docPagingState.getLastSeenDocId();
+    if (lastSeenId == null) {
+      return rows;
+    }
+
+    boolean idFound = docPagingState.isIdFound();
```
tomekl007, line 88: I wonder what's the worst-case time complexity here. What's the upper threshold of the number of rows that `List<Row> rows` can carry/

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * This class is in charge of keeping the data of the Docs API pagination process. It's not only a
+ * stateful component, but a state machine with a initial state. It's created on the REST layer and
+ * passed through to the DocumentService that is in charge of its changes. After being populated its
+ * `documentPageState` can be passed by the REST layer to the client as a JSON string.
+ */
+public class Paginator {
+
+  private final ObjectMapper mapper = new ObjectMapper();
```
ivansenic, line 21: I am not sure how the object mapper is handled in the project in general, but wouldn't it be better to pass it as the part of the constructor. This way any kind of mapper configuration is lost, I am not sure if this is intended or not?

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -866,45 +766,46 @@ void addRowsToMap(Map<String, List<Row>> rowsByDoc, List<Row> rows) {
...
-          if (fields.isEmpty() || fields.contains(row.getString("p0"))) {
-            rows.add(row);
-          }
-        }
-        docsResult.set(e.getKey(), convertToJsonDoc(rows, false, db.treatBooleansAsNumeric()).left);
-      }
-      DocumentSearchPageState pageState = new DocumentSearchPageState(lastIdSeen, currentPageState);
-      return ImmutablePair.of(docsResult, pageState);
+    if (rowsByDoc.keySet().size() > paginator.docPageSize) {
+      AtomicReference<String> lastIdSeen = new AtomicReference<>();
```
ivansenic, line 777: `MutableObject<String>` would also work here as you are not in parallel stream..

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -866,45 +766,46 @@ void addRowsToMap(Map<String, List<Row>> rowsByDoc, List<Row> rows) {
...
+      rowsByDoc.entrySet().stream()
+          .limit(paginator.docPageSize)
+          .forEach(
+              e -> {
+                List<Row> rows =
+                    e.getValue().stream()
+                        .filter(
+                            row -> {
+                              lastIdSeen.set(row.getString("key"));
```
ivansenic, line 786: it's kind a ugly to have side effects are part of the filter:

```
e.getValue().stream()
    .forEachOrdered(row -> lastIdSeen.set(row.getString("key")))
    .filter(..)
```

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
EricBorczuk: Yes, that's a good point - I didn't notice this before; I will use the same ObjectMapper that the rest of the project uses

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
EricBorczuk: Technically it'll only be called once per HTTP request as of now

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
EricBorczuk: rows can have as many as 1000 by default, as defined by:

```
public static final Integer SEARCH_PAGE_SIZE =
  Integer.getInteger("stargate.document_search_page_size", 1000);
```

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
EricBorczuk: oof, this is a problem throughout the stargate project, this pattern of creating ObjectMappers is all over SG. I will make a ticket to make this use a single object mapper, it's more organized wrt configuration...

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
EricBorczuk: This was verbatim copied from someplace else where it made a little bit more sense, I removed it now

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -616,74 +616,19 @@ public void deleteAtPath(
     db.delete(keyspace, collection, id, convertedPath, now);
   }
 
-  public JsonNode searchDocuments(
```
EricBorczuk, line 619: I removed this, as it's dead code and we are now moving far far away from it (nothing uses it)

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -616,74 +616,19 @@ public void deleteAtPath(
...
-
-    if (docsResult.isMissingNode() || (docsResult.isObject() && docsResult.isEmpty())) {
-      return null;
-    }
-
-    return docsResult;
-  }
-
-  public ImmutablePair<JsonNode, ByteBuffer> searchDocumentsV2(
+  public JsonNode searchDocumentsV2(
```
EricBorczuk, line 619: technically this should just be called `searchDocuments` from now on, but for the purposes of review I didn't want the diff to be illegible

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
ivansenic: Yea, ticket is fine..

dougwettlaufer: We just create new ones since so far we haven't had to add custom config for them.

dimas-b: This `ObjectMapper` is being removed in https://github.com/stargate/stargate/pull/872

PR #851, opened 2021-04-14
Add directive to index fields

olim7t:
TODO:
- [x] only allow `FULL` indexes on frozen collections
- [x] only allow `VALUES` indexes on non-frozen collections
- [x] move index parameters to a separate `@cql_index` directive (to allow multiple indexes in the future)

### DIFF
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraSchemaHelper.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraSchemaHelper.java
@@ -16,19 +16,18 @@
 package io.stargate.graphql.schema.schemafirst.migration;
 
 import io.stargate.db.schema.Column;
+import io.stargate.db.schema.SecondaryIndex;
 import io.stargate.db.schema.Table;
 import io.stargate.db.schema.UserDefinedType;
+import io.stargate.graphql.schema.schemafirst.processor.IndexTarget;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 public class CassandraSchemaHelper {
 
   /** @return a list of differences, or empty if the tables match. */
   public static List<Difference> compare(Table expectedTable, Table actualTable) {
-
-    // Note: we deliberately avoid comparing the keyspaces, they might differ because of keyspace
-    // decoration.
-
     String tableName = expectedTable.name();
 
     if (!tableName.equals(actualTable.name())) {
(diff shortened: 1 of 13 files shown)


graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraSchemaHelper.java
```diff
@@ -109,12 +115,63 @@ private static boolean equals(Column.ColumnType expectedType, Column.ColumnType
     return true;
   }
 
-  /** @return a list of differences, or empty if the UDTs match. */
-  public static List<Difference> compare(UserDefinedType expectedType, UserDefinedType actualType) {
+  private static SecondaryIndex findIndex(Table table, Column column) {
```
tomekl007, line 118: any reason why not call it `findSecondayIndex()`?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraSchemaHelper.java
```diff
@@ -109,12 +115,63 @@ private static boolean equals(Column.ColumnType expectedType, Column.ColumnType
...
+  private static void compareIndex(
+      SecondaryIndex expectedIndex,
+      SecondaryIndex actualIndex,
+      Column expectedColumn,
+      List<Difference> differences) {
+
+    if (actualIndex == null) {
+      differences.add(
+          new Difference(expectedColumn, expectedIndex, DifferenceType.MISSING_INDEX, null));
+      return;
```
tomekl007, line 141: what if we have > 1 missing indexes? would it find only the 1st one and return it?

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/processor/EntityModelBuilder.java
```diff
@@ -216,6 +224,29 @@ private static Table buildCqlTable(
...
+      field
+          .getIndex()
+          .ifPresent(
+              index -> {
+                ImmutableSecondaryIndex.Builder builder =
+                    ImmutableSecondaryIndex.builder()
+                        .keyspace(keyspaceName)
+                        .column(column)
+                        .name(index.getName())
+                        .indexingClass(index.getIndexClass().orElse(null))
```
tomekl007, line 243: at what level this `null` will be handled? (the `orElse(null)` case)

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraSchemaHelper.java
mpenick: This is for a single index comparison. There's nothing left to check here if the actual index is missing it'll add the difference and the loop where this is called will keep looping for other columns/indexes.

olim7t: Yes, this is in the context of a single column so there can be at most 1 index missing.

However I just realized that CQL _does_ allow more than one index per column, for a map column you can have both KEYS, VALUES and ENTRIES indexes at the same time. We don't support maps at the moment, but for the sake of future compatibility it might be better to have a separate `@cql_index` annotation.

graphqlapi/src/main/java/io/stargate/graphql/schema/schemafirst/migration/CassandraSchemaHelper.java
olim7t: Not really, I'll rename it.

PR #852, opened 2021-04-14
gRPC basic value handling

mpenick:


grpc/src/main/java/io/stargate/grpc/codec/cql/BytesCodec.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.google.protobuf.ByteString;
+import io.grpc.Status;
+import io.grpc.StatusException;
+import io.stargate.db.schema.Column;
+import io.stargate.proto.QueryOuterClass;
+import io.stargate.proto.QueryOuterClass.Value;
+import io.stargate.proto.QueryOuterClass.Value.InnerCase;
+import java.nio.ByteBuffer;
+import org.jetbrains.annotations.NotNull;
```
ivansenic, line 11: so I see that no other module except test uses the `jetbrain-annotations`, was this intentional?

grpc/src/main/java/io/stargate/grpc/codec/cql/BytesCodec.java
```diff
@@ -0,0 +1,30 @@
...
+import org.jetbrains.annotations.Nullable;
+
+public class BytesCodec implements ValueCodec {
+  @Nullable
+  @Override
+  public ByteBuffer encode(@Nullable QueryOuterClass.Value value, @NotNull Column.ColumnType type)
```
ivansenic, line 17: So non of the `ValueCodec` classes are handling cases when `value ` is `null`, which is quite strange considering that there is an explicit `@Nullable` on the param..

grpc/src/main/java/io/stargate/grpc/codec/cql/IntCodec.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class IntCodec implements ValueCodec {
+  @Nullable
+  @Override
+  public ByteBuffer encode(@Nullable QueryOuterClass.Value value, @NonNull Column.ColumnType type)
+      throws StatusException {
+    if (value.getInnerCase() != InnerCase.INT) {
+      throw Status.FAILED_PRECONDITION.withDescription("Expected integer type").asException();
+    }
+    int intValue = (int) value.getInt();
```
ivansenic, line 23: Maybe [Math.html#toIntExact(long)](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Math.html#toIntExact(long)) fits better here?

grpc/src/main/java/io/stargate/grpc/payload/cql/ValuesHandler.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.List;
+
+public class ValuesHandler implements PayloadHandler {
+  @Override
+  public BoundStatement bindValues(Prepared prepared, Payload payload)
+      throws InvalidProtocolBufferException, StatusException {
+    final Values values = payload.getValue().unpack(Values.class);
+    final List<Column> columns = prepared.metadata.columns;
+    final int columnCount = columns.size();
+    if (values == null && columnCount > 0 || columnCount != values.getValuesCount()) {
```
ivansenic, line 30: nit: maybe extra brackets would be easier to read?

grpc/src/main/java/io/stargate/grpc/payload/cql/ValuesHandler.java
```diff
@@ -0,0 +1,169 @@
...
+    final List<ByteBuffer> boundValues = new ArrayList<>(columnCount);
+    List<String> boundValueNames = null;
+    if (values.getValueNamesCount() != 0) {
+      final int namesCount = values.getValueNamesCount();
+      boundValueNames = new ArrayList<>(namesCount);
+      for (int i = 0; i < namesCount; ++i) {
+        String name = values.getValueNames(i);
+        Column column =
+            columns.stream()
+                .filter(c -> c.name() == name)
```
ivansenic, line 47: Seems you are comparing String by reference equality here?

grpc/src/main/java/io/stargate/grpc/payload/PayloadHandlers.java
```diff
@@ -0,0 +1,15 @@
+package io.stargate.grpc.payload;
+
+import io.stargate.grpc.payload.cql.ValuesHandler;
+import io.stargate.proto.QueryOuterClass.Payload;
+import io.stargate.proto.QueryOuterClass.Payload.Type;
+import java.util.EnumMap;
+
+public class PayloadHandlers {
+  public static final EnumMap<Payload.Type, PayloadHandler> HANDLERS =
```
ivansenic, line 9: Maybe it's better to expose the interface here `Map<Payload.Type, PayloadHandler>` ? ps this is modifiable from anybody, maybe better init in the `static` and wrap with unmodifiable

grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -0,0 +1,218 @@
...
+import io.grpc.Context;
+import io.grpc.Status;
+import io.grpc.StatusException;
+import io.grpc.StatusRuntimeException;
+import io.grpc.stub.StreamObserver;
+import io.stargate.auth.AuthenticationSubject;
+import io.stargate.core.metrics.api.Metrics;
+import io.stargate.db.AuthenticatedUser;
+import io.stargate.db.ClientInfo;
+import io.stargate.db.ImmutableParameters;
```
ivansenic, line 14: Intellij complains this is a missing class, but could be something local to me

ivansenic (commented):
I would expect way more tests for such a PR, testing all codec classes and `ValuesHandler` in details.. Not sure if you planned to add them?

mpenick:
@ianjevans This is a draft PR.

grpc/src/main/java/io/stargate/grpc/payload/cql/ValuesHandler.java
mpenick: Thanks. Good catch.

grpc/src/main/java/io/stargate/grpc/service/Service.java
mpenick: It means that the generated code for `ImmutableParameters` wasn't properly attached to the project.

grpc/src/main/java/io/stargate/grpc/codec/cql/BytesCodec.java
mpenick: Good catch. Not sure what happened there.

ianjevans:
> > I would expect way more tests for such a PR, testing all codec classes and ValuesHandler in details.. Not sure if you planned to add them?
> 
> @ianjevans This is a draft PR.

I think you meant @ivansenic.

ivansenic:
> > I would expect way more tests for such a PR, testing all codec classes and ValuesHandler in details.. Not sure if you planned to add them?
> 
> @ivansenic This is a draft PR.

@mpenick Ohh sorry, totally missed that.. Still I hope you could pull out some benefits from my review..

grpc-proto/proto/stargate.proto
```diff
@@ -8,7 +8,7 @@ package stargate;
 import "query.proto";
 
 service Stargate { // Maybe this should be call `Query` or `Cql`?
```
dougwettlaufer, line 10: Is a query both reads and writes or will we also have a "mutation"? I would suggest not using "Cql" to avoid any confusion.

mpenick: No mutation.

dimas-b (approved):
I'm not sure I understand all the implications, but LGTM :)

dougwettlaufer (approved):

PR #857, opened 2021-04-15
fixes #843: better responses for the mutation queries

### LINKED ISSUE #843: Update row using GQL with `ifExists: true` response needs updates when the row doesn't actually get inserted/upserted
I've a table with below schema:

```
CREATE TABLE restkeyspace.cyclist_semi_pro (
    id int PRIMARY KEY,
    affiliation map<text, text>,
    age int,
    country text,
    firstname text,
    lastname text,
    registration date
);
```

It has below values to begin with:

```
token@cqlsh:restkeyspace> SELECT * FROM restkeyspace.cyclist_semi_pro ;

 id | affiliation            | age | country | firstname | lastname | registration
----+------------------------+-----+---------+-----------+----------+--------------
  2 | {'name': 'Rotar Club'} |  31 |     FRA |       Joe |    Maddy |   2020-07-22
  4 |                   null |  31 |     FRA |       Joe |     Dude |   2020-10-02
  3 |      {'Smart': 'Dude'} |  31 |     IND |        JJ |     JACK |   2020-07-22

(3 rows)
```

If I try to run a GraphQL API to update a row (which doesn't exist `id=1`) using the following (notice the usage of `ifExists: true`):

```
curl --location --request POST 'https://DBID-REGION.apps.astra.datastax.com/api/graphql/restkeyspace' \
--header 'x-cassandra-token: AstraCS:blah:blah' \
--data-raw '{"query":"mutation cyclist_semi_pro {\n  update: updatecyclist_semi_pro(value: {id:1, firstname:\"Herman\"},ifExists: true ) { \n    value {\n      id\n      firstname\n    }\n  }\n}","variables":{}}'
```

I get an update/response that feels like it upserted the data into the table:

```
{
    "data": {
        "update": {
            "value": {
                "id": 1,
                "firstname": "Herman"
            }
        }
    }
}
```

When I query the table again, I don't see the row with `id = 1`:

```
token@cqlsh:restkeyspace> SELECT * FROM restkeyspace.cyclist_semi_pro ;

 id | affiliation            | age | country | firstname | lastname | registration
----+------------------------+-----+---------+-----------+----------+--------------
  2 | {'name': 'Rotar Club'} |  31 |     FRA |       Joe |    Maddy |   2020-07-22
  4 |                   null |  31 |     FRA |       Joe |     Dude |   2020-10-02
  3 |      {'Smart': 'Dude'} |  31 |     IND |        JJ |     JACK |   2020-07-22

(3 rows)
```

I would like to understand why? (or) if we need to modify the response to state it didn't actually insert/upsert that data into the table. FWIW, if I use ifExists: false, I see the row getting inserted which also has the same exact response structure. TIA!

> Olivier Michallat — Today at 11:02 AM
Yes, we need to change the response. The query works as expected (the row doesn't exist so it's not updated), but the response data simply mirrors the input data, which is wrong. We need to check the [applied]  field in the CQL response instead.

Discord reference: https://discord.com/channels/745831180213158001/766642967976476672/831536741093343232

ivansenic:
I fixed the response we send back for all mutation queries in case when `ifExist` or `ifNotExists` is set to `true`. Basically I am doing a generic handling here and only changing what we return back in case I get non-empty rows from Cassandra. 

I am also asking myself if we should actually return the input back in case we have no rows. If we are going to mirror the Cassandra behavior then we should maybe not do that. Not sure.

I added tests and added cleanup of the inserted products as they affect the `io.stargate.it.http.GraphqlTest#queryWithPaging` test if those inserts run before it.

### DIFF
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/MutationFetcher.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/MutationFetcher.java
@@ -90,10 +90,8 @@ public abstract class MutationFetcher extends DmlFetcher<CompletableFuture<Map<S
       // if we have no rows means that we got no information back from query execution, return
       // original value to the user and applied true to denote that query was accepted
       // not matter if the underlying data is not changed
-      ImmutableMap.Builder<String, Object> resultBuilder = ImmutableMap.builder();
-      resultBuilder.put("value", originalValue);
-      resultBuilder.put("applied", true);
-      return resultBuilder.build();
+      return ImmutableMap.of("value", originalValue, "applied", true);
+
     } else {
       // otherwise check what can we get from the results
       // mutation target only one row
@@ -103,10 +101,7 @@ public abstract class MutationFetcher extends DmlFetcher<CompletableFuture<Map<S
 
       // if applied we can return the original value, otherwise use database state
       Object finalValue = applied ? originalValue : value;
-      ImmutableMap.Builder<String, Object> resultBuilder = ImmutableMap.builder();
-      resultBuilder.put("value", finalValue);
-      resultBuilder.put("applied", applied);
-      return resultBuilder.build();
+      return ImmutableMap.of("value", finalValue, "applied", applied);
     }
   }
 
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/MutationFetcher.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/MutationFetcher.java
@@ -25,10 +25,13 @@ import io.stargate.auth.AuthenticationSubject;
 import io.stargate.auth.AuthorizationService;
 import io.stargate.db.datastore.DataStore;
 import io.stargate.db.datastore.DataStoreFactory;
+import io.stargate.db.datastore.ResultSet;
+import io.stargate.db.datastore.Row;
 import io.stargate.db.query.BoundQuery;
 import io.stargate.db.schema.Table;
 import io.stargate.graphql.schema.cqlfirst.dml.NameMapping;
 import io.stargate.graphql.web.HttpAwareContext;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 
@@ -74,9 +77,33 @@ public abstract class MutationFetcher extends DmlFetcher<CompletableFuture<Map<S
     }
 
     // Execute as a single statement
-    return dataStore
-        .execute(query)
-        .thenApply(rs -> ImmutableMap.of("value", environment.getArgument("value")));
+    return dataStore.execute(query).thenApply(rs -> toMutationResult(rs, environment));
+  }
+
+  private Map<String, Object> toMutationResult(
+      ResultSet resultSet, DataFetchingEnvironment environment) {
+
+    List<Row> rows = resultSet.currentPageRows();
+    Object originalValue = environment.getArgument("value");
+
+    if (rows.isEmpty()) {
+      // if we have no rows means that we got no information back from query execution, return
+      // original value to the user
+      return ImmutableMap.of("value", originalValue);
+    } else {
+      // otherwise check what can we get from the results
+      // mutation target only one row
+      Row row = rows.iterator().next();
+      boolean applied = row.getBoolean("[applied]");
+      Map<String, Object> value = DataTypeMapping.toGraphQLValue(nameMapping, table, row);
+
+      // return value from query, if given or not applied, fallback to the original
+      Object finalValue = !value.isEmpty() || !applied ? value : originalValue;
+      ImmutableMap.Builder<String, Object> mutationResultMapBuilder = ImmutableMap.builder();
+      mutationResultMapBuilder.put("value", finalValue);
+      mutationResultMapBuilder.put("applied", applied);
+      return mutationResultMapBuilder.build();
+    }
   }
 
   private CompletableFuture<Map<String, Object>> executeAsBatch(
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/MutationFetcher.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/MutationFetcher.java
@@ -88,8 +88,12 @@ public abstract class MutationFetcher extends DmlFetcher<CompletableFuture<Map<S
 
     if (rows.isEmpty()) {
       // if we have no rows means that we got no information back from query execution, return
-      // original value to the user
-      return ImmutableMap.of("value", originalValue);
+      // original value to the user and applied true to denote that query was accepted
+      // not matter if the underlying data is not changed
+      ImmutableMap.Builder<String, Object> resultBuilder = ImmutableMap.builder();
+      resultBuilder.put("value", originalValue);
+      resultBuilder.put("applied", true);
+      return resultBuilder.build();
     } else {
       // otherwise check what can we get from the results
       // mutation target only one row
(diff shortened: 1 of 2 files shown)


graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/MutationFetcher.java
```diff
@@ -74,9 +77,33 @@ protected MutationFetcher(
...
+  private Map<String, Object> toMutationResult(
+      ResultSet resultSet, DataFetchingEnvironment environment) {
+
+    List<Row> rows = resultSet.currentPageRows();
+    Object originalValue = environment.getArgument("value");
+
+    if (rows.isEmpty()) {
+      // if we have no rows means that we got no information back from query execution, return
+      // original value to the user
+      return ImmutableMap.of("value", originalValue);
```
dimas-b, line 92: Is it actually meaningful to the user? The user already has the original value :thinking:

ivansenic: Exactly what I asked in the PR description :thinking:

testing/src/main/java/io/stargate/it/http/GraphqlTest.java
```diff
@@ -827,6 +827,96 @@ public void insertProducts() {
     assertThat(product.getPrice()).hasValue(input.price());
     assertThat(product.getCreated()).hasValue(input.created());
     assertThat(product.getDescription()).hasValue(input.description());
+
+    // cleanup on the end
+    cleanupProduct(client, productId);
```
dimas-b, line 832: move to an `@AfterEach` method?

ivansenic: Yea, I was unaware you can get all products easily..  Done, resolving.

ivansenic:
@olim7t would be great if you can have a look on this today, so I can continue on Monday to wrap it up..

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/MutationFetcher.java
dimas-b: Sorry, TL;DR ;-)

mpenick: Would we also return `applied` is `true` here?

ivansenic: @mpenick Good question, I also have code it at first like this.. 

However, I was not sure if we can surely say that it was applied here? Reading Cassandra CQL reference I was a bit confused, not sure what happens if you hit insert with same primary keys twice?

olim7t: > Is it actually meaningful to the user? The user already has the original value 🤔

Legitimate question, but since the existing code was already doing it I think we have to preserve backward compatibility.

> Should we also return `applied` is `true` here?

I would lean towards yes. `applied` is only relevant for lightweight transactions, but you could say that if a query is not a LWT, then it's not a failed LWT, therefore it was "applied" in the sense that the backend did not reject it (regardless of whether it actually changed any data). That's how we did it in the Java driver at least.

testing/src/main/java/io/stargate/it/http/GraphqlTest.java
dimas-b: We could refactor the test code to keep track of generated product IDs and then remove them all inside an `@AfterEach` method.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/MutationFetcher.java
dimas-b: Although adding `applied` to the response to ordinary mutations makes sense to me, I think that will invalidate our backwards compatibility since we cannot mix it into the existing (original) value that is currently returned in this case. WDYT?

olim7t: If it's a completely new field so it won't break compatibility. Legacy requests did not include it in the selection set, so they'll keep returning exactly the same response.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/MutationFetcher.java
```diff
@@ -74,9 +77,33 @@ protected MutationFetcher(
...
+    } else {
+      // otherwise check what can we get from the results
+      // mutation target only one row
+      Row row = rows.iterator().next();
+      boolean applied = row.getBoolean("[applied]");
+      Map<String, Object> value = DataTypeMapping.toGraphQLValue(nameMapping, table, row);
+
+      // return value from query, if given or not applied, fallback to the original
+      Object finalValue = !value.isEmpty() || !applied ? value : originalValue;
```
olim7t, line 101: I wonder if we should consider `value.isEmpty()` in this condition.

The only cases when `!applied && value.isEmpty()` are failed `IF EXISTS` queries:
```
cqlsh:test> delete from foo where k = 1 if exists;

 [applied]
-----------
     False
```
With the current code we'll return the input data, which is not consistent with CQL (and to be clear this was already the case before, it's not a consequence of your changes).

I think the idea is when `applied == false` is to reflect the conflicting database state that caused the LWT to fail. If that state is "there was nothing to delete", then returning an empty value (like cqlsh does) is the right thing to do.

So I would simply do:
```suggestion
      Object finalValue = applied ? originalValue : value;
```

This is a breaking change, but I think the previous behavior was wrong.
WDYT?

ivansenic: This is more than valid imo..

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/MutationFetcher.java
ivansenic: Since this is also how it's done in the Java driver, maybe we can do it then.. It would seem a bit odd for the delete though, you would say applied, although the delete might not remove any data.. But if we look it from the *the sense that the backend did not reject it*, then it's fine I guess..

testing/src/main/java/io/stargate/it/http/GraphqlTest.java
ivansenic: no need, see here: https://github.com/stargate/stargate/blob/a39ec6014b9066ad091777c93bb5560a894a7949/testing/src/main/java/io/stargate/it/http/GraphqlTest.java#L181

ivansenic:
@olim7t @dimas-b applied suggestions & rebased, will merge on the approve..

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/MutationFetcher.java
```diff
@@ -74,9 +77,37 @@ protected MutationFetcher(
...
+    Object originalValue = environment.getArgument("value");
+
+    if (rows.isEmpty()) {
+      // if we have no rows means that we got no information back from query execution, return
+      // original value to the user and applied true to denote that query was accepted
+      // not matter if the underlying data is not changed
+      ImmutableMap.Builder<String, Object> resultBuilder = ImmutableMap.builder();
+      resultBuilder.put("value", originalValue);
+      resultBuilder.put("applied", true);
+      return resultBuilder.build();
```
olim7t, line 96: Nit: this could also be shortened as
```suggestion
      return ImmutableMap.of("value", originalValue, "applied", true);
```
But the builder version is slightly more readable. Feel free to keep whatever you prefer.

dimas-b (approved):

olim7t (approved):
LGTM, all conditional mutations now behave consistently with raw CQL.

Minor cosmetic comment below, feel free to apply or not.

dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/MutationFetcher.java
ivansenic: Yea, let's shorten it, less code, it's also readable without the builder.

PR #862, opened 2021-04-16
fixes #393: allow eq with null, trasnsform to in filter

### LINKED ISSUE #393: Add support for nullish operations
Such as: `$exists: false, $eq: null`

Most probably these will be "limited support" operations when SAI is not in the schema - not sure if SAI being present would actually help here!

ivansenic:
This seems like the simplest thing to do. I am validating `$eq` with `null` to OK, but then transforming this to in query internally.. I think there would be no other way around. Suggestions are welcome :ear_with_hearing_aid: 

I think I can not do anything for the other predicates that don't support null.

Should I update the documentation somewhere?

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
@@ -479,7 +479,7 @@ public class DocumentService {
     if (!filterOpt.isPresent()) return;
     FilterOp filterOp = filterOpt.get();
 
-    if (filterOp == FilterOp.NE) {
+    if (filterOp == FilterOp.EQ || filterOp == FilterOp.NE) {
       if (value.isArray() || value.isObject()) {
         throw new DocumentAPIRequestException(
             String.format(
@@ -494,8 +494,7 @@ public class DocumentService {
     } else if (filterOp == FilterOp.GT
         || filterOp == FilterOp.GTE
         || filterOp == FilterOp.LT
-        || filterOp == FilterOp.LTE
-        || filterOp == FilterOp.EQ) {
+        || filterOp == FilterOp.LTE) {
       if (value.isArray() || value.isObject() || value.isNull()) {
         throw new DocumentAPIRequestException(
             String.format(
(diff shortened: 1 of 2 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -595,7 +594,17 @@ private void validateOpAndValue(String op, JsonNode value, String fieldName) {
...
         } else {
-          conditions.add(new SingleFilterCondition(convertedFieldNamePath, op, (String) null));
+          // EQ does not support null values in Cassandra 3.x, thus we would manually move that to
+          // the IN [null]
+          SingleFilterCondition nullFilterCondition =
+              new SingleFilterCondition(convertedFieldNamePath, op, (String) null);
+          if (Objects.equals(nullFilterCondition.getFilterOp(), FilterOp.EQ)) {
+            conditions.add(
+                new ListFilterCondition(
+                    convertedFieldNamePath, FilterOp.IN.rawValue, Collections.singletonList(null)));
```
dimas-b, line 604: `IN` with `null` elements is not supported by CQL, e.g. in C* 3.11:
```
cqlsh:test> select * from d where p1 = 'a' and val in ('x') allow filtering;

 x | p1 | p2 | val
---+----+----+-----
 1 |  a |    |   x

(1 rows)
cqlsh:test> select * from d where p1 = 'a' and val in (null) allow filtering;
InvalidRequest: Error from server: code=2200 [Invalid query] message="Unsupported null value for column val"
```

So I assume the filtering for this `ListFilterCondition` will filter in memory. However, in this case could we avoid "hiding" the `null` in the list and instead add logic to process `null` values in `EQ` predicates as first class values?.. or at least perhaps create a dedicated condition for `EQ null` cases?

dimas-b: Also, as far as I understand, if `property EQ null` is the only query condition, we'll execute a full scan across the collection table, followed by PK queries to fetch specific values for the `property` of each document... That could cause a lot of work on the SG side :thinking:

ivansenic: @dimas-b @EricBorczuk OK, from your comments, I actually would think about not doing this, reasoning being:

* We actually hide from the user the fact that we will use the `IN` condition, and yes this could have huge impact if this is the only condition (which you would expect to execute fast). So maybe better would be to keep it forbidden, maybe send better error message back and suggest using `IN (null)` with consequences.
* I tried to understand what @dimas-b was saying about the *and instead add logic to process null values in EQ predicates as first class values*, how would this actually work, I see everywhere that null queering is not supported in C*? Maybe we should do a quick sync today once you are online.

dimas-b: > add logic to process null values in EQ predicates as first class values,

By that I meant just code clean-up to make it the handling on `EQ null` conditions explicit in Docs API code. It would not help with C*-level issues.

So, all in all, I tend to agree that it may be best to delay fixing this issue. Hopefully, when we have a more elaborate query planning engine in Doc API we could tackle this again. If we can detect that other more restrictive condition are present, then filtering for null values in memory would not be a problem, I think. It may even be doable in current code, but I think we need to put some more thought into how we communicate query complexities to the end user. I see some code in Docs API that denies queries that hit large volumes of data already (e.g. in [searchRows(...)](https://github.com/stargate/stargate/blob/1d8390b0402b7ea4221db7de71006bbe45afc109/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java#L1193))

ivansenic: OK @EricBorczuk I am closing this as not done.

PR #874, opened 2021-04-20
Add a way to insert multiple objects at once 

tomekl007:


testing/src/main/java/io/stargate/it/http/GraphqlTest.java
```diff
@@ -1411,7 +1461,7 @@ public void queryWithPaging() {
         .map(p -> p.getValues().map(v -> !v.isEmpty()).orElse(false))
         .orElse(false)); // Continue if there are still values
 
-    assertThat(names).containsExactlyInAnyOrder("a", "b", "c");
+    assertThat(names).contains("a", "b", "c");
```
tomekl007, line 1464: This test seems to be not properly isolated and depending on the "shared" state of the db.
When I inserted `Non-Legs` and `Shiny Legs` it returned it those as well. When running the `containsExactlyInAnyOrder` version in isolation it passes. Less strict validation (`contains`) works.
I am not sure if we are fine with that change, please let me know.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkInsertMutationFetcher.java
```diff
@@ -0,0 +1,92 @@
...
+          Scope.MODIFY,
+          SourceAPI.GRAPHQL);
+      boundQueries.add(query);
+    }
+    return boundQueries;
+  }
+
+  private List<ValueModifier> buildInsertValues(Map<String, Object> value) {
+
+    Preconditions.checkNotNull(value, "Insert statement must contain at least one field");
```
olim7t, line 83: I think this condition is never hit, `value` is an element of the list so it will always be present.

(And in fact I think it's not necessary in `InsertMutationFetcher` either, since the `value` is non-null in the schema)

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcher.java
```diff
@@ -0,0 +1,99 @@
...
+    if (values.size() != queries.size()) {
+      throw new IllegalStateException("Number of values to insert should match number of queries");
+    }
+
+    List<CompletableFuture<Map<String, Object>>> results = new ArrayList<>(values.size());
+    for (int i = 0; i < queries.size(); i++) {
+      if (operation.getDirectives().stream().anyMatch(d -> d.getName().equals(ATOMIC_DIRECTIVE))
+          && operation.getSelectionSet().getSelections().size() > 1) {
+        throw new UnsupportedOperationException(
+            "The @" + ATOMIC_DIRECTIVE + "in not supported for bulk inserts.");
```
olim7t, line 81: Why?
We support `@atomic` for multiple simple insert mutations. A bulk insert generates exactly the same statements under the hood, so we should be able to batch them the same way. We're just going to need a few adjustments in `BatchContext` (a mutation can now generate more than one query, so we can't use `queries.size()` anymore to test if everything has been added; but we can track the number of mutations separately).

olim7t (changes requested):

testing/src/main/graphql/betterbotz/products/BulkInsertProducts.graphql
```diff
@@ -0,0 +1,16 @@
+mutation BulkInsertProducts(
+    $values: [ProductsInput!]
+    $ifNotExists: Boolean
+    $options: MutationOptions
+) {
+    bulkInsertProducts(values: $values, ifNotExists: $ifNotExists, options: $options) {
+        applied
```
ivansenic, line 7: so there is no notion of the array return type here.. I can simply say what I want from each value right?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkInsertMutationFetcher.java
```diff
@@ -0,0 +1,92 @@
...
+          dataStore
+              .queryBuilder()
+              .insertInto(table.keyspace(), table.name())
+              .value(buildInsertValues(value))
+              .ifNotExists(ifNotExists)
+              .ttl(getTTL(environment))
+              .build()
+              .bind();
+
+      authorizationService.authorizeDataWrite(
```
ivansenic, line 69: do we really need to authorize every time, these are all same ops?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcher.java
```diff
@@ -0,0 +1,99 @@
...
+      }
+
+      // Execute as a single statement
+      int finalI = i;
+      results.add(
+          dataStore
+              .execute(queries.get(i))
+              .thenApply(rs -> ImmutableMap.of("value", values.get(finalI))));
```
ivansenic, line 89: this would not match with the updates in the #857, which will be merged right now..

testing/src/main/java/io/stargate/it/http/GraphqlTest.java
```diff
@@ -829,6 +826,48 @@ public void insertProducts() {
...
+    ProductsInput product2 =
+        ProductsInput.builder()
+            .id(productId2)
+            .name("Non-Legs")
+            .price("1000.99")
+            .created(now())
+            .description("Non-legs.")
+            .build();
+
+    bulkInsertProducts(client, Arrays.asList(product1, product2));
```
ivansenic, line 852: verify result of the insert as well..

testing/src/main/java/io/stargate/it/http/GraphqlTest.java
ivansenic: Will be solved in #857, I added the `@AfterEach cleanProducts()`, so you can revert to original code.

ivansenic (commented):
Looks good.. Have a look on the comments..

testing/src/main/java/io/stargate/it/http/GraphqlTest.java
tomekl007: There is a verification hidden in the `bulkInsertProducts` method.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcher.java
tomekl007: I will adapt my PR once your is merged

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkInsertMutationFetcher.java
tomekl007: yes.

graphqlapi/src/test/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcherTest.java
```diff
@@ -0,0 +1,168 @@
...
+  public void mutationAtomicBulkSingleSelectionTest() {
+    assertQuery(
+        "mutation @atomic { m1: bulkInsertbooks(values: [{ title: \"a\" }] ) { applied } }",
+        "INSERT INTO library.books (title) VALUES ('a')");
+  }
+
+  @Test
+  @DisplayName(
+      "Atomic bulk mutations with single selection with N values should use normal execution")
+  public void mutationAtomicBulkSingleSelectionTestNValues() {
```
tomekl007, line 49: for all mock tests that are using bulk with N elements in the array, I am getting a weird error, I.e.:
`Incompatible conditions title = a and title = b`
from the `WhereProcessor`:
```
format("Incompatible conditions %s and %s", existing.firstCondition, condition));
```
It seems that the insert here treated as one query with multiple AND statements. 
What Am I missing? Inserting N elements in array works in the integration test (see for example: `bulkInsertProductsWithAtomic`) Is the below syntax correct?
```
 "mutation @atomic { m1: bulkInsertbooks(values: [{ title: \"a\" }, { title: \"b\" }] ) { applied } }"
```
@olim7t

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcher.java
tomekl007: in progres

testing/src/main/graphql/betterbotz/products/BulkInsertProducts.graphql
tomekl007: yes, the list return type is generated here: https://github.com/stargate/stargate/pull/874/files#diff-0abbca7fbce169d57be0fe97d3d62960e061a295d2d0974c5cc8702165416527R310-R311 
dynamically

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcher.java
```diff
@@ -0,0 +1,161 @@
...
+          batchContext.setExecutionResult(batchDataStore.batch(batchContext.getQueries()));
+        }
+      }
+    }
+    return batchContext
+        .getExecutionFuture()
+        .thenApply(v -> ImmutableMap.of("values", environment.getArgument("values")));
+  }
+
+  private int calculateNumberOfAllQueries(OperationDefinition operationDefinition) {
```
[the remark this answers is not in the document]
olim7t: I think it would be easier to keep it based on the number of selections. You just need a separate variable in `BatchContext` because it's no longer equal to the number of queries.
```java
public static class BatchContext {
  private final List<BoundQuery> queries = new ArrayList<>();
  private final int operationCount;

  public synchronized int add(BoundQuery query) {
    queries.add(query);
    operationCount += 1;
    return operationCount;
  }

  public synchronized int add(List<BoundQuery> newQueries) {
    queries.add(newQueries);
    operationCount += newQueries.size();
    return operationCount;
  }
}
```

In `MutationFetcher.executeAsBatch`:
```java
    int selections = environment.getOperationDefinition().getSelectionSet().getSelections().size();
...
    } else if (batchContext.add(query) == selections) {
      ... // execute the batch
    }
```

In `BulkMutationFetcher.executeAsBatch`:
```java
    int selections = environment.getOperationDefinition().getSelectionSet().getSelections().size();
...
    } else if (batchContext.add(queries) == selections) {
      ... // execute the batch
    }
```

graphqlapi/src/test/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcherTest.java
olim7t: I can't reproduce it, I'm getting another error because `batchCaptor` didn't capture anything... I suppose that was a different version of the code. But anyway I would try debugging to see how the queries are being constructed.

tomekl007: Yes, I am getting the same error from `captor`. However, the mentioned `Incompatible conditions` is also generated but it is not propagated to the caller.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcher.java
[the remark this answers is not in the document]
tomekl007: This solution doesn't work. Let's consider a situation when we can have N selections, and each of the selections can have M queries. Let's assume that we have two selections. The first one has two queries, the second one has one query:
```
s1: [{q1}, {q2}]
s2: {q3}
```
for that case, the:
```
int selections = environment.getOperationDefinition().getSelectionSet().getSelections().size();
```
evaluates to `2`. When the first selection is evaluated, and both queries are added, the 
```
public synchronized int add(List<BoundQuery> newQueries) {
```
will return two. So, this condition:
```
batchContext.add(queries) == selections
```
will be met and the batch will be executed. I am not sure if this should be executed yet, because the second selection contains one additional query. Shouldn't all 3 queries be batched and executed together as one query?

This is why I initially thought that the `BulkMutationFetcher` will execute all 3 queries, so then the selections must be equal to the total number of queries (3 for that case), and not selections. 
However, when debugging this I see that there will be two separate queries executed:
1. Batch query for `q1` and `a2` using the `BulkMutationFetcher`
2. Single query for `q3` using the `BulkMutationFetcher`
Is it really atomic if we have two separate queries? 
 
For this situation, the proposed solution is also problematic:
1. When the first selection with `q1` and `q2` arrives at `BulkMutationFetcher` it will be executed once both queries are added (because `2 == queries == selections`)
2. The `MutationFetcher` tries to execute the remaining `q3`.
It will be added to the `batchContext`, but after the `batchContext.add(query)` is executed, it returns `3` (it had 2 queries added by `BulkMutationFetcher` and the `q3`. However, the `selections == 2`, so the `q3` won't be executed at all. The problem is that the `BatchContext` is shared between both mutation fetchers.

The situation is getting more complicated if the first selection has more queries (i.e. 10 queries). For that case, the 
`BulkMutationFetcher#batchContext.add(queries) == selections)` won't be evaluated properly. 
Another problem when we switch the order of queries (first the selection with on query q1, then the selection with two queries). For this situation, not all bulk queries from the second selection will be executed.

graphqlapi/src/test/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcherTest.java
tomekl007: There was some problem with mocks for that test. I've moved both test scenarios to the Integration test suit. Removing unit for the same use-case

dimas-b (commented):
It might have missed it (apologies if so), but would we want to add a guardrail for the input data size? CQL (for reference) appears to have per-request input size restriction.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcher.java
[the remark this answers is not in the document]
olim7t: 🤦 my bad, I meant to also increment by one when multiple queries are added:
```java
public static class BatchContext {
  private final List<BoundQuery> queries = new ArrayList<>();
  private final int operationCount;

  public synchronized int add(BoundQuery query) {
    queries.add(query);
    operationCount += 1;
    return operationCount;
  }

  public synchronized int add(List<BoundQuery> newQueries) {
    queries.add(newQueries);
    operationCount += 1;
    return operationCount;
  }
}
```

tomekl007:
> It might have missed it (apologies if so), but would we want to add a guardrail for the input data size? CQL (for reference) appears to have per-request input size restriction.

Do you mean batch size? Good idea, but let's add it in a separate ticket. The `@atomic` was already working before this PR without such a guardrail. So this new guardrail would need to impact the existing and the new code.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcher.java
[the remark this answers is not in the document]
tomekl007: Yeah, your solution looks simpler. I've implemented it in:
https://github.com/stargate/stargate/pull/874/commits/c358b1012c7b0eec0183f877a3bde8bb3d391059

dimas-b:
> On the other, if the batch size is too big, it will result in the exception from the backed, right? [...]

My concern is that just processing overly large query input in Stargate (before sending it to persistence) might be enough overhead to cause performance issues and allow DOS attacks.

Yes, we can certainly address this in a separate PR.

ivansenic (commented):
The comments I posted seem to be fixed :+1:

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcher.java
```diff
@@ -0,0 +1,145 @@
...
+import io.stargate.graphql.web.HttpAwareContext;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public abstract class BulkMutationFetcher
```
mpenick, line 40: Why is this abstracted out from `BulkInsertMutationFetcher`? Are there plans to have an bulk update mutation?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/InsertMutationFetcher.java
```diff
@@ -15,6 +15,8 @@
  */
 package io.stargate.graphql.schema.cqlfirst.dml.fetchers;
 
+import static io.stargate.graphql.schema.cqlfirst.dml.fetchers.TtlFromOptionsExtractor.getTTL;
```
mpenick, line 18: Is this needed?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/UpdateMutationFetcher.java
```diff
@@ -1,5 +1,7 @@
 package io.stargate.graphql.schema.cqlfirst.dml.fetchers;
 
+import static io.stargate.graphql.schema.cqlfirst.dml.fetchers.TtlFromOptionsExtractor.getTTL;
```
mpenick, line 3: Is this needed?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/InsertMutationFetcher.java
olim7t: It's used at line 67. That doesn't appear in the diff because it was previously an instance method in a parent class.

olim7t (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcher.java
tomekl007: This PR introduces bulk mutations

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/UpdateMutationFetcher.java
tomekl007: same as https://github.com/stargate/stargate/pull/874#discussion_r620786176

PR #878, opened 2021-04-21
Add a feature flag to enable GraphQL-first API

olim7t:


stargate-starter/src/main/java/io/stargate/starter/Starter.java
```diff
@@ -233,6 +233,10 @@
       name = {"--disable-mbean-registration", "Whether the mbean registration should be disabled"})
   protected boolean disableMBeanRegistration = false;
 
+  @Order(value = 21)
+  @Option(name = {"--enable-graphql-first", "Whether to enable the GraphQL-first API"})
+  protected boolean enableGraphqlFirst = false;
```
dimas-b, line 238: Is it worth adding a top-level cmd. line option for this? I imagine we are going to enable graphql-first by default at some point (soon?), so the option will become obsolete, but removing it at that time could break environments that happen to use it. If we stick to `-D` only, not checking it after graphql-first is default should be transparent to end users. WDYT?

olim7t: Good question. Yes, GraphQL-first will be enabled by default at some point. But I think the option could still be useful to disable it, in case someone just uses CQL-first and doesn't want their users to "accidentally" deploy schemas. In that case we would simply change the default, not remove the option entirely.

dougwettlaufer: My only concern with this is by adding the top-level cmd we're now going from an alpha feature flag to an actually supported feature of toggling graphql-first on and off. Likewise, @dimas-b also brings up a good point in "removing it at that time could break environments that happen to use it".

dimas-b: If we want a long-lived on/off flag, how about something like `--enable-feature graphql-first` (short: `-E graphql-first`) instead? The flag is more generic and allows arbitrary values, which we can later process (or ignore) in code.

olim7t: OK let's stick to a simple system property for now.

dimas-b (approved):

PR #879, opened 2021-04-21
Refactor JSON construction/conversion into its own service, with tests

EricBorczuk:
As a start to refactoring DocumentService into reasonable chunks, this PR takes `convertToJsonDoc` and all of its dependent functions and creates its own service, the `JsonConverterService`, with relevant unit tests

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
@@ -3,12 +3,7 @@ package io.stargate.web.docsapi.service;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
-import com.fasterxml.jackson.databind.node.BooleanNode;
-import com.fasterxml.jackson.databind.node.DoubleNode;
-import com.fasterxml.jackson.databind.node.LongNode;
-import com.fasterxml.jackson.databind.node.NullNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
-import com.fasterxml.jackson.databind.node.TextNode;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
@@ -24,7 +19,6 @@ import io.stargate.db.datastore.ResultSet;
 import io.stargate.db.datastore.Row;
 import io.stargate.db.query.Predicate;
 import io.stargate.db.query.builder.BuiltCondition;
-import io.stargate.db.schema.Column;
 import io.stargate.web.docsapi.dao.DocumentDB;
 import io.stargate.web.docsapi.dao.Paginator;
 import io.stargate.web.docsapi.exception.DocumentAPIErrorHandlingStrategy;
@@ -54,11 +48,17 @@ import org.slf4j.LoggerFactory;
 
 public class DocumentService {
   private static final Logger logger = LoggerFactory.getLogger(DocumentService.class);
-  private static final ObjectMapper mapper = new ObjectMapper();
   private static final Pattern PERIOD_PATTERN = Pattern.compile("\\.");
   private static final Splitter FORM_SPLITTER = Splitter.on('&');
   private static final Splitter PAIR_SPLITTER = Splitter.on('=');
   private static final Splitter PATH_SPLITTER = Splitter.on('/');
+
+  private final JsonConverterService jsonConverterService = new JsonConverterService();
+
+  @VisibleForTesting
+  JsonConverterService getJsonConverterService() {
+    return jsonConverterService;
+  }
 
   /*
    * Converts a JSON path string (e.g. "$.a.b.c[0]") into a JSON path string
@@ -436,7 +436,7 @@ public class DocumentService {
 
   public JsonNode getJsonAtPath(
       DocumentDB db, String keyspace, String collection, String id, List<PathSegment> path)
-      throws ExecutionException, InterruptedException, UnauthorizedException {
+      throws UnauthorizedException {
     List<BuiltCondition> predicates = new ArrayList<>();
     predicates.add(BuiltCondition.of("key", Predicate.EQ, id));
 
@@ -461,7 +461,7 @@ public class DocumentService {
       return null;
     }
     ImmutablePair<JsonNode, Map<String, List<JsonNode>>> result =
-        convertToJsonDoc(rows, false, db.treatBooleansAsNumeric());
+        getJsonConverterService().convertToJsonDoc(rows, false, db.treatBooleansAsNumeric());
     if (!result.right.isEmpty()) {
       logger.info(String.format("Deleting %d dead leaves", result.right.size()));
       db.deleteDeadLeaves(keyspace, collection, id, result.right);
@@ -628,6 +628,7 @@ public class DocumentService {
       throws UnauthorizedException {
     FilterCondition first = filters.get(0);
     List<String> path = first.getPath();
+    ObjectMapper mapper = getJsonConverterService().getMapper();
 
     List<Row> rows =
         searchRows(
@@ -668,7 +669,10 @@ public class DocumentService {
         for (Row row : docRows) {
           List<Row> wrapped = new ArrayList<>(1);
           wrapped.add(row);
-          JsonNode jsonDoc = convertToJsonDoc(wrapped, true, db.treatBooleansAsNumeric()).left;
+          JsonNode jsonDoc =
+              getJsonConverterService()
+                  .convertToJsonDoc(wrapped, true, db.treatBooleansAsNumeric())
+                  .left;
           ref.add(jsonDoc);
         }
       }
@@ -683,7 +687,10 @@ public class DocumentService {
         }
 
         List<Row> nonNull = chunk.stream().filter(x -> x != null).collect(Collectors.toList());
-        JsonNode jsonDoc = convertToJsonDoc(nonNull, true, db.treatBooleansAsNumeric()).left;
+        JsonNode jsonDoc =
+            getJsonConverterService()
+                .convertToJsonDoc(nonNull, true, db.treatBooleansAsNumeric())
+                .left;
 
         if (documentId == null) {
           ((ArrayNode) docsResult.get(key)).add(jsonDoc);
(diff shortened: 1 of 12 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverterService.java
```diff
@@ -0,0 +1,290 @@
...
+
+public class JsonConverterService {
+  private static final ObjectMapper mapper = new ObjectMapper();
+
+  public ObjectMapper getMapper() {
+    return mapper;
+  }
+
+  // TODO find a place to refactor this to, along with DocumentService#getBooleanFromRow
+  private static Boolean getBooleanFromRow(Row row, String colName, boolean numericBooleans) {
```
dimas-b, line 23: Would it be cleaner to make `numericBooleans` a instance-level field as opposed to a method parameter?

I guess this is value is fixed in each Stargate process (i.e. it does not change from call to call). WDYT?

restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverterService.java
```diff
@@ -0,0 +1,290 @@
...
+    if (row.isNull("bool_value")) return null;
+    if (numericBooleans) {
+      byte value = row.getByte(colName);
+      return value != 0;
+    }
+    return row.getBoolean(colName);
+  }
+
+  public ImmutablePair<JsonNode, Map<String, List<JsonNode>>> convertToJsonDoc(
+      List<Row> rows, boolean writeAllPathsAsObjects, boolean numericBooleans) {
```
dimas-b, line 33: Could we use simple `Pair` for the return type to make it more generic?

restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverterService.java
```diff
@@ -0,0 +1,290 @@
...
+    JsonNode doc = mapper.createObjectNode();
+    Map<String, Long> pathWriteTimes = new HashMap<>();
+    Map<String, List<JsonNode>> deadLeaves = new HashMap<>();
```
dimas-b, line 36: Why do we convert deal leaves to json? It seems unnecessary since we do not return them to the client :thinking:

dimas-b: What if we use custom classes to represent dead leaves? I suppose it should make the code more readable.

EricBorczuk: That is true - custom class would be better - also re: the `Pair` above it should be represented as a class of its own probably instead of a Pair...I'll look into this

restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverterService.java
EricBorczuk: Yea that would be clearer for sure

restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverterService.java
dimas-b: Also dead leaves appear to be ignored in some call paths. Would it make sense to make the building of this map optional?

restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverterService.java
dimas-b: Suggestion: use a custom inteface with `@Value.Immutable` for holding the result. It should be as efficient as the `ImmutablePair`, but more readable since we have assign meaningful names to the result components. It should able be easier to extend.

ivansenic (commented):
I know that generally refactoring is boring and sometimes if seems like a waist of time.. But the goal here is to increase our confidence about quality of specific component, to find new bugs that we did not know before, that we increase the ease of making changes for other developers.. Thus, this work is super important imo.. 

@EricBorczuk Good job! Keep it up! :+1:

restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverterService.java
```diff
@@ -0,0 +1,290 @@
...
+import io.stargate.db.datastore.Row;
+import io.stargate.db.schema.Column;
+import io.stargate.web.docsapi.dao.DocumentDB;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+
+public class JsonConverterService {
```
ivansenic, line 15: So copying the stuff form the `DocumentService` is a good first step. A second should be imo to:

* refactor that `convertToJsonDoc` into smaller junks that would make it easier to understand
* check for what shared logic between methods can be extracted/unified
* add docs to the methods so that reader can understand what's going on

restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverterService.java
```diff
@@ -0,0 +1,290 @@
...
+        if (isArray && !writeAllPathsAsObjects) {
+          if (!ref.isArray()) {
+            if (i == 0) {
+              doc = mapper.createArrayNode();
+              ref = doc;
+              pathWriteTimes.put(parentPath, rowWriteTime);
+            } else {
+              markObjectAtPathAsDead(ref, parentPath, deadLeaves);
+              ref = changeCurrentNodeToArray(row, parentRef, i);
+              pathWriteTimes.put(parentPath, rowWriteTime);
```
eribeiro, line 106: `pathWriteTimes.put(parentPath, rowWriteTime);`  at lines 102 and 106 could be extracted out of the inner if else (`if (i==0)`)

restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverterService.java
```diff
@@ -0,0 +1,290 @@
...
+            if (!ref.isObject()) {
+              markArrayAtPathAsDead(ref, parentPath, deadLeaves);
+              ref = changeCurrentNodeToObject(row, parentRef, i, writeAllPathsAsObjects);
+              pathWriteTimes.put(parentPath, rowWriteTime);
+            }
+
+            ((ObjectNode) ref).set(p, childRef);
+          }
+          parentRef = ref;
+          ref = childRef;
```
eribeiro, line 149: These statements are the same on the if branch at lines 130 and 131. They could be extracted from this inner loop.

restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverterService.java
eribeiro: +1 about using visitor interface to collect dead leaves.

restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverterService.java
eribeiro, line 15: `Nit:` Why not naming it as just `JsonConverter`? It's usually advisable not to overload a software's classes with suffixes like Manager, Service, Controller, etc.  

(please, feel free to ignore this comment 🙂 )

EricBorczuk: Yeah I could rename - only reason why I named it this way is because DocumentService is named that way.

However, since the package name already states that these are service logic, I think it's reasonable to leave Service out

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -668,7 +669,10 @@ public JsonNode searchDocumentsV2(
         for (Row row : docRows) {
           List<Row> wrapped = new ArrayList<>(1);
           wrapped.add(row);
-          JsonNode jsonDoc = convertToJsonDoc(wrapped, true, db.treatBooleansAsNumeric()).left;
+          JsonNode jsonDoc =
+              getJsonConverterService()
+                  .convertToJsonDoc(wrapped, true, db.treatBooleansAsNumeric())
+                  .left;
```
eribeiro, line 675: Instead of creating a list of size 1, pass a `Collections.singletonList(row)`.

restapi/src/main/java/io/stargate/web/docsapi/service/json/DeadLeafCollectorImpl.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public void addLeaf(String path, DeadLeaf leaf) {
+    Set<DeadLeaf> leavesAtPath = deadPaths.getOrDefault(path, new HashSet<>());
+    leavesAtPath.add(leaf);
+    deadPaths.put(path, leavesAtPath);
+  }
+
+  public void addArray(String path) {
+    Set<DeadLeaf> leavesAtPath = deadPaths.getOrDefault(path, new HashSet<>());
+    leavesAtPath.add(ImmutableDeadLeaf.builder().name(DeadLeaf.ARRAY).build());
```
dimas-b, line 20: Could we just make `ImmutableDeadLeaf.builder().name(DeadLeaf.ARRAY).build()` a constant on the `DeadLeaf` interface to avoid re-creating the same object?

restapi/src/main/java/io/stargate/web/docsapi/service/json/DeadLeafCollector.java
```diff
@@ -0,0 +1,17 @@
...
+
+public interface DeadLeafCollector {
+
+  void addLeaf(String path, DeadLeaf leaf);
+
+  void addArray(String path);
+
+  void addAll(String path);
+
+  Map<String, Set<DeadLeaf>> getLeaves();
```
dimas-b, line 14: Might be preferable to keep this at the impl. level to avoid requiring a particular data representation from all possible collectors.

restapi/src/main/java/io/stargate/web/docsapi/service/json/DeadLeafCollectorNoOp.java
```diff
@@ -0,0 +1,20 @@
...
+
+public class DeadLeafCollectorNoOp implements DeadLeafCollector {
+  public void addLeaf(String path, DeadLeaf leaf) {}
+
+  public void addArray(String path) {}
+
+  public void addAll(String path) {}
+
+  public Map<String, Set<DeadLeaf>> getLeaves() {
+    throw new UnsupportedOperationException("No-op implementation doesn't have leaves");
```
dimas-b, line 14: nit: this is probably dead code by now?

dimas-b (approved):

PR #882, opened 2021-04-22
Regenerate licenses-report.txt

dougwettlaufer:


licenses-report.txt
```diff
@@ -2,6 +2,13 @@
 Generated by fossa-cli (https://github.com/fossas/fossa-cli).
 This software includes the following software and licenses:
 
+========================================================================
+Android-Sdk
+========================================================================
+The following software have components provided under the terms of this license:
+
+- Android SDK (from https://www.android.com/)
```
dimas-b, line 10: SG uses Android SDK? :thinking:

ivansenic (approved):
Wanna add the instruction how is this generated to the `DEV_GUIDE.md`, could be useful for us that still don't know it?

licenses-report.txt
dougwettlaufer: At least a dependency or dependency of a dependency does... My guess would be something related to Apollo in the testing modules.

dimas-b (approved):

PR #884, opened 2021-04-23
adds bool expression logic for the document api filters

ivansenic:
This is a complete refactoring for the document API filters. This code should take over the responsibility for constructing expression tree, validating user input wrt to different types of predicates and testing against the database row in order to check if all filters match.

There are three base parts here:
-  `Predicates` - these are concrete operations like `$eq`, `$in`, etc, that might or might not have a corresponding database predicate. They know how to test filter value against one or more database values that can be String, Double or Boolean (this is what we store). 
- `Conditions` - they wrap predicates and bridge them with the concrete filter value and context. They can do 2 things: provide `BuiltCondition` if possible & test a `Row` for matching
- `Expressions` - the logic provided by the `jbool_expressions` that enables creating expression trees with And, Or, etc. Furthermore, it provides possibility to simplify expression provided by the user, filter, collect, etc.

Of course this is work in progress, so tests are not done. But with this refactoring, we can test more effectively for the each predicate & condition.

Besides being needed for query optimizer, I think with this in future we will have possibility to do things like:
* and or, not, etc
* remove non-Cassandra predicates from the expression tree (and vice versa)
* simplify expressions like `a < 3 && a > 3` => `false`
* provide transformations like `a in [2]` => `a == 2`
* order expressions by performance and/or expected return size, for example, given `a < 3 && b == 2` run first the second condition
* etc


restapi/src/main/java/io/stargate/web/docsapi/service/query/provider/ListConditionProvider.java
```diff
@@ -0,0 +1,80 @@
...
+                result.add(node.asBoolean());
+            } else if (node.isNumber()) {
+                result.add(node.asDouble());
+            } else if (node.isTextual()) {
+                result.add(node.asText());
+            } else if (node.isNull()) {
+                result.add(null);
+            }
+
+            // TODO should we throw the exception here if node is smth else?
```
EricBorczuk, line 74: Yea we could definitely throw in the `else` case

restapi/src/main/java/io/stargate/web/docsapi/service/query/provider/ConditionProviderService.java
```diff
@@ -0,0 +1,71 @@
...
+        Iterator<Map.Entry<String, JsonNode>> fields = conditionsNode.fields();
+        fields.forEachRemaining(field -> {
+            String filterOp = field.getKey();
+            ConditionProvider jsonNodeConditionProvider = providerMap.get(filterOp);
+            if (null != jsonNodeConditionProvider) {
+                Optional<? extends BaseCondition> condition = jsonNodeConditionProvider.createCondition(field.getValue(), numericBooleans);
+                condition.ifPresent(results::add);
+            }
+
+            // TODO what in case fo the non found, f.e $my, throw exception?
```
EricBorczuk, line 44: Meaning if an invalid filter is sent over the wire? This probably should throw and cause a 400 upstream

restapi/src/main/java/io/stargate/web/docsapi/service/query/predicate/impl/NotInPredicate.java
```diff
@@ -0,0 +1,81 @@
...
+    public boolean test(List<?> filterValue, Boolean dbValue) {
+        return !super.test(filterValue, dbValue);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean test(List<?> filterValue, Double dbValue) {
+        return !super.test(filterValue, dbValue);
```
EricBorczuk, line 78: :100:

restapi/src/main/java/io/stargate/web/docsapi/service/query/predicate/impl/InPredicate.java
```diff
@@ -0,0 +1,92 @@
...
+    @Override
+    public boolean test(List<?> filterValue, Double dbValue) {
+        // TODO maybe more correct number matching here as well
+        return filterValue.stream()
+                .filter(Number.class::isInstance)
+                .map(Number.class::cast)
+                .anyMatch(value -> Double.valueOf(value.doubleValue()).equals(dbValue));
+    }
+
+    // TODO empty list validation needed?
```
EricBorczuk, line 90: Yea - I think we might need to disallow empty list entirely, since $nin depends on $in's logic, we probably would like both to fail

I guess another option is to say that `in []` always is false, and `nin []` always is true, but I think that's kind of strange (postgres `IN` for instance doesn't allow empty lists)

EricBorczuk (commented):
This looks really nice! I added a few comments in the spots where you wrote TODO's/asked questions in code comments

I definitely like the direction here, it looks way cleaner than what I cobbled together before!

restapi/src/main/java/io/stargate/web/docsapi/service/query/provider/ConditionProviderService.java
ivansenic: I would agree. However, I just realized that there is no "mechanism" of throwing exceptions in the service that can later on be transformed to some kind of the status code + provide correct `Error` back. Usually I deal with such situations, by having something like this:

```
class BusinessRuntimeException extends RuntimeException {

  private String message;
  private ErrorCode errorCode;

  public Response getResponse() {
     ResponseStatus status = errorCode.getStatusProvider().get();
     String defaultMessage = errorCode.getStatusProvider().getDefaultMesssage();
     Error entity = new Error(Optional.ofNullable(message).orElse(defaultMessage), status.getCode());
     return Response.status(status)
        .entity(error)
       .build():
  }
}

enum ErrorCode {
   
   Supplier<ResponseStatus> statusSupplier;
   String defaultMessage;

   DOCSAPI_FILTER_WRONG_ARGUMENT(() -> ResponseStatus.BAD_REQUEST),
   DOCSAPI_FILTER_OPERATION_DOES_NOT_EXIST(() -> ResponseStatus.BAD_REQUEST),
   ....
   ....
}
```

Let's discuss this, would be a nice addition to the complete Rest API. This is also very nice for testing, as you can always confirm from the service that thrown exception has proper `ErrorCode`.

restapi/src/main/java/io/stargate/web/docsapi/service/query/predicate/impl/InPredicate.java
ivansenic: I was not sure here, but I added a validation for now. Empty or null will not be accepted by both in predicates.

restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterPath.java
```diff
@@ -0,0 +1,65 @@
...
+            return String.join(".", path);
+        } else {
+            return "";
+        }
+    }
+
+    /**
+     * @return Returns complete JSON path as String to the field, including the field.
+     */
+    default String getFullFieldPathString() {
```
dimas-b, line 56: not a major issue, but I find it a bit confusion when we separate the "field" from "path", which leads to having to maintain a "full path". This is a bit counter-intuitive with respect to how JRE's `File` and `Path` objects work, as far as far as I understand :) Could we always have "path" include "field" as the last element?

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/BaseCondition.java
```diff
@@ -0,0 +1,77 @@
...
+    Optional<BuiltCondition> getBuiltCondition();
+
+    /**
+     * Resolves {@link String} value from the document {@link Row}.
+     *
+     * @param row Row
+     * @return Returns resolved value or <code>null</code>
+     */
+    default String getStringDatabaseValue(Row row) {
+        return row.isNull(QueryConstants.STRING_VALUE_COLUMN_NAME) ? null : row.getString(QueryConstants.STRING_VALUE_COLUMN_NAME);
```
dimas-b, line 43: Why call `row.isNull`? Doesn't `getString` return java `null` or the column in null in the row?

restapi/src/main/java/io/stargate/web/docsapi/service/query/predicate/AnyValuePredicate.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package io.stargate.web.docsapi.service.query.predicate;
+
+/**
+ * Special type of predicate that can test a generic value against all database values.
+ *
+ * @param <V> Type of a value used in the predicate.
+ */
+public interface AnyValuePredicate<V> extends StringFilterPredicate<V>, DoubleFilterPredicate<V>, BooleanValuePredicate<V> {
```
dimas-b, line 24: nit: rename to `ValuePredicate`? "Any" in this context sounds a bit like a predicate that _allows_ any value? WDYT?

restapi/src/main/java/io/stargate/web/docsapi/service/query/predicate/BasePredicate.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * @return Returns raw value of the predicate, as user would specify in the query, f.e. <code>$eq</code>
+     */
+    String getRawValue();
+
+    /**
+     * @return If this predicate can provide database predicate.
+     */
+    default boolean canProvideDatabasePredicate() {
```
dimas-b, line 36: nit: rename to `hasDatabasePredicate`?

restapi/src/main/java/io/stargate/web/docsapi/service/query/provider/ConditionProviderService.java
EricBorczuk: I think this could be really nice to have, and extends well to anything that works over HTTP (maybe even graphql, though typically the HTTP status code is always 200 - but I could imagine it using the Exception's default message to make a response)

The only thing I could see becoming unwieldy with this setup is that the enum `ErrorCode` could get pretty long, but for the most part I feel like that can be fixed by keeping each enum value in some kind of order.

restapi/src/main/java/io/stargate/web/docsapi/service/query/provider/ListConditionProvider.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+        return Optional.empty();
+    }
+
+    private List<?> getListConditionValues(Iterator<JsonNode> iterator) {
+        List<Object> result = new ArrayList<>();
+        iterator.forEachRemaining(node -> {
+            // TODO Eric? copied int branch from the Document service, is this actually wanted?
+
+            if (node.isInt()) {
```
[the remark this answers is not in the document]
ivansenic: yea, I can simply do:

```
            if (node.isNumber()) {
                result.add(node.numberValue());
            }
```

restapi/src/main/java/io/stargate/web/docsapi/service/query/provider/ListConditionProvider.java
ivansenic: Yep, I will also note the operation and the node type passed.

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/BaseCondition.java
ivansenic: No, throws exception.

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/BaseCondition.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /**
+     * Resolves {@link Boolean} value from the document {@link Row}.
+     *
+     * @param row Row
+     * @param numericBooleans If <code>true</code> assumes booleans are stored as numeric values.
+     * @return Returns resolved value or <code>null</code>
+     */
+    default Boolean getBooleanDatabaseValue(Row row, boolean numericBooleans) {
```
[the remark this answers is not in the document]
ivansenic: It is in the state, at least here in conditions.. However, you can not make every component dependable on whoever is having the state.. Thus we need to pass it around a bit..

restapi/src/main/java/io/stargate/web/docsapi/service/query/predicate/AnyValuePredicate.java
ivansenic: Yes it will be something like `ValueFilterOperation` or smth like that at the end. Fine?

ivansenic:
So I am moving this to the review ready, there are couple of left `TODO`s, but that can be fixed along the way. Let's see once we actually use those expressions.. I applied fixes for all the existing comments btw.. And there is one failing test on purpose.

@EricBorczuk Can we resolve the situation around the exists condition? I have seen that this condition was using the `EQ` predicate and was targeting the `bool_value` column there.. So I am not sure how is this actually working, how would this apply once you have a field that is string and why would this target only the bool value? We can also take this into a quick call.

@dimas-b Since the PR is already huge, I did not add the 3 rules we talked about yesterday. I did rename everything per your suggestion.

ivansenic:
@EricBorczuk there is also an open issue with the comparing nulls wrt `gt, gte, lt, gte`. Are `null`s considers as last or not? Meaning is `null` greater than everything?

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/impl/ExistsCondition.java
```diff
@@ -0,0 +1,35 @@
...
+
+// TODO is this class needed at all? Can it provide a built in condition?
+//  if yes add test
+@Value.Immutable
+public abstract class ExistsCondition extends CombinedCondition<Boolean> implements BaseCondition {
+
+  /** {@inheritDoc} */
+  @Override
+  public Optional<BuiltCondition> getBuiltCondition() {
+    // TODO what's real predicate for the exists
```
EricBorczuk, line 32: The predicate for `exists` would just be that the row trivially exists...maybe you can match against the `leaf` column? But technically as of now, for `{"a.b":{ "$exists": true}}`

A query such as `select ... from table where p0 = 'a' and p1 = 'b' and leaf = 'b'` is performed, and each row that's returned is a matching document.

I'm not sure how easy that is to translate into a BuiltCondition...

dimas-b: > `select ... from table where p0 = 'a' and p1 = 'b' and leaf = 'b'`

Technically this query will also allow docs with an existing property like `a.b.c.b`, right? Is that indented?

dimas-b:
> there is also an open issue with the comparing nulls wrt gt, gte, lt, gte. Are nulls considers as last or not? Meaning is null greater than everything?

In traditional databases comparing `null` with anything always yields `false` if I'm not mistaken.

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/impl/ExistsCondition.java
ivansenic: This is perfect, I can easily integrate that. But just to confirm you are effectively ignoring the `exists` atm with: https://github.com/stargate/stargate/blob/87998bcbf3aaec481f7cd318f32cac26d08f421b/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java#L1203

```

```

dimas-b:
Re: failing tests, let fix or ignore them before merging... I believe they will only be a distraction when working with the code on the feature branch.

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/impl/ExistsCondition.java
ivansenic: > > `select ... from table where p0 = 'a' and p1 = 'b' and leaf = 'b'`
> 
> Technically this query will also allow docs with an existing property like `a.b.c.b`, right? Is that indented?

No because during the queering we are "nulling: the rest of the `p`s:
`select ... from table where p0 = 'a' and p1 = 'b' and p2 = '' and p3 = ''  and leaf = 'b'`

But I guess only one empty check is needed imo..

dimas-b: In that ^ case checking `leaf` is redundant, I think. The query might be more efficient without it too, since `leaf` is not part of the PK, WDYT?

restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterPath.java
```diff
@@ -0,0 +1,72 @@
...
+
+  /** @return The complete path to the field, including the field name as the last element. */
+  @Value.Parameter
+  List<String> getFullPath();
+
+  /** @return The name of the field. Effectively, the last element of {@link #getPath()}. */
+  default String getField() {
+    List<String> path = getFullPath();
+    if (null == path || path.isEmpty()) {
+      return null;
```
dimas-b, line 35: Is this not an error to have a `FilterPath` without any path elements? Even a field at the root level should have one path element `==` the field's name, right?

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/BaseCondition.java
```diff
@@ -0,0 +1,75 @@
...
+        : row.getString(QueryConstants.STRING_VALUE_COLUMN_NAME);
+  }
+
+  /**
+   * Resolves {@link Double} value from the document {@link Row}.
+   *
+   * @param row Row
+   * @return Returns resolved value or <code>null</code>
+   */
+  default Double getDoubleDatabaseValue(Row row) {
```
dimas-b, line 49: another naming nit: would a simple `getDouble(Row)` work? Having `Row` as a paramter kind of implies it's a database value. WDYT?

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/impl/CombinedCondition.java
```diff
@@ -0,0 +1,86 @@
...
+
+    if (filterOperation.isMatchAll()) {
+      return filterOperation.test(queryValue, dbValueBoolean)
+          && filterOperation.test(queryValue, dbValueDouble)
+          && filterOperation.test(queryValue, dbValueString);
+    } else {
+      return filterOperation.test(queryValue, dbValueBoolean)
+          || filterOperation.test(queryValue, dbValueDouble)
+          || filterOperation.test(queryValue, dbValueString);
+    }
```
dimas-b, line 84: What is the use case for testing all the value columns from a single row? AFAIK, only one of them should be non-null in any given row. :thinking:

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionService.java
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  /**
+   * Parses the root filter json node and returns back the list of the expressions found.
+   *
+   * @param prependedPath Given collection or document path segments
+   * @param filterJson Filter JSON node
+   * @return List of all expressions
+   */
+  public List<Expression<FilterExpression>> parse(
```
mpenick, line 83: Maybe this should have `@VisibleForTesting`?

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/impl/ExistsFilterOperation.java
```diff
@@ -0,0 +1,81 @@
...
+import io.stargate.db.query.Predicate;
+import io.stargate.web.docsapi.exception.DocumentAPIRequestException;
+import io.stargate.web.docsapi.service.query.filter.operation.CombinedFilterOperation;
+import java.util.Optional;
+import org.immutables.value.Value;
+
+/** Exists filter operation can resolve if any database value exists. */
+@Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE)
+@Value.Immutable(singleton = true)
+public abstract class ExistsFilterOperation implements CombinedFilterOperation<Boolean> {
```
dimas-b, line 28: Is this really a filter operation? 

Conceptually it should pass any row that it received on input since the row obviously exists at that point?

Should we perhaps handle "exists" at the query parsing level? We could produce something like a `FilterExpression` with conditions only on the primary key columns (also mentioned in another comment on this PR). WDYT?

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/BooleanValueFilterOperation.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package io.stargate.web.docsapi.service.query.filter.operation;
+
+/**
+ * Predicate that can compare values against a {@link Boolean} database value.
+ *
+ * @param <I> Type of the filter value to compare to.
+ */
+public interface BooleanValueFilterOperation<I> extends BaseFilterOperation {
```
dimas-b, line 24: Isn't `I` always `Boolean`?

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/BaseFilterOperation.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Optional;
+
+/** Base predicate that can be used in the filter expression. */
+public interface BaseFilterOperation {
+
+  /**
+   * @return Returns raw value of the predicate, as user would specify in the query, f.e. <code>$eq
+   *     </code>
+   */
+  String getRawValue();
```
dimas-b, line 29: Should we use an `enum` for the returned values perhaps? 

Also, how about renaming to `getOpCode` or something like that? "Raw" seems too vague, eh?

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/BaseFilterOperation.java
```diff
@@ -0,0 +1,38 @@
...
+package io.stargate.web.docsapi.service.query.filter.operation;
+
+import io.stargate.db.query.Predicate;
+import java.util.Optional;
+
+/** Base predicate that can be used in the filter expression. */
+public interface BaseFilterOperation {
+
+  /**
+   * @return Returns raw value of the predicate, as user would specify in the query, f.e. <code>$eq
```
dimas-b, line 26: `f.e.` == `e.g.`?

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/ConditionProviderService.java
```diff
@@ -0,0 +1,103 @@
...
+import io.stargate.web.docsapi.service.query.filter.operation.impl.*;
+import java.util.*;
+
+/**
+ * Simple service that wraps all available raw filter values and connects them to a {@link
+ * ConditionProvider}.
+ *
+ * <p>Currently uses a unmodifiable map, thus it's thread safe.
+ */
+public class ConditionProviderService {
```
dimas-b, line 34: nit: how about `ConditionParser`? "Service" in Stargate is historically something that responds to client requests directly... but it's up to you.

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/ConditionProviderService.java
```diff
@@ -0,0 +1,103 @@
...
+   * Creates the conditions for the node containing the raw filter ops as the keys. For example:
+   * <code>{ "$gt: { 5 }, "$lt": { 10 }}</code>.
+   *
+   * @param conditionsNode Node containing the filter ops as keys
+   * @param numericBooleans If numeric boolean should be applied to the created conditions
+   * @return Collection of created conditions.
+   * @throws io.stargate.web.docsapi.exception.DocumentAPIRequestException If filter op is not
+   *     found, condition constructions fails or filter value is not supported by the filter op.
+   */
+  public Collection<BaseCondition> getConditions(JsonNode conditionsNode, boolean numericBooleans) {
```
dimas-b, line 53: Could `numericBooleans` be an instance-level field in this class?

restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterExpression.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  @Override
+  public Expression<FilterExpression> replaceVars(
+      Map<FilterExpression, Expression<FilterExpression>> m,
+      ExprFactory<FilterExpression> exprFactory) {
+    if (m.containsKey(this)) {
+      return m.get(this);
+    }
+    return this;
```
dimas-b, line 89: Should it always be `this` since we do not use jbool variables at all?

restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterExpression.java
```diff
@@ -0,0 +1,91 @@
...
+import com.bpodgursky.jbool_expressions.rules.RuleList;
+import com.bpodgursky.jbool_expressions.util.ExprFactory;
+import io.stargate.web.docsapi.service.query.condition.BaseCondition;
+import java.util.*;
+import java.util.function.Function;
+import org.immutables.value.Value;
+
+/** Filter expression that stands as the node in the Expression tree. */
+@Value.Immutable
+public abstract class FilterExpression extends Expression<FilterExpression> {
```
dimas-b, line 30: Conceptually the type parameter of `Expression` appears to be the type of expression variable _names_, I think. Cf. `Variable.replaceVars`.

So, in our case it can probably be `Void` since we do not use variables.

Alternatively, since filters apply "conditions" to document properties, this type parameter could reasonably be `FilterPath`, I guess. The idea being that we should be able to resolve an `Expression<FilterPath>` if we have known values for all paths referenced by it (and its child expressions).

WDYT?

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionService.java
```diff
@@ -0,0 +1,154 @@
...
+import io.stargate.web.docsapi.exception.DocumentAPIRequestException;
+import io.stargate.web.docsapi.service.query.condition.BaseCondition;
+import io.stargate.web.docsapi.service.query.condition.ConditionProviderService;
+import java.util.*;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import javax.ws.rs.core.PathSegment;
+import org.apache.commons.lang3.StringUtils;
+
+public class ExpressionService {
```
dimas-b, line 34: naming nit: `ExpressionParser`?

dimas-b (commented):
LGTM overall :+1: 

By biggest concern at this point is `CombinedCondition`. I'm just not sure what the use case is for that condition.

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/impl/CombinedCondition.java
ivansenic: Yea you are right, considering there is only one value available, I can refactor this and even simplify.

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/impl/ExistsFilterOperation.java
ivansenic: It is. From the filter operation it's checks that the value is not null. 

Combined with the condition and the rule that each row must have a value, yes each row will pass it. But I don't care much, we need anyway to parse it and place it in the tree, thus I am keeping it as is ;)

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/BooleanValueFilterOperation.java
ivansenic: It's not:
- for the in predicate `<I>` is `List<?>`
- for the exists it's `Boolean` but for all operations, so `StringFilterOperation<Boolean>`

restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterExpression.java
ivansenic: I would agree here. I was also mentioning this in the talk we had. Let's fix it to `Void` for now, and then let's see along the path if this would change. OK?

ivansenic:
> LGTM overall +1
> 
> By biggest concern at this point is `CombinedCondition`. I'm just not sure what the use case is for that condition.

So `CombinedCondition` is currently used for `$in` and `$nin` where filter value is a `List<?>`. Thus I need to match potentially any of those three database values to a list. Let's picture this in a simple example:

* `EqFilterOperation` can match:
   * string vs string
   * bool vs bool
   * number vs double
*  `InFilterOperation` can match: 
   * List<?> vs string 
   * List<?> vs bool
   * List<?> vs double

* `StringCondition` uses `EqFilterOperation` to compare string filter value to a string db value (we already know we can limit to string only as the user specified string as the filter value)
* `CombinedCondition` uses `InFilterOperation` to compare list filter value to a string or double or boolean db value ( we can not limit to single db value as filter input is different from target db type)

In theory we could have only one `FilterOperation` generic interface and one generic `Condition`, as far as testing against a row is concerned. But we would not be able to construct `BuiltConditions` then and would need to cast the filter value to find a proper one and this would be too dirty.

Clearer now? Maybe naming is still not good enough?

tomekl007:
@ivansenic FYI CI is failing

restapi/src/main/java/io/stargate/web/docsapi/service/query/provider/ConditionProviderService.java
ivansenic: In separate ticket.

ivansenic:
> > there is also an open issue with the comparing nulls wrt gt, gte, lt, gte. Are nulls considers as last or not? Meaning is null greater than everything?
> 
> In traditional databases comparing `null` with anything always yields `false` if I'm not mistaken.

Thanks, applied that. Except for equal and notEqual they still can compare against a `null`.

restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterPath.java
ivansenic: Good catch, adding a check.. And changing the impl everywhere!

restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterPath.java
```diff
@@ -0,0 +1,72 @@
...
+    } else {
+      int length = path.size();
+      return path.get(length - 1);
+    }
+  }
+
+  /** @return The path without the field. Effectively, the last element of {@link #getPath()}. */
+  default List<String> getPath() {
```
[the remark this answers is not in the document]
ivansenic: no it's perfect, with every comment a PR gets better.. please never hesitate to write comments on my PRs! :pray:

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/BaseFilterOperation.java
```diff
@@ -0,0 +1,38 @@
...
+   */
+  String getRawValue();
+
+  /** @return If this predicate can provide database predicate. */
+  default boolean hasDatabasePredicate() {
+    return getDatabasePredicate().isPresent();
+  }
+
+  /** @return Mirrored database predicate, if one exists. */
+  Optional<Predicate> getDatabasePredicate();
```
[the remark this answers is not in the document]
ivansenic: Yep, better

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/BaseFilterOperation.java
ivansenic: but are we getting anything with this, yea we limit filter operations to return a fixed enum value, but in fact since we need to parse it, we are adding additional complexity here.. If the enum would bring anything else to the table it would be fine I guess.. or?

ivansenic: ok connected condition provider to the enum, even better now.. Resolved!

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/ConditionProviderService.java
ivansenic: I like it, done.

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/ConditionProviderService.java
ivansenic: ok, even better, removed from the expression service then..

restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterExpression.java
ivansenic: not sure if this is used in the eval later on? let s keep it like this until we actually find the usage?

restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterExpression.java
ivansenic: OK decided not to do it at the end. Then I can not collect all the filter expression from the tree and this was a very neat feature of having it like this.. Let's figure it out along the way..

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionService.java
ivansenic: done

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/impl/ExistsFilterOperation.java
ivansenic: Actually you are correct here.. This is not a real filter op. Removed!

ivansenic:
@EricBorczuk @dimas-b I resolved all remaining points, did lot of refactoring due to the Dmitri naming suggestions.. The current state seems really good to me..  If you give me a final approval I would merge on the green build :man_dancing:

restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterExpression.java
dimas-b: My guess is that it is actually not used during the expression simplification process. It looks like it will only be used when you call it explicitly, so we might even be able to throw an `UnsupportedOperationException` here to make it clear that we do not use this functionality.

As for me, the current code is harmless from the runtime perspective, but it requires the reader to try and understand it and that is where confusion sets it. So, avoiding unused code is preferable IMHO.

restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterExpression.java
dimas-b: Ah, so you use the `collectK` method for gathering all the `FilterExpression` objects? Yes, that's a nice approach. Sorry, I missed that.

dimas-b (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterExpression.java
ivansenic: throwing `UnsupportedOperationException`, that's fine.. I like fail fast..

dimas-b (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/ConditionParser.java
```diff
@@ -0,0 +1,77 @@
...
+            FilterOperationCode code = operationCode.get();
+            Optional<? extends BaseCondition> condition =
+                code.getConditionProvider().createCondition(valueNode, numericBooleans);
+            if (condition.isPresent()) {
+              results.add(condition.get());
+            } else {
+              // condition empty
+              String msg =
+                  String.format(
+                      "Operation %s is not supporting the provided value %s.",
```
mpenick, line 65: "... does not support the ..."

mpenick:
@ivansenic @dimas-b What do you think about this simplification to combined operations and conditions?  

https://github.com/stargate/stargate/commit/57c9d189c17ec4a3de07d700f179a3920721340e

mpenick:
@ivansenic @dimas-b What do you think about this simplification of combined ops and conditions? https://github.com/stargate/stargate/commit/54847ee3d5a6ea7ef3d0b35a555b636766ba8225

mpenick:
^ If you do that it also allows for simplification of value filters https://github.com/stargate/stargate/commit/a621efed894abd27e96124b6cd61b30b21ec8d69

mpenick (commented):
Looks great, well tested. Here's a branch https://github.com/stargate/stargate/commits/simplify-combined-ops with some possible simplifications. I think the type (String, Double, Boolean) specific filter operation interfaces could be collapsed into a single class and they don't need to be generic.

Also, there should be a an "in" filter operation test that uses a non-double number type:

```java
    @Test
    public void numberInNotDouble() {
      List<?> filterValue = Collections.singletonList(42); // Not a double

      boolean result = in.test(filterValue, 42d);

      assertThat(result).isTrue();
    }
```

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/provider/impl/BasicConditionProvider.java
```diff
@@ -0,0 +1,71 @@
...
+import io.stargate.web.docsapi.service.query.condition.impl.*;
+import io.stargate.web.docsapi.service.query.condition.provider.ConditionProvider;
+import io.stargate.web.docsapi.service.query.filter.operation.BooleanValueFilterOperation;
+import io.stargate.web.docsapi.service.query.filter.operation.DoubleValueFilterOperation;
+import io.stargate.web.docsapi.service.query.filter.operation.StringValueFilterOperation;
+import java.util.Optional;
+
+/**
+ * The basic {@link ConditionProvider} that usually handles all standard filter operations that
+ * works will String, Boolean and Number filter value.
```
mpenick, line 30: will --> with

ivansenic:
> Looks great, well tested. Here's a branch https://github.com/stargate/stargate/commits/simplify-combined-ops with some possible simplifications. I think the type (String, Double, Boolean) specific filter operation interfaces could be collapsed into a single class and they don't need to be generic.
> 
> Also, there should be a an "in" filter operation test that uses a non-double number type:
> 
> ```java
>     @Test
>     public void numberInNotDouble() {
>       List<?> filterValue = Collections.singletonList(42); // Not a double
> 
>       boolean result = in.test(filterValue, 42d);
> 
>       assertThat(result).isTrue();
>     }
> ```

@mpenick hey thanks for the suggestions.. they look quite interesting and good.. would you mind if we:

* merge the current state
* open PR with your simplifications on top
* this way I can check out in a cleaner way all the changes and we can discus them?

I would be free to rebase and push with force your branch if you don't mind? Or you can do it and open PR as it's your work anyway?

mpenick (approved):
+1

mpenick:
> I would be free to rebase and push with force your branch if you don't mind? Or you can do it and open PR as it's your work anyway?

PR #885, opened 2021-04-23
Start adding paging

dougwettlaufer:


grpc/src/main/java/io/stargate/grpc/payload/cql/ValuesHandler.java
```diff
@@ -29,6 +42,10 @@
   @Override
   public BoundStatement bindValues(Prepared prepared, Payload payload, ByteBuffer unsetValue)
       throws InvalidProtocolBufferException, StatusException {
+    if (!payload.hasValue()) {
```
dougwettlaufer, line 45: @mpenick is this valid or do I have to provide a value?

mpenick: Fixed and added test here: https://github.com/stargate/stargate/pull/852/commits/f3e5bbc6dd7ae8243d545659aad7069b65ca3b29

dougwettlaufer: So every request has to have a payload?

mpenick: No, good call, I'm tired. :) 

I think we can move handling for empty payloads up higher so that the handlers don't have to have duplicate logic: https://github.com/stargate/stargate/pull/852/commits/80aff752d4cc593e9183202b97d4dc50df6175ba

I open that PR up for review. I'll try to get that in ASAP so that we can get some more concurrency going.

mpenick (approved):
LGTM

dimas-b (approved):

PR #886, opened 2021-04-24
Fix type checks when a condition targets a mapped UDT

olim7t:


graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/EntityModel.java
```diff
@@ -155,42 +155,33 @@ public boolean isFederated() {
...
                 this.getGraphqlName(), describePartitionKey()));
       }
-      WhereConditionModel condition = maybeCondition.get();
-      if (condition.getPredicate() != Predicate.EQ && condition.getPredicate() != Predicate.IN) {
-        return Optional.of(
-            String.format(
-                "the only predicates allowed for partition key fields are EQ and IN "
-                    + "(got %s for %s).",
-                condition.getPredicate(), field.getGraphqlName()));
-      }
```
olim7t, line 193: Removed this because static checks on predicates are now all done at deployment time in `WhereConditionModelBuilder`.

As a reminder, `validateNoFiltering` is invoked both at deployment time, and every time a query is executed (because if the arguments are optional, omitting some of them changes the set of conditions).

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/migration/CreateIndexQuery.java
```diff
@@ -52,7 +52,7 @@ public Table getTable() {
   @Override
   public String getDescription() {
     return String.format(
-        "Create %s index on %s.%s", index.name(), table.name(), index.column().name());
+        "Create index %s on %s.%s", index.name(), table.name(), index.column().name());
```
olim7t, line 55: Only tangentially related, but I noticed it while testing. It's all going to be squashed when we merge the `graphql-first` PR anyway.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/EntityModel.java
```diff
@@ -155,42 +155,33 @@ public boolean isFederated() {
 
   /**
    * Validates a set of conditions to ensure that they will produce a valid CQL query that doesn't
-   * require ALLOW FILTERING.
+   * require ALLOW FILTERING for this entity.
    *
    * @return an optional error message if the validation failed, empty otherwise.
    */
-  public Optional<String> validateConditions(Collection<WhereConditionModel> conditions) {
+  public Optional<String> validateNoFiltering(Collection<WhereConditionModel> conditions) {
```
olim7t, line 162: Renamed for clarity. In `WhereConditionModelBuilder` we validate that argument types match their fields, and that the predicates are allowed. Here we validate that the set of conditions forms a valid query.

dimas-b (approved):

dimas-b:
For some reason `MetricsTest.restApiHttpRequestMetrics` failed in DSE CI, but I do not see how this PR can affect it... but the test seemed ok on `master`. @ivansenic : WDYT?

olim7t:
> For some reason `MetricsTest.restApiHttpRequestMetrics` failed in DSE CI, but I do not see how this PR can affect it... but the test seemed ok on `master`. @ivansenic : WDYT?

It seems to be failing randomly. I got the same error on C* 4 on another PR. I don't understand how that happens since it's just checking metric names.

ivansenic:
> > For some reason `MetricsTest.restApiHttpRequestMetrics` failed in DSE CI, but I do not see how this PR can affect it... but the test seemed ok on `master`. @ivansenic : WDYT?
> 
> It seems to be failing randomly. I got the same error on C* 4 on another PR. I don't understand how that happens since it's just checking metric names.

I ll check tomorrow to see if there is maybe an async falvor to the metric reporting and a possible record on a non-http thread. Nothing else comes on my mind here.

ivansenic:
Added small changes for the test in #902

PR #888, opened 2021-04-24
Handle where with  using types other than text fixes #887

### LINKED ISSUE #887: Unable to use "$in" with timestamp field
Given a table with a timestamp column, this request currently fails

```sh
localhost:8082/v2/keyspaces/testks/foo?where={"id": {"$eq":"454674e1-571d-45bc-bd46-be6e9b6165ed"}, "created": {"$in": ["2021-04-21T18:42:22.139Z"]}}
```

The reason for this is that no matter what the array for the `IN` condition is being treated as an array of Strings. Instead we need to convert to an array of whatever the type is for that column.

dougwettlaufer:
Fixes #887

### DIFF
--- a/restapi/src/test/java/io/stargate/web/resources/WhereParserTest.java
+++ b/restapi/src/test/java/io/stargate/web/resources/WhereParserTest.java
@@ -23,10 +23,12 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import io.stargate.db.query.Predicate;
 import io.stargate.db.query.builder.BuiltCondition;
 import io.stargate.db.schema.Column;
+import io.stargate.db.schema.Column.Type;
 import io.stargate.db.schema.ImmutableColumn;
 import io.stargate.db.schema.ImmutableTable;
 import io.stargate.web.service.WhereParser;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


ivansenic (approved):
Nice fix :+1:

restapi/src/test/java/io/stargate/web/resources/WhereParserTest.java
```diff
@@ -288,4 +308,24 @@ public void testInOperation() throws IOException {
...
     assertThat(where).isEqualTo(whereExpected);
   }
+
+  @Test
+  public void testInWithTimestampOperation() throws IOException {
+    String whereParam = "{\"created\":{\"$in\":[\"2021-04-23T18:42:22.139Z\"]}}";
+    List<BuiltCondition> whereExpected =
+        singletonList(
+            BuiltCondition.of(
+                "created", Predicate.IN, singletonList(Instant.parse("2021-04-23T18:42:22.139Z"))));
```
eribeiro, line 318: nit: maybe use `Converters.toCqlValue(type, object)` instead of directly calling `Instant`?

mpenick (approved):

PR #890, opened 2021-04-26
Json Conversion Refactor: Part 2

EricBorczuk:
- [x] Use DI to inject JsonConverter and ObjectMapper to DocumentService
- [x] Reformat DocumentServiceTest to no longer spy on DocumentService
- [x] Couple of small revisions from previous PR w.r.t convertToJsonDoc
- [x] Add unit tests to cover more behavior in JsonConverter

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
@@ -5,7 +5,6 @@ import static io.stargate.web.docsapi.resources.RequestToHeadersMapper.getAllHea
 import com.datastax.oss.driver.api.core.NoNodeAvailableException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.annotations.VisibleForTesting;
 import io.stargate.auth.UnauthorizedException;
 import io.stargate.db.schema.Keyspace;
 import io.stargate.db.schema.Table;
@@ -63,18 +62,8 @@ public class DocumentResourceV2 {
   @Inject private Db dbFactory;
   private static final Logger logger = LoggerFactory.getLogger(DocumentResourceV2.class);
   private static final ObjectMapper mapper = new ObjectMapper();
-  private final DocumentService documentService;
+  @Inject private DocumentService documentService;
   private final int DEFAULT_PAGE_SIZE = DocumentDB.SEARCH_PAGE_SIZE;
-
-  public DocumentResourceV2() {
-    documentService = new DocumentService();
-  }
-
-  @VisibleForTesting
-  DocumentResourceV2(Db dbFactory, DocumentService documentService) {
-    this.dbFactory = dbFactory;
-    this.documentService = documentService;
-  }
 
   @POST
   @ManagedAsync
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
@@ -14,6 +14,7 @@ import io.stargate.web.docsapi.examples.WriteDocResponse;
 import io.stargate.web.docsapi.exception.DocumentAPIRequestException;
 import io.stargate.web.docsapi.exception.ResourceNotFoundException;
 import io.stargate.web.docsapi.models.DocumentResponseWrapper;
+import io.stargate.web.docsapi.service.DocsApiConfiguration;
 import io.stargate.web.docsapi.service.DocumentService;
 import io.stargate.web.docsapi.service.filter.FilterCondition;
 import io.stargate.web.models.Error;
@@ -61,9 +62,9 @@ import org.slf4j.LoggerFactory;
 public class DocumentResourceV2 {
   @Inject private Db dbFactory;
   private static final Logger logger = LoggerFactory.getLogger(DocumentResourceV2.class);
-  private static final ObjectMapper mapper = new ObjectMapper();
+  @Inject private ObjectMapper mapper;
   @Inject private DocumentService documentService;
-  private final int DEFAULT_PAGE_SIZE = DocumentDB.SEARCH_PAGE_SIZE;
+  @Inject private DocsApiConfiguration docsApiConfiguration;
 
   @POST
   @ManagedAsync
@@ -664,7 +665,7 @@ public class DocumentResourceV2 {
                     dbFactory.getDataStore(),
                     pageStateParam,
                     pageSizeParam,
-                    pageSizeParam > 0 ? pageSizeParam : DEFAULT_PAGE_SIZE);
+                    pageSizeParam > 0 ? pageSizeParam : docsApiConfiguration.getSearchPageSize());
             DocumentDB db = dbFactory.getDocDataStoreForToken(authToken, getAllHeaders(request));
             JsonNode result =
                 documentService.searchDocumentsV2(
@@ -763,7 +764,10 @@ public class DocumentResourceV2 {
 
           final Paginator paginator =
               new Paginator(
-                  dbFactory.getDataStore(), pageStateParam, pageSizeParam, DEFAULT_PAGE_SIZE);
+                  dbFactory.getDataStore(),
+                  pageStateParam,
+                  pageSizeParam,
+                  docsApiConfiguration.getSearchPageSize());
 
           JsonNode results;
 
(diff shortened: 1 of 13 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -673,12 +669,11 @@ public JsonNode searchDocumentsV2(
...
           JsonNode jsonDoc =
-              getJsonConverterService()
-                  .convertToJsonDoc(
-                      Collections.singletonList(row),
-                      new DeadLeafCollectorNoOp(),
-                      true,
-                      db.treatBooleansAsNumeric());
+              jsonConverterService.convertToJsonDoc(
+                  Collections.singletonList(row),
+                  new DeadLeafCollectorNoOp(),
```
eribeiro, line 674: Instead of creating a new `DeadLeafCollectorNoOp` each time, you could just create it once and use it everywhere. 🙂 
It would be nice if this was "non instantiable" that could be reused across the code. My suggestions:

1) Make it a `enum` as below and call it `DeadLeafCollectorNoOp.INSTANCE`
```
public enum DeadLeafCollectorNoOp implements DeadLeafCollector {
     INSTANCE;
     (...)
 ```
 
 2) Make it a singleton  non instantiable class as below:
 ```
 public final class DeadLeafCollectorNoOp implements DeadLeafCollector {
     private DeadLeafCollectorNoOp() {}
     
     private static final DeadLeafCollectorNoOp INSTANCE = new DeadLeafCollectorNoOp();
     
     public static DeadLeafCollectorNoOp getInstance() {
                return INSTANCE;
     }
     (...)
 ```

Comment (eribeiro) [reply restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java:674]: I'd also recommend to have a second signature to this method that doesn't have the DeadLeafCollector for the cases it's not needed.
```
public JsonNode convertToJsonDoc(List<Row> rows, boolean writeAllPathsAsObjects, boolean numericBooleans) {
       return convertToJsonDoc(rows, DeadLeafCollector.INSTANCE, writeAllPathsAsObjects, numericBooleans);
}
```

Comment (ivansenic) [reply restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java:676]: `@Value.Immutable()` && `@Value.Style(Singleton = true)` is also good option for singletons?

File: restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java [hunk 2]
```diff
@@ -63,19 +62,9 @@
   @Inject private Db dbFactory;
   private static final Logger logger = LoggerFactory.getLogger(DocumentResourceV2.class);
   private static final ObjectMapper mapper = new ObjectMapper();
```
ivansenic, line 64: this can also be injected as well right? It's in the env..

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -63,19 +62,9 @@
...
-  private final DocumentService documentService;
+  @Inject private DocumentService documentService;
   private final int DEFAULT_PAGE_SIZE = DocumentDB.SEARCH_PAGE_SIZE;
```
ivansenic, line 66: Since we are already here with the dependency injection, can I propose the extraction of configuration options to:

```
interface DocsApiConfiguration {

    default getDefaultPageSize() {
        return DocumentDB.SEARCH_PAGE_SIZE;
    }

}
```

Then you can:
* register the configuration in the DI
* if and when needed fine tune config options in the test by mocking the configuration interface

Makes sense?

restapi/src/main/java/io/stargate/web/impl/Server.java
```diff
@@ -133,7 +135,20 @@ protected void configure() {
     environment.jersey().register(IndexesResource.class);
 
     // Documents API
+    JsonConverter jsonConverter = new JsonConverter(environment.getObjectMapper());
```
ivansenic, line 138: another suggestion:

1. let's register all classes, `JsonConverter` as well
2. let's extract all of this into a DocsApi bBinder class

``` 
public class DocsApiComponentsBinder extends AbstractBinder {

    private final Environment environment; // if needed, I guess not

    protected void configure() {
        /// all services and resources needed for the docs api
        ....
    }

} 

and then in the server just:

```
environment.jersey().register(new DocsApiComponentsBinder(environment));
```

clean and easy

File: restapi/src/test/java/io/stargate/web/docsapi/resources/DocumentResourceV2Test.java [hunk 5]
```diff
@@ -34,19 +34,22 @@
...
-  private DocumentResourceV2 documentResourceV2;
+  @InjectMocks private DocumentResourceV2 documentResourceV2;
+  @Mock private DocumentService documentServiceMock;
+  @Mock private Db dbFactoryMock;
   private HttpServletRequest httpServletRequest;
 
   @BeforeEach
   public void setup() {
-    documentResourceV2 = new DocumentResourceV2(dbFactoryMock, documentServiceMock);
+    MockitoAnnotations.initMocks(this);
```
ivansenic, line 52: you can use the `@ExtendWith(MockitoExtension.class)` on the class level

restapi/src/test/java/io/stargate/web/docsapi/service/DocumentServiceTest.java
```diff
@@ -89,12 +90,13 @@
...
   private Method checkLtOp;
   private Method searchRows;
+
   private static final ObjectMapper mapper = new ObjectMapper();
   private static final Map<String, String> EMPTY_HEADERS = Collections.emptyMap();
 
   @BeforeEach
   public void setup() throws NoSuchMethodException {
-    service = new DocumentService();
+    MockitoAnnotations.initMocks(this);
```
ivansenic, line 99: same here..

ivansenic (commented):
Stone by stone :european_castle:

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
EricBorczuk: Yes - I just missed this somehow! Thanks

restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverter.java
```diff
@@ -94,18 +110,6 @@ public JsonNode convertToJsonDoc(
         JsonNode childRef;
 
         if (isArray && !writeAllPathsAsObjects) {
-          if (!ref.isArray()) {
```
EricBorczuk, line 97: This is dead code, I added it as a safeguard long ago but it's never actually hit

restapi/src/test/java/io/stargate/web/docsapi/service/json/JsonConverterTest.java
```diff
@@ -1,130 +0,0 @@
...
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.node.BooleanNode;
-import com.fasterxml.jackson.databind.node.ObjectNode;
-import io.stargate.db.datastore.Row;
-import io.stargate.web.docsapi.service.DocumentServiceTest;
-import java.util.*;
-import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Test;
-
-public class JsonConverterTest {
```
dimas-b, line 16: Why does this change show up as a delete + add as opposed to a file move? Was there a lot of code changes in this file?

EricBorczuk: I did add unit tests, but other than that there weren't changes...so I'm not sure

restapi/src/main/java/io/stargate/web/docsapi/service/json/DeadLeafCollectorNoOp.java
```diff
@@ -1,13 +1,6 @@
...
-  public void addArray(String path) {}
-
-  public void addAll(String path) {}
-
-  public boolean isEmpty() {
-    return true;
-  }
-}
+@Value.Immutable(singleton = true)
+public interface DeadLeafCollectorNoOp extends DeadLeafCollector {}
```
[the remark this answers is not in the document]
EricBorczuk: I think that's what I had before, I'll revert :)

restapi/src/main/java/io/stargate/web/docsapi/service/json/DeadLeafCollector.java
```diff
@@ -1,11 +1,16 @@
 package io.stargate.web.docsapi.service.json;
 
+import org.immutables.value.Value;
+
+@Value.Immutable(singleton = true)
```
dimas-b, line 5: Interesting.... :thinking: I was thinking about something like `static DeadLeafCollector NOP = new DeadLeafCollector() {};` but `@Value.Immutable(singleton = true)` works too, I guess :)

dimas-b (approved):

Issue #130, opened 2020-09-24
Built-in C* aggregation functions


blerghh:
A common use case is to see the number of products in a cart. The functions are min, max, avg, sum, and count

Proposed format:

```
type Cart {
  cartId: TimeUuid @PrimaryKey
  userId: Uuid
  _count(field: CartFields): Integer
  _sum_float(field: CartFields): Float
  _min_float(field: CartFields): Float
  _max_float(field: CartFields): Float
  _avg_float(field: CartFields): Float
  _sum_int(field: CartFields): BigInteger
  _min_int(field: CartFields): BigInteger
  _max_int(field: CartFields): BigInteger
  _avg_int(field: CartFields): BigInteger
}

query {
  cart(value: {pk:"pk1"}) {
    values {
       cartId
       count: _count(field: id)
    }
  }
}
```

olim7t:
Related to #137 and #138: same issue on what to pass as the argument, handling collisions with column names...

tomekl007:
@olim7t just to double-check, this is for the `graphql-frist` API, right?

olim7t:
This is for CQL-first.

TTL and WRITETIME are not aggregations, I was just mentioning them because they share similarities as CQL functions. This ticket is specifically about functions like `count`, `max`, etc.

I think we need a little bit of research to refine the syntax. There are a few issues with the proposal above:

1. it requires a specific field for every (function, return type) combination. This doesn't seem very scalable in the long term, especially considering custom functions. At the very least we could take the function name as an argument:
    ```
    count: _int_function(name: "count", field: id)
    ```
    We still need one field per return type, but that's an improvement. We could go further but at the cost of type safety (e.g. return everything as strings, or a custom `Any` scalar), so I'm not sure if it's a good idea.

2. Having an enum for field names (`CartFields` in the example) sounds like a good idea at first, but it assumes that all function arguments will be field names, which is true right now but might not be forever (again, consider custom functions). Also, it's not possible to pass multiple arguments with the proposed syntax. Maybe we could go with something more generic like passing all arguments in a plain string.

tomekl007:
1. Good suggestion regarding passing the function name as a string. I think that's a good direction.
Regarding the return type. As far as I see, all functions (`count`, `sum`, `min`, `max`, `avg`) that we want support now are operating on `numeric` values, right? The graphQL allows only two numeric types: Float and Int: 
[...]

PR #901, opened 2021-04-28
Cql first build in aggregations

tomekl007:
in progress, todo:
- [x] Integration test and _int_function type definition in `schema.json`
- [x] is blocked by https://github.com/stargate/stargate/issues/907 (implementing in-progress)
- [x] handle `count()`
- [x] handle `min()`
- [x] handle `max()`
- [x] handle `avg()`
- [x] handle `sum()`
- [x] integration tests of all supported function types (_int_function, _bigint_function, etc)
- [x] improve query builder functions support
- [x] migrate ITs to a new infrastructure
- [x] handle case-sensitive functions' columns


testing/src/main/graphql/betterbotz/schema.json
```diff
@@ -5054,6 +5054,48 @@
               },
               "isDeprecated": false,
               "deprecationReason": null
+            },
+            {
+                "name": "_int_function",
```
tomekl007, line 5059: I had trouble with modeling the `https://github.com/stargate/stargate/pull/901/files#diff-cf660cd57cf1a5b7f88859dec98db42f352e3daa32a50ac012966a29c57e005eR13` in this `schema.json`.
@olim7t Is this what we want to achieve? (the current schema compiles properly)

testing/src/main/java/io/stargate/it/http/GraphqlTest.java
```diff
@@ -2119,6 +2121,50 @@ public void udtsTest() {
...
+    assertThat(getObservable(client.mutate(order2Mutation)).getInsertOrders().isPresent()).isTrue();
+
+    // execute aggregation with count()
+    GetOrdersWithCountQuery countQuery = GetOrdersWithCountQuery
+            .builder()
+            .value(order1)
+            .aggregationField("id")
+            .build();
+
+    GetOrdersWithCountQuery.Data result = getObservable(client.query(countQuery));
```
tomekl007, line 2163: Currently, this test fails with:
```
INFO  [OkHttp http://127.0.2.1:8080/...] 2021-04-28 15:06:11,189 GraphqlTest.java:2278 - GraphQL error found in test: [Validation error of type FieldUndefined: Field '_int_function' in type 'Orders' is undefined @ 'Orders/values/_int_function']
```
The submitted schema is:
```
INFO  [Exec Stream Pumper] 2021-04-28 15:06:11,176 ProcessRunner.java:78 - stargate0-0> WARN  [dw-133 - POST /graphql/betterbotz] 2021-04-28 15:06:11,176 GraphQL.java:508 - Query failed to validate : 'query GetOrdersWithCount($value: OrdersInput, $aggregationField: String!) { Orders(value: $value) { __typename pageState values { __typename id prodId prodName customerName address description price sellPrice count: _int_function(name: "count", args: [$aggregationField]) } } }'
```
If I understand this correctly, we need to generate now the `_int_function` at the java-backend side, right?
However, it seems that the validation happens before it goes to our logic. @olim7t Could you point me to a place where this should be handled?

olim7t: Yes, you need to implement the new field before you can reference it in a test.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/QueryFetcher.java
```diff
@@ -93,16 +97,63 @@ private BoundQuery buildQuery(DataFetchingEnvironment environment, DataStore dat
...
+    }
+    String column = getDBColumnName(table, args.get(0));
+    if (column == null) {
+      throw new IllegalArgumentException(
+          String.format(
+              "The column name: %s provided for the %s function does not exists.",
+              args.get(0), COUNT));
+    }
+
+    // todo after https://github.com/stargate/stargate/issues/907 is done
```
tomekl007, line 154: we need https://github.com/stargate/stargate/issues/907 to have aggregations available on the builder.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/DmlSchemaBuilder.java
```diff
@@ -528,9 +526,35 @@ public GraphQLObjectType buildType(Table table) {
       }
     }
 
+    buildAggregationFunctions(builder);
```
tomekl007, line 529: I assume that we want to add a possibility for all supported functions for all tables.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/QueryFetcher.java
```diff
@@ -93,16 +97,63 @@ private BoundQuery buildQuery(DataFetchingEnvironment environment, DataStore dat
...
+        dataStore
+            .queryBuilder()
+            .select()
+            .column(buildQueryColumns(environment))
+            .from(table.keyspace(), table.name())
+            .where(buildClause(table, environment))
+            .limit(limit)
+            .orderBy(buildOrderBy(environment));
+
+    addAggregationFunctions(environment, queryBuilder);
```
tomekl007, line 110: I think that we can leverage an existing `QueryFetcher` without the need for a new `AggregatedFunctionsFetcher`, @olim7t wdyt?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/QueryFetcher.java
```diff
@@ -84,6 +94,31 @@ public QueryFetcher(
...
+      for (SelectedField selectedField : valuesField.getSelectionSet().getFields()) {
+        if (selectedField.getName().equals(INT_FUNCTION)) {
+          String alias = selectedField.getAlias();
+          // put the returned value as alias
+          if (alias != null) {
+            columns.put(alias, row.getInt(alias));
+          }
+          //
+          else {
+            // if there is no alias, use the value as is.
```
tomekl007, line 113: I wonder if we should disallow using function calls without aliases. I think it may be reasonable as it offers a lot better UX for end-users than handling a generated function name. On the other hand, I am not sure if this is not too limiting. @olim7t wdyt?

testing/src/main/java/io/stargate/it/http/graphql/cqlfirst/GraphqlApolloTest.java
```diff
@@ -2100,6 +2102,232 @@ public void udtsTest() {
...
+
+    GetOrdersAllFunctionsQuery.Value res =
+        functionsResult.getOrders().get().getValues().get().get(0);
+    assertThat(res.get_int_function().get()).isEqualTo(value.intValue());
+    assertThat(res.get_double_function().get()).isEqualTo(value.doubleValue());
+    assertThat(res.get_bigint_function().get()).isEqualTo(value.toString());
+    assertThat(res.get_decimal_function().get()).isEqualTo(value.toString());
+    assertThat(res.get_varint_function().get()).isEqualTo(value.toString());
+    assertThat(res.get_float_function().get()).isEqualTo(value.doubleValue());
+    assertThat(res.get_smallint_function().get()).isEqualTo(BigDecimal.valueOf(value.shortValue()));
```
tomekl007, line 2327: the byte and short values are returned as `BigDecimal`:
```
smallint_function = {Optional@7638} "Optional[100]"
 value = {BigDecimal@7648} "100"
_tinyint_function = {Optional@7639} "Optional[100]"
 value = {BigDecimal@7650} "100"
```
It seems suspicious. @olim7t is it intended or do we have bugs in both `smallint` and `tinyint` GraphQl conversions?

graphqlapi/src/main/java/io/stargate/graphql/schema/scalars/IntCoercing.java
```diff
@@ -74,7 +74,7 @@ private DriverTypeT fromBigIntegerSafe(BigInteger i) {
 
   protected void checkRange(Number n, long min, long max) {
     long l = n.longValue();
-    if (l < Byte.MIN_VALUE || l > Byte.MAX_VALUE) {
+    if (l < min || l > max) {
```
tomekl007, line 77: there was a bug in this range check - both min and max were ignored. However, this function is called by both short and byte checks, so we need to use those arguments. Fixed

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/SupportedGraphqlFunction.java
```diff
@@ -0,0 +1,47 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.graphql.schema.cqlfirst.dml.fetchers.aggregations;
+
+public enum SupportedGraphqlFunction {
+  INT_FUNCTION("_int_function"),
```
dougwettlaufer, line 19: Do we have to specify a function for each type? We can't have something more generic and infer the type from the column?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
```diff
@@ -0,0 +1,210 @@
...
+  private final Table table;
+  private final DbColumnGetter dbColumnGetter;
+
+  public AggregationsFetcherSupport(NameMapping nameMapping, Table table) {
+    this.table = table;
+    this.dbColumnGetter = new DbColumnGetter(nameMapping);
+  }
+
+  public void addAggregationFunctions(
+      DataFetchingEnvironment environment, QueryBuilder.QueryBuilder__20 queryBuilder) {
```
dougwettlaufer, line 48: Would it be possible to pass the dataStore and return a QueryBuilder? Or Something to that effect? That way we can avoid passing around this autogenerated `QueryBuilder__20` and this will be easier to unit test?

testing/src/main/graphql/betterbotz/aggregations/GetOrdersAllFunctions.graphql
```diff
@@ -0,0 +1,23 @@
...
+        values {
+            id
+            prodId
+            prodName
+            customerName
+            address
+            description
+            price
+            sellPrice
+            _int_function(name: "max", args: ["value_int"])
```
dougwettlaufer, line 13: Related to my comment above, could these be `function(name: "max", args: ["value_int"])` or `numeric_function(name: "max", args: ["value_int"])`?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
```diff
@@ -0,0 +1,210 @@
...
+              "The column name: %s provided for the %s function does not exists.",
+              args.get(0), COUNT));
+    }
+    return column;
+  }
+
+  private List<String> getAndValidateArgs(
+      Map<String, Object> arguments, SupportedAggregationFunction supportedAggregationFunction) {
+    @SuppressWarnings("unchecked")
+    List<String> args = (List<String>) arguments.get("args");
```
dougwettlaufer, line 185: Is it possible for `arguments` or `args` to be null?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+  }
+
+  private String getAndValidateColumn(List<String> args) {
+    String column = dbColumnGetter.getDBColumnName(table, args.get(0));
+    if (column == null) {
+      throw new IllegalArgumentException(
+          String.format(
+              "The column name: %s provided for the %s function does not exists.",
+              args.get(0), COUNT));
```
dougwettlaufer, line 177: I don't think you want `COUNT` here, right?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/SupportedGraphqlFunction.java
tomekl007: yes, we need it. See this discussion for more details:
https://github.com/stargate/stargate/issues/130#issuecomment-825999565

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
tomekl007: good catch, fixed

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
tomekl007: good question, it cannot be null. I will add a null check.

tomekl007: done in: https://github.com/stargate/stargate/pull/901/commits/409b5cf79bca42bc17a20e5633e942e34139ed2c

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
tomekl007: I've tried to do it, but the problem is that it needs to operate on the pre-created builder in the `QueryFetcher`:
```

    QueryBuilder.QueryBuilder__20 queryBuilder =
        dataStore.queryBuilder().select().column(buildQueryColumns(environment));
```
I am aware that this is a bit problematic. I will try to find a better way, but I am not sure if we can find a cleaner and better solution.

dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
```diff
@@ -0,0 +1,222 @@
...
+
+  private String getAndValidateColumn(
+      List<String> args, SupportedAggregationFunction supportedAggregationFunction) {
+    String column = dbColumnGetter.getDBColumnName(table, args.get(0));
+    if (column == null) {
+      throw new IllegalArgumentException(
+          String.format(
+              "The column name: %s provided for the %s function does not exists.",
+              args.get(0), supportedAggregationFunction.getName()));
+    }
```
olim7t, line 179: We have a problem with case-sensitive columns here. For example, given the table:
```
CREATE TABLE foo(k int, cc int, "V" int, PRIMARY KEY (k, cc));
```
There is currently no way to write the GraphQL equivalent of:
```
SELECT count("V") FROM foo WHERE k=1 ;
```
If you pass the column name unquoted in the GraphQL query (as in `_bigint_function(name: "count", args: "V")`), Stargate tries to execute the invalid CQL `SELECT count(v)`.
If you try to quote the column argument (as in `_bigint_function(name: "count", args: "\"V\"")`), then this check fails.

I think the best approach is to expect the first form (exact case, unquoted). That way it's consistent with the DML API, and also the less surprising since it's exactly the same as the GraphQL field name.
We can add the quotes before we pass the argument to the query builder (no need to check if the name is mixed-case or not, `COUNT(v)` and `COUNT("v")` are equivalent, so we can just quote all the time).

Column names are not quoted in CQL result sets, so I _think_ `AggregationFetcherSupport.generateAggregationColumnName` will do the right thing. The best way to check is to add an integration test with case-sensitive names.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
```diff
@@ -0,0 +1,222 @@
...
+          String.format(
+              "The args provided for a function: %s are null, but it is not allowed.",
+              supportedAggregationFunction.getName()));
+    }
+    @SuppressWarnings("unchecked")
+    List<String> args = (List<String>) argsToValidate;
+    if (args.size() != 1) {
+      throw new IllegalArgumentException(
+          String.format(
+              "The %s function takes only one argument, " + "but more arguments: %s were provided.",
```
olim7t, line 197:
```suggestion
              "The %s function takes only one argument, but more arguments: %s were provided.",
```

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
```diff
@@ -0,0 +1,222 @@
...
+              supportedAggregationFunction.getName(), args));
+    }
+    return args;
+  }
+
+  private SupportedAggregationFunction getNameArgument(Map<String, Object> arguments) {
+    Object nameArgument = arguments.get("name");
+    if (nameArgument == null) {
+      throw new IllegalArgumentException("The function name cannot be null.");
+    }
```
olim7t, line 207: nit: this cannot happen, the argument is non-null in the schema.

graphqlapi/src/main/java/io/stargate/graphql/schema/scalars/IntCoercing.java
olim7t: 👍 good catch

testing/src/main/graphql/betterbotz/aggregations/GetOrdersAllFunctions.graphql
olim7t, line 13: Sorry about the conflicts, but I'm about to merge #906 and these new tests should definitely use the new "lightweight" approach.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/DmlSchemaBuilder.java
```diff
@@ -528,9 +520,43 @@ public GraphQLObjectType buildType(Table table) {
...
+        .argument(
+            GraphQLArgument.newArgument()
+                .name("name")
+                .description("Name of the function to invoke")
+                .type(new GraphQLNonNull(GraphQLString)))
+        .argument(
+            GraphQLArgument.newArgument()
+                .name("args")
+                .description("Arguments passed to a function. It can be a list of column names.")
+                .type(new GraphQLList(new GraphQLNonNull(GraphQLString))))
```
olim7t, line 555: Since `AggregationsFetcherSupport.getAndValidateArgs` fails if this argument is missing, we can make it non-null as well:
```suggestion
                .type(new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(GraphQLString)))))
```

If we support arg-less functions (like `now()`) in the future,  we might want to revisit this, but I'm fine with it since going from non-null to nullable is not breaking.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/DmlFetcher.java
```diff
@@ -150,16 +152,11 @@ protected Parameters getDatastoreParameters(DataFetchingEnvironment environment)
   }
 
   protected String getDBColumnName(Table table, String fieldName) {
-    Column column = getColumn(table, fieldName);
-    if (column == null) {
-      return null;
-    }
-    return column.name();
+    return dbColumnGetter.getDBColumnName(table, fieldName);
```
olim7t, line 155: Maybe we can make `dbColumnGetter` protected and inline and remove these methods?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
```diff
@@ -0,0 +1,222 @@
...
+
+  private List<String> getAndValidateArgs(
+      Map<String, Object> arguments, SupportedAggregationFunction supportedAggregationFunction) {
+    Object argsToValidate = arguments.get("args");
+    if (argsToValidate == null) {
+      throw new IllegalArgumentException(
+          String.format(
+              "The args provided for a function: %s are null, but it is not allowed.",
+              supportedAggregationFunction.getName()));
+    }
```
olim7t, line 191: Maybe remove this check if you apply my other suggestion of making it non-null in the schema.

olim7t (changes requested):

testing/src/main/java/io/stargate/it/http/graphql/cqlfirst/GraphqlApolloTest.java
olim7t: This might be specific to the Apollo client library. These types map to custom scalars. Apollo can't guess what the actual range is, so I imagine when it sees a number it picks the most general representation.
There are ways to register [custom adapters](https://www.apollographql.com/docs/android/essentials/custom-scalar-types/) on the client, but I'm not sure if it's worth it for these tests (especially given that I think they shouldn't use Apollo).

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
```diff
@@ -0,0 +1,222 @@
...
+      Map<String, Object> columns, DataFetchingEnvironment environment, Row row) {
+    List<SelectedField> valuesFields = environment.getSelectionSet().getFields("values");
+    if (valuesFields.isEmpty()) {
+      return columns;
+    }
+    for (SelectedField valuesField : valuesFields) {
+      for (SelectedField selectedField : getAllFieldsWithNameArg(valuesField)) {
+        switch (SupportedGraphqlFunction.valueOfIgnoreCase(selectedField.getName())) {
+          case INT_FUNCTION:
+            putResultValue(columns, row, selectedField, Row::getInt);
```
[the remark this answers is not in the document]
tomekl007: nice idea, done

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
tomekl007: good suggestions. Implemented it: https://github.com/stargate/stargate/pull/901/commits/75551240edd632a75ea51bfb8aea71ae5c23be18

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  public Map<String, Object> addAggregationResults(
+      Map<String, Object> columns, DataFetchingEnvironment environment, Row row) {
+    List<SelectedField> valuesFields = environment.getSelectionSet().getFields("values");
+    if (valuesFields.isEmpty()) {
+      return columns;
+    }
+    for (SelectedField valuesField : valuesFields) {
+      for (SelectedField selectedField : getAllFieldsWithNameArg(valuesField)) {
```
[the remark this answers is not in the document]
tomekl007: It is needed because the `SupportedGraphqlFunction.valueOfIgnoreCase`
throws if the name of the function is not found:
```
 throw new UnsupportedOperationException(
        String.format("The graphQL aggregation function: %s is not supported.", functionName));
```
I can change it to a less strict approach, and return optional from it. Next, execute the `putResultValue` only if the optional value is present. wdyt?

testing/src/main/java/io/stargate/it/http/graphql/cqlfirst/GraphqlApolloTest.java
tomekl007: yeah, that was it. Resolving

testing/src/main/graphql/betterbotz/aggregations/GetOrdersAllFunctions.graphql
tomekl007: converted to a new approach.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
tomekl007: good catch, fixed in:
https://github.com/stargate/stargate/pull/901/commits/ab21d68bf19b6f3f9a26380b2753830efee827af

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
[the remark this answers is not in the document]
olim7t: Yeah I think that would be cleaner, there could be other fields with a `name` attribute in the future.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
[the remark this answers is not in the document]
olim7t: 👍 
Sorry to only bring this up now, but a similar change could be done in `buildAggregatedFunctions()` above.

olim7t (approved):
LGTM, thanks for addressing my feedback.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/aggregations/AggregationsFetcherSupport.java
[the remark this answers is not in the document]
tomekl007: Ok, I will add the optional. However, we still need `getAllFieldsWithNameArg` because it provides the deduplication of fields (please note that it returns Set). The `valuesFields` contains a duplicate of every entry.

PR #902, opened 2021-04-28
small updates for the metrics test

ivansenic:
I tried actually getting this test to fail, but was no successful.. Since we pass all metric lines to the assertion, I changed it a bit so that we first do a small filter. Not that I am questioning the lib correctness, but more to see better error messages if we git it again.

### DIFF
--- a/DEV_GUIDE.md
+++ b/DEV_GUIDE.md
@@ -94,7 +94,13 @@ env JAVA_OPTS='-Dmy_property="some value"' ./starctl --cluster-name 'Some Cluste
 
 ### Debugging
 
-If you're an IntelliJ user you can use start the project with
+If you're an IntelliJ user you can create the *JAR Application* run configuration, pointing to the `stargate-lib/stargate-starter-[VERSION].jar` and specifying `stargate-lib/` as the working directory.
+
+Then please disable **Instrumenting agent** in `Settings | Build, Execution, Deployment | Debugger | Async Stacktraces`.
+This will allow you to debug directly using the IntelliJ debug run option.
+You can debug any run configuration and tests as well.
+
+#### Remote debugging
 
 ```sh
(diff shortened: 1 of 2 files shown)


dimas-b (approved):

testing/src/main/java/io/stargate/it/http/MetricsTest.java
```diff
@@ -67,12 +73,14 @@ public void restApiHttpRequestMetrics() throws IOException {
 
     String result = RestUtils.get("", String.format("%s:8084/metrics", host), HttpStatus.SC_OK);
 
-    String[] lines = result.split(System.getProperty("line.separator"));
+    List<String> lines = Arrays.stream(result.split(System.getProperty("line.separator")))
+            .filter(line -> line.startsWith("http_server_requests"))
+            .collect(Collectors.toList());
```
olim7t, line 76: nit: this is repeated a few times, maybe you could extract a `filter(String[] lines, String prefix)` method.

olim7t (approved):

testing/src/main/java/io/stargate/it/http/MetricsTest.java
ivansenic: So I tend not to extract common code for testing, as:

* reading a single test becomes more complicated, as you need to navigate to "helper" methods in order to realize what's going on
* changes in the extracted peaces start to influence too many tests
* of course this does not stand true for the some INT test setup for example

Would it be OK to leave it like this?

dougwettlaufer (approved):

DEV_GUIDE.md
```diff
@@ -94,7 +94,13 @@ env JAVA_OPTS='-Dmy_property="some value"' ./starctl --cluster-name 'Some Cluste
 
 ### Debugging
 
-If you're an IntelliJ user you can use start the project with
+If you're an IntelliJ user you can create the *JAR Application* run configuration, pointing to the `stargate-lib/stargate-starter-[VERSION].jar` and specifying `stargate-lib/` as the working directory.
```
mpenick, line 97: Nice!

PR #903, opened 2021-04-28
Never enable debug in Stargate nodes in test

olim7t:
It's currently impossible to debug an integration test that starts a Stargate node. The node fails with this error:
```
ERROR [Exec Stream Pumper] 2021-04-28 09:45:12,329 ProcessRunner.java:85 - stargate0-0> ERROR: transport error 202: connect failed: Connection refused
ERROR [Exec Stream Pumper] 2021-04-28 09:45:12,329 ProcessRunner.java:85 - stargate0-0> ERROR: JDWP Transport dt_socket failed to initialize, TRANSPORT_INIT(510)
ERROR [Exec Stream Pumper] 2021-04-28 09:45:12,329 ProcessRunner.java:85 - stargate0-0> JDWP exit error AGENT_ERROR_TRANSPORT_INIT(197): No transports initialized [debugInit.c:750]
```
I'm not sure exactly why this happens, but the whole idea of starting the nodes with debug sounds dubious to me:
* ~~if you are debugging the test, you already have a debugging session in the IDE. How would you start another one?~~
* ~~how do you deal with `suspend=y` when there are multiple Stargate nodes?~~

edit: you _can_ run multiple parallel debugging sessions in IDEA I guess... Still I'm not sure I often I would use that in practice.

I propose we remove the options altogether. At least that will give us the ability to set breakpoints in the test code, and observe the behavior of the the Stargate process from the client POV.
To debug the Stargate code, you'll need to start a standalone instance and execute the scenario manually.

### DIFF
--- a/testing/src/main/java/io/stargate/it/storage/StargateContainer.java
+++ b/testing/src/main/java/io/stargate/it/storage/StargateContainer.java
@@ -16,7 +16,6 @@
 package io.stargate.it.storage;
 
 import static io.stargate.starter.Starter.STARTED_MESSAGE;
-import static java.lang.management.ManagementFactory.getRuntimeMXBean;
 
 import com.datastax.oss.driver.api.core.Version;
 import io.stargate.it.exec.OutputListener;
@@ -214,11 +213,6 @@ public class StargateContainer extends ExternalResource<StargateSpec, StargateCo
     throw new IllegalStateException("Unknown parameter: " + pc);
   }
 
-  private static boolean isDebug() {
-    String args = getRuntimeMXBean().getInputArguments().toString();
-    return args.contains("-agentlib:jdwp") || args.contains("-Xrunjdwp");
-  }
-
   protected static class Container extends ExternalResource.Holder
       implements StargateEnvironmentInfo, AutoCloseable {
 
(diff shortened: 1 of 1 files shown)


ivansenic (commented):
Hah, I had a same issue today. Well, first having no debug possibility in test is terrible. So let me write what needs to be done.

You basically need to create and run the remote debugging session from the IDE before starting the test, with following config:

image

If you do this, your test debug will work. In addition, see the stuff from my metrics test PR to avoid class not found issues with OSGi:

Then please disable **Instrumenting agent** in `Settings | Build, Execution, Deployment | Debugger | Async Stacktraces`.
This will allow you to debug directly using the IntelliJ debug run option.
You can debug any run configuration and tests as well.

olim7t:
Wow, TIL 🤯 

I just tested it and having separate sessions is not that bad, especially given that we only have one Stargate node in most tests.

I'll close this but we need better documentation. This wasn't high priority for me but always annoying, I wish I had known the solution earlier.

dougwettlaufer:
Super helpful, thanks @ivansenic! Would you mind adding that to the docs?

ivansenic:
> Super helpful, thanks @ivansenic! Would you mind adding that to the docs?

@dougwettlaufer sure, the part for the Instrumenting agent is already added in the #902.. I ll added the picture with desc in a new PR..

PR #909, opened 2021-04-29
resolve possible request not flushed in metrics tests

ivansenic:
OK I figured out what could be causing the problem.. The call to the `client.newCall(request).execute()` seems not to guarantee that the request has been fully done on the other side:

```
Note that transport-layer success (receiving a HTTP response code, headers and body) does not necessarily indicate application-layer success: response may still indicate an unhappy HTTP response code like 404 or 500.
```

Thus I guess we can end up in the situation where the call is not yet registered in the metrics.. I fixed that by asserting on the status and body, closing the resources as well.

I am sorry I caused trouble in everyone's PRs.

### DIFF
--- a/testing/src/main/java/io/stargate/it/http/MetricsTest.java
+++ b/testing/src/main/java/io/stargate/it/http/MetricsTest.java
@@ -30,7 +30,6 @@ import java.util.List;
 import java.util.stream.Collectors;
 import net.jcip.annotations.NotThreadSafe;
 import okhttp3.*;
-import org.apache.commons.lang.mutable.MutableInt;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpStatus;
 import org.junit.jupiter.api.BeforeAll;
@@ -68,13 +67,7 @@ public class MetricsTest extends BaseOsgiIntegrationTest {
             .addHeader(TagMeHttpMetricsTagProvider.TAG_ME_HEADER, "test-value")
             .build();
 
-    MutableInt status = new MutableInt();
-    try (Response execute = client.newCall(request).execute()) {
-      assertThat(execute.body()).isNotNull();
-      assertThat(execute.code()).isNotZero();
-      status.setValue(execute.code());
-    }
-
+    int status = execute(client, request);
     String result = RestUtils.get("", String.format("%s:8084/metrics", host), HttpStatus.SC_OK);
 
     List<String> lines =
@@ -89,7 +82,7 @@ public class MetricsTest extends BaseOsgiIntegrationTest {
                     .contains("method=\"GET\"")
                     .contains("module=\"restapi\"")
                     .contains("uri=\"/v1/keyspaces\"")
-                    .contains(String.format("status=\"%d\"", status.intValue()))
+                    .contains(String.format("status=\"%d\"", status))
                     .contains(TagMeHttpMetricsTagProvider.TAG_ME_KEY + "=\"test-value\""));
   }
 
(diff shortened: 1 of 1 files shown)


testing/src/main/java/io/stargate/it/http/MetricsTest.java
```diff
@@ -66,7 +67,13 @@ public void restApiHttpRequestMetrics() throws IOException {
             .get()
             .addHeader(TagMeHttpMetricsTagProvider.TAG_ME_HEADER, "test-value")
             .build();
-    Response execute = client.newCall(request).execute();
+
+    MutableInt status = new MutableInt();
```
tomekl007, line 71: Maybe just `AtomicInteger`?

mpenick: `MutableInt` is a good fit here. `AtomicInteger` says this might be used on another thread and would cause me to go looking around for multi-threaded stuff.

mpenick (approved):

mpenick:
@ivansenic I know you tend to not like helpers in tests, but I think this cleans it up nicely. Then you don't have to use `MutableInt`.

```java
  private int execute(OkHttpClient client, Request request) throws IOException {
    try (Response execute = client.newCall(request).execute()) {
      assertThat(execute.body()).isNotNull();
      assertThat(execute.code()).isNotZero();
      return execute.code();
    }
  }
```

In the code:
```
 int status = execute(client, request);

// Use status
```

tomekl007 (approved):

testing/src/main/java/io/stargate/it/http/MetricsTest.java
tomekl007: Yes, sure. But `MutableInt` is a class from an external (3rd party) library, whereas the `AtomicInteger` already exists in the JDK.

testing/src/main/java/io/stargate/it/http/MetricsTest.java
```diff
@@ -66,7 +67,13 @@ public void restApiHttpRequestMetrics() throws IOException {
...
+    try (Response execute = client.newCall(request).execute()) {
+      assertThat(execute.body()).isNotNull();
+      assertThat(execute.code()).isNotZero();
+      status.setValue(execute.code());
```
dimas-b, line 75: Does `status` have to be a `MutableInt`? It appears to be accessed only by the test thread, so I guess it cab be a simple `int`. WDYT?

mpenick: The problem is it's used in a lambda below so it needs to be final. I suggested using a helper to with return value so it's not mutated.

dimas-b: Would it make sense to extract the response reading code into a method in this case and write to `status` on the declaration line?

dimas-b: Also, please add a comment about the need to full read and close the response :) Same as what we have in the PR, but it would be nice to have it in code too, please :)

testing/src/main/java/io/stargate/it/http/MetricsTest.java
ivansenic: `int` will work, no idea how `MutableInt` came on my keyboard :exploding_head:

testing/src/main/java/io/stargate/it/http/MetricsTest.java
ivansenic: `int` will work, I can extract...

dimas-b (approved):

dougwettlaufer (approved):

PR #911, opened 2021-04-29
Use the existing datastore cache pattern with the remainder of the rest resources

### LINKED ISSUE #551: Use the auth caching in docsapi in the remainder of restapi
Make use of the token cache in restapi that's currently used in the docsapi

https://github.com/stargate/stargate/blob/master/restapi/src/main/java/io/stargate/web/resources/Db.java#L44-L48

dougwettlaufer:
resolves #551

### DIFF
--- a/restapi/src/main/java/io/stargate/web/resources/Db.java
+++ b/restapi/src/main/java/io/stargate/web/resources/Db.java
@@ -21,16 +21,13 @@ import io.stargate.auth.AuthenticationService;
 import io.stargate.auth.AuthenticationSubject;
 import io.stargate.auth.AuthorizationService;
 import io.stargate.auth.UnauthorizedException;
-import io.stargate.db.Parameters;
 import io.stargate.db.datastore.DataStore;
 import io.stargate.db.datastore.DataStoreFactory;
 import io.stargate.db.datastore.DataStoreOptions;
 import io.stargate.web.docsapi.dao.DocumentDB;
-import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.concurrent.CompletionException;
 import java.util.stream.Collectors;
 
@@ -45,6 +42,13 @@ public class Db {
           .maximumSize(10_000)
           .expireAfterWrite(Duration.ofMinutes(1))
           .build(this::getDocDataStoreForTokenInternal);
+
+  private final LoadingCache<TokenAndHeaders, AuthenticatedDB> restTokensToDataStore =
+      Caffeine.newBuilder()
+          .maximumSize(10_000)
+          .expireAfterWrite(Duration.ofMinutes(1))
+          .build(this::getRestDataStoreForTokenInternal);
+
   private final DataStoreFactory dataStoreFactory;
 
   public Db(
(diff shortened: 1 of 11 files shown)


dougwettlaufer:
@EricBorczuk there's still some places in the docsapi that go directly to `io.stargate.web.resources.Db#getDataStoreForToken` but I didn't want to touch them since you're in the middle of a large refactor.

restapi/src/main/java/io/stargate/web/resources/Db.java
```diff
@@ -45,6 +42,13 @@
           .maximumSize(10_000)
           .expireAfterWrite(Duration.ofMinutes(1))
           .build(this::getDocDataStoreForTokenInternal);
+
+  private final LoadingCache<TokenAndHeaders, AuthenticatedDB> restTokensToDataStore =
+      Caffeine.newBuilder()
+          .maximumSize(10_000)
```
tomekl007, line 48: Could you elaborate more on why we pick 10_000 and expire after write 1 minute?

restapi/src/main/java/io/stargate/web/resources/Db.java
```diff
@@ -124,9 +95,35 @@ private DocumentDB getDocDataStoreForTokenInternal(TokenAndHeaders tokenAndHeade
...
+
+  public AuthenticatedDB getRestDataStoreForToken(String token, Map<String, String> headers)
       throws UnauthorizedException {
-    return authenticationService.validateToken(tokenAndHeaders.token, tokenAndHeaders.headers);
+    if (token == null) {
+      throw new UnauthorizedException("Missing token");
+    }
+
+    try {
+      return restTokensToDataStore.get(TokenAndHeaders.create(token, headers));
```
dimas-b, line 120: I wonder if using _all_ headers a part of the cache key can prevent hits... if one of the headers is irrelevant, but changes its value often, will it effectively make the cache ineffective?

dimas-b: I see that `TokenAndHeaders` filter out all but the `Host` header, but conceptually only the `AuthenticationService` knows that the relevant headers are... Should we somehow delegate the filtering of headers to `AuthenticationService`  for the sake of correctness?

dimas-b: By similar logic, should the `AuthenticationService` control the cache duration? Otherwise, there's risk of continuing to allow a user to execute requests after the token expires. WDYT?

dimas-b (approved):
Added some auth-related concerns in comments, but I think this PR preserves old caching behaviour, so it's ok to merge from my POV. We can address authentiation concerns in a follow-up PR.

restapi/src/main/java/io/stargate/web/resources/Db.java
dougwettlaufer: Following the pattern @EricBorczuk had. Seemed like reasonable defaults.

restapi/src/main/java/io/stargate/web/resources/Db.java
dougwettlaufer: Probably a good idea. I'll create a new issue.

dougwettlaufer: https://github.com/stargate/stargate/issues/915

PR #913, opened 2021-04-30
Add possibility to invoke C* aggregation functions on DataStore.queryBuilder()

tomekl007:


persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
```diff
@@ -88,7 +88,7 @@
       @SubExpr(
           name = "select",
           definedAs =
-              "select star? column* writeTimeColumn? from (where* perPartitionLimit? limit? orderBy*) allowFiltering?"),
+              "select star? column* writeTimeColumn? count? max? min? sum? avg? from (where* perPartitionLimit? limit? orderBy*) allowFiltering?"),
```
dimas-b, line 91: CQL allows multiple functions in the same select, AFAIK.

What do you think about a grammar like `(colium|funciton)*`, where `function` => `count|min|max|avg|sum`?

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
```diff
@@ -346,6 +357,51 @@ public void writeTimeColumn(String columnName, String alias) {
     this.writeTimeColumnAlias = alias;
   }
 
+  public void count(String countColumnName) {
```
dimas-b, line 360: I believe all of these new methods need the `@DSLAction` annotation.

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
```diff
@@ -346,6 +357,51 @@ public void writeTimeColumn(String columnName, String alias) {
...
+    count(countColumnName, null);
+  }
+
+  public void count(String countColumnName, String alias) {
```
dimas-b, line 364: If we're introducing selector aliases, why not support them for ordinary columns too?

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
dimas-b: ... and for supporting aliases, how about this: `((column|function) as?)*`?

We could keep the last selector (column or function) in a field on `QueryBuilderImpl` so that the `as` method can access it.

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
dimas-b, line 360: The `column` methods support passing a `Column` object as a parameter. Could we allow that for `count` and standard aggregates too?

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
```diff
@@ -1445,23 +1501,48 @@ protected BuiltSelect selectQuery() {
...
-                if (writeTimeColumnAlias != null) {
-                  builder.append("AS").append(cqlName(writeTimeColumnAlias));
-                }
-              })
+          .addIfNotNull(wtColumn, c -> addWriteTime(builder, c))
+          .addIfNotNull(countColumn, c -> addCount(builder, c))
+          .addIfNotNull(maxColumn, c -> addMax(builder, c))
+          .addIfNotNull(minColumn, c -> addMin(builder, c))
+          .addIfNotNull(sumColumn, c -> addSum(builder, c))
+          .addIfNotNull(avgColumn, c -> addAvg(builder, c))
```
dimas-b, line 1542: Could we handle functions in a generic way to avoid repetition?

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
tomekl007: I am basing that on the existing `writeTimeColumn?` -> it had the alias support but this is not specified here.
Should it be? If yes, how it worked if the AS for `writeTime` was not present?

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
tomekl007: Again - I am doing the same as for writeTimestamp. I don't need an alias for other columns. If we have a demand for this feature, we can address that in the separate PR.

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
tomekl007: suggestions welcome :) I've refactored the same code to `addFunctionCallWithAliasIfPresent` method

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
tomekl007: Sure, we can. However, I don't need an API that accepts `Column` in the ticket that needed this one (https://github.com/stargate/stargate/pull/901/files) so I am not sure if we should add it.

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
dimas-b: Re: multiple function I meant something like `select sum(x), sum(y) from table`... as far as I understand current grammar does not allow expressing that via the `QueryBuilder`.

dimas-b: Re: aliases, perhaps we do not need them for ordinary columns, but I thought it would be nice to extend that functionality to all selector kids just for the sake of completeness... totally optional, though.

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
dimas-b: added a comment about alias to the thread on line 91... resolving this conversation.

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
dimas-b: I do not have a concrete suggestion, but basically I was thinking about converting the `selection` field from a `List<String>` to a `List<Selector>`, `Selector` being a private class that encapsulates both simple columns and functions (perhaps in sub-classes).

tomekl007: I don't think that this PR adds a lot of duplication - It follows the same pattern. If we are not fine with the duplication in this class, I think we should handle that separately, in a different PR that focuses exclusively on refactoring and this one should be scoped to adding a new feature. Is that ok?

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
dimas-b: I guess it's a matter of convenience. In API modules I imagine we might have a `Column` object handy, so it might be worth supporting `Column` parameters if only to avoid using `Column.name()` all the time. WDYT?

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
dimas-b: Indeed, current impl. is fine with the current grammar from my POV.

However, if we support a grammar like `(column|function)*` then I think having more generic code for functions would be worth the effort.

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
tomekl007: Ok, I see. I will implement that support for N same invocations.
However, I don't see what you mean by 
> We could keep the last selector (column or function) in a field on QueryBuilderImpl so that the as method can access it.
I think that having one `as()` method is not enough. We need to carry the alias for each function call. Therefore we need those aliases per column, right?

dimas-b: `as` can be implemented differently, of course. I personally like the style of `builder.select().max("col1").as{"max1")` vs. `builder.select().max("col1", "max1")` but it's debatable, of course :)

With the former grammar we could keep a pointer to the last `Selector` object inside the `QueryBuilderImpl` and when `as(name)` is called, we'd simply redirect that call to `currentSelector.setAlias(name)` (or something like that)... This is just a rough sketch, feel free to use a different approach.

dougwettlaufer (approved):

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
tomekl007: After some more code investigation, I see that adding an alias to a `Column` is far from trivial. It requires changing the `Column` class and all `columns()` API (also potentially adding a new type to the API). Please note that the columns API accepts a list of columns as well. This is problematic in this context when we want to have aliases (especially alias using the `as()` method after the previous call). 
Because of those reasons, I think we should not implement aliases for columns in this ticket. At this point, we should not invest more time into it, especially because there is no demand for this feature.

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
tomekl007: fixed

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
tomekl007: done

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
```diff
@@ -337,13 +328,62 @@ public void column(String column, Column.ColumnType type) {
     column(column, type, Column.Kind.Regular);
   }
 
+  public void as(String alias) {
+    if (functionCalls.isEmpty()) {
+      throw new IllegalStateException(
```
mpenick, line 333: Maybe there should be a test for this condition?

```java
    BuiltQuery<?> query =
        newBuilder()
            .select()
            .column("k2", "v1")
            .as("alias1") // call without aggregate
            .build();
```

dimas-b: As for me, I do not view it as an error necessarily, but since we have `as` in the grammar, keeping the two argument method is probably not necessary and removing it will remove the possibility of confusion in this case.

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
```diff
@@ -88,7 +79,7 @@
       @SubExpr(
           name = "select",
           definedAs =
-              "select star? column* writeTimeColumn? from (where* perPartitionLimit? limit? orderBy*) allowFiltering?"),
+              "select column* star? ((count|min|max|avg|sum|writeTimeColumn) as?)* from (where* perPartitionLimit? limit? orderBy*) allowFiltering?"),
```
dimas-b, line 82: Do we have to move `star` after `column*`?

dimas-b (approved):

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
tomekl007: This code will not compile: 
```
      newBuilder()
            .select()
            .column("k2", "v1")
            .as("alias1") // call without aggregate
            .build();
```
So this it is not possible to call the `as` without adding a function call before. 
I've added this check to clearly state  that this situation should not occur.

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
tomekl007: good catch -  changed.

dimas-b (approved):

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
```diff
@@ -56,6 +46,7 @@
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
+import jdk.nashorn.internal.ir.FunctionCall;
```
dougwettlaufer, line 49: Not needed?

PR #916, opened 2021-05-03
gRPC: Add remaining primitive codecs

mpenick:
This adds the rest of the primitive value types w/ tests. Another PR will add collections.

### DIFF
--- a/grpc-proto/proto/query.proto
+++ b/grpc-proto/proto/query.proto
@@ -57,22 +57,10 @@ message UdtValue {
   map<string, Value> fields = 1;
 }
 
-message Inet {
-  bytes address = 1;
-}
-
 message Uuid {
   // TODO: Is this going to have endian problems representing it this way?
   fixed64 msb = 1;
   fixed64 lsb = 2;
-}
-
-message Date {
-  uint32 value = 1;
-}
-
-message Time {
-  uint64 value = 1;
 }
 
 message Value {
@@ -86,14 +74,13 @@ message Value {
     double double = 5; // double
     bool boolean = 6; // boolean
     string string = 7; // ascii, varchar, text
-    bytes bytes = 8; // blob, custom
-    Inet inet = 9; // inet
-    Uuid uuid = 10; // uuid, timeuuid
-    Date date = 11; // date
-    Time time = 12; // time
-    Collection collection = 13; // list, set, tuple
-    Map map = 14;
-    UdtValue udt = 15;
+    bytes bytes = 8; // blob, inet, custom
+    Uuid uuid = 9; // uuid, timeuuid
+    uint32 date = 10; // date
+    uint64 time = 11; // time
+    Collection collection = 12; // list, set, tuple
+    Map map = 13;
+    UdtValue udt = 14;
   }
 }
 
--- a/grpc-proto/proto/query.proto
+++ b/grpc-proto/proto/query.proto
@@ -69,17 +69,48 @@ message Value {
   oneof inner {
     Null null = 1;
     Unset unset = 2;
-    sint64 int = 3; // tinyint, smallint, int, bigint, counter, timestamp
-    float float = 4; // float
-    double double = 5; // double
-    bool boolean = 6; // boolean
-    string string = 7; // ascii, varchar, text
-    bytes bytes = 8; // blob, inet, custom
-    Uuid uuid = 9; // uuid, timeuuid
-    uint32 date = 10; // date
-    uint64 time = 11; // time
-    Collection collection = 12; // list, set, tuple
+
+    // CQL types: tinyint, smallint, int, bigint, counter, timestamp
+    sint64 int = 3;
+
+    // CQL types: float
+    float float = 4;
+
+    // CQL types: double
+    double double = 5;
+
+    // CQL types: boolean
+    bool boolean = 6;
+
+    // CQL types: ascii, varchar, text
+    string string = 7;
+
+    // CQL types: blob, inet, custom
+    bytes bytes = 8;
+
+    // CQL types: uuid, timeuuid
+    Uuid uuid = 9;
+
+    // CQL types: date
+    // An unsigned integer representing days with Unix epoch (January, 1 1970) at 2^31.
+    // Examples:
+    // 0:    -5877641-06-23
+    // 2^31: 1970-1-1
+    // 2^32: 5881580-07-11
+    uint32 date = 10;
+
+    // CQL types: time
+    // An unsigned integer representing the number of nanoseconds since midnight. Valid values are
+    // in the range 0 to 86399999999999 (inclusive).
+    uint64 time = 11;
+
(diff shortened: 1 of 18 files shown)


grpc/src/main/java/io/stargate/grpc/codec/cql/ValueCodecs.java
```diff
@@ -28,12 +28,23 @@
...
+              put(Type.Date, new DateCodec());
+              put(Type.Double, new DoubleCodec());
+              put(Type.Float, new FloatCodec());
               put(Type.Int, new IntCodec());
-              put(Type.Uuid, new UuidCodec(TypeCodecs.UUID));
+              put(Type.Inet, new InetCodec());
+              put(Type.Smallint, new SmallintCodec());
+              put(Type.Text, new StringCodec(TypeCodecs.TEXT));
+              put(Type.Time, new TimeCodec());
+              put(Type.Timestamp, new BigintCodec());
```
dougwettlaufer, line 43: Timestamp uses `BigintCodec`?

dougwettlaufer: Are we saying a timestamp is only numeric and we don't support ISO-8601 format?

mpenick: We could, but my goal was to support the simplest type on the wire. The client-side would use simple conversions to convert a `java.time.Instant` (Java) or a `time.Time` (golang) to a `int64`. Those types support parsing ISO string formats.

If we support string formats for `timestamp` should we then support string representations for all other types?

grpc-proto/proto/query.proto
```diff
@@ -86,14 +74,13 @@ message Value {
...
     string string = 7; // ascii, varchar, text
-    bytes bytes = 8; // blob, custom
-    Inet inet = 9; // inet
-    Uuid uuid = 10; // uuid, timeuuid
-    Date date = 11; // date
-    Time time = 12; // time
-    Collection collection = 13; // list, set, tuple
-    Map map = 14;
-    UdtValue udt = 15;
+    bytes bytes = 8; // blob, inet, custom
```
dimas-b, line 77: Would it be meaningful to split blob, inet and custom into distinct cases even though each of them will end up being represented as `bytes`?

grpc/src/main/java/io/stargate/grpc/codec/cql/IntCodec.java
```diff
@@ -30,18 +29,18 @@ public ByteBuffer encode(@NonNull QueryOuterClass.Value value, @NonNull Column.C
     if (value.getInnerCase() != InnerCase.INT) {
       throw new IllegalArgumentException("Expected integer type");
     }
-    try {
-      return TypeCodecs.INT.encodePrimitive(
-          Math.toIntExact(value.getInt()), ProtocolVersion.DEFAULT);
-    } catch (ArithmeticException e) {
-      throw new IllegalArgumentException("Integer overflow", e);
+    int intValue = (int) value.getInt();
```
dimas-b, line 32: Why not have a dedicated `Value` field for `int` to avoid overflow?

grpc/src/main/java/io/stargate/grpc/codec/cql/TimeCodec.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class TimeCodec implements ValueCodec {
+
+  @Override
+  public ByteBuffer encode(@NonNull Value value, @NonNull ColumnType type) {
+    if (value.getInnerCase() != InnerCase.TIME) {
+      throw new IllegalArgumentException("Expected time type");
+    }
+    long time = value.getTime();
+    if (time < 0 || time > 86399999999999l) {
```
dimas-b, line 18: Could capitalizie the `l`... it looks like `1` in my browser :)

grpc-proto/proto/query.proto
```diff
@@ -86,14 +74,13 @@ message Value {
...
+    Uuid uuid = 9; // uuid, timeuuid
+    uint32 date = 10; // date
+    uint64 time = 11; // time
```
dimas-b, line 80: Is it trivial for clients to encode date and time values? Could you add a comment about the expected representation here?

grpc/src/main/java/io/stargate/grpc/codec/cql/ValueCodec.java
```diff
@@ -15,13 +15,16 @@
...
 
+import com.datastax.oss.driver.api.core.ProtocolVersion;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import io.stargate.db.schema.Column.ColumnType;
 import io.stargate.proto.QueryOuterClass.Value;
 import java.nio.ByteBuffer;
 
 /** Codec to convert to/from gRPC and CQL native protocol ({@link ByteBuffer}) values. */
 public interface ValueCodec {
+  ProtocolVersion PROTOCOL_VERSION = ProtocolVersion.DEFAULT;
```
dimas-b, line 26: Should the protocol version be the `Persistence` default rather than the driver's default? Even if we use driver codecs, I guess it might be preferable to use the latest version that `Persistence` supports. WDYT?

grpc-proto/proto/query.proto
mpenick: It's best to have the minimum number of fields as possible: https://developers.google.com/protocol-buffers/docs/overview#assigning_field_numbers

grpc/src/main/java/io/stargate/grpc/codec/cql/IntCodec.java
mpenick: gRPC uses variable encoded integers and there's no type for `short` or `byte` so I think it makes sense to a have a single integer value that can represent `tinyint`, `smallint`, `int` and `bigint`. It also helps minimize the number of fields per the last comment.

Note: We should likely be using `sint64` for this type as per: https://developers.google.com/protocol-buffers/docs/encoding#signed_integers

grpc/src/main/java/io/stargate/grpc/codec/cql/ValueCodec.java
mpenick: I agreee, but I tried doing this, but the java-driver's `ProtocolVersion` is an interface and was not straightforward to map from Stargate's protocol version. I'll give it another try.

grpc/src/main/java/io/stargate/grpc/codec/cql/IntCodec.java
mpenick: Ha, I was already using `sint64` :)

grpc-proto/proto/query.proto
dimas-b: TIL :+1:

dimas-b (approved):

dougwettlaufer (approved):

Issue #828, opened 2021-04-07
Document API query improvements


dimas-b:
This is a high-level issue to track some improvements in Docs API queries:

* Improving paging state handling (#825, #855)
* Supporting conjunction (AND) of predicates (#748)
* Supporting disjunction (OR) of predicates
* Supporting negation (NOT) of predicates

PR #919, opened 2021-05-05
Support custom paging positions within a partition

dimas-b:
This functionality is intended to help with #828

### DIFF
--- a/persistence-api/src/main/java/io/stargate/db/PagingPosition.java
+++ b/persistence-api/src/main/java/io/stargate/db/PagingPosition.java
@@ -20,9 +20,12 @@ import io.stargate.db.ImmutablePagingPosition.Builder;
 import io.stargate.db.Result.ResultMetadata;
 import io.stargate.db.datastore.Row;
 import io.stargate.db.schema.Column;
+import io.stargate.db.schema.QualifiedSchemaEntity;
 import java.nio.ByteBuffer;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Set;
 import java.util.stream.Collectors;
 import org.immutables.value.Value;
 
@@ -57,6 +60,44 @@ public interface PagingPosition {
   default Map<String, ByteBuffer> currentRowValuesByColumnName() {
     return currentRow().entrySet().stream()
         .collect(Collectors.toMap(e -> e.getKey().name(), Entry::getValue));
+  }
+
+  @Value.Lazy
+  default TableName table() {
+    Set<TableName> tables =
+        currentRow().keySet().stream()
+            .map(
+                c ->
+                    ImmutableTableName.builder()
+                        .keyspace(Objects.requireNonNull(c.keyspace()))
+                        .name(Objects.requireNonNull(c.table()))
+                        .build())
+            .collect(Collectors.toSet());
+
+    if (tables.isEmpty()) {
+      throw new IllegalArgumentException("Missing table information in custom paging request.");
+    }
+
+    if (tables.size() > 1) {
+      throw new IllegalArgumentException("Too many tables are referenced: " + tables);
+    }
+
+    return tables.iterator().next();
+  }
+
+  default ByteBuffer requiredValue(String columnName) {
+    TableName table = table(); // indirectly validates that there's only one table referenced
+
+    ByteBuffer value = currentRowValuesByColumnName().get(columnName);
+
+    if (value == null) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Required value is not present in current row (table: %s, column: %s)",
+              table.name(), columnName));
+    }
+
+    return value;
   }
 
   /** Defines how to resume paging relative to the {@link #currentRow() reference row}. */
@@ -100,4 +141,7 @@ public interface PagingPosition {
     /** Paging is resumed from the row following the {@link #currentRow() reference row}. */
     NEXT_ROW,
   }
+
+  @Value.Immutable
+  interface TableName extends QualifiedSchemaEntity {}
 }
--- a/persistence-api/src/main/java/io/stargate/db/PagingPosition.java
+++ b/persistence-api/src/main/java/io/stargate/db/PagingPosition.java
@@ -63,14 +63,14 @@ public interface PagingPosition {
   }
 
   @Value.Lazy
-  default TableName table() {
+  default TableName tableName() {
     Set<TableName> tables =
         currentRow().keySet().stream()
             .map(
-                c ->
+                column ->
                     ImmutableTableName.builder()
-                        .keyspace(Objects.requireNonNull(c.keyspace()))
-                        .name(Objects.requireNonNull(c.table()))
+                        .keyspace(Objects.requireNonNull(column.keyspace()))
+                        .name(Objects.requireNonNull(column.table()))
                         .build())
             .collect(Collectors.toSet());
 
@@ -86,7 +86,7 @@ public interface PagingPosition {
   }
 
   default ByteBuffer requiredValue(String columnName) {
-    TableName table = table(); // indirectly validates that there's only one table referenced
+    TableName table = tableName(); // indirectly validates that there's only one table referenced
 
     ByteBuffer value = currentRowValuesByColumnName().get(columnName);
 
--- a/persistence-api/src/main/java/io/stargate/db/PagingPosition.java
+++ b/persistence-api/src/main/java/io/stargate/db/PagingPosition.java
@@ -96,5 +96,8 @@ public interface PagingPosition {
      * #currentRow() reference row}.
      */
     NEXT_PARTITION,
+
+    /** Paging is resumed from the row following the {@link #currentRow() reference row}. */
+    NEXT_ROW,
   }
 }
--- a/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
+++ b/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
@@ -37,9 +37,7 @@ import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
 import org.apache.cassandra.config.CFMetaData;
 import org.apache.cassandra.config.Schema;
 import org.apache.cassandra.cql3.ColumnSpecification;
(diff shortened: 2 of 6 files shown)


persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java
```diff
@@ -162,6 +164,20 @@
     TYPE_MAPPINGS = ImmutableMap.copyOf(types);
   }
 
+  private static ByteBuffer getKeyValue(
```
ivansenic, line 167: is there no chance that all these duplicates are extracted to some to a common persistence.. Is there no chance that at least some of these depending classes are shared between all persistence modules?

ivansenic (commented):
Looks good, but too much C* specific for me to give an approval..

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java
dimas-b: Yes, I was thinking about moving it to `PagingPosition`, but the trick is that all the related table and column classes are slightly different in each persistence implementation and cannot be shared. I suppose we could use `String` arguments instead. I'll try and see what it looks like.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
```diff
@@ -187,33 +202,38 @@ public static ByteBuffer toPagingState(PagingPosition pos, Parameters parameters
...
+        break;
+
+      case NEXT_ROW:
+        ByteBuffer[] ccValues =
+            cfm.clusteringColumns().stream()
+                .map(c -> getKeyValue(pos, cfm, c))
+                .toArray(ByteBuffer[]::new);
+
+        Clustering rowClustering = Clustering.make(ccValues);
+        BTreeRow row = BTreeRow.noCellLiveRow(rowClustering, LivenessInfo.create(0, 0));
```
tomekl007, line 225: what's `LivenessInfo.create(0, 0)`? It seems that it is on the hot path, maybe extract to constant?

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
```diff
@@ -166,6 +168,19 @@
     TYPE_MAPPINGS = ImmutableMap.copyOf(types);
   }
 
+  private static ByteBuffer getKeyValue(PagingPosition pos, CFMetaData cfm, ColumnSpecification c) {
+    ByteBuffer value = pos.currentRowValuesByColumnName().get(c.name.toCQLString());
+
+    if (value == null) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Key value is not present in current row (table: %s, column: %s)",
```
tomekl007, line 177: why `Key value `? What the `Key` stands for? I don't see it used anywhere in this context.

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
```diff
@@ -1729,6 +1733,86 @@ public void testPagingStateForNextPartition() throws ExecutionException, Interru
...
+        dataStore
+            .queryBuilder()
+            .select()
+            .column("pk")
+            .column("val")
+            .from(keyspace, table)
+            .build()
+            .bind();
+
+    java.util.List<Row> allRows = dataStore.execute(selectAll).get().rows();
```
tomekl007, line 1751: nit `java.util.List` import

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
```diff
@@ -1729,6 +1733,86 @@ public void testPagingStateForNextPartition() throws ExecutionException, Interru
...
+            .build()
+            .bind();
+
+    ResultSet rs = dataStore.execute(select, p -> p.toBuilder().pageSize(5).build()).get();
+    java.util.List<Row> rows = rs.currentPageRows();
+    assertThat(rows).hasSize(5); // full page
+    Row row2 = rows.get(1);
+    assertThat(row2.getInt(0)).isEqualTo(key2);
+    assertThat(row2.getInt(1)).isEqualTo(val2);
+    // with 3 row per partition, the last row in page 1 is the second row of the second partition
```
tomekl007, line 1782: I think that this test will be more readable if it will be split into N test cases. (or parameterized test)

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
dimas-b: It's a required parameter for `Row` but it's not used for constructing the `PagingState`. Will extract.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
dimas-b: Key as in Primary _Key_. I can rename for clarity.

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
dimas-b: good catch. thx! will fix.

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
dimas-b: I'm not sure it can be split. It's really just one case with two reads. The first read gets some data and the second read uses a custom paging position to re-fetch some of the row from the first read, essentially back-tracking the pagination a bit. Information is shared between these operations from the test case perspective.

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
dimas-b: fixed

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
dimas-b: renamed to "required value" and moved to `PagingPosition`

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
dimas-b: extracted to a `private` constant.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java
dimas-b: Moved common code to `PagingPosition`... Does it look better this way? :)

persistence-api/src/main/java/io/stargate/db/PagingPosition.java
```diff
@@ -59,6 +62,44 @@
         .collect(Collectors.toMap(e -> e.getKey().name(), Entry::getValue));
   }
 
+  @Value.Lazy
+  default TableName table() {
+    Set<TableName> tables =
+        currentRow().keySet().stream()
+            .map(
+                c ->
```
mpenick, line 70: minor nit: Maybe `column` instead of `c`. It makes it easier to see this is iterating through key columns of without having to look at the types returned by `currentRow()`.

mpenick: Thinking about this more. It's not a big deal at all. Ignore. :)

persistence-api/src/main/java/io/stargate/db/PagingPosition.java
```diff
@@ -59,6 +62,44 @@
...
+      throw new IllegalArgumentException("Too many tables are referenced: " + tables);
+    }
+
+    return tables.iterator().next();
+  }
+
+  default ByteBuffer requiredValue(String columnName) {
+    TableName table = table(); // indirectly validates that there's only one table referenced
+
+    ByteBuffer value = currentRowValuesByColumnName().get(columnName);
```
mpenick, line 91: Interesting implementation detail. Lazy initialized values in the generated immutable wrapper are implemented  in a thread-safe way (`lazyInitBitmap` is volatile). There's probably a primary key size threshold where a linear search without caching the map is much faster. Maybe document store is bigger than this threshold with its big-ish primary key?
```java
  public Map<String, ByteBuffer> currentRowValuesByColumnName() {
    if ((lazyInitBitmap & CURRENT_ROW_VALUES_BY_COLUMN_NAME_LAZY_INIT_BIT) == 0) {
      synchronized (this) {
        if ((lazyInitBitmap & CURRENT_ROW_VALUES_BY_COLUMN_NAME_LAZY_INIT_BIT) == 0) {
          this.currentRowValuesByColumnName = Objects.requireNonNull(PagingPosition.super.currentRowValuesByColumnName(), "currentRowValuesByColumnName");
          lazyInitBitmap |= CURRENT_ROW_VALUES_BY_COLUMN_NAME_LAZY_INIT_BIT;
        }
      }
    }
    return currentRowValuesByColumnName;
  }
```

mpenick (approved):
+1

persistence-api/src/main/java/io/stargate/db/PagingPosition.java
dimas-b: no worries. I'll change :)

persistence-api/src/main/java/io/stargate/db/PagingPosition.java
dimas-b: I agree, there are ways to improve this code, but I did not want to do any "preliminary optimization" :) At this point Docs API has other areas where the efficiency can be improved with more gain, I believe.

Also, for there should be at most one custom pagination requests per Docs API HTTP call, so hash lookup inefficiencies may not even manifest :shrug:

persistence-api/src/main/java/io/stargate/db/PagingPosition.java
dimas-b: In 9c9c6af2 I renamed the variable, as well as the method (to `tableName()` just in case we need to add a method return the related `Table` object).

persistence-api/src/main/java/io/stargate/db/PagingPosition.java
tomekl007: Is it possible to micro-benchmark this (i..e using JMH)?

dimas-b: Certainly, but I'm just not sure it's worth the time at this point. I assume we'll test it out as part of our Docs API tests and if it shows up as a bottleneck we'll optimize. Would that be ok?

dimas-b: At this point Docs API is the only "client" of `PagingPosition`... other REST endpoints may need it too, but only when and if we support more advanced queries.

mpenick: I'm fine with merging it as is.

mpenick (approved):

dougwettlaufer (approved):

PR #920, opened 2021-05-05
generic constructor for ServiceAndProperties

ivansenic:
I was actually able to compile code that supplies to the `ServiceAndProperties` an object that is not instance of a class passed.. Seems that this is allowed, not sure when and how it fails later on, but I think it would be fine to already prevent this in the compile time?

### DIFF
--- a/core/src/main/java/io/stargate/core/activator/BaseActivator.java
+++ b/core/src/main/java/io/stargate/core/activator/BaseActivator.java
@@ -280,15 +280,15 @@ public abstract class BaseActivator implements BundleActivator {
 
     private final Hashtable<String, String> properties;
 
-    public ServiceAndProperties(
-        Object service, Class<?> targetServiceClass, Hashtable<String, String> properties) {
+    public <T, E extends T> ServiceAndProperties(
+        E service, Class<T> targetServiceClass, Hashtable<String, String> properties) {
       this.service = service;
       this.targetServiceClass = targetServiceClass;
       this.properties = properties;
     }
 
     @SuppressWarnings("JdkObsolete")
-    public ServiceAndProperties(Object service, Class<?> targetServiceClass) {
+    public <T, E extends T> ServiceAndProperties(E service, Class<T> targetServiceClass) {
       this(service, targetServiceClass, new Hashtable<>());
     }
   }
--- a/core/src/main/java/io/stargate/core/activator/BaseActivator.java
+++ b/core/src/main/java/io/stargate/core/activator/BaseActivator.java
@@ -280,15 +280,15 @@ public abstract class BaseActivator implements BundleActivator {
 
     private final Hashtable<String, String> properties;
 
-    public <T, E extends T> ServiceAndProperties(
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -280,15 +280,15 @@ protected void stopService() {
 
     private final Hashtable<String, String> properties;
 
-    public ServiceAndProperties(
-        Object service, Class<?> targetServiceClass, Hashtable<String, String> properties) {
+    public <T, E extends T> ServiceAndProperties(
```
ivansenic, line 283: do I actually need `E`?

mpenick: Do you mean `T service, Class<T> targetServiceClass` instead of `E service, Class<T>`?

ivansenic: Yup..

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -280,15 +280,15 @@ protected void stopService() {
...
+        E service, Class<T> targetServiceClass, Hashtable<String, String> properties) {
       this.service = service;
       this.targetServiceClass = targetServiceClass;
       this.properties = properties;
     }
 
     @SuppressWarnings("JdkObsolete")
-    public ServiceAndProperties(Object service, Class<?> targetServiceClass) {
+    public <T, E extends T> ServiceAndProperties(E service, Class<T> targetServiceClass) {
```
dimas-b, line 291: Do we have to re-declared generics on the constructor (`<T, E extends T>`)?

core/src/main/java/io/stargate/core/activator/BaseActivator.java
dimas-b: Having a distinction here may be convenient in cases where the same instance is used for multiple services under different interfaces.

ivansenic: Since we are not keeping the generics it does not matter much, but I made it clear with single generic param:

`T service, Class<? super T> targetServiceClass`

This is actually the correct form imo.

core/src/main/java/io/stargate/core/activator/BaseActivator.java
ivansenic: See above

mpenick (approved):

core/src/main/java/io/stargate/core/activator/BaseActivator.java
dimas-b: sgtm :+1:

core/src/main/java/io/stargate/core/activator/BaseActivator.java
dimas-b: You're right. My mistake. I though the first type param was on the class declaration :facepalm:

dimas-b (approved):

dougwettlaufer (approved):

PR #921, opened 2021-05-05
introduced error code for better handling of the responses from…

ivansenic:
Here is a quick draft for introducing a well-defined error code(s) that can return correct responses back to the user, filled with correct status, entity and message.

The benefits of this approach:
* single exception thrown with different error codes, easy to test from service perspective (unit tests catch and check error code)
* error code connected to status, rather than exception type to status
   *  more flexibility
   * easier to catch wrong status codes we have now (`DocumentDB.java` was throwing 400 instead of 404 on not existing keyspace imo)
* single place where error response is created, easy to adapt in future (for example adding extra response header, changing entity, etc)
* easy way to adapt default messages in future
* no need to repeat default messages all around the code
* easier to integrate with metrics/tracing and "record" error code as well

Goal of the finished PR:
* remove `DocumentAPIRequestException` completely
* adapt tests to fully check the expected error code


restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -169,15 +170,18 @@ public QueryBuilder builder() {
...
+    }
 
     if (!tableName.matches("^[a-zA-Z0-9_]+$")) {
-      throw new DocumentAPIRequestException(
+      String message =
           String.format(
               "Could not create collection %s, it has invalid characters. Valid characters are alphanumeric and underscores.",
-              tableName));
+              tableName);
+      throw new ErrorCodeRuntimeException(ErrorCode.DATASTORE_TABLE_NAME_INVALID, message);
```
dimas-b, line 184: Can we avoid duplicating the error message in the exception (line 182) and in the error code enum (`DATASTORE_TABLE_NAME_INVALID`)?

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCodeRuntimeException.java
```diff
@@ -0,0 +1,71 @@
...
+    return errorCode;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+    ErrorCodeRuntimeException that = (ErrorCodeRuntimeException) o;
+    return errorCode == that.errorCode;
+  }
```
dimas-b, line 65: What is the purpose of this particular way of doing `equals`? Exceptions can differ by origin/stack trace as well as by error code :thinking:

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCodeRuntimeException.java
```diff
@@ -0,0 +1,71 @@
...
+
+/** {@link RuntimeException} that carries an {@link ErrorCode}. */
+public class ErrorCodeRuntimeException extends RuntimeException {
+
+  /** Dedicated {@link ErrorCode}. */
+  private final ErrorCode errorCode;
+
+  /** ErrorCode only constructor. */
+  public ErrorCodeRuntimeException(ErrorCode errorCode) {
+    this.errorCode = errorCode;
```
dimas-b, line 30: Should we perhaps use `super(errorCode.defaultMessage())` in this case? Otherwise the exception will have a `null` message, AFAIK, which does not show up nicely in logs.

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCodeRuntimeException.java
```diff
@@ -0,0 +1,71 @@
...
+
+  /** @see Exception#Exception(String, Throwable) */
+  public ErrorCodeRuntimeException(ErrorCode errorCode, String message, Throwable cause) {
+    super(message, cause);
+    this.errorCode = errorCode;
+  }
+
+  /** Pipe to {@link ErrorCode#toResponse()} with the exception message. */
+  public Response getResponse() {
+    return errorCode.toResponse(getLocalizedMessage());
```
dimas-b, line 47: I'm not sure about `getLocalizedMessage` in this context. It give the impression of producing a localized message for the end user, but will use the server's default locale, which is probably not user-specific.

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
```diff
@@ -0,0 +1,75 @@
...
+  private final String defaultMessage;
+
+  ErrorCode(Response.Status responseStatus, String defaultMessage) {
+    this.responseStatus = responseStatus;
+    this.defaultMessage = defaultMessage;
+  }
+
+  /** @return Returns {@link Response} using the ErrorCode default message. */
+  public Response toResponse() {
+    return toResponse(null);
```
dimas-b, line 53: should we `return toResponse(defaultMessage)` to match javadoc?

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
```diff
@@ -0,0 +1,75 @@
...
+  /** Generic data store errors. */
+  DATASTORE_KEYSPACE_DOES_NOT_EXIST(Response.Status.NOT_FOUND, "An unknown namespace provided."),
+  DATASTORE_TABLE_DOES_NOT_EXIST(Response.Status.NOT_FOUND, "An unknown table provided."),
+  DATASTORE_TABLE_NAME_INVALID(
+      Response.Status.BAD_REQUEST,
+      "The table name contains invalid characters. Valid characters are alphanumeric and underscores."),
+
+  /** Document API. */
+  DOCS_API_GENERAL_PAGE_SIZE_EXCEEDED(
+      Response.Status.BAD_REQUEST, "The parameter `page-size` is limited to 20.");
```
dimas-b, line 35: Could we have a constant for `20` so that the code using the error can be sure to use the same value in its checks?

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
ivansenic: Well, if we want to have a non-static messages that can depend on the user params for example, then I don't see a way to do it.. I think it's quite valuable sometimes that you specify the more details in the error message than just a static one? 

Any idea on your mind?

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCodeRuntimeException.java
ivansenic: Yea, this will be removed.. Thanks..

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCodeRuntimeException.java
ivansenic: Perfect, I can then further simplify in the `ErrorCode` I guess..

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCodeRuntimeException.java
ivansenic: Fine, will adapt..

dimas-b (commented):
LGTM overall :+1: Minor comments added.

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
ivansenic: Yep, this is also a mistake, wrote it was quick :clock11:

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
ivansenic: I don't think so, this has to be generic for all error codes?

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCodeRuntimeException.java
ivansenic: Done.

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
ivansenic: yep that's doable.. thanks for clarification :+1:.. Adding..

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
ivansenic: You missed the difference I guess, in the above example it depends directly on the collection name (table) provided by the user.

ivansenic:
Finalized, rebased and ready for final review.

dimas-b (approved):

ivansenic:
Blocked by #937

PR #922, opened 2021-05-05
Break up searchRows into more reasonable parts

EricBorczuk:
Before this change, searchRows was doing far too much - so I broke it up into 3 reasonable parts.

1) getUnfilteredRows - selects the next page from cassandra
2) getRowsForDocument - selects a page of rows for a particular document
3) getCandidateRows - selects rows for the "candidacy filtering" algorithm we use for evaluating search filters with AND

The common parts of these 3 functions were also refactored (most of them have to do with converting data into C* predicates)

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
@@ -1,6 +1,5 @@
 package io.stargate.web.docsapi.dao;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import io.stargate.db.PagingPosition;
 import io.stargate.db.PagingPosition.ResumeMode;
 import io.stargate.db.datastore.DataStore;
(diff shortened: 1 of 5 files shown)


eribeiro:
Hi @EricBorczuk, I would like to suggest some changes that could be part of this PR.

1) Rename the `addRowsToMap` method to `groupRowsByKey` or `groupByKey`;

2) The following lines in `updateExistenceForMap` can be changed to ``groupByKey(documentChunks, rows)`:

```
    for (int i = 0; i < rows.size(); i++) {
      Row row = rows.get(i);
      String key = row.getString("key");
      List<Row> chunk = documentChunks.getOrDefault(key, new ArrayList<>());
      chunk.add(row);
      documentChunks.put(key, chunk);
    }
```

dimas-b:
I'm actually working on a parallel refactoring under #828 that touches a lot of the code refactored here... it might be worth syncing up on this offline @EricBorczuk :)

dimas-b:
`searchRows` was removed in #937... which should be ready for review/merge soon.

EricBorczuk:
#937 removed searchRows, closing this

PR #37, opened 2020-09-12
Handle missing token on requests

dougwettlaufer:

PR #925, opened 2021-05-06
Standardize use of Paginator limits

EricBorczuk:
The "search within a document" functionality now properly honors the page-size default of 1. That default could change in the future, but is now shared with the "search across collections" functionality.

Note that In order to do this, the db page size actually had to be set to be equal to the doc page size when performing search within a document.  This is so that no rows are discarded in memory after querying and the pagination state is maintained properly.

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
@@ -1,6 +1,5 @@
 package io.stargate.web.docsapi.dao;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import io.stargate.db.PagingPosition;
 import io.stargate.db.PagingPosition.ResumeMode;
 import io.stargate.db.datastore.DataStore;
@@ -54,7 +53,7 @@ public class Paginator {
     this.currentDbPageState = resultSet.getPagingState();
   }
 
-  public String makeExternalPagingState() throws JsonProcessingException {
+  public String makeExternalPagingState() {
     ByteBuffer pagingState;
 
     if (lastDocumentId != null && resultSet != null) {
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
@@ -6,7 +6,6 @@ import io.stargate.db.datastore.DataStore;
 import io.stargate.db.datastore.ResultSet;
 import io.stargate.db.schema.Column;
 import io.stargate.db.schema.Column.Type;
-import io.stargate.web.docsapi.exception.DocumentAPIRequestException;
 import java.nio.ByteBuffer;
 import java.util.Base64;
 
@@ -31,10 +30,6 @@ public class Paginator {
     docPageSize = pageSize;
     if (docPageSize < 1) {
       throw new IllegalStateException("Document page size cannot be less than 1.");
-    }
-
-    if (docPageSize > DocumentDB.MAX_PAGE_SIZE) {
-      throw new DocumentAPIRequestException("The parameter `page-size` is limited to 20.");
     }
 
     this.dbPageSize = dbPageSize;
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
@@ -29,12 +29,19 @@ public class Paginator {
 
   public Paginator(DataStore dataStore, String pageState, int pageSize, int dbPageSize) {
     this.dataStore = dataStore;
-    docPageSize = Math.max(1, pageSize);
+    docPageSize = pageSize;
+    if (docPageSize < 1) {
+      throw new IllegalStateException("Document page size cannot be less than 1.");
+    }
+
     if (docPageSize > DocumentDB.MAX_PAGE_SIZE) {
       throw new DocumentAPIRequestException("The parameter `page-size` is limited to 20.");
     }
 
     this.dbPageSize = dbPageSize;
+    if (this.dbPageSize < 1) {
+      throw new IllegalStateException("Database page size cannot be less than 1.");
+    }
 
     if (pageState != null) {
       byte[] decodedBytes = Base64.getDecoder().decode(pageState);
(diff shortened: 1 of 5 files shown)


restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -660,12 +660,13 @@ public Response getDocPath(
             logger.debug(json);
             return Response.ok(json).build();
           } else {
+            int pageSize = Math.max(1, pageSizeParam);
```
dimas-b, line 663: What about ` defaultValue = "100"` for the `page-size` HTTP parameter on line 587? Why is it not effective? Is it just for swagger documention? In the latter case I think we ought to fix it too :)

EricBorczuk: Oh yes - my fault - that is just for swagger, will change

dimas-b (approved):

grpc/pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>io.stargate</groupId>
     <artifactId>stargate</artifactId>
-    <version>1.0.23</version>
+    <version>1.0.24-SNAPSHOT</version>
```
dimas-b, line 7: This should help, but #926 takes priority for updating this version on `master`, I guess

EricBorczuk: Yea I will wait until after #926 to merge

dimas-b: #926 has been merged

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -1217,11 +1213,6 @@ private JsonNode searchWithOnlyInMemoryFilters(
               .collect(Collectors.toList());
     }
 
-    if (!inMemoryFilters.isEmpty() && paginator.hasDbPageState()) {
-      throw new DocumentAPIRequestException(
-          "The results as requested must fit in one page, try increasing the `page-size` parameter.");
```
dimas-b, line 1222: Is this no longer an issue? It is taking me some extra efforts to implement on my branch :smiley: So, I'll wait for this to merge and rebase my refactoring.

dimas-b (approved):

dimas-b:
@EricBorczuk : if test are causing trouble, we can probably just pull the default page size fixes into my branch and I will open a PR from it. I'm almost ready and I have all Docs inntegratio tests passing except `testPaginationDisallowedLimitedSupport` but you removed it anyway. Let's discuss tomorrow.

EricBorczuk:
Sounds good - yea, I just realized that searching within a document shouldn't have the same max limit as searching across collections (especially because of how limited support filters work when searching within a document) - the limit of 20 isn't enough in some cases, and I only made a limit for the collection search because documents could be arbitrarily large.

But yeah, we will discuss tomorrow :) I suspect this test run will succeed though

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: So - I realized along the way that this can be left out - the page size for searching within a document just works in a strange way for limited support filters as it is...but I think it is fine to remove this behavior, it is too restrictive as it is

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
```diff
@@ -29,12 +27,15 @@
 
   public Paginator(DataStore dataStore, String pageState, int pageSize, int dbPageSize) {
     this.dataStore = dataStore;
-    docPageSize = Math.max(1, pageSize);
-    if (docPageSize > DocumentDB.MAX_PAGE_SIZE) {
```
tomekl007, line 33: should we keep the upper `pageSize` validation as well (besides the lower bound)?

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -1375,7 +1375,7 @@ public void testBasicSearch() throws IOException {
             hostWithPort
```
tomekl007, line 1375: can we add two test cases for edge conditions:
1. Validate that it returns an error when page-size > upper bound
2. Validate that it returns an error when page-size < lower bound

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -762,18 +759,20 @@ public Response searchDoc(
 
           DocumentDB db = dbFactory.getDocDataStoreForToken(authToken, getAllHeaders(request));
 
+          int docPageSize = Math.max(1, pageSizeParam);
```
dougwettlaufer, line 762: Looks like Paginator already does this? 
https://github.com/stargate/stargate/blob/master/restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java#L32

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -762,18 +759,20 @@ public Response searchDoc(
...
                   dbFactory.getDataStore(),
                   pageStateParam,
-                  pageSizeParam,
+                  docPageSize,
                   docsApiConfiguration.getSearchPageSize());
 
-          JsonNode results;
-
-          if (pageSizeParam > 20) {
+          if (docPageSize > DocumentDB.MAX_PAGE_SIZE) {
```
dougwettlaufer, line 770: Also duplicated by Paginator constructor?

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -660,12 +660,9 @@ public Response getDocPath(
...
             return Response.ok(json).build();
           } else {
+            int pageSize = Math.max(1, pageSizeParam);
             final Paginator paginator =
-                new Paginator(
-                    dbFactory.getDataStore(),
-                    pageStateParam,
-                    pageSizeParam,
-                    pageSizeParam > 0 ? pageSizeParam : docsApiConfiguration.getSearchPageSize());
+                new Paginator(dbFactory.getDataStore(), pageStateParam, pageSize, pageSize);
```
dougwettlaufer, line 665: Why does this one get `pageSize` twice? Just curious.

Would it make sense to have another constructor that takes 3 args?

ivansenic:
> #937 implements the same page size defaults. As discussed, @EricBorczuk , would you be ok with closing this PR in favour of #37 and perhaps rebased it after #937 merges and re-apply whatever is not covered by #937? WDYT?

@EricBorczuk Also, I just started on adding validation layer to the resources, maybe you can completely skip this and let me handle it there..

PR #933, opened 2021-05-07
Add checks for document table validity before querying

EricBorczuk:
Instead of throwing a schema-related exception and returning a 500 when someone tries to use the docs API to access a non-docs table, give them a better error message and a 400

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
@@ -37,7 +37,7 @@ public enum ErrorCode {
   /** Document API. */
   DOCS_API_TABLE_NOT_A_COLLECTION(
       Response.Status.BAD_REQUEST,
-      "The requested Cassandra table is not a Documents API Collection."),
+      "The requested database table is not a Documents API Collection."),
   DOCS_API_GENERAL_ARRAY_LENGTH_EXCEEDED(
       Response.Status.BAD_REQUEST,
       String.format(
--- a/restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
@@ -35,6 +35,9 @@ public enum ErrorCode {
       "The table name contains invalid characters. Valid characters are alphanumeric and underscores."),
 
   /** Document API. */
+  DOCS_API_TABLE_NOT_A_COLLECTION(
+      Response.Status.BAD_REQUEST,
+      "The requested Cassandra table is not a Documents API Collection."),
   DOCS_API_GENERAL_ARRAY_LENGTH_EXCEEDED(
       Response.Status.BAD_REQUEST,
       String.format(
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
@@ -655,7 +655,6 @@ public class DocumentResourceV2 {
                     pageStateParam,
                     pageSizeParam,
                     pageSizeParam > 0 ? pageSizeParam : docsApiConfiguration.getSearchPageSize());
-            db = dbFactory.getDocDataStoreForToken(authToken, getAllHeaders(request));
             JsonNode result =
                 documentService.searchDocumentsV2(
                     db, namespace, collection, filters, selectionList, id, paginator);
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
@@ -6,8 +6,6 @@ import com.datastax.oss.driver.api.core.NoNodeAvailableException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.stargate.auth.UnauthorizedException;
-import io.stargate.db.schema.Keyspace;
-import io.stargate.db.schema.Table;
 import io.stargate.web.docsapi.dao.DocumentDB;
 import io.stargate.web.docsapi.dao.Paginator;
 import io.stargate.web.docsapi.examples.WriteDocResponse;
@@ -15,10 +13,10 @@ import io.stargate.web.docsapi.exception.ErrorCode;
 import io.stargate.web.docsapi.exception.ErrorCodeRuntimeException;
 import io.stargate.web.docsapi.models.DocumentResponseWrapper;
 import io.stargate.web.docsapi.service.DocsApiConfiguration;
+import io.stargate.web.docsapi.service.DocsSchemaChecker;
 import io.stargate.web.docsapi.service.DocumentService;
 import io.stargate.web.docsapi.service.filter.FilterCondition;
 import io.stargate.web.models.Error;
-import io.stargate.web.resources.AuthenticatedDB;
 import io.stargate.web.resources.Db;
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
@@ -65,6 +63,7 @@ public class DocumentResourceV2 {
   @Inject private ObjectMapper mapper;
   @Inject private DocumentService documentService;
   @Inject private DocsApiConfiguration docsApiConfiguration;
+  @Inject private DocsSchemaChecker schemaChecker;
 
   @POST
   @ManagedAsync
@@ -630,23 +629,11 @@ public class DocumentResourceV2 {
           }
 
           // check first that namespace and table exist
-          AuthenticatedDB authenticatedDB = dbFactory.getDataStoreForToken(authToken, allHeaders);
-          Keyspace keyspace = authenticatedDB.getKeyspace(namespace);
-          if (null == keyspace) {
-            String message = String.format("Namespace %s does not exist.", namespace);
-            throw new ErrorCodeRuntimeException(
-                ErrorCode.DATASTORE_KEYSPACE_DOES_NOT_EXIST, message);
-          }
-          Table table = keyspace.table(collection);
-          if (null == table) {
-            String message = String.format("Collection %s does not exist.", collection);
-            throw new ErrorCodeRuntimeException(ErrorCode.DATASTORE_TABLE_DOES_NOT_EXIST, message);
-          }
+          DocumentDB db = dbFactory.getDocDataStoreForToken(authToken, allHeaders);
+          schemaChecker.checkValidity(namespace, collection, db);
 
           JsonNode node;
           if (filters.isEmpty()) {
-
-            DocumentDB db = dbFactory.getDocDataStoreForToken(authToken, allHeaders);
             node = documentService.getJsonAtPath(db, namespace, collection, id, path);
             if (node == null) {
               return Response.noContent().build();
@@ -668,7 +655,7 @@ public class DocumentResourceV2 {
                     pageStateParam,
                     pageSizeParam,
                     pageSizeParam > 0 ? pageSizeParam : docsApiConfiguration.getSearchPageSize());
-            DocumentDB db = dbFactory.getDocDataStoreForToken(authToken, getAllHeaders(request));
+            db = dbFactory.getDocDataStoreForToken(authToken, getAllHeaders(request));
             JsonNode result =
                 documentService.searchDocumentsV2(
                     db, namespace, collection, filters, selectionList, id, paginator);
@@ -763,6 +750,7 @@ public class DocumentResourceV2 {
           }
 
           DocumentDB db = dbFactory.getDocDataStoreForToken(authToken, getAllHeaders(request));
+          schemaChecker.checkValidity(namespace, collection, db);
 
           final Paginator paginator =
               new Paginator(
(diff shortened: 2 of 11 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -413,6 +423,8 @@ public void putAtPath(
       db.maybeCreateTableIndexes(keyspace, collection);
     }
 
+    checkTableIsDocsCollection(db, keyspace, collection);
```
eribeiro, line 426: Sincere question: wouldn't this be checked in `DocumentResourceV2` as we already do with `keyspace` and `table`?

ivansenic: Seems reasonable. Maybe all checks could be grouped into a single `Checker`. Maybe we can also cache the results there, I mean check should be done before doing any operation, but with each check we add extra latency. So if check is "green" for keyspace/table combo we could cache it for like 5-10mins? This way we would not recheck for tables that we know are fine.. And if something changes in the meantime, let's fail on the query then.. It's a tradeoff..

@EricBorczuk @eribeiro What do you both think?

dimas-b: +1 to caching schema compatibility checks.

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -668,7 +655,7 @@ public Response getDocPath(
                     pageStateParam,
                     pageSizeParam,
                     pageSizeParam > 0 ? pageSizeParam : docsApiConfiguration.getSearchPageSize());
-            DocumentDB db = dbFactory.getDocDataStoreForToken(authToken, getAllHeaders(request));
+            db = dbFactory.getDocDataStoreForToken(authToken, getAllHeaders(request));
```
dimas-b, line 658: Why do we need to call `getDocDataStoreForToken` again? It seems that we already have a `DocumentDB` on line 632 :thinking:

restapi/src/main/java/io/stargate/web/docsapi/service/DocsSchemaChecker.java
```diff
@@ -0,0 +1,66 @@
...
+import io.stargate.web.docsapi.exception.ErrorCode;
+import io.stargate.web.docsapi.exception.ErrorCodeRuntimeException;
+import java.util.Objects;
+
+/**
+ * A helper class for checking that particular keyspace/table combinations exist and are valid
+ * document API collections.
+ */
+public class DocsSchemaChecker {
+  public static class KeyspaceAndTable {
```
dimas-b, line 15: suggestion: use `@Value.Immutable` to auto-generate getters and `.equals`. It's likely going to be as efficient as a custom class, but takes less code on our side.

restapi/src/main/java/io/stargate/web/docsapi/service/DocsSchemaChecker.java
```diff
@@ -0,0 +1,66 @@
...
+    db.tableExists(keyspace, table);
+  }
+
+  public boolean isDocumentsTable(String keyspace, String table, DocumentDB db) {
+    KeyspaceAndTable keyspaceAndTable = new KeyspaceAndTable(keyspace, table);
+    Boolean valid = validatedDocCollections.getIfPresent(keyspaceAndTable);
+    if (valid == null) {
+      valid = db.isDocumentsTable(keyspace, table);
+      validatedDocCollections.put(keyspaceAndTable, valid);
+    }
```
dimas-b, line 51: suggestion: `Map.computeIfAbsent(...)`

restapi/src/main/java/io/stargate/web/docsapi/service/DocsSchemaChecker.java
```diff
@@ -0,0 +1,66 @@
...
+    return valid;
+  }
+
+  public void checkValidity(String keyspace, String table, DocumentDB db) {
+    checkTableExists(keyspace, table, db);
+    boolean isDocsTable = isDocumentsTable(keyspace, table, db);
```
dimas-b, line 57: Is there a reason for calling `checkTableExists` outside of `isDocumentsTable`? 

If we check existence separately, would it make sense to also report it as different kind of error?

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
EricBorczuk: oh we probably don't have to anymore...good point

DocumentDB used to get newed up with page size and page state, but now the Paginator handles that

restapi/src/main/java/io/stargate/web/docsapi/service/DocsSchemaChecker.java
EricBorczuk: checkTableExists throws an exception that marshalls to 404, but isDocumentsTable will produce a 400 (that's why I kept them separate, though maybe it should be clearer that `checkTableExists` is a void method that throws)

EricBorczuk: I still have to do the cache invalidation on schema changes part btw - looking into that now

dimas-b: Ah, I missed the 404 part, sorry :facepalm:

restapi/src/main/java/io/stargate/web/docsapi/service/DocsSchemaChecker.java
```diff
@@ -0,0 +1,66 @@
...
+    KeyspaceAndTable keyspaceAndTable = new KeyspaceAndTable(keyspace, table);
+    Boolean valid = validatedDocCollections.getIfPresent(keyspaceAndTable);
+    if (valid == null) {
+      valid = db.isDocumentsTable(keyspace, table);
+      validatedDocCollections.put(keyspaceAndTable, valid);
+    }
+    return valid;
+  }
+
+  public void checkValidity(String keyspace, String table, DocumentDB db) {
```
ivansenic, line 55: are these public methods above needed at all? Or better said, can they be private? I would expect that everybody from outside just hit the `checkValidity`.

restapi/src/main/java/io/stargate/web/docsapi/service/DocsSchemaChecker.java
ivansenic: ahh no let's not hit the `checkTableExists` every time, that was the reason why we wanted to have the cache at first place.. can we cache an enum as the value : `VALID`, `KEYSPACE_NOT_EXISTING`, `TABLE_NOT_EXISTING`, `TABLE_NOT_COLLECTION` and throw exception based on that??

restapi/src/main/java/io/stargate/web/docsapi/service/DocsSchemaChecker.java
```diff
@@ -0,0 +1,66 @@
...
+import io.stargate.web.docsapi.dao.DocumentDB;
+import io.stargate.web.docsapi.exception.ErrorCode;
+import io.stargate.web.docsapi.exception.ErrorCodeRuntimeException;
+import java.util.Objects;
+
+/**
+ * A helper class for checking that particular keyspace/table combinations exist and are valid
+ * document API collections.
+ */
+public class DocsSchemaChecker {
```
ivansenic, line 14: I think you missed to impl the listening to the schema/table changes to invalidate the cache, or?

ivansenic (commented):
Please check my comments, I guess they are important.

restapi/src/main/java/io/stargate/web/docsapi/service/DocsSchemaChecker.java
EricBorczuk: Just finished the cache invalidation part!

restapi/src/main/java/io/stargate/web/docsapi/service/DocsSchemaChecker.java
EricBorczuk: I removed these in the most recent checks - also I made the assumption in the caching that if the schema is a valid collection, then the table must have existed - so I decided to cache them both together (if the table exists and all is well, it caches until the schema changes again)

restapi/src/main/java/io/stargate/web/docsapi/service/DocsSchemaChecker.java
```diff
@@ -0,0 +1,53 @@
...
+    boolean isDocsTable = db.isDocumentsTable(keyspace, table);
+    if (!isDocsTable) {
+      throw new ErrorCodeRuntimeException(
+          ErrorCode.DOCS_API_TABLE_NOT_A_COLLECTION,
+          String.format(
+              "The Cassandra table %s.%s is not a Documents collection. Accessing arbitrary tables via the Documents API is not permitted.",
+              keyspace, table));
+    }
+    validatedDocCollections.put(keyspaceAndTable, true);
+    this.lastCheckedSchema = db.schema();
```
dimas-b, line 51: Would it be more appropriate to do this assignment in `clearCacheOnSchemaChange()`?

restapi/src/main/java/io/stargate/web/docsapi/service/DocsSchemaChecker.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * A helper class for checking that particular keyspace/table combinations exist and are valid
+ * document API collections.
+ */
+public class DocsSchemaChecker {
+  private Schema lastCheckedSchema;
+
+  private final Cache<ImmutableKeyspaceAndTable, Boolean> validatedDocCollections =
+      Caffeine.newBuilder().maximumSize(10_000).build();
```
dimas-b, line 19: nit: For this simple use case a `Caffeine` cache might be an overkill. A simple `ConcurrentHashMap` could do. WDYT?

EricBorczuk: Oh yeah that's true - since there are no time-based invalidation rules or anything

dimas-b (approved):
Let's run our usual perf. tests on this, just in case :)

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
```diff
@@ -35,6 +35,9 @@
       "The table name contains invalid characters. Valid characters are alphanumeric and underscores."),
 
   /** Document API. */
+  DOCS_API_TABLE_NOT_A_COLLECTION(
+      Response.Status.BAD_REQUEST,
+      "The requested Cassandra table is not a Documents API Collection."),
```
dougwettlaufer, line 40: Minor but this should be more generic
```suggestion
      "The requested database table is not a Documents API Collection."),
```

restapi/src/main/java/io/stargate/web/docsapi/service/DocsSchemaChecker.java
```diff
@@ -0,0 +1,48 @@
...
+  private boolean performValidityCheck(ImmutableKeyspaceAndTable keyspaceAndTable, DocumentDB db) {
+    String keyspace = keyspaceAndTable.getKeyspace();
+    String table = keyspaceAndTable.getTable();
+    db.tableExists(keyspace, table);
+    boolean isDocsTable = db.isDocumentsTable(keyspace, table);
+    if (!isDocsTable) {
+      throw new ErrorCodeRuntimeException(
+          ErrorCode.DOCS_API_TABLE_NOT_A_COLLECTION,
+          String.format(
+              "The Cassandra table %s.%s is not a Documents collection. Accessing arbitrary tables via the Documents API is not permitted.",
```
dougwettlaufer, line 43: Minor but this should be more generic

```suggestion
              "The database table %s.%s is not a Documents collection. Accessing arbitrary tables via the Documents API is not permitted.",
```

restapi/src/main/java/io/stargate/web/docsapi/service/DocsSchemaChecker.java
```diff
@@ -0,0 +1,48 @@
...
+ * document API collections.
+ */
+public class DocsSchemaChecker {
+  private Schema lastCheckedSchema;
+
+  private final ConcurrentHashMap<ImmutableKeyspaceAndTable, Boolean> validatedDocCollections =
+      new ConcurrentHashMap<>();
+
+  private void clearCacheOnSchemaChange(DocumentDB db) {
+    if (!db.schema().equals(lastCheckedSchema)) {
```
dougwettlaufer, line 21: If we're comparing the entire schema on every call is it worth having a cache?

EricBorczuk: According to Dmitri, Schema pre-hashes its equals / hashCode and so this comparison should be quick

dimas-b: The idea is to short-circuit on stable schema. I'm not actually sure if it's going to have a measurable impact. So I'd be ok with removing the schema check. However, without it we'd have to refresh by time, which is prone to false positives and we really wanted to avoid false positive 500 errors.

We discussed using a schema change listener for invalidation, but it seems to be about as effective as a schema-global check in serverless env. We you have any suggestions for cache invalidation, they'd be very welcome, considering the proverbial complexity of the problem here :)

EricBorczuk: After running the same perf tests using YCSB, I didn't find any major change in latency, so I'm going to merge this one as-is; if we think of a better way to do this caching in the future we can always change it :) no one-way doors here

Issue #935, opened 2021-05-07
Add support for collections to gRPC


mpenick:
Add support for list, set, map, and tuple. 

Note: This will not add support for UDTs that will be handled in another issue.

mpenick:
Resolved by https://github.com/stargate/stargate/pull/936

PR #936, opened 2021-05-07
gRPC: Collections

mpenick:
Support list, set, map, tuple

### DIFF
--- a/grpc/src/main/java/io/stargate/grpc/codec/cql/CollectionCodec.java
+++ b/grpc/src/main/java/io/stargate/grpc/codec/cql/CollectionCodec.java
@@ -1,0 +1,59 @@
+package io.stargate.grpc.codec.cql;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import io.stargate.db.schema.Column;
+import io.stargate.db.schema.Column.ColumnType;
+import io.stargate.proto.QueryOuterClass;
+import io.stargate.proto.QueryOuterClass.Collection;
+import io.stargate.proto.QueryOuterClass.Value;
+import java.nio.ByteBuffer;
+
+public class CollectionCodec extends CompositeCodec {
+
+  @Override
+  public ByteBuffer encode(@NonNull QueryOuterClass.Value value, @NonNull Column.ColumnType type) {
+    if (!value.hasCollection()) {
+      throw new IllegalArgumentException("Expected collection type");
+    }
+    assert type.isCollection();
+
+    Collection collection = value.getCollection();
+
+    Column.ColumnType elementType = type.parameters().get(0);
+    ValueCodec elementCodec = ValueCodecs.get(elementType.rawType());
+
+    int elementCount = collection.getElementsCount();
+    ByteBuffer[] encodedElements = new ByteBuffer[elementCount];
+    int toAllocate = 4;
+    for (int i = 0; i < elementCount; ++i) {
+      ByteBuffer encodedElement =
+          ValueCodec.encodeValue(elementCodec, collection.getElements(i), elementType);
+      if (encodedElement == null) {
+        throw new IllegalArgumentException(
+            String.format("null is not supported inside %ss", type.rawType().cqlDefinition()));
+      }
+      encodedElements[i] = encodedElement;
+      toAllocate += 4 + encodedElement.remaining();
+    }
+
+    ByteBuffer result = ByteBuffer.allocate(toAllocate);
+    result.putInt(elementCount);
+    return encodeValues(encodedElements, result);
+  }
+
+  @Override
+  public Value decode(@NonNull ByteBuffer bytes, @NonNull ColumnType type) {
+    Collection.Builder builder = Collection.newBuilder();
+
+    Column.ColumnType elementType = type.parameters().get(0);
+    ValueCodec elementCodec = ValueCodecs.get(elementType.rawType());
+
+    ByteBuffer input = bytes.duplicate();
+    int elementCount = input.getInt();
+    for (int i = 0; i < elementCount; i++) {
+      builder.addElements(decodeValue(input, elementCodec, elementType));
+    }
+
+    return Value.newBuilder().setCollection(builder).build();
+  }
+}
--- a/grpc/src/main/java/io/stargate/grpc/codec/cql/CollectionCodec.java
+++ b/grpc/src/main/java/io/stargate/grpc/codec/cql/CollectionCodec.java
@@ -3,7 +3,6 @@ package io.stargate.grpc.codec.cql;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import io.stargate.db.schema.Column;
 import io.stargate.db.schema.Column.ColumnType;
-import io.stargate.proto.QueryOuterClass;
 import io.stargate.proto.QueryOuterClass.Collection;
 import io.stargate.proto.QueryOuterClass.Value;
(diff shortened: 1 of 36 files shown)


mpenick:
Resolves https://github.com/stargate/stargate/issues/935

grpc/src/main/java/io/stargate/grpc/codec/cql/MapCodec.java
```diff
@@ -0,0 +1,76 @@
...
+    ColumnType keyType = type.parameters().get(0);
+    ValueCodec keyCodec = ValueCodecs.get(keyType.rawType());
+
+    ColumnType valueType = type.parameters().get(1);
+    ValueCodec valueCodec = ValueCodecs.get(valueType.rawType());
+
+    int elementCount = map.getElementsCount();
+    if (elementCount % 2 != 0) {
+      throw new IllegalArgumentException(
+          "Missing pair value (expected an even number of elements)");
```
dimas-b, line 29: nit: how do we know it's the pair _value_ that is missing and not the key? :)

grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -153,12 +152,12 @@ private void handleError(Throwable throwable, StreamObserver<Result> responseObs
...
   private CompletableFuture<Prepared> prepareQuery(
-      Connection connection, String cql, StringValue keyspace, boolean tracing) {
+      Connection connection, String cql, String keyspace, boolean tracing) {
     CompletableFuture<Prepared> future = new CompletableFuture<>();
     final StringBuilder keyBuilder = new StringBuilder();
     connection.loggedUser().ifPresent(user -> keyBuilder.append(user.name()));
-    if (keyspace.isInitialized()) {
-      keyBuilder.append(keyspace.getValue());
+    if (keyspace != null) {
+      keyBuilder.append(keyspace);
```
dimas-b, line 160: It may be worth using explicit separators between key components to avoid confusing (`key`, `statement`) with (`null`, `keystatement`). It's not likely to manifest, I guess, but might be worth implementing for good measure. WDYT?

dimas-b: suggestion: a dedicated `@Value.Immutable` inner class might be as efficient a key as a constructed `String` and will solve the component separation problem.

grpc/src/main/java/io/stargate/grpc/codec/cql/MapCodec.java
mpenick: Am I missing something? Keys are always odd.  `<key> (1), <value> (2), <key> (3), <value> (4) ...`.

grpc/src/main/java/io/stargate/grpc/codec/cql/TupleCodec.java
```diff
@@ -0,0 +1,67 @@
...
+    assert type.isTuple();
+
+    Collection tuple = value.getCollection();
+
+    int fieldCount = tuple.getElementsCount();
+    if (fieldCount > type.parameters().size()) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Too many tuple fields. Expected %d, but received %d",
+              type.parameters().size(), fieldCount));
```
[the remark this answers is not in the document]
mpenick: Good question! Yes. This is a case I have in the tests too.

grpc/src/main/java/io/stargate/grpc/service/Service.java
mpenick: Good suggestion. Let me give that a try.

grpc/src/main/java/io/stargate/grpc/codec/cql/MapCodec.java
dimas-b: Sorry, I did not mean to cause confusion. What I mean is that since the array length is odd, something is mixed up, but can we be certain it's the "value" that is missing? The client might have submitted something completely irrelevant here, so it might be best to return `expected an even number of elements` without trying to interpret the contents. WDYT?

grpc/src/main/java/io/stargate/grpc/codec/cql/CollectionCodec.java
```diff
@@ -0,0 +1,58 @@
...
+import java.nio.ByteBuffer;
+
+public class CollectionCodec extends CompositeCodec {
+
+  @Override
+  public ByteBuffer encode(@NonNull Value value, @NonNull Column.ColumnType type) {
+    if (!value.hasCollection()) {
+      throw new IllegalArgumentException("Expected collection type");
+    }
+    assert type.isCollection();
```
dougwettlaufer, line 17: Why an assert rather than an actual check? A few lines down will throw an NPE if it's not a collection, right?

```java
Column.ColumnType elementType = type.parameters().get(0);
```

grpc/src/main/java/io/stargate/grpc/codec/cql/CollectionCodec.java
```diff
@@ -0,0 +1,58 @@
...
+
+    Collection collection = value.getCollection();
+
+    Column.ColumnType elementType = type.parameters().get(0);
+    ValueCodec elementCodec = ValueCodecs.get(elementType.rawType());
+
+    int elementCount = collection.getElementsCount();
+    ByteBuffer[] encodedElements = new ByteBuffer[elementCount];
+    int toAllocate = 4;
```
dougwettlaufer, line 26: What's this?

grpc/src/main/java/io/stargate/grpc/codec/cql/CollectionCodec.java
dimas-b: +1 to converting `assert` to an always-on runtime check. It should not have any significant overhead in this case, right?

grpc/src/main/java/io/stargate/grpc/codec/cql/CollectionCodec.java
mpenick: Space for the number of elements in the collection.

grpc/src/main/java/io/stargate/grpc/codec/cql/CollectionCodec.java
mpenick: This truly is an assert. The codec is determined by the type. Something's really gone wrong if that assert fails and it's really not possible.

grpc/src/main/java/io/stargate/grpc/codec/cql/CollectionCodec.java
mpenick: https://github.com/stargate/stargate/pull/936/files/bff0e34c80a4cdcfcd34c2b07e6a6fd11b8c7533#diff-8379a2c1845030448fcd49119bf131681bb724d112e78d0974cc91c837abc622R39

grpc/src/main/java/io/stargate/grpc/codec/cql/MapCodec.java
mpenick: Fixed this.

grpc/src/main/java/io/stargate/grpc/service/Service.java
mpenick: Fixed this.

dougwettlaufer (approved):

PR #940, opened 2021-05-10
Add async writes

tomekl007:


graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcher.java
```diff
@@ -65,7 +66,15 @@ protected BulkMutationFetcher(
       buildException = e;
     }
     OperationDefinition operation = environment.getOperationDefinition();
-    if (operation.getDirectives().stream().anyMatch(d -> d.getName().equals(ATOMIC_DIRECTIVE))
+
+    // if the caller expects the atomic mutations,
+    // he should not use fire-and-forget behavior provided by the async directive
```
dimas-b, line 71: Is this not what logged batches provide?

tomekl007: yes, this is why we allow logged batch (`atomic` directive) or the new `async`.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/DmlFetcher.java
```diff
@@ -149,6 +158,33 @@ protected Parameters getDatastoreParameters(DataFetchingEnvironment environment)
...
+                log.warn(
+                    String.format(
+                        "The query %s executed within the %s directive failed.",
+                        query, ASYNC_DIRECTIVE),
+                    throwable);
+              }
+            });
+    // complete immediately with applied=true without waiting for the result
+    return CompletableFuture.completedFuture(
+        toAppliedMutationResultWithOriginalValue(originalValue));
```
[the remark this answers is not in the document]
dimas-b: Well, since `@async` annotates specific mutations, we should be able to use a different return type. Put it another way, making a mutation async is a schema change from the client's perspective, so it should not be a surprise the user that the return value is different. WDYT?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcher.java
dimas-b: Yes, but why not allow both? If the client does not want to wait for the acknowledgement of a batch write, it's the client's choice. Why would we want to explicitly forbid it?

dougwettlaufer: +1 an async batch makes sense to me.

testing/src/main/graphql/betterbotz/async/BulkInsertProductsWithAsync.graphql
```diff
@@ -0,0 +1,16 @@
+mutation BulkInsertProductsWithAsync(
```
dougwettlaufer, line 1: These tests should probably make use of @olim7t change in https://github.com/stargate/stargate/pull/906

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/DmlFetcher.java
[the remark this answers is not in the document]
tomekl007: As far as I can see, we don't have the knowledge of the semantic of the GraphQL query at the `GraphqlResourceBase` where we decide about the HTTP response codes. Also, all responses all returned with the standard `200` code.
We may be able to refactor the whole `GraphQL` worklow to give the `GraphqlResourceBase`:
```
   .thenApply(result -> asyncResponse.resume(result.toSpecification()))
```
feedback about whether it should return `202` or `200` but I am not sure if this is worth the effort (@olim7t wdyt?)
I think that a good middle ground is returning `accepted` in the GraphQL response body as @dimas-b suggested.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcher.java
tomekl007: Here is my reasoning around forbidding it:

On the one hand, we may allow both operations - it seems that batch without waiting for the `ack`. On the other hand, I think that we should enforce good patterns and prohibit the usage of harmful patterns. 

I assume that when the client annotated its operations with `@atomic`, he wants a form of atomic "transactions". In that case, I think that the caller SHOULD wait for the ack. `Atomic` without `ack` does not give any value to the caller. It is tough when there are N operations in one `Atomic` update. How will the caller validate if this batch was indeed executed correctly? He would need to perform N GET queries for all mutations in the `atomic` batch and wait for the result. Sure, it is feasible but far from ideal.
When prohibiting `async` and `atomic` we are enforcing good usage patterns of our clients. 

Please note that `atomic` =! `bulk` inserts. The current implementation allows bulk inserts (N records) with `@async` directive. Wdyt?The accepted field is relevant and fulfilled with data only when used with the @%s directive.
wdyt?

ivansenic: I think there is bunch of use cases where I don't care about the result of the atomic operation.. So `@atomic` & `@async` makes sense for me from the developer point of view.. It's his choice at the end..

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/DmlFetcher.java
```diff
@@ -149,6 +163,33 @@ protected Parameters getDatastoreParameters(DataFetchingEnvironment environment)
     return ImmutableMap.of("value", finalValue, "applied", applied);
   }
 
+  protected boolean containsDirective(OperationDefinition operation, String atomicDirective) {
```
ivansenic, line 166: `atomicDirective` -> `directive` naming?

ivansenic (approved):
Seems good to me, let's just resolve this atomic & async discussion in proper way.. Rest is fine imo..

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/DmlFetcher.java
```diff
@@ -149,6 +163,33 @@ protected Parameters getDatastoreParameters(DataFetchingEnvironment environment)
...
+            (r, throwable) -> {
+              if (throwable != null) {
+                log.warn(
+                    String.format(
+                        "The query %s executed within the %s directive failed.",
+                        query, ASYNC_DIRECTIVE),
+                    throwable);
+              }
+            });
+    // complete immediately with applied=true without waiting for the result
```
mpenick, line 188: Should this comment say `accepted=true`?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/DmlSchemaBuilder.java
```diff
@@ -163,6 +161,26 @@ public GraphQLSchema build() {
...
+            .build());
+  }
+
+  private void addAsyncDirective(GraphQLSchema.Builder builder) {
+    builder.additionalDirective(
+        GraphQLDirective.newDirective()
+            .validLocation(Introspection.DirectiveLocation.MUTATION)
+            .name(ASYNC_DIRECTIVE)
+            .description(
+                "Instructs the server to apply the mutations in an Async way. "
```
mpenick, line 179: minor: `... apply the mutations in an Async way.` --> `... apply the mutations asynchronously without waiting for the result` ?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcher.java
tomekl007: It seems that I am out-voted here :D 
Implemented `async` AND `atomic`:
https://github.com/stargate/stargate/pull/940/commits/49746aca90892df8674f9409bb8632e8bc077d52

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/DmlFetcher.java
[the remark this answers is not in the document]
dimas-b: Yeah, I think the GraphQL infrastructure takes over HTTP code generation, so it would probably be an overkill for us to try to override the HTTP code in this PR.

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/BulkMutationFetcher.java
mpenick: > I believe C* provides atomic batches in the sense that all of none of the mutations in the batch take effect eventually

mpenick (approved):

dimas-b (approved):

testing/src/main/graphql/betterbotz/async/BulkInsertProductsWithAsync.graphql
tomekl007: @olim7t, do you want me to wait until your PR is merged, or can I merge and you will adapt yours?

dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/DmlFetcher.java
```diff
@@ -149,6 +172,33 @@ protected Parameters getDatastoreParameters(DataFetchingEnvironment environment)
     return ImmutableMap.of("value", finalValue, "applied", applied);
   }
 
+  protected boolean containsDirective(OperationDefinition operation, String directive) {
+    return operation.getDirectives().stream().anyMatch(d -> d.getName().equals(directive));
+  }
+
+  /**
+   * Executes the query in an async way in a fire and forget fashion. Completes immediately with
+   * applied=true without waiting for the result.
```
olim7t, line 181:
```suggestion
   * accepted=true without waiting for the result.
```

olim7t (approved):
👍 
I'm not convinced that this will be used very often, but the implementation looks good.

testing/src/main/graphql/betterbotz/async/BulkInsertProductsWithAsync.graphql
olim7t: Whoops, I just merged it 😬  So the former...

tomekl007: Ok, I've adapted to the new IT testing approach.

PR #941, opened 2021-05-10
Added client info metric tag provider

ivansenic:
PR provides a way to add additional tags from the `ClientInfo` into the `cql_` metrics. Details and open questions:

* same approach as for the HTTP tags, added one new interface in the `persistance-api`, provided default impl that s registered if no other impl is set via the props
* I refactored the `ClientMetrics`, so that we can use micrometer and the new interface together
   * my intention here is that we only use the micrometer and go away from the metric registry
   * I changed some signatures to relieve class clients from depending on metric impl behind (see bytes metrics)
* everything was quite easy except for the connected clients, Doug said `connectedNativeClients` should have tags as well, so I used the multi gauge to achieve this
   * it would be awesome if anybody can advise any other option than the timer?
* the rest of the connected clients metrics is quite confusing to me, in two cases we are returning the `List<Map<String, String>>` as the gauge value, how can this work?
   * can we dismiss any of those metrics? I am not sure how can we handle this with micrometer
   * do we actually need now `connectedNativeClientsByUser`, as we allow anybody to report tags based on the `ClientInfo` and thus distinguish users

I tried to keep everything to be backwards compatible wrt to naming (there is even a typo I kept -> `RequestDiscarded` vs `RequestsProcessed`). However in case of distribution summaries I am not exposing the `quantiles` as before, if this is needed I can add exactly the ones we had before.. Also the `Meter` concept in DW metrics does not exist in the Micrometer as it's not needed, so I moved there to classic counters for requests processed and discarded.

### DIFF
--- a/cql/src/main/java/io/stargate/cql/impl/CqlImpl.java
+++ b/cql/src/main/java/io/stargate/cql/impl/CqlImpl.java
@@ -22,6 +22,7 @@ import io.netty.channel.nio.NioEventLoopGroup;
 import io.stargate.auth.AuthenticationService;
 import io.stargate.core.metrics.api.Metrics;
 import io.stargate.db.Persistence;
+import io.stargate.db.metrics.api.ClientInfoMetricsTagProvider;
 import java.net.InetAddress;
 import java.util.Arrays;
 import java.util.Collection;
@@ -43,17 +44,20 @@ public class CqlImpl {
   private final Persistence persistence;
   private final Metrics metrics;
   private final AuthenticationService authentication;
+  private final ClientInfoMetricsTagProvider clientInfoTagProvider;
 
   public CqlImpl(
       Config config,
       Persistence persistence,
       Metrics metrics,
-      AuthenticationService authentication) {
+      AuthenticationService authentication,
+      ClientInfoMetricsTagProvider clientInfoTagProvider) {
     TransportDescriptor.daemonInitialization(config);
 
     this.persistence = persistence;
     this.metrics = metrics;
     this.authentication = authentication;
+    this.clientInfoTagProvider = clientInfoTagProvider;
 
     if (useEpoll()) {
       workerGroup = new EpollEventLoopGroup();
@@ -91,7 +95,8 @@ public class CqlImpl {
       }
     }
 
-    ClientMetrics.instance.init(servers, metrics.getRegistry("cql"));
+    ClientMetrics.instance.init(
+        servers, metrics.getRegistry("cql"), metrics.getMeterRegistry(), clientInfoTagProvider);
     servers.forEach(Server::start);
     persistence.setRpcReady(true);
   }
--- a/cql/src/main/java/io/stargate/cql/impl/CqlImpl.java
+++ b/cql/src/main/java/io/stargate/cql/impl/CqlImpl.java
@@ -95,10 +95,10 @@ public class CqlImpl {
       }
     }
 
-    int metricUpdateRate =
-        Integer.parseInt(System.getProperty("stargate.cql.metrics.updateRate", "10000"));
+    double metricsUpdatePeriodSeconds =
+        Double.parseDouble(System.getProperty("stargate.cql.metrics.updatePeriodSeconds", "0.1"));
     ClientMetrics.instance.init(
-        servers, metrics.getMeterRegistry(), clientInfoTagProvider, metricUpdateRate);
+        servers, metrics.getMeterRegistry(), clientInfoTagProvider, metricsUpdatePeriodSeconds);
     servers.forEach(Server::start);
     persistence.setRpcReady(true);
   }
--- a/cql/src/main/java/io/stargate/cql/impl/CqlImpl.java
+++ b/cql/src/main/java/io/stargate/cql/impl/CqlImpl.java
@@ -95,8 +95,10 @@ public class CqlImpl {
       }
     }
 
+    int metricUpdateRate =
+        Integer.parseInt(System.getProperty("stargate.cql.metrics.updateRate", "10000"));
     ClientMetrics.instance.init(
-        servers, metrics.getRegistry("cql"), metrics.getMeterRegistry(), clientInfoTagProvider);
+        servers, metrics.getMeterRegistry(), clientInfoTagProvider, metricUpdateRate);
     servers.forEach(Server::start);
     persistence.setRpcReady(true);
   }
@@ -104,6 +106,7 @@ public class CqlImpl {
   public void stop() {
     persistence.setRpcReady(false);
     servers.forEach(Server::stop);
+    ClientMetrics.instance.shutdown();
   }
 
   public static boolean useEpoll() {
--- a/cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
+++ b/cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
@@ -32,7 +32,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -132,14 +131,14 @@ public final class ClientMetrics {
    * @param servers List of servers to get metrics for
    * @param meterRegistry Micrometer MeterRegistry to report to
    * @param clientInfoTagProvider Tag provider for the connection
-   * @param updateRateMillis Number of milliseconds to schedule the internal metric update task. If
-   *     zero or less task will not be scheduled.
+   * @param updatePeriodSeconds The period for internal metric update in 1/s (f.e. for 10 sec rate,
+   *     use 0.1). If zero or less task will not be scheduled.
    */
   public synchronized void init(
       Collection<Server> servers,
       MeterRegistry meterRegistry,
       ClientInfoMetricsTagProvider clientInfoTagProvider,
-      int updateRateMillis) {
+      double updatePeriodSeconds) {
 
     if (initialized) return;
 
@@ -170,8 +169,9 @@ public final class ClientMetrics {
 
     initialized = true;
 
-    // if we have the update rate, init the executor service and submit the update task
-    if (updateRateMillis > 0) {
+    // if we have the positive period, init the executor service and submit the update task
+    if (updatePeriodSeconds > 0) {
+      long fixedRateMillis = Double.valueOf(1000d / updatePeriodSeconds).longValue();
       executorService =
           Executors.newSingleThreadScheduledExecutor(
               new BasicThreadFactory.Builder()
@@ -189,7 +189,7 @@ public final class ClientMetrics {
             }
           },
           0,
-          updateRateMillis,
+          fixedRateMillis,
           TimeUnit.MILLISECONDS);
     }
   }
@@ -205,6 +205,7 @@ public final class ClientMetrics {
           executorService.shutdownNow();
         }
       } catch (InterruptedException e) {
+        logger.warn("Interrupted during executor termination, going for shutdownNow().", e);
         executorService.shutdownNow();
       }
     }
@@ -227,16 +228,15 @@ public final class ClientMetrics {
   }
 
   void updateConnectedClients() {
-    Map<ConnectionMetrics, Integer> total = new HashMap<>();
+    Map<Tags, Integer> total = new HashMap<>();
 
     for (Server server : servers) {
       server
-          .countConnectedClientsByConnectionMetrics()
-          .forEach(
-              (metrics, count) -> total.compute(metrics, (ci, v) -> v == null ? count : v + count));
+          .countConnectedClientsByConnectionTags()
+          .forEach((tags, count) -> total.compute(tags, (t, v) -> v == null ? count : v + count));
     }
 
-    recordMapToMultiGauge(connectedNativeClients, total, ConnectionMetrics::getTags);
+    recordMapToMultiGauge(connectedNativeClients, total, Function.identity());
   }
 
   void updateConnectedClientsByUser() {
@@ -270,7 +270,7 @@ public final class ClientMetrics {
    * @return full name
    */
   private static String metric(String name) {
-    // to keep the back-ward compatibility, init all metric names with cql_ + DefaultNameFactory
+    // to keep the back-ward compatibility, init all metric names with cql. + DefaultNameFactory
     String metricName = factory.createMetricName(name).getMetricName();
     return "cql." + metricName;
   }
@@ -325,19 +325,6 @@ public final class ClientMetrics {
     @Override
     public Tags getTags() {
       return tags;
-    }
-
-    @Override
-    public boolean equals(Object o) {
-      if (this == o) return true;
-      if (o == null || getClass() != o.getClass()) return false;
-      ConnectionMetricsImpl that = (ConnectionMetricsImpl) o;
-      return tags.equals(that.tags);
-    }
-
-    @Override
-    public int hashCode() {
-      return Objects.hash(tags);
     }
   }
 }
--- a/cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
+++ b/cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
@@ -49,6 +49,7 @@ public final class ClientMetrics {
   private static final String AUTH_SUCCESS_METRIC;
   private static final String AUTH_FAILURE_METRIC;
   private static final String AUTH_ERROR_METRIC;
+  private static final String CONNECTED_NATIVE_CLIENTS_METRIC;
 
   // init to avoid re-computing on each record
   static {
@@ -57,6 +58,7 @@ public final class ClientMetrics {
     AUTH_SUCCESS_METRIC = metric("AuthSuccess");
     AUTH_FAILURE_METRIC = metric("AuthFailure");
     AUTH_ERROR_METRIC = metric("AuthError");
+    CONNECTED_NATIVE_CLIENTS_METRIC = metric("connectedNativeClients");
   }
 
   private volatile boolean initialized = false;
@@ -67,7 +69,6 @@ public final class ClientMetrics {
   private ClientInfoMetricsTagProvider clientInfoTagProvider;
 
   private AtomicInteger pausedConnections;
-
   private Counter totalBytesRead;
   private Counter totalBytesWritten;
   private DistributionSummary bytesReceivedPerFrame;
@@ -75,31 +76,6 @@ public final class ClientMetrics {
   private MultiGauge connectedNativeClients;
 
   private ClientMetrics() {}
-
-  public void markRequestProcessed(ClientInfo clientInfo) {
-    Tags tags = clientInfoTagProvider.getClientInfoTags(clientInfo);
-    meterRegistry.counter(REQUESTS_PROCESSED_METRIC, tags).increment();
-  }
-
-  public void markRequestDiscarded(ClientInfo clientInfo) {
-    Tags tags = clientInfoTagProvider.getClientInfoTags(clientInfo);
-    meterRegistry.counter(REQUESTS_DISCARDED_METRIC, tags).increment();
-  }
-
-  public void markAuthSuccess(ClientInfo clientInfo) {
-    Tags tags = clientInfoTagProvider.getClientInfoTags(clientInfo);
-    meterRegistry.counter(AUTH_SUCCESS_METRIC, tags).increment();
-  }
-
-  public void markAuthFailure(ClientInfo clientInfo) {
-    Tags tags = clientInfoTagProvider.getClientInfoTags(clientInfo);
-    meterRegistry.counter(AUTH_FAILURE_METRIC, tags).increment();
-  }
-
-  public void markAuthError(ClientInfo clientInfo) {
-    Tags tags = clientInfoTagProvider.getClientInfoTags(clientInfo);
-    meterRegistry.counter(AUTH_ERROR_METRIC, tags).increment();
-  }
 
   public void pauseConnection() {
     pausedConnections.incrementAndGet();
@@ -123,6 +99,14 @@ public final class ClientMetrics {
 
   public void recordBytesTransmittedPerFrame(double value) {
     bytesTransmittedPerFrame.record(value);
+  }
+
+  public ConnectionMetrics connectionMetrics(ClientInfo clientInfo) {
+    if (!initialized) {
+      throw new IllegalStateException("Client metrics not initialized yet.");
+    }
+
+    return new ConnectionMetricsImpl(clientInfo);
   }
 
   public synchronized void init(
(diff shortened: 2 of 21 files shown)


cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -88,62 +109,69 @@ public void unpauseConnection() {
...
+    Timer connectedNativeClientsTimer = new Timer();
+    connectedNativeClientsTimer.schedule(
+        new TimerTask() {
+          @Override
+          public void run() {
+            updateConnectedClients();
+          }
+        },
+        10000L,
+        10000L);
```
olim7t, line 161: > can we avoid using the timer, but get notified from somebody on the changes?

Yeah a timer has a few drawbacks: holds a thread, updates only happen at a fixed interval...

And `Server` does not provide a way to monitor when client connections get added/removed. And even if we added that, I'm not sure if that's the best approach because we only really need to update the values when the metrics get scraped.

Maybe we could extend `PrometheusMeterRegistry` to add a `beforeScrape` callback?

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -19,65 +19,86 @@
...
+  private DistributionSummary bytesTransmittedPerFrame;
+  private MultiGauge connectedNativeClients;
 
   private ClientMetrics() {}
 
-  public void markRequestProcessed() {
-    requestsProcessed.mark();
+  public void markRequestProcessed(ClientInfo clientInfo) {
+    Tags tags = clientInfoTagProvider.getClientInfoTags(clientInfo);
+    meterRegistry.counter(REQUESTS_PROCESSED_METRIC, tags).increment();
```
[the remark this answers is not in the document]
ivansenic: Hmm, OK after looking deeper, this is only one get on the `ConcurrentHashMap`, I don't think we'll gain that much in total.. But maybe it makes sense if we mix this up with the counting of native connections..

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -162,12 +179,32 @@ private long nettyHeapMemory() {
     return -1;
   }
 
-  private int countConnectedClients() {
-    int count = 0;
-
-    for (Server server : servers) count += server.countConnectedClients();
```
mpenick, line 168: `Server#countConnectedClients()` is dead code now. Should it be removed?

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -162,12 +179,32 @@ private long nettyHeapMemory() {
...
+                  Tags clientInfoTags = clientInfoTagProvider.getClientInfoTags(entry.getKey());
+                  return MultiGauge.Row.of(clientInfoTags, entry.getValue());
+                })
+            .collect(Collectors.toList());
+
+    // ensure overwrite is called to re-write existing values to new ones
+    connectedNativeClients.register(rows, true);
   }
 
   private Map<String, Integer> countConnectedClientsByUser() {
```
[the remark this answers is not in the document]
ivansenic: Also an option.. So we could just have at the end: 

`cql_org_..._connectedNativeClients_total{tag1=, tag2=, username=,}`

or we can keep in separate metrics:

`cql_org_..._connectedNativeClients_total{tag1=, tag2=,}`
`cql_org_..._connectedNativeClientsByUser_total{username=,}`

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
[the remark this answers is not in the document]
dimas-b: Re: HTTP: I believe the CQL protocol+impl. is much more tuned for performance :)

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Connection.java
```diff
@@ -28,16 +29,24 @@
...
+      ProtocolVersion version,
+      Tracker tracker,
+      ConnectionMetrics connectionMetrics) {
     this.channel = channel;
     this.version = version;
     this.tracker = tracker;
+    this.connectionMetrics = connectionMetrics;
 
     tracker.addConnection(channel, this);
+    // TODO decrease where
```
dimas-b, line 48: `decrease where`?

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -183,39 +220,79 @@ private int countConnectedClients() {
...
-  }
+    @Override
+    public void increaseConnectedNativeClients() {
+      connectedClients.incrementAndGet();
+    }
 
-  private Meter registerMeter(String name) {
-    return metricRegistry.meter(factory.createMetricName(name).getMetricName());
+    @Override
+    public void decreaseConnectedNativeClients() {
```
dimas-b, line 294: Is this method called? I could not find the call side, sorry.

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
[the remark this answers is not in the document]
dimas-b: Re: `per user`, user names may clash in a serverless env. Is it really meaningful to count specifically by user _name_?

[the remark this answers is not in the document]
dimas-b: Why do not treat the username separately and not as a tag?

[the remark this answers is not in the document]
ivansenic: I also vote for kicking it out..

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
ivansenic: No matter if we take this proposal, **we need to fix the connection tracking removing anyway as @mpenick suggested above**.. This is such a nice way to have memory leaks..

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -162,12 +179,32 @@ private long nettyHeapMemory() {
...
+        total.entrySet().stream()
+            .map(
+                entry -> {
+                  Tags clientInfoTags = clientInfoTagProvider.getClientInfoTags(entry.getKey());
+                  return MultiGauge.Row.of(clientInfoTags, entry.getValue());
+                })
+            .collect(Collectors.toList());
+
+    // ensure overwrite is called to re-write existing values to new ones
+    connectedNativeClients.register(rows, true);
```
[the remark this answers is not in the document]
ivansenic: Resolving as not doing.

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
ivansenic: Removed.

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -162,12 +179,32 @@ private long nettyHeapMemory() {
...
+                          combined.compute(clientInfo, (ci, v) -> v == null ? count : v + count));
+                  return combined;
+                });
+
+    // then map to multi gauge rows
+    List<MultiGauge.Row<?>> rows =
+        total.entrySet().stream()
+            .map(
+                entry -> {
+                  Tags clientInfoTags = clientInfoTagProvider.getClientInfoTags(entry.getKey());
```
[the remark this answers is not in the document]
ivansenic: In the final version the connection setup increased by getting all the counters initialized and calculating the tags for the connection once. I think this is good. Furthermore, I will reuse calculated tags in all other actions, so there is no need to recalculate every again! OK?

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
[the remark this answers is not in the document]
ivansenic: Keeping as agreed, but moving to micrometer multi-gauge.

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
ivansenic: OK confirmed that there is no need to remove channels & connections, the `DefaultChannelGroup` does this automatically..

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
ivansenic: I kept it in order to be still aligned to `Server` the Cassandra 4 & 3.11

ivansenic:
@dimas-b @mpenick Pushed updates as agreed yesterday.. I changed the impl in the MultiGauge update method, and I am taking already precalculated tags exposed in the `ConnectionMetrics`.. So for every connection there is only one time we calculate tags!

I am confident to merge on green, but you can have another look if you want.

cql/src/main/java/io/stargate/cql/impl/CqlImpl.java
```diff
@@ -91,14 +95,18 @@ public void start() {
       }
     }
 
-    ClientMetrics.instance.init(servers, metrics.getRegistry("cql"));
+    int metricUpdateRate =
+        Integer.parseInt(System.getProperty("stargate.cql.metrics.updateRate", "10000"));
```
dimas-b, line 99: nit: rename `updateRate` to `updatePeriodSeconds` for the same of clarity? The "rate" for updating once every 10 sec would be 0.1 (unit 1/s), right?

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -88,64 +102,114 @@ public void unpauseConnection() {
...
+  public synchronized void shutdown() {
+    if (!initialized) return;
+
+    if (null != executorService && !executorService.isShutdown()) {
+      executorService.shutdown();
+      try {
+        if (!executorService.awaitTermination(1, TimeUnit.SECONDS)) {
+          executorService.shutdownNow();
+        }
+      } catch (InterruptedException e) {
```
dimas-b, line 207: nit: I'd log the `InterruptedException`... just for good measure and debuggability.

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
dimas-b: If it's dead code, lets remove from Stargate. We do not have to be 1:1 with C* code, I think :thinking:

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -180,42 +249,95 @@ private int countConnectedClients() {
...
 
-    return clients;
+  /**
+   * Resolves a metric name for the micrometer registry.
+   *
+   * @param name metric name
+   * @return full name
+   */
+  private static String metric(String name) {
+    // to keep the back-ward compatibility, init all metric names with cql_ + DefaultNameFactory
```
dimas-b, line 273: `cql_` or `cql.` (`_` vs. `.`)?

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Server.java
```diff
@@ -44,10 +44,7 @@
 import io.netty.util.internal.logging.Slf4JLoggerFactory;
 import io.stargate.auth.AuthenticationService;
 import io.stargate.cql.impl.CqlImpl;
-import io.stargate.db.AuthenticatedUser;
-import io.stargate.db.EventListener;
-import io.stargate.db.EventListenerWithChannelFilter;
-import io.stargate.db.Persistence;
+import io.stargate.db.*;
```
dimas-b, line 47: please expand imports

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Server.java
```diff
@@ -321,12 +323,25 @@ int countConnectedClients() {
...
     }
+
+    Map<ConnectionMetrics, Integer> countConnectedClientsByConnectionMetrics() {
+      Map<ConnectionMetrics, Integer> result = new HashMap<>();
+      for (Channel c : allChannels) {
+        Connection connection = c.attr(Connection.attributeKey).get();
+        if (connection instanceof ServerConnection) {
+          ServerConnection conn = (ServerConnection) connection;
+          ConnectionMetrics connectionMetrics = conn.getConnectionMetrics();
+          result.compute(connectionMetrics, (ci, v) -> v == null ? 1 : v + 1);
```
dimas-b, line 340: Sorry, I might have missed the comparison logic, but it is possible to have more that one value per key here?

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -180,42 +249,95 @@ private int countConnectedClients() {
...
 
-  private Counter registerCounter(String name) {
-    return metricRegistry.counter(factory.createMetricName(name).getMetricName());
-  }
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) return true;
+      if (o == null || getClass() != o.getClass()) return false;
+      ConnectionMetricsImpl that = (ConnectionMetricsImpl) o;
+      return tags.equals(that.tags);
```
dimas-b, line 335: Would it be more human-readable to do all aggregations explicitly by `Tags` as opposed to by `ConnectionMetrics`, which delegates `equals` to `Tags`?

dimas-b (commented):
LGTM overall. Minor comments added.

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -88,64 +102,114 @@ public void unpauseConnection() {
...
 
-    bytesReceivedPerFrame = registerHistogram("BytesReceivedPerFrame");
-    bytesTransmittedPerFrame = registerHistogram("BytesTransmittedPerFrame");
+    bytesReceivedPerFrame = meterRegistry.summary(metric("BytesReceivedPerFrame"));
+    bytesTransmittedPerFrame = meterRegistry.summary(metric("BytesTransmittedPerFrame"));
 
     initialized = true;
+
+    // if we have the update rate, init the executor service and submit the update task
+    if (updateRateMillis > 0) {
```
tomekl007, line 174: Why do we need this if? If someone sets it to `0`, shouldn't the below action just execute immediately?

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -88,64 +102,114 @@ public void unpauseConnection() {
...
+      executorService =
```
tomekl007, line 175: why don't make this `executorService` static and/or initialize it in the constructor (or `static` block)?

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Server.java
```diff
@@ -321,12 +323,25 @@ int countConnectedClients() {
...
         }
       }
       return result;
     }
+
+    Map<ConnectionMetrics, Integer> countConnectedClientsByConnectionMetrics() {
+      Map<ConnectionMetrics, Integer> result = new HashMap<>();
+      for (Channel c : allChannels) {
+        Connection connection = c.attr(Connection.attributeKey).get();
+        if (connection instanceof ServerConnection) {
```
tomekl007, line 337: Just for curiosity: what other instances of the `Connection` we may get here? Is it possible to extract the number of connections from them?

cql/src/test/java/org/apache/cassandra/stargate/metrics/ClientMetricsTest.java
```diff
@@ -0,0 +1,486 @@
...
+    clientInfo2 = mock(ClientInfo.class);
+
+    when(clientTagProvider.getClientInfoTags(clientInfo1)).thenReturn(Tags.of("client", "one"));
+    when(clientTagProvider.getClientInfoTags(clientInfo2)).thenReturn(Tags.of("client", "two"));
+
+    List<Server> servers = Arrays.asList(server1, server2);
+    clientMetrics.init(servers, meterRegistry, clientTagProvider, 0);
+  }
+
+  @Nested
```
tomekl007, line 74: why do we need those nested classes with test cases `happyPath()`? I think that the existing pattern for tests is to have it in a more flat structure, without nesting, AND having more descriptive names.

cql/src/test/java/org/apache/cassandra/stargate/metrics/ClientMetricsTest.java
```diff
@@ -0,0 +1,486 @@
...
+  class MarkRequestProcessed {
+
+    @Test
+    public void happyPath() {
```
tomekl007, line 78: can't we create more descriptive names for those tests that `happyPath`? and/or possibly use `@DisplayName` for a more human-readable test case? How the maven surefire output looks like when using N nested classes with N test suites where each is named `happyPath`?
```
 @DisplayName("Should calculate requests processed")
  public void shouldCalculateRequestsProcessed() {
```

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -180,42 +249,95 @@ private int countConnectedClients() {
...
+  }
 
-    return clients;
+  /**
+   * Resolves a metric name for the micrometer registry.
+   *
+   * @param name metric name
+   * @return full name
+   */
+  private static String metric(String name) {
```
tomekl007, line 272: I think that we are missing a test case for a backward-compatibility.

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
ivansenic: This is there so somebody can disable the update with the value zero. Since we make it configurable rate, we should also make it configurable to be turned off.

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
ivansenic: Because I don't want to initialize the executor service if it will not be used, no need to keep extra thread around doing nothing.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Server.java
ivansenic: True, I don't need this "cast", can simply check for non null. Thanks, will update.

cql/src/test/java/org/apache/cassandra/stargate/metrics/ClientMetricsTest.java
ivansenic: I am a huge fan of Nested tests as there are so many advantages over the flat structure you are proposing to follow here. Let me explain:

First, if you open my test you can clearly see from the nested class name what (public) method is being tested in the cut. Meaning you have clear separation for target tested code/path. Open any flat structure test and you'll need to dig deep to actually figure out what's being tested in each test. And descriptive names don't help here imo, this is unit test, you are testing a peace of code no behavior.

Second, since you have that nested structure per the method, you can then actually have test methods in the nested class to handle different paths / situations. This gives you a clear view on what's covered for each method.

Third, once you have such a clear structure, adding new tests is easy. If you extended a existing method, add a new test method in the target nested class. If you added new method, add new nested class.

And, the last results outputs in the IDE and reports are we structured as well. Clear and readable, you can even directly associate the possible failing test with the peace of code that might influence it:

image

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
ivansenic: Adding..

cql/src/test/java/org/apache/cassandra/stargate/metrics/ClientMetricsTest.java
tomekl007: Yeah, I agree that this IDE structure looks very clear!
However, IDE is one side of the coin. Is it also so user-friendly when running the `mvn` test runner?

cql/src/test/java/org/apache/cassandra/stargate/metrics/ClientMetricsTest.java
ivansenic: The `happyPath` name is to denote that we are testing a very simple straightforward path in the target method. It's usually there to test expected behavior. Since you know what's being tested from the nested class, it often does not make sence to now name it differently.. 

DisplayName is nice, but it's pure luxury to have and maintain it, as everything should be figured out from the test method name. If you take my tests names for example, you have smth like:

```
ClientMetricsTest.UpdateConnectedClients#happyPath
ClientMetricsTest.UpdateConnectedClients#noConnections
ClientMetricsTest.UpdateConnectedClients#removedConnections
```

* From `ClientMetricsTest` you know we are testing `ClientMetrics` class. 
* From `UpdateConnectedClients` you know we are testing `updateConnectedClients()` method
* From `#name` you know what specific situation are we testing in the method

Surefire reports looks good, more structured anyway that the flat test and quicker to realize what and where got wrong. But might I ask how often do you look into the surefire reports? When was the last time you looked?

cql/src/main/java/io/stargate/cql/impl/CqlImpl.java
ivansenic: Done.

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
ivansenic: Done.

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
ivansenic: Code is correct, comment not..

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Server.java
ivansenic: Done. Old setting.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/ServerConnection.java
```diff
@@ -49,13 +51,12 @@
...
-            proxyInfo != null
-                ? proxyInfo.sourceAddress
-                : (InetSocketAddress) channel.remoteAddress(),
-            proxyInfo != null ? proxyInfo.publicAddress : null);
+    super(
+        channel,
+        version,
+        tracker,
+        ClientMetrics.instance.connectionMetrics(getClientInfo(channel, proxyInfo)));
+    this.clientInfo = getClientInfo(channel, proxyInfo);
```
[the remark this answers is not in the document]
ivansenic: Done.

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
ivansenic: Sure, done.

cql/src/test/java/org/apache/cassandra/stargate/metrics/ClientMetricsTest.java
tomekl007: Sorry, by surefire I meant the standard `mvn` output when running test via `mvn test` command line.

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
ivansenic: Actually it's already tested? All tests in the both unit and int `ClientMetricsTest` confirm the metric names. Or you had something else in mind?

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
ivansenic: Removed

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
tomekl007: yes, the `ClientMetricsTest` tests but I think that only the subset of all metrics (only memory-related)

ivansenic: `ClientMetricsTest` unit test every single metric. `ClientMetricsTest` int tests targets few, because there is no need to check all imo, first fill fail anyway. Or?

cql/src/test/java/org/apache/cassandra/stargate/metrics/ClientMetricsTest.java
ivansenic: Same as for the other.

cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
tomekl007: Maybe we should consider adding dedicated on/off settings instead of relying on `updateRateMillis ==0`?

mpenick (approved):

dimas-b (commented):
Belated +1 :)

cql/src/test/java/org/apache/cassandra/stargate/metrics/ClientMetricsTest.java
dougwettlaufer: Neat, that structuring does look clean.

Regarding `happyPath` I'm personally cool with it and use it as well. It's a pretty common naming convention for testing in general

https://en.wikipedia.org/wiki/Happy_path

cql/src/test/java/org/apache/cassandra/stargate/metrics/ClientMetricsTest.java
dougwettlaufer: > DisplayName is nice, but it's pure luxury to have and maintain it, as everything should be figured out from the test method name.

+1

PR #942, opened 2021-05-11
Simplify combined ops for the docs api filter expressions

ivansenic:
Original work from @mpenick, opening so that I can review first and then add anything needed before requesting a review from others..


restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/ValueFilterOperation.java
```diff
@@ -0,0 +1,86 @@
...
+  boolean test(Number filterValue, Double dbValue);
+
+  /**
+   * Validates if the filter input can be used against this predicate.
+   *
+   * <p>Implementations should throw proper exception in case the validation fails.
+   *
+   * @param filterValue Filter value, can be <code>null</code>
+   */
+  default void validateDoubleFilterInput(Number filterValue) {
```
ivansenic, line 62: ok let's call this `validateNumberFilterInput`..

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/CombinedFilterOperation.java
```diff
@@ -21,7 +21,10 @@
  *
  * @param <V> Type of a value used in the predicate.
  */
-public interface CombinedFilterOperation<V>
-    extends StringValueFilterOperation<V>,
-        DoubleValueFilterOperation<V>,
-        BooleanValueFilterOperation<V> {}
+public interface CombinedFilterOperation<V> extends BaseFilterOperation {
+  boolean test(V filterValue, Object dbValue);
```
ivansenic, line 25: no let's have three test methods here as well, there is no reason to force impls to do any casts..

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/CombinedFilterOperation.java
```diff
@@ -21,7 +21,10 @@
  *
  * @param <V> Type of a value used in the predicate.
  */
-public interface CombinedFilterOperation<V>
-    extends StringValueFilterOperation<V>,
-        DoubleValueFilterOperation<V>,
-        BooleanValueFilterOperation<V> {}
+public interface CombinedFilterOperation<V> extends BaseFilterOperation {
```
ivansenic, line 24: This should be called `InputFilterOperation`..

ivansenic (commented):
OK, these changes seems very logical, thanks @mpenick ..

I'll update these comments below and move from draft to the review ready.

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/BaseCondition.java
```diff
@@ -36,6 +36,16 @@ default boolean isPersistenceCondition() {
   /** @return Returns persistence built condition, if this condition supports database querying. */
   Optional<BuiltCondition> getBuiltCondition();
 
+  default Object getValue(Row row, boolean numericBooleans) {
```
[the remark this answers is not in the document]
ivansenic: Removed.

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/CombinedFilterOperation.java
ivansenic: Done.

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/CombinedFilterOperation.java
ivansenic: `GenericFilterOperation` at the end..

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/ValueFilterOperation.java
ivansenic: Done.

ivansenic:
@dimas-b @EricBorczuk Could you have a quick view here and give approval. This is for the feature branch. I think the combined effort Michael and I did here makes things way simpler..

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/impl/CombinedCondition.java
```diff
@@ -64,11 +64,20 @@ protected void validate() {
...
+    Boolean dbValueBoolean = getBoolean(row, isNumericBooleans());
+    Double dbValueDouble = getDouble(row);
+    String dbValueString = getString(row);
 
     CombinedFilterOperation<V> filterOperation = getFilterOperation();
     V queryValue = getQueryValue();
 
-    return filterOperation.test(queryValue, dbValue);
+    // compare against the non-null values, fallback to text compare even if null
+    if (null != dbValueBoolean) {
```
mpenick: It seems off to me to do a `Row#isNull()` then do another check for `!= null`  (then again in the filter operation).

mpenick: It might be better to do something like:

```java
public interface GenericFilterOperation<FV> extends BaseFilterOperation {
  boolean test(FV filterValue, Row row, boolean isNumericBoolean);
}
```

```java
public abstract class InFilterOperation implements GenericFilterOperation<List<?>> {

  @Override
  public boolean test(List<?> filterValue, Row row, boolean isNumericBoolean) {
    // if null, check any is null, avoid .contains(null) as some impl could throw NPE
    if (row.isNull(QueryConstants.BOOLEAN_VALUE_COLUMN_NAME)) {
      Boolean b = row.getBoolean(QueryConstants.BOOLEAN_VALUE_COLUMN_NAME);
      // Do stuff and deal with numeric bool
    } else if (row.isNull(QueryConstants.DOUBLE_VALUE_COLUMN_NAME)) {
      Double d = row.getDouble(QueryConstants.DOUBLE_VALUE_COLUMN_NAME);
      // Do stuff
    } else if (row.isNull(QueryConstants.STRING_VALUE_COLUMN_NAME)) {
      String s = row.getString(QueryConstants.BOOLEAN_VALUE_COLUMN_NAME);
      // Do stuff
    }
    return filterValue.stream().anyMatch(Objects::isNull);
  }
}
```

This would also allow for the removal of the row specific code from `BaseCondition`  (which is sort of out of place) and it could be moved into each specific condition type e.g.

```java
public abstract class NumberCondition implements BaseCondition {

  @Override
  public boolean test(Row row) {
    Double dbValue = row.isNull(QueryConstants.DOUBLE_VALUE_COLUMN_NAME) ? null :
        row.getDouble(QueryConstants.DOUBLE_VALUE_COLUMN_NAME);
    return getFilterOperation().test(getQueryValue(), dbValue);
  }
}

Comment (mpenick) [reply restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/impl/CombinedCondition.java]: Could we also get rid of the numeric boolean settings by adding a utility function like?

```java
public static getBoolean(Row row) {
  int index = row.firstIndexOf(QueryConstants.BOOLEAN_VALUE_COLUMN_NAME);
  if (!row.isNull(index)) {
    ColumnType type = row.columns().get(index).type().rawType();
    return type == Type.Boolean ? row.getBoolean(index) : (row.getByte(index) != 0);
  } else {
    return null;
  }
}
```

ivansenic: There are few suggestions here, so let me comment back on all of them:

* the whole intention was that the `FilterOperation` is not depending on the Row or any other data store related structure.. So I would like to keep that, FilterOperation should not know anything about the data store.. One benefit would be that we could re-use filter operations in other peaces of code, making a unified, well tested filtering available.
* having row specific code in the `BaseCondition` was intentional in order to keep it all in one single class and re-usable by all implementing ones   
* idea about the numeric boolean is not bad at all, @dimas-b @EricBorczuk  what do you think?

mpenick: > the whole intention was that the FilterOperation is not depending on the Row or any other data store related structure.. So I would like to keep that, FilterOperation should not know anything about the data store.. One benefit would be that we could re-use filter operations in other peaces of code, making a unified, well tested filtering available.

A better way to achieve that would be to wrap `Row` in a `Value`-like interface:

```java
interface Value {
  boolean isBoolean();
  Boolean getBoolean();
  boolean isDouble();
  Double getDouble();
  // etc.
}
```

Then:

```java
public interface GenericFilterOperation<FV> extends BaseFilterOperation {
  boolean test(FV filterValue, Value value);
}
```

mpenick: > having row specific code in the BaseCondition was intentional in order to keep it all in one single class and re-usable by all implementing ones

Are we planning to have many more of these? It seems off to have concrete details in the base interface.

mpenick: All this said, I'm fine with moving on. :)

dougwettlaufer (approved):

Issue #945, opened 2021-05-12
Pass all fetcher context via HttpAwareContext


olim7t:
`CassandraFetcher` and all of its subclasses receive `authorizationService` and `dataStoreFactory` via the constructor. This generates a very tedious refactoring if we need to add a new argument or remove an old one (there is no automated way to change all the constructors at once AFAICT).

Such context components should be passed via `HttpAwareContext` instead.

Incidentally, we should rename that class, it already handles more than HTTP-related stuff. `CassandraFetcherContext` feels appropriate.

ivansenic:
@olim7t you can change all consturctors with refactoring in Intellij, right click on constructor `Refactor -> Change signature`. If you do it in a base class, it will be applied to all sub-classes as well.

olim7t:
I think I tried that, are you sure it applies to constructors as well?

But anyway, beyond "refactorability" I still think using the context is the better option.

PR #946, opened 2021-05-12
GraphQL schema-first: handle persistence backends that default to SAI indexes

olim7t:


graphqlapi/src/main/java/io/stargate/graphql/web/HttpAwareContext.java
```diff
@@ -41,9 +43,10 @@
   // For more information.
   private final BatchContext batchContext = new BatchContext();
 
-  public HttpAwareContext(HttpServletRequest request) {
+  public HttpAwareContext(HttpServletRequest request, Persistence persistence) {
```
olim7t, line 46: This is more convenient than modifying all the fetcher constructors.
I think we should do the same for existing constructor arguments, see #945.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IndexModelBuilder.java
```diff
@@ -59,6 +59,10 @@
   IndexModel build() {
     Optional<String> indexClass =
         DirectiveHelper.getStringArgument(cqlIndexDirective, "class", context);
+    // Some persistence backends default to SAI when no class is specified:
+    if (!indexClass.isPresent() && !context.defaultsToRegularIndexes()) {
+      indexClass = Optional.of("org.apache.cassandra.index.sai.StorageAttachedIndex");
```
dimas-b, line 64: Should we also check `Persistence.supportsSAI)` in this case?

Although, in practice we do not have `Persistence` implementations that support neither 2i nor SAI, we could probably produce a nicer error message if it ever happens in the future. WDYT?

olim7t: 👍 you're right, I'll make the change.

tomekl007: Also, maybe log the info, i.e:
`log.info("The index is not set, defaulting to sai")`
?

olim7t: 👍 I added a processing message

dimas-b (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IndexModelBuilder.java
dimas-b: Maybe add a warning to the context, similar to `invalidMapping(...)` but a warning if we default to SAI? Would it be helpful to the end user?

dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IndexModelBuilder.java
olim7t: Yeah that's pretty much what `info()` does. The message will be visible to the user if they select `logs` in the `deploySchema` response.

mpenick (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/FieldModelBuilder.java
```diff
@@ -319,9 +319,15 @@ private static boolean isCompatible(Column.ColumnType type1, Column.ColumnType t
         throw SkipException.INSTANCE;
       }
     }
-    return cqlIndexDirective.map(
-        d ->
-            new IndexModelBuilder(d, parentCqlName, cqlName, cqlType, messagePrefix, context)
-                .build());
+    return cqlIndexDirective.flatMap(
```
mpenick, line 322: Why not move this up into the if statement above?

```java
if (cqlIndexDirective.isPresent()) {
   // ...
   return Optional.of(
     new IndexModelBuilder(d, parentCqlName, cqlName, cqlType, messagePrefix, context)
        .build());
}
```

mpenick: Or move the other checks into the `flatMap()`? 

```java
return cqlIndexDirective.flatMap(
        d -> {
          try {

            if (isPk) {
              invalidMapping("%s: partition or clustering columns can't have an index", messagePrefix);
              throw SkipException.INSTANCE;
            }
            if (isUdtField) {
              invalidMapping("%s: UDT fields can't have an index", messagePrefix);
              throw SkipException.INSTANCE;
            }
            return Optional.of(
                new IndexModelBuilder(d, parentCqlName, cqlName, cqlType, messagePrefix, context)
                    .build());
          } catch (SkipException e) {
            return Optional.empty();
          }
        });
```

olim7t: 👍 I applied your first suggestion.

mpenick (approved):

PR #949, opened 2021-05-13
fix the introspection query for an empty keyspace

tomekl007:


ivansenic (approved):
Perfect.

The playground unfortunately shows:

```
{
  "error": "Response not successful: Received status code 404"
}
```
although we send quite a nice message back:

```
{"errors":[{"message":"Unknown keyspace 'yourKeyspace'"}]}
```

Can it be that something is missed here, I usually get very good error messages in the playground when using the GQL?

tomekl007:
> Perfect.
> 
> The playground unfortunately shows:
> 
> ```
> {
>   "error": "Response not successful: Received status code 404"
> }
> ```
> 
> although we send quite a nice message back:
> 
> ```
> {"errors":[{"message":"Unknown keyspace 'yourKeyspace'"}]}
> ```
> 
> Can it be that something is missed here, I usually get very good error messages in the playground when using the GQL?

I think that's expected behavior. 
When querying for non-existing keyspace:
Screenshot 2021-05-13 at 13 57 47
When querying for existing empty keyspace:
Screenshot 2021-05-13 at 13 57 54

So both situations look ok.

dougwettlaufer (approved):

PR #950, opened 2021-05-13
update schema cache from DeploySchemaFetcherBase

tomekl007:


graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/admin/DeploySchemaFetcherBase.java
```diff
@@ -100,7 +106,8 @@ protected DeploySchemaResponseDto get(
       SchemaSource newSource = schemaSourceDao.insert(keyspaceName, input);
       schemaSourceDao.purgeOldVersions(keyspaceName);
       response.setVersion(newSource.getVersion());
-      // TODO update SchemaFirstCache from here instead of letting it reload from the DB
+      graphqlCache.getGraphQLAndUpdateCache(
```
ivansenic, line 109: result ignored?

ivansenic (approved):
Nothing special to add, except that I see no test updates, I am not sure why is that? At least some unit tests should be updated imo and interaction with the `graphqlCache` confirmed.

tomekl007:
> Nothing special to add, except that I see no test updates, I am not sure why is that? At least some unit tests should be updated imo and interaction with the `graphqlCache` confirmed.

Yes, I was trying to create a unit test for it. However, it requires deep mocking created from scratch as we are not testing those components in a unit way. Fortunately, all tests from the `graphqlfirst` are validating that code path via `CLIENT.deploySchema`.  Maybe that's good enough for this change. @olim7t wdyt? Should we create a unit test that examines the content of the `graphqlCache` or leave it as it is?

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/admin/DeploySchemaFetcherBase.java
tomekl007: yes, that's expected

olim7t, line 109: One of the motivations behind this ticket was to avoid computing the GraphQL schema twice. We do it in this class at line 86:
```java
      ProcessedSchema processedSchema =
          new SchemaProcessor(authorizationService, dataStoreFactory, false)
              .process(input, keyspace);
      ...
      queries =
          CassandraMigrator.forDeployment(migrationStrategy)
              .compute(processedSchema.getMappingModel(), keyspace);
```

But when we call the cache here, the same work will be done a second time (see `DmlGraphHolder.buildGraphql`).

So we should expose a different cache method to directly put the already computed result:
```java
graphqlCache.put(newSource, processedSchema.getGraphql(), authenticationSubject.customProperties()));
```
That method will simply compute the decorated keyspace and call `dmlGraphqls.put()`.
We also need a new `DmlGraphqlHolder` constructor that takes a `GraphQL` directly, but that should be a pretty straightforward change.

(PS: this can race if someone accesses the schema after we've written the `SchemaSource` but before we've completed the cache update, but that doesn't matter, the worst that can happen is that the schema gets computed twice)

olim7t (changes requested):

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
```diff
@@ -223,6 +224,23 @@ public void onKeyspaceChanged(
...
+
+    Keyspace keyspace = dataStore.schema().keyspace(keyspaceName);
+    if (keyspace == null) {
+      LOG.trace("Keyspace {} does not exist", decoratedKeyspaceName);
+      return;
+    }
+    LOG.trace(
+        "Putting new schema version: {} for {}", newSource.getVersion(), decoratedKeyspaceName);
+    DmlGraphqlHolder schemaHolder = new DmlGraphqlHolder(newSource, keyspace, graphql);
+    dmlGraphqls.put(decoratedKeyspaceName, schemaHolder);
```
tomekl007, line 241: @olim7t do you think that we can safely put the value, without this:
```
(currentHolder == null)
            ? dmlGraphqls.putIfAbsent(decoratedKeyspaceName, newHolder) == null
            : dmlGraphqls.replace(decoratedKeyspaceName, currentHolder, newHolder);
```
I think that we can `put()` it without checking the previous version because we are sure that if someone is deploying a new schema, it is the newest one. wdyt?

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/admin/DeploySchemaFetcherBase.java
tomekl007: Done, one question regarding possible race:
https://github.com/stargate/stargate/pull/950#discussion_r633419019

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
olim7t: Yes, think we are safe here. The LWT in `SchemaSourceDao` guarantees that a deployment will never be overwritten. The worst that can happen here is:
* thread A starts a deployment and writes the new row in `schema_source`
* thread B executes a GraphQL query for this keyspace; `getDml()` sees the new row and refreshes the entry
* thread A finishes the deployment, `put()` doesn't see that the new version is already computed, and computes it again

We can tolerate a double init in this corner case, it's just a few more CPU cycles and has no impact for users.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
```diff
@@ -223,6 +224,23 @@ public void onKeyspaceChanged(
     // It is assumed that the data model will only evolve by deploying new GraphQL schema versions.
   }
 
+  public void put(
```
olim7t, line 227: nit: maybe we should name this `putDml()` to be symmetric with `getDml()` (my bad, I didn't think of it when I suggested the name before).

Also: could you place it below `getDml()`, since the two are closely related?

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
```diff
@@ -241,6 +259,12 @@ public void onKeyspaceChanged(
       this.keyspace = keyspace;
     }
 
+    DmlGraphqlHolder(SchemaSource source, Keyspace keyspace, GraphQL graphQL) {
+      this.source = source;
+      this.graphqlFuture.complete(graphQL);
+      this.keyspace = keyspace;
```
olim7t, line 265: We don't need the keyspace here, it's only used by `init()`, which will never be called if this constructor was used. You can remove the constructor argument, that will also simplify the `get()` code.

You can just set the field to `null` for now. The class is getting a bit confusing though, I think I'll revisit it in a follow-up PR.

olim7t (changes requested):

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
tomekl007: done

tomekl007: done

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
```diff
@@ -141,6 +142,23 @@ public GraphQL getDml(
...
+      String keyspaceName, SchemaSource newSource, GraphQL graphql, AuthenticationSubject subject) {
+    Map<String, String> headers = subject.customProperties();
+    DataStore dataStore = buildUserDatastore(subject, headers);
+    String decoratedKeyspaceName = persistence.decorateKeyspaceName(keyspaceName, headers);
+
+    Keyspace keyspace = dataStore.schema().keyspace(keyspaceName);
+    if (keyspace == null) {
+      LOG.trace("Keyspace {} does not exist", decoratedKeyspaceName);
+      return;
+    }
```
olim7t, line 155: I think we can skip that check. We're coming from the deploy schema fetcher that already retrieved the keyspace in order to build the `GraphQL`.
That way you don't need to instantiate a `DataStore` either.

Granted, the keyspace could have been dropped in the meantime, but this can already happen anyway. The schema-first API is not completely immune to external schema changes (that would be nearly impossible to do since there is no way to lock the schema).

olim7t (approved):
Approving now so that you can merge tomorrow, but see one last minor remark below.

olim7t:
I merged #956 which introduces a few conflicts, but they should be pretty straightforward to address.

In fact I think we can store a reference to `GraphqlCache` in `StargateGraphqlContext` to follow the same pattern. `GraphqlCache` is already bound to the Jersey context, so you can inject it in `GraphqlResourceBase` in order to add it whenever we create the GraphQL context.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
tomekl007: Yes, from the current context it should not happen. However, the `putDml` is public on the `GraphqlCache`, therefore it can be used by other places in code in the future. In that case, I think that guarding against this situation may be beneficial, wdyt?

olim7t: Even if calling from another place, you'll need to already have the `GraphQL` instance, so that implies you've built it.

PR #953, opened 2021-05-14
GraphQL schema-first: use user-scoped datastore to fetch schema files

olim7t:


graphqlapi/src/main/java/io/stargate/graphql/web/resources/FilesResource.java
```diff
@@ -98,33 +101,51 @@ public Response getSchema(
...
     }
 
     UUID versionUuid = null;
     if (version != null) {
       try {
         versionUuid = UUID.fromString(version);
       } catch (IllegalArgumentException e) {
-        return notFound(keyspace, version);
+        LOG.warn("Malformed version in URI, this could be an XSS attack: {}", version);
+        return Response.status(Response.Status.BAD_REQUEST).entity("Malformed version").build();
```
olim7t, line 115: This is only tangentially related, but as I moved this check to the top of the method I realized it should be consistent with the keyspace one.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/FilesResource.java
```diff
@@ -98,33 +101,51 @@ public Response getSchema(
...
-    return Response.ok(schemaSource.getContents())
-        .header("Content-Disposition", "inline; filename=" + createFileName(schemaSource))
-        .build();
+      DataStore dataStore =
+          dataStoreFactory.create(
+              authenticationSubject.asUser(),
+              DataStoreOptions.builder().putAllCustomProperties(headers).build());
+      SchemaSource schemaSource =
+          new SchemaSourceDao(dataStore)
+              .getSingleVersion(keyspace, Optional.ofNullable(versionUuid));
```
olim7t, line 136: This is the main change: create a request-scoped datastore every time, instead of using a global, non-scoped one.

testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/FilesResourceTest.java
```diff
@@ -71,13 +71,18 @@ public void schemaFile(@TestKeyspace CqlIdentifier keyspaceId) {
   }
 
   @Test
-  @DisplayName("Should return 404 if schema file coordinates are invalid")
+  @DisplayName("Should return 404 if schema file coordinates do not exist")
   public void schemaFileNotFound(@TestKeyspace CqlIdentifier keyspaceId) {
     CLIENT.expectSchemaFileStatus("unknown_keyspace", Status.NOT_FOUND);
-    CLIENT.expectSchemaFileStatus("malformed keyspace #!?", Status.NOT_FOUND);
```
olim7t, line 77: This was not covering the intended scenario: because the URL parameter was not encoded, it produced an invalid URL that was caught by Jersey and it would never actually hit the method in `FilesResource`.

testing/src/main/java/io/stargate/it/http/graphql/GraphqlClient.java
```diff
@@ -86,4 +89,12 @@ protected String getGraphqlError(String authToken, String url, String graphqlQue
       return null; // never reached
     }
   }
+
+  protected String urlEncode(String parameter) {
+    try {
+      return URLEncoder.encode(parameter, StandardCharsets.UTF_8.name());
+    } catch (UnsupportedEncodingException e) {
+      throw new AssertionError(e);
```
dimas-b, line 97: Why an `Error` as opposed to `RuntimeException` here?

dimas-b (approved):

dougwettlaufer (approved):

PR #954, opened 2021-05-14
Reset warnings in finally for dse

dougwettlaufer:


persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -561,6 +562,13 @@ protected void loginInternally(io.stargate.db.AuthenticatedUser user) {
         CompletableFuture<T> exceptionalFuture = new CompletableFuture<>();
         exceptionalFuture.completeExceptionally(convertExceptionWithWarnings(e));
         return exceptionalFuture;
+      } finally {
+        // this is to clean the thread local in non-TPC thread.
+        // note that: even though request execution happens asynchronously in TPC thread, but the
+        // thread local
+        // values should been already copied by TPC threads when enqueueing TPC tasks via {@link
+        // ExecutorLocals#create()).
+        ClientWarn.instance.resetWarnings();
```
dimas-b, line 571: Why not `ExecutorLocals.set(null)` as we do at the start of this method?

Currently there are two kinds of thread-locals in DSE: 1) warnings 2) CQL traces. I believe if we reset warnings we ought to reset all of them. WDYT?

mpenick: CQL tracing seems to be completely handled on the TPC thread and its init/cleanup is handled in `Message` (in the DSE code).

mpenick (approved):

PR #955, opened 2021-05-17
Prevent "USE <keyspace>" requests

mpenick:
Because we create a new `ClientInfo` and `QueryState` on every request
it's okay to just allow the query then return an error when a set
keyspace result type is detected. The keyspace state will not persist
between requests.

### DIFF
--- a/testing/src/main/java/io/stargate/it/grpc/ExecuteQueryTest.java
+++ b/testing/src/main/java/io/stargate/it/grpc/ExecuteQueryTest.java
@@ -1,11 +1,13 @@
 package io.stargate.it.grpc;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import com.datastax.oss.driver.api.core.CqlIdentifier;
 import com.google.protobuf.Int32Value;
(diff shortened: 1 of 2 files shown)


dougwettlaufer (approved):

tomekl007 (approved):

testing/src/main/java/io/stargate/it/grpc/ExecuteQueryTest.java
```diff
@@ -81,4 +83,49 @@ public void simpleQueryWithPaging() throws InvalidProtocolBufferException {
     assertThat(rs.getPagingState()).isNotNull();
     assertThat(rs.getPageSize().getValue()).isGreaterThan(0);
   }
+
+  @Test
+  public void useKeyspace(@TestKeyspace CqlIdentifier keyspace) {
```
tomekl007, line 88: nit, I think that it can be split into 3 test cases:
- use keyspace
- query without a keyspace
- query with a keyspace
Lower granularity tests give better feedback. wdyt?

mpenick: @ivansenic @tomekl007 It's not three different uses case though. It verifying that state doesn't persist between request when using the same stub/channel.

PR #956, opened 2021-05-18
GraphQL: pass all fetcher dependencies through environment.getContext() (fixes #945)

### LINKED ISSUE #945: Pass all fetcher context via HttpAwareContext

olim7t:
This is just refactoring, there are no functional changes.

* rename `HttpAwareContext` to `StargateGraphqlContext`
* make `authorizationService` and `dataStoreFactory` fields of that object.
* don't pass them to fetcher constructors anymore (and transitively remove them from other components where the only goal was to propagate them to the fetchers, e.g. `DmlSchemaBuilder`)
* `StargateGraphqlContext` is initialized in `GraphqlResourceBase`

### DIFF
--- a/graphqlapi/src/test/java/io/stargate/graphql/schema/GraphQlTestBase.java
+++ b/graphqlapi/src/test/java/io/stargate/graphql/schema/GraphQlTestBase.java
@@ -31,8 +31,8 @@ import io.stargate.db.query.TypedValue.Codec;
 import io.stargate.db.query.builder.AbstractBound;
 import io.stargate.db.query.builder.QueryBuilder;
 import io.stargate.db.schema.Schema;
-import io.stargate.graphql.web.HttpAwareContext;
-import io.stargate.graphql.web.HttpAwareContext.BatchContext;
+import io.stargate.graphql.web.StargateGraphqlContext;
+import io.stargate.graphql.web.StargateGraphqlContext.BatchContext;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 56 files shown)


ivansenic (approved):
Perfect :+1:

graphqlapi/src/test/java/io/stargate/graphql/schema/GraphQlTestBase.java
```diff
@@ -31,8 +31,8 @@
 import io.stargate.db.query.builder.AbstractBound;
 import io.stargate.db.query.builder.QueryBuilder;
 import io.stargate.db.schema.Schema;
-import io.stargate.graphql.web.HttpAwareContext;
-import io.stargate.graphql.web.HttpAwareContext.BatchContext;
+import io.stargate.graphql.web.StargateGraphqlContext;
```
tomekl007, line 34: I think that we need a change in c2 (and possibility CNDB) because those changes are not backward compatible and those projects might use this API.

olim7t: I'll double check, but I'm pretty sure C2 and CNDB do not reference that API.

dougwettlaufer (approved):

tomekl007 (approved):

PR #957, opened 2021-05-18
don't wait for create docs api indexes

ivansenic:
Speeds up the insert of the first document in the collection by not waiting for all indexes to be created. As this can be done independently from the doc write, that actually only depends on the table existing. Performance updates from 5s to under 1s, but only influences the first write.

I would appreciate if anybody has an idea how to confirm in the INT tests that indexes are created?

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
@@ -32,6 +32,7 @@ import java.nio.ByteBuffer;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.*;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.function.UnaryOperator;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -223,20 +224,25 @@ public boolean maybeCreateTable(String keyspaceName, String tableName) {
...
+  public CompletableFuture<Void> maybeCreateTableIndexes(String keyspaceName, String tableName) {
+    CompletableFuture<Void> result;
+    if (dataStore.supportsSAI()) {
+      result = createSAIIndexes(keyspaceName, tableName);
+    } else {
+      result = createDefaultIndexes(keyspaceName, tableName);
     }
+
+    // make sure errors are logger, so that we don't swallow the exception
+    return result.whenComplete(
```
dougwettlaufer, line 236: What's the failure case look like for this on a user perspective? I just inserted my first document and got back a 201 but the indexes didn't create, now what? When I try to get the document back is there an obvious failure with a clear remediation path?

dougwettlaufer:
I like the idea but I'm a little unsure of what this means for the user when something goes wrong.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
ivansenic: Well, currently if the index creation fails, you won't as a user get any feedback. The getting of the document will work I guess, there is not requirement on the existence of the index. Yes it's kind a tricky if we end up failing on the creation.

I see two options:
* dismiss this change and keep the first entry into the collection to wait for the indexes
* update the Checker that was created by Eric and extend to check for the indexes as well (but then what if indexes are not there)

Since this is only affecting the first doc and the collection creation, I am fine with having this closed. Although, there was a feedback on Astra that the first insert takes seconds and that we have those latency spikes..

dougwettlaufer: Is there any sort of remediation we can do? Retry or fix on next read/write?

ivansenic: This is what I would like to avoid as then we require a check if everything is ok on the next read/write.. That's making way more overhead that what we are originally trying to gain. 

I'll put blocked on this one, until #937 because we anyway agreed that we want to move to RxJava here as well.. I'll try to find a solution that works with the checker Eric implemented.

PR #958, opened 2021-05-18
Custom if clause for delete

tomekl007:
Add a new `cql_if` directive that provides a way to use IF on CQL delete query:
https://docs.datastax.com/en/cql-oss/3.3/cql/cql_reference/cqlDelete.html
from the GraphQL-first API:
i.e.:
```
type Mutation {
  deleteFoo(
    k: Int
    minV: Int @cql_if(field: "v", predicate: GTE)
  ) @cql_delete(targetEntity: "Foo")
}
````

### CODE REVIEW

File: graphqlapi/src/main/resources/schemafirst/cql_directives.graphql [hunk 1]
```diff
@@ -218,6 +218,18 @@ enum Predicate {
     CONTAINS
 }
 
+"""
+A predicate used in `@cql_if` to define a condition.
+"""
+enum IfPredicate {
```
tomekl007, line 224: We don't support:
- IN, because:
```
InvalidRequest: Error from server: code=2200 [Invalid query] message="IN predicates on non-primary-key columns (firstname) is not yet supported"
```
but DELETE can target only non-primary-key

- CONTAINS: it doesn't see to work for DELETE `IF` (I was not able to do it from cqlsh)

ivansenic (approved):
Seems good to me..

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/DeployedFetcher.java
```diff
@@ -380,7 +377,40 @@ protected ResultSet executeUnchecked(
...
+        .validateNoFiltering(activeConditions)
+        .ifPresent(
+            message -> {
+              throw new IllegalArgumentException("Invalid arguments: " + message);
+            });
+    return result;
+  }
+
+  protected List<BuiltCondition> bindIf(
+      List<IfConditionModel> whereConditions,
```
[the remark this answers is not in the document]
tomekl007: I don't see a type, what do you mean?

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IfConditionModel.java
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.graphql.schema.graphqlfirst.processor;
+
+import io.stargate.db.query.Predicate;
+import io.stargate.db.query.builder.BuiltCondition;
+
+/** Maps a GraphQL operation argument to a CQL IF clause that is used with DELETE statements. */
+public class IfConditionModel implements ConditionModel {
```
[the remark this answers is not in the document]
tomekl007: The existing WhereCondtiionModel was not created using `@Value.Immutable`. I am not sure if this is on purpose or we can refactor that, @olim7t wdyt?

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/DeployedFetcher.java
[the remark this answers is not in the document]
ivansenic: `List<IfConditionModel> whereConditions` -> `List<IfConditionModel> ifConditions`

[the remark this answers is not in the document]
mpenick: Maybe it can just be `conditions` for both?

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/DeleteModelBuilder.java
```diff
@@ -21,11 +21,7 @@
 import io.stargate.graphql.schema.graphqlfirst.processor.OperationModel.ReturnType;
 import io.stargate.graphql.schema.graphqlfirst.processor.OperationModel.SimpleReturnType;
 import io.stargate.graphql.schema.graphqlfirst.processor.ResponsePayloadModel.TechnicalField;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
dimas-b, line 24: Please expand imports.

mpenick (approved):
:+1:

mpenick (approved):

dimas-b (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IfConditionModel.java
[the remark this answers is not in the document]
olim7t: It just didn't occur to me at the time. These classes don't override equals and hashcode, reference equality is good enough for their purpose. The added benefit would be close to zero so I'm not sure it's worth adding a layer of annotation processing to the module for that.

graphqlapi/src/main/resources/schemafirst/cql_directives.graphql
olim7t: IN seems to work:
```
cqlsh:test> CREATE TABLE foo(k int PRIMARY KEY, v int);
cqlsh:test> INSERT INTO foo (k, v) VALUES (1, 1);
cqlsh:test> DELETE FROM foo WHERE k = 1 IF v in (1, 3);

 [applied]
-----------
      True
```

CONTAINS is indeed invalid. The CQL grammar only allows IN and the arithmetic operators ([source](https://github.com/apache/cassandra/blob/trunk/src/antlr/Parser.g#L1624-L1647)).

I wonder if it's worth creating an entirely separate predicate enum though. We could reuse `Predicate`, even if some values are not allowed in this particular context. This already happens with WHERE conditions (e.g. some predicates are not allowed for primary key fields, require a particular type of index, etc). Checking this during schema validation seems acceptable.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/DeployedFetcher.java
```diff
@@ -371,16 +368,49 @@ protected ResultSet executeUnchecked(
...
+  }
+
+  private <T extends ConditionModel> void bind(
+      List<T> conditions,
+      Predicate<String> hasArgument,
+      Function<String, Object> getArgument,
+      Keyspace keyspace,
+      List<BuiltCondition> result,
+      List<T> activeConditions) {
+    for (T whereCondition : conditions) {
```
olim7t, line 413: Nit:
```suggestion
    for (T condition : conditions) {
```

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/DeployedFetcher.java
```diff
@@ -371,16 +368,49 @@ protected ResultSet executeUnchecked(
...
+
+  protected List<BuiltCondition> bindIf(
+      List<IfConditionModel> conditions,
+      Predicate<String> hasArgument,
+      Function<String, Object> getArgument,
+      Keyspace keyspace) {
+
+    List<BuiltCondition> result = new ArrayList<>();
+    List<IfConditionModel> activeConditions = new ArrayList<>();
+    bind(conditions, hasArgument, getArgument, keyspace, result, activeConditions);
```
olim7t, line 401: We don't use `activeConditions` in this method (which is right BTW: removing IF conditions cannot make a request invalid).

Maybe we could adapt `bind()` to only build it when needed, but that would make it a bit more complicated. Creating one extra list is not the end of the world, I'm fine keeping it like this.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/DeployedFetcher.java
```diff
@@ -371,16 +368,49 @@ protected ResultSet executeUnchecked(
...
+    return result;
+  }
+
+  private <T extends ConditionModel> void bind(
+      List<T> conditions,
+      Predicate<String> hasArgument,
+      Function<String, Object> getArgument,
+      Keyspace keyspace,
+      List<BuiltCondition> result,
+      List<T> activeConditions) {
```
olim7t, line 412: Given how things were refactored, it would make sense to return `activeConditions` as the result of this method, instead of taking it as a parameter.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/ConditionModel.java
```diff
@@ -0,0 +1,15 @@
+package io.stargate.graphql.schema.graphqlfirst.processor;
+
+import io.stargate.db.query.Predicate;
+import io.stargate.db.query.builder.BuiltCondition;
+
+public interface ConditionModel {
```
olim7t, line 6: Why extract that interface?
`WhereConditionModel` and `IfConditionModel` are 100% identical, we could use a single class.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IfConditionModelBuilder.java
```diff
@@ -0,0 +1,85 @@
...
+        ifDirective
+            .flatMap(d -> DirectiveHelper.getEnumArgument(d, "predicate", Predicate.class, context))
+            .orElse(Predicate.EQ);
+
+    FieldModel field = findField(fieldName);
+
+    // The CQL IF works only for regular columns (non PK, CK)
+    if (field.isPartitionKey()) {
+      invalidMapping(
+          "Operation %s: predicate %s is not supported for partition keys (field %s)",
```
olim7t, line 52: Mentioning the predicate is a bit misleading, it implies that other predicates might be supported. Maybe it should just say:
> `@cql_if` is not supported for partition keys.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IfConditionModelBuilder.java
```diff
@@ -0,0 +1,85 @@
...
+
+    Optional<Directive> ifDirective = DirectiveHelper.getDirective("cql_if", argument);
+    String fieldName =
+        ifDirective
+            .flatMap(d -> DirectiveHelper.getStringArgument(d, "field", context))
+            .orElse(argument.getName());
+    Predicate predicate =
+        ifDirective
+            .flatMap(d -> DirectiveHelper.getEnumArgument(d, "predicate", Predicate.class, context))
+            .orElse(Predicate.EQ);
```
olim7t, line 45: `build()` could be pulled up in `ConditionModelBuilderBase` if you introduce two abstract methods:
```java
protected abstract String getDirectiveName();
protected abstract void validate(FieldModel field, Predicate predicate) throws SkipException;
```

(also assuming that you use a single condition class per my other comment, so that the return statement at the end is the same)

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/OperationModelBuilderBase.java
```diff
@@ -90,16 +92,42 @@ protected OperationModelBuilderBase(
...
+   */
+  protected List<IfConditionModel> buildIfConditions(EntityModel entity) throws SkipException {
+    return buildConditionModel(
+        entity,
+        (inputValue, e) ->
+            new IfConditionModelBuilder(inputValue, operationName, e, entities, context),
+        (value) -> !DirectiveHelper.getDirective("cql_if", value).isPresent());
+  }
+
+  private <C extends ConditionModel> List<C> buildConditionModel(
```
olim7t, line 114: Maybe call this `buildConditions` to be consistent with the other methods.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/OperationModelBuilderBase.java
```diff
@@ -90,16 +92,42 @@ protected OperationModelBuilderBase(
    */
   protected List<WhereConditionModel> buildWhereConditions(EntityModel entity)
       throws SkipException {
-    ImmutableList.Builder<WhereConditionModel> whereConditionsBuilder = ImmutableList.builder();
+    return buildConditionModel(
+        entity,
+        (inputValue, e) ->
+            new WhereConditionModelBuilder(inputValue, operationName, e, entities, context),
+        (value) -> DirectiveHelper.getDirective("cql_if", value).isPresent());
+  }
```
olim7t, line 100: This is a bit hard to follow (especially the fact that the predicate is exclusive).

Also it won't detect some invalid usages:
- a parameter is annotated with both `@cql_if` and `@cql_where`
- a parameter is annotated with `@cql_if` in an operation that doesn't support it (e.g. a SELECT).

Maybe we could have a single method that builds the two lists.

graphqlapi/src/main/resources/schemafirst/cql_directives.graphql
```diff
@@ -235,6 +247,23 @@ directive @cql_where (
     predicate: Predicate = EQ
 ) on ARGUMENT_DEFINITION
 
+"""
+Annotates a parameter to customize the DELETE condition that is generated from it.
```
olim7t, line 251: I think we need to be more detailed here. As much as "WHERE condition" should be clear to everyone because of the similarity with SQL, conditional updates are more specific to CQL.

Proposal:
```suggestion
Annotates a parameter to indicate that it will be used as a condition that must test true on the selected entity in order for the mutation to be applied.

This is only allowed for delete and update mutations.
```

olim7t (changes requested):

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IfConditionModelBuilder.java
tomekl007: Nice idea! done

graphqlapi/src/main/resources/schemafirst/cql_directives.graphql
tomekl007: I think that keeping a separate `enum` for both has a couple of important benefits; mainly, it allows GraphQL to catch the wrong condition at the parsing time. It is faster feedback, and we don't need to build the query and process it.
Also, the existing `Predicate` does not support `NEQ`, whereas the new `IfPredicate` is. Are we ok with adding the `NEQ` to the existing `Predicate`

tomekl007: Regarding `IN` support. You are right, I was trying to implement it but there was an error from `QueryBuilder`. It turns out that `QueryBuilder` had a bug for the `IN` predicate. I've fixed it and implemented IN for a custom IF DELETE in:
https://github.com/stargate/stargate/pull/958/commits/eee0a1a56a0cc5b72b3e683f1075b7523afdb77c

persistence-api/src/main/java/io/stargate/db/query/builder/BuiltDML.java
```diff
@@ -278,7 +268,13 @@ private void handleTimestamp() {
     private void handleConditions() {
       for (BuiltCondition bc : dml.data.conditions) {
         Condition.LHS lhs = createLHS(bc.lhs());
-        TypedValue v = handleValue(lhs.toString(), lhs.valueType(), bc.value());
+        TypedValue v;
+        if (bc.predicate().equals(Predicate.IN)) {
```
tomekl007, line 272: We had a bug for `Predicate.IN`, that I've reproduced in this unit test:
https://github.com/stargate/stargate/pull/958/files#diff-f4326ae765fc08166ff59d0f5c07f7c2478f35032e4aa1efd74f8e47b77249baR202
It is fixed now

persistence-api/src/main/java/io/stargate/db/query/builder/BuiltDML.java
```diff
@@ -278,7 +268,21 @@ private void handleTimestamp() {
...
+          // IN works only on a LIST type
+          v = handleValue(lhs.toString(), Column.Type.List.of(lhs.valueType()), bc.value());
+        } else if (bc.predicate().equals(Predicate.CONTAINS)) {
```
tomekl007, line 275: CONTAINS and CONTAINS `bind` was also broken and not tested. I've fixed both of those

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/OperationModelBuilderBase.java
tomekl007: Implemented in: https://github.com/stargate/stargate/pull/958/commits/26da160b1ef7d7b2c899fb047ca13d32e263785b

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/ConditionsModelBuilder.java
```diff
@@ -0,0 +1,106 @@
...
+  public Conditions build(EntityModel entity) throws SkipException {
+    ImmutableList.Builder<ConditionModel> ifConditionsBuilder = ImmutableList.builder();
+    ImmutableList.Builder<ConditionModel> whereConditionsBuilder = ImmutableList.builder();
+    boolean foundErrors = false;
+    for (InputValueDefinition inputValue : operation.getInputValueDefinitions()) {
+
+      if (isCqlIfDirective(inputValue) && isCqlWhereDirective(inputValue)) {
+        invalidMappingReporter.invalidMapping(
+            "You cannot set both: %s and %s directives on the same field: %s",
+            CQL_IF, CQL_IF, inputValue.getName());
```
olim7t, line 56:
```suggestion
            CQL_IF, CQL_WHERE, inputValue.getName());
```

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IfConditionModelBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+  protected String getDirectiveName() {
+    return "cql_if";
+  }
+
+  @Override
+  protected void validate(FieldModel field, Predicate predicate) throws SkipException {
+    // The CQL IF works only for regular columns (non PK, CK)
+    if (field.isPartitionKey()) {
+      invalidMapping(
+          "The cql_if is not supported for partition keys (field %s)", field.getGraphqlName());
```
olim7t, line 43:
```suggestion
          "@cql_if is not supported for partition keys (field %s)", field.getGraphqlName());
```
To be consistent with other messages that refer to directives (e.g. `FieldModelBuilder`, line 213).

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/ConditionModelBuilderBase.java
```diff
@@ -0,0 +1,134 @@
...
+              return SkipException.INSTANCE;
+            });
+  }
+
+  protected abstract String getDirectiveName();
+
+  protected abstract void validate(FieldModel field, Predicate predicate) throws SkipException;
+
+  @Override
+  protected ConditionModel build() throws SkipException {
```
olim7t, line 99: Nit: could you move this method under the constructor? It's the "main" method so I'd rather see it first when I open the class, instead of in the middle of utility methods. (not sure if that's a shared preference across the team, but it's how I tend to organize things...)

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/ConditionModelBuilderBase.java
```diff
@@ -0,0 +1,134 @@
...
+
+    Optional<Directive> ifDirective = DirectiveHelper.getDirective(getDirectiveName(), argument);
```
olim7t, line 101: Nit: at this level we don't know which directive it is.
```suggestion
    Optional<Directive> directive = DirectiveHelper.getDirective(getDirectiveName(), argument);
```

olim7t (approved):
LGTM but please see last round of minor remarks below.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/OperationModelBuilderBase.java
olim7t: > Is there a way to get the info if the operation is Select

You can check in `QueryModelBuilder`:
```java
    ConditionsModelBuilder.Conditions conditions = buildConditions(entity);
    if (!conditions.getIfConditions().isEmpty) {
      ... // fail
    }
    List<ConditionModel> whereConditions = conditions.getWhereConditions();
```

Or maybe pass an additional boolean parameter `allowIfConditions` to `ConditionsModelBuilder.build()`.

Issue #947, opened 2021-05-12
Improve Document API parameter validation


dimas-b:
* Enforce declared defaults
* Validate HTTP parameters early
* Improve / add error messages where applicable

PR #962, opened 2021-05-20
releates to #947: validation for the CollectionsResource

ivansenic:
This is first PR in series of proper validation of resources for the docs api. I started with `CollectionResource` so we can agree what strategy to keep along the path.

Changes with respect to the original ticket:
* separated DTO's so thay can be properly validated
* added `ViolationExceptionMapper` that transforms Jerset exception to response with our `Error` entity and provides meaningful messages back
*  completely refactored the `CollectionTest` - added new cases, optimized for speed, remove not needed calls

Changes that are not related, but I went and did it:
* removed the calls to the  `Converters.writeResponse` -> not needed
* moved to `ErrorCode` usage where possible
* proper dependency injection, with direct class injection (not the initialized object)
* minor fixes where needed

Open questions:
* update test name to `CollectionsResorceIntTest`?
* what about validation messages, do we want to group them in a single class (and reuse when possible)?
* should I go for immutables for bodies (simple interfaces with api definition, impl used in the method param)?

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
@@ -35,9 +35,6 @@ public enum ErrorCode {
       "The table name contains invalid characters. Valid characters are alphanumeric and underscores."),
 
   /** Document API. */
-  DOCS_API_TABLE_NOT_A_COLLECTION(
-      Response.Status.BAD_REQUEST,
-      "The requested database table is not a Documents API Collection."),
   DOCS_API_GENERAL_ARRAY_LENGTH_EXCEEDED(
       Response.Status.BAD_REQUEST,
       String.format(
@@ -62,6 +59,13 @@ public enum ErrorCode {
       String.format(
           "The parameter `page-size` is limited to %d.",
           DocsApiConfiguration.DEFAULT.getMaxPageSize())),
+
+  DOCS_API_GENERAL_UPGRADE_INVALID(
+      Response.Status.BAD_REQUEST, "The collection cannot be upgraded in given manner."),
+
+  DOCS_API_GENERAL_TABLE_NOT_A_COLLECTION(
+      Response.Status.BAD_REQUEST,
+      "The requested database table is not a Documents API Collection."),
 
   DOCS_API_GET_FIELDS_WITHOUT_WHERE(
       Response.Status.BAD_REQUEST, "Selecting fields is not allowed without `where`."),
--- a/restapi/src/main/java/io/stargate/web/docsapi/models/dto/UpgradeCollectionDto.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/models/dto/UpgradeCollectionDto.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.web.docsapi.models.dto;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.stargate.web.docsapi.models.CollectionUpgradeType;
+import io.swagger.annotations.ApiModelProperty;
+import javax.validation.constraints.NotNull;
+
+/** The DTO for the collection upgrade. */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class UpgradeCollectionDto {
+
+  @JsonProperty("upgradeType")
+  @NotNull(message = "`upgradeType` is required to upgrade a collection")
+  private final CollectionUpgradeType upgradeType;
+
+  @JsonCreator
+  public UpgradeCollectionDto(@JsonProperty("upgradeType") CollectionUpgradeType upgradeType) {
+    this.upgradeType = upgradeType;
+  }
+
+  @ApiModelProperty(required = true, value = "The upgrade type, if an upgrade is available.")
+  @JsonProperty("upgradeType")
+  public CollectionUpgradeType getUpgradeType() {
+    return upgradeType;
+  }
+
+  @Override
+  public String toString() {
+    return String.format("UpgradeCollectionDto(upgradeType=%s)", upgradeType);
+  }
+}
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
@@ -10,8 +10,8 @@ import io.stargate.db.schema.Table;
 import io.stargate.web.docsapi.dao.DocumentDB;
 import io.stargate.web.docsapi.exception.ErrorCode;
 import io.stargate.web.docsapi.models.DocCollection;
-import io.stargate.web.docsapi.models.dto.CreateCollectionDto;
-import io.stargate.web.docsapi.models.dto.UpgradeCollectionDto;
+import io.stargate.web.docsapi.models.dto.CreateCollection;
+import io.stargate.web.docsapi.models.dto.UpgradeCollection;
 import io.stargate.web.docsapi.service.CollectionService;
 import io.stargate.web.models.Error;
 import io.stargate.web.models.ResponseWrapper;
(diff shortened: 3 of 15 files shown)


testing/src/main/java/io/stargate/it/http/RestUtils.java
```diff
@@ -74,22 +75,18 @@ public static String postWithHeader(
       Headers headers, String path, String requestBody, int expectedStatusCode) throws IOException {
     OkHttpClient client = client();
 
-    Request request;
+    RequestBody rb =
+        Optional.ofNullable(requestBody)
```
mpenick, line 79: nit: Use of `Optional` seems unnecessary.  Any reason not to use `if (requestBody != null)` ?

mpenick:
> update test name to CollectionsResorceIntTest?

It's already in the `it` namespace. Maybe it's redundant to also have it in the class name?

> should I go for immutables for bodies (simple interfaces with api definition, impl used in the method param)?

TIL, http://immutables.github.io/json.html. That's pretty clean.

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
```diff
@@ -63,6 +60,13 @@
           "The parameter `page-size` is limited to %d.",
           DocsApiConfiguration.DEFAULT.getMaxPageSize())),
 
+  DOCS_API_GENERAL_UPGRADE_INVALID(
+      Response.Status.BAD_REQUEST, "The collection cannot be upgraded in given manner."),
```
dimas-b, line 64: Does the user actually have a choice for how the upgrade is done? I'd just state `The collection cannot be upgraded`. WDYT?

dimas-b: nit: why `GENERAL` in the enum name?

restapi/src/main/java/io/stargate/web/docsapi/models/dto/UpgradeCollectionDto.java
```diff
@@ -0,0 +1,48 @@
...
+import io.stargate.web.docsapi.models.CollectionUpgradeType;
+import io.swagger.annotations.ApiModelProperty;
+import javax.validation.constraints.NotNull;
+
+/** The DTO for the collection upgrade. */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class UpgradeCollectionDto {
+
+  @JsonProperty("upgradeType")
+  @NotNull(message = "`upgradeType` is required to upgrade a collection")
```
dimas-b, line 30: Are these annotations actually effective? I'm just wondering... the field is `private` and hidden behind the annotated constructor argument :thinking:

restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
```diff
@@ -117,34 +132,33 @@ public Response createCollection(
...
           if (res) {
             return Response.status(Response.Status.CREATED).build();
           } else {
             return Response.status(Response.Status.CONFLICT)
                 .entity(
                     new Error(
                         String.format(
-                            "Create failed: collection %s already exists.", info.getName()),
+                            "Create failed: collection %s already exists.", body.getName()),
                         Response.Status.CONFLICT.getStatusCode()))
```
dimas-b, line 162: Is there a way to avoid duplicate references to `Response.Status.CONFLICT` in this fragment?

restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
```diff
@@ -96,8 +110,9 @@ public Response getCollections(
   @ApiResponses(
       value = {
         @ApiResponse(code = 201, message = "Created"),
-        @ApiResponse(code = 409, message = "Conflict", response = Error.class),
         @ApiResponse(code = 401, message = "Unauthorized", response = Error.class),
+        @ApiResponse(code = 409, message = "Conflict", response = Error.class),
+        @ApiResponse(code = 422, message = "Unprocessable entity", response = Error.class),
```
dimas-b, line 115: Where do we actually report `422`?

restapi/src/main/java/io/stargate/web/docsapi/service/CollectionService.java
```diff
@@ -52,6 +54,6 @@ public boolean upgradeCollection(
     if (upgradeType == CollectionUpgradeType.SAI_INDEX_UPGRADE) {
       return docDB.upgradeTableIndexes(keyspaceName, tableName);
     }
-    throw new RuntimeException("Invalid upgrade type: " + upgradeType);
+    throw new ErrorCodeRuntimeException(ErrorCode.DOCS_API_GENERAL_UPGRADE_INVALID);
```
dimas-b, line 57: Why not include `upgradeType`? It appears to be set by the user, so it'd be nice to indicate what exactly is not valid.

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
dimas-b, line 64: nit: why `GENERAL` in the enum name?

dimas-b (changes requested):
Fairly minor comments, but I think it's worth improving the error messages a bit (as commented)

ivansenic:
> > update test name to CollectionsResorceIntTest?
> 
> It's already in the `it` namespace. Maybe it's redundant to also have it in the class name?
> 
> > should I go for immutables for bodies (simple interfaces with api definition, impl used in the method param)?
> 
> TIL, http://immutables.github.io/json.html. That's pretty clean.

Ahh this is way better, trying this! Thanks @mpenick

restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
ivansenic: Sure, but let's keep this for now, as we would further address this once we refactor to Rx and actually move everything away from this class to the service I guess, ok? Then it's gonna be:

```
collectionService.something()
   .map(result -> Response...)
   .onErrorReturn(ErrorCodeRuntimeException.class, e -> e.toResponse())
   .onErrorReturn(Throwable.class, e -> ErrorCode.INTERAL_ERROR.toResponse(e.getMessage())
```

or smth in this direction..

restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
ivansenic: Yea this is per RFC-4918, we were able to process the body, but it was not valid -> https://datatracker.ietf.org/doc/html/rfc4918#section-11.2

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
ivansenic: No idea, to separate from `DOCS_API_SEARCH` for example..

testing/src/main/java/io/stargate/it/http/RestUtils.java
ivansenic: Done.

ivansenic:
> > > update test name to CollectionsResorceIntTest?
> > 
> > 
> > It's already in the `it` namespace. Maybe it's redundant to also have it in the class name?
> > > should I go for immutables for bodies (simple interfaces with api definition, impl used in the method param)?
> > 
> > 
> > TIL, http://immutables.github.io/json.html. That's pretty clean.
> 
> Ahh this is way better, trying this! Thanks @mpenick

@mpenick Unfortunately, the validation does not work with Immutables (see https://github.com/immutables/immutables/issues/26). I am keeping as it s for now.. That's one more point for Guava I guess..

restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
dimas-b: sgmt

restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
dimas-b: I mean, where do we use this error in our code?

ivansenic: ahhh -> https://github.com/stargate/stargate/pull/962/files#diff-cbe6e8c417a014d3779c711a4c0ebfa09d2ab8688fdb66020cdf4ab31500f283R46

dimas-b (approved):

dimas-b (changes requested):
Just noticed that DSE CI still fails:
```
Step #4 - "dse-6.8": [ERROR] Failures: 
Step #4 - "dse-6.8": [ERROR]   CollectionsResourceIntTest$UpgradeCollection.upgradeNotAvailable:208 
Step #4 - "dse-6.8": Expecting:
Step #4 - "dse-6.8":  <"{"description":"The collection cannot be upgraded in given manner.","code":400}">
Step #4 - "dse-6.8": to be equal to:
Step #4 - "dse-6.8":  <"{"description":"The collection cannot be upgraded in given manner.","code":404}">
Step #4 - "dse-6.8": but was not.
```

dimas-b (approved):
Changes LGTM. +1 to merge once the CI is fixed.

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
dimas-b: ok

restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
dimas-b: Thx! TIL

PR #966, opened 2021-05-21
Reduce the number of predicates in nested CQL queries for Docs API filters

dimas-b:
As discussed in https://github.com/stargate/stargate/pull/937#discussion_r632547769


ivansenic (approved):
:+1:

mpenick (approved):
Nice change.

PR #969, opened 2021-05-24
Change to 404 on empty GET

EricBorczuk:
If a document (or path) request is empty, return a 404 instead of a 204 No Content

This would mean that either the data at the particular path is `null` or missing; if it is the empty object it will still be returned with a 200 OK

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
@@ -636,7 +636,7 @@ public class DocumentResourceV2 {
           if (filters.isEmpty()) {
             node = documentService.getJsonAtPath(db, namespace, collection, id, path);
             if (node == null) {
-              return Response.noContent().build();
+              return Response.status(Response.Status.NOT_FOUND).build();
             }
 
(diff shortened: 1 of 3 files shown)


restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -636,7 +636,7 @@ public Response getDocPath(
           if (filters.isEmpty()) {
             node = documentService.getJsonAtPath(db, namespace, collection, id, path);
             if (node == null) {
-              return Response.noContent().build();
+              return Response.status(Response.Status.NOT_FOUND).build();
```
ivansenic, line 639: this change influences the `getDoc` method as well, is this fine?

ivansenic (approved):
Nothing I can really review ;)

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
EricBorczuk: Yep! that's intentional :)

dimas-b (approved):

PR #970, opened 2021-05-24
gRPC API cleanup

mpenick:
This PR is not as scary as it looks I promise. It's mostly renaming and a few shuffles of member variables inside of different messages. It also adds additional tests.

* Moved payload out of `QueryParameters` and into `Query`. This is because
  `QueryParameters` can be reused by queries with different payloads and it
  allows the parameters to be reused.
* Made all the enums more concise.
* Removed redundant `Type` from `TypeSpec`. This is now used for `Basic`
  types only.
* `Result` renamed to `Response` because not all responses are results.
* Renamed generic `payload` to `values` for `Query` and `result_set` for
  `Response`.

### DIFF
--- a/grpc-proto/proto/query.proto
+++ b/grpc-proto/proto/query.proto
@@ -24,17 +24,17 @@ option go_package = "stargate/proto";
 package stargate;
 
 enum Consistency {
-  CONSISTENCY_ANY = 0x00;
-  CONSISTENCY_ONE = 0x01;
-  CONSISTENCY_TWO = 0x02;
-  CONSISTENCY_THREE = 0x03;
-  CONSISTENCY_QUORUM = 0x04;
-  CONSISTENCY_ALL = 0x05;
-  CONSISTENCY_LOCAL_QUORUM = 0x06;
-  CONSISTENCY_EACH_QUORUM = 0x07;
-  CONSISTENCY_SERIAL = 0x08;
-  CONSISTENCY_LOCAL_SERIAL = 0x09;
-  CONSISTENCY_LOCAL_ONE = 0x0A;
+  ANY = 0x00;
+  ONE = 0x01;
+  TWO = 0x02;
+  THREE = 0x03;
+  QUORUM = 0x04;
+  ALL = 0x05;
+  LOCAL_QUORUM = 0x06;
+  EACH_QUORUM = 0x07;
+  SERIAL = 0x08;
+  LOCAL_SERIAL = 0x09;
+  LOCAL_ONE = 0x0A;
 }
 
 message ConsistencyValue {
@@ -106,7 +106,8 @@ message Value {
 
 message Query {
   string cql = 1;
-  QueryParameters parameters = 2;
+  Payload values = 2;
+  QueryParameters parameters = 3;
 }
 
 message Values {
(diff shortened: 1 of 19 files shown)


grpc-proto/proto/query.proto
```diff
@@ -116,95 +117,90 @@ message Values {
...
-  bool tracing = 6;
-  bool skip_metadata = 7;
-  google.protobuf.Int64Value timestamp = 8;
-  ConsistencyValue serial_consistency = 9;
-  google.protobuf.Int32Value now_in_seconds = 10;
+  ConsistencyValue consistency = 2;
+  google.protobuf.Int32Value page_size = 3;
+  google.protobuf.BytesValue paging_state = 4;
+  bool tracing = 5;
+  bool populate_metadata = 6;
```
mpenick, line 135: Changed the name to `populate_metadata` so that this now defaults to `false`.

grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -204,42 +204,35 @@ private void executePrepared(
...
-                        if (rows.rows.isEmpty()
-                            && (parameters.getSkipMetadata()
-                                || rows.resultMetadata.columns.isEmpty())) {
-                          resultBuilder.setPayload(
-                              Payload.newBuilder().setType(payload.getType()).build());
-                        } else {
-                          resultBuilder.setPayload(
-                              handler.processResult((Rows) result, parameters));
-                        }
+                        responseBuilder.setResultSet(
```
mpenick, line 231: This now just returns an empty payload data for empty result sets. Which is both less complex and more consistent.

grpc/src/main/java/io/stargate/grpc/payload/PayloadHandlers.java
```diff
@@ -20,18 +20,18 @@
 import edu.umd.cs.findbugs.annotations.NonNull;
 import io.grpc.Status;
 import io.stargate.grpc.payload.cql.ValuesHandler;
-import io.stargate.proto.QueryOuterClass.Payload.Type;
```
tomekl007, line 23: why dropping this import?

grpc/src/main/java/io/stargate/grpc/payload/cql/ValuesHandler.java
```diff
@@ -35,17 +35,22 @@
...
+import io.stargate.proto.QueryOuterClass.Value;
+import io.stargate.proto.QueryOuterClass.Values;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.List;
 
 public class ValuesHandler implements PayloadHandler {
   @Override
-  public BoundStatement bindValues(Prepared prepared, Payload payload, ByteBuffer unsetValue)
+  public BoundStatement bindValues(Prepared prepared, Any payload, ByteBuffer unsetValue)
```
tomekl007, line 51: Nit, If this will be used from the Scala language, there will be a clash with the Scala syntax. `Any` is equivalent of Java's `Object`

grpc/src/main/java/io/stargate/grpc/payload/PayloadHandlers.java
mpenick: Following usage from other parts of the code. Using `Payload.Type` in most places because `Type` is super generic.

grpc/src/main/java/io/stargate/grpc/payload/cql/ValuesHandler.java
mpenick: Heh, I'll make it fully qualified if we start using it from Scala. :)

grpc-proto/proto/query.proto
```diff
@@ -116,95 +117,90 @@ message Values {
...
-  bool skip_metadata = 7;
-  google.protobuf.Int64Value timestamp = 8;
-  ConsistencyValue serial_consistency = 9;
-  google.protobuf.Int32Value now_in_seconds = 10;
+  ConsistencyValue consistency = 2;
+  google.protobuf.Int32Value page_size = 3;
+  google.protobuf.BytesValue paging_state = 4;
+  bool tracing = 5;
+  bool skip_metadata = 6;
+  google.protobuf.Int64Value timestamp = 7;
```
dimas-b, line 136: nit: Could you add a comment about the expected representation of timestamp values in this case?

grpc/src/test/java/io/stargate/grpc/payload/ProcessResultTests.java
```diff
@@ -46,32 +46,31 @@
   @ParameterizedTest
   @MethodSource("results")
   public void processResult(
```
dimas-b, line 48: The class name ends in `Tests` (with an `s`)... I'm just wondering if maven picks up and actually executes this test :thinking: By convention unit tests have names without the `s`.

dimas-b (approved):

grpc/src/test/java/io/stargate/grpc/payload/ProcessResultTests.java
mpenick: Thanks, typo.

grpc-proto/proto/query.proto
mpenick: Will followup in https://github.com/stargate/stargate/issues/990.

PR #971, opened 2021-05-24
GraphQL: simplify cache internals (fixes #959)

### LINKED ISSUE #959: Revisit `GraphqlCache.DmlGraphqlHolder`
That class could probably benefit from a small refactoring to make the code more readable.

After #950, it can work in 3 different ways:
* CQL-first: the entry will initialize once it has been put in the cache
* GraphQL-first, when the update is triggered by a new row in `schema_source`: same.
* GraphQL-first, when the update is triggered by a `deploySchema` operation: no delayed initialization, the `GraphQL` object is already available.

Maybe we could have 3 separate classes.
Also, with a few changes in `getDml`, we can use lazy init instead of the future/init() mechanism.

olim7t:


graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
```diff
@@ -232,70 +220,99 @@ public void onKeyspaceChanged(
...
+  }
 
-    DmlGraphqlHolder(SchemaSource source, GraphQL graphQL) {
+  /**
+   * An entry used when we need to recompute the schema (because we're reacting to the insertion of
+   * a new row in the source table).
+   */
+  class LazyGraphqlHolder implements GraphqlHolder {
+    private final SchemaSource source;
+    private final Supplier<GraphQL> computeGraphql;
```
tomekl007, line 244: Why we don't use `CompletableFuture`? 
I think that it may be cleaner, and also we may not need explicit synchronization:
https://github.com/stargate/stargate/pull/971/files#diff-9e565971d8098ad42c987560f96a7f22a8ca0ae1987d11e9c2ebfee6d87ee84eR268

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
```diff
@@ -232,70 +220,99 @@ public void onKeyspaceChanged(
...
       return (source == null && otherSource == null)
           || (source != null
               && otherSource != null
               && source.getVersion().equals(otherSource.getVersion()));
     }
 
-    boolean isCqlFirst() {
+    @Override
+    public boolean isCqlFirst() {
       return source == null;
```
tomekl007, line 286: I don't fully understand why it is cqlFirst in that case, can you elaborate more?

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
```diff
@@ -113,21 +113,14 @@ public GraphQL getDml(
...
-    boolean installed =
-        (currentHolder == null)
-            ? dmlGraphqls.putIfAbsent(decoratedKeyspaceName, newHolder) == null
-            : dmlGraphqls.replace(decoratedKeyspaceName, currentHolder, newHolder);
-
-    if (installed) {
-      LOG.trace("Installing new version for {} in the cache", decoratedKeyspaceName);
-      newHolder.init();
-      return newHolder.getGraphql();
+    GraphqlHolder newHolder = new LazyGraphqlHolder(latestSource, keyspace);
```
ivansenic, line 116: you could replace all this below with the:

```
GraphqlHolder result = dmlGraphqls.compute(decoratedKeyspaceName, (k, v) -> Objects.equals(v, currentHolder) ? newHolder : v);
return result == null ? null : currentHolder.getGraphql();
```

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
```diff
@@ -232,70 +220,99 @@ public void onKeyspaceChanged(
...
     }
 
-    void init() {
-      try {
-        graphqlFuture.complete(buildGraphql());
-      } catch (Exception e) {
-        graphqlFuture.completeExceptionally(e);
+    @Override
+    public GraphQL getGraphql() {
+      GraphQL result = graphql;
```
ivansenic, line 264: you can use the suppliers from the guava library that calculate the result only once in order to not care about the concurrency here: https://guava.dev/releases/29.0-jre/api/docs/com/google/common/base/Suppliers.html#memoize-com.google.common.base.Supplier-

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
```diff
@@ -232,70 +220,99 @@ public void onKeyspaceChanged(
...
+      this.graphql = graphql;
+    }
+
+    @Override
+    public GraphQL getGraphql() {
+      return graphql;
+    }
+
+    @Override
+    public boolean matches(@Nullable SchemaSource otherSource) {
```
ivansenic, line 309: this `matches` is not the same as the one for the lazy impl?

ivansenic (commented):
We should really test this class in a well defined unit test.. Even hit the concurrency if possible..

Review resolved as *Comment*: no need to wait for me for merging.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
olim7t: When there is no entry in the `schema_source` table, we expose the legacy schema.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
olim7t: Great idea, will do.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlCache.java
olim7t: Lazy init is a bit simpler than the two-step "put / init" mechanism we had before. We Ivan's suggestion of using Guava memoizer, we don't need the explicit synchronization anymore.

mpenick (approved):

dimas-b (approved):

PR #972, opened 2021-05-24
GraphQL schema-first: check authorization for the DDL queries of a deploySchema

olim7t:


graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/admin/DeploySchemaFetcherBase.java
```diff
@@ -101,6 +97,23 @@ protected DeploySchemaResponseDto get(
     return response;
   }
 
+  private void checkPermissions(StargateGraphqlContext context, String keyspaceName)
+      throws UnauthorizedException {
+
+    AuthorizationService authorizationService = context.getAuthorizationService();
+    AuthenticationSubject subject = context.getSubject();
+    // Ensure we can read the data model to validate the schema:
+    authorizationService.authorizeSchemaRead(
```
tomekl007, line 106: is this for preventing a later error if the user has a write, but doesn't have read permission?

tomekl007 (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/admin/DeploySchemaFetcherBase.java
ivansenic: Is it really possible that somebody could have write, but not read permissions?

dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/admin/DeploySchemaFetcherBase.java
olim7t: My bad, I need to revisit this, there are a number of issues:
* the check at line 113 should be an `authorizeDataWrite` instead of `authorizeSchemaWrite`. It's only about writing the new row in `stargate_graphql.schema_source`.
* this check is about reading the user keyspace's model (which tables are present, what their columns are, etc). But in fact it should also check for write privilege, since the `deploySchema` operation can modify the model (create tables, add columns, etc).
    I'm not sure if write access implies read access.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/migration/CreateIndexQuery.java
```diff
@@ -55,6 +61,18 @@ public String getDescription() {
...
+  @Override
+  public void authorize(AuthorizationService authorizationService, AuthenticationSubject subject)
+      throws UnauthorizedException {
+    authorizationService.authorizeSchemaWrite(
+        subject,
+        table.keyspace(),
+        table.name(),
+        Scope.ALTER,
+        SourceAPI.GRAPHQL,
+        ResourceKind.TABLE);
```
olim7t, line 73: Not sure about this one, so I used the rule that is implemented [in Cassandra](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/cql3/statements/schema/CreateIndexStatement.java#L218): creating an index requires ALTER permission on the target table.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/migration/CreateTableQuery.java
```diff
@@ -62,6 +68,18 @@ public String getDescription() {
...
+  @Override
+  public void authorize(AuthorizationService authorizationService, AuthenticationSubject subject)
+      throws UnauthorizedException {
+    authorizationService.authorizeSchemaWrite(
+        subject,
+        table.keyspace(),
+        table.name(),
+        Scope.CREATE,
+        SourceAPI.GRAPHQL,
+        ResourceKind.TABLE);
```
olim7t, line 80: Not sure here either. It feels wrong to reference a table that does not exist yet.

It looks like [Cassandra](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/cql3/statements/schema/CreateTableStatement.java#L128) checks the `CREATE` permission on the keyspace. Should I leave the table name null?

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/migration/CreateUdtQuery.java
```diff
@@ -51,6 +57,13 @@ public String getDescription() {
     return "Create UDT " + type.name();
   }
 
+  @Override
+  public void authorize(AuthorizationService authorizationService, AuthenticationSubject subject)
+      throws UnauthorizedException {
+    authorizationService.authorizeSchemaWrite(
+        subject, type.keyspace(), type.name(), Scope.CREATE, SourceAPI.GRAPHQL, ResourceKind.TYPE);
+  }
```
olim7t, line 65: Same remark as `CreateTableQuery`.

olim7t (commented):
Second attempt.
It looks like it's not possible to express a global DDL permission on the keyspace, so I'm authorizating each migration request individually.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/migration/CreateIndexQuery.java
dimas-b: `StargateQueryHandler` [uses](https://github.com/stargate/stargate/blob/1bfee7b1d080017c2496ee43e4225e6d9891d3cf/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateQueryHandler.java#L492) `CREATE` and `INDEX` for authorizing `CreateIndexStatement`.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/migration/CreateTableQuery.java
dimas-b: Current code matches the auth logic in `StargateQueryHandler` :+1:

dimas-b (commented):
LGTM :+1: yet, I think we have to align create index permissions between GraphQL and `StargateQueryHandler`

dimas-b (changes requested):
I think we have to align create index permissions between GraphQL and `StargateQueryHandler`

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/migration/CreateIndexQuery.java
olim7t: Good call, it didn't occur to me to look at `StargateQueryHandler`, but it makes sense to align the checks with it.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/migration/CreateUdtQuery.java
```diff
@@ -51,6 +57,13 @@ public String getDescription() {
     return "Create UDT " + type.name();
   }
 
+  @Override
+  public void authorize(AuthorizationService authorizationService, AuthenticationSubject subject)
+      throws UnauthorizedException {
+    authorizationService.authorizeSchemaWrite(
+        subject, type.keyspace(), null, Scope.CREATE, SourceAPI.GRAPHQL, ResourceKind.TYPE);
```
olim7t, line 64: Passing a `null` table name for all UDT permissions, similar to what is done in `StargateQueryHandler`.

dimas-b (approved):

PR #975, opened 2021-05-25
Allow IF conditions for UPDATE queries

tomekl007:
It should allow clients to use `cql_if` directive to specify the IF condition(s) for the CQL UPDATE as per documentation:
https://docs.datastax.com/en/cql-oss/3.3/cql/cql_reference/cqlUpdate.html#Conditionallyupdatingcolumns
todo:
- [x] implement and test `cql_if` for all possible predicates
- [x] add IF EXISTS clause (similar to DELETE already has)
- [x] allow callers to return custom payload for UPDATE, not only boolean stating whether it was applied. It is needed to propagate the info about partial IF UPDATES. 
- [x] integration tests for all predicates


graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/UpdateFetcher.java
```diff
@@ -43,43 +44,51 @@ public UpdateFetcher(UpdateModel model, MappingModel mappingModel) {
...
-        modifiers.add(
-            ValueModifier.set(
-                column.getCqlName(), toCqlValue(graphqlValue, column.getCqlType(), keyspace)));
-      }
-    }
-    if (modifiers.isEmpty()) {
-      throw new IllegalArgumentException(
-          "Input object must have at least one non-PK field set for an update");
-    }
+    prepopulateWithInputData(
```
tomekl007, line 81: I am using a similar pattern to `InsertFetcher` to pre-populate response with data from the input selection. Later, they are overridden if the LWT (IF) fails.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/UpdateFetcher.java
```diff
@@ -93,8 +102,40 @@ protected Boolean get(
             Scope.MODIFY,
             SourceAPI.GRAPHQL);
 
-    executeUnchecked(query, Optional.empty(), Optional.empty(), dataStore);
+    ResultSet resultSet = executeUnchecked(query, Optional.empty(), Optional.empty(), dataStore);
+
+    populateResponseWithResultSet(
+        selectionSet, response, isLwt, resultSet, model.getResponsePayload(), model.getEntity());
+    return response;
```
tomekl007, line 109: we are returning a map of objects - applied and OPTIONAL `responsePayload` now.

testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/UpdateCustomConditionsTest.java
```diff
@@ -0,0 +1,192 @@
...
+
+  @BeforeEach
+  public void cleanupData() {
+    SESSION.execute("truncate table \"User\"");
+  }
+
+  @Test
+  @DisplayName(
+      "Should conditional update user using EQ predicate with a custom payload return type")
+  public void testConditionalUpdateUserUsingPredicateWithCustomPayloadReturnType() {
```
tomekl007, line 78: I will add more integration tests for the rest of the predicates.

testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/UpdateCustomConditionsTest.java
```diff
@@ -0,0 +1,192 @@
...
+            + "}\n"
+            + "type Mutation {\n"
+            + "  updateUser(user: UserInput!): UpdateUserResponse @cql_update(targetEntity: \"User\")\n"
+            + "  updateUserEQCustomPayload(\n"
+            + "    pk: Int\n"
+            + "    username: String\n"
+            + "    age: Int @cql_if(field: \"age\", predicate: EQ)\n"
+            + "  ): UpdateUserResponse @cql_update(targetEntity: \"User\")\n"
+            + "  updateUserIfExists(user: UserInput!): UpdateUserResponse @cql_update\n"
+            + "  updateUserIfExistsCustomPayload(user: UserInput!): UpdateUserResponse @cql_update(ifExists: true)\n"
```
olim7t, line 66: Could you add a test with WHERE...IN conditions?
(I assume your comment below refers to IF conditions)

olim7t (changes requested):

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/UpdateModelBuilder.java
```diff
@@ -43,37 +44,74 @@ MutationModel build() throws SkipException {
...
+        invalidMapping(
+            "Mutation %s: if an update takes an entity input type, it must be the only argument",
+            operationName);
+        throw SkipException.INSTANCE;
+      }
+      entity = entityFromFirstArgument.get();
+      ifConditions = Collections.emptyList();
+    } else {
+      entity = entityFromDirective(cqlUpdateDirective, "update", "cql_update");
+      ConditionsModelBuilder.Conditions conditions = buildOnlyIfConditions(entity);
```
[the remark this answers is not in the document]
tomekl007: The problem is that if we build the where conditions based on the current `buildConditions` method, it is trying to add the to WHERE clause for all fields not annotated with `@cql_if`. So for example, when we have this query:
```
    + "  updateUserEQCustomPayload(\n"
            + "    pk: Int\n"
            + "    username: String\n"
            + "    age: Int @cql_if(field: \"age\", predicate: EQ)\n"
```
It will try to add both `pk` and `username` to WHERE conditions, however, this is not correct. We want to have only the PK in the `where` conditions, age in the `IF` clause, and `username` needs to be updated.
The proposed logic (similar to `DeleteModelBuilder`) will put `username` in the IF clause, and it will not update the value as expected.
(It will also fail with:
` non-primary key argument username must be indexed in order to use it in a condition`
)

If we want to keep the signature of the update as is (without requiring the `cql_where` on each field used in the WHERE) we need to use `getPrimaryKeyWhereConditions`. 
Otherwise, we have two options:
1. Require `cql_where` on all fields that will be used in the `WHERE`clause. If the field is not annotated with the `cql_where`, we are not adding it to a WHERE clause.
2. Use the existing `getPrimaryKeyWhereConditions` to generate WHERE clause + all fields annotated with `cql_where`. If the field is present in both `getPrimaryKeyWhereConditions` and also annotated with `cql_where`, we are overriding its predicate (i.e.: EQ`, `IN`) with the predicate supplied in the `cql_where`

I think that the latter approach is more user-friendly because it does not require `cql_where` directive for all PK fields.
@olim7t Wdyt?
Lastly, as this PR is about IF conditions and already adds more logic (IF EXISTS, returning payload) maybe we should handle `cql_where` in a separate PR (and ticket)?

testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/UpdateCustomConditionsTest.java
tomekl007: added test for IN

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/UpdateFetcher.java
```diff
@@ -16,24 +16,25 @@
...
-import io.stargate.graphql.schema.graphqlfirst.processor.MappingModel;
-import io.stargate.graphql.schema.graphqlfirst.processor.UpdateModel;
+import io.stargate.graphql.schema.graphqlfirst.processor.*;
 import io.stargate.graphql.web.StargateGraphqlContext;
 import java.util.*;
+import java.util.function.Function;
+import java.util.function.Predicate;
 
-public class UpdateFetcher extends DeployedFetcher<Boolean> {
+public class UpdateFetcher extends DeployedFetcher<Map<String, Object>> {
```
olim7t, line 37: This needs to be `Object`, we still allow the mutation to return a simple `Boolean`.

olim7t (commented):
As agreed I continued working on this PR in @tomekl007's absence. Here are a few comments explaining my upcoming changes.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/UpdateFetcher.java
```diff
@@ -44,15 +47,14 @@ public UpdateFetcher(UpdateModel model, MappingModel mappingModel) {
   }
 
   @Override
-  protected Map<String, Object> get(
+  protected Object get(
```
tomekl007, line 50: Isn't it better to have a more strict contract and always return `Map<String, Object>` instead of two cases: (boolean and map)?

tomekl007 (commented):
LGTM, @olim7t  are your changes done? If so, please approve (I cannot do it since I've created this PR)

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/UpdateFetcher.java
olim7t: But then it wouldn't be possible to implement this signature anymore:
```
updateUser(user: UserInput!): Boolean
```
I think we need to support it, it's the most straightforward and logical for non-conditional queries.

olim7t (approved):

PR #976, opened 2021-05-25
Add metrics support for gRPC

dougwettlaufer:


grpc/src/main/java/io/stargate/grpc/service/interceptors/metrics/MetricCollectingServerInterceptor.java
```diff
@@ -0,0 +1,289 @@
+/*
```
dougwettlaufer, line 1: Had to pull this in directly or else there were issues with the classloader.

ivansenic: Let me investigate a bit on this class loading, should not be hapenning..

grpc/src/main/java/io/stargate/grpc/service/interceptors/metrics/MetricCollectingServerInterceptor.java
```diff
@@ -0,0 +1,289 @@
...
+  /** The metrics tag key that belongs to the called service name. */
+  private static final String TAG_SERVICE_NAME = "service";
+  /** The metrics tag key that belongs to the called method name. */
+  private static final String TAG_METHOD_NAME = "method";
+  /** The metrics tag key that belongs to the type of the called method. */
+  private static final String TAG_METHOD_TYPE = "methodType";
+  /** The metrics tag key that belongs to the result status code. */
+  private static final String TAG_STATUS_CODE = "statusCode";
+
+  private final Map<MethodDescriptor<?, ?>, MetricSet> metricsForMethods =
```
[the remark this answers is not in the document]
ivansenic: You don't need the cache, as they are caching already and you can just use it.. It's a perfectly normal situation when you don't have a set of fixed metrics/tags combo.. But let's see if we can just use the micrometer stuff directly..

grpc/src/main/java/io/stargate/grpc/service/interceptors/metrics/MetricCollectingServerInterceptor.java
ivansenic: OK so here is the patch that solves it. Solution:

* load the `io.micrometer.core.instrument.binder.grpc` with the class loader of the `grpc` module (needed because this depends on the `io.grpc` classes)
* load the rest of the micrometer stuff by the `core` class loader (needed not to break the interaction with the MeterRegistry)

I know it looks strange, but this is one of the bizarre situations with the OSGi. Starup now works, did not test further, I have no idea how to simulate grpc traffic. Patch :arrow_down: , I did not include the clean-up.

```diff --git a/grpc/pom.xml b/grpc/pom.xml
index 7d807865..5ab328dc 100644
--- a/grpc/pom.xml
+++ b/grpc/pom.xml
@@ -52,6 +52,10 @@
       <version>${grpc.version}</version>
     </dependency>
     <!-- Other -->
+    <dependency>
+      <groupId>io.micrometer</groupId>
+      <artifactId>micrometer-core</artifactId>
+    </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
@@ -116,11 +120,13 @@
                             org.apache.cassandra.stargate,
                             org.apache.cassandra.stargate.*,
                             org.javatuples,
-                            com.codahale.metrics,
                             com.datastax.oss.driver.api.core,
                             com.datastax.oss.driver.api.core.*,
                             com.datastax.oss.driver.shaded.guava.*,
-                            io.micrometer.*
+                            com.codahale.metrics,
+                            com.codahale.metrics.*,
+                            !io.micrometer.core.instrument.binder.grpc,
+                            io.micrometer.core.*
                         ]]></Import-Package>
             <Export-Package>!*</Export-Package>
             <Embed-Dependency>*;scope=compile|runtime</Embed-Dependency>
diff --git a/grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java b/grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
index 0d32143d..6f94962c 100644
--- a/grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
+++ b/grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
@@ -17,13 +17,13 @@ package io.stargate.grpc.impl;
 
 import io.grpc.Server;
 import io.grpc.ServerBuilder;
+import io.micrometer.core.instrument.binder.grpc.MetricCollectingServerInterceptor;
 import io.stargate.auth.AuthenticationService;
 import io.stargate.core.metrics.api.Metrics;
 import io.stargate.db.Persistence;
 import io.stargate.grpc.service.Service;
 import io.stargate.grpc.service.interceptors.AuthenticationInterceptor;
 import io.stargate.grpc.service.interceptors.RemoteAddressInterceptor;
-import io.stargate.grpc.service.interceptors.metrics.MetricCollectingServerInterceptor;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import org.slf4j.Logger;
diff --git a/pom.xml b/pom.xml
index 748358c6..537587a7 100644
--- a/pom.xml
+++ b/pom.xml
@@ -29,8 +29,8 @@
     <slf4j.version>1.7.30</slf4j.version>
     <logback.version>1.2.3</logback.version>
     <!-- Make sure micrometer and prometheus versions are in sync  -->
-    <micrometer.version>1.6.5</micrometer.version>
-    <prometheus.version>0.9.0</prometheus.version>
+    <micrometer.version>1.7.0</micrometer.version>
+    <prometheus.version>0.10.0</prometheus.version>
     <junit.version>5.7.0</junit.version>
     <mockito.version>3.5.13</mockito.version>
     <jacoco.version>0.8.6</jacoco.version>
@@ -174,6 +174,7 @@
         <module>rate-limiting-global</module>
         <module>grpc-proto</module>
         <module>grpc</module>
+        <module>metrics-jersey</module>
       </modules>
     </profile>
     <profile>
/*
```

ivansenic (changes requested):
Requesting changes just because I think we can simply everything and because I want to be in the loop :nerd_face:

grpc/src/main/java/io/stargate/grpc/service/interceptors/metrics/MetricCollectingServerInterceptor.java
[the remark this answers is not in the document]
dougwettlaufer: If we pull it in from micrometer the map will still be in use.

grpc/src/main/java/io/stargate/grpc/service/interceptors/metrics/MetricCollectingServerCall.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * A simple forwarding server call that collects metrics.
+ *
+ * @param <Q> The type of message received one or more times from the client.
+ * @param <A> The type of message sent one or more times to the client.
+ * @author Daniel Theuke (daniel.theuke@heuboe.de)
+ */
+class MetricCollectingServerCall<Q, A> extends SimpleForwardingServerCall<Q, A> {
+
+  private final Counter responseCounter;
```
tomekl007, line 34: Isn't `Meter` a better type here? It will give us `Counter` + rates.

grpc/src/main/java/io/stargate/grpc/service/interceptors/metrics/MetricCollectingServerInterceptor.java
```diff
@@ -0,0 +1,289 @@
...
+    /**
+     * Uses the given registry to create a {@link Sample Timer.Sample} that will be reported if the
+     * returned consumer is invoked.
+     *
+     * @param registry The registry used to create the sample.
+     * @return The newly created consumer that will report the processing duration since calling
+     *     this method and invoking the returned consumer along with the status code.
+     */
+    public Consumer<Status.Code> newProcessingDurationTiming(final MeterRegistry registry) {
+      final Timer.Sample timerSample = Timer.start(registry);
```
tomekl007, line 285: Shouldn't this be wrapped in the `Consumer` lambda as well? Right now, the invocation of `newProcessingDurationTiming` will start the timer, but it stops it somewhere in the future when the `Consumer<Status.Code>` is invoked. It may result in wrong incorrect reporting.

grpc/src/main/java/io/stargate/grpc/service/interceptors/metrics/MetricCollectingServerCall.java
dougwettlaufer: This is directly from Micrometer

grpc/src/main/java/io/stargate/grpc/service/interceptors/metrics/MetricCollectingServerInterceptor.java
dougwettlaufer: This is directly from Micrometer, few modifications were made to this class. The plan is to actually drop all this and change what we need directly in upstream. Working on the PR now.

grpc/src/main/java/io/stargate/grpc/service/interceptors/metrics/MetricCollectingServerInterceptor.java
dougwettlaufer: Okay so this works for start up because the `grpc` classloader is loading `MetricCollectingServerInterceptor` but looks like I've got some more work to do because it fails when a request is made since the `core` classloader loads `MeterRegistry`.

pom.xml
```diff
@@ -29,14 +29,15 @@
     <slf4j.version>1.7.30</slf4j.version>
     <logback.version>1.2.3</logback.version>
     <!-- Make sure micrometer and prometheus versions are in sync  -->
-    <micrometer.version>1.6.5</micrometer.version>
-    <prometheus.version>0.9.0</prometheus.version>
+    <micrometer.version>1.7.0</micrometer.version>
```
ivansenic, line 32: any reason not to go with the latest, `1.7.3` was just released?

grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
```diff
@@ -59,6 +60,7 @@ public GrpcImpl(
             .intercept(new AuthenticationInterceptor(authenticationService))
             .intercept(new RemoteAddressInterceptor())
             .intercept(new HeadersInterceptor())
+            .intercept(new MetricCollectingServerInterceptor(metrics.getMeterRegistry()))
```
ivansenic, line 63: So we are going with the default stuff from Micrometer right? Any chance we can add a small INT test here, just to confirm it works when being hit?

ivansenic (commented):
Seems fine, I would add a small INT test to check it all..

grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
dougwettlaufer: Correct. I'll add some tests

ivansenic (approved):
Awesome :+1:

testing/src/main/java/io/stargate/it/grpc/ServerMetricsTest.java
```diff
@@ -0,0 +1,249 @@
...
+              System.out.println(lines);
+              assertThat(lines)
+                  .anySatisfy(
+                      metric ->
+                          assertThat(metric)
+                              .contains("grpc_server_processing_duration_seconds_count")
+                              .contains("method=\"ExecuteBatch\"")
+                              .contains("methodType=\"UNARY\"")
+                              .contains("service=\"stargate.Stargate\"")
+                              .contains("statusCode=\"OK\""))
```
tomekl007, line 154: Is it also adding HealthChecks? If not, then we should have a separate issue for adding gRPC health checks

dougwettlaufer: https://github.com/stargate/stargate/issues/1219

Issue #923, opened 2021-05-06
Improve gRPC error handling


mpenick:
The gRPC server implementation returns `Status.UNKNOWN` error for all unhandled exceptions to prevent leaking sensitive information.  We either need to specifically handle all known exception types and map those nicely into grpc error statuses or we can do a catch all interceptor: https://grpc.github.io/grpc-java/javadoc/io/grpc/util/TransmitStatusRuntimeExceptionInterceptor.html

PR #978, opened 2021-05-25
gRPC errors

mpenick:
This PR resolves https://github.com/stargate/stargate/issues/923. 

My first attempt was to add a custom interceptor to like `TransmitStatusRuntimeExceptionInterceptor` (https://grpc.github.io/grpc-java/javadoc/io/grpc/util/TransmitStatusRuntimeExceptionInterceptor.html). However, I think this is more for debugging or trusted internal services. Also, the internal implementation has some complexity around threading and `Server.Listeners`.  Let me know if you think this is a better approach.

I think the preferred approach is to use `StreamObserver#onError(Throwable t)`. So this PR just handles all the unhandled and known exception types from the `Persistence` layer. Otherwise, code inside the service, where we control error handling, should use `StatusException` or `StatusRuntimeException`.

### DIFF
--- a/grpc/src/main/java/io/stargate/grpc/service/Service.java
+++ b/grpc/src/main/java/io/stargate/grpc/service/Service.java
@@ -41,7 +41,6 @@ import io.stargate.db.Result.Rows;
 import io.stargate.db.Statement;
 import io.stargate.grpc.payload.PayloadHandler;
 import io.stargate.grpc.payload.PayloadHandlers;
-import io.stargate.grpc.service.Service.PrepareInfo;
 import io.stargate.proto.QueryOuterClass.AlreadyExists;
 import io.stargate.proto.QueryOuterClass.Batch;
 import io.stargate.proto.QueryOuterClass.BatchParameters;
@@ -76,7 +75,6 @@ import org.apache.cassandra.stargate.exceptions.ReadTimeoutException;
 import org.apache.cassandra.stargate.exceptions.UnavailableException;
 import org.apache.cassandra.stargate.exceptions.WriteFailureException;
 import org.apache.cassandra.stargate.exceptions.WriteTimeoutException;
-import org.apache.tinkerpop.gremlin.structure.T;
 import org.immutables.value.Value;
 
 public class Service extends io.stargate.proto.StargateGrpc.StargateImplBase {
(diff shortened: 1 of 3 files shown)


grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -170,6 +209,174 @@ private void handleError(Throwable throwable, StreamObserver<?> responseObserver
...
+      case UNPREPARED: // Fallthrough
+        onError(responseObserver, Status.INTERNAL, pe);
+        break;
+      case INVALID:
+      case SYNTAX_ERROR: // Fallthrough
+        onError(responseObserver, Status.INVALID_ARGUMENT, pe);
+        break;
+      case TRUNCATE_ERROR:
+      case CDC_WRITE_FAILURE: // Fallthrough
+        onError(responseObserver, Status.ABORTED, pe);
```
dimas-b, line 226: What is the difference between `ABORTED` and `INTERNAL` is this case? It looks like we handle the lower-level error in a generic manner in this case and when we return `INTERNAL` so why is this client-side error code different here?

dimas-b: Are we sure the mutation does not make any changes in this case?

dimas-b (approved):

grpc/src/main/java/io/stargate/grpc/service/Service.java
mpenick: Here's the criteria I used for choosing different error codes:

> A litmus test that may help a service implementor in deciding between FAILED_PRECONDITION, ABORTED, and UNAVAILABLE: (a) Use UNAVAILABLE if the client can retry just the failing call. (b) Use ABORTED if the client should retry at a higher-level (e.g., restarting a read-modify-write sequence). (c) Use FAILED_PRECONDITION if the client should not retry until the system state has been explicitly fixed. E.g., if an "rmdir" fails because the directory is non-empty, FAILED_PRECONDITION should be returned since the client should not retry unless they have first fixed up the directory by deleting files from it.

> Why not expose more specific error codes to the client here?

These are gRPC specific status code. https://grpc.github.io/grpc-java/javadoc/io/grpc/Status.html

I guess we could have our own `Error` type in the result payload instead of using gRPC errors and `Status`, but I'd prefer to stay the gRPC way of handling errors unless there's some critical reason we shouldn't.

dimas-b: Makes sense. Thanks for the info :+1:

mpenick:
Rebased on `master`.

PR #983, opened 2021-06-01
Add RequestFailureReason enum values for DSE

### LINKED ISSUE #982: Inconsistencies in C* RequestFailureReason mapping for DSE
Stargate has its own enum for "RequestFailureReason" - `org.apache.cassandra.stargate.exceptions.RequestFailureReason` that defines error codes for various failure reasons.

However, these codes differ between C* 3.11/ 4.0, Stargate and DSE.

Stargate handles C* OSS `RequestFailureReason` correctly, but may map DSE reasons _incorrectly_.

Cf. `persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java`

E.g. in DSE error code `2` means `INDEX_NOT_AVAILABLE` but in Stargate it means `TIMEOUT`. Also DSE has more error code than Stargate.

dimas-b:
Also add unit tests for C* OSS failure codes (no functional change for OSS)

Fixes #982

### DIFF
--- a/persistence-api/src/main/java/org/apache/cassandra/stargate/exceptions/RequestFailureReason.java
+++ b/persistence-api/src/main/java/org/apache/cassandra/stargate/exceptions/RequestFailureReason.java
@@ -23,12 +23,98 @@ public enum RequestFailureReason {
   UNKNOWN(0),
   READ_TOO_MANY_TOMBSTONES(1),
   TIMEOUT(2),
-  INCOMPATIBLE_SCHEMA(3);
+  INCOMPATIBLE_SCHEMA(3),
+
+  /** The request queried an index but that index wasn't build on the data node. */
+  INDEX_NOT_AVAILABLE,
+
+  /**
+   * The request was writing some data on a CDC enabled table but the CDC commit log segment doesn't
+   * have space anymore (slow CDC consumer).
+   */
+  CDC_SEGMENT_FULL,
+
+  /**
+   * We executed a forwarded counter write but got a failure (any {@link RequestExecutionException}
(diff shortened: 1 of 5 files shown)


dougwettlaufer (approved):

persistence-api/src/main/java/org/apache/cassandra/stargate/exceptions/RequestFailureReason.java
```diff
@@ -23,12 +23,98 @@
...
     this.code = code;
+    this.hasProtocolSupport = true;
+  }
+
+  /**
+   * Enum values without a failure code will be represented as {@link #UNKNOWN} at the protocol
+   * level. These failure reasons have dedicated enum values so that they could be meaningfully
+   * represented in Stargate internal exceptions and log messages.
+   */
+  RequestFailureReason() {
```
mpenick, line 115: Seems like a good solution to me. Esp. because `INCOMPATIBLE_SCHEMA` conflicts with the value of `CDC_SEGMENT_FULL`. The funny thing is the protocol doesn't have any specified values:  https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v5.spec#L1332-L1337

mpenick (approved):

persistence-api/src/main/java/org/apache/cassandra/stargate/exceptions/RequestFailureReason.java
dimas-b: Yeah, and the java driver does not attempt to interpret them :thinking:

PR #985, opened 2021-06-02
Add dependency to the legacy driver in OSS persistence modules

### LINKED ISSUE #645: Stargate does not load non-system keyspaces if it tries to throw an InvalidTypeException
I had an invalid UDF in a test Cassandra cluster and Stargate couldn't parse it, which is expected. However, it caused a side effect: the Stargate node appeared to have started ok but I could only access a few system keyspaces via the GraphQL API: system, data_endpoint_auth, system_schema, stargate_system, system_auth.

Here's an excerpt from Stargate log:

```
ERROR [InternalResponseStage:1] 2021-02-05 17:07:11,020 CassandraDaemon.java:235 - Exception in thread Thread[InternalResponseStage:1,5,main]
java.lang.NoClassDefFoundError: com/datastax/driver/core/exceptions/InvalidTypeException
	at org.apache.cassandra.cql3.functions.JavaBasedUDFunction.<init>(JavaBasedUDFunction.java:193)
	at org.apache.cassandra.cql3.functions.UDFunction.create(UDFunction.java:233)
	at org.apache.cassandra.schema.SchemaKeyspace.createUDFFromRow(SchemaKeyspace.java:1301)
	at org.apache.cassandra.schema.SchemaKeyspace.fetchUDFs(SchemaKeyspace.java:1255)
	at org.apache.cassandra.schema.SchemaKeyspace.fetchFunctions(SchemaKeyspace.java:1240)
	at org.apache.cassandra.schema.SchemaKeyspace.fetchKeyspace(SchemaKeyspace.java:961)
	at org.apache.cassandra.schema.SchemaKeyspace.fetchKeyspacesOnly(SchemaKeyspace.java:951)
	at org.apache.cassandra.schema.SchemaKeyspace.mergeSchema(SchemaKeyspace.java:1401)
	at org.apache.cassandra.schema.SchemaKeyspace.mergeSchemaAndAnnounceVersion(SchemaKeyspace.java:1380)
	at org.apache.cassandra.service.MigrationTask$1.response(MigrationTask.java:91)
	at org.apache.cassandra.net.ResponseVerbHandler.doVerb(ResponseVerbHandler.java:53)
	at org.apache.cassandra.net.MessageDeliveryTask.run(MessageDeliveryTask.java:66)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at org.apache.cassandra.concurrent.NamedThreadFactory.lambda$threadLocalDeallocator$0(NamedThreadFactory.java:84)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.ClassNotFoundException: com.datastax.driver.core.exceptions.InvalidTypeException not found by io.stargate.db.cassandra [1]
	at org.apache.felix.framework.BundleWiringImpl.findClassOrResourceByDelegation(BundleWiringImpl.java:1597)
	at org.apache.felix.framework.BundleWiringImpl.access$300(BundleWiringImpl.java:79)
	at org.apache.felix.framework.BundleWiringImpl$BundleClassLoader.loadClass(BundleWiringImpl.java:1982)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
```

I see that it couldn't find InvalidTypeException class. I found it in one of the Datastax Maven packages and added it to pom.xml.

```
<dependency>
  <groupId>com.datastax.cassandra</groupId>
  <artifactId>cassandra-driver-core</artifactId>
  <version>3.0.8</version>
</dependency>
```

Once the class was available, Stargate handled this case more gracefully and I could access non-system keyspaces in the cluster that still contained the invalid UDF.

```
ERROR [InternalResponseStage:1] 2021-01-20 18:19:11,395 SchemaKeyspace.java:1305 - Cannot load function 'test.bad_udf' from schema: this function won't be available (on this node)
org.apache.cassandra.exceptions.InvalidRequestException: Java source compilation failed:
GENERATED SOURCE ERROR: line 6 (in generated source): Only a type can be imported. org.apache.cassandra.cql3.functions.JavaUDF resolves to a package
[generated code]
	at org.apache.cassandra.cql3.functions.JavaBasedUDFunction.<init>(JavaBasedUDFunction.java:298)
	at org.apache.cassandra.cql3.functions.UDFunction.create(UDFunction.java:233)
	...
INFO  [InternalResponseStage:1] 2021-01-20 18:19:11,420 ColumnFamilyStore.java:427 - Initializing ...
```

Would it be possible to make com.datastax.driver.core.exceptions.InvalidTypeException available to Stargate so that it handles this (and other similar cases) more gracefully?

Here's an example of an invalid UDF in case you'd like to reproduce this.

```
CREATE FUNCTION test_keyspace.invalid_udf(val int)
    CALLED ON NULL INPUT
    RETURNS int
    LANGUAGE java
    AS $$
        if (val == null)
            return 0;
        else
            return val;
        $$;
```

olim7t:
Cassandra uses the driver internally for UDFs. The dependency is
optional, but if the Stargate backend has UDFs, keyspaces can't be
parsed correctly if the driver is not present.

Fixes #645

### DIFF
--- a/persistence-cassandra-3.11/pom.xml
+++ b/persistence-cassandra-3.11/pom.xml
@@ -9,7 +9,13 @@
   <groupId>io.stargate.db.cassandra</groupId>
   <artifactId>persistence-cassandra-3.11</artifactId>
   <properties>
+    <!-- If you update this, make sure to keep `cassandra.bundled-driver.version` in sync -->
     <cassandra.version>3.11.8</cassandra.version>
+    <!--
+      The driver used internally by cassandra-all for UDFs (must match the version declared in
+      cassandra-all's POM).
+    -->
+    <cassandra.bundled-driver.version>3.0.1</cassandra.bundled-driver.version>
   </properties>
   <dependencies>
     <dependency>
@@ -46,10 +52,14 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <!--
+      Redeclare the driver dependency because it's optional in cassandra-all (this is required to
+      correctly handle UDFs).
+    -->
     <dependency>
       <groupId>com.datastax.cassandra</groupId>
       <artifactId>cassandra-driver-core</artifactId>
-      <version>3.0.1</version>
+      <version>${cassandra.bundled-driver.version}</version>
     </dependency>
     <dependency>
       <groupId>org.gridkit.jvmtool</groupId>
--- a/persistence-cassandra-3.11/pom.xml
+++ b/persistence-cassandra-3.11/pom.xml
@@ -47,6 +47,11 @@
       </exclusions>
     </dependency>
     <dependency>
+      <groupId>com.datastax.cassandra</groupId>
+      <artifactId>cassandra-driver-core</artifactId>
+      <version>3.0.1</version>
+    </dependency>
+    <dependency>
       <groupId>org.gridkit.jvmtool</groupId>
       <artifactId>sjk-core</artifactId>
       <version>0.14</version>
--- a/persistence-cassandra-4.0/pom.xml
+++ b/persistence-cassandra-4.0/pom.xml
@@ -47,6 +47,11 @@
       </exclusions>
     </dependency>
     <dependency>
+      <groupId>com.datastax.cassandra</groupId>
+      <artifactId>cassandra-driver-core</artifactId>
+      <version>3.10.0</version>
+    </dependency>
+    <dependency>
       <groupId>org.gridkit.jvmtool</groupId>
       <artifactId>sjk-core</artifactId>
       <version>0.14</version>
--- a/persistence-cassandra-4.0/pom.xml
+++ b/persistence-cassandra-4.0/pom.xml
@@ -9,7 +9,13 @@
   <groupId>io.stargate.db.cassandra</groupId>
   <artifactId>persistence-cassandra-4.0</artifactId>
   <properties>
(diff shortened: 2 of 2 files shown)


olim7t:
No need to change the DSE module, the driver dependency is already explicit.

persistence-cassandra-3.11/pom.xml
```diff
@@ -46,6 +46,11 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>com.datastax.cassandra</groupId>
+      <artifactId>cassandra-driver-core</artifactId>
+      <version>3.0.1</version>
```
dougwettlaufer, line 52: Just curious, why is this an older version?

dougwettlaufer (approved):

persistence-cassandra-3.11/pom.xml
olim7t: It matches the version in `cassandra-all`'s POM. Unfortunately there's no way to infer it, we'll have to maintain the two dependencies in sync manually. I'll add a comment in the code.

mpenick (approved):

dimas-b (approved):

PR #988, opened 2021-06-02
Avoid proactively fetching empty pages when one result set row is expected

dimas-b:
* Use LIMIT in nested filter queries

* Make QueryExecutor enforce non-zero page sizes

* Use non-zero page sizes in document populating queries.
  Note: previous usage of zero page size in these cases was
  incorrect and would lead to error if/when the whole doc does
  not fit into one page.

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
@@ -489,7 +489,17 @@ public class DocumentService {
     }
 
     List<RawDocument> docs =
-        db.executeSelect(keyspace, collection, predicates, 0, null, nestedPopulate(context))
+        db.executeSelect(
+                keyspace,
+                collection,
+                predicates,
+                // We have to use a non-zero page size here in case the whole doc does not
+                // fit into one page because C* will ignore paging state on subsequent requests
+                // if the page size is <= 0.
+                // TODO: use a separate page size parameter for full doc queries?
+                docsApiConfiguration.getSearchPageSize(),
+                null,
+                nestedPopulate(context))
             .take(1)
             .toList()
             .blockingGet();
@@ -906,7 +916,11 @@ public class DocumentService {
                           keyspace,
                           collection,
                           ImmutableList.of(keyPredicate),
-                          0,
+                          // We have to use a non-zero page size here in case the whole doc does not
+                          // fit into one page because C* will ignore paging state on subsequent
+                          // requests if the page size is <= 0.
+                          // TODO: use a separate page size parameter for populating queries?
+                          docsApiConfiguration.getSearchPageSize(),
                           null,
                           populateCtx));
                 })
@@ -1008,13 +1022,16 @@ public class DocumentService {
                         .from(keyspace, collection)
                         .where(keyCondition)
                         .where(buildConditions(nestedCondition, db.treatBooleansAsNumeric()))
+                        .limit(1)
                         .allowFiltering()
                         .build()
                         .bind();
 
                 // If the nested query finds any docs, return the input doc to preserve
-                // the paging order of the main query
-                return db.executeSelect(nestedQuery, 1, null, nestedCtx).take(1).map(nested -> d);
+                // the paging order of the main query.
+                // Note: use page size larger than the query limit to make sure the backend
+                // paginator is exhausted on the first page.
+                return db.executeSelect(nestedQuery, 2, null, nestedCtx).take(1).map(nested -> d);
               });
     }
 
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
@@ -54,6 +54,12 @@ public class QueryExecutor {
       int pageSize,
       ByteBuffer pagingState,
       ExecutionContext context) {
+    if (pageSize <= 0) {
+      // Note: if page size is not set, C* will ignore paging state, but subsequent pages may be
+      // requested by the execute() method, so we require a specific page size for all queries here.
+      throw new IllegalArgumentException("Unsupported page size: " + pageSize);
+    }
+
     BuiltSelect select = (BuiltSelect) query.source().query();
     if (keyDepth < 1 || keyDepth > select.table().primaryKeyColumns().size()) {
       throw new IllegalArgumentException("Invalid document identity depth: " + keyDepth);
@@ -86,10 +92,6 @@ public class QueryExecutor {
                     .execute(
                         query,
                         p -> {
-                          if (pageSize <= 0) {
-                            return p; // if page size is not set pagingState is ignored by C*
-                          }
-
                           ImmutableParameters.Builder builder = p.toBuilder();
                           builder.pageSize(pageSize);
                           if (pagingState != null) {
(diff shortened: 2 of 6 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -1008,13 +1022,16 @@ private ExecutionContext nestedPopulate(ExecutionContext context) {
...
+                        .limit(1)
                         .allowFiltering()
                         .build()
                         .bind();
 
                 // If the nested query finds any docs, return the input doc to preserve
-                // the paging order of the main query
-                return db.executeSelect(nestedQuery, 1, null, nestedCtx).take(1).map(nested -> d);
+                // the paging order of the main query.
+                // Note: use page size larger than the query limit to make sure the backend
```
ivansenic, line 1032: Can you maybe explain me what the page size actually means in this concept? I am not sure if I follow 100% here. And why the limit fix is not good enough?

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -1008,13 +1022,16 @@ private ExecutionContext nestedPopulate(ExecutionContext context) {
                         .from(keyspace, collection)
                         .where(keyCondition)
                         .where(buildConditions(nestedCondition, db.treatBooleansAsNumeric()))
+                        .limit(1)
```
ivansenic, line 1025: this seems smart, I need to pull these changes in my reactive flow after this is merged.

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
```diff
@@ -54,6 +54,12 @@ public QueryExecutor(DataStore dataStore) {
       int pageSize,
       ByteBuffer pagingState,
       ExecutionContext context) {
+    if (pageSize <= 0) {
+      // Note: if page size is not set, C* will ignore paging state, but subsequent pages may be
+      // requested by the execute() method, so we require a specific page size for all queries here.
+      throw new IllegalArgumentException("Unsupported page size: " + pageSize);
```
ivansenic, line 60: can we map this to our new error code exception?

ivansenic (commented):
@dimas-b can we shortly go over it together, this is heavily related to my current work.

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
```diff
@@ -54,6 +54,12 @@ public QueryExecutor(DataStore dataStore) {
       int pageSize,
       ByteBuffer pagingState,
       ExecutionContext context) {
+    if (pageSize <= 0) {
```
ivansenic, line 57: aren't you using the page size 0 when populating the docs?

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -906,7 +916,11 @@ private JsonNode searchUsingCassandraFilters(
                           keyspace,
                           collection,
                           ImmutableList.of(keyPredicate),
-                          0,
+                          // We have to use a non-zero page size here in case the whole doc does not
+                          // fit into one page because C* will ignore paging state on subsequent
+                          // requests if the page size is <= 0.
+                          // TODO: use a separate page size parameter for populating queries?
+                          docsApiConfiguration.getSearchPageSize(),
```
ivansenic, line 923: use max here

restapi/src/test/java/io/stargate/web/docsapi/service/QueryExecutorTest.java
```diff
@@ -205,8 +205,10 @@ void testPartialFullScan(int pageSize) {
   }
 
   @ParameterizedTest
-  @CsvSource({"1", "3", "5", "100"})
+  @CsvSource({"3", "5", "100"})
   void testFullScanFinalPagingState(int pageSize) {
+    // Note: page size must not be a divisor of the result set size for this test to work normally
+    // Otherwise, the last doc will have a non-null paging state
```
dougwettlaufer, line 211: Does this have implications for actual usage or is this only a testing consideration?

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
dougwettlaufer: I think default is 1000

```java
  public static final Integer SEARCH_PAGE_SIZE =
      Integer.getInteger("stargate.document_search_page_size", 1000);
```

ivansenic (approved):
I already checked this but forgot to approve.

restapi/src/test/java/io/stargate/web/docsapi/service/QueryExecutorTest.java
dimas-b: This is just to be able to do asserts on the paging state in a simple way. If the page size aligns exactly with the result set size, there will be an extra empty page fetched. We can account for that in the test, but it'll complicate the assertion logic.

In real applications this is not likely, I guess. Plus our default page size is 1000 rows, so fetching one more empty page at the end should not cause too much overhead.

However, now that you mention it, I think it might be worth splitting the test methods so that we validate paging state separately from doc groupling since pages size should not matter for the latter.... I'll see if that's too much refactoring.

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
dimas-b: This page size maps to the C* result set page size. 

Our Rx pipeline may have to process multiple pages, however C* will ignore paging state if page _size_ is not set. Therefore, we require page size to be set up front because we _may_ have to use it to request page #2, 3, etc...

The `LIMIT 1` is the real fix that makes sure we do not request more than 1 page for "filter" queries. We only care whether we have any rows or no rows. Previously we requested page size `1` for these queries without `LIMIT`, but in this case C* does not always know whether more matching rows are available and will return a non-null paging state, which in turn, will cause our Rx pipeline to fetch page #2 only to reliase there are no more rows available. With `LIMIT 1` and page size 2 (> 1) we can be sure that the first page will not have a paging state. In C* terms the result set will be "exhausted".

Hypothetically, we might do fine without any page size in this case, but as noted above, in the general case page size is required, so I thought it would be ok just to use page size == `2` here.

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
dimas-b: This is not a user-level page size. So if that fails (unexpected) it would be a `500` on the HTTP size. Does it make sense to have a dedicated error code enum in this case?

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
dimas-b: I think I replaced all places where `0` was used with "search page size". Please let me know if I missed something.

dougwettlaufer (approved):

mpenick (approved):

restapi/src/test/java/io/stargate/web/docsapi/service/QueryExecutorTest.java
dimas-b: Updated tests. @dougwettlaufer : is the new test code clearer? I hope it is :)

Issue #991, opened 2021-06-03
Failed to open hints directory


john-gress:
After running for several weeks we are seeing frequent exceptions (see below) around failing to open the hints directory. It looks like this is a result of CentOS automatically deleting files/directories in /tmp after a period of time.

Is there a recommendation for how to deal with this issue? Can the hints directory be relocated somewhere besides /tmp?

See https://stackoverflow.com/questions/61264024/unable-to-open-hint-directory

Thanks,
John Gress

<pre><code>
2021-06-03 14:08:17,384 [HintsWriteExecutor:1] WARN  o.a.cassandra.utils.NativeLibrary - open(/tmp/stargate-cassandra-3.117232734951422798831/hints, O_RDONLY) failed, errno (2).
2021-06-03 14:08:17,384 [HintsWriteExecutor:1] ERROR o.a.cassandra.hints.HintsCatalog - Unable to open directory /tmp/stargate-cassandra-3.117232734951422798831/hints
2021-06-03 14:08:17,385 [HintsWriteExecutor:1] ERROR o.a.c.service.CassandraDaemon - Exception in thread Thread[HintsWriteExecutor:1,5,main]
org.apache.cassandra.io.FSWriteError: java.io.IOException: Unable to open hint directory /tmp/stargate-cassandra-3.117232734951422798831/hints
	at org.apache.cassandra.hints.HintsCatalog.fsyncDirectory(HintsCatalog.java:168)
	at org.apache.cassandra.hints.HintsWriteExecutor$FsyncWritersTask.run(HintsWriteExecutor.java:204)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at org.apache.cassandra.concurrent.NamedThreadFactory.lambda$threadLocalDeallocator$0(NamedThreadFactory.java:84)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.io.IOException: Unable to open hint directory /tmp/stargate-cassandra-3.117232734951422798831/hints
	... 8 common frames omitted
[...]

PR #993, opened 2021-06-04
Allow the persistence base directory to be overridden

mpenick:
On some systems the temp directory is not a suitable location for
Stargate's ephemeral files. This adds the system property
"stargate.basedir" to allow this to be configured.


mpenick:
Resolves  https://github.com/stargate/stargate/issues/991

persistence-api/src/main/java/io/stargate/db/PersistenceActivator.java
```diff
@@ -0,0 +1,32 @@
...
+   * temporary directory is periodically cleaned.
+   *
+   * @return A file handle to the base directory.
+   * @throws IOException if the base directory is invalid or unable to be created.
+   */
+  protected File getBaseDir() throws IOException {
+    if (BASEDIR == null || BASEDIR.isEmpty()) {
+      return Files.createTempDirectory("stargate-" + getActivatorName()).toFile();
+    } else {
+      return Paths.get(BASEDIR).toFile();
```
mpenick, line 29: This is wrong. Fixing.

persistence-api/src/main/java/io/stargate/db/PersistenceActivator.java
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+   * Creates a throw away data directory for Stargate's ephemeral files. It can be overridden using
+   * the system property {@code "stargate.basedir"} which can be useful on systems where the
+   * temporary directory is periodically cleaned.
+   *
+   * @return A file handle to the base directory.
+   * @throws IOException if the base directory is invalid or unable to be created.
+   */
+  protected File getBaseDir() throws IOException {
```
ivansenic, line 25: nit: maybe this can be part of the `BaseActivator` directly.. No need to extend into a dedicated sub-class.. I know that other modules will not use it now, but in general base directory can be helpful to every module.

ivansenic (approved):
:+1: 

Maybe a int test would be nice, just to detect if for example writing to the new directory is possible.

persistence-api/src/main/java/io/stargate/db/PersistenceActivator.java
dimas-b: +1 to moving into `BaseActivator`

persistence-api/src/main/java/io/stargate/db/PersistenceActivator.java
```diff
@@ -0,0 +1,32 @@
...
+    if (BASEDIR == null || BASEDIR.isEmpty()) {
+      return Files.createTempDirectory("stargate-" + getActivatorName()).toFile();
+    } else {
+      return Files.createDirectories(Paths.get(BASEDIR)).toFile();
```
dimas-b, line 29: Should we append `getActivatorName()` to the path in this case too?

persistence-api/src/main/java/io/stargate/db/PersistenceActivator.java
mpenick: What advantage does that give us?

persistence-api/src/main/java/io/stargate/db/PersistenceActivator.java
dimas-b: Or perhaps do the appending in each activator sub-class?

dimas-b (changes requested):

persistence-api/src/main/java/io/stargate/db/PersistenceActivator.java
dimas-b: It may be used later, I guess. It is fairly generic if other modules need it in the future it would be nice to manage the "base" dir in one place.

mpenick: Nm, I just moved it. :)

persistence-api/src/main/java/io/stargate/db/PersistenceActivator.java
```diff
@@ -0,0 +1,32 @@
...
+  private String BASEDIR = System.getProperty("stargate.basedir");
+
+  public PersistenceActivator(String activatorName) {
+    super(activatorName);
+  }
+
+  /**
+   * Creates a throw away data directory for Stargate's ephemeral files. It can be overridden using
+   * the system property {@code "stargate.basedir"} which can be useful on systems where the
+   * temporary directory is periodically cleaned.
```
dimas-b: Just curious: where do this happen?

core/src/main/java/io/stargate/core/activator/BaseActivator.java
```diff
@@ -37,6 +42,8 @@
 public abstract class BaseActivator implements BundleActivator {
   private static final Logger logger = LoggerFactory.getLogger(BaseActivator.class);
 
+  private String BASEDIR = System.getProperty("stargate.basedir");
```
dimas-b, line 45: nit: `final`?

persistence-api/src/main/java/io/stargate/db/PersistenceActivator.java
mpenick: Good idea. Done.

dimas-b (approved):

core/src/main/java/io/stargate/core/activator/BaseActivator.java
mpenick: :+1:

PR #994, opened 2021-06-07
Add resource to attach a JSON schema to a collection

EricBorczuk:
Adds an endpoint to attach a JSON schema to a given collection; When PUT/POST ing a document to that collection, if a schema is present, the document will be checked against the schema, and the request will be rejected if the document doesn't match the schema.

Note that with a schema defined, you can no longer do partial updates of your document (it is disallowed), because you need to be able to check the _entire_ document for correctness, all at once.

In order to ensure that the schema is always available, I hacked the comment on the table to have the schema saved there

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/JsonSchemaResource.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/JsonSchemaResource.java
@@ -1,0 +1,116 @@
+package io.stargate.web.docsapi.resources;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.stargate.web.docsapi.dao.DocumentDB;
+import io.stargate.web.docsapi.models.JsonSchemaResponse;
+import io.stargate.web.docsapi.service.DocsSchemaChecker;
+import io.stargate.web.docsapi.service.JsonSchemaHandler;
+import io.stargate.web.models.Error;
+import io.stargate.web.resources.Db;
+import io.stargate.web.resources.RequestHandler;
+import io.swagger.annotations.*;
+import java.util.concurrent.CompletableFuture;
+import javax.inject.Inject;
+import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.*;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import org.glassfish.jersey.server.ManagedAsync;
+
+@Api(
+    produces = MediaType.APPLICATION_JSON,
+    consumes = MediaType.APPLICATION_JSON,
+    tags = {"documents"})
+@Path("/v2/namespaces/{namespace-id: [a-zA-Z_0-9]+}")
+@Produces(MediaType.APPLICATION_JSON)
+public class JsonSchemaResource {
+  @Inject private Db dbFactory;
+  @Inject private ObjectMapper mapper;
+  @Inject private JsonSchemaHandler jsonSchemaHandler;
+  @Inject private DocsSchemaChecker schemaChecker;
+
+  @PUT
+  @ManagedAsync
+  @ApiOperation(
+      value =
+          "Assign a JSON schema to a collection. This will erase any schema that already exists.")
+  @ApiResponses(
+      value = {
+        @ApiResponse(code = 200, message = "OK"),
+        @ApiResponse(code = 400, message = "Bad request", response = Error.class),
+        @ApiResponse(code = 401, message = "Unauthorized", response = Error.class),
+        @ApiResponse(code = 403, message = "Forbidden", response = Error.class),
+        @ApiResponse(code = 500, message = "Internal Server Error", response = Error.class)
+      })
+  @Path("collections/{collection-id}/json-schema")
+  @Consumes(MediaType.APPLICATION_JSON)
+  @Produces(MediaType.APPLICATION_JSON)
+  public Response attachJsonSchema(
+      @ApiParam(
+              value =
+                  "The token returned from the authorization endpoint. Use this token in each request.",
+              required = true)
+          @HeaderParam("X-Cassandra-Token")
+          String token,
+      @ApiParam(value = "the namespace of the collection", required = true)
+          @PathParam("namespace-id")
+          String namespace,
+      @ApiParam(value = "the collection to add a JSON schema to", required = true)
+          @PathParam("collection-id")
+          String collection,
+      @ApiParam(value = "The JSON schema to attach") String payload,
+      @Context HttpServletRequest request) {
+    return RequestHandler.handle(
+        () -> {
+          DocumentDB db =
+              dbFactory.getDocDataStoreForToken(
+                  token, RequestToHeadersMapper.getAllHeaders(request));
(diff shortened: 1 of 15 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/JsonSchemaHandler.java
```diff
@@ -0,0 +1,163 @@
...
+  /**
+   * Attaches a comment to the given C* table that represents a JSON schema.
+   *
+   * @param namespace
+   * @param collection
+   * @param schemaData
+   */
+  private void writeSchemaToCollection(
+      DocumentDB db, String namespace, String collection, String schemaData) {
+    db.builder().alter().table(namespace, collection).withComment(schemaData).build().execute();
```
dimas-b, line 117: Would it be worth wrapping `schemaData` in another JSON object here, just to make upgrades easier in case we need to store something else in the comment field in the future?

restapi/src/main/java/io/stargate/web/docsapi/service/JsonSchemaHandler.java
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  private CompletableFuture<String> getRawJsonSchemaForCollection(
+      DocumentDB db, String namespace, String collection) {
+    BuiltQuery q =
+        db.builder()
+            .select()
+            .column("comment")
+            .from("system_schema", "tables")
+            .where("keyspace_name", Predicate.EQ, namespace)
```
dimas-b, line 79: This looks risky to me. We may not know the low-level keyspace name in all cases. Could we expose comments via the main Stargate `Schema` API?

dimas-b (changes requested):

restapi/src/main/java/io/stargate/web/docsapi/service/JsonSchemaHandler.java
EricBorczuk: Ah, perhaps - that would make more sense to do

EricBorczuk: yea I forgot about CNDB here and how tenant is added into KS names 😓

restapi/src/main/java/io/stargate/web/docsapi/service/JsonSchemaHandler.java
```diff
@@ -0,0 +1,138 @@
...
+      DocumentDB db, String namespace, String collection, JsonNode schema) {
+    ProcessingReport report = schemaFactory.getSyntaxValidator().validateSchema(schema);
+    JsonSchemaResponse resp = reportToResponse(schema, report);
+    if (report.isSuccess()) {
+      ObjectNode wrappedSchema = mapper.createObjectNode();
+      wrappedSchema.set("schema", schema);
+      writeSchemaToCollection(db, namespace, collection, wrappedSchema.toString());
+      ImmutableKeyspaceAndTable info =
+          ImmutableKeyspaceAndTable.builder().keyspace(namespace).table(collection).build();
+      schemasPerCollection.put(info, schema);
```
dimas-b, line 63: In the face of potentially concurrent "attach" operations, I believe it would be more correct to do `.remove(info)` here and let `getCachedJsonSchema()` fetch the latest schema on request.

restapi/src/main/java/io/stargate/web/docsapi/service/JsonSchemaHandler.java
dimas-b: Also, should be perhaps wait for schema agreement here to ensure subsequent "get" operations work with the latest data? WDYT?

In this case we can probably remove the "wait for schema agreement" call from the integration test.

dimas-b (approved):

restapi/src/main/java/io/stargate/web/docsapi/resources/JsonSchemaResource.java
```diff
@@ -0,0 +1,115 @@
...
+          String collection,
+      @Context HttpServletRequest request) {
+    return RequestHandler.handle(
+        () -> {
+          DocumentDB db =
+              dbFactory.getDocDataStoreForToken(
+                  token, RequestToHeadersMapper.getAllHeaders(request));
+          schemaChecker.checkValidity(namespace, collection, db);
+          JsonSchemaResponse resp =
+              jsonSchemaHandler.getJsonSchemaForCollection(db, namespace, collection);
```
dougwettlaufer, line 111: Late but won't this return 200 and null if the schema doesn't exist?

restapi/src/main/java/io/stargate/web/docsapi/service/JsonSchemaHandler.java
```diff
@@ -0,0 +1,138 @@
...
+    return comment.isEmpty() ? null : comment;
+  }
+
+  public JsonSchemaResponse getJsonSchemaForCollection(
+      DocumentDB db, String namespace, String collection) {
+    String schemaData = getRawJsonSchemaForCollection(db, namespace, collection);
+    try {
+      return new JsonSchemaResponse(mapper.readTree(schemaData).requiredAt("/schema"));
+    } catch (JsonProcessingException e) {
+      return null;
```
dougwettlaufer, line 81: Should this at least log a warn since this shouldn't happen?

restapi/src/main/java/io/stargate/web/docsapi/service/JsonSchemaHandler.java
```diff
@@ -0,0 +1,138 @@
...
+    if (report.isSuccess()) {
+      ObjectNode wrappedSchema = mapper.createObjectNode();
+      wrappedSchema.set("schema", schema);
+      writeSchemaToCollection(db, namespace, collection, wrappedSchema.toString());
+      ImmutableKeyspaceAndTable info =
+          ImmutableKeyspaceAndTable.builder().keyspace(namespace).table(collection).build();
+      schemasPerCollection.remove(info);
+      return resp;
+    } else {
+      throw new ErrorCodeRuntimeException(ErrorCode.DOCS_API_JSON_SCHEMA_INVALID);
```
dougwettlaufer, line 66: Would it be worth returning the report here as well to help troubleshoot what was wrong with the schema?

testing/src/main/java/io/stargate/it/http/docsapi/JsonSchemaResourceIntTest.java
```diff
@@ -0,0 +1,144 @@
...
+    obj = OBJECT_MAPPER.readTree("{\"id\":1, \"name\":\"a\", \"price\":-1}");
+    resp = RestUtils.put(authToken, collectionPath + "/1", obj.toString(), 400);
+    data = OBJECT_MAPPER.readTree(resp);
+    assertThat(data.requiredAt("/description").textValue())
+        .isEqualTo(
+            "Invalid JSON: [numeric instance is lower than the required minimum (minimum: 0, found: -1)]");
+  }
+
+  @Test
+  public void testInvalidSchema() throws IOException {
```
dougwettlaufer, line 136: What about a test for getting a schema from a collection that doesn't have one?

dimas-b: Good point +1

PR #995, opened 2021-06-07
Add support for user-defined types to gRPC

mpenick:
Resolves https://github.com/stargate/stargate/issues/979

Don't be worry about the size. This also fixes licence headers.

### DIFF
--- a/grpc-proto/src/main/java/io/stargate/grpc/Values.java
+++ b/grpc-proto/src/main/java/io/stargate/grpc/Values.java
@@ -74,7 +74,7 @@ public class Values {
         .build();
   }
 
-  public static Value of(Map<String, Value> fields) {
+  public static Value udtOf(Map<String, Value> fields) {
     return Value.newBuilder().setUdt(UdtValue.newBuilder().putAllFields(fields).build()).build();
   }
 }
--- a/grpc-proto/src/main/java/io/stargate/grpc/Values.java
+++ b/grpc-proto/src/main/java/io/stargate/grpc/Values.java
@@ -75,8 +75,6 @@ public class Values {
   }
 
   public static Value of(Map<String, Value> fields) {
-    return Value.newBuilder()
-        .setUdt(UdtValue.newBuilder().putAllFields(fields).build())
-        .build();
+    return Value.newBuilder().setUdt(UdtValue.newBuilder().putAllFields(fields).build()).build();
   }
 }
--- a/grpc-proto/src/main/java/io/stargate/grpc/Values.java
+++ b/grpc-proto/src/main/java/io/stargate/grpc/Values.java
@@ -2,6 +2,7 @@ package io.stargate.grpc;
 
 import com.google.protobuf.ByteString;
 import io.stargate.proto.QueryOuterClass.Collection;
+import io.stargate.proto.QueryOuterClass.UdtValue;
 import io.stargate.proto.QueryOuterClass.Uuid;
 import io.stargate.proto.QueryOuterClass.Value;
 import io.stargate.proto.QueryOuterClass.Value.Null;
@@ -11,6 +12,7 @@ import java.time.LocalDate;
 import java.time.LocalTime;
 import java.time.temporal.ChronoUnit;
 import java.util.Arrays;
+import java.util.Map;
 import java.util.UUID;
 
 public class Values {
@@ -69,6 +71,12 @@ public class Values {
   public static Value of(Value... elements) {
     return Value.newBuilder()
         .setCollection(Collection.newBuilder().addAllElements(Arrays.asList(elements)).build())
+        .build();
+  }
+
+  public static Value of(Map<String, Value> fields) {
+    return Value.newBuilder()
+        .setUdt(UdtValue.newBuilder().putAllFields(fields).build())
         .build();
   }
 }
--- a/grpc-proto/src/main/java/io/stargate/grpc/Values.java
+++ b/grpc-proto/src/main/java/io/stargate/grpc/Values.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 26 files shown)


grpc/src/main/java/io/stargate/grpc/codec/cql/UdtCodec.java
```diff
@@ -0,0 +1,84 @@
...
+      int i = 0;
+      while (input.hasRemaining()) {
+        if (i > fieldCount) {
+          throw new IllegalArgumentException(
+              String.format("Too many fields in encoded UDT, expected %d", fieldCount));
+        }
+        String fieldName = udtFields.get(i).name();
+        ColumnType fieldType = Objects.requireNonNull(udtFields.get(i).type());
+        ValueCodec fieldCodec = ValueCodecs.get(fieldType.rawType());
+        builder.putFields(fieldName, decodeValue(input, fieldCodec, fieldType));
```
mpenick, line 75: As it's currently implemented. If you have a UDT as defined: `CREATE TYPE udt (f1 int, f2 text, f3 uuid)`. If you insert a subset of the fields (which is valid) the result will return `null`s for the values that weren't provided e.g.

Insert: `Values.of(ImmutableMap.of("f2", Values.of("abc")))`
Result: `Values.of(ImmutableMap.of("f1", Values.NULL, "f2", Values.of("abc"), "f3", Values.NULL))`

There is no way to distinguish between not provided and `null` provided.

My question: Should we remove `null` values from the result? So that if we would get back `Values.of(ImmutableMap.of("f2", Values.of("abc")))` in the example above when the other fields (f1 and f3) are either `null` or not provided?

dimas-b: Can we keep track of _provided_ property names and remove `null` values only for not provided properties?

dimas-b: If that is not possible, I think it would still make sense to remove `null` values from the response for the same of avoiding ambiguity with respect to what is stored in the database.

grpc/src/main/java/io/stargate/grpc/codec/cql/UdtCodec.java
```diff
@@ -0,0 +1,84 @@
+package io.stargate.grpc.codec.cql;
```
dimas-b, line 1: Should we add copyright headers to new files? I think some other gRPC files are missing them too :thinking:

dimas-b (changes requested):

grpc/src/main/java/io/stargate/grpc/codec/cql/UdtCodec.java
mpenick: thanks! I wish we had a plug-in for this...

grpc/src/main/java/io/stargate/grpc/codec/cql/UdtCodec.java
mpenick: > Can we keep track of provided property names and remove null values only for not provided properties?

No. Unfortunately we *can't* track that. The storage engine just stores `null` for both actual `null` and not provided.

grpc-proto/src/main/java/io/stargate/grpc/Values.java
```diff
@@ -71,4 +73,8 @@ public static Value of(Value... elements) {
         .setCollection(Collection.newBuilder().addAllElements(Arrays.asList(elements)).build())
         .build();
   }
+
+  public static Value of(Map<String, Value> fields) {
```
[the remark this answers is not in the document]
mpenick: I agree. I will think of a way to disambiguate.

grpc/src/main/java/io/stargate/grpc/codec/cql/UdtCodec.java
mpenick: I do like the idea of removing `null` values from the result, but this is not what's stored in the DB.

For reference this is what's returned from persistence:

```
token@cqlsh:testks> CREATE TABLE udt_test (key text PRIMARY KEY, value udt);
token@cqlsh:testks> INSERT INTO udt_test (key, value) VALUES ( 'k1', { f1: 1 });
token@cqlsh:testks> SELECT * FROM udt_test ;

 key | value
-----+-------------------
  k1 | {f1: 1, f2: null}

(1 rows)
token@cqlsh:testks> INSERT INTO udt_test (key, value) VALUES ( 'k1', { f2: 'v1' });
token@cqlsh:testks> SELECT * FROM udt_test ;

 key | value
-----+----------------------
  k1 | {f1: null, f2: 'v1'}

(1 rows)
token@cqlsh:testks> INSERT INTO udt_test (key, value) VALUES ( 'k1', { f1: null, f2: null });
token@cqlsh:testks> SELECT * FROM udt_test ;

 key | value
-----+-------
  k1 |  null

(1 rows)
token@cqlsh:testks> INSERT INTO udt_test (key, value) VALUES ( 'k1', { f1: 1, f2: null });
token@cqlsh:testks> SELECT * FROM udt_test ;

 key | value
-----+-------------------
  k1 | {f1: 1, f2: null}

(1 rows)
```

grpc/src/main/java/io/stargate/grpc/codec/cql/UdtCodec.java
dimas-b: I believe we discussed using `spotless` for this, just did not get around to implementing it :)

grpc/src/main/java/io/stargate/grpc/codec/cql/UdtCodec.java
dimas-b: Sorry, I'm bit lost here. `cqlsh` probably does not show the difference, but can we tell `UNSET` from `NULL` in our code?

grpc-proto/src/main/java/io/stargate/grpc/Values.java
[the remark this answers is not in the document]
mpenick: Done. Now `Values.udtOf(...)`.

grpc/src/main/java/io/stargate/grpc/codec/cql/UdtCodec.java
mpenick: Fixed.

dimas-b (approved):

grpc/src/main/java/io/stargate/grpc/codec/cql/UdtCodec.java
dimas-b: thx! It looks like `Values.java` is still missing the header.

dimas-b (approved):

dougwettlaufer (approved):

PR #996, opened 2021-06-07
GraphQL schema-first: allow custom consistency levels

olim7t:


graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/DeleteFetcher.java
```diff
@@ -101,7 +97,7 @@ protected Object get(
             Scope.DELETE,
             SourceAPI.GRAPHQL);
 
-    ResultSet resultSet = executeUnchecked(query, Optional.empty(), Optional.empty(), dataStore);
+    ResultSet resultSet = executeUnchecked(query, dataStore);
```
olim7t, line 100: All fetchers now override `getDatastoreParameters()` to set those arguments, instead of passing them here.

graphqlapi/src/main/resources/schemafirst/cql_directives.graphql
```diff
@@ -114,6 +114,16 @@ directive @cql_index (
...
+"""
+The consistency level of the CQL SELECT generated for a query.
+"""
+enum QueryConsistency {
+    LOCAL_ONE
+    LOCAL_QUORUM
+    ALL
+    SERIAL
+    LOCAL_SERIAL
+}
```
olim7t, line 126: The available constants for each enum, and the default values, mimic the existing behavior of the CQL-first API.

olim7t (commented):
I'm not sure how to write integration tests for this. As far as I know there's no way to spy the CLs that get passed to the persistence backend.
Tracing would be a possibility, but we don't expose that through GraphQL currently (and I don't think it's something we'll add).

mpenick:
> I'm not sure how to write integration tests for this. As far as I know there's no way to spy the CLs that get passed to the persistence backend.

One possibility might be to make it a unit test and mock out `DataStore` to validate parameters make it all the way through. This is what I'm doing  for gRPC, but I can understand if you decide against it because it could be huge effort.

graphqlapi/src/main/resources/schemafirst/cql_directives.graphql
```diff
@@ -114,6 +114,16 @@ directive @cql_index (
 
 ) on FIELD_DEFINITION
 
+"""
+The consistency level of the CQL SELECT generated for a query.
+"""
+enum QueryConsistency {
+    LOCAL_ONE
+    LOCAL_QUORUM
+    ALL
```
dimas-b, line 123: We have `ALL`, but no `QUORUM`?

dimas-b (approved):

dimas-b:
> [...] mock out DataStore

Docs API code has something like that already.... It should be possible to move it into a common module and share.

dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/MutationFetcher.java
```diff
@@ -0,0 +1,51 @@
...
+    super(mappingModel);
+    this.model = model;
+  }
+
+  @Override
+  protected Parameters getDatastoreParameters(DataFetchingEnvironment environment) {
+    ConsistencyLevel consistencyLevel = model.getConsistencyLevel().orElse(DEFAULT_CONSISTENCY);
+    ConsistencyLevel serialConsistencyLevel =
+        model.getSerialConsistencyLevel().orElse(DEFAULT_SERIAL_CONSISTENCY);
+    if (consistencyLevel == DEFAULT_CONSISTENCY
```
tomekl007, line 40: maybe we could do:
```
if (consistencyLevel == DEFAULT_PARAMETERS.defaultConsistency 
&& serialConsistencyLevel == DEFAULT_PARAMETERS.defaultSerialConsistency)
```
?
By doing so, we are not risking that when someone change CLs in the `DEFAULT_PARAMETERS` this code will get out of sync

graphqlapi/src/main/resources/schemafirst/cql_directives.graphql
olim7t: I used the same levels as the CQL-first API, I wasn't there since the beginning so I don't know how they were decided.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/MutationFetcher.java
olim7t: `DEFAULT_PARAMETERS` is itself defined in terms of `DEFAULT_CONSISTENCY`, so it's kind of cyclical.

olim7t:
> One possibility might be to make it a unit test and mock out `DataStore` to validate parameters make it all the way through. This is what I'm doing for gRPC, but I can understand if you decide against it because it could be huge effort.

Actually we already have some infrastructure in place for unit tests, I'll give it a try.

olim7t:
Force-pushed to rebase on latest master (in particular #1003).

dimas-b (approved):

PR #998, opened 2021-06-08
Persist processing messages in graphql schema

tomekl007:


testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/SchemaDeploymentTest.java
```diff
@@ -386,4 +387,52 @@ public void deploySchemaErrorNoStacktrace(@TestKeyspace CqlIdentifier keyspaceId
...
+
+    // when
+    Object response =
+        CLIENT.executeSchemaQuery(
+            String.format(
+                "{\n"
+                    + "  schemas(keyspace: \"%s\") {\n"
+                    + "    version\n"
+                    + "    contents\n"
+                    + "    logs { message, category, locations } \n"
```
tomekl007, line 425: I think that we should not propagate the tuple to callers, but encapsulate that using a dedicated type (https://github.com/stargate/stargate/pull/998/files#diff-c7af9322a34dce3e3db2ca1af6aad38ddb19a2d8f831a2c67e2ea0f636daebf6R55). It seems a lot more user friendly, than letting our users handle tuples. @olim7t do you agree?

dougwettlaufer (approved):

testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/SchemaDeploymentTest.java
olim7t: We already have such a type in the schema: `AdminSchemaBuilder.LOG_TYPE`.

tomekl007: The existing `LOG` type uses enum for category and custom type for location:
```
   newObject()          .name("Log")          .description("A log emitted by a deploy mutation.")          .field(newFieldDefinition().name("message").type(nonNull(GraphQLString)).build())          .field(newFieldDefinition().name("category").type(nonNull(LOG_CATEGORY_ENUM)).build())          .field(newFieldDefinition().name("locations").type(list(LOCATION_TYPE)).build())          .build();
```
As we discussed in the ticket, we wanted to have a simpler solution here (`tuple<string,string,string>`). Therefore I think we need this new type.

tomekl007:
Because it impacts the schema of the main grapqhl table and this feature was not requested by any user, we decided to close it without merging. If there will be a customer's demand for this ticket, we will re-open and add the schema migration to it.

tomekl007:
Because it impacts the schema of the main grapqhl table and this feature was not requested by any user, we decided to close it without merging. If there will be a customer's demand for this ticket, we will re-open and add the schema migration to it.

Issue #1001, opened 2021-06-08
Cql client encryption is not working


irumman:
Hi,

I am trying to use stargate 1.0.25 in front of Cassandra 3.11 with enabling ssl for both server and client. Server ssl is working.
But client encryption is not working. 

```
env JAVA_OPTS='-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=5005 -Dstargate.unsafe.cassandra_config_path=/stargate/stargate/cass_configs/cassandra.yaml' \
./starctl --cluster-name cass \
--cluster-seed 10.10.10.21 \
--cluster-version 3.11 \
--listen 10.0.0.161 \
--dc cass \
--rack 3a \
--enable-auth

```

In cassandra.yaml, I have

```
client_encryption_options:
    enabled: true
    optional: true
    require_client_auth: true

```
In log, I can see:

`INFO  [main] 2021-06-08 15:11:05,959 Server.java:157 - Starting listening for CQL clients on /0.0.0.0:9042 (unencrypted)...`
    
Is there any other configuration that I should use to enable client ssl for cql?

Thanks.

mpenick:
Thanks for reporting the issue. These options were not available because, in the past, we've assumed that TLS termination would happen in front of Stargate, but I think Stargate should be able to handle this. I've added a PR with a possible fix,  it resolves it using `-Dstargate.unsafe.cassandra_config_path` as you've done, but I'm wondering if we should add these to the `startctl` cli in some way?

irumman:
With the latest version of code where it is using `<groupId>org.apache.cassandra</groupId><artifactId>cassandra-all</artifactId><version>4.0.0</version>`, we were getting the following error during cql ssl connection: 

```
WARN  [nioEventLoopGroup-2-2] 2021-10-07 09:18:12,529 ChannelInitializer.java:97 - Failed to initialize a channel. Closing: [id: 0x9fbab0a0, L:/127.0.0.1:9042 - R:/127.0.0.1:60082]
java.lang.NoSuchMethodError: io.netty.handler.ssl.SslContextBuilder.protocols(Ljava/lang/Iterable;)Lio/netty/handler/ssl/SslContextBuilder;
        at org.apache.cassandra.security.SSLFactory.createNettySslContext(SSLFactory.java:329)

```

It was basically happening due to conflict on netty. To solve it, we has to modify the netty version to the same as `org.apache.cassandra 4.0.0` which is:

```
    <dependency>
[...]

PR #1002, opened 2021-06-09
Fix client encryption

mpenick:
One possible resolution to https://github.com/stargate/stargate/issues/1001.

I'm wonder if this should be configurable from the `starctl` cli, but this is a bunch of options to put on the command line.

### DIFF
--- a/cql/src/main/java/io/stargate/cql/CqlActivator.java
+++ b/cql/src/main/java/io/stargate/cql/CqlActivator.java
@@ -22,11 +22,14 @@ import io.stargate.cql.impl.CqlImpl;
 import io.stargate.db.DbActivator;
 import io.stargate.db.Persistence;
 import io.stargate.db.metrics.api.ClientInfoMetricsTagProvider;
+import java.io.File;
 import java.net.InetAddress;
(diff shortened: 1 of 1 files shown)


cql/src/main/java/io/stargate/cql/CqlActivator.java
```diff
@@ -95,13 +98,20 @@ private static Config makeConfig() {
 
       Integer cqlPort = Integer.getInteger("stargate.cql_port", 9042);
 
-      Config c = new Config();
+      Config c;
+      String cassandraConfigPath = System.getProperty("stargate.unsafe.cassandra_config_path", "");
```
tomekl007, line 102: I don't see how this change relates to `Fix client encryption`. Can you elaborate more on it?

mpenick: It allows configuring the client TLS using a `cassandra.yaml` file. But this property name is already taken: https://github.com/stargate/stargate/pull/602

mpenick:
This PR is really stale and the property name is already taken. I'm not sure if it's even a good fix anymore.

PR #1005, opened 2021-06-10
Support incremental updates for UPDATE queries

tomekl007:
Adds a new `@cql_increment` directive that allows customers to use CQL increment operations on Counter, List, and Set via Grapqhl-first API, for example:
```
type Mutation {
  updateCounter(
    k: Int
    cInc: Int @cql_increment(field: "c")
  ): Boolean
  
  appendList(
    k: Int
    l: [Int] @cql_increment
  ): Boolean

  prependList(
    k: Int
    l: [Int] @cql_increment(prepend: true)
  ): Boolean
}
```
todo: 
- [x] support for counter 
- [x] support for append list
- [x] support for prepend list
- [x] validation


graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/DirectiveModelsBuilder.java
```diff
@@ -0,0 +1,296 @@
...
+          "Operation %s: could not find field %s in type %s",
+          operationName, fieldName, entity.getGraphqlName());
+    }
+    return result;
+  }
+
+  static class BuildState {
+
+    private boolean foundErrors = false;
+    private boolean isWhereCondition = false;
```
olim7t, line 270: Hrmm, not a big fan of this, although I understand that you did it to allow helper methods to modify the booleans.

The whole class is kind of hard to follow now. Maybe breaking down the three cases (select/delete/update) into submethods would help...

olim7t (changes requested):

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/UpdateModelBuilder.java
```diff
@@ -117,7 +120,26 @@ MutationModel build() throws SkipException {
...
+  }
+
+  private Optional<IncrementModel> getAndValidate(List<IncrementModel> incrementModel)
+      throws SkipException {
+
+    if (incrementModel.size() > 1) {
+      invalidMapping(
+          "The %s directive can be set only on one field, but it was set on %s fields.",
+          CqlDirectives.INCREMENT, incrementModel.size());
+      throw SkipException.INSTANCE;
```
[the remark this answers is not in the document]
tomekl007: done

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IncrementModelBuilder.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+  }
+
+  private void validateCounter(TypeName fieldInputType) throws SkipException {
+    String typeName = fieldInputType.getName();
+    if (typeName.equalsIgnoreCase(COUNTER_TYPE_NAME)) {
+      // counter graph-ql field can be BIGINT or INT
+      checkArgumentIsAnyOfTypes(
+          Arrays.asList(CqlScalar.BIGINT.getGraphqlType(), Scalars.GraphQLInt));
+    }
```
olim7t, line 103: This should fail if the type is not `Counter`.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/UpdateModel.java
```diff
@@ -82,6 +85,10 @@ public ReturnType getReturnType() {
     return responsePayload;
   }
 
+  public List<IncrementModel> getIncrementModel() {
```
olim7t, line 88:
```suggestion
  public List<IncrementModel> getIncrementModels() {
```

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/UpdateModelBuilder.java
```diff
@@ -94,14 +95,16 @@ MutationModel build() throws SkipException {
       entity = entityFromFirstArgument.get();
       whereConditions = entity.getPrimaryKeyWhereConditions();
       ifConditions = Collections.emptyList();
+      incrementModels = Collections.emptyList();
     } else {
       entity = entityFromDirective(cqlUpdateDirective, "update", CqlDirectives.UPDATE);
-      ConditionModels conditions =
-          new ConditionModelsBuilder(operation, OperationType.UPDATE, entity, entities, context)
+      ArgumentDirectiveModels directives =
+          new DirectiveModelsBuilder(operation, OperationType.UPDATE, entity, entities, context)
```
olim7t, line 102: Since we renamed the class, the builder should be renamed as well.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IncrementModelBuilder.java
```diff
@@ -0,0 +1,146 @@
...
+
+    Type<?> argumentType = TypeHelper.unwrapNonNull(argument.getType());
+
+    boolean hasProperType =
+        scalarTypes.stream()
+            .anyMatch(
+                type -> {
+                  if (argumentType instanceof TypeName) {
+                    return type.getName().equals(((TypeName) argumentType).getName());
+                  }
```
olim7t, line 131: Nit: you can extract `argumentType.getName()` outside of the loop.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IncrementModelBuilder.java
```diff
@@ -0,0 +1,146 @@
...
+    Type<?> fieldInputType =
+        toInput(field.getGraphqlType(), argument, entity, field, entities, operationName);
+    // if it is non-list type
+    if (fieldInputType instanceof TypeName) {
+      validateCounter((TypeName) fieldInputType);
+      validatePrepend(field, prepend);
+    } else {
+      // otherwise it must be a list
+      checkArgumentIsListOf(argument, entity, entities, operationName, field);
+    }
```
olim7t, line 94: We should also fail if `prepend` is set and the column is a a CQL set.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IncrementModelBuilder.java
```diff
@@ -0,0 +1,146 @@
...
+
+    Type<?> fieldInputType =
+        toInput(field.getGraphqlType(), argument, entity, field, entities, operationName);
+    // if it is non-list type
+    if (fieldInputType instanceof TypeName) {
+      validateCounter((TypeName) fieldInputType);
+      validatePrepend(field, prepend);
+    } else {
+      // otherwise it must be a list
+      checkArgumentIsListOf(argument, entity, entities, operationName, field);
```
olim7t, line 93: This not the right check.

What that method does is check that the argument _is a list of_ the field type (e.g. field = `Int` => argument = `[Int]`).

What we want is to check that both the argument and the field are lists of the same element type. For example:
```graphql
    type List {
      id: ID
      l: [Int]
    }
    type Mutation {
      updateList(
        id: ID
        lSuffix: [Int] @cql_increment(field: "l")
      ): Boolean @cql_update(targetEntity: "List")
    }
```

PS: I tried to deploy the above and to my surprise it worked. Turns out that there is a bug in `checkArgumentIsListOf`. I'll open a separate ticket to fix it.

olim7t (changes requested):

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IncrementModelBuilder.java
tomekl007: To use it in lambda we would still need to copy it:
```
 String argumentTypeName = null;
    if (argumentType instanceof TypeName) {
      argumentTypeName = ((TypeName) argumentType).getName();
    }
    String finalArgumentTypeName = argumentTypeName;
```
so I am not sure if extracting it is a good choice.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IncrementModelBuilder.java
tomekl007: Is it even possible to create a set using the GrapqhQL API?
I don't see any test covering that case.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IncrementModelBuilder.java
tomekl007: fixed in:
https://github.com/stargate/stargate/pull/1015

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IncrementModelBuilder.java
olim7t: Addressed in #e9b4f103

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IncrementModelBuilder.java
olim7t: The CQL type is known at this stage, via `field.getCqlType()`. Fixed in 3e2b33f0.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IncrementModelBuilder.java
olim7t: Even after fixing the method, we still have to change the call here, as I said it's not the right check. Also addressed in 3e2b33f0.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/IncrementModelBuilder.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.*;
+
+public class IncrementModelBuilder extends ArgumentDirectiveModelBuilderBase<IncrementModel> {
+
+  // The types that we allow for an argument that represents an increment to a counter:
+  private static final List<String> COUNTER_INCREMENT_TYPES =
+      ImmutableList.of(
+          Scalars.GraphQLInt.getName(),
+          CqlScalar.BIGINT.getGraphqlType().getName(),
+          CqlScalar.COUNTER.getGraphqlType().getName());
```
olim7t, line 37: I've added `Counter`, it feels natural to use the same type as the field, and this didn't require any additional code.

olim7t (approved):
I've addressed a couple of remaining items, and attempted a refactoring of the builder in 43e418e9.

PR #1008, opened 2021-06-10
Gracefully handle missing default keyspace (fixes #1007)

### LINKED ISSUE #1007: Gracefully handle missing default keyspace
When handling requests to `/graphql`, we default to the oldest non-system keyspace:`GraphqlCache.getDefaultKeyspaceName()`.

It's possible that that keyspace does not exist. Currently this causes an NPE, we should return a clear GraphQL error instead.

IIRC this used to work, maybe it got lost when the cache was refactored for schema-first.

olim7t:


tomekl007 (approved):
can we add an IT for it?

dougwettlaufer (approved):

PR #1010, opened 2021-06-10
Expose decorated partition keys for comparison at Stargate level

dimas-b:


persistence-api/src/main/java/io/stargate/db/ComparableKey.java
```diff
@@ -0,0 +1,87 @@
...
+    this.delegate = delegate;
+  }
+
+  @Override
+  public int compareTo(@NotNull ComparableKey<?> o) {
+    if (equals(o)) {
+      return 0;
+    }
+
+    if (delegateClass != o.delegateClass) {
```
tomekl007, line 46: As I understand, when someone will create an alternative implementation of `ComparableKey` for which the `equals(o)` returns false, but the delegation classes are the same, this compare will return 0. However, those alternative implementations may differ. Not sure if this is a real problem.

ivansenic, line 46: why not `equal` check?

tomekl007: Making this class `final` would be enough. Thx

ivansenic (commented):
Seems quite clean, not able to say if this part below is correct though:

```
Clustering key = metadata.partitionKeyAsClusteringComparator().make(rawKeyValues);
DecoratedKey decoratedKey = metadata.partitioner.decorateKey(key.serializeAsPartitionKey());
return new ComparableKey<>(PartitionPosition.class, decoratedKey);
```

persistence-api/src/main/java/io/stargate/db/ComparableKey.java
```diff
@@ -0,0 +1,87 @@
...
+      // This is not an expected use case. It is handled only for the sake of supporting the
+      // contract of the compareTo() method whose signature in this class allows comparing objects
+      // with different delegate classes.
+      int cmp = delegateClass.getName().compareTo(o.delegateClass.getName());
+      if (cmp != 0) {
+        return cmp;
+      }
+
+      throw new IllegalStateException(
+          "Mixing different delegate classes with the same name: " + delegateClass.getName());
```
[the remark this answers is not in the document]
dimas-b: No, `compareTo` is a generic method. Implementations are usually scoped to a particular class so passing in an incompatible class will yield a `ClassCastException`.

persistence-api/src/main/java/io/stargate/db/ComparableKey.java
dimas-b: AFAIK, class objects are interned. `==` is equivalent to `.equals()`

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/RowDecoratorImpl.java
```diff
@@ -0,0 +1,46 @@
...
+            .map(c -> c.name.toCQLString())
+            .collect(Collectors.toList()));
+    this.metadata = tableMetadata;
+  }
+
+  @Override
+  protected ComparableKey<?> decoratePrimaryKey(Object... rawKeyValues) {
+    Clustering key = metadata.partitionKeyAsClusteringComparator().make(rawKeyValues);
+    DecoratedKey decoratedKey = metadata.partitioner.decorateKey(key.serializeAsPartitionKey());
+    return new ComparableKey<>(PartitionPosition.class, decoratedKey);
```
dimas-b, line 44: This is modelled after `Cql3Utils.createPartitionKeyFromSolrKey()` CC: @ivansenic

persistence-api/src/main/java/io/stargate/db/ComparableKey.java
dimas-b: Made `final` and rebased in a13660492e87dfffd22775d865b4df298fe6f7c0

persistence-api/src/main/java/io/stargate/db/ComparableKey.java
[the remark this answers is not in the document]
dimas-b: We can have different classes with the same name because of OSGi (potentially).

persistence-api/src/test/java/io/stargate/db/ComparableKeyTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+  @Test
+  void differentClass() {
+    // This is not an expected use case, but it's included for the sake of coverage
+    ComparableKey<?> key1 = new ComparableKey<>(String.class, "a");
+    ComparableKey<?> key2 = new ComparableKey<>(Integer.class, 2);
+    assertThat(key1).isNotEqualTo(key2);
+    assertThat(key2).isNotEqualTo(key1);
+
+    assertThat(key1.compareTo(key2)).isGreaterThan(0);
```
EricBorczuk, line 69: This is essentially because `S` comes after `I` alphabetically, right? If I read the comparison code way up there correctly

persistence-api/src/main/java/io/stargate/db/ComparableKey.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    T other = (T) o.delegate;
+    return delegate.compareTo(other);
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
EricBorczuk, line 65: Can this just reuse `ComparableKey#compareTo` somehow? Should it?

EricBorczuk (approved):
This looks good! I left a few comments for my own understanding, and a suggestion that is non-blocking

persistence-api/src/main/java/io/stargate/db/ComparableKey.java
```diff
@@ -0,0 +1,87 @@
...
+  public int compareTo(@NotNull ComparableKey<?> o) {
+    if (equals(o)) {
+      return 0;
+    }
+
+    if (delegateClass != o.delegateClass) {
+      // This is not an expected use case. It is handled only for the sake of supporting the
+      // contract of the compareTo() method whose signature in this class allows comparing objects
+      // with different delegate classes.
+      int cmp = delegateClass.getName().compareTo(o.delegateClass.getName());
```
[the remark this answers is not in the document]
dimas-b: I'd say, it is a safety valve for us to avoid throwing `ClassCastExcepion` during comparison because the signature of `compareTo` in this class allows passing in irrelevant objects.

In normal use cases this line should not be executed (but we have test code for it just in case).

persistence-api/src/test/java/io/stargate/db/ComparableKeyTest.java
dimas-b: yes

persistence-api/src/main/java/io/stargate/db/ComparableKey.java
dimas-b: In general this is recommended, but the contract of `compareTo` does not _require_ that `x.compareTo(y) == 0` must imply that `x.equals(y)`.

Since this code may have to work with delegate classes that we do not know about, I'd rather delegate to the real key separately in `.equals()` and in `.compareTo`.

persistence-api/src/main/java/io/stargate/db/ComparableKey.java
ivansenic: OK, fine..

persistence-api/src/main/java/io/stargate/db/ComparableKey.java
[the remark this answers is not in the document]
ivansenic: Resolving, I was reading the return from the compare wrongly.

ivansenic (approved):
OK let merge this, we need some extra INT tests once this is actually used..

PR #1014, opened 2021-06-14
GraphQL schema-first: fix type checks for conditions (fixes #1013)

### LINKED ISSUE #1013: GraphQL schema-first: fix type checks for conditions
`ConditionModelBuilder.checkArgumentIsListOf` uses `isEqualTo` to compare the expected and actual types. This will always succeed if both types are lists, because it doesn't recurse into subtypes. We need to compare the element types instead.

`ConditionModelBuilder.checkArgumentIsSameAs` might be subject to the same issue.

olim7t:


dimas-b (approved):
:+1:

dimas-b (approved):

tomekl007 (approved):

PR #1016, opened 2021-06-15
extending INT tests for the docs api

ivansenic:
This is a set of test updates for the docs API that were added as part of the #981. We decided to add these to master first, so we can see in the #981 the test diff only for bug fixing.

### DIFF
--- a/testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
+++ b/testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
@@ -1148,70 +1148,6 @@ public class BaseDocumentApiV2Test extends BaseOsgiIntegrationTest {
   }
 
   @Test
-  public void testFullCollectionSearchFilterNesting() throws IOException {
-    JsonNode fullObj1 =
-        OBJECT_MAPPER.readTree("{\"someStuff\": {\"someOtherStuff\": {\"value\": \"a\"}}}");
-    JsonNode fullObj2 = OBJECT_MAPPER.readTree("{\"value\": \"a\"}");
-    RestUtils.put(authToken, collectionPath + "/cool-search-id", fullObj1.toString(), 200);
-    RestUtils.put(authToken, collectionPath + "/cool-search-id-2", fullObj2.toString(), 200);
-
-    // Any filter on full collection search should only match the level of nesting of the where
-    // clause
-    String r =
-        RestUtils.get(
-            authToken,
-            hostWithPort
-                + "/v2/namespaces/"
-                + keyspace
-                + "/collections/collection?page-size=2&where={\"value\": {\"$eq\": \"a\"}}&raw=true",
-            200);
-
-    String expected = "{\"cool-search-id-2\":{\"value\":\"a\"}}";
-    assertThat(OBJECT_MAPPER.readTree(r)).isEqualTo(OBJECT_MAPPER.readTree(expected));
-
-    r =
-        RestUtils.get(
-            authToken,
-            hostWithPort
-                + "/v2/namespaces/"
-                + keyspace
-                + "/collections/collection?page-size=2&where={\"someStuff.someOtherStuff.value\": {\"$eq\": \"a\"}}&raw=true",
-            200);
-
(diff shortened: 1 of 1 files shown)


testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -1147,70 +1147,6 @@ public void testBasicSearchProfile() throws IOException {
     assertThat(OBJECT_MAPPER.readTree(r).get("profile").isEmpty()).isFalse();
   }
 
-  @Test
-  public void testFullCollectionSearchFilterNesting() throws IOException {
```
dimas-b, line 1151: Do we have to remove this now?

dimas-b: Is this test included in the other test methods?

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -1571,6 +1507,394 @@ public void testPaginationSingleDocSearch() throws IOException {
...
+
+    // Any filter on full collection search should only match the level of nesting of the where
+    // clause
+    String r =
+        RestUtils.get(
+            authToken,
+            hostWithPort
+                + "/v2/namespaces/"
+                + keyspace
+                + "/collections/collection?page-size=3&where={\"n,m.value\": {\"$gte\": 5}}&raw=true",
```
dimas-b, line 1705: unrelated to this PR, but IMHO `,` is a stronger separator than `.` so I personally tend to interpret this as `n` or `m.value`... just my 2 cents :)

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -2073,42 +2397,6 @@ public void testInvalidFullDocPageSize() throws IOException {
             "{\"description\":\"The parameter `page-size` is limited to 20.\",\"code\":400}");
   }
 
-  @Test
-  public void createCollection() throws IOException {
```
dimas-b, line 2077: nit: could we re-arrange the code to avoid this showing up as a change? Actual code of `createCollection` appears to be the same :thinking:

dimas-b (approved):

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
ivansenic: I think this is valid as it is per json path spec: https://tools.ietf.org/id/draft-goessner-dispatch-jsonpath-00.html#section-1.3-12

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
ivansenic: I moved it up in the order to separate the search stuff to be at the bottom of the class.. I grouped it in the nested test, but it did not work somehow.. Let's merge as it is..

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
dimas-b: TIL, thx!

dimas-b: JSONPath seems to be different. It requires the alternatives to be wrapped in `[ ]`, does it not? E.g. `["n","m"]["value"]`

ivansenic: OK maybe we need to revisit this, not sure if we have someone in the team to help out here? This would be something related to the JS & JSON path maybe.

dimas-b: @ashryanbeats : Do you have any suggestions on this ^ (the Q is not specific to this PR and we do not have to fix this immediately)

PR #1018, opened 2021-06-16
Allow custom timestamp on operations

tomekl007:
todo:
- [x] update 
- [x] insert


graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/UpdateModelBuilder.java
```diff
@@ -108,6 +110,9 @@ MutationModel build() throws SkipException {
       incrementModels = directives.getIncrementModels();
     }
 
+    // we are using GraphQLString because it will be coerced to BigInteger (Long)
```
dougwettlaufer, line 113: I'm not sure I follow. Why a string instead of something numeric?

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/CqlDirectives.java
```diff
@@ -485,6 +485,7 @@
   public static final String WHERE_OR_IF_OR_INCREMENT_FIELD = "field";
   public static final String WHERE_OR_IF_PREDICATE = "predicate";
   public static final String INCREMENT_PREPEND = "prepend";
+  public static final String TIMESTAMP = "cql_timestamp";
```
dougwettlaufer, line 488: Might be good to be more specific and call this a `WRITE_TIMESTAMP` and `cql_write_timestamp` to avoid confusion later if there's some other timestamp that gets added.

testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/CqlTimestampDirectiveTest.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @BeforeEach
+  public void cleanupData() {
+    SESSION.execute("truncate table \"User\"");
+  }
+
+  @Test
+  @DisplayName("Should update user with write timestamp using @cql_timestamp directive")
+  public void shouldUpdateUserWithWriteTimestampUsingCqlTimestampDirective() {
```
dougwettlaufer, line 79: What about something that doesn't parse? Too big? Too small?

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/CqlDirectives.java
```diff
@@ -587,6 +588,15 @@
           .validLocation(ARGUMENT_DEFINITION)
           .build();
 
+  private static final GraphQLDirective TIMESTAMP_DIRECTIVE =
+      newDirective()
+          .name(TIMESTAMP)
+          .description(
+              "Annotates a parameter to indicate that it will be used as a write timestamp for this row."
+                  + "This is only allowed for insert and update mutations. It must be a long value.")
```
dougwettlaufer, line 596: Just a thought but for convenience what about accepting an ISO 8601 timestamp and converting that to a long?

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/UpdateModelBuilder.java
tomekl007: `BigInteger` is represented as `String` in GraphQL

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/CqlDirectives.java
tomekl007: the naming is unified with other directives:
- `cql_increment`
- `cql_ttl`
- `cql_timestamp`
So I am not sure if that necessary. Do you have a particular use case in mind?
(@olim7t wdyt about making this directive "longer": `cql_write_timestamp`?)

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/CqlDirectives.java
tomekl007: Yeah, we can implement a similar path that Oliver did in https://github.com/stargate/stargate/pull/1022/files#diff-afd1e86c5d7ba50a9f13687223a01d326662d1cfe10432a18e7fd242af038bd0R125-R150.
However, we need to support here: `BigInteger`, `Integer` and I wonder if also adding the `ISO 8601` won't be too wide and confusing. Maybe support for both `BigInteger` and `Integer` is good enough?

testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/CqlTimestampDirectiveTest.java
tomekl007: All values are passed to `USING TIMESTAMP`, so everything that the `USING` timestamp accepts is passed there. As far as I can see, negative values are allowed (added test case for it).
Regarding out-of-range, the `Long.MAX_VALUE` is handled properly.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/UpdateModelBuilder.java
olim7t: To be clear, we allow both `Int` and `String` (it's the user's choice, they define the type of the argument in their schema).

Large integer literals are subject to interoperability issues (the server receives truncated values). This only happens for values larger than (2^53 - 1). Timestamps are in microseconds, so they will reach that limit in the year 2255.

Most applications deal with contemporary timestamps, so they can safely choose `Int`. But if for any reason they were to send timestamps over the limit, then they must use `String`.

All this should be explained in the description of the `@cql_timestamp` directive.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/CqlDirectives.java
olim7t: I think `cql_timestamp` is specific enough. There is no other timestamp that can be sent with a request. As for reading back timestamps, this is done with the `writetime` CQL function; we don't have a way to do that yet, but I think we would do it with a way to provide a formula for an entity field, not via a dedicated directive.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/CqlDirectives.java
olim7t: I know I'm contradicting what I said through another channel, but maybe it's not that complicated after all. If we standardize on the format of [DateTimeFormatter.ISO_ZONED_DATE_TIME](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_ZONED_DATE_TIME), we can do it like this:

```java
  private static long parseTimestamp(String spec) {
    ZonedDateTime dateTime = ZonedDateTime.parse(spec);
    return dateTime.toEpochSecond() * 1_000_000 + dateTime.getNano() / 1000;
  }
```

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/UpdateModelBuilder.java
olim7t: Actually, it should be `BigInt`, not `String`:
```java
    Optional<String> cqlTimestampArgumentName =
        findFieldNameWithDirective(
            CqlDirectives.TIMESTAMP, Arrays.asList(Scalars.GraphQLInt, CqlScalar.BIGINT.getGraphqlType()));
```
That way the coercion is already done automatically for you (instead of doing it manually in `TimestampParser`), and you'll get the value as a `Long` directly in the environment.

If you decide to support ISO strings per the other comment below, then you'll have `GraphQLString` as a third option.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/MutationFetcher.java
```diff
@@ -48,4 +57,28 @@ protected Parameters getDatastoreParameters(DataFetchingEnvironment environment)
...
+            ValueModifier.set(
+                column.getCqlName(), toCqlValue(graphqlValue, column.getCqlType(), keyspace)));
+      }
+    }
+    if (modifiers.isEmpty()) {
+      throw new IllegalArgumentException(
+          "Input object must have at least one non-PK field set for an update");
+    }
+    return modifiers;
+  }
```
olim7t, line 83: This is only called in `UpdateFetcher`. I think you started moving it when we were considering unifying processing between UPDATE and INSERT, but now it's not necessary to move it anymore.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/ArgumentDirectiveModelsBuilder.java
```diff
@@ -110,6 +111,11 @@ private void buildArgument(
       return;
     }
 
+    if (is(directive, CqlDirectives.TIMESTAMP)) {
+      // It's a technical field, ignore it, it will be handled later
+      return;
+    }
+
```
olim7t, line 118: 👍 
Maybe merge the two ifs for paging state and timestamp.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/InsertModelBuilder.java
```diff
@@ -77,11 +82,15 @@ InsertModel build() throws SkipException {
           operationName, entity.getGraphqlName());
     }
 
+    // we are using GraphQLString because it will be coerced to BigInteger (Long)
+    Optional<String> cqlTimestampArgumentName =
+        findFieldNameWithDirective(
+            CqlDirectives.TIMESTAMP, Arrays.asList(Scalars.GraphQLString, Scalars.GraphQLInt));
```
olim7t, line 88: What if there is a second field, but it's not annotated with `@cql_timestamp`?

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/OperationModelBuilderBase.java
```diff
@@ -90,4 +94,66 @@ protected void validateNoFiltering(List<ConditionModel> whereConditions, EntityM
...
+      InputValueDefinition inputValue, String directiveName, List<GraphQLScalarType> expectedTypes)
+      throws SkipException {
+    boolean hasDirective = DirectiveHelper.getDirective(directiveName, inputValue).isPresent();
+    if (!hasDirective) {
+      return false;
+    }
+    Type<?> type = TypeHelper.unwrapNonNull(inputValue.getType());
+    boolean typeMatchFound = false;
+    for (GraphQLScalarType expectedType : expectedTypes) {
+      if (type instanceof TypeName && ((TypeName) type).getName().equals(expectedType.getName())) {
```
olim7t, line 130: You can use `TypeHelper.deepEquals` to shorten this.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/OperationModelBuilderBase.java
```diff
@@ -90,4 +94,66 @@ protected void validateNoFiltering(List<ConditionModel> whereConditions, EntityM
...
+              operationName, directiveName, result.get(), inputValue.getName());
+          throw SkipException.INSTANCE;
+        }
+        result = Optional.of(inputValue.getName());
+      }
+    }
+    return result;
+  }
+
+  private boolean isDirective(
```
olim7t, line 120: `hasDirectiveAndType`?

olim7t (changes requested):

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/UpdateModelBuilder.java
olim7t: One more thing 😬 

I just realized that allowing `GraphQLInt` makes no sense. The highest value it can represent is `Integer.MAX_VALUE` microseconds, or about 35 minutes from the epoch.

So the only type we should allow are:
*  `CqlScalar.BIGINT.getGraphqlType()`
* *if we decide to support ISO strings*, `Scalars.GraphQLString`.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/CqlDirectives.java
tomekl007: For the latter case, even if we would want to use a directive, I don't see a clear problem with using the same directive name as for request.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/MutationFetcher.java
tomekl007: Yes, sorry for that. Reverted. I will do this is a separate PR.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/InsertModelBuilder.java
tomekl007: Currently, it will be ignored (I added a test case for it). I am not sure if that's ideal. We can do 2 things:
1. Take and use the second parameter as the `writeTimestamp` even if it is not annotated with a `@cql_timestamp`. I think it is a bit risky - users may have accidentally updated write. Also, If we need to add another insert directive support, the ordering would be critical, and we don't want to do it.
2. If there is a second argument, but without any directive, we may throw an exception and report all possible directives supported on the insert. This option seems better to me.
wdyt?

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/OperationModelBuilderBase.java
tomekl007: As far as I can see, the `expectedType` works only on the `Type` class. However, here I need to compare the `GraphQLScalarType`. I don't see a clear way to extract the `Type` out of `GraphQLScalarType`. Is it possible?

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/UpdateModelBuilder.java
tomekl007: Regarding using `CqlScalar.BIGINT` instead of `GraphQLString` to carry the value of `BigInt`, I am not sure how it suppose to work automatically. 
It is impossible to create such a GraphQL schema:
```
       + " updateWithWriteTimestamp(\n"
            + "    k: Int\n"
            + "    v: Int\n"
            + "    write_timestamp: BigInt @cql_timestamp\n"
            + "  ): Boolean\n"
```
, it throws:
```
       + " updateWithWriteTimestamp(\n"
            + "    k: Int\n"
            + "    v: Int\n"
            + "    write_timestamp: BigInt @cql_timestamp\n"
            + "  ): Boolean\n"
```
So the `write_timestamp` must be passed as `String`. Even when having this:
```
       findFieldNameWithDirective(
            CqlDirectives.TIMESTAMP,
            Arrays.asList(CqlScalar.BIGINT.getGraphqlType(), Scalars.GraphQLString));
```
The ` environment.getArgument(name);` for the `write_timestamp` returns a `String`  (not `Long`) so there is a need for manual coercion. Am I missing something?

olim7t: I think you meant to paste an error but you pasted the schema a second time.

I tried locally. After passing the right types to `findFieldNameWithDirective` there is indeed an issue: CQL scalars are only added to the schema if they are referenced, and that is currently done only for entity fields. We need to do it for regular arguments too.

I'm pushing a commit (b096de2b) since I have it already.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/InsertModelBuilder.java
olim7t: I agree 2 is better. Since there are only two possible arguments we can use a simple check like `inputs.size()==2 && ˜cqlTimestampArgumentName.isPresent()`.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/OperationModelBuilderBase.java
olim7t: You're right, I was mixing different things: `GraphQLScalarType` is not a subtype of `Type`.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/CqlDirectives.java
```diff
@@ -583,6 +584,15 @@
...
```
olim7t, line 592: Needs an update to mention ISO strings. Also it needs to be more descriptive for people not familiar with CQL and our custom scalars:
```suggestion
                  + "This is only allowed for insert and update mutations. The parameter can be "
                  + "either a `BigInt` (that represents a number of microseconds since the epoch), "
                  + "or a `String` (that represents an ISO-8601 zoned date time, e.g. "
                  + "`2007-12-03T10:15:30+01:00`).")
```

testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/CqlTimestampDirectiveValidationTest.java
```diff
@@ -0,0 +1,205 @@
...
+                    + "    v: Int\n"
+                    + "    write_timestamp: String @cql_timestamp\n"
+                    + "    write_timestamp2: String @cql_timestamp\n"
+                    + "  ): Boolean\n"
+                    + "@cql_update(targetEntity: \"User\")\n"
+                    + "}")
+            .get(0);
+
+    // then
+    AssertionsForClassTypes.assertThat(getMappingErrors(errors))
```
olim7t, line 70: nit: should be `Assertions` and can be statically imported (IDEA does weird things on copy/paste sometimes)

olim7t (approved):
LGTM, pending a small update to the directive description.

PR #1019, opened 2021-06-16
GraphQL CQL-first: support counter increments (fixes #986)

### LINKED ISSUE #986: GraphQL CQL-first: support counter increments
Counter tables only support increment updates, so we don't even need a special syntax for this. The convention could be that if it's a counter table, then every non-PK field is interpreted as an increment:

```
CREATE TABLE counters(k int PRIMARY KEY, c1 counter, c2 counter);

mutation {
  updatecounters(value: { k: 1, c1: 1, c2: -1 }) {
    applied
  }
}
```

olim7t:


dougwettlaufer (approved):

testing/src/main/java/io/stargate/it/http/graphql/cqlfirst/CounterUpdateTest.java
```diff
@@ -0,0 +1,78 @@
...
+  private static CqlFirstClient CLIENT;
+  private static CqlIdentifier KEYSPACE_ID;
+
+  @BeforeAll
+  public static void setup(StargateConnectionInfo cluster, @TestKeyspace CqlIdentifier keyspaceId) {
+    String host = cluster.seedAddress();
+    CLIENT = new CqlFirstClient(host, RestUtils.getAuthToken(host));
+    KEYSPACE_ID = keyspaceId;
+  }
+
```
tomekl007, line 54: We are missing a test case for collections (list)

olim7t: This PR only covers counters.
For collections we'll need additional syntax, because an update can either append or completely replace them.

PR #1020, opened 2021-06-16
Add system property to disable graphql playground

### LINKED ISSUE #791: Add config option to disable graphql playground
Playground is a great tool for local development but not always something you want to expose in production. To allow for running playground locally or in non-prod environments but disabling it for a prod deployment add a new system property that will turn on/off playground in a configurable manner.

dougwettlaufer:
Resolves #791


graphqlapi/src/main/java/io/stargate/graphql/GraphqlActivator.java
```diff
@@ -44,6 +44,8 @@
       System.getProperty("stargate.auth_id", "AuthTableBasedService");
   private static final boolean ENABLE_GRAPHQL_FIRST =
       Boolean.getBoolean("stargate.graphql_first.enabled");
+  private static final boolean DISABLE_GRAPHQL_PLAYGROUND =
```
dimas-b, line 47: nit: would it be worth using `ENABLE` in the name to align with `ENABLE_GRAPHQL_FIRST`?

It would also make the `if` in DropwizardServer.java  a bit more readable IMHO.

dimas-b: `Boolean.parseBoolean(System.getProperty("stargate.graphql_playground.enabled", "true"))`

olim7t (approved):
LGTM, but I agree with Dmitri on inverting the flag.

graphqlapi/src/main/java/io/stargate/graphql/GraphqlActivator.java
dougwettlaufer: I was initially going to use "enable" but it felt weird to have such a flag default to "true". I agree that it makes the `if` more legible but from a user perspective this feels like unexpected behavior. This also aligns with the existing property `stargate.graphql.default_keyspace.disabled`.

dimas-b: The default value logic makes sense to me. Could this be a middle ground if we had something like this:
```
boolean ENABLE_GRAPHQL_PLAYGROUND = !Boolean.getBoolean("stargate.graphql_playground.disabled");
```
... and used positive "enable" parameter names in code? WDYT?

dougwettlaufer: Done

olim7t: > This also aligns with the existing property `stargate.graphql.default_keyspace.disabled`

Hrmm I didn't remember that property existed, that's unfortunate. FWIW I prefer to avoid double negatives ("not disabled"), and it would be slightly easier if you look at a bunch of properties at once and true means enabled all the time.
But maybe I'm overthinking it, +0.

dimas-b (approved):

PR #1022, opened 2021-06-16
GraphQL schema-first: allow custom TTL for INSERT and UPDATE

olim7t:


testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/TtlTest.java
```diff
@@ -0,0 +1,133 @@
...
+    SCHEMA_VERSION =
+        CLIENT.deploySchema(
+            KEYSPACE,
+            "type Foo @cql_input{\n"
+                + "  k: Int @cql_column(partitionKey:true)\n"
+                + "  v: Int\n"
+                + "}\n"
+                + "type Mutation {\n"
+                + "  insertFoo(foo: FooInput): Foo @cql_insert(ttl: \"3600\")\n"
+                + "  updateFoo(k: Int, v: Int): Boolean @cql_update(targetEntity: \"Foo\", ttl: \"PT2H\")\n"
```
olim7t, line 64: I went with static directive arguments. I think passing the TTL as an argument is not a likely use case. However that can always be added later if it gets requested.

dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/MutationModelBuilder.java
```diff
@@ -112,4 +114,39 @@ protected boolean computeIfExists(Optional<Directive> directive) {
...
   }
+
+  protected Optional<Integer> getTtl(Optional<Directive> directive) {
+    return directive.flatMap(
+        d ->
+            DirectiveHelper.getStringArgument(d, CqlDirectives.UPDATE_OR_INSERT_TTL, context)
+                .flatMap(this::parseTtl));
+  }
+
+  private Optional<Integer> parseTtl(String spec) {
```
tomekl007, line 125: I wonder if this does not deserve a unit test. wdyt?

tomekl007:
Tests are failing

Issue #944, opened 2021-05-12
Adding per-keyspace and percentile Dropwizard metrics


marksurnin:
Currently the `http_server_requests_seconds` Dropwizard metrics that are exposed at `:8084/metrics` look like this:

```
http_server_requests_seconds_count{method="POST",module="graphqlapi",status="200",uri="/graphql/{keyspaceName}",} 172.0
http_server_requests_seconds_sum{method="POST",module="graphqlapi",status="200",uri="/graphql/{keyspaceName}",} 5.934070816
http_server_requests_seconds_max{method="POST",module="graphqlapi",status="200",uri="/graphql/{keyspaceName}",} 0.076524805
```

We'd like to include keyspace name here and record these metrics at common percentiles (p95, p99) so we could set up per-keyspace SLOs. Is this something you'd be open to including as an opt-in config setting?

- - -
Here are some details for reference:

1. To get the keyspace name in the metric, we could substitute it in `uri="/graphql/{keyspaceName}` instead of the template or add it as an extra tag.

I briefly tested the latter approach in [ResourceTagsProvider](https://github.com/stargate/stargate/blob/6452957cd078f17f1c0067c943e3fc2129f61ceb/metrics-jersey/src/main/java/io/stargate/metrics/jersey/ResourceTagsProvider.java#L54) and this tag was added to the metric: `keyspaceName="hello_cassandra"`.

```
private Tag convertPathParamToTag(Map.Entry<String, List<String>> pathParamEntry) {
    String values = String.join(",", pathParamEntry.getValue());
    return Tag.of(pathParamEntry.getKey(), values);
}

@Override
public Iterable<Tag> httpRequestTags(RequestEvent event) {
    ...
    List<Tag> pathParamTags = request.getUriInfo().getPathParameters().entrySet().stream()
        .map(this::convertPathParamToTag)
        .collect(Collectors.toList());

    return defaultTags
        .and(pathParamTags)
        ...
```

1. To record percentile metrics, I added a meter filter to the `meterRegistry` config in [MetricsImpl.java](https://github.com/stargate/stargate/blob/6452957cd078f17f1c0067c943e3fc2129f61ceb/core/src/main/java/io/stargate/core/metrics/impl/MetricsImpl.java#L36). This is just a rough version as a PoC:

```
MeterFilter meterFilter = new MeterFilter() {
      @Override
      public DistributionStatisticConfig configure(Meter.Id id, DistributionStatisticConfig config) {
[...]

PR #1035, opened 2021-06-18
relates to #944: property for collecting path params as tags in the h…

ivansenic:
Added the `stargate.metrics.http_server_requests_path_param_tags` property option to declare what path params should be taken into account and use them as the tags in the `http_server_resuests`. Examples:

- ~`stargate.metrics.http_server_requests_path_param_tags=*`~ - can not be supported, produces diff amount of tags every time
- `stargate.metrics.http_server_requests_path_param_tags=keyspaceName,collection` - only path params that are matching `keyspaceName` or `collection`.

**This is a global option and affects all endpoints.**

### DIFF
--- a/metrics-jersey/src/main/java/io/stargate/metrics/jersey/PathParametersTagsProvider.java
+++ b/metrics-jersey/src/main/java/io/stargate/metrics/jersey/PathParametersTagsProvider.java
@@ -19,10 +19,11 @@ package io.stargate.metrics.jersey;
 
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.jersey2.server.JerseyTagsProvider;
+import io.stargate.core.metrics.StargateMetricConstants;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Objects;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import javax.ws.rs.core.MultivaluedMap;
 import org.glassfish.jersey.server.ExtendedUriInfo;
@@ -44,7 +45,7 @@ public class PathParametersTagsProvider implements JerseyTagsProvider {
   /** {@inheritDoc} */
   @Override
   public Iterable<Tag> httpRequestTags(RequestEvent event) {
-    if (config.matchNone) {
+    if (config.whitelistedPathParams.isEmpty()) {
       return Collections.emptyList();
     }
 
@@ -54,7 +55,7 @@ public class PathParametersTagsProvider implements JerseyTagsProvider {
   /** {@inheritDoc} */
   @Override
   public Iterable<Tag> httpLongRequestTags(RequestEvent event) {
-    if (config.matchNone) {
+    if (config.whitelistedPathParams.isEmpty()) {
       return Collections.emptyList();
     }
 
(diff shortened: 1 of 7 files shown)


metrics-jersey/src/main/java/io/stargate/metrics/jersey/PathParametersTagsProvider.java
```diff
@@ -0,0 +1,112 @@
...
+            System.getProperty("stargate.metrics.http_server_requests_path_param_tags");
+        if (null == property || property.length() == 0) {
+          matchNone = true;
+        } else if (Objects.equals(property, "*")) {
+          matchAll = true;
+        } else {
+          matchParams = Arrays.asList(property.split(","));
+        }
+      } catch (Exception e) {
+        matchNone = true;
```
dimas-b, line 106: Please log the exception for trouble-shooting.

metrics-jersey/src/test/java/io/stargate/metrics/jersey/PathParametersTagsProviderTest.java
```diff
@@ -0,0 +1,116 @@
...
+    lenient().when(requestEvent.getUriInfo()).thenReturn(extendedUriInfo);
+    System.clearProperty("stargate.metrics.http_server_requests_path_param_tags");
+  }
+
+  @Nested
+  class HttpRequestTags {
+
+    @Test
+    public void matchSpecific() {
+      System.setProperty("stargate.metrics.http_server_requests_path_param_tags", "k2");
```
dimas-b, line 55: Use the custom `Config` constructor?

metrics-jersey/src/main/java/io/stargate/metrics/jersey/PathParametersTagsProvider.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    public static Config fromSystemProps() {
+      boolean matchAll = false;
+      boolean matchNone = false;
+      Collection<String> matchParams = Collections.emptyList();
+
+      try {
+        String property =
+            System.getProperty("stargate.metrics.http_server_requests_path_param_tags");
```
dimas-b, line 97: Suggestion: make a method parameter for `"stargate.metrics.http_server_requests_path_param_tags"`. This way unit tests can work independently of each other (potentially in parallel).

dimas-b (changes requested):

metrics-jersey/src/main/java/io/stargate/metrics/jersey/PathParametersTagsProvider.java
ivansenic: No exception catching need at all, my bad.

metrics-jersey/src/test/java/io/stargate/metrics/jersey/PathParametersTagsProviderTest.java
ivansenic: Done.

metrics-jersey/src/main/java/io/stargate/metrics/jersey/PathParametersTagsProvider.java
ivansenic: Done.

ivansenic:
Was lucky enough to get INT test to fail due to the different amount of tags in the same process.. Moved to a fixed amount of tags.

dimas-b (approved):

PR #1036, opened 2021-06-18
closes #944: default http tag provider to return selected headers

### LINKED ISSUE #944: Adding per-keyspace and percentile Dropwizard metrics

ivansenic:
Added `stargate.metrics.http_server_requests_header_tags` to select what headers you want added as tags. This is now default impl for the `HttpMetricsTagProvider`.

### DIFF
--- a/core/src/main/java/io/stargate/core/metrics/api/DefaultHttpMetricsTagProvider.java
+++ b/core/src/main/java/io/stargate/core/metrics/api/DefaultHttpMetricsTagProvider.java
@@ -52,7 +52,7 @@ public class DefaultHttpMetricsTagProvider implements HttpMetricsTagProvider {
       Tag[] collect =
           headers.entrySet().stream()
               .filter(e -> whitelist.contains(e.getKey().toLowerCase()))
-              .map(e -> Tag.of(e.getKey(), String.join(",", e.getValue())))
+              .map(e -> Tag.of(e.getKey().toLowerCase(), String.join(",", e.getValue())))
               .toArray(Tag[]::new);
       return Tags.of(collect);
     }
@@ -63,16 +63,24 @@ public class DefaultHttpMetricsTagProvider implements HttpMetricsTagProvider {
     private final Collection<String> whitelistedHeaderNames;
 
     public Config(Collection<String> whitelistedHeaderNames) {
-      this.whitelistedHeaderNames = whitelistedHeaderNames;
+      if (null == whitelistedHeaderNames) {
+        this.whitelistedHeaderNames = Collections.emptyList();
+      } else {
+        this.whitelistedHeaderNames = whitelistedHeaderNames;
+      }
     }
 
     public static Config fromSystemProps() {
-      try {
         String property = System.getProperty("stargate.metrics.http_server_requests_header_tags");
-        if (null != property) {
-          String[] headers = property.split(",");
+        return fromPropertyString(property);
+    }
+
+    public static Config fromPropertyString(String value) {
+      try {
+        if (null != value) {
+          String[] headers = value.split(",");
           List<String> lowercaseHeaders =
-              Arrays.stream(headers).map(String::toLowerCase).collect(Collectors.toList());
+                  Arrays.stream(headers).map(String::toLowerCase).collect(Collectors.toList());
           return new Config(lowercaseHeaders);
         } else {
           return new Config(Collections.emptyList());
--- a/core/src/main/java/io/stargate/core/metrics/api/DefaultHttpMetricsTagProvider.java
+++ b/core/src/main/java/io/stargate/core/metrics/api/DefaultHttpMetricsTagProvider.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.stargate.core.metrics.api;
+
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * Default {@link HttpMetricsTagProvider} that adds headers as tags based on the whitelisted header
+ * names from the system prop.
+ */
+public class DefaultHttpMetricsTagProvider implements HttpMetricsTagProvider {
+
+  private final Config config;
+
+  public DefaultHttpMetricsTagProvider() {
+    this(Config.fromSystemProps());
+  }
+
+  public DefaultHttpMetricsTagProvider(Config config) {
+    this.config = config;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public Tags getRequestTags(Map<String, List<String>> headers) {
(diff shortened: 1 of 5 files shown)


core/src/test/java/io/stargate/core/metrics/api/DefaultHttpMetricsTagProviderTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+      DefaultHttpMetricsTagProvider provider = new DefaultHttpMetricsTagProvider();
+      Tags result = provider.getRequestTags(headers);
+
+      assertThat(result).containsOnly(Tag.of("header1", "value1"));
+    }
+
+    @Test
+    public void caseIrrelevant() {
+      System.setProperty("stargate.metrics.http_server_requests_header_tags", "HEADER1,header2");
```
dimas-b, line 60: Shall we use a custom instance of the `Config` class for this? Was that why the `Config(Collection<String>)` constructor was introduced?

core/src/main/java/io/stargate/core/metrics/api/DefaultHttpMetricsTagProvider.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private final Collection<String> whitelistedHeaderNames;
+
+    public Config(Collection<String> whitelistedHeaderNames) {
+      this.whitelistedHeaderNames = whitelistedHeaderNames;
+    }
+
+    public static Config fromSystemProps() {
+      try {
+        String property = System.getProperty("stargate.metrics.http_server_requests_header_tags");
```
dimas-b, line 71: Suggestion: make `"stargate.metrics.http_server_requests_header_tags"` a parameter to this method. This way unit tests can work independently of each other (potentially in parallel).

dimas-b (changes requested):

core/src/main/java/io/stargate/core/metrics/api/DefaultHttpMetricsTagProvider.java
```diff
@@ -0,0 +1,85 @@
...
+  @Override
+  public Tags getRequestTags(Map<String, List<String>> headers) {
+    Collection<String> whitelist = config.whitelistedHeaderNames;
+    if (null == whitelist || whitelist.isEmpty()) {
+      return Tags.empty();
+    } else {
+      Tag[] collect =
+          headers.entrySet().stream()
+              .filter(e -> whitelist.contains(e.getKey().toLowerCase()))
+              .map(e -> Tag.of(e.getKey(), String.join(",", e.getValue())))
```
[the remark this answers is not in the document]
ivansenic: Good point, I wanted to leave the headers as is in tags, but completely forgot that it could cause double tag keys for different header cases.. Thanks :+1:

core/src/main/java/io/stargate/core/metrics/api/DefaultHttpMetricsTagProvider.java
ivansenic: Added extra method..

core/src/test/java/io/stargate/core/metrics/api/DefaultHttpMetricsTagProviderTest.java
ivansenic: Using added method

core/src/main/java/io/stargate/core/metrics/api/DefaultHttpMetricsTagProvider.java
```diff
@@ -0,0 +1,85 @@
...
+
+  public DefaultHttpMetricsTagProvider(Config config) {
+    this.config = config;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public Tags getRequestTags(Map<String, List<String>> headers) {
+    Collection<String> whitelist = config.whitelistedHeaderNames;
+    if (null == whitelist || whitelist.isEmpty()) {
```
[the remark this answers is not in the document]
ivansenic: Fix this as requested. But you do know that branch prediction would kick in here, as the whitelist value is not changing, there is no overhead of doing any check (not sure if this was the problem)?

ivansenic:
I made a huge mistake by not returning always the same amount of `Tags`. Fortunately caught by the INT test in #1035. Funny that I was aware of this, and then tried in a test with `SimpleMeterRegistry`, which is different from the Prom one. Fixed now. 

See https://github.com/micrometer-metrics/micrometer/issues/877 for details.

core/src/main/java/io/stargate/core/metrics/api/DefaultHttpMetricsTagProvider.java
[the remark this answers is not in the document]
dimas-b: No, perf. is not a concern, just easier to read :)

core/src/main/java/io/stargate/core/metrics/api/DefaultHttpMetricsTagProvider.java
```diff
@@ -0,0 +1,103 @@
...
+      return Tags.empty();
+    } else {
+      // note that we need to return same set of tags
+      // this is why we iterate white list and try to find header value
```
dimas-b, line 55: +1

dimas-b (approved):

PR #1037, opened 2021-06-18
closes #947: document api validation finalized

### LINKED ISSUE #947: Improve Document API parameter validation

ivansenic:
Done with the docs api validation.. Open questions:

* Can it be that the default value for the full search is `1` and for the in document search is `100`?
* Is it ok not to have a `max` limit for the in document search?

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
@@ -29,12 +29,16 @@ import java.net.URI;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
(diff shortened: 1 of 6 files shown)


dimas-b:
> Can it be that the default value for the full search is 1 and for the in document search is 100?

Sure. Why not?

In-document search is limited to one partition, so the amount of data is already somewhat restricted.

For global search, I imagine the default page size of `1` is there to avoid long scans if the `where` clause happens to be very narrow.

dimas-b (approved):

dimas-b:
> Is it ok not to have a max limit for the in document search?

I think it's ok. We already allow loading the doc as whole even if it is huge. A search within a doc cannot yield more data that the whole doc.

PR #1038, opened 2021-06-18
allow insert to return Boolean to have the unified API between insert and update

tomekl007:
After more investigation of the data retrieval code for Update and Insert fetchers, I think that the logic should stay separate. There are more differences between those code paths. However, I think that we should allow inserting to support return `Boolean` to have the unified API. I've implemented it in this PR.

### DIFF
--- a/testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/InsertTest.java
+++ b/testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/InsertTest.java
@@ -56,6 +56,7 @@ public class InsertTest extends GraphqlFirstTestBase {
             + "}\n"
             + "type Mutation {\n"
             + "  insertUser(user: UserInput!): User\n"
+            + "  insertUserReturnBoolean(user: UserInput!): Boolean\n"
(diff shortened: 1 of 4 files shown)


testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/InsertTest.java
```diff
@@ -69,6 +70,21 @@ public void testSimpleInsert() {
...
+  public void testSimpleInsertReturnBoolean() {
+    Object response =
+        CLIENT.executeKeyspaceQuery(
+            KEYSPACE,
+            "mutation {\n"
+                + "  result: insertUserReturnBoolean(user: {name: \"Ada Lovelace\", username: \"@ada\"})\n"
+                + "}");
+
+    // Should have generated an id
+    Boolean applied = JsonPath.<Boolean>read(response, "$.result");
```
ivansenic, line 84: Awesome, I like this `JsonPath` helper, way more easier to assert the results. We should use this everywhere..

ivansenic (approved):

PR #1041, opened 2021-06-18
Use URL-safe base64 alphabet for REST API paging state

### LINKED ISSUE #1039: Incorrect pageState when get subdocument (AstraDb)
I have a document
`{
        "careerStartDate": "2012-04-23T18:25:43.511Z",
        "certified": true,
        "description": "Мы уже более 5 лет занимаемся лидогенерацией и продажами в ВК.<br>Легко настроим таргетированную рекламу и возьмем вашу группу на комплексное ведение и продвижение.",
        "name": "Сергей Смирнов",
        "reviews": {
            "review1": {
                "authorName": "Артем",
                "comment": "Пользуюсь услугами этой компании уже около года. Группа растет, клиенты звонят. Ребята всегда на связи. Рекомендую к сотрудничеству.",
                "rating": 5,
                "status": "moderation"
            },
            "review2": {
                "authorName": "Артем",
                "comment": "Пользуюсь услугами этой компании уже около года. Группа растет, клиенты звонят. Ребята всегда на связи. Рекомендую к сотрудничеству.",
                "rating": 4,
                "status": "resolved"
            },
            "review3": {
                "authorName": "Артем",
                "comment": "Пользуюсь услугами этой компании уже около года. Группа растет, клиенты звонят. Ребята всегда на связи. Рекомендую к сотрудничеству.",
                "rating": 3,
                "status": "rejected"
            },
            "review4": {
                "authorName": "Артем",
                "comment": "Пользуюсь услугами этой компании уже около года. Группа растет, клиенты звонят. Ребята всегда на связи. Рекомендую к сотрудничеству.",
                "rating": 4,
                "status": "resolved"
            },
            "review5": {
                "authorName": "Артем",
                "comment": "Пользуюсь услугами этой компании уже около года. Группа растет, клиенты звонят. Ребята всегда на связи. Рекомендую к сотрудничеству.",
                "rating": 4,
                "status": "resolved"
            },
            "review6": {
                "authorName": "Артем",
                "comment": "Пользуюсь услугами этой компании уже около года. Группа растет, клиенты звонят. Ребята всегда на связи. Рекомендую к сотрудничеству.",
                "rating": 4,
                "status": "resolved"
            },
            "review7": {
                "authorName": "Артем",
                "comment": "Пользуюсь услугами этой компании уже около года. Группа растет, клиенты звонят. Ребята всегда на связи. Рекомендую к сотрудничеству.",
                "rating": 4,
                "status": "resolved"
            }
        },
        "serviceUsingStartDate": "2012-04-23T18:25:43.511Z",
        "services": [
            "Создание чат-ботов",
            "Консультирование",
            "Настройка таргета ВК"
        ],
        "specialty": "Основатель SMMZAVOD"
    }
`

When i try load "reviews" with pagination `.../path/to/reviews/*?where={...}&page-size=3`,
I get pageState `JGRlNDgxMjk3LTBiN2QtNDMwZi04ZWRmLTAyOWE5NjhjZThkZST/VVVVVVVVVUAHcmV2aWV3cwdyZXZpZXc1BnN0YXR1c+FVVVXwf/////B/////`

When i paste this page-state to next request, i gets response with error:
`{
    "description": "Server error: Illegal base64 character 20",
    "code": 500
}`

dimas-b:
Fixes #1039


restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentSearchPageState.java
```diff
@@ -1,65 +0,0 @@
-package io.stargate.web.docsapi.dao;
-
-import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.annotation.JsonProperty;
-import java.nio.ByteBuffer;
-import java.util.Base64;
-
-public class DocumentSearchPageState {
```
dimas-b, line 9: This class was not used. I noticed it now because it used `Base64` (incorrectly).

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
```diff
@@ -24,11 +39,19 @@ public Paginator(String pageState, int pageSize) {
...
 
     if (pageState != null) {
-      byte[] decodedBytes = Base64.getDecoder().decode(pageState);
+      byte[] decodedBytes = decode(pageState);
       this.currentPageState = ByteBuffer.wrap(decodedBytes);
     }
   }
 
+  public static byte[] decode(String pageState) {
+    return Base64.getUrlDecoder().decode(pageState);
```
dougwettlaufer, line 48: Do we need to temporarily decode both encodings? Wondering if when this rolls out it could break inflight requests.

dougwettlaufer (approved):
Looks good. I don't want to expand scope so would you mind opening an issue to add this to the rest side?

restapi/src/main/java/io/stargate/web/docsapi/dao/Paginator.java
dimas-b: Good point.

An old paging state with a `+` would be corrupted at the HTTP level, it reaches our code already broken.

However, old paging state that only contain `/` chars (the seconds and last difference for the [URL-safe scheme](https://www.ietf.org/rfc/rfc4648.html#section-5)) should be reusable. I'll update.

dimas-b: I think we can un-break old paging state too... :hourglass:

dimas-b: I think I'll include plain REST in this PR too... the problem is the same in both code areas.

dimas-b: updated, force pushed

core/src/main/java/io/stargate/core/util/ByteBufferUtils.java
```diff
@@ -31,6 +31,22 @@ public static ByteBuffer fromBase64(String base64) {
     return ByteBuffer.wrap(Base64.getDecoder().decode(base64));
   }
 
+  public static String toBase64ForUrl(ByteBuffer buffer) {
+    return Base64.getUrlEncoder().encodeToString(getArray(buffer));
+  }
+
+  public static ByteBuffer fromBase64UrlParam(String base64) {
```
ivansenic, line 38: Few questions here:

1. Is this only to be backwards-compatible with page state already returned? 
2. So how do you know what's the diff between the base encoder and the url encoder? Why searching only for these char and why this single replace?

Going a bit deeper and looking into this (https://datatracker.ietf.org/doc/html/rfc4648#section-4), seems that the diff in encoding is not what you have here? 

```
This encoding is technically identical to the previous one, except
   for the 62:nd and 63:rd alphabet character, as indicated in Table 2.
```   

Maybe I am wrong, but let's double check on this.

ivansenic (changes requested):
Blocking until we double check the diff in encodings.

core/src/main/java/io/stargate/core/util/ByteBufferUtils.java
ivansenic: Yea so `/` should also be checked for I guess, as it can be returned in the base encoding, same as the `+`. Url encoded value is `%2F`.

dimas-b: > Is this only to be backwards-compatible with page state already returned?

backward-compatibility is handled inside the `if` on line 40.

Line 47 is the regular use case (transparent to URL encoding).

> So how do you know what's the diff between the base encoder and the url encoder? Why searching only for these char and why this single replace?

Replacing ` ` (space) chars is to fix older base64 text blocks that get broken at the HTTP level where `+` gets replaces with ` ` during URL parameter decoding. Since the base64 alphabet does not use spaces, this transformation should be safe as far as I can tell.

If neither `/` nor `+` are in the base64 encoded string, then the "URL decoder" should be able to decode the string correctly, I believe, even if it was produced by the "plain" encoder.

Do you see a particular failure mode that is not handled by this code?

ivansenic (approved):

core/src/main/java/io/stargate/core/util/ByteBufferUtils.java
ivansenic: My bad, resolving.

dougwettlaufer (approved):

PR #1043, opened 2021-06-21
Add endpoint to write multiple docs

EricBorczuk:
Add an endpoint to allow writing multiple documents in a single request.

Data sent to this endpoint is expected to be in JSON lines format (1 document per line).

Additionally, you can supply an `id-path` query parameter to use the value at a particular path in each document as the document's key in the database, so if all your documents have an `id` field, you could set `id-path=id` and treat the value in `id` as the document's key. You can also use any valid path syntax (globs not allowed), e.g. `id-path=user.emails.[0].id`

If `id-path` is excluded, random UUID's will be assigned to every document, and the response will have the ID's created corresponding in the same order as the documents were supplied in.


restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -173,6 +175,72 @@ public Response postDoc(
...
+          String authToken,
+      @ApiParam(value = "the namespace that the collection is in", required = true)
+          @PathParam("namespace-id")
+          String namespace,
+      @ApiParam(value = "the name of the collection", required = true) @PathParam("collection-id")
+          String collection,
+      @ApiParam(
+              value = "A JSON Lines payload where each line is a document to write",
+              required = true)
+          InputStream payload,
```
ivansenic, line 213: can we add a `@NonNull` now when we have everything validated..

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -390,6 +391,110 @@ private Object convertToBackendBooleanValue(boolean value, boolean numericBoolea
...
+      InputStream payload,
+      Optional<String> idPath,
+      Db dbFactory,
+      Map<String, String> headers)
+      throws IOException, UnauthorizedException {
+
+    DocumentDB db = dbFactory.getDocDataStoreForToken(authToken, headers);
+    JsonSurfer surfer = JsonSurferGson.INSTANCE;
+
+    boolean created = db.maybeCreateTable(keyspace, collection);
```
ivansenic, line 411: can we extract this to some private method, isn't there some other places that uses the same as well? what about inserting the single document?

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -390,6 +391,110 @@ private Object convertToBackendBooleanValue(boolean value, boolean numericBoolea
...
+      while (iter.hasNext()) {
+        chunkIndex++;
+        Map<String, String> docsInChunk = new LinkedHashMap<>();
+        while (docsInChunk.size() < CHUNK_SIZE && iter.hasNext()) {
+          String doc = iter.next();
+          JsonNode json = mapper.readTree(doc);
+
+          String docId = UUID.randomUUID().toString();
+          if (idPath.isPresent()) {
+            String docsPath = convertToJsonPtr(idPath.get());
```
ivansenic, line 433: so everything related to the `idPath` conversion to the doc path can be moved out of the loop, this way you are not doing the same thing over and over again as this seems to be a constant..

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -692,6 +695,63 @@ public void testPostProfile() throws IOException {
     assertThat(OBJECT_MAPPER.readTree(resp).get("profile").isEmpty()).isFalse();
   }
 
+  @Test
+  public void testWriteManyDocs() throws IOException {
```
ivansenic, line 699: can we also add one test that targets the overwrite?

ivansenic (changes requested):
Few small fixes needed imo.. :+1:

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -390,6 +391,122 @@ private Object convertToBackendBooleanValue(boolean value, boolean numericBoolea
...
+      ExecutionContext context = ExecutionContext.NOOP_CONTEXT;
+      String docsPath = convertToJsonPtr(idPath.get());
+      int chunkIndex = 0;
+      final int CHUNK_SIZE = 250;
+      while (iter.hasNext()) {
+        chunkIndex++;
+        Map<String, String> docsInChunk = new LinkedHashMap<>();
+        while (docsInChunk.size() < CHUNK_SIZE && iter.hasNext()) {
+          String doc = iter.next();
+          JsonNode json = mapper.readTree(doc);
```
dimas-b, line 442: Why do we require one doc per line? Is it not possible to parse JSON input in a stream fashion?

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -390,6 +391,122 @@ private Object convertToBackendBooleanValue(boolean value, boolean numericBoolea
...
+
+          String docId = UUID.randomUUID().toString();
```
dimas-b, line 444: The generated ID is ignored if we have an `idPath`, but generating it takes time and drains randomness pools.

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -390,6 +391,122 @@ private Object convertToBackendBooleanValue(boolean value, boolean numericBoolea
...
+    try (BufferedReader reader = new BufferedReader(new InputStreamReader(payload, "UTF-8"))) {
+      Iterator<String> iter = reader.lines().iterator();
+      ExecutionContext context = ExecutionContext.NOOP_CONTEXT;
+      String docsPath = convertToJsonPtr(idPath.get());
+      int chunkIndex = 0;
+      final int CHUNK_SIZE = 250;
+      while (iter.hasNext()) {
+        chunkIndex++;
+        Map<String, String> docsInChunk = new LinkedHashMap<>();
+        while (docsInChunk.size() < CHUNK_SIZE && iter.hasNext()) {
```
dimas-b, line 440: Do I read it correctly that `CHUNK_SIZE` corresponds to the number of docs in each CQL batch? Why would be break input into multiple batches?

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -390,6 +391,122 @@ private Object convertToBackendBooleanValue(boolean value, boolean numericBoolea
...
+        long now = timeSource.currentTimeMicros();
+        try {
+          db.deleteManyThenInsertBatch(
+              keyspace,
+              collection,
+              ids,
+              bindVariableList,
+              Collections.emptyList(),
+              now,
+              context.nested("ASYNC INSERT"));
```
dimas-b, line 493: Could you add a profiling unit test for this, please?

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: Technically, the input isn't JSON, since it's actually multiple delimited JSON objects (JSON lines format). I'll look into possible methods to stream JSON lines, but Jackson doesn't have anything for JSON lines afaik

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: I really don't have a reason for this; my only question being: if there are some large number of documents that someone sends, is it fine to put all of those documents in a single batch? is there a point where that would break?

If there isn't really a feasible breaking point, then I can just remove this

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: Another option would be to make the input valid JSON (i.e. require that it's a JSON array) and then streaming probably becomes possible/easy

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
dimas-b: Storage may have limits on max batch size, but I do not think we have visibility into that at SG level.

Since we do not really limit HTTP payload sizes for single doc inserts (do we?) it does not make batch operations worse in terms of overloading the server. It may be worth adding guardrails at HTTP level, but that's a separate concern for sure.

I believe it would not more intuitive for the end user if we had one batch per request.

If it fails at storage level, the user will have to consider using smaller batches, but at least it will be consistent with other APIs for that matter. WDYT?

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
dimas-b: Yeah, I guess it may depend on how the library works with input streams. Please post your findings :)

dimas-b: Requiring input to be a JSON array is not unreasonable from my POV.

gconaty: Perhaps an option to import and/or export in either format?

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -390,6 +391,122 @@ private Object convertToBackendBooleanValue(boolean value, boolean numericBoolea
...
+        } catch (Exception e) {
+          throw new ErrorCodeRuntimeException(
+              ErrorCode.DOCS_API_WRITE_BATCH_FAILED,
```
dimas-b, line 496: We do not wait for the `CompletableFuture` associated with the batch operation (or did I miss that), so this catch will not reflect storage-level failure. Would it make sense to wait for batch acknowledgement so that a successful HTTP response would mean the batch was accepted (we should probably use the `202` response code in that case)? WDYT?

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -173,6 +176,73 @@ public Response postDoc(
...
+              documentService.writeManyDocs(
+                  authToken,
+                  namespace,
+                  collection,
+                  payload,
+                  Optional.ofNullable(idPath),
+                  dbFactory,
+                  getAllHeaders(request));
+
+          return Response.created(
```
dimas-b, line 238: WDYT about `202` (Accepted) instead? Currently our code does not guarantee that anything was created at this point or will eventually be created (see my other comment about exceptions).

Even if we wait for the ack from storage, batches may or may not be applied by the time we get the ack, so `201` (Created) might be an overstatement. WDYT?

dougwettlaufer: +1 for `202` since we're just "accepting" here

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: I think that's reasonable, yea - especially the point you made about how stargate doesn't have visibility into any batch size limits in storage.  Also, it's true that the single document write doesn't have a strict limit. So I will remove this!

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -173,6 +176,73 @@ public Response postDoc(
         });
   }
 
+  @POST
+  @ManagedAsync
+  @ApiOperation(
+      value = "Write multiple documents in one request",
+      notes =
+          "Auto-generates an ID for the newly created document if an idPath is not provided as a query parameter. When an idPath is provided, this operation is idempotent.",
```
dougwettlaufer, line 184: > this operation is idempotent

As in a NOOP or it's an upsert?

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
dougwettlaufer: What's the error response from dropwizard look like for that annotation?

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
EricBorczuk: It will be an upsert - if you provide an `idPath` and do two back-to-back requests, the second request will update instead of inserting

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: It seems like `deleteManyThenInsertBatch` calls `executeBatch`, which then calls

``` 
if (useLoggedBatches) {
      dataStore.batch(queries, ConsistencyLevel.LOCAL_QUORUM).join();
} else {
      dataStore.unloggedBatch(queries, ConsistencyLevel.LOCAL_QUORUM).join();
}
```

This would mean that it is already blocking on some sort of ack, right? If the ack may not mean that data was written, then I agree with changing this to 202...but most probably that should be changed in the other write endpoints too, which I would consider out of scope for this PR; so I think I will leave as 201 just for now, to remain consistent with the other document write endpoints

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: So using a JsonParser (Jackson's impl for streaming JSON) is feasible, but only with valid JSON, so I am going to stick with expecting array data over the wire; then, each object in the array can get parsed one by one using an ObjectMapper, which won't pull in much into memory at once :+1: impl is [here](https://github.com/stargate/stargate/pull/1043/files#diff-caf69767609da0fee30d28b2b410ba5e400ed01e1f42af082f495ef82fec8416R435)

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: removed!

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: added [here](https://github.com/stargate/stargate/pull/1043/files#diff-93b9952f26f7639ef23892552f8d4059d445215cdf6bef3ed4eaacc93357a399R865)

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
dimas-b: Ah, right, I missed that :facepalm:

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -390,6 +393,111 @@ private Object convertToBackendBooleanValue(boolean value, boolean numericBoolea
...
+                    json, idPath.get(), json.at(docsPath.get()).toString()));
+          }
+          docId = json.requiredAt(docsPath.get()).asText();
+        } else {
+          docId = UUID.randomUUID().toString();
+        }
+        docs.put(docId, json.toString());
+      }
+
+      // Write the chunk of (at most) 250 documents by firing a single batch with the row inserts
```
dimas-b, line 462: I thought we removed "chunks" :thinking: is this a leftover comment?

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
dimas-b: Ok, let's switch to `202` :)

dimas-b (changes requested):
LGTM overall :+1: Just some minor, but critical tweaks remain to be done, I think.

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -173,6 +176,73 @@ public Response postDoc(
...
+      code = 201)
+  @ApiResponses(
+      value = {
+        @ApiResponse(code = 201, message = "Created", response = WriteDocResponse.class),
```
dougwettlaufer, line 188:
```suggestion
        @ApiResponse(code = 202, message = "Accepted", response = MultiDocsResponse.class),
```

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -173,6 +176,73 @@ public Response postDoc(
         });
   }
 
+  @POST
+  @ManagedAsync
+  @ApiOperation(
+      value = "Write multiple documents in one request",
+      notes =
+          "Auto-generates an ID for the newly created document if an idPath is not provided as a query parameter. When an idPath is provided, this operation is idempotent.",
+      code = 201)
```
dougwettlaufer, line 185:
```suggestion
      code = 202)
```

dougwettlaufer:
@EricBorczuk would you mind writing up a short example and filing an issue in stargate/docs so we don't forget to document this?

dougwettlaufer (approved):

EricBorczuk:
Done https://github.com/stargate/docs/issues/132

dimas-b (approved):

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
ivansenic: it's generated in `ViolationExceptionMapper`, but it requires a message, please do:

`@NotNull(message = "payload must not be null")`

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -397,6 +404,110 @@ private Object convertToBackendBooleanValue(boolean value, boolean numericBoolea
...
+
+      Map<String, String> docs = new LinkedHashMap<>();
+      if (jsonParser.nextToken() != JsonToken.START_ARRAY) {
+        throw new IllegalArgumentException("Payload must be an array.");
+      }
+
+      while (jsonParser.nextToken() != JsonToken.END_ARRAY) {
+        JsonNode json = mapper.readTree(jsonParser);
+        String docId;
+        if (idPath.isPresent()) {
```
ivansenic, line 457: shouldn't this be `docsPath.isPresent()`?

ivansenic (approved):
I think my comments are resolved now..

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: technically it is logically equivalent, but yea that would make way more sense to read :) I'll change

clun: Wondering if we could hit a `JsonToken.END_ARRAY` inside a document itself and loosing track witch the streaming parser here?

```json
[
 { "id":1, "sample":[1,2,3] },
 { "id":2, "sample":[4,5,6] }
]
```

EricBorczuk: I've tested this out, and the parser actually works properly! This is because mapper.readTree(...) on the parser reads the next full object.

EricBorczuk: See https://github.com/stargate/stargate/pull/1043/files#diff-585290b1d8a4cd9f6b594dd2e6a35102c4ab7c86988b9df67452177d38c73868R1, which is used in a few tests

PR #1047, opened 2021-06-22
relates to #1024: refactored doc search rows to node conversion

### LINKED ISSUE #1024: Document population and full search could be restricted to the field paths
Currently when the documents are populated in the docs api, the query is fetching all the rows of that document. This could be improved if the user specifies the fields to be returned:

* fetching only the columns in needed max depth (max field depth + 1)
* limit the rows by actually setting shared pre-path between all rows

Example: If the user requests the fields `['a.b', 'a.c.d']`, the query should fetch only `p0 = a`, `p1 > ''` and make `p3 = ''`.

This also requires that `JsonConverter` supports population based on the requested max depth.

ivansenic:
Before being able to optimize on the query level, we had to do small refactoring and bug fixing:

**New features:**
* supports globs and paths segments in fields in the same way as in the where - no difference now. This all works -> `fields=["field.*.value","field.one,two,three.value","field.[*].value","field.[2].value","field.[0],[1],[2].value"]`

**Fixed bugs:**
* fail on wrong array index was not working - `[not_allowed]` would pass :disappointed: 
* improved matching on globs
   * `{ field: value}` would pass a condition and selection on `[*].value`
   * `[{ field: value}]` would pass a condition and selection on `*.value`

**Tasks done:**
* [x] moved code related to the search to `ReactiveDocumentService`
* [x] extracted loading values from rows to a utils class 
* [x] improved tests
* [x] break connection from  `ReactiveDocumentService` to `DocumentService`, move `convertToSelectionList` to utils
* [x] move `DocumentServiceUtils` to proper package and add tests for every method
* [x] check where `DocumentServiceUtils` can be used more to decrease code in `DocumentService`
* [x] blocked by #1050

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
@@ -27,16 +27,20 @@ import io.reactivex.rxjava3.core.Single;
 import io.stargate.auth.AuthenticationSubject;
 import io.stargate.auth.AuthorizationService;
 import io.stargate.auth.SourceAPI;
+import io.stargate.db.datastore.Row;
 import io.stargate.web.docsapi.dao.DocumentDB;
 import io.stargate.web.docsapi.dao.Paginator;
 import io.stargate.web.docsapi.exception.ErrorCode;
 import io.stargate.web.docsapi.exception.ErrorCodeRuntimeException;
 import io.stargate.web.docsapi.models.DocumentResponseWrapper;
 import io.stargate.web.docsapi.service.query.DocumentSearchService;
+import io.stargate.web.docsapi.service.query.DocumentServiceUtils;
 import io.stargate.web.docsapi.service.query.ExpressionParser;
 import io.stargate.web.docsapi.service.query.FilterExpression;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
 import javax.inject.Inject;
 
 public class ReactiveDocumentService {
@@ -44,6 +48,7 @@ public class ReactiveDocumentService {
   @Inject ExpressionParser expressionParser;
   @Inject DocumentSearchService searchService;
   @Inject DocumentService documentService;
+  @Inject JsonConverter jsonConverter;
   @Inject ObjectMapper objectMapper;
 
   public ReactiveDocumentService() {}
@@ -52,10 +57,12 @@ public class ReactiveDocumentService {
       ExpressionParser expressionParser,
       DocumentSearchService searchService,
       DocumentService documentService,
+      JsonConverter jsonConverter,
       ObjectMapper objectMapper) {
     this.expressionParser = expressionParser;
     this.searchService = searchService;
     this.documentService = documentService;
+    this.jsonConverter = jsonConverter;
     this.objectMapper = objectMapper;
   }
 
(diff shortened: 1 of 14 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/query/DocumentServiceUtils.java
```diff
@@ -78,7 +79,52 @@ public static long maxFieldDepth(String field) {
...
+    return fields.stream().mapToInt(DocumentServiceUtils::maxFieldDepth).max();
+  }
+
+  /**
+   * Simple utility to return a string value from doc api {@link Row}.
+   *
+   * @param row Row
+   * @return String or null if a row column {@value QueryConstants#STRING_VALUE_COLUMN_NAME) is null.
+   */
+  public static String getStringFromRow(Row row) {
```
dimas-b, line 92: nit: rename to `getString`? `Row` is obvious from the parameter type.... however, it's a matter of personal preference, I do not insist :)

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -138,4 +137,44 @@ public ReactiveDocumentService(
...
+  private String getFieldPathFromRow(Row row, long maxDepth) {
+    List<String> path = new ArrayList<>();
+    for (int i = 0; i < maxDepth; i++) {
+      String value = row.getString("p" + i);
+      if (value == null || value.isEmpty()) {
+        break;
+      }
+      path.add(value);
+    }
+    return String.join(".", path);
```
dimas-b, line 178: Each request has a few fields, but may have many rows. Could we refactor the filtering logic to use a list of individual field names (at each level) and compare them to `pNNN` values to avoid constructing transient `String`s?

dimas-b (commented):
If fields have a common prefix, I believe we can still optimize the "populate" query by restricting `pNNN` columns to the common prefix.

Also, if the doc is large, but few fields are selected it may be more efficient to run two or more "populate" query for each common field prefix than fetching the whole doc and filtering in memory. Since we do not have stats ATM, we may need a user-level parameter to control whether to use this strategy or not. WDYT?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
ivansenic: Sure we can, I already optimized here, but due to some reverts this was gone.. Good catch :+1:

ivansenic:
@EricBorczuk @dimas-b Please check the updated description of the PR.. This was all needed to actually go for the query optimization.

dimas-b (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  /**
+   * Simple utility to return a string value from doc api {@link Row}.
+   *
+   * @param row Row
+   * @return String or null if a row column {@value QueryConstants#STRING_VALUE_COLUMN_NAME) is null.
+   */
+  public static String getStringFromRow(Row row) {
+    return row.isNull(QueryConstants.STRING_VALUE_COLUMN_NAME)
```
mpenick, line 134: It might be better to avoid the double hash lookup e.g.
```
int index = row.firstIndexOf(QueryConstants.STRING_VALUE_COLUMN_NAME);
return row.isNull(index) ? null : row.getString(index);
```

PR #1050, opened 2021-06-23
fixes #1049: path segment search with in-memory filter fixed

### LINKED ISSUE #1049: Document search with path segment and $in returns wrong results
Document search with `where={"n,m.value": {"$in": [5]}}` will match both `{"n": { "value": 5}}` and `{"m": { "value": 8}}`. Path segment is not considered when doing the row matches filter path check.

ivansenic:
See #1049 for details. Two fixes:

- `convertArrayPath` to transform also the path segments and convert directly to expected path
   - this enables complete removal of calls to the `convertArrayPath` outside the expression parsing
- filter expression to consider path segments when evaluating

Not sure if this was working before, there were no tests.

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/query/DocumentServiceUtils.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/query/DocumentServiceUtils.java
@@ -20,10 +20,12 @@ package io.stargate.web.docsapi.service.query;
 import io.stargate.web.docsapi.dao.DocumentDB;
 import io.stargate.web.docsapi.exception.ErrorCode;
 import io.stargate.web.docsapi.exception.ErrorCodeRuntimeException;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Objects;
 import java.util.OptionalLong;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
 
 public final class DocumentServiceUtils {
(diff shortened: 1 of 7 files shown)


restapi/src/test/java/io/stargate/web/docsapi/service/query/FilterExpressionTest.java
```diff
@@ -342,6 +342,32 @@ public void conditionFalse() {
       verifyNoMoreInteractions(condition);
     }
 
+    @Test
+    public void conditionFalsePathSegment() {
```
dimas-b, line 346: Do we have a unit test for a positive segment match? Sorry, I could not find it.

restapi/src/main/java/io/stargate/web/docsapi/service/query/DocumentServiceUtils.java
```diff
@@ -32,7 +34,30 @@
...
+   * Converts given path to an array one if needed:
+   *
+   * <ol>
+   *   <li>[1] - converted to [000001]
+   *   <li>[0],[1] - converted to [000000],[000001]
+   *   <li>[*] - not converted
+   *   <li>anyPath - not converted
+   * </ol>
+   *
+   * @param path filter or filed path
```
EricBorczuk, line 47: nit: `filed path` -> `field path`

restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterExpression.java
```diff
@@ -161,8 +162,16 @@ private boolean matchesFilterPath(Row row) {
         return false;
       }
 
-      // if not equal, fail
-      if (!Objects.equals(path, target)) {
+      boolean pathSegment = target.contains(",");
+      // if we have the path segment, we need to check if any matches
+      if (pathSegment) {
```
EricBorczuk, line 167: :+1:

dimas-b (approved):

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -1778,6 +1778,30 @@ public void searchPathSegmentMatching() throws Exception {
...
+    String r =
+        RestUtils.get(
+            authToken,
+            hostWithPort
+                + "/v2/namespaces/"
+                + keyspace
+                + "/collections/collection?page-size=3&where={\"n,m.value\": {\"$in\": [5]}}&raw=true",
+            200);
+
+    String expected = "{\"matching1\":{\"n\":{\"value\":5}}}";
```
EricBorczuk, line 1801: :+1: thanks for adding this case!

EricBorczuk (approved):
This looks good to me! Thanks for fixing and adding test cases

restapi/src/test/java/io/stargate/web/docsapi/service/query/FilterExpressionTest.java
ivansenic: Yea there is no need for this, we wanted to makes sure here that the path match, and I can not do that positive condition.

dimas-b: ok :+1:

dougwettlaufer (approved):

PR #1053, opened 2021-06-23
Get the user from the original clientState when cloning

dougwettlaufer:
Only applies to 3.11. Neither 4.0 or DSE implement the same cloning functionality.

### DIFF
--- a/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
+++ b/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
@@ -489,7 +489,7 @@ public class CassandraPersistence
           original.isInternal
               ? ClientState.forInternalCalls()
               : ClientState.forExternalCalls(original.getRemoteAddress());
-      clone.login(clone.getUser());
+      clone.login(original.getUser());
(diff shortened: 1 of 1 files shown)


persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/CassandraPersistence.java
```diff
@@ -489,7 +489,7 @@ private ClientState cloneWithKeyspace(ClientState original, String keyspace) {
           original.isInternal
               ? ClientState.forInternalCalls()
               : ClientState.forExternalCalls(original.getRemoteAddress());
-      clone.login(clone.getUser());
+      clone.login(original.getUser());
```
tomekl007, line 492: what's the reason behind this change? Does the `clone` not contain the same user? 
If so, maybe the method responsible for cloning should be fixed as well?

dougwettlaufer: This is the method responsible for the clone :). The line above initializes `clone` with a default ClientState, it's not until we call `login()` that a user is added so what was happening is we were calling `login()` with a null user.

dimas-b (approved):
Nice catch :+1:

PR #1054, opened 2021-06-23
Allow fetching document fields without WHERE

dimas-b:
Fixes #1046

PR #1058, opened 2021-06-24
relates to #1024: extract row matching

ivansenic:
* @dimas-b extracted row matching to use in #1054 
* ~added fix suggested from @mpenick -> https://github.com/stargate/stargate/pull/1047/files#r657991035~ kills way too many tests, will try in separate commit


dougwettlaufer (approved):

dimas-b (approved):

dimas-b:
> [...] kills way too many tests, will try in separate commit

Yes, test code does not cover 100% of the `Row` API... `ValidatingDataStore` and related classes need to be enhanced for @mpenick 's improvement to pass tests.

PR #1061, opened 2021-06-25
closes #1031: support parsing of the $or and $and expressions

### LINKED ISSUE #1031: Support parsing of $or and $and conditions
As the first step towards the `$or` in the document service, we should support the parsing of the `$or`. The `$or` should have a array JSON value with one or more conditions.

We should also explicitly support the `$and`, although we already do support it as all the condition in the same JSON object level would be wrapped in the and.

Example:

```
{
   "fixed":{
      "$eq":1
   },
   "$or":[
      {
         "one":{
            "$eq":1
         }
      },
      {
         "two":{
            "$eq":2
         }
      },
      {
         "$and":[
            {
               "three":{
                  "$eq":3
               }
            },
            {
               "four":{
                  "$eq":4
               }
            }
         ]
      }
   ]
}
```

should resolve as `AND(fixed=1, OR(one=1, two=2, AND(three=4, four=4)))`

ivansenic:
This supports parsing (only!) of the `$or` and `$and` conditions. The requests still fail if the `$or` is found with a `404: Searching documents with the $or condition is not yet supported.` 

Minimal changes in the parser, mostly tests.

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
@@ -19,23 +19,33 @@ package io.stargate.web.docsapi.service.query;
 import com.bpodgursky.jbool_expressions.And;
 import com.bpodgursky.jbool_expressions.Expression;
 import com.bpodgursky.jbool_expressions.Literal;
+import com.bpodgursky.jbool_expressions.Or;
 import com.bpodgursky.jbool_expressions.rules.RuleSet;
 import com.fasterxml.jackson.databind.JsonNode;
+import io.stargate.web.docsapi.exception.ErrorCode;
+import io.stargate.web.docsapi.exception.ErrorCodeRuntimeException;
 import io.stargate.web.docsapi.service.query.condition.BaseCondition;
 import io.stargate.web.docsapi.service.query.condition.ConditionParser;
 import io.stargate.web.docsapi.service.util.DocsApiUtils;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.ws.rs.core.PathSegment;
+import org.apache.commons.lang3.mutable.MutableInt;
 
 public class ExpressionParser {
+
+  private static final String OR_OPERATOR = "$or";
+
+  private static final String AND_OPERATOR = "$and";
 
   private static final Pattern PERIOD_PATTERN = Pattern.compile("\\.");
 
@@ -76,7 +86,8 @@ public class ExpressionParser {
   }
 
   /**
-   * Parses the root filter json node and returns back the list of the expressions found.
+   * Parses the root filter json node and returns back the list of the expressions found on the
+   * first level.
    *
    * @param prependedPath Given collection or document path segments
    * @param filterJson Filter JSON node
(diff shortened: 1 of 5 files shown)


restapi/src/test/java/io/stargate/web/docsapi/service/query/ExpressionParserIntTest.java
```diff
@@ -131,6 +136,124 @@ public void andWrappingRespected() {
...
+                          mapper
+                              .createArrayNode()
+                              .add(
+                                  mapper
+                                      .createObjectNode()
+                                      .set("c", mapper.createObjectNode().put("$eq", "c")))
+                              .add(
+                                  mapper
+                                      .createObjectNode()
+                                      .set("d", mapper.createObjectNode().put("$eq", "d"))))));
```
dimas-b, line 164: Would it be more readable to parse this JSON doc from a string?

restapi/src/test/java/io/stargate/web/docsapi/service/query/ExpressionParserIntTest.java
```diff
@@ -131,6 +136,124 @@ public void andWrappingRespected() {
...
+                                    }))
+                    .anySatisfy(
+                        c ->
+                            assertThat(c)
+                                .isInstanceOfSatisfying(
+                                    Or.class,
+                                    or -> {
+                                      List<?> orChildren = or.getChildren();
+                                      assertThat(orChildren)
+                                          .hasSize(2)
```
dimas-b, line 193: WDYT about using `toString()` of the filter expression for testing? It does not have to be a well-defined representation for end-users to work with, just stable enough for asserting in test. I believe it would make test code more readable.

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
```diff
@@ -19,24 +19,34 @@
...
 import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.ws.rs.core.PathSegment;
+import org.apache.commons.lang3.mutable.MutableInt;
 
 public class ExpressionParser {
 
+  private static final String OR_OPERATOR = "$or";
+
+  private static final String AND_OPERATOR = "$and";
```
dimas-b, line 48: I think if we introduce `OR` we ought to support `NOT` too, for the same of completeness... WDYT?

restapi/src/test/java/io/stargate/web/docsapi/service/query/ExpressionParserIntTest.java
```diff
@@ -509,5 +632,213 @@ public void multipleFields() {
...
+    }
+
+    @Test
+    public void orNotInArray() {
+      ObjectNode root = mapper.createObjectNode();
+      root.set(
+          "$or",
+          mapper
+              .createObjectNode()
+              .set("myField", mapper.createObjectNode().put("$eq", "some-value")));
```
dimas-b, line 731: Does the test name match actual use case here? I do not see an `IN` :thinking:

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
```diff
@@ -85,26 +96,99 @@ public ExpressionParser(ConditionParser predicateProvider) {
...
+   * Parses set of json nodes and returns back the list of the expressions found. Note that
+   * expression in the list contain only first level conditions in each node.
+   *
+   * @param prependedPath Given collection or document path segments
+   * @param nodes JSON nodes to resolve
+   * @param numericBooleans If number booleans should be used when creating conditions.
+   * @param nextIndex index counter for the expressions
+   * @return List of all expressions
+   */
+  private List<Expression<FilterExpression>> parse(
```
dimas-b, line 113: Would it make sense to return a single `Expression` to encapsulate all boolean operators? With current method signature I guess the individuals expressions in the list are implicitly joined by an `AND`, right? I think it would be nice to make that explicit.

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
```diff
@@ -85,26 +96,99 @@ public ExpressionParser(ConditionParser predicateProvider) {
...
+      boolean numericBooleans,
+      MutableInt nextIndex) {
+    if (!node.isArray()) {
+      throw new ErrorCodeRuntimeException(
+          ErrorCode.DOCS_API_SEARCH_FILTER_INVALID,
+          "The $or requires an array json node as value.");
+    }
+
+    List<Expression<FilterExpression>> orConditions =
+        parse(prependedPath, node, numericBooleans, nextIndex);
```
dimas-b, line 172: Now expression indexing depends on the JSON tree iteration order, which may be too obscure to the end user.

It might be worth adding explicit (but optional) selectivity hints. If a hint is _not_ provided we'd derive selectivity from relation positions. E.g. `{ field: {$eq: 1, $selectivity: 1000} }`.

For filters without an explicit selectivity, assume is it less (better) than the selectivity of all of the following filters in the same JSON array, and their nested filters.

When comparing filters with know selectivity with unknown selectivity assume that know is always better. WDYT?

I know it adds quite a bit of work to this PR, but I think we have to address the problem of how to allow users to control query plans in a simple and hopefully intuitive way since we do not have stats about real data distribution.

restapi/src/test/java/io/stargate/web/docsapi/service/query/ExpressionParserIntTest.java
ivansenic: This is a case when you have something like `"$or": {}` instead of `"$or: []`.. Will adapt..

restapi/src/test/java/io/stargate/web/docsapi/service/query/ExpressionParserIntTest.java
ivansenic: Nope, this can hide too many stuff.. Keeping it like this for now, it's not readable due to the formatting now the actual code.

restapi/src/test/java/io/stargate/web/docsapi/service/query/ExpressionParserIntTest.java
ivansenic: Not a bad idea, let's have this in all test cases, would be more understandable I guess.. :+1:

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
ivansenic: Pufff, I would keep this separated.. Also I am not sure how would we process this atm.. Additional issue?

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
ivansenic: This is a private method.. It's clearly specified in the JavaDocs what it does. There is another method for wrapping. Plus, it can not be done due to the recursion!

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
ivansenic: I like the idea, but this is a complete other issue imo.. Can make a ticket for it.. Let's make small steps faster.. OK?

restapi/src/test/java/io/stargate/web/docsapi/service/query/ExpressionParserIntTest.java
ivansenic: Done.

restapi/src/test/java/io/stargate/web/docsapi/service/query/ExpressionParserIntTest.java
ivansenic: Done.

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
dimas-b: yes. let's file an separate issue for this, but I think we'd have to address it before we enable `ORs` to be used in queries.

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
dimas-b: ok, pls file another issue.

ivansenic: https://github.com/stargate/stargate/issues/1071

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
dimas-b: Re: `private` - do we not have the same return type in the `public` entry point?

dimas-b: Re: recursion - agreed. I should have made the comment on the `public` method, sorry about the confusion.

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
ivansenic: https://github.com/stargate/stargate/issues/1072

restapi/src/test/java/io/stargate/web/docsapi/service/query/ExpressionParserIntTest.java
```diff
@@ -131,17 +132,114 @@ public void andWrappingRespected() {
...
+                    .hasSize(2)
+                    .anySatisfy(
+                        c ->
+                            assertThat(c)
+                                .isInstanceOfSatisfying(
+                                    FilterExpression.class,
+                                    first -> {
+                                      assertThat(first.getOrderIndex()).isZero();
+                                      assertThat(first.getFilterPath().getField()).isEqualTo("a");
+                                    }))
```
dimas-b, line 161: Should we assert this on the _first_ element, rather than "any"?

ivansenic: No, the order is not important and will be defined by the stupid hash comparator. What's important is that `orderIndex=0` is there for the field `a`.

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
ivansenic: The public method that returns the `List` is not used actually, I can make it private and make all tests go through that only public method?

dimas-b (approved):

restapi/src/test/java/io/stargate/web/docsapi/service/query/ExpressionParserIntTest.java
dimas-b: right, sorry about the false alarm :facepalm:

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
dimas-b: Yes, please. I think it would be more consistent to test through the public method that returns an `Expression`. Let's keep the internal method `private`.

dimas-b (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
ivansenic: Done.

PR #1062, opened 2021-06-25
gRPC tracing

tomekl007:


grpc/pom.xml
```diff
@@ -92,6 +92,12 @@
       <version>2.0.1</version>
       <scope>compile</scope>
     </dependency>
+      <dependency>
```
tomekl007, line 95: I've added this dependency because we need to transform UUID to `bytes[]`, and this logic is already implemented in the `UUIDGen.decompose`. However, the `UUIDGen` class is in this dependency. If that's problematic, we may decide to inline the decompose method. wdyt?

dougwettlaufer: I'd much rather inline it than pull in the cql module with all of it's dependencies.

dimas-b: `UUIDGen` comes from C* OSS, right? I think we wanted to remove that dependency from the `cql` module too (or at least contain it). +1 to adding a local utility class for this in SG.

grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -432,14 +435,24 @@ private void executePrepared(
...
                       case Rows:
                         responseBuilder.setResultSet(
                             Payload.newBuilder()
                                 .setType(query.getValues().getType())
                                 .setData(
                                     handler.processResult((Rows) result, query.getParameters())));
+                        handleTraceId(
+                            result.getTracingId(),
+                            query.getParameters().getTracing(),
+                            responseBuilder);
```
dimas-b, line 455: Could we use simple parameters here and move the logic of extracting sub-values to `handleTraceId`? This would reduce code duplication (cf. lines 439-442), WDYT?

grpc/pom.xml
mpenick: Maybe use a `string` instead of `bytes`?  It might also be nicer for the user too, since it's a UUID.

tomekl007: +1 on using `string`. I thought that the schema in:
```
message Response {
  Payload result_set = 1;
  bytes tracing_id = 2;
  repeated string warnings = 3;
}
```
cannot be changed in a non-compatible way. But if we can, That would solve this. 
Done in: https://github.com/stargate/stargate/pull/1062/commits/432a7954ddba51f670aa14ad45387bb0c9f4664f

grpc/src/main/java/io/stargate/grpc/service/Service.java
tomekl007: There are two types of parameters: `QueryParameters` AND `BatchParameters` so we cannot just pass parameters there. I can extract two helper methods that are extracting `tracingId`. 
Done in:
https://github.com/stargate/stargate/pull/1062/commits/231e839134e9ad213b725dba7bbaebc2eafe5d7e

mpenick (approved):
+1. Looks great.

dimas-b (commented):
belated +1 :)

PR #1066, opened 2021-06-25
Retry queries in grpc on PreparedQueryNotFoundException

### LINKED ISSUE #1065: gRPC's prepared query cache gets out of sync with persistence's query cache
The prepared query cache in gRPC does not currently have a cache invalidation mechanism and because of this when a prepared query is removed from the cache in the persistence module it does not also get removed from the gRPC cache. 

I was able to reproduce this issue by

1. Create a keyspace and table
1. Execute an insert query on the table
1. Drop the keyspace
1. Recreate the same keyspace and table
1. Try to run the same insert query again and it fails

To resolve this we need to either a) remove the local cache in gRPC or b) add cache invalidation in order to keep the caches in sync.

dougwettlaufer:
Resolves #1065

### DIFF
--- a/grpc/src/main/java/io/stargate/grpc/service/Service.java
+++ b/grpc/src/main/java/io/stargate/grpc/service/Service.java
@@ -70,6 +70,7 @@ import org.apache.cassandra.stargate.exceptions.AlreadyExistsException;
 import org.apache.cassandra.stargate.exceptions.CasWriteUnknownResultException;
 import org.apache.cassandra.stargate.exceptions.FunctionExecutionException;
 import org.apache.cassandra.stargate.exceptions.PersistenceException;
+import org.apache.cassandra.stargate.exceptions.PreparedQueryNotFoundException;
 import org.apache.cassandra.stargate.exceptions.ReadFailureException;
 import org.apache.cassandra.stargate.exceptions.ReadTimeoutException;
 import org.apache.cassandra.stargate.exceptions.UnavailableException;
@@ -156,7 +157,7 @@ public class Service extends io.stargate.proto.StargateGrpc.StargateImplBase {
                 if (t != null) {
                   handleException(t, responseObserver);
                 } else {
-                  executePrepared(connection, prepared, query, responseObserver);
+                  executePrepared(connection, prepared, query, responseObserver, prepareInfo);
                 }
               });
     } catch (Throwable t) {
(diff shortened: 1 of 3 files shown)


grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -426,7 +428,22 @@ private void executePrepared(
           .whenComplete(
               (result, t) -> {
                 if (t != null) {
-                  handleException(t, responseObserver);
+                  if (t instanceof PreparedQueryNotFoundException) {
```
ivansenic, line 431: Are we handling those in other modules?

dougwettlaufer: Logic borrowed from persistence-api 
https://github.com/stargate/stargate/blob/6452957cd078f17f1c0067c943e3fc2129f61ceb/persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedDataStore.java#L115

grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -426,7 +428,22 @@ private void executePrepared(
...
+                    preparedCache.invalidate(prepareInfo);
+
+                    prepareQuery(connection, prepareInfo, parameters.getTracing())
+                        .whenComplete(
+                            (p, t1) -> {
+                              if (t1 != null) {
+                                handleException(t1, responseObserver);
+                              } else {
+                                executePrepared(
```
dimas-b, line 440: Should we be checking for infinite loops here? It's probably not likely, but still... :thinking:

mpenick:
I think we need to also handle unprepared queries in batches.

grpc/src/main/java/io/stargate/grpc/service/Service.java
dougwettlaufer: I thought about that but noticed `PersistenceBackedDataStore.java` doesn't check so I figured we'd be fine without it.

dimas-b: I do believe the same problem exists there too, but I think it is not likely to result in real issues. The schema would have to change very often or the statement preparation has to have bugs for this infinite loop to manifest. +0

tomekl007: Maybe we should do an exponential backoff before retry?

dimas-b (approved):

PR #1077, opened 2021-06-29
Include trace table content in the gRPC message

tomekl007:
todo:
- [x] propagate tracing consistency level in the `QueryParameters` and `BatchParameters`
- [x] create a complex trace type that contains session data and a list of events
- [x] extract more tracing data from a session and events
- [x] decide whether to use executor per trace query or one executor with more number threads
- [x] unit or integration test `QueryTracingFetcher` (regarding the unit, it is very hard to mock the `DataStore`, so it may be not feasible). There is already an integration test `TracingQueryTest` but is specific to `gRPC` module, whereas the `QueryTracingFetcher` is in the `api` module). However, maybe it is good enough?


grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -143,6 +156,8 @@ public void executeQuery(Query query, StreamObserver<Response> responseObserver)
       Connection connection = newConnection(authenticationSubject.asUser());
       QueryParameters queryParameters = query.getParameters();
 
+      CompletableFuture<Prepared> tracingPrepareQuery = prepareTracingQuery(connection);
```
tomekl007, line 159: we could prepare this query only if tracing is enabled. However, there is only one prepared query so overhead is low, and paid only once (during the first query). I am not sure if introducing the if is worth complicating the code.

grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -563,6 +731,20 @@ private Parameters makeParameters(QueryParameters parameters) {
     return builder.tracingRequested(parameters.getTracing()).build();
   }
 
+  private Parameters makeTracingParameters(QueryParameters parameters) {
+    ImmutableParameters.Builder builder = ImmutableParameters.builder();
+    // only consistency levels can be inherited by tracing query
+    if (parameters.hasConsistency()) {
```
mpenick, line 737: I'm not sure it makes sense to inherit the consistency levels from the original queries for the tracing queries. For example the java-driver uses a default consistency of ONE for tracing queries. 

https://github.com/datastax/java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/api/core/config/OptionsMap.java#L293

grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -421,41 +454,60 @@ private void executePrepared(
...
+        Response response = responseBuilder.build();
+        responseObserver.onNext(response);
+        responseObserver.onCompleted();
+      } else {
+        try {
+          long queryStartNanoTime = System.nanoTime();
+
+          CompletableFuture<Result> queryExecute =
+              connection.execute(
+                  bindValues(handler, preparedTracing, cqlPayload(responseBuilder.getTracingId())),
```
mpenick, line 494: It was not the original intent of the payload handler logic for handling bind values and converting result sets for internal queries. It's purpose is to convert to/from the internal representation to gRPC payload values.

Perhaps it would be better to use `DataStore` for tracing/internal queries? I wonder if we could add a way to make a `DataStore` from an existing `Connection`.  If not, I don't think it's very expensive to construct a new `DataStore`.

grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -563,6 +731,20 @@ private Parameters makeParameters(QueryParameters parameters) {
...
   }
 
+  private Parameters makeTracingParameters(QueryParameters parameters) {
+    ImmutableParameters.Builder builder = ImmutableParameters.builder();
+    // only consistency levels can be inherited by tracing query
+    if (parameters.hasConsistency()) {
+      builder.consistencyLevel(
+          ConsistencyLevel.fromCode(parameters.getConsistency().getValue().getNumber()));
+    }
+    if (parameters.hasSerialConsistency()) {
```
mpenick, line 741: Serial consistency should have no effect on tracing queries.

grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -77,12 +84,18 @@
...
 
 public class Service extends io.stargate.proto.StargateGrpc.StargateImplBase {
-
+  private static final Logger log = LoggerFactory.getLogger(Service.class);
   public static final Context.Key<AuthenticationSubject> AUTHENTICATION_KEY =
       Context.key("authentication");
   public static final Context.Key<SocketAddress> REMOTE_ADDRESS_KEY = Context.key("remoteAddress");
+  private static final String SYSTEM_TRACES_KEYSPACE = "system_traces";
+  public static final String TRACING_PREPARE_QUERY =
+      "select activity, source, source_elapsed, thread from events where session_id = ?";
```
mpenick, line 98: I'm pretty sure you need to query the  `system_traces.session` table first. I think it acts as "gate" to signifying tracing is complete.  I'm digging into the server-side implementation more to verify.

Check the implementation in the java-driver: https://github.com/datastax/java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/internal/core/cql/QueryTraceFetcher.java

It queries "SELECT * FROM system_traces.sessions WHERE session_id = ?" and validates `duration` and `started_`:

`row == null || row.isNull("duration") || row.isNull("started_at")`

mpenick: minor: would prefer: `SELECT activity, source, source_elapsed, thread FROM events WHERE session_id = ?`. Makes this easier to see in code as a query.

grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -77,12 +84,18 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class Service extends io.stargate.proto.StargateGrpc.StargateImplBase {
-
+  private static final Logger log = LoggerFactory.getLogger(Service.class);
   public static final Context.Key<AuthenticationSubject> AUTHENTICATION_KEY =
       Context.key("authentication");
   public static final Context.Key<SocketAddress> REMOTE_ADDRESS_KEY = Context.key("remoteAddress");
+  private static final String SYSTEM_TRACES_KEYSPACE = "system_traces";
```
mpenick, line 96: minor:  Seem unnecessary: `private static final String SYSTEM_TRACES_KEYSPACE = "system_traces";`  It's used one time and it's very unlikely keyspace name will ever change.

grpc/src/main/java/io/stargate/grpc/service/Service.java
mpenick: Do tracing queries need to be prepared at all? Maybe just use `DataStore` and it's handled automatically.

mpenick (changes requested):
I'll dig more into the server-side implementations of tracing tomorrow to better inform decisions on handling tracing data availability. In general, I think this could be done using `DataStore` instead of modifying `PayloadHandler`.

grpc/src/main/java/io/stargate/grpc/service/Service.java
tomekl007: Regarding adding the tracing query logic to the `DataStore`, are you sure that this is the correct place?
The `DataStore` can have N implementations. We already have `PersistenceBackedDataStore` and `ValidatingDataStore`. If we decide to add a new method for fetching tracing, we would need to duplicate it. Maybe the best middle ground would be a dedicated `TraceFetcher` that takes the connection as an argument wdyt?

grpc/src/main/java/io/stargate/grpc/service/Service.java
tomekl007: thank you for this, I've implemented a new `QueryTracingFetcher` that uses a similar mechanism to java driver: firstly queries the `session` table then `events`. 
I've moved the tracing fetcher to the `persistence-api` module. The client of this class needs to map `Rows` to the expected type ourselves - for `gRPC`, see the `TraceEventsMapper`.
I didn't move it directly to the `DataStore` because of: https://github.com/stargate/stargate/pull/1077#discussion_r662044829

persistence-api/src/main/java/io/stargate/db/tracing/QueryTracingFetcher.java
```diff
@@ -0,0 +1,159 @@
...
+  private final Parameters parameters;
+  private final ScheduledExecutorService executorService;
+  private final ByteBuffer tracingIdBytes;
+
+  public QueryTracingFetcher(UUID tracingId, Persistence.Connection connection) {
+    this.tracingId = tracingId;
+    this.tracingIdBytes = decompose(tracingId);
+    this.connection = connection;
+    this.parameters = createTracingQueryParameters();
+    this.executorService = Executors.newSingleThreadScheduledExecutor();
```
tomekl007, line 52: This is not ideal. 
Right now it will create a dedicated executor per tracing invocation. We may consider creating one executor with a higher number of threads, wdyt?

mpenick:
I took some time to understand how tracing works. It's basically asynchronous unless it's configured to wait,  but even then it can silently (except for a logging entry). 

The keyspace schema is (written at consistency ANY):

```
CREATE KEYSPACE system_traces WITH replication = {'class': 'NetworkTopologyStrategy', 'us-east-1': '3'}
```


Cassandra 3.11:

Here's the wait logic:
https://github.com/apache/cassandra/blob/cassandra-3.11.8/src/java/org/apache/cassandra/tracing/TraceStateImpl.java#L77-L98

The default timeout is 0:
https://github.com/apache/cassandra/blob/cassandra-3.11.8/src/java/org/apache/cassandra/tracing/TraceStateImpl.java#L52-L53

Also, traces are written with consistency ANY:
https://github.com/apache/cassandra/blob/casandra-3.11.8/src/java/org/apache/cassandra/tracing/TraceStateImpl.java#L122


Cassandra 4.0 and DSE 6.8 are essentially the same implementation:
https://github.com/apache/cassandra/blob/cassandra-4.0-beta4/src/java/org/apache/cassandra/tracing/TraceStateImpl.java
https://github.com/riptano/bdp/blob/6.8.13/dse-db/src/java/org/apache/cassandra/tracing/TraceStateImpl.java

grpc/src/main/java/io/stargate/grpc/service/Service.java
mpenick: The code is already using `Persistence` directly. Why use the concrete impl `PersistenceBackedDataStore`?

tomekl007:
> I took some time to understand how tracing works. It's basically asynchronous unless it's configured to wait, but even then it can silently fail (except for a logging entry).
> 
> The keyspace schema is (written at consistency ANY):
> 
> ```
> CREATE KEYSPACE system_traces WITH replication = {'class': 'NetworkTopologyStrategy', 'us-east-1': '3'}
> ```
> 
> Cassandra 3.11:
> 
> Here's the wait logic:
> https://github.com/apache/cassandra/blob/cassandra-3.11.8/src/java/org/apache/cassandra/tracing/TraceStateImpl.java#L77-L98
> 
> The default timeout is 0:
> https://github.com/apache/cassandra/blob/cassandra-3.11.8/src/java/org/apache/cassandra/tracing/TraceStateImpl.java#L52-L53
> 
> Also, traces are written with consistency ANY:
> https://github.com/apache/cassandra/blob/casandra-3.11.8/src/java/org/apache/cassandra/tracing/TraceStateImpl.java#L122
> 
> Cassandra 4.0 and DSE 6.8 are essentially the same implementation:
> https://github.com/apache/cassandra/blob/cassandra-4.0-beta4/src/java/org/apache/cassandra/tracing/TraceStateImpl.java
> https://github.com/riptano/bdp/blob/6.8.13/dse-db/src/java/org/apache/cassandra/tracing/TraceStateImpl.java

Thanks. In that case, the solution with scheduled polling (same as the driver, and same as I've implemented now) should be good for our use case. Do you agree?

persistence-api/src/main/java/io/stargate/db/tracing/QueryTracingFetcher.java
mpenick: Agreed, a shared executor would be preferred. I'm wondering if there's an existing one we can use?

persistence-api/src/main/java/io/stargate/db/tracing/TracingData.java
```diff
@@ -0,0 +1,22 @@
+package io.stargate.db.tracing;
+
+import io.stargate.db.datastore.Row;
+import java.util.List;
+
+public class TracingData {
+  private final List<Row> events;
```
tomekl007, line 7: We could create a structure similar to java-driver: https://github.com/datastax/java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/internal/core/cql/DefaultTraceEvent.java
However, we still would need to map it to the `gRPC` structure, so I think it's better to propagate rows and let the caller do the actual mapping. wdyt?

persistence-api/src/main/java/io/stargate/db/tracing/QueryTracingFetcher.java
tomekl007: I've changed the logic to use one static `ScheduledExecutorService` here. It is created with the same config, as the java-driver scheduler for this component. Since the defaults are proven for drivers, it should meet our needs here as well.

grpc-proto/proto/query.proto
```diff
@@ -199,10 +200,25 @@ message ColumnSpec {
...
+  }
+  string id = 1;
+  int64 duration = 2;
+  int64 started_at = 3;
+  repeated Event events = 4;
+}
+
 message Response {
   Payload result_set = 1;
   string tracing_id = 2;
```
mpenick, line 219: Should `tracing_id` be removed? It's now in `traces`.

tomekl007: That's a good point. I've removed the `tracing_id` from this level. 
However, I need a way to propagate the `tracing_id` from the `handleBatchQuery` and `handleQuery` to  `executeTracingQueryIfNeeded`. For that purpose, I've created a `ResponseAndTraceId` wrapper class.
The con of this approach is that I don't need to transform `UUID -> String -> UUID`, but I can carry over a `UUID` directly.

dougwettlaufer (approved):

mpenick (approved):
Looks good.

PR #1078, opened 2021-06-29
GraphQL CQL-first: Fix handling of conditional batches

### LINKED ISSUE #1067: GraphQL CQL-first: batched mutations don't handle LWTs correctly
Data model:

```
create table foo(k int primary key, v int);
insert into foo(k,v) values(1,1);
```

A single update works as expected, `applied` is false and it returns the conflicting value:

```graphql
mutation {
  update1: updatefoo(value: { k: 1, v: 2 }, ifCondition: { v: { eq: 0 } }) {
    applied
    value { v }
  }
}

# Response:
    "update1": {
      "applied": false,
      "value": {
        "v": 1
      }
    }
```

However if we batch two of these mutations (stupid query but it is technically valid CQL), the results are incorrect:

```graphql
mutation @atomic {
  update1: updatefoo(value: { k: 1, v: 2 }, ifCondition: { v: { eq: 0 } }) {
    applied
    value { v }
  }
  update2: updatefoo(value: { k: 1, v: 3 }, ifCondition: { v: { eq: 0 } }) {
    applied
    value { v }
  }
}

# Response:
    "update1": {
      "applied": null,
      "value": {
        "v": 2
      }
    },
    "update2": {
      "applied": null,
      "value": {
        "v": 3
      }
    }
```

This is because `MutationFetcher.executeAsPartOfBatch` always reflect the input value:

```java
      return batchContext
          .getExecutionFuture()
          .thenApply(v -> ImmutableMap.of("value", environment.getArgument("value")));
```

It should invoke `toMutationResult` instead, like the non-batched variant.

NB: once we do that, we'll run into another issue because the result set cannot be consumed twice; the second mutation will see no rows and assume the mutation has succeeded. We can solve that by making `batchContext.getExecutionFuture()` a `CompletableFuture<Row>` (a batch returns at most one row).

### LINKED ISSUE #1068: GraphQL CQL-first: bulk insert does not handle conditional batch
Disclaimer: the query is convoluted, I'm not sure if there's any realistic use case.

CQL:

```
create table foo(k int primary key, v int);
insert into foo(k,v) values(1,1);
```

GraphQL:

```graphql
mutation @atomic {
  bulkInsertfoo(values: [{ k: 1, v: 2 }, { k: 1, v: 3 }], ifNotExists: true) {
    applied
    value {
      k
    }
  }
  insertfoo(value: { k: 1, v: 4 }, ifNotExists: true) {
    applied
    value {
      v
    }
  }
}
```

This causes:

```
java.lang.IndexOutOfBoundsException: Index: 1, Size: 1
        at io.stargate.graphql.schema.cqlfirst.dml.fetchers.DmlFetcher.toListOfMutationResults(DmlFetcher.java:122)
```

The methods assumes that if there are rows then it's one per input value, but if the batch is conditional, only one row is returned.

olim7t:
Fixes #1067
Fixes #1068

### DIFF
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/DmlFetcher.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/DmlFetcher.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.UnaryOperator;
+import java.util.stream.Collectors;
 import org.apache.cassandra.stargate.db.ConsistencyLevel;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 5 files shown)


graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/DmlFetcher.java
```diff
@@ -109,24 +110,64 @@ protected DmlFetcher(Table table, NameMapping nameMapping) {
...
   }
 
+  protected Map<String, Object> toBatchResult(List<Row> rows, Map<String, Object> originalValue) {
+    return isAppliedBatch(rows)
+        ? toAppliedMutationResultWithOriginalValue(originalValue)
+        : toUnappliedBatchResult(rows, originalValue);
+  }
+
+  private boolean isAppliedBatch(List<Row> batchRows) {
+    if (batchRows.isEmpty()) {
```
tomekl007, line 135: It means that if there are no rows returned we treat it as success. However, also if one row is returned we are extracting it's applied. Could you elaborate on when the first and the second situation are happening?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/DmlFetcher.java
```diff
@@ -109,24 +110,64 @@ protected DmlFetcher(Table table, NameMapping nameMapping) {
...
+    }
+    Row row = batchRows.get(0);
+    return row.columns().stream().map(Column::name).anyMatch("[applied]"::equals)
+        && row.getBoolean("[applied]");
+  }
+
+  private Map<String, Object> toUnappliedBatchResult(
+      List<Row> rows, final Map<String, Object> originalValue) {
+    Map<String, Object> primaryKey =
+        table.primaryKeyColumns().stream()
```
tomekl007, line 149: isn't the `table` containing all the fields?
What if the graphQL selection does not specify all the fields to be returned?
Are we filling it anyway, but the GraphQL returns only those fields that are in the user's selection?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/DmlFetcher.java
```diff
@@ -109,24 +110,64 @@ protected DmlFetcher(Table table, NameMapping nameMapping) {
...
+            .collect(
+                Collectors.toMap(
+                    Column::name, column -> toDBValue(column, originalValue.get(column.name()))));
+    return rows.stream()
+        .filter(row -> matches(row, primaryKey))
+        .findFirst()
```
tomekl007, line 155: why only the first result is relevant?

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/DmlFetcher.java
olim7t: No rows is when a batch has no conditions:
```
cqlsh:test> begin batch
        ... insert into foo(k,cc,v) values(1,1,1);
        ... insert into foo(k,cc,v) values(1,2,2);
        ... apply batch;

<< no rows (void) >>
```
One row with `applied` is when the batch has conditions that are met:
```
cqlsh:test> truncate table foo;
cqlsh:test> begin batch
        ... insert into foo(k,cc,v) values(1,1,1) if not exists;
        ... insert into foo(k,cc,v) values(1,2,2);
        ... apply batch;

 [applied]
-----------
      True
```

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/DmlFetcher.java
olim7t: There can't be more than one result. A conditional batch response only contains each primary key once, even if there were multiple queries on that row (one of the integration tests covers that).

graphqlapi/src/main/java/io/stargate/graphql/schema/cqlfirst/dml/fetchers/DmlFetcher.java
olim7t: We're filling the response with `row`, which may or may not contain all the fields depending on the original request (this PR doesn't change anything in that regard).

The goal of this method is to pick the correct row, because they are not necessarily in the right order (see `AtomicBatchTest. failedConditionalBatchOutOfOrder`).

tomekl007 (approved):

PR #1079, opened 2021-06-29
Add handling/testing for 400's on malformed JSON in requests

EricBorczuk:
This ensures that on any user input that involves JSON, a 400 will be returned when there is malformed data.

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/error/ErrorHandler.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/error/ErrorHandler.java
@@ -18,7 +18,6 @@
 package io.stargate.web.docsapi.resources.error;
 
 import com.datastax.oss.driver.api.core.NoNodeAvailableException;
-import com.fasterxml.jackson.core.JsonProcessingException;
 import io.stargate.auth.UnauthorizedException;
 import io.stargate.web.docsapi.exception.ErrorCodeRuntimeException;
 import io.stargate.web.models.Error;
@@ -48,10 +47,6 @@ public final class ErrorHandler
               new Error(
                   "Role unauthorized for operation: " + throwable.getMessage(),
                   Response.Status.UNAUTHORIZED.getStatusCode()))
-          .build();
-    } else if (throwable instanceof JsonProcessingException) {
-      return Response.status(Response.Status.BAD_REQUEST)
-          .entity(new Error("Unable to process JSON", Response.Status.BAD_REQUEST.getStatusCode()))
           .build();
     } else if (throwable instanceof NoNodeAvailableException) {
       return Response.status(Response.Status.SERVICE_UNAVAILABLE)
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/error/ErrorHandler.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/error/ErrorHandler.java
@@ -18,6 +18,7 @@
 package io.stargate.web.docsapi.resources.error;
 
 import com.datastax.oss.driver.api.core.NoNodeAvailableException;
+import com.fasterxml.jackson.core.JsonProcessingException;
 import io.stargate.auth.UnauthorizedException;
 import io.stargate.web.docsapi.exception.ErrorCodeRuntimeException;
 import io.stargate.web.models.Error;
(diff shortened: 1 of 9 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -169,7 +169,17 @@ private boolean isEmptyArray(Object v) {
       boolean isJson) {
     String trimmed = payload.trim();
     if (isJson) {
-      return shredJson(surfer, db, path, key, trimmed, patching);
+      try {
+        return shredJson(surfer, db, path, key, trimmed, patching);
+      } catch (RuntimeException e) {
```
ivansenic, line 174: Shouldn't we simply catch the concrete `RuntimeException` that you want to transform to the `ErrorCodeRuntimeException`? What s the exception thrown here?

Again I would catch a proper exception in `shredJson` based on the user input only (see other comment).

restapi/src/main/java/io/stargate/web/docsapi/resources/error/ErrorHandler.java
```diff
@@ -48,6 +49,10 @@ public Response apply(Throwable throwable) {
                   "Role unauthorized for operation: " + throwable.getMessage(),
                   Response.Status.UNAUTHORIZED.getStatusCode()))
           .build();
+    } else if (throwable instanceof JsonProcessingException) {
+      return Response.status(Response.Status.BAD_REQUEST)
```
ivansenic, line 53: Although this will work and is good for being safe, shouldn't we catch exceptions at the place where they can occur and provide more info to the user? See `ReactiveDocumentsService` to see how I handled that so that user has a nicer feedback.

Furthermroe if the `JsonProcessingException` is caught here, then you don't know if this is related to user input or not, and you are potentially hiding a real `500` that occurred due to the bug in our code. So I would strongly advise to not do it this way.

ivansenic (changes requested):
Check the comments and let's discuss if needed :call_me_hand:

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: So unfortunately, the exception thrown by `JsonSurfer` on malformed JSON is a plain RuntimeException, which is why I took this approach...since `shredJson` is really only managing user input, I chose to do the handling here...not ideal but there's not much for me to work with since the JsonSurfer API is throwing a plain RuntimeException, specificity is lost

restapi/src/main/java/io/stargate/web/docsapi/resources/error/ErrorHandler.java
EricBorczuk: Yeah, I can change this up and maybe throw the appropriate `ErrorCodeRuntimeException` at each location where processing input can find malformed JSON

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
ivansenic: OK appraoch is fine then, as we can not really do anything about it.. Can we just move the whole to the method itself, then callers of the method don't have to care to repeat over and over the same procedure? Then I guess we can remove changes here and on the https://github.com/stargate/stargate/pull/1079/files#diff-caf69767609da0fee30d28b2b410ba5e400ed01e1f42af082f495ef82fec8416R494

ivansenic (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -476,15 +491,22 @@ private DocumentDB maybeCreateTableAndIndexes(
...
+                                  surfer,
+                                  finalDb,
+                                  Collections.emptyList(),
+                                  data.getKey(),
+                                  data.getValue(),
+                                  false)
+                              .left);
+                    } catch (RuntimeException e) {
+                      throw new ErrorCodeRuntimeException(
+                          ErrorCode.DOCS_API_INVALID_JSON_VALUE,
```
dougwettlaufer, line 506: This is catching RuntimeException, do we need to check the type of the exception to see if it's json related or do we know for sure that's what gets thrown from here?

EricBorczuk: I think I am going to change `shredJson` to just directly throw an `ErrorCodeRuntimeException`, but yes - because the function only handles JSON input (and JsonSurfer only throws plain RuntimeExceptions), I think it is safe to assume this

Issue #1027, opened 2021-06-17
Support $exists with false


ivansenic:
Refactoring in the #981 allowed now to support `$exists` filter op with the `false` value. It would allow to check if the field is not existing in a document.

ivansenic:
@EricBorczuk @dimas-b Can you help me out a bit. 

I the old implementation we allowed not equal checks with nulls (`"$ne": null`), however in the new implementation this is not supported (seems like a bug). However, since the other ops (`$eq, $gt, $gte, $lt, $lte`) do not support `null` as filter value, would it be fine if we simply implement this issue and keep `$ne` not supporting the nulls. It would be kind a breaking change. Or do we want to actually support both, exists with false and not equals with nulls, which is essentially the same. Or?

dimas-b:
I'd say since `$eq: null` is not supported (user error, I assume), it makes sense for `$ne: null` to report the same kind of error.

My rationale is that if we have full boolean expression support, then `field $ne null` would be equivalent to `NOT( field $eq null)`, while the inner predicate in the latter is not supported.

tatu-at-datastax:
Looks like Documents API might be implementing "$exists" different from REST API -- to me, Doc API got it right, whereas REST API, as per #1518 seems to consider it an alias for "booleanField $EQ true" which does not seem very useful.

Not sure what to do about that but noting this as relevant.

PR #1080, opened 2021-06-30
closes #1028: refactored in-document search

### LINKED ISSUE #1028: Document search service to support in document search
The new reactive service should also support in document search. This would align the two search types to the same. Right now, there are some differences as the collection search provides some extra features and better performance.

Implementing this would allow us to remove a majority of code from the `DocumentService`.

ivansenic:
This is draft for the refactored in-document search for sub-documents. The majority of implementation will stay as it is, however, we still need to discuss the input parameters meaning (and limitations) and output format. The main concern here being backward compatible.

Things that have been done:

* [x] loading the sub-paths once and doing complete tree evaluation (this can be optimized in case when we have one single expression -> direct matching)
* [x] out-of-the-box support for ORs
* [x] build on RxJava, reusing lot of stuff from the full collection search
* [x] tests
* [x] final agreement on inputs and outputs  
* [ ] remove old code
* [x] blocked by #1089 
* [x] blocked by #1116 (revert tests)
* [ ] align with #1105 
* [x] rebase on #1125 (upgrade tests)

~**Current issues ~ HELP NEEDED**~

~1. *Failing tests* - I am getting `404` in all int tests that have the array notation in the path (f.e. `/doc-id/some/path/[0]`), this worked before in the old resource class, I changed nothing here. HTML page returned (@EricBorczuk any idea?)~
~2. *Failing tests* - `$ne: null` fails, not supported atm, see [discussion](https://github.com/stargate/stargate/issues/1027#issuecomment-875641915),  remove the test for now?~
~3.  *Updated tests* - updated [two cases](https://github.com/stargate/stargate/pull/1080/files#diff-9fa942a58c33d5742059dedc43ada743adb2c6c83159c1a624fc1e05d11521c1R1525) where `*` was used to match array element. Should we rather allow it, basically `*` would match path or array, `[*]` would match array only?~


ivansenic:
@EricBorczuk @dimas-b Although this is in draft still, we could start reviewing it.. We need to resolve issues mentioned in the PR description. In addition, we need to confirm that loading all rows on the sub-path is fine.. I am open for questions :call_me_hand:

restapi/src/main/java/io/stargate/web/impl/Server.java
```diff
@@ -150,8 +150,8 @@ protected void configure() {
 
     // Documents API
     environment.jersey().register(new DocsApiComponentsBinder(environment));
-    environment.jersey().register(DocumentResourceV2.class);
     environment.jersey().register(ReactiveDocumentResourceV2.class);
+    environment.jersey().register(DocumentResourceV2.class);
```
EricBorczuk, line 154: Is it possible that the 404's you are observing are due to this reordering? I forget how Dropwizard chooses to resolve endpoints when the path you provide techncially matches multiple paths

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -1523,7 +1522,7 @@ public void testMultiSearch() throws IOException {
         RestUtils.get(
             authToken,
             collectionPath
-                + "/cool-search-id?where={\"quiz.nests.q2.options.*.this.that.them\": {\"$eq\": false}}&raw=true",
+                + "/cool-search-id?where={\"quiz.nests.q2.options.[*].this.that.them\": {\"$eq\": false}}&raw=true",
```
dimas-b, line 1525: > Should we rather allow it, basically * would match path or array, [*] would match array only?

I think this ^ makes sense.

restapi/src/main/java/io/stargate/web/docsapi/resources/ReactiveDocumentResourceV2.java
```diff
@@ -128,16 +336,20 @@ public void searchDoc(
...
+        .map(rawDocumentHandler(raw))
         .safeSubscribe(
             AsyncObserver.forResponseWithHandler(
                 asyncResponse, ErrorHandler.EXCEPTION_TO_RESPONSE));
   }
+
+  private Function<DocumentResponseWrapper<? extends JsonNode>, Response> rawDocumentHandler(
+      @QueryParam("raw") @ApiParam(value = "Unwrap results", defaultValue = "false") Boolean raw) {
+    return results -> {
+      if (raw == null || !raw) {
```
dimas-b, line 348: nit: invert the `if` condition for more readability, i.e. `raw != null && raw`?

dimas-b (commented):
Sorry, I could not do a comprehensive review today... I'll continue tomorrow

restapi/src/main/java/io/stargate/web/impl/Server.java
ivansenic: Yea I added reordering in hope to actually solve it.. Forgot to revert.. Are there any matching ones to this one?

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
ivansenic: Allow it? So revert the test and fix the matching on `master`?

ivansenic: Agreed star matches everything.

ivansenic: Done in #1116

restapi/src/main/java/io/stargate/web/docsapi/resources/ReactiveDocumentResourceV2.java
ivansenic: Done.

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -135,28 +145,353 @@ public ReactiveDocumentService(
...
+                            fieldPathsFinal,
+                            collector,
+                            false,
+                            db.treatBooleansAsNumeric());
+
+                    // dead leaf deletion init on non-empty collection
+                    if (!collector.isEmpty()) {
+                      logger.info(
+                          String.format("Deleting %d dead leaves", collector.getLeaves().size()));
+                      long now = timeSource.currentTimeMicros();
```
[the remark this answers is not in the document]
ivansenic: Done.

restapi/src/main/java/io/stargate/web/impl/Server.java
ivansenic: Fixed.

ivansenic:
@dimas-b @EricBorczuk OK I am marking this as ready to review.. I really gave my best to keep it all in the backward compatible mode and unfortunately some peaces of the new code are also very ugly and quite confusing..

We can remove the dead code in the follow-up commit..

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -135,28 +144,367 @@ public ReactiveDocumentService(
...
+    for (RawDocument doc : docs) {
+      // create document node and set to result
+      JsonNode node =
+          documentToNode(doc, fieldPaths, writeAllPathsAsObjects, db.treatBooleansAsNumeric());
+
+      // TODO Confirm: adding empty nodes to the results array makes no sense?
+      //  should be revisited
+      if (!node.isEmpty()) {
+        docsResult.add(node);
+      }
```
dimas-b, line 474: +1 to exclude empty nodes. @EricBorczuk : WDYT?

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -1424,49 +1413,48 @@ public void testInvalidSearch() throws IOException {
...
     assertThat(r)
-        .isEqualTo(
-            "{\"description\":\"The operation $exists only supports the value `true`\",\"code\":400}");
+        .isEqualTo("{\"description\":\"$exists only supports the value `true`\",\"code\":400}");
 
     r =
         RestUtils.get(
             authToken, collectionPath + "/cool-search-id?where={\"a\": {\"exists\": true}}}", 400);
-    assertThat(r).startsWith("{\"description\":\"Invalid operator: exists, valid operators are:");
+    assertThat(r).startsWith("{\"description\":\"Operation exists is not supported.");
```
dimas-b, line 1429: I'd quote the invalid operator for clarity. 

Also, was there any particular reason for removing valid operators from this message?

dimas-b (changes requested):

restapi/src/test/java/io/stargate/web/docsapi/service/query/DocumentSearchServiceTest.java
```diff
@@ -477,4 +480,339 @@ public void fullSearch() {
...
+      Paginator paginator = new Paginator(null, 20);
+      ExecutionContext context = ExecutionContext.create(true);
+      FilterPath filterPath = ImmutableFilterPath.of(Collections.singletonList("field"));
+      BaseCondition condition = ImmutableStringCondition.of(EqFilterOperation.of(), "find-me");
+      FilterExpression expression = ImmutableFilterExpression.of(filterPath, condition, 0);
+
+      String cql =
+          "SELECT key, leaf, text_value, dbl_value, bool_value, p0, p1, p2, p3, WRITETIME(leaf) FROM %s WHERE key = ?";
+      ValidatingDataStore.QueryAssert cqlAssert =
+          withQuery(TABLE, cql)
```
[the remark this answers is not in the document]
dimas-b: The test code should be fixed in #1125. If it looks ok from your POV, please feel free to merge it and rebase this PR to take advantage of the fixes in `ValidatingDataStore`.

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
ivansenic: Eric is on the holiday, I would also vote for the removing the empty nodes, OK?

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/resolver/impl/SubDocumentsResolver.java
```diff
@@ -0,0 +1,138 @@
...
+      List<String> subDocumentPath,
+      ExecutionContext context,
+      boolean splitOnSubPathKeys) {
+    this.expression = expression;
+    this.context = createContext(context, subDocumentPath);
+    this.queryBuilder = new SubDocumentSearchQueryBuilder(documentId, subDocumentPath);
+    // key depth explained:
+    //  - one extra for the document id
+    //  - one extra on size if we need to split based on keys ids after sub-document path
+    this.keyDepth = subDocumentPath.size() + (splitOnSubPathKeys ? 2 : 1);
```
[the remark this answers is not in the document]
ivansenic: I can remove for now, this was made to support possible changes that are not moved for the V3.

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/resolver/impl/SubDocumentsResolver.java
```diff
@@ -0,0 +1,138 @@
...
+                  query,
+                  configuration.getSearchPageSize(),
+                  paginator.getCurrentDbPageState(),
+                  context);
+            })
+        .filter(
+            document -> {
+              Map<String, EvalRule<FilterExpression>> rules = EvalEngine.booleanRules();
+              rules.put(FilterExpression.EXPR_TYPE, new RawDocumentEvalRule(document));
+              return EvalEngine.evaluate(expression, rules);
```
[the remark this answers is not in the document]
ivansenic: Yea, this was intentionally changed:

1. To support multi field search (yes the `DocumentSearchService` does support it in the current state, we block from the outisde)
2. To be able to resolve ORs right away or actually any kind of the expression (`$not` would be also supported right away)

If you think about this, it's the same cost as loading the whole document, imo if search has same latency as the get, we should be fine, right. And we would be adding support for any kind of expression.

Furthermore, if there is sub-path, loading is only going for that sub-path, thus we never load more than we would actually need. 

Plus, since this is all in the same partition, loading all once could be even more performant that going for the 2 network hops to get first the row to evaluate and then the needed fields/structure.

We can discuss this, but I am thinking there are multiple benefits without actually degrading anything here..

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
ivansenic: This was changed in the error messages refactoring, now we hit that new path.. Wanna update to the old style of the error message?

restapi/src/test/java/io/stargate/web/docsapi/service/query/DocumentSearchServiceTest.java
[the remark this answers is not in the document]
ivansenic: Done.

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
dimas-b: ok, let's remove

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/resolver/impl/SubDocumentsResolver.java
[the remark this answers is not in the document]
dimas-b: yeah, let's keep the code lean and remove `splitOnSubPathKeys` for now.

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/resolver/impl/SubDocumentsResolver.java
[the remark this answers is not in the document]
dimas-b: > Furthermore, if there is sub-path, loading is only going for that sub-path, thus we never load more than we would actually need.

Good point. I missed it :facepalm:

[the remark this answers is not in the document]
dimas-b: Current approach LGTM. Resolving this comment.

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
dimas-b: The new message is fine from my POV. I meant quoting `"exists"` in this case, otherwise it makes parsing the English text harder, IMHO :)

dimas-b (approved):

ivansenic:
@dimas-b Perfect, I am adding one INT test for the `$or` and finalizing this.

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/resolver/impl/SubDocumentsResolver.java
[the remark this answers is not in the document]
ivansenic: Done.

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
ivansenic: Done.

ivansenic:
@dimas-b There is one more issue, that's the auth for the dead leaves removal.. Can we quickly agree here how to handle?

I would advise trying to auth here only if there is a need to do a removal.. If the auth is success then removal is done, otherwise it's skipped. This way read only users would be able to read here, which should be good imo. WDYT?

dimas-b:
> There is one more issue, that's the auth for the dead leaves removal.. Can we quickly agree here how to handle?
> 
> I would advise trying to auth here only if there is a need to do a removal.. If the auth is success then removal is done, otherwise it's skipped. This way read only users would be able to read here, which should be good imo. WDYT?

Makes sense, let's go with this approach :+1:

Issue #1082, opened 2021-07-01
GraphQL schema-first: allow entity field in delete response payload


olim7t:
Currently we only allow the `applied` field.

But a conditional delete can return data:
```
cqlsh:test> select * from "Foo";

 k | cc | v
---+----+---
 1 |  1 | 1
 1 |  2 | 2

(2 rows)
cqlsh:test> delete from "Foo" where k = 1 and cc = 1 if v = 2;

 [applied] | v
-----------+---
     False | 1
```
We should allow the entity field, if the client wants to inspect the conflicting data.

olim7t:
I pushed a branch with a fix: `gql_delete_entity`. Not opening the PR yet because it is based on #1083.

PR #1083, opened 2021-07-02
Add `@atomic` directive

olim7t:


graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/MutationFetcher.java
```diff
@@ -32,6 +49,162 @@ protected MutationFetcher(MutationModelT model, MappingModel mappingModel) {
...
+      Parameters parameters,
+      Exception buildException,
+      DataFetchingEnvironment environment,
+      StargateGraphqlContext context) {
+    StargateGraphqlContext.BatchContext batchContext = context.getBatchContext();
+    if (buildException != null) {
+      batchContext.setExecutionResult(buildException);
+    } else if (!batchContext.setParameters(parameters)) {
+      batchContext.setExecutionResult(
+          new GraphQLException(
```
mpenick, line 108: Is consistency level the only parameter that could be different?

the `cqlfirst` implementation uses the following error. Is this different?

````
  buildException =
          new GraphQLException("options can only de defined once in an @atomic mutation selection")

```

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/MutationPayload.java
```diff
@@ -0,0 +1,74 @@
...
+import io.stargate.db.query.BoundQuery;
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Function;
+
+/**
+ * Abstract the queries needed to process a mutation operation, and how the results will be turned
+ * into a GraphQL response.
+ */
+class MutationPayload<ResultT> {
```
mpenick, line 28: Does this need to be generic? As far as I can tell, all the parameterized inheritors use `DataFetcherResult<Object>`.

testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/DeleteCustomConditionsTest.java
```diff
@@ -147,7 +147,7 @@ public void deleteWithGreaterThanPredicate() {
         CLIENT.executeKeyspaceQuery(KEYSPACE, "mutation { deleteFooGT(pk: 1, v: 100) }");
 
     // then
-    assertThat(JsonPath.<Boolean>read(response, "$.deleteFooGT")).isTrue();
+    assertThat(JsonPath.<Boolean>read(response, "$.deleteFooGT")).isFalse();
```
dougwettlaufer, line 150: Why did this change?

olim7t: Yeah that one is puzzling to me: `false` is the right answer, because we execute `DELETE ... IF v > 100` and the value is 100. The mutation is not applied, as evidenced by the fact that the row still exists (checked on the next line).

However the response is indeed `true` on `master`, so there must be a production bug there. I'll check.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/UpdateFetcher.java
```diff
@@ -200,4 +161,87 @@ protected Object get(DataFetchingEnvironment environment, StargateGraphqlContext
...
+            primaryKey,
+            Scope.MODIFY,
+            SourceAPI.GRAPHQL);
+  }
+
+  private Function<List<MutationResult>, DataFetcherResult<Object>> getResultBuilder(
+      DataFetchingFieldSelectionSet selectionSet, DataFetchingEnvironment environment) {
+    return queryResults -> {
+      DataFetcherResult.Builder<Object> result = DataFetcherResult.newResult();
+      assert queryResults.size() == 1;
```
dougwettlaufer, line 206: Is it possible to share anything in here with the DeleteFetcher? Seems like there's a lot of similarities.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/UpdateFetcher.java
```diff
@@ -200,4 +161,87 @@ protected Object get(DataFetchingEnvironment environment, StargateGraphqlContext
...
   }
+
+  private AbstractBound<?> buildUpdateQuery(
+      EntityModel entityModel,
+      Collection<ValueModifier> modifiers,
+      List<BuiltCondition> whereConditions,
+      List<BuiltCondition> ifConditions,
+      Optional<Long> timestamp,
+      StargateGraphqlContext context) {
+    AbstractBound<?> query =
```
dougwettlaufer, line 172: Couldn't we just return here?

testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/DeleteCustomConditionsTest.java
olim7t: OK I see. In `DeleteFetcher` on `master`:
```java
boolean applied = !model.ifExists() || resultSet.one().getBoolean("[applied]");
```
This doesn't take into account `IF` conditions (the line was written at a time when `ifExists` was the only possible condition, and wasn't adapted later).

This got fixed as part of the larger refactoring in this PR.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/MutationFetcher.java
olim7t: Yes, the CLs are the only ones that can change, and there is no `options` objects in this version of the API so I changed the message so that it would be less confusing.

The only other mutation option in GraphQL-first is TTL, and it's appended directly in the CQL query string, so it can actually be different across queries in the same batch.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/MutationPayload.java
olim7t: Fair point. And this also applies to `MutationFetcher` itself.

`ResultT` made sense in the beginning because the mutations had specific return types. But as we added more ways to map results, `Object` is convenient for flexibility.

I'll make the change.

dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/UpdateFetcher.java
olim7t: Right, even more so after my changes for #1082. I'm adding a todo item to make those changes in that PR.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/UpdateFetcher.java
olim7t: 👍 
Refactoring mishap

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/MutationPayload.java
olim7t: Actually, there's one argument to keep the code generic: I think the CQL-first fetchers could be retrofitted to use it too, and they have more specific return types.

PR #1086, opened 2021-07-02
closes #1085: cassandra-4 move driver to 4.0-rc2

### LINKED ISSUE #1085: Move persistence-cassandra-4.0 to trunk branch until there is a GA release
This will make driver and INT test be up to date with the 4.0 main branch.

ivansenic:
The driver version `rc2` uses `com.datastax.cassandra » cassandra-driver-core (optional) ` in version `3.11.0`. Everything else should be clear.

### DIFF
--- a/testing/pom.xml
+++ b/testing/pom.xml
@@ -391,7 +391,7 @@
                 </goals>
                 <configuration>
                   <systemPropertyVariables>
-                    <ccm.version>trunk</ccm.version>
+                    <ccm.version>4.0-rc2</ccm.version>
                     <stargate.test.nodes>1</stargate.test.nodes>
                   </systemPropertyVariables>
                   <failIfNoTests>true</failIfNoTests>
--- a/testing/pom.xml
+++ b/testing/pom.xml
@@ -391,7 +391,7 @@
                 </goals>
                 <configuration>
                   <systemPropertyVariables>
-                    <ccm.version>4.0-beta4</ccm.version>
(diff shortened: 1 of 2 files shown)


testing/pom.xml
```diff
@@ -391,7 +391,7 @@
                 </goals>
                 <configuration>
                   <systemPropertyVariables>
-                    <ccm.version>4.0-beta4</ccm.version>
+                    <ccm.version>trunk</ccm.version>
```
dougwettlaufer, line 394: Shouldn't this still match what's in the persistent module?

ivansenic: I think that idea that Pushkala had was to test against the latest commits in the Cassandra repo in order to catch things early.. But I do agree, we could go with the driver latest as well.. 

The build does fail with the setup I provided.

ivansenic: OK so @dougwettlaufer was correct, versions must be the same:

```
Step #4 - "cassandra-4.0": Caused by: java.lang.IllegalArgumentException: Invalid version number: trunk
Step #4 - "cassandra-4.0": 	at com.datastax.oss.driver.api.core.Version.parse(Version.java:95)
Step #4 - "cassandra-4.0": 	at com.datastax.oss.driver.api.testinfra.ccm.CcmBridge.<clinit>(CcmBridge.java:73)
```

As there is no version for `trunk`, let's go with `rc-2` for now.

dimas-b (approved):

PR #1089, opened 2021-07-06
closes #1088: fixed comparing filters returning opposite results

### LINKED ISSUE #1088: Comparing filter operations tests wrongly
The doc api filter operations `$lt`, `$lte` `$gt` and `$gte` are testing wrongly and should provide actually opposite results :exploding_head: . We did not catch this until now as their tests are currently not used at all, as these are all Cassandra predicates that get executed directly on the database. Again, these buggy methods are not on any path currently. 

However, we should fix this asap for the sake of future work. Unfortunately, all unit tests were also written wrongly, there was a mix between filter value and actual database value.

ivansenic:
I have no idea how did this bug got into both tests and code.. Thank god this is not used atm, as there is a Cassandra predicate for those and their test method is not needed atm.. Also the reason INT tests are green.. 

So to clarify, first parameter is the filter value, second parameter is the database value. Op should make sure that the DB value is greater (or equal) or less (or equal) than the filter value!

Note: second commit is IDE based refactoring of the `test` methods parameters order.

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/impl/LteFilterOperation.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/impl/LteFilterOperation.java
@@ -46,6 +46,6 @@ public abstract class LteFilterOperation extends NotNullValueFilterOperation {
   /** {@inheritDoc} */
   @Override
   public boolean isSatisfied(int compareValue) {
-    return compareValue >= 0;
+    return compareValue <= 0;
   }
 }
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/impl/LteFilterOperation.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/impl/LteFilterOperation.java
@@ -46,6 +46,6 @@ public abstract class LteFilterOperation extends NotNullValueFilterOperation {
   /** {@inheritDoc} */
   @Override
   public boolean isSatisfied(int compareValue) {
-    return compareValue <= 0;
+    return compareValue >= 0;
   }
 }
--- a/restapi/src/test/java/io/stargate/web/docsapi/service/query/filter/operation/impl/GtFilterOperationTest.java
+++ b/restapi/src/test/java/io/stargate/web/docsapi/service/query/filter/operation/impl/GtFilterOperationTest.java
@@ -40,17 +40,17 @@ class GtFilterOperationTest {
     }
 
     @Test
-    public void stringGreater() {
+    public void stringLess() {
       boolean result = gt.test("filterValue", "aaa");
 
-      assertThat(result).isTrue();
+      assertThat(result).isFalse();
     }
 
     @Test
-    public void stringLess() {
+    public void stringGreater() {
       boolean result = gt.test("filterValue", "www");
 
-      assertThat(result).isFalse();
+      assertThat(result).isTrue();
     }
 
     @Test
@@ -71,17 +71,17 @@ class GtFilterOperationTest {
     }
 
     @Test
-    public void booleanGreater() {
+    public void booleanLess() {
       boolean result = gt.test(true, false);
 
-      assertThat(result).isTrue();
+      assertThat(result).isFalse();
     }
 
     @Test
-    public void booleanLess() {
+    public void booleanGreater() {
       boolean result = gt.test(false, true);
 
-      assertThat(result).isFalse();
+      assertThat(result).isTrue();
     }
 
     @Test
@@ -99,17 +99,17 @@ class GtFilterOperationTest {
     }
 
     @Test
-    public void numbersGreater() {
+    public void numbersLess() {
       boolean result = gt.test(22.1d, 22d);
 
-      assertThat(result).isTrue();
+      assertThat(result).isFalse();
     }
 
     @Test
-    public void numbersLess() {
+    public void numbersGreater() {
       boolean result = gt.test(21.9d, 22d);
 
-      assertThat(result).isFalse();
+      assertThat(result).isTrue();
     }
 
     @Test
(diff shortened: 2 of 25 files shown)


restapi/src/test/java/io/stargate/web/docsapi/service/query/filter/operation/impl/GtFilterOperationTest.java
```diff
@@ -40,17 +40,17 @@ public void stringEquals() {
     }
 
     @Test
-    public void stringGreater() {
+    public void stringLess() {
       boolean result = gt.test("filterValue", "aaa");
 
-      assertThat(result).isTrue();
+      assertThat(result).isFalse();
```
dimas-b, line 46: nit: WDYT about  `assertThatgt.test("filterValue", "aaa")).isFalse()`... I find it a bit hard to read through the `result` variable indirection.

restapi/src/test/java/io/stargate/web/docsapi/service/query/filter/operation/impl/LtFilterOperationTest.java
```diff
@@ -40,17 +40,17 @@ public void stringEquals() {
     }
 
     @Test
-    public void stringGreater() {
+    public void stringLess() {
       boolean result = lt.test("filterValue", "aaa");
 
-      assertThat(result).isFalse();
+      assertThat(result).isTrue();
```
dimas-b, line 46: I'm not sure this change correct. `filterValue` is indeed _greater than_ `aaa` lexicographically.

dimas-b (changes requested):

restapi/src/test/java/io/stargate/web/docsapi/service/query/filter/operation/impl/LtFilterOperationTest.java
ivansenic: Yes that's the point:

* `aaa` is value from the database
* `filterValue` is what's specified from the user (eq. `smth: { lt: 'filterValue'}}
* `aaa` should pass as it's less than `filterValue`

Right?

dimas-b: I think I see the problem now. `filterValue` comes from the HTTP request as in `{field: {$lt: "filterValue"}}` as we compare it to "dbValue", right?

IMHO, it would be preferable to reverse the order of parameters to the `test` method, i.e. define it as `test(String dbValue, String filterValue)`. This was the operation will be consistent with `Comparator.compare(dbValue, filterValue))` and with the order of predicate args in the `where` clause at HTTP level. WDYT?

ivansenic: Hmm although this is gonna be a simple IDE refactoring, this will create a huge amount of changes. Are we sure we want this? And if it really gives us a better "understanding"?

dimas-b: From my POV it will clarity the meaning of those `FilterOperation` classes - closer match with API-level parameters.

A `{field: {$lt: "filterValue"}}` input predicate will translate into a `LtFilterOperation` object. In runtime the `LtFilterOperation.test()` method will be called with values like `test("dbValue", "filterValue")` - it's the same order or parameters as `field` and `filterValue` in the user input and the `true/false` result will be consistent with the contract of `Comparator.test("dbValue", "filterValue")`. WDYT?

ivansenic: Yep, I'll add as separate commit.

dimas-b: Thanks! Hopefully, in the end, the `isSatisfied()` implementations will not have to change wrt `master`

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/impl/LteFilterOperation.java
```diff
@@ -46,6 +46,6 @@ public FilterOperationCode getOpCode() {
   /** {@inheritDoc} */
   @Override
   public boolean isSatisfied(int compareValue) {
-    return compareValue <= 0;
+    return compareValue >= 0;
```
dimas-b: Is this change required after the paramerter reordering in the `test` method? I assumed we'd change the order of parameters when we call `Comparator.compare()` and leave the `isSatisfied()` methods as they are on `trunk`, i.e. following the contract of `Comparator` return values... WDYT?

dimas-b (changes requested):

dimas-b (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/impl/LteFilterOperation.java
ivansenic: Outdated, resolved.

restapi/src/test/java/io/stargate/web/docsapi/service/query/filter/operation/impl/GtFilterOperationTest.java
```diff
@@ -40,22 +40,22 @@ public void stringEquals() {
     }
 
     @Test
-    public void stringGreater() {
-      boolean result = gt.test("filterValue", "aaa");
+    public void stringLess() {
+      boolean result = gt.test("aaa", "filterValue");
```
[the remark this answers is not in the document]
ivansenic: Now, this is as it should be now.

restapi/src/test/java/io/stargate/web/docsapi/service/query/filter/operation/impl/GtFilterOperationTest.java
ivansenic: I think now it's OK to read after the change of params and the method name updates..

PR #1091, opened 2021-07-07
Support merging multiple result sets in QueryExecutor

dimas-b:
Contributes to #828 

-----

This PR adds support to `QueryExecutor` for running more than one query (against the same table) while merging the result sets on Document identity columns (C* primary key).

Paging state is extended to allow advancing pages in each of the underlying queries independently, as the iteration over the merged result set progresses.

This include minor changes to the handling of document pagination at the `RawDocument` level. In particular the `hasNext` flag is no longer necessary.

No functional changes are expected in the single query case.

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
@@ -16,19 +16,55 @@
 package io.stargate.web.docsapi.service;
 
 import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
+import io.stargate.db.PagingPosition.ResumeMode;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.stream.Collectors;
 
 /**
  * A utility class for combining one or more per-query paging state buffers into one byte buffer.
+ *
+ * <p>It handles the edge cases of {@code null} and {@link #EXHAUSTED_PAGE_STATE "exhausted"} query
+ * states.
  */
-public class CombinedPagingState {
+public class CombinedPagingState implements PagingStateSupplier {
 
-  private CombinedPagingState() {
-    // private constructor to prevent instantiation
+  /**
+   * A special paging state buffer that indicates that the query reach the end of its result set.
+   *
+   * <p>Normally individual queries return a {@code null} paging state when they are exhausted.
+   * However, if case of merging two or more results sets we need to distinguish queries that have
+   * not yet produced any rows (also having a {@code null} initial paging state) from queries that
+   * reach the end of their result sets.
+   */
+  public static final ByteBuffer EXHAUSTED_PAGE_STATE = ByteBuffer.allocate(0);
+
+  private final List<PagingStateSupplier> pagingState;
+
+  CombinedPagingState(List<PagingStateSupplier> pagingState) {
+    this.pagingState = pagingState;
+  }
+
+  public static boolean isExhausted(ByteBuffer pagingState) {
+    return pagingState != null && pagingState.remaining() == 0;
+  }
+
+  @Override
+  public ByteBuffer makePagingState(ResumeMode resumeMode) {
+    List<ByteBuffer> pagingStateBuffers =
+        pagingState.stream().map(s -> s.makePagingState(resumeMode)).collect(Collectors.toList());
+
+    // An empty paging state means the query was exhausted during previous execution.
+    // A null paging state mean the query has not been executed yet.
+    if (pagingStateBuffers.stream().allMatch(CombinedPagingState::isExhausted)) {
+      // Return a null overall paging state if all queries have been exhausted.
+      return null;
+    }
+
+    return CombinedPagingState.serialize(pagingStateBuffers);
   }
 
   public static ByteBuffer serialize(List<ByteBuffer> nestedStates) {
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
@@ -18,22 +18,20 @@ package io.stargate.web.docsapi.service;
 import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 
+/**
+ * A utility class for combining one or more per-query paging state buffers into one byte buffer.
+ */
 public class CombinedPagingState {
 
-  private final List<ByteBuffer> nestedStates;
-
-  private CombinedPagingState(List<ByteBuffer> nestedStates) {
-    this.nestedStates = nestedStates;
+  private CombinedPagingState() {
+    // private constructor to prevent instantiation
   }
 
-  public static CombinedPagingState of(List<ByteBuffer> nestedStates) {
-    return new CombinedPagingState(nestedStates);
-  }
-
-  public ByteBuffer serialize() {
+  public static ByteBuffer serialize(List<ByteBuffer> nestedStates) {
     if (nestedStates.stream().allMatch(Objects::isNull)) {
       return null;
     }
@@ -42,9 +40,9 @@ public class CombinedPagingState {
       return nestedStates.get(0);
     }
 
-    int toAllocate = 4; // size of int for element count
+    int toAllocate = Integer.BYTES; // int for element count
     for (ByteBuffer state : nestedStates) {
-      toAllocate += 4; // size of int for element size
+      toAllocate += Integer.BYTES; // int for element size
       toAllocate += state == null ? 0 : state.remaining();
     }
 
@@ -66,7 +64,16 @@ public class CombinedPagingState {
     return result;
   }
 
-  public static CombinedPagingState deserialize(int expectedSize, ByteBuffer data) {
+  /**
+   * Breaks down a combined paging state into per-query paging state buffers.
+   *
+   * @param expectedSize the expected number of nested paging states. This number servers as
+   *     simplistic sanity check to ensure we have as many queries using the paging states as we
+   *     have nested elements. Also it is used to support rolling upgrades where we have to be able
+   *     to correctly interpret one-element paging states in the old format.
+   * @param data serialized form of the combined paging state
+   */
+  public static List<ByteBuffer> deserialize(int expectedSize, ByteBuffer data) {
     if (expectedSize <= 0) {
       throw new IllegalArgumentException("Invalid paging state size: " + expectedSize);
     }
@@ -77,11 +84,14 @@ public class CombinedPagingState {
         buffers.add(null);
       }
 
-      return of(buffers);
+      return Collections.unmodifiableList(buffers);
     }
 
+    // The special case for one nested element is mostly to support rolling upgrades where older
+    // nodes may be producing one-elelement paging states, while newer (upgraded) nodes might have
+    // to interpret those paging states.
     if (expectedSize == 1) {
-      return of(ImmutableList.of(data));
+      return ImmutableList.of(data);
     }
 
     if (data.remaining() < 4) {
@@ -113,10 +123,6 @@ public class CombinedPagingState {
       nested.add(element);
     }
 
-    return of(nested);
-  }
-
-  public List<ByteBuffer> nested() {
-    return nestedStates;
+    return Collections.unmodifiableList(nested);
   }
 }
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.web.docsapi.service;
+
+import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+public class CombinedPagingState {
+
+  private final List<ByteBuffer> nestedStates;
+
+  private CombinedPagingState(List<ByteBuffer> nestedStates) {
+    this.nestedStates = nestedStates;
+  }
+
+  public static CombinedPagingState of(List<ByteBuffer> nestedStates) {
+    return new CombinedPagingState(nestedStates);
+  }
+
+  public ByteBuffer serialize() {
+    if (nestedStates.stream().allMatch(Objects::isNull)) {
+      return null;
+    }
+
+    if (nestedStates.size() == 1) {
+      return nestedStates.get(0);
+    }
+
+    int toAllocate = 4; // size of int for element count
+    for (ByteBuffer state : nestedStates) {
+      toAllocate += 4; // size of int for element size
+      toAllocate += state == null ? 0 : state.remaining();
+    }
+
+    ByteBuffer result = ByteBuffer.allocate(toAllocate);
+
+    result.putInt(nestedStates.size());
+    for (ByteBuffer state : nestedStates) {
+      if (state != null) {
+        ByteBuffer nested = state.slice();
+
+        result.putInt(nested.remaining());
+        result.put(nested);
+      } else {
+        result.putInt(-1); // no state
+      }
+    }
+
+    result.flip();
+    return result;
+  }
+
+  public static CombinedPagingState deserialize(int expectedSize, ByteBuffer data) {
+    if (expectedSize <= 0) {
+      throw new IllegalArgumentException("Invalid paging state size: " + expectedSize);
+    }
+
+    if (data == null) {
+      ArrayList<ByteBuffer> buffers = new ArrayList<>(expectedSize);
+      for (int i = 0; i < expectedSize; i++) {
+        buffers.add(null);
+      }
+
+      return of(buffers);
+    }
+
+    if (expectedSize == 1) {
+      return of(ImmutableList.of(data));
+    }
+
+    if (data.remaining() < 4) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Invalid paging state: unable to read size, available bytes: %d", data.remaining()));
+    }
+
+    int count = data.getInt();
+    if (expectedSize != count) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Invalid paging state: expected element count: %d, actual: %d", expectedSize, count));
+    }
+
+    ArrayList<ByteBuffer> nested = new ArrayList<>(count);
+    while (count-- > 0) {
+      int size = data.getInt();
+
+      ByteBuffer element;
+      if (size >= 0) {
+        element = data.slice();
+        element.limit(size);
+        data.position(data.position() + size);
+      } else {
+        element = null;
+      }
+
+      nested.add(element);
+    }
+
+    return of(nested);
+  }
+
+  public List<ByteBuffer> nested() {
+    return nestedStates;
+  }
+}
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
@@ -130,7 +130,7 @@ public class CombinedPagingState implements PagingStateSupplier {
       return ImmutableList.of(data);
     }
 
-    if (data.remaining() < 4) {
+    if (data.remaining() < Integer.BYTES) {
       throw new IllegalArgumentException(
           String.format(
               "Invalid paging state: unable to read size, available bytes: %d", data.remaining()));
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/PagingStateSupplier.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/PagingStateSupplier.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.web.docsapi.service;
+
+import io.stargate.db.PagingPosition.ResumeMode;
+import java.nio.ByteBuffer;
+
+public abstract class PagingStateSupplier {
+
+  public abstract boolean hasPagingState();
+
+  public abstract ByteBuffer makePagingState(ResumeMode resumeMode);
+}
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/PagingStateSupplier.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/PagingStateSupplier.java
@@ -18,9 +18,22 @@ package io.stargate.web.docsapi.service;
 import io.stargate.db.PagingPosition.ResumeMode;
 import java.nio.ByteBuffer;
 
-public abstract class PagingStateSupplier {
+/**
+ * An abstraction layer for sources of a query paging state.
+ *
+ * <p>In general a paging state may come from:
+ * <li>a byte buffer, which in turn is a paging state returned from the previous query execution
+ *     that was not utilized during the current processing round and hence can be reused "as is"
+ *     regardless of {@link ResumeMode ResumeMode}.
+ * <li>a row that was received from the current (fresh) result set page.
+ */
+public interface PagingStateSupplier {
 
-  public abstract boolean hasPagingState();
+  /** Constructs paging state for the specified {@link ResumeMode ResumeMode}. */
+  ByteBuffer makePagingState(ResumeMode resumeMode);
 
-  public abstract ByteBuffer makePagingState(ResumeMode resumeMode);
+  /** Creates a {@link PagingStateSupplier} for a pre-built, fixed paging state buffer. */
+  static PagingStateSupplier fixed(ByteBuffer pagingState) {
+    return resumeMode -> pagingState;
+  }
 }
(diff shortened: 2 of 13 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
```diff
@@ -57,30 +74,109 @@ public QueryExecutor(DataStore dataStore) {
...
+      }
+    }
+
+    if (pathColumns == null) {
+      throw new IllegalArgumentException("No query provided");
+    }
+
+    Comparator<DocProperty> comparator = Comparator.comparing(DocProperty::comparableKey);
+    for (Column column : pathColumns) {
+      comparator = comparator.thenComparing(p -> p.keyValue(column));
```
ivansenic, line 123: why we are including this extra? Isn't enough that the order is per key? Why do we care for the other sub-document props here? Who cares how are the sub-documents of the single document order, that's all on the same partition?

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
```diff
@@ -116,20 +212,37 @@ public QueryExecutor(DataStore dataStore) {
...
+    String id = row.getString("key");
+    Builder<String> docKey = ImmutableList.builder();
+    for (Column c : keyColumns) {
+      docKey.add(Objects.requireNonNull(row.getString(c.name())));
+    }
+
+    return new Accumulator(id, comparator, docKey.build(), property);
+  }
+
+  private Iterable<DocProperty> properties(
```
ivansenic, line 227: can you explain this a bit more to me please? also some kind of inline docs would be so helpful to understand what s going on, nobody will ever be able to understand this except the writer :)

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
```diff
@@ -139,39 +252,51 @@ public DataStore getDataStore() {
   private class Accumulator {
```
ivansenic, line 252: could be worth to extract this to a dedicated file now.. Stands true for the other private classes as well..

ivansenic (changes requested):
It's very hard to follow the changes in the `QueryExecutor` without any inline comments or PR description. I did a first look on the whole code, but I would need some guidance to finalize the review..

restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
```diff
@@ -0,0 +1,122 @@
...
+      } else {
+        result.putInt(-1); // no state
+      }
+    }
+
+    result.flip();
+    return result;
+  }
+
+  public static CombinedPagingState deserialize(int expectedSize, ByteBuffer data) {
```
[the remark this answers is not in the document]
dimas-b: Yes, it probably makes sense to make it `static` too... will do.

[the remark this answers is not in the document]
dimas-b: The single element case is not really a workaround, but more of a backward compatibility logic for supporting rolling upgrades where single query paging state may be generated by older code, but interpreted by newer code.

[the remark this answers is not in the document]
dimas-b: `expectedSize` is a sanity check. Ideally it should be something like a hash of the whole query plan, but doing that seemed like an overkill at this point.

restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
```diff
@@ -0,0 +1,122 @@
...
+              "Invalid paging state: expected element count: %d, actual: %d", expectedSize, count));
+    }
+
+    ArrayList<ByteBuffer> nested = new ArrayList<>(count);
+    while (count-- > 0) {
+      int size = data.getInt();
+
+      ByteBuffer element;
+      if (size >= 0) {
+        element = data.slice();
```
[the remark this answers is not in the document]
dimas-b: sure

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
```diff
@@ -57,30 +74,109 @@ public QueryExecutor(DataStore dataStore) {
...
+  private List<Column> idColumns(int keyDepth, List<BoundQuery> queries) {
+    List<Column> idColumns = null;
+    for (BoundQuery query : queries) {
+      BuiltSelect select = (BuiltSelect) query.source().query();
+      if (keyDepth < 1 || keyDepth > select.table().primaryKeyColumns().size()) {
+        throw new IllegalArgumentException("Invalid document identity depth: " + keyDepth);
+      }
+
+      if (idColumns == null) {
+        idColumns = select.table().primaryKeyColumns().subList(0, keyDepth);
```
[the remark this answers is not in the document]
dimas-b: All queries must reference the same table and have the same selectors. I'll add a check for that. Thanks for flagging this!

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
```diff
@@ -57,30 +74,109 @@ public QueryExecutor(DataStore dataStore) {
...
+      int pageSize,
+      CombinedPagingState pagingState,
+      ExecutionContext context) {
+    List<Flowable<DocProperty>> flows = new ArrayList<>(queries.size());
+    int idx = 0;
+    for (BoundQuery query : queries) {
+      final int finalIdx = idx++;
+      ByteBuffer queryPagingState = pagingState.nested().get(finalIdx);
+      if (queryPagingState != null) {
+        queryPagingState = queryPagingState.slice();
```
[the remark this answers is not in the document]
dimas-b: good point.

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
dimas-b: We need strong order to be able to identify duplicate rows without loading all of the data into memory. Currently we assume all queries have the same sorting order (according to `pNNN` clustering columns). Then, we can merge the result sets by comparing only the top-level elements.

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
dimas-b: I'll add comments

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
dimas-b: I'm not sure about that. The logic encapsulated into `Accumulator` is tightly coupled to the way our Rx pipeline is constructed, in particular to the use of the `scan` operator. Let's keep it private for now. I'd be open to factoring it out to a top-level class later, though.

I will move other general-purpose helper classes to top level.

restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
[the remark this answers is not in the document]
dimas-b: I made the `Paginator`'s buffer read only instead in fe2d63fd36a102c70a867ac58b63bb72d447a6cb, WDYT? A slices or a R/O buffer will be read-only too.

restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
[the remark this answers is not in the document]
dimas-b: done in fe2d63fd36a102c70a867ac58b63bb72d447a6cb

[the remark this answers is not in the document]
dimas-b: comments added in fe2d63fd36a102c70a867ac58b63bb72d447a6cb

restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
```diff
@@ -0,0 +1,122 @@
...
+  public ByteBuffer serialize() {
+    if (nestedStates.stream().allMatch(Objects::isNull)) {
+      return null;
+    }
+
+    if (nestedStates.size() == 1) {
+      return nestedStates.get(0);
+    }
+
+    int toAllocate = 4; // size of int for element count
```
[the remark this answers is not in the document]
dimas-b: done in fe2d63fd36a102c70a867ac58b63bb72d447a6cb

restapi/src/test/java/io/stargate/db/datastore/AlphabeticalOrderPartitionKeyDecorator.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package io.stargate.db.datastore;
+
+import io.stargate.db.ComparableKey;
+import io.stargate.db.RowDecorator;
+import io.stargate.db.schema.Column;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class AlphabeticalOrderPartitionKeyDecorator implements RowDecorator {
```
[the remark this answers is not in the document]
dimas-b: Added in 274d75aece0c491c4c367d30c7b246ae6280d036

restapi/src/test/java/io/stargate/db/datastore/AlphabeticalOrderPartitionKeyDecorator.java
```diff
@@ -0,0 +1,44 @@
...
+  private final List<Column> partitionKeyColumns;
+
+  public AlphabeticalOrderPartitionKeyDecorator(List<Column> allColumns) {
+    this.partitionKeyColumns =
+        allColumns.stream().filter(Column::isPartitionKey).collect(Collectors.toList());
+  }
+
+  @Override
+  public <T extends Comparable<T>> ComparableKey<T> decoratePartitionKey(Row row) {
+    StringBuilder decorated = new StringBuilder();
```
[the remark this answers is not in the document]
dimas-b: Thanks for the tip! Implemented in 274d75aece0c491c4c367d30c7b246ae6280d036

restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+public class CombinedPagingState {
+
+  private final List<ByteBuffer> nestedStates;
+
+  private CombinedPagingState(List<ByteBuffer> nestedStates) {
+    this.nestedStates = nestedStates;
```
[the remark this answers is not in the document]
dimas-b: I guess this is no longer necessary after the refactoring done in fe2d63fd36a102c70a867ac58b63bb72d447a6cb

restapi/src/main/java/io/stargate/web/docsapi/service/PagingStateSupplier.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.web.docsapi.service;
+
+import io.stargate.db.PagingPosition.ResumeMode;
+import java.nio.ByteBuffer;
+
+public abstract class PagingStateSupplier {
```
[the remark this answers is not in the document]
dimas-b: Added in 27dc6ca381a43d4285718bf58e20192ed7e40c4c

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
dimas-b: I've refactored all of the paging state handlers and suppliers in 27dc6ca381a43d4285718bf58e20192ed7e40c4c. Please review again.

restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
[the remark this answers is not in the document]
ivansenic: Perfect

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
ivansenic: Hmm very good point.. I did not think about the duplication of the rows from the different queries..

So how is this handled? Where are duplicates dismissed? I guess `orderedMerge` will not do that? Is this something we handle in the Accumulator..

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
```diff
@@ -183,34 +308,52 @@ public RawDocument toDoc() {
...
-      lastResultSet = other.lastResultSet;
+      if (rowComparator == null || lastRow == null) {
+        throw new IllegalStateException("Cannot compare");
+      }
+
+      DocProperty otherRow = other.lastRow;
+
+      if (rowComparator.compare(this.lastRow, otherRow) != 0) {
+        rows.add(otherRow);
+        this.lastRow = otherRow;
```
ivansenic, line 350: so is this correct, you'll always update the last row, no matter if the comparator returns `1` or `-1`? How's that?

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
```diff
@@ -219,15 +362,103 @@ private Accumulator combine(Accumulator buffer) {
...
+    public void track(DocProperty row) {
+      states.set(row.queryIndex(), row);
+    }
+
+    public List<PagingStateSupplier> slice() {
+      return new ArrayList<>(states);
+    }
+  }
+
+  /**
```
ivansenic, line 398: the comments here really helped :+1:

restapi/src/test/java/io/stargate/web/docsapi/service/CombinedPagingStateTest.java
```diff
@@ -0,0 +1,183 @@
...
+          new CombinedPagingState(ImmutableList.of(resumeMode -> null, resumeMode -> null));
+      assertThat(combined.makePagingState(ResumeMode.NEXT_PARTITION)).isNull();
+    }
+
+    @Test
+    void mixed() {
+      CombinedPagingState combined =
+          new CombinedPagingState(
+              ImmutableList.of(resumeMode -> null, resumeMode -> EXHAUSTED_PAGE_STATE));
+      assertThat(combined.makePagingState(ResumeMode.NEXT_PARTITION)).isNotNull();
```
ivansenic, line 78: is this correct considering the tests above and below?

restapi/src/test/java/io/stargate/web/docsapi/service/QueryExecutorTest.java
```diff
@@ -414,4 +422,160 @@ void testSubDocumentsPaged(int pageSize) {
...
+        executor.queryDocs(ImmutableList.of(q1, q2), pageSize, null, context).test();
+    test.assertNoErrors();
+
+    List<RawDocument> docs = test.values();
+
+    assertThat(docs).extracting(RawDocument::id).containsExactly("a", "b", "c", "d");
+    assertThat(docs.get(0).rows()).extracting(r -> r.getString("p0")).contains("x1", "x2");
+    assertThat(docs.get(1).rows()).extracting(r -> r.getString("p0")).contains("y2");
+    assertThat(docs.get(2).rows()).extracting(r -> r.getString("p0")).contains("x2", "x3");
+    assertThat(docs.get(3).rows()).extracting(r -> r.getString("p0")).contains("z1", "z2");
```
ivansenic, line 463: :partying_face:

restapi/src/test/java/io/stargate/web/docsapi/service/QueryExecutorTest.java
```diff
@@ -414,4 +422,160 @@ void testSubDocumentsPaged(int pageSize) {
       assertThat(docs).isEmpty();
     }
   }
+
+  @ParameterizedTest
+  @CsvSource({"1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "100"})
+  void testMergeByDocKey(int pageSize) {
```
ivansenic, line 428: can we have the same extended depth please?

ivansenic (commented):
This is really awesome work.. Well done @dimas-b 

Tests failing though :(

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
dimas-b: `orderedMerge` will merge source flows placing equal rows next to each other. `Accumulator.combine(...)` later will take only one of a group of equal rows.

restapi/src/test/java/io/stargate/web/docsapi/service/QueryExecutorTest.java
dimas-b: will add

dimas-b:
@ivansenic : Re: test failures: I had to undo (f0adb4940e9cd06e7504192d0f03ebd0bf213f04) the read-only ByteBuffer change that you previously suggested.

The reason is that C* (3.11 at least) appears to handle read-only byte buffers incorrectly in `FastByteOperations$UnsafeOperations.copy`. It gets the "address" as `0` and tries to copy memory from there, resulting in `SIGSEGV` :(

restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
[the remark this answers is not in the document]
dimas-b: fe2d63f had to be reverted, unfortunately: https://github.com/stargate/stargate/pull/1091#issuecomment-877358241

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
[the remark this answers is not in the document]
dimas-b: I've refactored this code  and added comments in the last round of updates. Please re-review.

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
```diff
@@ -57,30 +74,109 @@ public QueryExecutor(DataStore dataStore) {
...
         .filter(Accumulator::isComplete)
         .map(Accumulator::toDoc);
   }
 
+  private Comparator<DocProperty> rowComparator(List<BoundQuery> queries) {
+    List<Column> pathColumns = null;
+    for (BoundQuery query : queries) {
+      BuiltSelect select = (BuiltSelect) query.source().query();
+
+      if (pathColumns == null) {
```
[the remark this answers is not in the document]
dimas-b: I've refactored this code  and added comments in the last round of updates. Please re-review.

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
dimas-b: javadoc added

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
dimas-b: We only care about equality at this point.

I hope it's a bit clearer in the refactored code. Please see the latest commits.

restapi/src/test/java/io/stargate/web/docsapi/service/CombinedPagingStateTest.java
dimas-b: I think it is correct. If one of the results sets was never processed (`null`), while the other  is exhausted, we want a non-null paging state so that the user could ask for more data and cause the result set of the former query to be processed.

restapi/src/test/java/io/stargate/web/docsapi/service/QueryExecutorTest.java
dimas-b: tests added in the new set of commits

restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    // The special case for one nested element is mostly to support rolling upgrades where older
+    // nodes may be producing one-elelement paging states, while newer (upgraded) nodes might have
+    // to interpret those paging states.
+    if (expectedSize == 1) {
+      return ImmutableList.of(data);
+    }
+
+    if (data.remaining() < 4) {
```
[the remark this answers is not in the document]
dimas-b: thx! fixed in cf003f15866f12f2c303f583a85e18fcc2f931f6

restapi/src/main/java/io/stargate/web/docsapi/service/CombinedPagingState.java
```diff
@@ -0,0 +1,164 @@
...
+  @Override
+  public ByteBuffer makePagingState(ResumeMode resumeMode) {
+    List<ByteBuffer> pagingStateBuffers =
+        pagingState.stream().map(s -> s.makePagingState(resumeMode)).collect(Collectors.toList());
+
+    // An empty paging state means the query was exhausted during previous execution.
+    // A null paging state mean the query has not been executed yet.
+    if (pagingStateBuffers.stream().allMatch(CombinedPagingState::isExhausted)) {
+      // Return a null overall paging state if all queries have been exhausted.
+      return null;
```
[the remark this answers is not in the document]
dimas-b: The trick is that "exhausted" paging states may occurs as a result of a round trip through the HTTP-level paging state, in which case they will be separate buffers, empty, but not identical to `EXHAUSTED_PAGE_STATE`.

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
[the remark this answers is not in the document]
dimas-b: In the end I think slicing is easier to understand and safer. Paging state buffers may be read by different pieces of code, including C* code and our `QueryExecutor` code (for building the "combined" paging state). Slicing ensures that the buffer pointers we use are independent. Resetting should work, but then we have think about the correctness of the read/reset sequence of operations, which is a bit tricky in this context. I hope slicing does not add any measurable overhead... WDYT?

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
```diff
@@ -57,30 +76,120 @@ public QueryExecutor(DataStore dataStore) {
       int pageSize,
       ByteBuffer pagingState,
       ExecutionContext context) {
+    return queryDocs(keyDepth, ImmutableList.of(query), pageSize, pagingState, context);
+  }
+
+  /**
+   * Runs the provided queries in parallel and merges their result sets according to the selected
```
ivansenic, line 83: ohh this is so nice, we needed this kind of detailed explanation what s possible what not

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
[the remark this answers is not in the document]
ivansenic: Fine with me

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
```diff
@@ -116,20 +225,42 @@ public QueryExecutor(DataStore dataStore) {
     }
   }
 
-  private Iterable<Accumulator> seeds(
-      BoundQuery query, ResultSet rs, List<Column> keyColumns, ExecutionContext context) {
+  private Accumulator toSeed(
+      DocProperty property, Comparator<DocProperty> comparator, List<Column> keyColumns) {
+    Row row = property.row();
+    String id = row.getString("key");
```
ivansenic, line 231: nit: I think we should stop having the column names used by the DocsAPI directly as the literal, wanna use the constant value from the `io.stargate.web.docsapi.service.query.QueryConstants`?

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
```diff
@@ -219,15 +389,191 @@ private Accumulator combine(Accumulator buffer) {
...
+
+    private boolean isSelected(Column column) {
+      // An empty selection set means `*` (all columns)
+      return selectedColumns().isEmpty() || selectedColumns().contains(column);
+    }
+
+    /**
+     * Retrurn the list of columns whose values are used to distinguish one document from another.
+     */
+    private List<Column> docIdColumns(int keyDepth) {
```
ivansenic, line 549: could this be `@Value.Lazy` as well?

restapi/src/test/java/io/stargate/db/datastore/AlphabeticalOrderPartitionKeyDecorator.java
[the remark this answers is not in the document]
ivansenic: You see I missed that this is a test class, with the comment I get it immediately :smile:

restapi/src/test/java/io/stargate/web/docsapi/DocsApiTestSchemaProvider.java
```diff
@@ -83,7 +84,7 @@ public DocsApiTestSchemaProvider(int maxDepth, String keyspaceName, String colle
           ImmutableColumn.builder()
               .name(QueryConstants.P_COLUMN_NAME.apply(i))
               .type(Column.Type.Text)
-              .kind(PartitionKey)
+              .kind(Clustering)
```
ivansenic, line 87: :warning: Wait is this correct change? You are changing the schema here?

restapi/src/test/java/io/stargate/web/docsapi/service/CombinedPagingStateTest.java
ivansenic: OK, resolve.

ivansenic (changes requested):
As far as I am concerned we can merge this.. Awesome work @dimas-b, great testing.. There is only one change that worries me, thus set as *Request changes* until this is resolved...

restapi/src/test/java/io/stargate/web/docsapi/DocsApiTestSchemaProvider.java
dimas-b: No, I believe old code was a mistake here. `pNNN` columns are clustering keys, not partition keys.

The reason why I changed it now, is that it breakes the new validation logic in `QueryExecutor`.

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
dimas-b: sure, but I thought `@Value.Lazy` was not required in this case since this method is not a "getter" (it has an arg).

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
dimas-b: good point. will fix :+1:

dimas-b: fixed in ee5522221621015605a3fab3b196e37452a7843d

dimas-b:
Rebased to `master`

restapi/src/test/java/io/stargate/web/docsapi/DocsApiTestSchemaProvider.java
ivansenic: OK but once we redeploy this will only influence new tables right? What about the existing tables? If it brakes the validation logic will it not work in the existing tables?

dimas-b: Sorry, I'm not sure I understand... the `key` columns has always been the only partition key, the `pNNN` have always need clustering keys. This change affects only the test code. Production code is not affected and should work as before.

ivansenic: Fu*** I did not see this is a test class.. Sorry my bad :exploding_head:

ivansenic (approved):
Approved after all is resolved.

PR #1093, opened 2021-07-08
Add restriction on tables that aren't docs collections

### LINKED ISSUE #1052: Docs API getCollections should only return collections
The `io.stargate.web.docsapi.resources.CollectionsResource#getCollections` method is returning all tables in a keyspace rather than just the collections. Instead we should either make use of `io.stargate.web.docsapi.dao.DocumentDB#isDocumentsTable` or the internal schema cache.

EricBorczuk:
Resolves #1052

Don't allow users to see or delete tables that aren't documents collections

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
@@ -5,6 +5,7 @@ import static io.stargate.web.docsapi.resources.RequestToHeadersMapper.getAllHea
 import io.stargate.auth.Scope;
 import io.stargate.auth.SourceAPI;
 import io.stargate.auth.entity.ResourceKind;
+import io.stargate.db.schema.Keyspace;
 import io.stargate.db.schema.SchemaEntity;
 import io.stargate.db.schema.Table;
 import io.stargate.web.docsapi.dao.DocumentDB;
@@ -13,6 +14,7 @@ import io.stargate.web.docsapi.models.DocCollection;
 import io.stargate.web.docsapi.models.dto.CreateCollection;
 import io.stargate.web.docsapi.models.dto.UpgradeCollection;
 import io.stargate.web.docsapi.service.CollectionService;
+import io.stargate.web.docsapi.service.DocsSchemaChecker;
 import io.stargate.web.models.Error;
 import io.stargate.web.models.ResponseWrapper;
 import io.stargate.web.resources.AuthenticatedDB;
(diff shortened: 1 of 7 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/DocsSchemaChecker.java
```diff
@@ -24,24 +24,30 @@ private void clearCacheOnSchemaChange(DocumentDB db) {
...
         ImmutableKeyspaceAndTable.builder().keyspace(keyspace).table(table).build();
     clearCacheOnSchemaChange(db);
-    validatedDocCollections.computeIfAbsent(keyspaceAndTable, ks -> performValidityCheck(ks, db));
+    return validatedDocCollections.computeIfAbsent(
+        keyspaceAndTable, ks -> performValidityCheck(ks, db, throwOnFail));
   }
 
-  private boolean performValidityCheck(ImmutableKeyspaceAndTable keyspaceAndTable, DocumentDB db) {
+  private boolean performValidityCheck(
+      ImmutableKeyspaceAndTable keyspaceAndTable, DocumentDB db, boolean throwOnFail) {
```
dougwettlaufer, line 36: I'm not sure I understand the point of the `throwOnFail` flag (which in general boolean flags are a code smell). Why not have the caller catch that exception or add another method?

restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
```diff
@@ -83,20 +87,25 @@ public Response getCollections(
...
-                  authenticatedDB.getAuthenticationSubject(),
+                  docDB.getAuthenticationSubject(),
                   Collections.singletonList(namespace),
                   tables.stream().map(SchemaEntity::name).collect(Collectors.toList()),
                   SourceAPI.REST,
                   ResourceKind.TABLE);
 
           List<DocCollection> result =
               tables.stream()
+                  .filter(
```
dougwettlaufer, line 107: Should we filter first and then authenticate? Doesn't seem like it would make sense to return a 403 for a keyspace that isn't being returned anyway.

restapi/src/main/java/io/stargate/web/docsapi/service/DocsSchemaChecker.java
EricBorczuk: Yea I removed this by just adding another method :) thanks

restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
EricBorczuk: Right, ofc - this should be done in opposite order...

EricBorczuk: fixed

dimas-b (approved):

PR #1099, opened 2021-07-12
Throw a user error in case of value type mismatch in Document filters

dimas-b:


dougwettlaufer (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
```diff
@@ -157,6 +159,26 @@ public ExpressionParser(ConditionParser predicateProvider) {
     return expressions;
   }
 
+  private void validateFieldConditions(
+      FilterPath filterPath, Collection<BaseCondition> conditions) {
+    // If some conditions imply specific and different value types it is a user error since each
+    // field can have a value of only one type at a time.
+    // Example: `field GT 1` and `field EQ "string"`
+    Set<? extends Class<?>> specificTypes =
```
mpenick, line 167: Would it ever be possible to have field condition for multiple fields? 

`field1 GT 1 ... field2 EQ "string"`

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/BaseCondition.java
```diff
@@ -43,6 +44,18 @@ default boolean isPersistenceCondition() {
...
+  /**
+   * Returns the most specific value type implied by this condition.
+   *
+   * <p>Note: if a specific value type is not known, {@code Object.class} should be returned.
+   *
+   * @return the most specific type of value implied by this condition (must not be null).
+   */
+  @NotNull
+  default Class<?> getQueryValueType() {
+    return Object.class;
```
mpenick, line 56: I your stream expression you filter `Object.class != c`. Could this not have a default implementation (or throw in the default implementation)?

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
dimas-b: That example is valid even now. Current change applies only to collections of conditions on one particular field.

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/BaseCondition.java
dimas-b: @mpenick : I'm not sure what you mean, sorry. `Object.class` is a valid return value, it just means the condition does not imply a specific type, which is why it is excluded from the validation logic.

For example: `"field": { "$in": ["string", 12345], "$gt": "a" }` is meaningful from the execution perspective (if not from the user-level logic perspective) because the same field name may map to different values in different documents. The `$in` operation does not imply a particular value type, therefore mixing it with a `$gt` operation on a string is allowed.

However, `"field": { "$in": ["string", 12345], "$gt": "a", "$lt": 10}` is not allowed because both the `$gt` and `$lt` conditions have to apply to the same values in each document, but they imply different value types. The fact that the `$in` operation does not imply a specific type here is irrelevant because we have a conflict between `$gt` and `$lt` conditions anyway.

dimas-b: We could, of course, enhance  the validation logic to detect whether the inner values in an `$in` condition conflict with another condition, but I thought it'd be an overkill for now. The main intention of this PR is to provide some more level of validation to use out accidental end-user mistakes. I can certainly try and enhance the validation of inner `$in` values. WDYT?

mpenick: :+1: Thanks for the explanation.

mpenick (approved):

PR #1102, opened 2021-07-13
Update openapi spec annotations for specific return types rather than generic responsewrapper

### LINKED ISSUE #1101: OpenAPI specs don't return typed responses in any API with a "raw" parameter
Using `CollectionsResource` as an example, any of the resources that use the `raw` parameter generate an OpenAPI spec that loses the type information. `ResponseWrapper` is a generic data holder.

When generating any clients in a typed language (e.g. Java, Kotlin) from the specs the generated have two current limitations

1. In many of the methods the `@ApiOperation` is missing and so the generated stubs have a void return type.
1. With the recent fix on the `CollectionsResource` https://github.com/stargate/stargate/pull/1100, the new return type is invalid if called with `raw=true`

I would propose that anyone using the OpenAPI spec to generate API clients to Stargate is using a language where they would default to `raw=true` so as to have typed data in their application.

As an example making this change from `ResponseWrapper` to `DocCollection` in `CollectionsResource`

```
@GET                                                                                     
@ApiOperation(                                                                           
    value = "List collections in namespace",                                             
    response = DocCollection.class,                                                      
    responseContainer = "List")                                                          
@ApiResponses(                                                                           
    value = {                                                                            
      @ApiResponse(code = 200, message = "OK", response = DocCollection.class),          
      @ApiResponse(code = 401, message = "Unauthorized", response = Error.class),        
      @ApiResponse(code = 500, message = "Internal server error", response = Error.class)
    })                                                                                   
```

generates

```
data class DocCollection (
    /* The name of the collection. */
    @SerializedName("name")
    val name: kotlin.String,
    /* Whether an upgrade is available. Use the 'upgrade a collection' endpoint with the upgrade type to perform upgrade. */
    @SerializedName("upgradeAvailable")
    val upgradeAvailable: kotlin.Boolean? = null,
    /* The upgrade type, if an upgrade is available. */
    @SerializedName("upgradeType")
    val upgradeType: DocCollection.UpgradeType? = null
) {

    /**
     * The upgrade type, if an upgrade is available.
     * Values: SAI_INDEX_UPGRADE
     */
    enum class UpgradeType(val value: kotlin.String) {
        @SerializedName(value = "SAI_INDEX_UPGRADE") SAI_INDEX_UPGRADE("SAI_INDEX_UPGRADE");
    }
}

    @GET("v2/namespaces/{namespace-id}/collections")
    suspend fun getCollections(@Header("X-Cassandra-Token") xCassandraToken: kotlin.String,
        @Path("namespace-id") namespaceId: kotlin.String,
        @Query("raw") raw: kotlin.Boolean? = null): Response<DocCollection>
```

which is much more useful than

```
data class ResponseWrapper (
    /* Response data returned by the request. */
    @SerializedName("data")
    val data: kotlin.Any? = null
)

@GET("v2/namespaces/{namespace-id}/collections")
suspend fun getCollections(@Header("X-Cassandra-Token") xCassandraToken: kotlin.String,
    @Path("namespace-id") namespaceId: kotlin.String,
    @Query("raw") raw: kotlin.Boolean? = null): Response<ResponseWrapper>
```

I'd speculate that any clients accessing the doc api are not accessing it by generating OpenAPI client stubs because in fact the return types are missing for some methods.  It might be preferable to use the `oneof` part of the OAS 3.0 spec to specify the two different return types whether or not `raw=true`, but AFAIK there is no way via the Swagger annotations to automate this.  Therefore I'd request that the spec defaults as though `raw=true` and specifies the correct types, rather than the wrapper.

gconaty:
Wherever a response entity was set to use a specific class, namely when raw=true, modify the annotations to include that class rather than ResponseWrapper so that clients can generate the same types.

Fixes #1101

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
@@ -64,11 +64,11 @@ public class CollectionsResource {
   @GET
   @ApiOperation(
       value = "List collections in namespace",
-      response = ResponseWrapper.class,
+      response = DocCollection.class,
       responseContainer = "List")
   @ApiResponses(
       value = {
-        @ApiResponse(code = 200, message = "OK", response = ResponseWrapper.class),
+        @ApiResponse(code = 200, message = "OK", response = DocCollection.class),
         @ApiResponse(code = 401, message = "Unauthorized", response = Error.class),
         @ApiResponse(code = 500, message = "Internal server error", response = Error.class)
       })
@@ -233,11 +233,12 @@ public class CollectionsResource {
   @POST
   @ApiOperation(
       value = "Upgrade a collection in a namespace",
+      response = DocCollection.class,
       notes =
           "WARNING: This endpoint is expected to cause some down-time for the collection you choose.")
   @ApiResponses(
       value = {
-        @ApiResponse(code = 200, message = "OK", response = ResponseWrapper.class),
+        @ApiResponse(code = 200, message = "OK", response = DocCollection.class),
         @ApiResponse(code = 400, message = "Bad Request", response = Error.class),
         @ApiResponse(code = 401, message = "Unauthorized", response = Error.class),
         @ApiResponse(code = 404, message = "Collection not found", response = Error.class),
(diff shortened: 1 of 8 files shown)


ashryanbeats (approved):

ashryanbeats:
Looks good! 

As a future note for us, if we do need to update the annotations to support `raw=false`, @gconaty has provided some info on using `oneOf` annotations the issue associated with this PR.

dimas-b:
Would `oneOf` be equally convenient from the generated code POV on the client side?

dougwettlaufer (approved):

mpenick (approved):

PR #1105, opened 2021-07-14
Encode special characters and decode them on read

EricBorczuk:
This PR removes restrictions on characters in field names.  Because certain characters have a special meaning in the search/path syntax, the characters `, . *` were previously disallowed; however now they are allowed if they are escaped on ingestion. This is the case for both writing documents and searching with a path.

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
@@ -34,7 +34,6 @@ import java.nio.ByteBuffer;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
-import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import org.apache.cassandra.stargate.db.ConsistencyLevel;
 import org.slf4j.Logger;
@@ -45,7 +44,6 @@ public class DocumentDB {
   private static final List<Column> allColumns;
   private static final List<String> allColumnNames;
   private static final List<Column.ColumnType> allColumnTypes;
-  private static final Pattern UNICODE_ESCAPE_PATTERN = Pattern.compile("\\\\u");
   private static final List<String> allPathColumnNames;
   private static final List<Column.ColumnType> allPathColumnTypes;
   public static final int MAX_PAGE_SIZE = 20;
@@ -145,7 +143,7 @@ public class DocumentDB {
   }
 
   public static boolean containsIllegalSequences(String x) {
-    return x.contains("[") || UNICODE_ESCAPE_PATTERN.matcher(x).find();
+    return x.contains("[") || x.contains(".");
   }
 
   public static List<Column> allColumns() {
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
@@ -144,8 +144,11 @@ public class DocumentDB {
   }
 
   public static boolean containsIllegalSequences(String x) {
-    String replaced = x.replaceAll(DocsApiUtils.ESCAPED_FORBIDDEN_CHAR_PATTERN.pattern(), "");
-    return replaced.contains("[") || replaced.contains(".") || replaced.contains("'");
+    String replaced = x.replaceAll(DocsApiUtils.ESCAPED_PATTERN.pattern(), "");
+    return replaced.contains("[")
+        || replaced.contains(".")
+        || replaced.contains("'")
+        || replaced.contains("\\");
   }
 
   public static List<Column> allColumns() {
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
@@ -30,7 +30,6 @@ import io.stargate.web.docsapi.service.ExecutionContext;
 import io.stargate.web.docsapi.service.QueryExecutor;
 import io.stargate.web.docsapi.service.RawDocument;
 import io.stargate.web.docsapi.service.json.DeadLeaf;
-import io.stargate.web.docsapi.service.util.DocsApiUtils;
 import java.nio.ByteBuffer;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
@@ -141,14 +140,6 @@ public class DocumentDB {
 
   public boolean treatBooleansAsNumeric() {
     return !dataStore.supportsSecondaryIndex();
-  }
-
-  public static boolean containsIllegalSequences(String x) {
-    String replaced = x.replaceAll(DocsApiUtils.ESCAPED_PATTERN.pattern(), "");
-    return replaced.contains("[")
-        || replaced.contains(".")
-        || replaced.contains("'")
-        || replaced.contains("\\");
   }
 
   public static List<Column> allColumns() {
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
@@ -30,6 +30,7 @@ import io.stargate.web.docsapi.service.ExecutionContext;
 import io.stargate.web.docsapi.service.QueryExecutor;
 import io.stargate.web.docsapi.service.RawDocument;
 import io.stargate.web.docsapi.service.json.DeadLeaf;
+import io.stargate.web.docsapi.service.util.DocsApiUtils;
 import java.nio.ByteBuffer;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
@@ -143,7 +144,8 @@ public class DocumentDB {
   }
 
   public static boolean containsIllegalSequences(String x) {
-    return x.contains("[") || x.contains(".") || x.contains("'");
+    String replaced = x.replaceAll(DocsApiUtils.ESCAPED_FORBIDDEN_CHAR_PATTERN.pattern(), "");
+    return replaced.contains("[") || replaced.contains(".") || replaced.contains("'");
   }
 
   public static List<Column> allColumns() {
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
@@ -2,7 +2,6 @@ package io.stargate.web.docsapi.dao;
 
 import com.datastax.oss.driver.api.core.servererrors.AlreadyExistsException;
 import com.google.common.base.Splitter;
-import com.google.common.collect.ImmutableList;
 import io.reactivex.rxjava3.core.Flowable;
 import io.reactivex.rxjava3.core.Single;
 import io.stargate.auth.AuthenticationSubject;
@@ -35,6 +34,7 @@ import java.nio.ByteBuffer;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import org.apache.cassandra.stargate.db.ConsistencyLevel;
 import org.slf4j.Logger;
@@ -42,10 +42,11 @@ import org.slf4j.LoggerFactory;
 
 public class DocumentDB {
   private static final Logger logger = LoggerFactory.getLogger(DocumentDB.class);
-  private static final List<Character> forbiddenCharacters;
   private static final List<Column> allColumns;
   private static final List<String> allColumnNames;
   private static final List<Column.ColumnType> allColumnTypes;
+  private static final Pattern ARRAY_INDEX_PATTERN = Pattern.compile("\\[[0-9]+\\]");
+  private static final Pattern UNICODE_ESCAPE_PATTERN = Pattern.compile("\\\\u");
   private static final List<String> allPathColumnNames;
   private static final List<Column.ColumnType> allPathColumnTypes;
   public static final int MAX_PAGE_SIZE = 20;
@@ -103,8 +104,6 @@ public class DocumentDB {
     allColumnTypes.add(Type.Boolean);
     allColumns.add(Column.create("bool_value", Type.Boolean));
 
-    forbiddenCharacters = ImmutableList.of('[', ']', ',', '.', '\'', '*');
-
     if (MAX_ARRAY_LENGTH > 1000000) {
       throw new IllegalStateException(
           "stargate.document_max_array_len cannot be greater than 1000000.");
@@ -146,20 +145,8 @@ public class DocumentDB {
     return !dataStore.supportsSecondaryIndex();
   }
 
-  public static List<String> getForbiddenCharactersMessage() {
-    return forbiddenCharacters.stream().map(ch -> "`" + ch + "`").collect(Collectors.toList());
-  }
-
-  public static boolean containsIllegalChars(String x) {
-    return forbiddenCharacters.stream().anyMatch(ch -> x.indexOf(ch) >= 0);
-  }
-
-  public static String replaceIllegalChars(String x) {
-    String newStr = x;
-    for (Character y : forbiddenCharacters) {
-      newStr = newStr.replace(y, '_');
-    }
-    return newStr;
+  public static boolean containsIllegalSequences(String x) {
+    return ARRAY_INDEX_PATTERN.matcher(x).matches() || UNICODE_ESCAPE_PATTERN.matcher(x).matches();
   }
 
   public static List<Column> allColumns() {
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
@@ -45,7 +45,6 @@ public class DocumentDB {
   private static final List<Column> allColumns;
   private static final List<String> allColumnNames;
   private static final List<Column.ColumnType> allColumnTypes;
-  private static final Pattern ARRAY_INDEX_PATTERN = Pattern.compile("\\[[0-9]+\\]");
   private static final Pattern UNICODE_ESCAPE_PATTERN = Pattern.compile("\\\\u");
   private static final List<String> allPathColumnNames;
   private static final List<Column.ColumnType> allPathColumnTypes;
@@ -146,7 +145,7 @@ public class DocumentDB {
   }
 
   public static boolean containsIllegalSequences(String x) {
-    return ARRAY_INDEX_PATTERN.matcher(x).matches() || UNICODE_ESCAPE_PATTERN.matcher(x).matches();
+    return x.contains("[") || UNICODE_ESCAPE_PATTERN.matcher(x).find();
   }
 
   public static List<Column> allColumns() {
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
@@ -143,7 +143,7 @@ public class DocumentDB {
   }
 
   public static boolean containsIllegalSequences(String x) {
-    return x.contains("[") || x.contains(".");
+    return x.contains("[") || x.contains(".") || x.contains("'");
   }
 
   public static List<Column> allColumns() {
--- a/restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
@@ -49,7 +49,7 @@ public enum ErrorCode {
 
   DOCS_API_GENERAL_INVALID_FIELD_NAME(
       Response.Status.BAD_REQUEST,
-      String.format("Array paths with brackets are not permitted in JSON field names.")),
+      "Array paths with brackets are not permitted in JSON field names."),
 
   DOCS_API_GENERAL_PAGE_SIZE_EXCEEDED(
       Response.Status.BAD_REQUEST,
--- a/restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
@@ -16,7 +16,6 @@
 
 package io.stargate.web.docsapi.exception;
 
-import io.stargate.web.docsapi.dao.DocumentDB;
 import io.stargate.web.docsapi.service.DocsApiConfiguration;
 import io.stargate.web.models.Error;
 import javax.ws.rs.core.MediaType;
@@ -50,9 +49,7 @@ public enum ErrorCode {
 
   DOCS_API_GENERAL_INVALID_FIELD_NAME(
       Response.Status.BAD_REQUEST,
-      String.format(
-          "The characters %s are not permitted in JSON field names.",
-          DocumentDB.getForbiddenCharactersMessage())),
+      String.format("Array paths with brackets are not permitted in JSON field names.")),
 
   DOCS_API_GENERAL_PAGE_SIZE_EXCEEDED(
       Response.Status.BAD_REQUEST,
--- a/restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
@@ -49,7 +49,7 @@ public enum ErrorCode {
 
   DOCS_API_GENERAL_INVALID_FIELD_NAME(
       Response.Status.BAD_REQUEST,
-      "Array paths with brackets are not permitted in JSON field names."),
+      "Array paths contained in square brackets, periods, single quotes, and backslash are not allowed in field names."),
 
   DOCS_API_GENERAL_PAGE_SIZE_EXCEEDED(
       Response.Status.BAD_REQUEST,
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
@@ -44,7 +44,7 @@ import org.slf4j.LoggerFactory;
 
 public class DocumentService {
   private static final Logger logger = LoggerFactory.getLogger(DocumentService.class);
-  private static final Pattern PERIOD_PATTERN = Pattern.compile("\\.");
+  private static final Pattern PERIOD_PATTERN = Pattern.compile("(?<!\\\\)\\.");
   private static final Splitter FORM_SPLITTER = Splitter.on('&');
   private static final Splitter PAIR_SPLITTER = Splitter.on('=');
 
@@ -164,7 +164,7 @@ public class DocumentService {
 
                     // pv always starts with a square brace because of the above conversion
                     String innerPath =
-                        DocsApiUtils.convertUnicodeCodePoints(pv.substring(1, pv.length() - 1));
+                        DocsApiUtils.convertEscapedCharacters(pv.substring(1, pv.length() - 1));
                     boolean isArrayElement = op.getType() == PathOperator.Type.ARRAY;
                     if (isArrayElement) {
                       if (i == path.size() && patching) {
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
@@ -29,7 +29,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
-import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.ws.rs.core.PathSegment;
@@ -44,7 +43,6 @@ import org.slf4j.LoggerFactory;
 
 public class DocumentService {
   private static final Logger logger = LoggerFactory.getLogger(DocumentService.class);
-  private static final Pattern PERIOD_PATTERN = Pattern.compile("(?<!\\\\)\\.");
   private static final Splitter FORM_SPLITTER = Splitter.on('&');
   private static final Splitter PAIR_SPLITTER = Splitter.on('=');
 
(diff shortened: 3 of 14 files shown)


EricBorczuk:
I still have to adapt search to work with this change, working on that now

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -102,7 +106,8 @@
     allColumnTypes.add(Type.Boolean);
     allColumns.add(Column.create("bool_value", Type.Boolean));
 
-    forbiddenCharacters = ImmutableList.of('[', ']', ',', '.', '\'', '*');
+    escapableCharacters = ImmutableList.of('[', ']', ',', '.', '\'', '*');
+    forbiddenCharacters = ImmutableList.of('%');
```
dimas-b, line 110: Why  do we forbid `%` chars? The URL encoding scheme should be able to support escaping them, I belive.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -153,14 +158,37 @@ public static boolean containsIllegalChars(String x) {
...
+  /**
+   * Decodes all hex-encoded characters into their natural form, effectively reversing
+   * DocumentDB#replaceEscapableChars
+   *
+   * @param x A string with encoded characters, such as `eric@gmail%2ecom`
+   * @return A string with all encoded characters decoded, such as `eric@gmail.com`
+   */
+  public static String decodeEscapableChars(String x) {
+    try {
+      return URLDecoder.decode(x, StandardCharsets.UTF_8.name());
```
dimas-b, line 185: Using `URLDecoder` in this context will probably require the end user to escape `%` signs twice because otherwise the HTTP/Servlet libs are going to interpret them before we get the field name in our code. WDYT?

restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverter.java
```diff
@@ -65,7 +65,7 @@ private JsonNode convertToJsonDoc(
 
     for (Row row : rows) {
       Long rowWriteTime = row.getLong(writeTimeCol.name());
-      String rowLeaf = row.getString("leaf");
+      String rowLeaf = DocumentDB.decodeEscapableChars(row.getString("leaf"));
```
dimas-b, line 68: Would it not be easier to keep document contents "as is" (unescaped), but support special char escaping when dealing with paths (where `[` and other "forbidden" charts have special meaning)?

EricBorczuk: There are a few cases where I think things might fall apart...I remember thinking about this a while back and realizing that it wouldn't work sometimes to do it the simple way.

If we allow all characters - imagine you have a document A like `{ "a": { "b.c": "value" }}`
that would be stored as p0 = 'a', p1 = 'b.c'

And you would likely want to query it like `{ "a.b.c": {"$eq": "value"}}`

But also, imagine you have a document B `{ "a": { "b": { "c": "value" }}}`
that would be stored as p0 = 'a', p1 = 'b', p2 = 'c'

And you would ALSO want to query it like `{ "a.b.c": {"$eq": "value"}}`

So then there is some ambiguity; however with my change you could theoretically search like
`{ "a.b%2ec": {"$eq": "value"}}` if you want to match document A, and you could do `{ "a.b.c": {"$eq": "value"}}` to match document B.  It's a bit ugly, but removes the ambiguity.

The other situation that is nasty is that if we allow all characters without encoding, someone could save data like `{"[000001]": "a"}` and it would get read out as an array inadvertently - but by encoding the literal value in the database becomes `%5b000001%5d` and so it won't be interpreted as an array (while legitimate array values will have non-encoded `[ ]` in them)

EricBorczuk: These are just two examples, I'm guessing there is a whole lot of nasty cases that could be reached as well...and I agree that this implementation isn't necessarily simple.

To take a step back, the value of doing this is really that people can save keys in their documents that have `.` in them, which is very common since people love to key on emails, e.g. `{ "eric@gmail.com": { "enabled": true }}` or something similar. But I'm second-guessing now whether it's worth the complexity required to accept this?

EricBorczuk: ah I think I see what you're saying about that first case - we could just have url encoding be supported in the filter and then store as-is for that case.

I think the array case is an issue still though

dimas-b: > So then there is some ambiguity; however with my change you could theoretically search like
{ "a.b%2ec": {"$eq": "value"}} if you want to match document A, and you could do { "a.b.c": {"$eq": "value"}} to match document B. It's a bit ugly, but removes the ambiguity.

Correct, but I think we do not have to encode stored values for this to work. We only need to parse the path specification from JSON into POJOs correctly, eh?

dimas-b: > The other situation that is nasty is that if we allow all characters without encoding, someone could save data like {"[000001]": "a"} and it would get read out as an array inadvertently - but with encoding the literal value in the database becomes %5b000001%5d and so it won't be interpreted as an array (while legitimate array values will have non-encoded [ ] in them)

Good point. But could we not solve this by forbidding `[` and `]` as the first and last property name chars specifically? I'd guess it might be an acceptable restrictions if we could avoid unescaping property names on every read... WDYT?

I do not object, just bringing up points for discusstion :)

EricBorczuk: Yea I'm leaning towards what you just said: `forbidding \[[0-9]+\] (regex) property names` and then allowing data to be written as-is to the database...and then parsing the `where` filter has to deal with encoding of special chars - that would work way better!

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
```diff
@@ -50,9 +49,7 @@
 
   DOCS_API_GENERAL_INVALID_FIELD_NAME(
       Response.Status.BAD_REQUEST,
-      String.format(
-          "The characters %s are not permitted in JSON field names.",
-          DocumentDB.getForbiddenCharactersMessage())),
+      String.format("Array paths with brackets are not permitted in JSON field names.")),
```
tatu-at-datastax, line 52: No need to use format as there are not replacements

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -145,20 +143,8 @@ public boolean treatBooleansAsNumeric() {
...
-  }
-
-  public static String replaceIllegalChars(String x) {
-    String newStr = x;
-    for (Character y : forbiddenCharacters) {
-      newStr = newStr.replace(y, '_');
-    }
-    return newStr;
+  public static boolean containsIllegalSequences(String x) {
+    return x.contains("[") || UNICODE_ESCAPE_PATTERN.matcher(x).find();
```
tatu-at-datastax, line 147: Might not matter a lot but these 2 checks can be combined into single regexp. Or, since UNICODE_ESCAPE_PATTERN looks like constant String (unless I misread it -- so many backslashes :) ), could just use `contains()`. But java.util Regexps have more overhead than simple substring check.

(actually, funnily enough I was playing with this part of code & test performance for fun:

https://github.com/tatu-at-datastax/stargate-micro-benchmarks

... but I digress)

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
```diff
@@ -61,6 +62,37 @@ public static String convertArrayPath(String path) {
     }
   }
 
+  /**
+   * Converts any literal unicode points into their actual character. E.g. if the input string is
+   * literally abc\u002E123, this function returns abc.123 This allows a user to use unicode escape
+   * sequence in where filters when it would otherwise be ambiguous to use the corresponding
+   * character. Any path that already has square brackets in it will be left alone.
```
tatu-at-datastax, line 69: So this skips bracketed-only, like "[1234]"?

restapi/src/test/java/io/stargate/web/docsapi/service/util/DocsApiUtilsTest.java
```diff
@@ -108,6 +108,40 @@ public void numberFormatException() {
...
+
+      result =
+          DocsApiUtils.convertUnicodeCodePoints(
+              "\\u but without valid code points after are ignored: \\ufg00");
+      assertThat(result).isEqualTo("\\u but without valid code points after are ignored: \\ufg00");
+    }
+
+    @Test
+    public void arraySegmentsIgnored() {
+      String result = DocsApiUtils.convertUnicodeCodePoints("[1],[44],[555]");
```
tatu-at-datastax, line 132: Since there are no Unicode escapes here, why should the result be anything else than the original String?
(I am assuming encoding is handled elsewhere)

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
EricBorczuk: Oh yea, I was doing something more complex before, but now it can just be a simple `contains` :) I will change

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
EricBorczuk: Yes, that's correct

restapi/src/test/java/io/stargate/web/docsapi/service/util/DocsApiUtilsTest.java
EricBorczuk: This was just a blind copy paste from the corresponding test above - and I think I'm going to remove it :)

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
tatu-at-datastax: Yeah, it should be fine for escaping to work at multiple levels I think? But I could be missing something.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -145,20 +141,8 @@ public boolean treatBooleansAsNumeric() {
...
-    return forbiddenCharacters.stream().anyMatch(ch -> x.indexOf(ch) >= 0);
-  }
-
-  public static String replaceIllegalChars(String x) {
-    String newStr = x;
-    for (Character y : forbiddenCharacters) {
-      newStr = newStr.replace(y, '_');
-    }
-    return newStr;
+  public static boolean containsIllegalSequences(String x) {
```
tatu-at-datastax, line 144: Not directly related to changes here, but is there a specification of how handling should work? Like a bigger document that describes intended behavior, including disallowed characters and so on?
And if there isn't, would it make to start writing/maintaining such document(s) now? :)

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
tatu-at-datastax: I should have asked: why is this needed/intended?

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
```diff
@@ -61,6 +62,37 @@ public static String convertArrayPath(String path) {
...
+   *
+   * @param path single filter or field path
+   * @return Converted to a string with no literal unicode code points.
+   */
+  public static String convertUnicodeCodePoints(String path) {
```
tatu-at-datastax, line 74: Minor naming suggestion: this relates to escaping more than Unicode code point handling (although escapes are used to encode code points of course) so maybe naming should be something indicating that here we are decoding possibly escaped unicode characters? I mean, indicate direction in name itself, in addition to Javadocs.

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
```diff
@@ -61,6 +62,37 @@ public static String convertArrayPath(String path) {
...
+    if (ARRAY_PATH_PATTERN.matcher(path).matches()) {
+      return path;
+    }
+
+    String[] segments = path.split("\\\\u");
```
tatu-at-datastax, line 79: I think it should be possible to express actual full match in regexp `Pattern`, to avoid further checks?

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
dimas-b: AFAIK this is to avoid clashes with indexes of JSON array elements when they are stored in rows.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -145,20 +141,8 @@ public boolean treatBooleansAsNumeric() {
...
+    return x.contains("[") || x.contains(".") || x.contains("'");
```
dimas-b, line 145: Why are `.` and `'` illegal?

As for `[` I think we can be less stringent and disallow a smaller sub-set of fields names as in regex `^\[[0-9]+\]$`, but allow something like `abc[def`. WDYT?

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
```diff
@@ -50,9 +49,7 @@
 
   DOCS_API_GENERAL_INVALID_FIELD_NAME(
       Response.Status.BAD_REQUEST,
-      String.format(
-          "The characters %s are not permitted in JSON field names.",
-          DocumentDB.getForbiddenCharactersMessage())),
+      "Array paths with brackets are not permitted in JSON field names."),
```
dimas-b, line 52: The message appears to be too far away from the meaning that can be deduced from the name of  the constant: `DOCS_API_GENERAL_INVALID_FIELD_NAME` :thinking:

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
tatu-at-datastax: One quick note on this: as a general rule, it is much easier to relax restrictions than add new ones -- so I think it'd be best to be cautious about what to allow, maybe even reserve certain characters just in case.
And definitely needs to be well documented: I am not sure what is the actual state of Documents API wrt alpha/beta/GA, but at some point these cannot be easily changed by us.

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -226,7 +226,8 @@ private boolean isEmptyArray(Object v) {
                     if (pv.equals("$")) continue;
 
                     // pv always starts with a square brace because of the above conversion
-                    String innerPath = pv.substring(1, pv.length() - 1);
+                    String innerPath =
+                        DocsApiUtils.convertUnicodeCodePoints(pv.substring(1, pv.length() - 1));
```
dimas-b, line 230: I'm not sure I understand why we need to unescape `pv` in this case :thinking: Could you clarify, please?

I assume we always take any JSON name in a doc body "as is" and only apply escaping to names from query parameters (e.g. field paths). Is that a valid assumption?

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
```diff
@@ -61,6 +62,37 @@ public static String convertArrayPath(String path) {
...
+    for (int i = 1; i < segments.length; i++) {
+      String segment = segments[i];
+      if (segment.length() >= 4
+          && UNICODE_POINT_PATTERN.matcher(segment.substring(0, 4)).matches()) {
```
dimas-b, line 83: IMHO, Unicode points are inconvenient from the end-user's perspective. I guess in many cases query parameter may be written by hand, so figuring out the right escape sequence is going to be tedious and error-prone.

Would `\` escapes be an option (like in java: `"\""`, in Docs API: `field1\.partTwo`, `\\` - literal `\` char)?

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
dimas-b: +1 to the point on adding restrictions.

However, I really hope that we do not have to forbid any names valid in JSON document bodies. I hope all on the escaping can be on the query parameters side, where we have more control over the data structures.

The array index case seems to be the only point of contention, as far as I can see, because that is the only case where we have to decide whether to interpret the `pNNN` column value as a "field" or as an "index".

ivansenic:
@tatu-at-datastax @dimas-b  Since Eric is on holiday, do we wait for him to finish this himself, or are we jumping on top here?

dimas-b:
From my POV this PR is not critical. I'd prefer to wait for @EricBorczuk

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
dimas-b: @ashryanbeats : Do you have any guidelines wrt escaping methods for Docs API query parameters (this is to happen after URL decoding).

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
EricBorczuk: Yea I think we are getting to that point; there is so document but it would be good to have.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
EricBorczuk: To answer the question about `[`, `.`, and `'`, they can't be present because internally right here (https://github.com/stargate/stargate/pull/1105/files#diff-caf69767609da0fee30d28b2b410ba5e400ed01e1f42af082f495ef82fec8416R97) JSON paths like `$.a.b.c` get converted to `$['a']['b']['c']`. The point of this conversion is to allow nearly any character to be present without need to escape (this code has existed since virtually the beginning). But that means that if you have literal `'` or `[` in your paths then the conversion creates malformed paths like `$['Eric's']`, etc. Periods need escaping over the wire for a similar reason, since if the input path is `$.a......b.c` or something, you'll end up with problems in this conversion as well

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
EricBorczuk: I was hoping for that to be the case, but because of what I named above https://github.com/stargate/stargate/pull/1105#discussion_r677464700 there is actually a need to handle escaping while writing a document as well.

For example, a document such as `{ "a.b": "c" }` will be converted by the JsonSurfer library to `$.a.b` (which isn't quite right!!) - and then our code turns that into `$['a']['b']` - essentially this will write `p0='a', p1='b', text_value='c'` to cassandra, which isn't quite what we want here (we would want `p0='a.b', text_value='c'`)

So, by escaping like `{ "a\u002eb": "c" }`, the JsonSurfer library makes the path `$.a\u002eb`, and we convert it to `$['a\u002eb']` - which would cause `p0='a\u002eb', text_value='c'` to be written to the db; close but not quite right.

So the last thing that has to be done is to take each of pN and convert unicode escape sequences to actual literal characters, and then it's all correct in the database.

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
EricBorczuk: Yea this is a good question...one of the benefits of allowing unicode is that you can put other types of characters as well, but that's not necessarily the most useful thing in the world...

I think there might be a way to accept frontslash escaping for periods, braces, and quotes, and maybe that would do the trick? I'm not married to using unicode escapes, I agree that it might be a bit unwieldy - but I also figured for Javascript and Java developers it would be familiar to them at the very least

ivansenic (changes requested):
So after rebasing please also take into account the changes needed here https://github.com/stargate/stargate/blob/5095e9f1111b28e9b00e04a5f796a1005542cca6/restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java#L442

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
dimas-b: Hmmm... it might be an overkill, but could we avoid using JSON paths so as to reduce the need for escaping and allow a larger set of chars? Could we perform the same operation "manually" using simpler APIs? I suppose we only use a small sub-set of the JSON path functionality :thinking: WDYT?

EricBorczuk:
Yep! rebasing now

ivansenic (changes requested):
Changes in the `ReactiveDocumentServcie` still not included, failing only so that we don't forget.

EricBorczuk:
ReactiveDocumentService change is done

So in the end, I had to make a decision between using `\uXXXX` to escape special characters or just a simple `\` in front of the character.  It turns out that because of how the JsonSurfer parses paths out, ANY path containing the chars `[` or `'` will cause JsonSurfer to fail...using `\u` I can get around this issue and allow the user to use braces in their field names, but it is unwieldy and looks odd. Also, the main reason for this PR is to allow fields to be email addresses or similar, i.e. allowing periods is the main thing that we want here.

So I decided to allow periods, commas, and asterisk by a simple `\`. But `[`. `]`, and `'` remain disallowed because they break the parser.

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -439,7 +439,6 @@ public ReactiveDocumentService(
 
   // we need to transform the stuff to support array elements
   private List<String> processSubDocumentPath(List<String> subDocumentPath) {
-    // TODO after https://github.com/stargate/stargate/pull/1105 encode chars as well
```
ivansenic, line 442: I don't get this now why is the path not escaped here, are we only doing so for the fields?

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/db/impl/FilterPathSearchQueryBuilder.java
```diff
@@ -76,7 +77,7 @@ public FilterPath getFilterPath() {
     if (parentSize >= DocumentDB.MAX_DEPTH) {
       throw new ErrorCodeRuntimeException(ErrorCode.DOCS_API_GENERAL_DEPTH_EXCEEDED);
     } else {
-      String field = filterPath.getField();
+      String field = DocsApiUtils.convertEscapedCharacters(filterPath.getField());
```
ivansenic, line 80: I would strongly disagree to do any conversions in the QueryBuilders. The unescaped stuff should already be in the in the path, no need to change here..

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/db/impl/PathSearchQueryBuilder.java
```diff
@@ -42,6 +44,12 @@ public PathSearchQueryBuilder(List<String> path) {
     this.path = path;
   }
 
+  private List<String> unescape(List<String> escaped) {
```
ivansenic, line 47: same here, no unesaping in the query builders, paths should already be unescaped.. This is a non-needed overhead for the query builders..

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
```diff
@@ -234,7 +250,7 @@ public static boolean isRowMatchingPath(Row row, List<String> path) {
     // short-circuit if the field is not matching
     String field = path.get(targetPathSize - 1);
     String leaf = row.getString(QueryConstants.LEAF_COLUMN_NAME);
-    if (!Objects.equals(field, leaf)) {
+    if (!Objects.equals(DocsApiUtils.convertEscapedCharacters(field), leaf)) {
```
ivansenic, line 253: same here, the path should already be unescaped, not need to alter here.

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
```diff
@@ -292,11 +308,12 @@ public static boolean isRowOnPath(Row row, Iterable<String> pathIterable) {
       // if we have the path segment, we need to check if any matches
       if (pathSegment) {
         boolean noneMatch =
-            Arrays.stream(target.split(",")).noneMatch(t -> Objects.equals(t, path));
+            Arrays.stream(target.split(COMMA_PATTERN.pattern()))
+                .noneMatch(t -> Objects.equals(DocsApiUtils.convertEscapedCharacters(t), path));
```
ivansenic, line 312: ahd here

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
```diff
@@ -292,11 +308,12 @@ public static boolean isRowOnPath(Row row, Iterable<String> pathIterable) {
...
         if (noneMatch) {
           return false;
         }
-      } else if (!Objects.equals(path, target)) {
+      } else if (!Objects.equals(path, DocsApiUtils.convertEscapedCharacters(target))) {
```
ivansenic, line 316: and here

ivansenic (changes requested):
This PR is very confusing. It's confused what's being encoded and where.. We should do any encoding before entering any query and matching process. This is exactly how we have it now, we do convert all before (or when) we construct filter expressions and then there is no need to do this in the single query builders and matching logic.

We need to make thid right, otherwise we will screw up.

@EricBorczuk Let's sync if you want in order to create proper strategy here.

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
EricBorczuk: So this case right here is actually the reason for waiting til the "last moment" before query building to do the unescaping...

If you have a value for p0 in the database that is literally `a,b` and you want to match it, you would have to send `{ "a\,b": { "$exists": true }` as your where clause. If you unescape "a\,b" too _early_, then splitting on a comma produces `["a", "b"]` here (which is the "a or b path" syntax that we allow with commas). By keeping the `\` sequence until the last moment, this COMMA_PATTERN (which uses a negative lookback to check for `\`) does NOT split "a\,b" at all, and then only _after_ the attempt to split does unescaping occur.  So instead of producing `["a", "b"]` after the split on this line, you instead get `["a,b"]` which actually matches the value in the database...

There are two potentially bad things here but I couldn't think of another way to do it...one is the COMMA_PATTERN (and also PERIOD_PATTERN) which has a negative lookback, essentially the COMMA_PATTERN regex is "any comma that is not preceded by a `\`"

And the second thing, as you pointed out, is that the unescaping occurs right before querying or matching in memory...any thoughts on how to make this cleaner?

ivansenic: @EricBorczuk Understood. I would like us to sync on this and think what's the best solution here?

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -146,20 +143,12 @@ public boolean treatBooleansAsNumeric() {
...
-    return forbiddenCharacters.stream().anyMatch(ch -> x.indexOf(ch) >= 0);
-  }
-
-  public static String replaceIllegalChars(String x) {
-    String newStr = x;
-    for (Character y : forbiddenCharacters) {
-      newStr = newStr.replace(y, '_');
-    }
-    return newStr;
+  public static boolean containsIllegalSequences(String x) {
```
ivansenic, line 146: Can we move this to the DocsApiUtils and add dedicated unit tests for this method?

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
ivansenic: Yea let's refactor the error code here or?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
ivansenic: OK I think this can be resolved now, right @EricBorczuk ?

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/db/impl/FilterPathSearchQueryBuilder.java
ivansenic: Can be resolved.

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/db/impl/PathSearchQueryBuilder.java
ivansenic: Can be resolved.

ivansenic, line 47: Can we create `DocsApiUtils.convertEscapedCharacters` that accepts a list and then use the utility rather than implement the method here..

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
ivansenic: Can be resolved.

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
```diff
@@ -296,13 +312,15 @@ public static boolean isRowOnPath(Row row, Iterable<String> pathIterable) {
 
       boolean pathSegment = target.contains(",");
```
ivansenic, line 313: Not refactored as we talked yesterday? Is it not possible or? If not can we use 

`boolean pathSegment = COMMA_PATTERN.matches(target);` for the sake of being correct here..

ivansenic (commented):
Seems good, let's resolve the remaining comments and I can give a final approval.

ivansenic (approved):
Perfect :+1:

PR #1106, opened 2021-07-14
GraphQL schema-first: allow schema fetch operation to return null

### LINKED ISSUE #1103: GraphQL schema-first: improve `schema` response when no custom schema is deployed
If a keyspace is still using the legacy CQL-first schema, but a user tries the `schema` query to see if a schema exists, a confusing technical error message gets returned:

```graphql
// Query (on graphql-admin)
{
  schema(keyspace: "legacyKs") {
    version
    deployDate
  }
}

// Response
{
  "errors": [
    {
      "message": "The field at path '/schema' was declared as a non null type, but the code involved in retrieving data has wrongly returned a null value.  The graphql specification requires that the parent field be set to null, or if that is non nullable that it bubble up null to its parent and so on. The non-nullable type is 'Schema' within parent type 'Query'",
      "path": [
        "schema"
      ],
      "extensions": {
        "classification": "NullValueInNonNullableField"
      }
    }
  ],
  "data": null
}
```

We should detect this condition and return a more appropriate error message.

olim7t:
Fixes #1103

### DIFF
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/admin/SingleSchemaFetcher.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/admin/SingleSchemaFetcher.java
@@ -46,8 +46,8 @@ public class SingleSchemaFetcher extends SchemaFetcher<SchemaSource> {
 
     authorize(context, keyspace);
 
-    return schemaSourceDaoProvider
-        .apply(context.getDataStore())
-        .getSingleVersion(keyspace, version);
+    SchemaSource schemaSource =
+        schemaSourceDaoProvider.apply(context.getDataStore()).getSingleVersion(keyspace, version);
+    return schemaSource;
   }
 }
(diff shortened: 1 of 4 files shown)


dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/admin/SingleSchemaFetcher.java
```diff
@@ -46,8 +46,8 @@ protected SchemaSource get(DataFetchingEnvironment environment, StargateGraphqlC
 
     authorize(context, keyspace);
 
-    return schemaSourceDaoProvider
-        .apply(context.getDataStore())
-        .getSingleVersion(keyspace, version);
+    SchemaSource schemaSource =
+        schemaSourceDaoProvider.apply(context.getDataStore()).getSingleVersion(keyspace, version);
+    return schemaSource;
```
mpenick, line 51: Was this change to debug that `null` was being returned?

mpenick (approved):

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/admin/SingleSchemaFetcher.java
olim7t: It's a remnant from an earlier attempt. I'll revert it.

tomekl007 (approved):

PR #1108, opened 2021-07-15
Fix invalid (e.g `0.0.0.0`) `rpc_address` in `system.local`

mpenick:
Use the broadcast RPC address as the `rpc_address` in `system.local`. It
was populated with the bound RPC address which was often `0.0.0.0` when
deployed.


dougwettlaufer (approved):

dimas-b (commented):
+1

PR #1118, opened 2021-07-19
GraphQL schema-first: allow entity field in delete response payload

### LINKED ISSUE #1082: GraphQL schema-first: allow entity field in delete response payload

olim7t:
Fixes #1082.

### DIFF
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/MutationFetcher.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/MutationFetcher.java
@@ -18,6 +18,7 @@ package io.stargate.graphql.schema.graphqlfirst.fetchers.deployed;
 import com.google.common.collect.Lists;
 import graphql.ExceptionWhileDataFetching;
 import graphql.GraphQLException;
+import graphql.execution.DataFetcherResult;
 import graphql.language.OperationDefinition;
 import graphql.language.SourceLocation;
 import graphql.schema.DataFetchingEnvironment;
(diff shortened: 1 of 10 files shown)


graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/MutationFetcher.java
```diff
@@ -245,4 +251,43 @@ protected ExceptionWhileDataFetching toGraphqlError(
   protected SourceLocation getCurrentFieldLocation(DataFetchingEnvironment environment) {
     return environment.getMergedField().getSingleField().getSourceLocation();
   }
+
+  protected Function<List<MutationResult>, DataFetcherResult<Object>>
+      getDeleteOrUpdateResultBuilder(DataFetchingEnvironment environment) {
```
olim7t, line 256: Shared the code between update and delete per [this discussion](https://github.com/stargate/stargate/pull/1083#discussion_r671433875).

dougwettlaufer (approved):

testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/DeleteCustomConditionsTest.java
```diff
@@ -97,6 +94,11 @@ public static void setup(
             + "    vs: [Int] @cql_if(field: \"v\", predicate: IN)\n"
             + "    ): Boolean\n"
             + "    @cql_delete(targetEntity: \"Foo\")\n"
+            + "  deleteFooWithResponsePayload(\n"
```
tomekl007, line 97: do we have a corresponding test for an update?

tomekl007 (approved):

testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/DeleteCustomConditionsTest.java
olim7t: Yes, see `UpdateCustomConditionsTest`.

PR #1122, opened 2021-07-20
closes #1117: never return html code as the API response

### LINKED ISSUE #1117: APIs should not return HTML content
In some cases, the Docs API returns HTML response instead of the JSON.

Seems to be very similar to the #740, however it seems that not all cases were covered with the fix in the #630. We must set the entity apparently so that HTML is not returned. However, imo this should be solved on the DW/Jersey level, so that no HTML is returned by default on missing entity:

As suggested in [SO](https://stackoverflow.com/questions/63056592/dropwizard-response-statusresponse-status-not-found-build-returns-html):

```
environment.jersey()
        .property(ServerProperties.RESPONSE_SET_STATUS_OVER_SEND_ERROR, true);
```

ivansenic:
:heavy_check_mark: Confirmed with a request like:

```
curl -H "x-cassandra-token: ..." http://127.0.0.4:8082/v2/namespaces/namespace/collections/payu/1/quiz/maths/q1/options/[0] -v 
*   Trying 127.0.0.4:8082...
* TCP_NODELAY set
* Connected to 127.0.0.4 (127.0.0.4) port 8082 (#0)
> GET /v2/namespaces/namespace/collections/payu/1/quiz/maths/q1/options/[0] HTTP/1.1
> Host: 127.0.0.4:8082
> User-Agent: curl/7.68.0
> Accept: */*
> x-cassandra-token: ...
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 404 Not Found
< Date: Tue, 20 Jul 2021 11:30:06 GMT
< Content-Length: 0
< 
* Connection #0 to host 127.0.0.4 left intact
```

which used to return the HTML.

:heavy_check_mark:  Checked the GraphQL playground, works fine.
:heavy_check_mark: INT tests should be fine as well.

### DIFF
--- a/auth-api/src/main/java/io/stargate/auth/api/impl/Server.java
+++ b/auth-api/src/main/java/io/stargate/auth/api/impl/Server.java
@@ -39,6 +39,7 @@ import javax.servlet.DispatcherType;
 import javax.servlet.FilterRegistration;
 import org.eclipse.jetty.servlets.CrossOriginFilter;
 import org.glassfish.hk2.utilities.binding.AbstractBinder;
+import org.glassfish.jersey.server.ServerProperties;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.FrameworkUtil;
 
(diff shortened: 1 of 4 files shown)


dimas-b (approved):

dougwettlaufer (approved):

dougwettlaufer:
> ✔️ Checked the GraphQL playground, works fine.

@ivansenic the swagger ui still works, correct?

ivansenic:
> > heavy_check_mark Checked the GraphQL playground, works fine.
> 
> @ivansenic the swagger ui still works, correct?

Yep it works, I double checked.. I think this settings is only how the errors are reported, thus has nothing with actually serving content.

PR #1123, opened 2021-07-20
Grpc fix response types for batches to support LWTs

tomekl007:


grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -644,9 +645,18 @@ private ConsistencyLevel getTracingConsistency(BatchParameters parameters) {
...
-          if (result.kind != Kind.Void) {
+
+          if (result.kind != Kind.Void && result.kind != Kind.Rows) {
             throw Status.INTERNAL.withDescription("Unhandled result kind").asException();
           }
+
+          if (result.kind == Kind.Rows) {
+            Payload.Type type = validateAndGetTypeForBatchQueries(batch);
+            PayloadHandler handler = PayloadHandlers.get(type);
+            Any data = handler.processResult((Rows) result, false);
```
mpenick, line 656: Should skipping the metadata be a parameter exposed to the user? (maybe in `Batch`) I could go either way here.

grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -644,9 +645,18 @@ private ConsistencyLevel getTracingConsistency(BatchParameters parameters) {
...
           Response.Builder responseBuilder = makeResponseBuilder(result);
           handleTraceId(result.getTracingId(), parameters, responseAndTraceId);
-          if (result.kind != Kind.Void) {
+
+          if (result.kind != Kind.Void && result.kind != Kind.Rows) {
             throw Status.INTERNAL.withDescription("Unhandled result kind").asException();
           }
+
+          if (result.kind == Kind.Rows) {
+            Payload.Type type = validateAndGetTypeForBatchQueries(batch);
```
mpenick, line 654: Maybe type validation could be done earlier (before the queries are executed)?

grpc/src/main/java/io/stargate/grpc/service/Service.java
tomekl007: good point, I've added a version of method that takes `BatchParameters`

grpc/src/main/java/io/stargate/grpc/service/Service.java
tomekl007: done

mpenick (approved):

PR #1124, opened 2021-07-20
GraphQL schema-first: support federated tracing

### LINKED ISSUE #1113: GraphQL schema-first: support federated tracing
See [the Apollo spec](https://www.apollographql.com/docs/federation/metrics/) and [graphql-java docs](https://www.graphql-java.com/documentation/master/instrumentation/#apollo-tracing-instrumentation).

olim7t:
Fixes #1113

### DIFF
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/SchemaProcessor.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/SchemaProcessor.java
@@ -17,6 +17,7 @@ package io.stargate.graphql.schema.graphqlfirst.processor;
 
 import com.apollographql.federation.graphqljava.Federation;
 import com.apollographql.federation.graphqljava._FieldSet;
+import com.apollographql.federation.graphqljava.tracing.FederatedTracingInstrumentation;
 import com.google.common.collect.ImmutableMap;
 import graphql.GraphQL;
 import graphql.GraphQLError;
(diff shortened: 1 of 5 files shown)


graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/SchemaProcessor.java
```diff
@@ -210,22 +211,24 @@ private GraphQL buildGraphql(
...
+              .fetchEntities(new FederatedEntityFetcher(mappingModel))
+              .resolveEntityType(
+                  environment -> {
+                    FederatedEntity entity = environment.getObject();
+                    return environment.getSchema().getObjectType(entity.getTypeName());
+                  })
+              .build();
+      graphqlBuilder =
+          GraphQL.newGraphQL(federationReadySchema)
+              .instrumentation(new FederatedTracingInstrumentation());
```
tomekl007, line 227: isn't the `instrumentation` costly? If so, shouldn't we set it only if some setting is enabled? (and allow to disable that if needed)

testing/src/main/java/io/stargate/it/http/graphql/graphqlfirst/FederationTest.java
```diff
@@ -131,4 +132,25 @@ public void udtKeyTest() {
...
+            KEYSPACE,
+            "query {\n"
+                + "_entities(representations: [ "
+                + "{ __typename: \"Entity2\", k: 1 }, "
+                + " ]) { "
+                + "... on Entity2 { k } "
+                + "} }");
+
+    String trace = JsonPath.read(response, "$.extensions.ftv1");
+    // The value is an opaque string (Base64 encoding of the protobuf representation of the trace).
```
tomekl007, line 151: interesting, could you elaborate more on how this `trace_id` will be used by end-users?

olim7t: The trace data is consumed by the [Apollo Gateway](https://www.apollographql.com/docs/federation/metrics/):

> These federated traces capture the subgraph-level details in the shape of the query plan, which is sent to Apollo's metrics ingress by default, and aggregated into query-level stats and field-level stats.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/SchemaProcessor.java
olim7t: It is already conditioned by an HTTP header (see the integration test).

tomekl007: So is this `apollo-federation-include-trace` handled internally in the `FederatedTracingInstrumentation`?

dougwettlaufer:
Looks like you need to merge in https://github.com/stargate/stargate/pull/1149 to get the tests to pass.

dougwettlaufer (approved):

ivansenic:
> Looks like you need to merge in #1149 to get the tests to pass.

#1149 is merged, you can rebase.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/SchemaProcessor.java
olim7t: Correct, the HTTP header is handled internally [here](https://github.com/apollographql/federation-jvm/blob/master/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java#L469-L476).

PR #1125, opened 2021-07-20
Fix prepared statement handling in ValidatingDataStore

### LINKED ISSUE #1026: Validating data store should compare against externalQueryString
Currently the validating data store compares against the prepared query, which is wrong, because it should compare against the query defined by the user. The placeholder `?` should be used in the asserting query only for the values that are directly binded after the preparation.

dimas-b:
Fixes #1026

### DIFF
--- a/restapi/src/test/java/io/stargate/db/datastore/ValidatingDataStore.java
+++ b/restapi/src/test/java/io/stargate/db/datastore/ValidatingDataStore.java
@@ -19,10 +19,8 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 import io.stargate.db.BatchType;
 import io.stargate.db.Parameters;
-import io.stargate.db.query.BindMarker;
 import io.stargate.db.query.BoundQuery;
 import io.stargate.db.query.Query;
-import io.stargate.db.query.QueryType;
 import io.stargate.db.query.TypedValue;
 import io.stargate.db.query.TypedValue.Codec;
 import io.stargate.db.schema.Schema;
@@ -32,6 +30,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -44,10 +43,11 @@ import org.assertj.core.api.AbstractIntegerAssert;
 import org.assertj.core.api.Assertions;
 
 public class ValidatingDataStore implements DataStore {
+
   private final Schema schema;
+  private final AtomicInteger preparedIdSeq = new AtomicInteger();
   private final List<QueryExpectation> expectedQueries = new ArrayList<>();
-  private final List<Prepared<?>> prepared = new ArrayList<>();
-  private final List<ExpectedExecution> expectedExecutions = new ArrayList<>();
+  private final Map<MD5Digest, Prepared> prepared = new HashMap<>();
 
   public ValidatingDataStore(Schema schema) {
     this.schema = schema;
@@ -55,7 +55,6 @@ public class ValidatingDataStore implements DataStore {
 
   public void reset() {
     prepared.clear();
-    expectedExecutions.clear();
     expectedQueries.clear();
   }
 
@@ -64,8 +63,8 @@ public class ValidatingDataStore implements DataStore {
   }
 
   public void validate() {
-    prepared.forEach(Prepared::validate);
-    expectedExecutions.forEach(ExpectedExecution::validate);
+    prepared.values().forEach(Prepared::validate);
+    expectedQueries.forEach(QueryExpectation::validate);
   }
 
   @Override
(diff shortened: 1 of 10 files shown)


restapi/src/test/java/io/stargate/db/datastore/ValidatingDataStore.java
```diff
@@ -96,22 +96,36 @@ public boolean supportsLoggedBatches() {
...
       BoundQuery query, UnaryOperator<Parameters> parametersModifier) {
-    ExpectedExecution execution;
-    if (query instanceof ExpectedExecution) {
-      execution = (ExpectedExecution) query;
-    } else {
-      Prepared<?> prepared = (Prepared<?>) prepare(query.source().query()).join();
-      execution = (ExpectedExecution) prepared.bindValues(query.values());
+    Optional<MD5Digest> preparedId = query.source().query().preparedId();
+    if (!preparedId.isPresent()) {
+      preparedId = prepare(query.source().query()).join().preparedId();
```
tomekl007, line 114: why do we use a blocking operation (`join`) in the method that is non-blocking (returns `CompletableFuture<ResultSet>`)? Shouldn't it be fixed

tomekl007, line 114: why do we use a blocking operation (`join`) in the method that is non-blocking (returns `CompletableFuture<ResultSet>`)? Shouldn't it be fixed?

restapi/src/test/java/io/stargate/web/docsapi/service/query/search/resolver/filter/impl/InMemoryCandidatesFilterTest.java
```diff
@@ -199,7 +199,11 @@ public void fixedPath() {
           withQuery(
                   TABLE,
                   "SELECT key, leaf, text_value, dbl_value, bool_value, p0, p1, WRITETIME(leaf) FROM %s WHERE p0 = ? AND leaf = ? AND p1 = ? AND key = ? LIMIT ? ALLOW FILTERING",
-                  documentId)
+                  "field",
```
tomekl007, line 202: why do we change the `documentApi` tests in the ticket that is fixing data store API? Shouldn't those changes be separated?

restapi/src/test/java/io/stargate/db/datastore/ValidatingDataStore.java
```diff
@@ -96,22 +96,36 @@ public boolean supportsLoggedBatches() {
...
-    prepared.add(p);
-    return CompletableFuture.completedFuture(p);
+    // Generate unique IDs for each `prepare` call to validate that all prepared statements get
+    // executed.
+    ByteBuffer buffer = ByteBuffer.allocate(Integer.BYTES);
+    buffer.putInt(preparedIdSeq.incrementAndGet());
+    MD5Digest preparedId = MD5Digest.wrap(buffer.array());
+    Query<B> queryWithId = query.withPreparedId(preparedId);
+    prepared.computeIfAbsent(preparedId, p -> new Prepared<>(queryWithId));
+    return CompletableFuture.completedFuture(queryWithId);
```
tomekl007, line 106: nit: maybe extract this logic to a separate method with a name stating the intent clearly. (emphasize more what's in the comment)

restapi/src/test/java/io/stargate/db/datastore/ValidatingDataStore.java
```diff
@@ -223,9 +210,11 @@ public B bindValues(List<TypedValue> values) {
       ExpectedExecution exec = new ExpectedExecution(this, expectation);
       expectedExecutions.add(exec);
       bound = true;
-      this.source = new BoundQuery.Source<>(query, values);
-      //noinspection unchecked
-      return (B) exec;
+      return exec;
+    }
+
+    public boolean hasId(MD5Digest preparedId) {
```
tomekl007, line 216: where this method is called? I don't see any invocation.

restapi/src/test/java/io/stargate/db/datastore/ValidatingDataStore.java
dimas-b: The whole call chain in this case is owned by this test class. The `prepare` call returns a "complete" future in all cases.

However, I can probably refactor this for the sake of clarity.

dimas-b: duplicate comment?... resolving

restapi/src/test/java/io/stargate/web/docsapi/service/query/search/resolver/filter/impl/InMemoryCandidatesFilterTest.java
dimas-b: These test were incorrect. They did not validate bound variables because the bugs in `ValidatingDataStore` did not make it possible. After fixing `ValidatingDataStore` I re-added bound variable validation to the tests.

`withQuery(...)` methods are meant to match both the query text and specific parameter values so that the test code could provide different result sets for different parameters.

restapi/src/test/java/io/stargate/db/datastore/ValidatingDataStore.java
dimas-b: good catch, thx! It is not used... leftover from a previous iteration. will remove.

dimas-b: removed in the refactored code (force-pushed)

restapi/src/test/java/io/stargate/db/datastore/ValidatingDataStore.java
dimas-b: refactored (force-pushed)

restapi/src/test/java/io/stargate/db/datastore/ValidatingDataStore.java
dimas-b: refactored (force-pushed)

ivansenic (approved):
That's what I am talking about :partying_face: Thanks @dimas-b for the effort, I am gonna merge asap so I can rebase tests it in the #1080

tomekl007 (approved):

PR #1131, opened 2021-07-22
gRPC: authorize requests

olim7t:
Apply a similar approach to the CQL module: store the authenticated user in the connection's clientInfo, then serialize it into the custom payload for every request.
This allows us to leverage the existing checks in `StargateQueryHandler.authorizeByToken`.

No test yet but I think we can do something like `JwtAuthTest`. I'll push a follow-up commit.

### DIFF
--- a/grpc/src/main/java/io/stargate/grpc/service/Service.java
+++ b/grpc/src/main/java/io/stargate/grpc/service/Service.java
@@ -65,7 +65,9 @@ import java.net.SocketAddress;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
@@ -450,7 +452,9 @@ public class Service extends io.stargate.proto.StargateGrpc.StargateImplBase {
 
       connection
           .execute(
-              bindValues(handler, prepared, values), makeParameters(parameters), queryStartNanoTime)
+              bindValues(handler, prepared, values),
+              makeParameters(parameters, connection.clientInfo()),
+              queryStartNanoTime)
           .handle(
               handleQuery(
                   query,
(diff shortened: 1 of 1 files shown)


dougwettlaufer:
FYI this will do authz but not the same sort of RLAC as the other HTTP services do.

grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -742,6 +764,9 @@ private Connection newConnection(AuthenticatedUser user) {
       connection = persistence.newConnection();
     }
     connection.login(user);
+    if (user.token() != null) {
```
dimas-b, line 767: Why do we check for token to be present here? Would we not want process users that were authenticated by other means?

dimas-b:
> FYI this will do authz but not the same sort of RLAC as the other HTTP services do.

With this PR gRPC is going to be on par with CQL as far as authZ goes, right?

grpc/src/main/java/io/stargate/grpc/service/Service.java
olim7t: The persistence API can only check authorization if a token is present: https://github.com/stargate/stargate/blob/master/persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/StargateQueryHandler.java#L212

olim7t:
> > FYI this will do authz but not the same sort of RLAC as the other HTTP services do.
> 
> With this PR gRPC is going to be on par with CQL as far as authZ goes, right?

Correct. gRPC is closer to CQL than the other HTTP services: requests come in as plain query strings, so in order to perform fine-grained RLAC we'd have to parse them.

grpc/src/main/java/io/stargate/grpc/service/Service.java
dimas-b: True, but would it be possible to have unconditional logic for injecting user data into `ClientState` and let `Persistence` deal with specific user details?

olim7t:
> I think we can do something like `JwtAuthTest`

Actually no, that test delegates all checks to the internal Cassandra authorizer, so it does not cover the code we're interested in here.

I don't think we can write that test with the current test infrastructure, none of the built-in authorization services handle table-level permissions.

grpc/src/main/java/io/stargate/grpc/service/Service.java
```diff
@@ -742,6 +764,9 @@ private Connection newConnection(AuthenticatedUser user) {
...
+      connection.clientInfo().ifPresent(c -> c.setAuthenticatedUser(user));
```
mpenick, line 768: I know this is not really part of supporting auth in grpc directly, but shouldn't `connection.login()` set the authenticated user of the `ClientInfo` if it's present? It seems strange to have to set it as a separate state.

olim7t: Yeah, I copied what was already in place for CQL, but indeed it seems like this could be done directly in `login()`.

mpenick (approved):

PR #1134, opened 2021-07-22
Add selectivity hints to Docs API query filters

### LINKED ISSUE #1072: Enable user to explicitly specify filter priority in order to control the execution
Once we have two or more filters with the same internal priority (for example two filters that we can execute against C*), we fail back to the user given order of the condition (first wins). However, with supporting `$or` and `$and`, this is now related to the tree iteration order of conditions and it's not intuitive anymore. 

We should support adding of the `$selectivity` parameter along side the condition in order to give user the explicit control. For example, `{ field: {$eq: 1, $selectivity: 1000} }`. When `$selectivity` is not specified fallback to the current approach.

dimas-b:
* Allow "$selectivity" to be used as a key in JSON objects
  defining a single filter. The expected value is a number.

* Selectivity hints in JSON objects defining multiple filters
  are not supported to avoid ambiguity in case the filters
  later get separated during transformations of the boolean
  expression that contains them.

Fixes #1072

Contributes to #828

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
@@ -22,10 +22,13 @@ import com.bpodgursky.jbool_expressions.Literal;
 import com.bpodgursky.jbool_expressions.Or;
 import com.bpodgursky.jbool_expressions.rules.RuleSet;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.collect.Streams;
 import io.stargate.web.docsapi.exception.ErrorCode;
 import io.stargate.web.docsapi.exception.ErrorCodeRuntimeException;
+import io.stargate.web.docsapi.service.query.ImmutableFilterExpression.Builder;
 import io.stargate.web.docsapi.service.query.condition.BaseCondition;
 import io.stargate.web.docsapi.service.query.condition.ConditionParser;
+import io.stargate.web.docsapi.service.query.filter.operation.FilterHintCode;
 import io.stargate.web.docsapi.service.util.DocsApiUtils;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -35,6 +38,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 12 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
```diff
@@ -177,6 +216,17 @@ private void validateFieldConditions(
               specificTypes.stream().map(Class::getSimpleName).collect(Collectors.joining(", ")));
       throw new ErrorCodeRuntimeException(ErrorCode.DOCS_API_SEARCH_FILTER_INVALID, msg);
     }
+
+    // If selectivity is explicitly provided, allow at most one condition to avoid ambiguity.
+    if (selectivity.isPresent() && conditions.size() > 1) {
```
ivansenic, line 221: but multiple are anyway wrapped by and, I would gave them all same selectivity. WDYT? But maybe this is even better.. Let's keep it like this..

ivansenic (approved):
This is really perfect, I can not add anything, except two advises:

* add int test with $selectivity only for backward compatibility
* add additional test in the `DocumentServiceTest` with selectivity winning over the index and asserting in the profile

restapi/src/main/java/io/stargate/web/docsapi/service/query/ExpressionParser.java
dimas-b: Yes, but I'm not sure it is going to be obvious to the user :thinking: from my POV selectivity of an AND should be better than the selectivity of its sub-conditions, so using the same value there may be confusing.... I'd prefer to error out in that case and let the user define explicit values for each condition.

dimas-b:
Tests added. Also added a minor fix in `DocumentService` (bug exposed by the new tests)

restapi/src/main/java/io/stargate/web/docsapi/service/query/filter/operation/FilterHintCode.java
```diff
@@ -0,0 +1,45 @@
...
+
+/** Enumerates tags for query hints. */
+public enum FilterHintCode {
+
+  /**
+   * Filter selectivity.
+   *
+   * @see FilterExpression#getSelectivity()
+   */
+  SELECTIVITY("$selectivity");
```
dougwettlaufer, line 32: Can we get this added to the swagger in https://github.com/stargate/stargate/blob/master/restapi/src/main/java/io/stargate/web/docsapi/resources/ReactiveDocumentResourceV2.java#L292 and create an issue in stargate/docs to make sure full documentation doesn't get lost?

dimas-b: sure. Sorry, I missed that. Follow-up PR: #1138

dimas-b: Docs issue: https://github.com/stargate/docs/issues/144

PR #1137, opened 2021-07-26
closes #1063: full support for $or queries in the docs api

### LINKED ISSUE #1063: Support `$or` document search
After having parsing for `$or` (#1031) and after having support for merging queries (#1091) we can fully support ORs. Note that #1080 will add the support for `$or` in the in-document search.
**Support this**:

* `a = 1 && (b = 2 || c = 3) && (d = 4 || e = 5)`
*  `(b = 2 || c = 3) && (b = 2 || e = 5)` -> because it resolves to `b = 2 && (c = 3 || e = 5)`
* `(b = 2 || c = 3)`
* `(b = 2 || c = 3) && (d = 4 || e = 5)`

ivansenic:
Full support for the `$or` conditions when searching for the documents in the complete collection.

This is fully functional and done, but we need to agree on two final questions:

1. Can we do a optimization to run merging queries with different columns. This would save us both data we are fetching (less columns), as well as time to evaluate in some cases (although evaluation is fast).
2. Do we want to have implemented order of resolving conditions in the CNF form in the following order:
    a. Single persistence filter
    b. `OR` filters (order by "best")
    c. Single in-memory filter

Missing tasks:

* [x] Complete unit test for the `OrExpressionDocumentsResolver`
* [x] Add one INT test for paging
* [x] Fix codacy
* [x] Create documentation ticket once merged -> https://github.com/stargate/docs/issues/145
* [x] blocked by https://github.com/stargate/stargate/issues/1162

@polandll Lorina this is a new feature that needs to be documented. Simple example  would be 
`collection?where={\"$or\": [{\"price\": {\"$lt\": 200}}, {\"discountPrice\": {\"$lt\": 150}}]}`.

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
@@ -102,8 +102,7 @@ public enum ErrorCode {
       Response.Status.BAD_REQUEST, "Search was expecting a JSON object as input."),
 
   DOCS_API_SEARCH_EXPRESSION_NOT_RESOLVED(
-      Response.Status.INTERNAL_SERVER_ERROR,
-      "Can not find the document resolver for the given expression."),
+      Response.Status.INTERNAL_SERVER_ERROR, "Unable to resolve the given expression."),
 
   DOCS_API_SEARCH_RESULTS_NOT_FITTING(
       Response.Status.BAD_REQUEST,
--- a/restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
@@ -101,9 +101,9 @@ public enum ErrorCode {
   DOCS_API_SEARCH_OBJECT_REQUIRED(
       Response.Status.BAD_REQUEST, "Search was expecting a JSON object as input."),
 
-  DOCS_API_SEARCH_OR_NOT_SUPPORTED(
-      Response.Status.BAD_REQUEST,
-      "Searching documents with the $or condition is not yet supported."),
+  DOCS_API_SEARCH_EXPRESSION_NOT_RESOLVED(
+      Response.Status.INTERNAL_SERVER_ERROR,
+      "Can not find the document resolver for the given expression."),
 
   DOCS_API_SEARCH_RESULTS_NOT_FITTING(
       Response.Status.BAD_REQUEST,
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/query/rules/TrueFilterExpressions.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/query/rules/TrueFilterExpressions.java
@@ -39,7 +39,6 @@ public class TrueFilterExpressions extends Rule<Expression<FilterExpression>, Fi
   @Override
   public Expression<FilterExpression> applyInternal(
       Expression<FilterExpression> input, ExprOptions<FilterExpression> options) {
-    // cast OK because of the isApply
     boolean test = truePredicate.test(input);
     if (test) {
       return Literal.getTrue();
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/query/rules/TrueFilterExpressions.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/query/rules/TrueFilterExpressions.java
@@ -25,14 +25,14 @@ import io.stargate.web.docsapi.service.query.FilterExpression;
 import java.util.function.Predicate;
 
 /**
- * A {@link Rule} that coverts any {@link FilterExpression} matching the given predicate to {@link
- * Literal#getTrue()}.
+ * A {@link Rule} that coverts any {@link Expression<FilterExpression>} matching the given predicate
+ * to {@link Literal#getTrue()}.
  */
 public class TrueFilterExpressions extends Rule<Expression<FilterExpression>, FilterExpression> {
 
-  private final Predicate<FilterExpression> truePredicate;
+  private final Predicate<Expression<FilterExpression>> truePredicate;
 
-  public TrueFilterExpressions(Predicate<FilterExpression> truePredicate) {
+  public TrueFilterExpressions(Predicate<Expression<FilterExpression>> truePredicate) {
     this.truePredicate = truePredicate;
   }
 
@@ -40,7 +40,7 @@ public class TrueFilterExpressions extends Rule<Expression<FilterExpression>, Fi
   public Expression<FilterExpression> applyInternal(
       Expression<FilterExpression> input, ExprOptions<FilterExpression> options) {
     // cast OK because of the isApply
-    boolean test = truePredicate.test((FilterExpression) input);
+    boolean test = truePredicate.test(input);
     if (test) {
       return Literal.getTrue();
     } else {
@@ -51,6 +51,6 @@ public class TrueFilterExpressions extends Rule<Expression<FilterExpression>, Fi
   // only instance of FilterExpression
   @Override
   protected boolean isApply(Expression<FilterExpression> input) {
-    return input instanceof FilterExpression;
+    return true;
   }
 }
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/query/search/resolver/impl/AnyFiltersResolver.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/query/search/resolver/impl/AnyFiltersResolver.java
@@ -2,7 +2,7 @@
  * Copyright The Stargate Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in comHpliance with the License.
+ * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
  * http://www.apache.org/licenses/LICENSE-2.0
@@ -29,33 +29,27 @@ import java.util.List;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
-public class AllFiltersResolver extends AbstractFiltersResolver {
+public class AnyFiltersResolver extends AbstractFiltersResolver {
 
   private final Collection<CandidatesFilter> candidatesFilters;
 
   private final DocumentsResolver candidatesResolver;
 
-  public AllFiltersResolver(
+  public AnyFiltersResolver(
       Function<ExecutionContext, CandidatesFilter> candidatesFilterSupplier,
       ExecutionContext context,
       DocumentsResolver candidatesResolver) {
     this(Collections.singleton(candidatesFilterSupplier), context, candidatesResolver);
   }
 
-  public AllFiltersResolver(
+  public AnyFiltersResolver(
       Collection<Function<ExecutionContext, CandidatesFilter>> candidatesFilterSuppliers,
       ExecutionContext context,
       DocumentsResolver candidatesResolver) {
-    ExecutionContext nested = context.nested("PARALLEL [ALL OF]");
+    ExecutionContext nested = context.nested("PARALLEL [ANY OF]");
     this.candidatesFilters =
         candidatesFilterSuppliers.stream().map(s -> s.apply(nested)).collect(Collectors.toList());
     this.candidatesResolver = candidatesResolver;
-  }
-
-  /** {@inheritDoc} */
-  @Override
-  protected Collection<CandidatesFilter> getCandidatesFilters() {
-    return candidatesFilters;
   }
 
   /** {@inheritDoc} */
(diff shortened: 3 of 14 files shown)


dimas-b:
> Can we do a optimization to run merging queries with different columns. This would save us both data we are fetching (less columns), as well as time to evaluate in some cases (although evaluation is fast).

I think this optimization is important if the two queries target different fields. Even though evaluation is fast, Stargate may have to fetch a lot of data into memory if the condition is not pushed down to storage. Since we index all values, using the index at storage level is probably more efficient.

If the conditions target the _same_ field, the optimization of pushing down one of the predicates is probably going to be overshadowed by having to fetch all rows for the second predicate (the one that must be evaluated in memory).

dimas-b:
> Do we want to have implemented order of resolving conditions in the CNF form in the following order:
> a. Single persistence filter
> b. OR filters (order by "best")
> c. Single in-memory filter

LGTM. I assume `a` means that we can simplify the expression to this form: `Filter1 AND (any expression)` and `b` means we can simplify it to `(Filter2 OR Filter3) AND (any expression)`, right?

So, in case we have an expression like `Filter1 AND (Filter1 OR Filter2) AND (any expression)` then, we could estimate the selectivity of the OR as the worst selectivity of its components and them chose the AND sub-expression with the lowest selectivity. WDYT?

If selectivity hints are not provided, then, it does look reasonable to execute `Filter1` first.

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
```diff
@@ -101,9 +101,9 @@
   DOCS_API_SEARCH_OBJECT_REQUIRED(
       Response.Status.BAD_REQUEST, "Search was expecting a JSON object as input."),
 
-  DOCS_API_SEARCH_OR_NOT_SUPPORTED(
-      Response.Status.BAD_REQUEST,
-      "Searching documents with the $or condition is not yet supported."),
+  DOCS_API_SEARCH_EXPRESSION_NOT_RESOLVED(
+      Response.Status.INTERNAL_SERVER_ERROR,
+      "Can not find the document resolver for the given expression."),
```
dimas-b, line 106: IMHO, the new message is a bit cryptic for the end-user. How about something like `Unable to resolve the given expression`?

Also, is it really a server error? I'd say if we're not able to process something because of a feature that is know to be not implemented, and we can detect that condition, we ought to report `BAD_REQUEST`. WDYT?

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/resolver/impl/AnyFiltersResolver.java
```diff
@@ -0,0 +1,74 @@
...
+  @Override
+  protected Collection<CandidatesFilter> getCandidatesFilters() {
+    return candidatesFilters;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  protected Flowable<RawDocument> resolveSources(RawDocument rawDocument, List<Maybe<?>> sources) {
+    // only one signal is needed here, we can dispose the rest immediately
+    // when one emits, map to the document
```
dimas-b, line 71: nit: Could you add a comment about why we map to the original document? IIRC, it is to preserve pagination order of the higher-level query.

restapi/src/main/java/io/stargate/web/docsapi/service/query/rules/TrueFilterExpressions.java
```diff
@@ -25,22 +25,22 @@
...
 
-  public TrueFilterExpressions(Predicate<FilterExpression> truePredicate) {
+  public TrueFilterExpressions(Predicate<Expression<FilterExpression>> truePredicate) {
     this.truePredicate = truePredicate;
   }
 
   @Override
   public Expression<FilterExpression> applyInternal(
       Expression<FilterExpression> input, ExprOptions<FilterExpression> options) {
     // cast OK because of the isApply
```
dimas-b, line 42: nit: there's no cast anymore

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/resolver/impl/OrExpressionDocumentsResolver.java
```diff
@@ -0,0 +1,189 @@
...
+            })
+        .filter(
+            doc -> {
+              // now we can get doc as a result of the persistence or in memory query
+              // if we can locate that it was a result of the persistence ones, then nothing is
+              // needed anymore
+              // we can do this by finding any row that does not have any path related column
+              // TODO enable if solved in the query executor
+              // boolean noPaths = doc.rows().stream().anyMatch(r ->
+              // !r.columnExists(QueryConstants.P_COLUMN_NAME.apply(0)));
```
dimas-b, line 121: As discussed, I think we can determine whether a doc was produced by a qurry with a "persistence" predicate by validating whether it has any of the `*_value` columns.

Alternatively, and this may actually be cleaner, we could update `QueryExecutor` to link `RawDocument` objects to  queries that provided rows for the doc. In that case, if any of the queries with "persistence" predicates contributed to the doc, we do not have to evaluate rules in memory.

I suppose we could introduce a boolean flag on `RawDocument`, e.g. `matched`, then add a function parameter to `QueryExecutor` for mapping `Query` object to the `matched` flag, and have `QueryExecutor` invoke that function on every row collected (inside `PagingStateTracker.track()` probably) and aggregate the final value for the `matched` flag. WDYT?

By comparison, using rows with different selected column sets is risky, because rows differing only in `*_value` columns are considered "equal" by the merge logic and there's no telling which of those equal rows will be stored in the doc object.

dimas-b: On a second though, we can probably do this optimization later.

For now, if query with "persistence" predicates return `*_value` columns anyway, re-evaluating their expression in memory is a bit of extra work, but it should not change the filter outcome, right? So, it might be ok, to merge as is for now, and optimize in a follow-up PR.

OR conditions are a new feature, so having a bit of extra perf. overhead initially is probably ok. WDYT?

dimas-b (changes requested):
PR LGTM, nice work :)

Let's clarify and remove the TODO comments and I think this PR is good to merge.

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/resolver/impl/AnyFiltersResolver.java
ivansenic: I am gonna update to:

```
Maybe<?> resolveSources(List<Maybe<?>> sources)
```

and the abstract class can deal with the mapping to the doc..

ivansenic: Ok this will not work, adding comment..

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
ivansenic: This is really a server exception and should be reported like this so that we can look into it.. Should never happen, but I need to throw something here.. I'll update the message.. :+1:

ivansenic:
> > Do we want to have implemented order of resolving conditions in the CNF form in the following order:
> > a. Single persistence filter
> > b. OR filters (order by "best")
> > c. Single in-memory filter
> 
> LGTM. I assume `a` means that we can simplify the expression to this form: `Filter1 AND (any expression)` and `b` means we can simplify it to `(Filter2 OR Filter3) AND (any expression)`, right?
> 
> So, in case we have an expression like `Filter1 AND (Filter2 OR Filter3) AND (any expression)` then, we could estimate the selectivity of the OR as the worst selectivity of its components and then chose the AND sub-expression with the lowest selectivity. WDYT?
> 
> If selectivity hints are not provided, then, it does look reasonable to execute `Filter1` first.
> 
> I think we can do the selectivity-based fine tuning in a follow-up PR. For now we can prefer single filter AND sub-expressions.

Hmm, not sure if I can follow here.. Let's quickly sync and decide what do to now and create a ticket for the selectivity optimizations.

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/resolver/impl/AnyFiltersResolver.java
```diff
@@ -0,0 +1,75 @@
...
+    return candidatesFilters;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  protected Flowable<RawDocument> resolveSources(RawDocument rawDocument, List<Maybe<?>> sources) {
+    // only one signal is needed here, we can dispose the rest immediately
+    // when one emits, map to the document because we need to return the document that passes
+    // it's important to keep with the concatMap approach in the abstract class
+    return Maybe.concat(sources).take(1).map(any -> rawDocument);
```
dimas-b, line 73: I guess this will wait for each source _in sequence_... How about `Flowable.merge(sources).take(1)` instead? That should complete as soon as any source produce an item... may want to set the concurrently parameter to match the number of sources. WDYT?

ivansenic: Done. Merge takes the `Integer.MAX_VALUE` for the concurrency, so we are fine there..

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/resolver/impl/OrExpressionDocumentsResolver.java
```diff
@@ -0,0 +1,180 @@
...
+            })
+        .filter(
+            doc -> {
+              // now we can get doc as a result of the persistence or in memory query
+              // if we only run persistence queries we can return true immediately
+              // running only persistence queries means we had no path columns
+              // since all rows must have same columns we can find any that does not have it
+              boolean noPaths =
+                  doc.rows().stream()
+                      .anyMatch(r -> !r.columnExists(QueryConstants.P_COLUMN_NAME.apply(0)));
```
dimas-b, line 115: nit: I believe we can precompute this flag outside of the Rx pipeline (for the sake of optimization)

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -2347,6 +2347,247 @@ public void searchArrayWildcardExists() throws Exception {
...
+    String r =
+        RestUtils.get(
+            authToken,
+            hostWithPort
+                + "/v2/namespaces/"
+                + keyspace
+                + "/collections/collection?page-size=1&where={\"$or\": [{\"value\": {\"$eq\": \"a\"}}, {\"value\": {\"$in\": [\"b\"]}}]}",
+            200);
+
+    String expected = "{\"matching1\":{\"value\":\"a\"}}";
```
dimas-b, line 2442: Can we be sure `a` always comes back first? I believe it ultimately depends on the C* partitioner and the doc ID... could make the test unstable.

I'd suggest to run the same query twice, make sure we get both `a` and `b`, but without depending on the order of those results. WDYT?

dimas-b (approved):

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
ivansenic: Well all other tests with paging are the same and are expecting a certain order of elements.. Not sure if it's key what's the order here, but let's keep it until we hit some issues.

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/resolver/impl/OrExpressionDocumentsResolver.java
ivansenic: Good catch. Done.

ivansenic:
The DSE build fails with 
```
Step #4 - "dse-6.8": [ERROR] Failures: 
Step #4 - "dse-6.8": [ERROR]   DocumentApiV2Test>BaseDocumentApiV2Test.searchOrMixedFilterWithPaging:2461 
Step #4 - "dse-6.8": Expecting:
Step #4 - "dse-6.8":  <[]>
Step #4 - "dse-6.8": to be equal to:
Step #4 - "dse-6.8":  <[[[]]]>
Step #4 - "dse-6.8": but was not.
Step #4 - "dse-6.8": [INFO] 
Step #4 - "dse-6.8": [ERROR] Tests run: 920, Failures: 1, Errors: 0, Skipped: 10
```

The commit that added this method has been green yesterday.. I did rebase on the latest master today.. I'll try to debug locally..

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
dimas-b: Ok

ivansenic:
@dimas-b unfortunately seems there is an issue with the merging page state in DSE..

* I am getting correct document and paging state in the first query :heavy_check_mark: 
* However, no rows are returned using this state later on the the second query :warning: 

Since you are using the `CombinedPagingState` to use the state, can it be that I need to produce the state for the user in the same way? Right now I was not changing this and was using the `String state = Paginator.makeExternalPagingState(paginator, rawDocuments);`.. 

Any help is appreciated.

PR #1141, opened 2021-07-27
[WIP] fix dse ci to use the latest builder image version

ivansenic:
Trying to see if `1.0.4` will work..

Fails with `1.0.4` as well, mabe the problem was introduced in #733 this change, when we updated the DSE version to the `6.8.5` to `6.8.9`. 

And seems that there was already a revert from `1.0.3` to `1.0.4` in https://github.com/stargate/stargate/commit/f8073debd799b8bcabd2cc029b6dfb3a69ae3f18.


ivansenic:
Closing in favor of #1155

Issue #1071, opened 2021-06-28
Support `$not` for the collection search


ivansenic:
Add a possibility to add `$not` as the filter operation for the document API. This should effectively negate anything that is wrapped in the `$not`.

ivansenic:
@dimas-b Don't we need the #1027 done first here? Or you have another idea?

dimas-b:
Yes, for some reason I assumed #1027 was already done :thinking:

dimas-b:
However, since we never push `$exists` down to persistence, this issue may be doable before #1027 :thinking:

PR #1150, opened 2021-07-28
closes #1027: support `$exists`  with false

### LINKED ISSUE #1027: Support $exists with false

ivansenic:
Supports `$exists: false` for the docs api search. This is a quick fix, because it's a blocker for the #1071. 

It's implemented in the way that if `false` is given, this filter condition becomes in-memory one. The behavior with `true` is not changed and we can still leverage the C query here. 

@dimas-b I implemented this so you can focus on the #1071.
@polandll This is an extension of the existing feature

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/impl/ExistsCondition.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/impl/ExistsCondition.java
@@ -18,8 +18,6 @@ package io.stargate.web.docsapi.service.query.condition.impl;
 
 import io.stargate.db.datastore.Row;
 import io.stargate.db.query.builder.BuiltCondition;
-import io.stargate.web.docsapi.exception.ErrorCode;
-import io.stargate.web.docsapi.exception.ErrorCodeRuntimeException;
 import io.stargate.web.docsapi.service.query.condition.BaseCondition;
 import io.stargate.web.docsapi.service.query.filter.operation.FilterOperationCode;
 import java.util.Optional;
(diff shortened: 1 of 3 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/impl/ExistsCondition.java
```diff
@@ -70,13 +56,18 @@ public FilterOperationCode getFilterOperationCode() {
...
-    return false;
+    return !getQueryValue();
   }
 
   /** {@inheritDoc} */
   @Override
   public boolean test(Row row) {
-    // always test true when row is there :)
-    return true;
+    // if row exists then the test is true if query value is true
```
tomekl007, line 65: could you elaborate more what's the use case for this "convoluted" logic?

ivansenic: This is just to be "conceptually correct". We will never call `test(null)`, however, I wanted to ensure that the logic behind the condition is mapped here.

But I can also revert this and assume row is never null (as we actually do in the other conditions). WDYT?

tomekl007: shouldn't we just return false in case `row == null`?

dimas-b: I guess `row` is never `null` here, except when test code calls it.

However, a `null` row means the row _does not exist_ so if the condition is `$exists: false`, it makes sense to return `true`, I think.

dimas-b (approved):
Thanks @ivansenic ! :)

dimas-b:
#1153 should fix the DSE CI errors. Let's merge it first and rebase.

dougwettlaufer (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/impl/ExistsCondition.java
ivansenic: Since all the other conditions assume `Row` is not null, I will apply the same logic here..

Issue #1145, opened 2021-07-27
gRPC: Waiting for schema agreement should be asynchronous 


mpenick:
Maybe not critical, but waiting for schema agreement after DDL queries currently uses `Persistence.waitForSchemaAgreement()` which synchronously waits (via `Uninterruptibles.sleepUninterruptibly()`) which blocks gRPC request processing event loops. Ideally, this would use an asynchronous mechanism to allow event loops to continue processing other requests.

ivansenic:
Can this be connected to #1151. Not sure if we have there the DDL queries..

PR #1152, opened 2021-07-28
gRPC: refactor Service class to simplify control flow (fixes #1148)

### LINKED ISSUE #1148: gRPC: refactor main service class to simplify control flow
There are a few things we could do to make the code easier to follow:

* try to limit the use of `whenComplete()`, in order to reduce the duplication of error handling branches. This was alluded to in [this comment](https://github.com/stargate/stargate/pull/1132#discussion_r675932362) (although it won't be as trivial as the example given there).
* I think it might also be helpful to introduce a temporary handler object for each query: `new QueryHandler(query, connection, responseObserver, ....).executeAsync()`. This will allow us to store final state as fields in this object, instead of passing it to every method along the chain.

olim7t:
Draft implementation to test the ideas suggested in #1148.

The idea is that `Service.executeQuery()` would call `new QueryHandler(...).handle()` (this is not wired yet).

If we agree to pursue this, I'll do the same for batches, wire both handlers, and most of the code in `Service` should go away.

### DIFF
--- a/grpc/src/main/java/io/stargate/grpc/service/BatchHandler.java
+++ b/grpc/src/main/java/io/stargate/grpc/service/BatchHandler.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.grpc.service;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import io.grpc.stub.StreamObserver;
+import io.stargate.db.Persistence;
+import io.stargate.db.Persistence.Connection;
+import io.stargate.db.Result.Prepared;
+import io.stargate.grpc.service.Service.PrepareInfo;
+import io.stargate.proto.QueryOuterClass.Response;
+import java.util.concurrent.CompletionStage;
+
+class BatchHandler extends MessageHandler {
+
+  protected BatchHandler(
+      Connection connection,
+      Cache<PrepareInfo, CompletionStage<Prepared>> preparedCache,
+      Persistence persistence,
+      StreamObserver<Response> responseObserver) {
+    super(connection, preparedCache, persistence, responseObserver);
+  }
+
+  // TODO
+}
--- a/grpc/src/main/java/io/stargate/grpc/service/BatchHandler.java
+++ b/grpc/src/main/java/io/stargate/grpc/service/BatchHandler.java
@@ -16,23 +16,252 @@
 package io.stargate.grpc.service;
 
 import com.github.benmanes.caffeine.cache.Cache;
+import com.google.protobuf.Any;
+import io.grpc.Status;
 import io.grpc.stub.StreamObserver;
+import io.stargate.db.AuthenticatedUser;
+import io.stargate.db.BatchType;
+import io.stargate.db.ClientInfo;
+import io.stargate.db.ImmutableParameters;
+import io.stargate.db.Parameters;
 import io.stargate.db.Persistence;
 import io.stargate.db.Persistence.Connection;
+import io.stargate.db.Result;
 import io.stargate.db.Result.Prepared;
+import io.stargate.db.Statement;
+import io.stargate.grpc.payload.PayloadHandler;
+import io.stargate.grpc.payload.PayloadHandlers;
 import io.stargate.grpc.service.Service.PrepareInfo;
+import io.stargate.grpc.service.Service.ResponseAndTraceId;
+import io.stargate.proto.QueryOuterClass.Batch;
+import io.stargate.proto.QueryOuterClass.BatchParameters;
+import io.stargate.proto.QueryOuterClass.BatchQuery;
+import io.stargate.proto.QueryOuterClass.Payload;
 import io.stargate.proto.QueryOuterClass.Response;
+import java.nio.ByteBuffer;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.apache.cassandra.stargate.db.ConsistencyLevel;
 
 class BatchHandler extends MessageHandler {
 
-  protected BatchHandler(
+  /** The maximum number of batch queries to prepare simultaneously. */
+  private static final int MAX_CONCURRENT_PREPARES_FOR_BATCH =
+      Math.max(Integer.getInteger("stargate.grpc.max_concurrent_prepares_for_batch", 1), 1);
+
+  private final Batch batch;
+
+  BatchHandler(
+      Batch batch,
       Connection connection,
       Cache<PrepareInfo, CompletionStage<Prepared>> preparedCache,
       Persistence persistence,
       StreamObserver<Response> responseObserver) {
     super(connection, preparedCache, persistence, responseObserver);
+    this.batch = batch;
   }
 
-  // TODO
+  void handle() {
+    CompletionStage<Result> resultFuture =
+        CompletableFuture.<Void>completedFuture(null)
+            .thenApply(this::validateQueries)
+            .thenCompose(__ -> new BatchPreparer().prepare())
+            .thenCompose(this::executePrepared);
+    resultFuture = handleUnprepared(resultFuture, this::reprepareAndRetry);
+    resultFuture
+        .thenApply(this::buildResponse)
+        .thenCompose(this::executeTracingQueryIfNeeded)
+        .whenComplete(
+            (response, error) -> {
+              if (error != null) {
+                handleException(error);
+              } else {
+                setSuccess(response);
+              }
+            });
+  }
+
+  private Void validateQueries(Void ignored) {
+    if (batch.getQueriesCount() == 0) {
+      throw new CompletionException(
+          Status.INVALID_ARGUMENT.withDescription("No queries in batch").asException());
+    }
+
+    Payload.Type type = batch.getQueries(0).getValues().getType();
+    boolean allTypesMatch =
+        batch.getQueriesList().stream().allMatch(v -> v.getValues().getType().equals(type));
+    if (!allTypesMatch) {
+      throw new CompletionException(
+          Status.INVALID_ARGUMENT
+              .withDescription(
+                  "Types for all queries within batch must be the same, and equal to: " + type)
+              .asException());
+    }
+    return null;
+  }
+
+  private CompletionStage<Result> executePrepared(io.stargate.db.Batch preparedBatch) {
+    long queryStartNanoTime = System.nanoTime();
+    BatchParameters parameters = batch.getParameters();
+    try {
+      return connection.batch(
+          preparedBatch, makeParameters(parameters, connection.clientInfo()), queryStartNanoTime);
+    } catch (Exception e) {
+      return failedFuture(e);
+    }
+  }
+
+  private CompletionStage<Result> reprepareAndRetry() {
+    return new BatchPreparer().prepareForRetry().thenCompose(this::executePrepared);
+  }
+
+  private ResponseAndTraceId buildResponse(Result result) {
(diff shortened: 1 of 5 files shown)


grpc/src/main/java/io/stargate/grpc/service/MessageHandler.java
```diff
@@ -0,0 +1,305 @@
...
+import org.apache.cassandra.stargate.exceptions.CasWriteUnknownResultException;
+import org.apache.cassandra.stargate.exceptions.FunctionExecutionException;
+import org.apache.cassandra.stargate.exceptions.PersistenceException;
+import org.apache.cassandra.stargate.exceptions.ReadFailureException;
+import org.apache.cassandra.stargate.exceptions.ReadTimeoutException;
+import org.apache.cassandra.stargate.exceptions.UnavailableException;
+import org.apache.cassandra.stargate.exceptions.WriteFailureException;
+import org.apache.cassandra.stargate.exceptions.WriteTimeoutException;
+
+abstract class MessageHandler {
```
olim7t, line 46: Shared logic that is common to query and batch. 
There's possibly stuff currently in `QueryHandler` that will eventually be pulled up here.

grpc/src/main/java/io/stargate/grpc/service/MessageHandler.java
```diff
@@ -0,0 +1,305 @@
...
+      ProtoUtils.keyForProto(QueryOuterClass.FunctionFailure.getDefaultInstance());
+  private static final Metadata.Key<QueryOuterClass.WriteFailure> WRITE_FAILURE_KEY =
+      ProtoUtils.keyForProto(QueryOuterClass.WriteFailure.getDefaultInstance());
+  private static final Metadata.Key<QueryOuterClass.AlreadyExists> ALREADY_EXISTS_KEY =
+      ProtoUtils.keyForProto(QueryOuterClass.AlreadyExists.getDefaultInstance());
+  private static final Metadata.Key<QueryOuterClass.CasWriteUnknown> CAS_WRITE_UNKNOWN_KEY =
+      ProtoUtils.keyForProto(QueryOuterClass.CasWriteUnknown.getDefaultInstance());
+
+  protected final Connection connection;
+  private final Cache<PrepareInfo, CompletionStage<Prepared>> preparedCache;
```
olim7t, line 66: Storing futures in the cache allows us to prepare exactly once if multiple threads are racing on the same statement (see below).

grpc/src/main/java/io/stargate/grpc/service/MessageHandler.java
```diff
@@ -0,0 +1,305 @@
...
+      Cache<PrepareInfo, CompletionStage<Prepared>> preparedCache,
+      Persistence persistence,
+      StreamObserver<Response> responseObserver) {
+    this.connection = connection;
+    this.preparedCache = preparedCache;
+    this.persistence = persistence;
+    this.responseObserver = responseObserver;
+  }
+
+  protected CompletionStage<Prepared> prepare(PrepareInfo prepareInfo) {
```
olim7t, line 81: I'm not propagating the `tracing` flag here anymore. In the `Service` implementation it's actually not used (we do send it to the persistence, but don't read the traces).
We want to trace the execution of the prepared statement, not the preparation itself.

grpc/src/main/java/io/stargate/grpc/service/QueryHandler.java
```diff
@@ -0,0 +1,258 @@
...
+    }
+  }
+
+  private CompletionStage<Result> reprepareAndRetry() {
+    return prepare(prepareInfo, true).thenCompose(this::executePrepared);
+  }
+
+  private ResponseAndTraceId buildResponse(Result result) {
+    ResponseAndTraceId responseAndTraceId = new ResponseAndTraceId();
+    responseAndTraceId.setTracingId(result.getTracingId());
```
olim7t, line 116: The original code also checks that the tracing flag is set in `query.getParameters()`. I think we can reasonably assume that the response contains an id if and only if the client asked for it.

grpc/src/main/java/io/stargate/grpc/service/QueryHandler.java
```diff
@@ -0,0 +1,258 @@
...
+        : new QueryTracingFetcher(responseAndTraceId.tracingId, connection, getTracingConsistency())
+            .fetch()
+            .handle(
+                (traces, error) -> {
+                  if (error == null) {
+                    responseBuilder.setTraces(
+                        TraceEventsMapper.toTraceEvents(
+                            traces, responseBuilder.getTraces().getId()));
+                  }
+                  // If error != null, ignore and still return the main result with an empty trace
```
olim7t, line 162: This is a departure from the original implementation, but it sounds better to me.

tomekl007: Do you mean to return a response without tracing? That sounds like a sane idea +1

grpc/src/main/java/io/stargate/grpc/service/BatchHandler.java
```diff
@@ -0,0 +1,38 @@
...
+import com.github.benmanes.caffeine.cache.Cache;
+import io.grpc.stub.StreamObserver;
+import io.stargate.db.Persistence;
+import io.stargate.db.Persistence.Connection;
+import io.stargate.db.Result.Prepared;
+import io.stargate.grpc.service.Service.PrepareInfo;
+import io.stargate.proto.QueryOuterClass.Response;
+import java.util.concurrent.CompletionStage;
+
+class BatchHandler extends MessageHandler {
```
tomekl007, line 27: Probably there will be even more common parts between `BatchHandler` and `QueryHandler`.

Side note: The one problem that I see is that `QueryParameters` and `BatchParameters` don't share the same interface nor inherit from the same pattern (as far as I know, it is not possible in protobuf).
We could reduce even more duplication if that would be possible.

tomekl007 (commented):
Very good refactoring. I like this direction and separate handlers per batch and query.

grpc/src/main/java/io/stargate/grpc/service/QueryHandler.java
olim7t: Yes: if the trace query fails, we still have the main result, so don't fail the whole request.

olim7t:
> I'll do the same for batches, wire both handlers, and most of the code in `Service` should go away.

Done. Marking this ready for review since all the tests pass.

The two `handle()` methods are almost similar, there's an opportunity to push up even more stuff into `MessageHandler` but I'm not sure if it's worth it. I'll try this afternoon and we can revert it if it's too complicated.

grpc/src/main/java/io/stargate/grpc/service/MessageHandler.java
```diff
@@ -0,0 +1,455 @@
...
+    this.connection = connection;
+    this.preparedCache = preparedCache;
+    this.persistence = persistence;
+    this.responseObserver = responseObserver;
+  }
+
+  void handle() {
+    CompletionStage<Result> resultFuture =
+        CompletableFuture.<Void>completedFuture(null)
+            .thenApply(this::invokeValidate)
```
olim7t, line 105: This is a bit ugly, but it allows us to handle all errors with a single `handleException` at the end of the future chain.

grpc/src/main/java/io/stargate/grpc/service/QueryHandler.java
mpenick: Good catch.

grpc/src/main/java/io/stargate/grpc/service/QueryHandler.java
```diff
@@ -0,0 +1,167 @@
...
+  @Override
+  protected ResponseAndTraceId buildResponse(Result result) {
+    ResponseAndTraceId responseAndTraceId = new ResponseAndTraceId();
+    responseAndTraceId.setTracingId(result.getTracingId());
+    Response.Builder responseBuilder = makeResponseBuilder(result);
+    switch (result.kind) {
+      case Void:
+        break;
+      case SchemaChange:
+        // TODO make this non-blocking?
```
mpenick, line 98: https://github.com/stargate/stargate/issues/1145

grpc/src/main/java/io/stargate/grpc/service/BatchHandler.java
```diff
@@ -0,0 +1,232 @@
...
+   * Concurrently prepares queries in a batch. It'll prepare up to {@link
+   * #MAX_CONCURRENT_PREPARES_FOR_BATCH} queries simultaneously.
+   */
+  class BatchPreparer {
+
+    private final AtomicInteger queryIndex = new AtomicInteger();
+    private final List<Statement> statements = new CopyOnWriteArrayList<>();
+    private final CompletableFuture<io.stargate.db.Batch> future = new CompletableFuture<>();
+
+    CompletionStage<io.stargate.db.Batch> prepare() {
```
mpenick, line 169: `prepare()` looks to be dead code.

grpc/src/main/java/io/stargate/grpc/service/BatchHandler.java
```diff
@@ -0,0 +1,232 @@
...
+      return prepare(false);
+    }
+
+    CompletionStage<io.stargate.db.Batch> prepareForRetry() {
```
mpenick, line 173: Same here, dead code.

grpc/src/main/java/io/stargate/grpc/service/MessageHandler.java
```diff
@@ -0,0 +1,455 @@
...
+            .thenCompose(__ -> prepare(false))
+            .thenCompose(this::executePrepared);
+    resultFuture = handleUnprepared(resultFuture, this::reprepareAndRetry);
```
mpenick, line 108: `this::reprepareAndRetry` is always the second parameter. Maybe remove the second parameter?

grpc/src/main/java/io/stargate/grpc/service/BatchHandler.java
olim7t: Right, I meant to delete those.

grpc/src/main/java/io/stargate/grpc/service/MessageHandler.java
mpenick: I think I'd prefer something like the following even though we have duplicate calls to `handleException()`:

```
    try {
      validate();

      CompletionStage<Result> resultCompletionStage = prepare(false)
          .thenCompose(this::executePrepared);

      resultCompletionStage = handleUnprepared(resultCompletionStage);
      resultCompletionStage
          .thenApply(this::buildResponse)
          .thenCompose(this::executeTracingQueryIfNeeded)
          .whenComplete(
              (response, error) -> {
                if (error != null) {
                  handleException(error);
                } else {
                  setSuccess(response);
                }
              });
    } catch (Throwable t) {
      handleException(t);
    }
```

1) It's simpler; it doesn't require a temporary future and we don't have to have the method `invokeValidate()`.
2) It's a bit safer because if any of the async setup code throws it'll be handled and won't hang the request. The code might be exception safe now until someone adds something in the future that's not.

grpc/src/main/java/io/stargate/grpc/service/MessageHandler.java
mpenick: Nice fix.

grpc/src/main/java/io/stargate/grpc/service/MessageHandler.java
```diff
@@ -0,0 +1,451 @@
...
+    return values.hasData()
+        ? handler.bindValues(prepared, values.getData(), persistence.unsetValue())
+        : new BoundStatement(prepared.statementId, Collections.emptyList(), null);
+  }
+
+  protected CompletionStage<Prepared> prepare(PrepareInfo prepareInfo, boolean shouldInvalidate) {
+    // In the event a query is being retried due to a PreparedQueryNotFoundException invalidate the
+    // local cache to refresh with the remote cache
+    if (shouldInvalidate) {
+      preparedCache.invalidate(prepareInfo);
```
mpenick, line 163: Any worry about invalidating entries we just replaced?

grpc/src/main/java/io/stargate/grpc/service/MessageHandler.java
olim7t: OK, I was a bit on the fence about this, I don't mind changing it back.

grpc/src/main/java/io/stargate/grpc/service/MessageHandler.java
olim7t: I see, one problematic scenario would be:
```
[thread1] prepare, cache hit
[thread1] execute, receive UNPREPARED
[thread2] prepare, cache hit
[thread2] execute, receive UNPREPARED
[thread1] invalidate, put new future in cache, reprepare
[thread2] invalidate, put new future in cache, reprepare ==> wasteful, new entry was already there
```

We could keep a reference to the entry we got on the first hit, and only invalidate if it's still in the cache by then.

But maybe that's a bit overkill, it complicates the code, and repreparing twice is just slightly wasteful, it does not introduce any bug.

mpenick (approved):
:+1: Great refactor, simpler and fixes 2 major issues (prepare cache race and tracing query failure).

PR #1153, opened 2021-07-28
Do not block Rx operations when deleting "dead leaves" in Docs API

### LINKED ISSUE #1151: Possible lock in the reactive document service RX chain
When getting a document and performing a dead leaves deletion, we might end-up in the dead-lock (or some other kind of issue) inside the `CoreThread`s that are used to actually communicate with DSE. This results in timeout error and is shown in the `DocumentApiV2Test.testPutReplacingWithArray` test.

I managed to grab the thread dump in the middle of the test when we are actually waiting for the result of dead leaves deletion. The full log is [here](https://gist.githubusercontent.com/ivansenic/eece0dffdd7b48b90e8915808c3ba532/raw/9683a785cf09dc89089c43e7113a92313adca430/gistfile1.txt).

As we can see, the log the `[CoreThread-4]` deletion of the dead leaves. However in the thread dump the thread is `parking to wait for  <0x000000073ed5d7b0>`. And it stays there until timeout is signaled. Right after we start the leave deletion, the message:

`WARN  [GossipStage:1] 2021-07-28 15:38:55,293 StorageService.java:3061 - /127.0.0.1 has the same server ID C0-3C-59-BF-C0-ED, and same DC dc1, but is placed in a different rack r1 - (/127.0.2.1, dc1, rack1) vs. (/127.0.0.1, dc1, r1)`

However, in the CI when the test fails the above message is not shown and instead we have:

```
Step #4 - "dse-6.8": INFO  [Exec Stream Pumper] 2021-07-28 10:00:33,633 ProcessRunner.java:78 - storage0-0> INFO  [OptionalTasks:1] 2021-07-28 10:00:33,633  Gossiper.java:546 - Destroying messaging connection to /127.0.2.33 due to endpoint being removed from cluster
Step #4 - "dse-6.8": INFO  [Exec Stream Pumper] 2021-07-28 10:00:56,392 ProcessRunner.java:78 - stargate33-0> INFO  [OptionalTasks:1] 2021-07-28 10:00:56,392 Gossiper.java:546 - Destroying messaging connection to /127.0.2.33 due to endpoint being removed from cluster
```

Any help is welcome.

dimas-b:
* Make DocumentDB.deleteDeadLeaves non-blocking

* Add a separate method to DocumentDB for authorizing
  "dead leaf" deletion separately from the deleteDeadLeaves
  call.

* Make the Rx pipeline wait for the DELETE batch submission
  in a non-blocking way (no user-level behaviour change)

Fixes #1151

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
@@ -29,7 +29,6 @@ import io.reactivex.rxjava3.core.Single;
 import io.stargate.auth.AuthenticationSubject;
 import io.stargate.auth.AuthorizationService;
 import io.stargate.auth.SourceAPI;
-import io.stargate.auth.UnauthorizedException;
 import io.stargate.db.datastore.Row;
 import io.stargate.web.docsapi.dao.DocumentDB;
 import io.stargate.web.docsapi.dao.Paginator;
@@ -44,6 +43,7 @@ import io.stargate.web.docsapi.service.query.ExpressionParser;
 import io.stargate.web.docsapi.service.query.FilterExpression;
 import io.stargate.web.docsapi.service.query.FilterPath;
 import io.stargate.web.docsapi.service.util.DocsApiUtils;
+import io.stargate.web.rx.RxUtils;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
@@ -188,6 +188,10 @@ public class ReactiveDocumentService {
           authorizationService.authorizeDataRead(
               authenticationSubject, namespace, collection, SourceAPI.REST);
 
+          // Don't fail this read request if the corrective DELETE statements are not authorized,
+          // simply skip DELETE batches in that case (see the Rx pipeline below).
+          boolean deleteAuthorized = db.authorizeDeleteDeadLeaves(namespace, collection);
+
           // we need to check that we have no globs and transform the stuff to support array
           // elements
           List<String> subDocumentPathProcessed = processSubDocumentPath(subDocumentPath);
@@ -224,21 +228,32 @@ public class ReactiveDocumentService {
                             false,
                             db.treatBooleansAsNumeric());
 
+                    Maybe<?> deleteBatch;
+
                     // dead leaf deletion init on non-empty collection
-                    if (!collector.isEmpty()) {
+                    if (deleteAuthorized && !collector.isEmpty()) {
                       logger.info(
                           String.format("Deleting %d dead leaves", collector.getLeaves().size()));
 
-                      // don't fail here if the unauthorized exception occurs, this is a read
-                      // request
-                      try {
-                        db.deleteDeadLeaves(
-                            namespace, collection, documentId, collector.getLeaves(), context, now);
-                      } catch (UnauthorizedException e) {
-                        logger.debug("Unauthorized when trying to delete dead leaves.", e);
-                      }
+                      deleteBatch =
+                          Maybe.fromSingle(
+                              RxUtils.singleFromFuture(
+                                  () ->
+                                      db.deleteDeadLeaves(
+                                          namespace,
+                                          collection,
+                                          documentId,
+                                          now,
+                                          collector.getLeaves(),
+                                          context)));
+                    } else {
+                      deleteBatch = Maybe.just("delete not issued");
                     }
 
+                    return deleteBatch.map(__ -> docsResult);
+                  })
+              .flatMap(
+                  docsResult -> {
                     // create json pattern expression if sub path is defined
                     if (!subDocumentPath.isEmpty()) {
                       String jsonPtrExpr =
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
@@ -24,8 +24,11 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.google.common.annotations.VisibleForTesting;
 import io.reactivex.rxjava3.core.Maybe;
 import io.reactivex.rxjava3.core.Single;
+import io.reactivex.rxjava3.disposables.Disposable;
+import io.reactivex.rxjava3.schedulers.Schedulers;
 import io.stargate.auth.AuthenticationSubject;
 import io.stargate.auth.AuthorizationService;
 import io.stargate.auth.SourceAPI;
@@ -52,6 +55,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
+import org.javatuples.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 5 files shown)


testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -561,7 +562,7 @@ public void testPutReplacingArrayElement() throws IOException {
     assertThat(OBJECT_MAPPER.readTree(r)).isEqualTo(wrapResponse(fullObjNode, "1", null));
   }
 
-  @Test
+  @RepeatedTest(1000) // TODO: remove repeats once CI passes
```
dimas-b, line 565: This change is to be reverted before merging.

dougwettlaufer: Looks like CI passed. Good to revert?

dougwettlaufer (approved):

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
dimas-b: Nice :) reverted

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -188,6 +188,10 @@ public ReactiveDocumentService(
           authorizationService.authorizeDataRead(
               authenticationSubject, namespace, collection, SourceAPI.REST);
 
+          // Don't fail this read request if the corrective DELETE statements are not authorized,
+          // simply skip DELETE batches in that case (see the Rx pipeline below).
+          boolean deleteAuthorized = db.authorizeDeleteDeadLeaves(namespace, collection);
```
ivansenic, line 193: I would not have this out of the pipe and would do it only if we need to delete the leaves, otherwise we add extra latency to each call..

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -224,21 +228,32 @@ public ReactiveDocumentService(
...
 
-                      // don't fail here if the unauthorized exception occurs, this is a read
-                      // request
-                      try {
-                        db.deleteDeadLeaves(
-                            namespace, collection, documentId, collector.getLeaves(), context, now);
-                      } catch (UnauthorizedException e) {
-                        logger.debug("Unauthorized when trying to delete dead leaves.", e);
-                      }
+                      deleteBatch =
```
ivansenic, line 238: I would not wait for the deletion, the user issued a GET and we can simply do the deletion in the async way, non blocking on the response.. If it fails, the read would still be successful. We save both on time and on potential error:

```
Single.fromCallable(() -> db.authorizeDeleteDeadLeaves(namespace, collection))
   .filter(authenticated -> authenticated)
   .observeOn(Schedulers.io())
   .subscibeOn(Schedulers.io())
   .flatMap(_ -> RxUtils.singleFromFuture(() ->
                          db.deleteDeadLeaves(
                              namespace,
                              collection,
                              documentId,
                              now,
                              collector.getLeaves(),
                              context))
   .subscribe((r, t) -> {
    // log result or throwable
  })
   
```

ivansenic (changes requested):
Please have a look on the proposed changes.

I am still worried that the source of the problem is not solved. We are executing the RX pipe line on the CoreThreads, tomorrow we would hit similar problem again. I would strongly advise to move to the RX threads once we have the data from the C. I will create a PR for that as well, we can discuss further in that PR.

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
ivansenic: I provided the patch so we can quicker resolve this once you are online..


```
diff --git a/restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java b/restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
index 2b0e44ff..2caaebea 100644
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
@@ -26,6 +26,7 @@ import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.reactivex.rxjava3.core.Maybe;
 import io.reactivex.rxjava3.core.Single;
+import io.reactivex.rxjava3.schedulers.Schedulers;
 import io.stargate.auth.AuthenticationSubject;
 import io.stargate.auth.AuthorizationService;
 import io.stargate.auth.SourceAPI;
@@ -188,8 +189,6 @@ public class ReactiveDocumentService {
           authorizationService.authorizeDataRead(
               authenticationSubject, namespace, collection, SourceAPI.REST);
 
-          // Don't fail this read request if the corrective DELETE statements are not authorized,
-          // simply skip DELETE batches in that case (see the Rx pipeline below).
           boolean deleteAuthorized = db.authorizeDeleteDeadLeaves(namespace, collection);
 
           // we need to check that we have no globs and transform the stuff to support array
@@ -228,32 +227,44 @@ public class ReactiveDocumentService {
                             false,
                             db.treatBooleansAsNumeric());
 
-                    Maybe<?> deleteBatch;
-
                     // dead leaf deletion init on non-empty collection
-                    if (deleteAuthorized && !collector.isEmpty()) {
-                      logger.info(
-                          String.format("Deleting %d dead leaves", collector.getLeaves().size()));
-
-                      deleteBatch =
-                          Maybe.fromSingle(
-                              RxUtils.singleFromFuture(
-                                  () ->
-                                      db.deleteDeadLeaves(
-                                          namespace,
-                                          collection,
-                                          documentId,
-                                          now,
-                                          collector.getLeaves(),
-                                          context)));
-                    } else {
-                      deleteBatch = Maybe.just("delete not issued");
+                    if (!collector.isEmpty()) {
+                      // Don't fail this read request if the corrective DELETE statements are not
+                      // authorized,
+                      // simply skip DELETE batches in that case (see the Rx pipeline below).
+                      Single.fromCallable(() -> db.authorizeDeleteDeadLeaves(namespace, collection))
+                          .filter(
+                              auth -> {
+                                if (auth) {
+                                  logger.info(
+                                      String.format(
+                                          "Deleting %d dead leaves", collector.getLeaves().size()));
+                                } else {
+                                  logger.info(
+                                      String.format(
+                                          "No authority to delete %d dead leaves",
+                                          collector.getLeaves().size()));
+                                }
+                                return auth;
+                              })
+                          .observeOn(Schedulers.io())
+                          .subscribeOn(Schedulers.io())
+                          .flatMapSingle(
+                              any ->
+                                  RxUtils.singleFromFuture(
+                                      () ->
+                                          db.deleteDeadLeaves(
+                                              namespace,
+                                              collection,
+                                              documentId,
+                                              now,
+                                              collector.getLeaves(),
+                                              context)))
+                          .subscribe(
+                              r -> logger.info("Dead leaves successfully deleted."),
+                              t -> logger.warn("Error deleting dead leaves.", t));
                     }
 
-                    return deleteBatch.map(__ -> docsResult);
-                  })
-              .flatMap(
-                  docsResult -> {
                     // create json pattern expression if sub path is defined
                     if (!subDocumentPath.isEmpty()) {
                       String jsonPtrExpr =
```

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
dimas-b: good point, but can we defer that until the execution is moved to non-TPC (non-core) threads? Auth calls can block, I guess. In any case they are not guaranteed to be non-blocking, and we should never block DSE core threads.

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
dimas-b: I did not want to change existing behaviour wrt surfacing DELETE errors, but you have a point. Since the DELETEs use explicit timestamps, and we only wait for the batch to be _accepted_ (it may not have been fully executed by the time we continue), running the DELETEs asynchronously is probably ok.

dimas-b: Implemented that with some minor tweaks in 0bd2d386913abe87601f1174d7ac742a57d558cc

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
dimas-b: resolved by 0bd2d386913abe87601f1174d7ac742a57d558cc

ivansenic:
Codacy complains, tried to ignore issue and re-analyze, but it's not working.. I am merging this so we can continue forward.

ivansenic (approved):

PR #1155, opened 2021-07-28
WIP: using stargate-builder:v1.0.5 for DSE

dimas-b:


ivansenic:
Pushed one commit to read the seed address from the backend config.

ivansenic:
OK my changes are irrelevant, the failure is occurring in the external storage already.. I am testing from the image itself now.

ivansenic:
@dimas-b I spent considerable amount of time on this today, and can it be that the `bionic` image was upgraded since we builded and that this command is not valid anymore:

```
mvn -B verify --file pom.xml \
-P \${C3}it-cassandra-3.11 \
-P \${C4}it-cassandra-4.0 \
-P \${DSE}dse -P \${DSE}it-dse-6.8 \
-P default,grpc \
-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
```

Meaning that `\${C3}it-cassandra-3.11` is not valid. In the image bash I am getting this:

```
ubuntu@20309e2200f1:/stargate$ echo  ${C4}it-cassandra-4.0
!it-cassandra-4.0
ubuntu@20309e2200f1:/stargate$ echo  "${C4}it-cassandra-4.0"
!it-cassandra-4.0
ubuntu@20309e2200f1:/stargate$ echo  \${C4}it-cassandra-4.0 
${C4}it-cassandra-4.0
```

I have no idea why the `\` was there at all? @dougwettlaufer 

Also, is there a need to run the CassandraPersistenceIT when we are building for the dse? This does not seem write. We run it both for the C 3.11 and C 4.0, but if the `it-dse-6.8` profile kicks in and sets `ccm.dse=true` how can these tests even be valid to run?

dimas-b:
@ivansenic : that scriptlet goes through a couple of reinterpretation rounds - that's why we need to escape `$` chars - they should be interpreted by the inner shell.

dimas-b:
> is there a need to run the CassandraPersistenceIT when we are building for the dse? This does not seem right

It should _not_ run. That's what those `-P` options are supposed to control.

ivansenic:
> > is there a need to run the CassandraPersistenceIT when we are building for the dse? This does not seem right
> 
> It should _not_ run. That's what those `-P` options are supposed to control.

It seems they are not running.. Only the CassandraPersistenceTest`s are, but those are different. OK I am out of ideas here..

ivansenic:
Closing in favor of https://github.com/stargate/stargate/pull/1521

PR #1158, opened 2021-07-28
Support $not operators in Docs API WHERE clauses

### LINKED ISSUE #1071: Support `$not` for the collection search

dimas-b:
* Parse $not operators in JSON allowing exactly one
  sub-expression.

* Add negation methods to filter expressions and
  conditions

* Add a rule for eliminating NOT boolean expressions
  by negating the child filter during expression
  resolution.

* Simplify expressions after eliminating negation

Note: Also remove selectivity from the identity of
FilterExpression. This is to give preference to boolean
expression simplification when equivalent filters happen
to have different selectivity hints specified by the user.

Fixes #1071

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterExpression.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterExpression.java
@@ -65,6 +65,10 @@ public abstract class FilterExpression extends Expression<FilterExpression>
    * (the worst possible selectivity).
    */
   @Value.Default
+  // Do not use selectivity for equality comparisons. This is critical to allow expression
+  // simplification (see `ExpressionUtilsTest`) in cases where filters do not have proper
+  // selectivity hints defined (hints are optional).
+  @Value.Auxiliary
   public double getSelectivity() {
     return 1.0;
   }
@@ -76,6 +80,16 @@ public abstract class FilterExpression extends Expression<FilterExpression>
         .condition(condition)
         .orderIndex(orderIndex)
         .selectivity(selectivity)
+        .build();
+  }
+
+  public FilterExpression negate() {
+    return ImmutableFilterExpression.builder()
+        .filterPath(getFilterPath())
+        .condition(getCondition().negate())
+        .orderIndex(getOrderIndex())
+        // The negated filter will select rows complementing the ones selected by source filter
+        .selectivity(1.0 - getSelectivity())
         .build();
   }
 
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/BaseCondition.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/BaseCondition.java
@@ -27,6 +27,10 @@ import javax.validation.constraints.NotNull;
 /** Interface for the base filtering condition. */
 public interface BaseCondition extends Predicate<Row> {
 
+  /** Returns a {@link BaseCondition} that is the logical negation of this condition. */
+  @Override
+  BaseCondition negate();
+
   /**
    * @return If this condition can be executed on the persistence level. The default implementation
    *     resolves to true if the {@link #getBuiltCondition()} returns non-empty value.
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/query/rules/ExpressionUtils.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/query/rules/ExpressionUtils.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.stargate.web.docsapi.service.query.rules;
+
+import com.bpodgursky.jbool_expressions.Expression;
+import com.bpodgursky.jbool_expressions.rules.RuleSet;
+import io.stargate.web.docsapi.service.query.FilterExpression;
+
+public class ExpressionUtils {
+
+  /**
+   * Converts the given expression to CNF and eliminates boolean NOT operators but inverting the
+   * negated {@link FilterExpression}s.
+   *
+   * <p>Note: If selectivity hints are not defined on all filters, the simplification process may
+   * lose user-provided hints. This can be avoided by the user by providing hints on all filters or
+   * by simplifying the expression up front.
+   */
+  public static Expression<FilterExpression> toSimplifiedCnf(
+      Expression<FilterExpression> expression) {
+    // Note: this will push NOTs down to the leaf (FilterExpression) level
+    Expression<FilterExpression> cnf = RuleSet.toCNF(expression);
+
+    // Resolve NOTs by inverting negated FilterExpressions
+    Expression<FilterExpression> negationResolved = NegateFilters.resolve(cnf);
+
+    // Re-simplify the expression after eliminating NOTs
+    return RuleSet.simplify(negationResolved);
+  }
+}
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/query/rules/NegateFilters.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/query/rules/NegateFilters.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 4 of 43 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/query/search/resolver/BaseResolver.java
```diff
@@ -61,6 +62,8 @@ public static DocumentsResolver resolve(
     // execute cnf optimization
     Expression<FilterExpression> cnf = RuleSet.toCNF(expression);
 
+    cnf = NegateFilters.resolve(cnf);
```
ivansenic, line 65: I would add this rule to the expression parsing and immediately simplify there. We could potentially even eliminate some of the "stupid" conditions like `eq: x | not(neq: x)`. WDYT? 

I guess there is no need to take not out of the parsing..

ivansenic (changes requested):
Looks great and clean :+1:

But I strongly opt for optimization when parsing and not later. That's why I am requesting changes, so that we can sync on that..

restapi/src/main/java/io/stargate/web/docsapi/service/query/rules/NegateFilters.java
```diff
@@ -0,0 +1,64 @@
...
+    return child.negate();
+  }
+
+  @Override
+  protected boolean isApply(Expression<FilterExpression> input) {
+    if (!(input instanceof Not)) {
+      return false;
+    }
+
+    Not<FilterExpression> negated = (Not<FilterExpression>) input;
```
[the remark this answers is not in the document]
ivansenic: That said I would really like to have an INT test that tackles this situation!

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/BaseCondition.java
```diff
@@ -27,6 +29,12 @@
 /** Interface for the base filtering condition. */
 public interface BaseCondition extends Predicate<Row> {
 
+  @Override
+  default BaseCondition negate() {
```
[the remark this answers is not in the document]
dimas-b: We need it to narrow the return type. The base method returns `Predicate`. I can probably keep it abstract in `BaseCondition`, though.

restapi/src/main/java/io/stargate/web/docsapi/service/query/rules/NegateFilters.java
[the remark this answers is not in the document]
dimas-b: The intent behind this rule is to convert something like `Not(GT(1))` into `LTE(1)`. So, the rule only accepts `Not` expression that have a `FilterExpression` as a direct child.

It should work both for CNF and DNF as those forms have `NOT`s pushed down to the predicate (variable) level.

This rule is a bit useless during expression simplification. Granted it can simplify some cases (e.g. `And(Not(GT(1))`), but it will not be able to eliminate `NOT`s completely, e.g. in `Not(And(GT(1), LT(2))`. Conversion to CNF or DNF is required to push the negation to predicate level in that case (simplification alone is not sufficient).

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/resolver/BaseResolver.java
dimas-b: Please see my comment on `NegateFilters` (above) regarding simplification.

Regarding the `eq: x | not(neq: x)` example, the conversion to CNF includes simplification rules, so that should work already. I'll see if I can add a test case for that.

dimas-b: Refactored the conversion code to support simplification after NOTs are eliminated (6db3bf767cc5a7918170c99be9ed2c1dddef9ba7)

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/BaseCondition.java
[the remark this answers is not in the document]
ivansenic: That's what I mean, remove `default` and make abstract.

restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterExpression.java
```diff
@@ -65,6 +65,7 @@
    * (the worst possible selectivity).
    */
   @Value.Default
+  @Value.Auxiliary // do not use selectivity for equality comparisons
```
ivansenic, line 68: Why not?

restapi/src/main/java/io/stargate/web/docsapi/service/query/rules/ExpressionUtils.java
```diff
@@ -0,0 +1,37 @@
...
+
+package io.stargate.web.docsapi.service.query.rules;
+
+import com.bpodgursky.jbool_expressions.Expression;
+import com.bpodgursky.jbool_expressions.rules.RuleSet;
+import io.stargate.web.docsapi.service.query.FilterExpression;
+
+public class ExpressionUtils {
+
+  public static Expression<FilterExpression> toSimplifiedCnf(
```
ivansenic, line 26: small java doc please to explain what's this for, especially because method name does not specify anything about negating..

restapi/src/test/java/io/stargate/web/docsapi/service/query/rules/ExpressionUtilsTest.java
```diff
@@ -0,0 +1,199 @@
...
+import io.stargate.web.docsapi.service.query.condition.impl.ImmutableNumberCondition;
+import io.stargate.web.docsapi.service.query.filter.operation.FilterOperationCode;
+import io.stargate.web.docsapi.service.query.filter.operation.ValueFilterOperation;
+import io.stargate.web.docsapi.service.query.filter.operation.impl.GtFilterOperation;
+import io.stargate.web.docsapi.service.query.filter.operation.impl.LteFilterOperation;
+import java.util.Collections;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+class ExpressionUtilsTest {
```
ivansenic, line 43: hell yea :1st_place_medal:

restapi/src/test/java/io/stargate/web/docsapi/service/query/rules/ExpressionUtilsTest.java
```diff
@@ -0,0 +1,199 @@
...
+      assertThat(cnf.toLexicographicString())
+          .isEqualTo("((field GT 2 | field LTE 3) & field LTE 1)");
+    }
+
+    @Test
+    void orOfAndWithNegation() {
+      Expression<FilterExpression> cnf =
+          toSimplifiedCnf(Or.of(Not.of(e(LTE, 1)), e(GT, 2), And.of(e(GT, 3), e(LTE, 4))));
+      // Note: the first predicate got inverted
+      // Note: predicate-specific simplification is not supported (yet?), as in:
```
ivansenic, line 114: let's add a ticket for this.. should be supper easy to add a rule for this..

ivansenic (approved):
This is awesome stuff.. I am merging the #1137 so we could also add a test for not with ors in the `BaseResolverTest`. That and rebasing would be it imo..

restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterExpression.java
dimas-b: Sorry, I put the rationale into git commit message - it's to allow expression simplification in cases where only some of the filter have proper selectivity hints, e.g. `{field: {$eq: 1}, field: {$not: {$ne: 1}}` will not simplify because by default selectivity is `1.0` but inverting the filter makes it `0.0`, so we end up with `{field: {$eq: 1, $selectivity: 1.0}, field: {$eq: 1, $selectivity: 0.0}}`. Those need to compare equal for the simplification rules to kick in.

dimas-b: Of course, we could lose user-provided hints this way, but I think it is still better than to miss simplification.

restapi/src/test/java/io/stargate/web/docsapi/service/query/rules/ExpressionUtilsTest.java
dimas-b: #1165

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/resolver/impl/OrExpressionDocumentsResolver.java
```diff
@@ -186,6 +186,7 @@ public OrExpressionDocumentsResolver(Or<FilterExpression> expression, ExecutionC
 
   private ExecutionContext createContext(
       ExecutionContext context, Or<FilterExpression> expression) {
-    return context.nested("MERGING OR: expression '" + expression + "'");
+    // Note: use toLexicographicString to ensure a stable order of sub-expressions.
+    return context.nested("MERGING OR: expression '" + expression.toLexicographicString() + "'");
```
dimas-b, line 190: This is needed to unbreak the unit tests now that the `.toString()` representation became unstable due to expression transformations in `ExpressionUtils`.

restapi/src/main/java/io/stargate/web/docsapi/service/query/FilterExpression.java
dimas-b: Added direct comments about this in the rebased version.

restapi/src/main/java/io/stargate/web/docsapi/service/query/rules/ExpressionUtils.java
dimas-b: Javadoc added

restapi/src/main/java/io/stargate/web/docsapi/service/query/condition/BaseCondition.java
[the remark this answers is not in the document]
dimas-b: Removed the default method body in the rebased version. Please review again. There were some extra changes related to ORs.

ivansenic (approved):
Looks great, merging..

PR #1159, opened 2021-07-29
relates to #1151: switch away from the core thread in the RX

ivansenic:
For me this is tackling the root cause for the #1151. We should switch away from the `CoreThread` after that data has been loaded from the C*. This would help us avoid problems like the one observed in future and release us from thinking what's done in the RX sequence and if there could be issues. In addition, we are releasing the `CoreThread` early.

The fix proposed by @dimas-b in the #1153 can be complimentary to this one if we decide to not-block on the dead leaves deletion..

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
@@ -24,6 +24,7 @@ import hu.akarnokd.rxjava3.operators.FlowableTransformers;
 import hu.akarnokd.rxjava3.operators.Flowables;
 import io.reactivex.rxjava3.core.Flowable;
 import io.reactivex.rxjava3.core.Single;
+import io.reactivex.rxjava3.schedulers.Schedulers;
 import io.stargate.db.ComparableKey;
 import io.stargate.db.ImmutableParameters;
 import io.stargate.db.PagingPosition;
@@ -212,6 +213,7 @@ public class QueryExecutor {
                 });
 
     return RxUtils.singleFromFuture(supplier)
+        .observeOn(Schedulers.io())
         .toFlowable()
         .compose(FlowableConnectOnRequest.with()) // separate subscription from query execution
         .take(1);
--- a/testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
+++ b/testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
@@ -40,6 +40,7 @@ import okhttp3.Response;
 import okhttp3.ResponseBody;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.RepeatedTest;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 
@@ -561,7 +562,7 @@ public abstract class BaseDocumentApiV2Test extends BaseOsgiIntegrationTest {
     assertThat(OBJECT_MAPPER.readTree(r)).isEqualTo(wrapResponse(fullObjNode, "1", null));
   }
 
-  @Test
+  @RepeatedTest(1000)
   public void testPutReplacingWithArray() throws IOException {
     JsonNode fullObj =
         OBJECT_MAPPER.readTree(this.getClass().getClassLoader().getResource("example.json"));
--- a/testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
+++ b/testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
@@ -40,7 +40,6 @@ import okhttp3.Response;
 import okhttp3.ResponseBody;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.RepeatedTest;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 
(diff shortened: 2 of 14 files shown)


testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -561,7 +562,7 @@ public void testPutReplacingArrayElement() throws IOException {
     assertThat(OBJECT_MAPPER.readTree(r)).isEqualTo(wrapResponse(fullObjNode, "1", null));
   }
 
-  @Test
+  @RepeatedTest(1000)
```
ivansenic, line 565: To be reverted after CI.

ivansenic:
This is failing the unit test as the `TestSubscriber` is missing he `await()` call now before actually doing any assertions.. I would update the tests, let's just agree if we wanna do this at the end.

@dimas-b :arrow_up:

ivansenic:
Agreed with @dimas-b to fix unit tests and go for small perf examination to understand overhead for the thread switch. However, we concluded that `CoreThread` could be even used to write to the AsyncResponse, which would not be optimal.

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
```diff
@@ -212,6 +213,7 @@ public QueryExecutor(DataStore dataStore) {
                 });
 
     return RxUtils.singleFromFuture(supplier)
+        .observeOn(Schedulers.io())
```
dimas-b, line 216: As discussed, we should certainly clean up thread management in Docs API, but this small change may not be sufficient.

I suppose C* OSS should generally work fine regardless of how we switch threads. The DSE use cases are probably going to be the most tricky, though.

Points to consider:
* DSE core (aka TPC) threads should never be blocked by I/O or `.sleep()` or `CompletableFuture.join()`, etc.
* Blocking I/O should be performed on another thread pool. This probably means we have to write HTTP responses on another thread pool or at least verify whether the response can be written without block from a DSE core thread. I'm not sure how exactly this works with async servlets. Please double check.
* `DataStore` queries will run callbacks on DSE core threads initially. If we have to switch to another thread pool, we should be careful with performance implications since DSE core thread generally occupy almost all CPUs under load (CPU count - 1, IIRC).
* If we use another thread pool it might be worth adding controls for min/max pool size parameters (to avoid creating too many threads under load)

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
ivansenic: Reverted.

ivansenic:
@mpenick @dougwettlaufer @EricBorczuk This is ready for reviewing. I tried to finalize the discussion Dmitri and I had before he left with a quick nosqlbench test I did today. The results are in the [comment above](https://github.com/stargate/stargate/pull/1159#discussion_r686827054).

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
dimas-b: Nice results @ivansenic !

I believe it would also be valuable to review perf. under high load, though. I'm not sure what the overall CPU utilization was on the test system(s) above, but in general DSE-specific behaviours are likely to surface under high load on the coordinator, plus multiple storage nodes.

Also, single (or few) rows reads are likely to be fulfilled with a single inter-node message, which will probably not trigger any on the deep DSE-specific behaviours. It might be worth testing with complex `WHERE` conditions that require range scans or full scans to have a more complete picture.

ivansenic: I'll create tests and add results for using the `where`. I'll target even larger documents and match many of them (max `page-size`).

ivansenic: Run a quick search scenario as well:

**Setup:**
* Local test
* Stargate single node, with DSE single node on same machine
* Measuring search of 10k documents, each having 23 fields (rows in the database)
* First filters grabs app. 5% of the document, second grabs app. 10% of the candidates
* Results (not that absolute numbers don't matter here, this was run with 120 threads all on the single node)
   * master:  p90 ~6.95s, p99.9 ~11.89s
   * 1151:  p90 ~6.95ms, p99.9 ~13.68ms

*master*
image

*#1151*
image

ivansenic:
@dougwettlaufer @EricBorczuk @mpenick @tatu-at-datastax @olim7t Can anybody please approve this?

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
dimas-b: One more point, sorry for keeping this thread long :)

Have you considered switching to the I/O scheduler at the end of the chain? I suppose it is only really required when we _write_ the HTTP response. Filtering and posting nested queries can probably be done on the thread that processes previous query results (i.e. DSE core threads). I suppose it might help to reduce the number of context switches in highly filtered cases. WDYT?

dougwettlaufer (approved):

EricBorczuk (approved):
Based on the results you posted in thread, this looks good! I was just wondering about Dmitri's most recent comment, about switching to IO scheduler at the end of the chain; also this is really not something I'm very familiar with so I don't want to merge this outright

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
ivansenic: @dimas-b The HTTP write was not my only concern, it's also to protect that we don't do anything in the core threads that can lead to the problems we had seen with the delete leafs.. You are most likely write, perf might be best if we do the switch prior to the HTTP write, but I am also looking for that safety.

PR #1160, opened 2021-07-29
Support more complex types grpc

tomekl007:
Provides support for C* `varint` and `decimal` in the gRPC API.

| Cassandra Type  | Java Type | gRPC Type |
| ------------- | ------------- | ------------- |
| Decimal | BigDecimal | Decimal { uint32 scale, bytes value} |
| Varint | BigInteger | Varint { bytes value } |

### DIFF
--- a/grpc-proto/proto/query.proto
+++ b/grpc-proto/proto/query.proto
@@ -59,6 +59,11 @@ message BigInteger {
   bytes value = 1;
 }
 
+message BigDecimal {
+  uint32 scale = 1;
+  bytes value = 3;
+}
+
 message Value {
   message Null {}
   message Unset {}
@@ -108,6 +113,9 @@ message Value {
 
     // CQL types: varint
     BigInteger varint = 14;
+
+    // CQL types: decimal
+    BigDecimal decimal = 15;
   }
 }
 
--- a/grpc-proto/proto/query.proto
+++ b/grpc-proto/proto/query.proto
@@ -55,11 +55,11 @@ message Uuid {
   fixed64 lsb = 2;
 }
 
-message BigInteger {
+message Varint {
   bytes value = 1;
 }
 
-message BigDecimal {
+message Decimal {
   uint32 scale = 1;
   bytes value = 3;
 }
@@ -112,10 +112,10 @@ message Value {
     UdtValue udt = 13;
 
     // CQL types: varint
-    BigInteger varint = 14;
+    Varint varint = 14;
 
     // CQL types: decimal
-    BigDecimal decimal = 15;
+    Decimal decimal = 15;
   }
 }
 
--- a/grpc-proto/proto/query.proto
+++ b/grpc-proto/proto/query.proto
@@ -55,6 +55,10 @@ message Uuid {
   fixed64 lsb = 2;
 }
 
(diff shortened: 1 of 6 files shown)


grpc/src/main/java/io/stargate/grpc/codec/cql/VarintCodec.java
```diff
@@ -0,0 +1,46 @@
...
+import java.nio.ByteBuffer;
+
+public class VarintCodec implements ValueCodec {
+  @Override
+  public ByteBuffer encode(@NonNull Value value, @NonNull ColumnType type) {
+    if (value.getInnerCase() != InnerCase.VARINT) {
+      throw new IllegalArgumentException("Expected varint type");
+    }
+    BigInteger varint = value.getVarint();
+    return ByteBuffer.wrap(varint.getValue().toByteArray());
```
tomekl007, line 33: we could leverage the java-driver codecs(`TypeCodecs.VARINT`):
```
  @Override
  public ByteBuffer encode(@NonNull Value value, @NonNull ColumnType type) {
    if (value.getInnerCase() != InnerCase.VARINT) {
      throw new IllegalArgumentException("Expected varint type");
    }
    BigInteger bigInteger = new BigInteger(value.getVarint().toByteArray());
    return TypeCodecs.VARINT.encode(bigInteger, PROTOCOL_VERSION);
  }
```
but It would mean that we need to create a `BigInteger` instance only for the purpose of encoding. I am not sure if that is a good value from the performance perspective. Instead, we can inline the logic and just return `ByteBuffer`.
Same for `decode(...)`. Wdyt?

tomekl007 (commented):
Besides the `BigInteger` and `BigDecimal`, Java-driver supports the `Duration` type.
However, it is very CQL specific (see `com.datastax.oss.driver.api.core.data.CqlDuration`) so I am not sure if we want to handle it as well.

For the DSE, there are:
```
  public static final TypeCodec<LineString> LINE_STRING = new LineStringCodec();

  public static final TypeCodec<Point> POINT = new PointCodec();

  public static final TypeCodec<Polygon> POLYGON = new PolygonCodec();

  public static final TypeCodec<DateRange> DATE_RANGE = new DateRangeCodec();
```
I am also not sure if we want to support them.
@mpenick wdyt?

dougwettlaufer (approved):

grpc-proto/proto/query.proto
```diff
@@ -55,6 +55,15 @@ message Uuid {
   fixed64 lsb = 2;
 }
 
+message BigInteger {
```
mpenick, line 58: Could we name these `Varint` and `Decimal` to match the CQL protocol types?

grpc/src/main/java/io/stargate/grpc/codec/cql/VarintCodec.java
mpenick: Seems like overkill for how simple these codecs are.

grpc-proto/proto/query.proto
tomekl007: done

mpenick (approved):

PR #1166, opened 2021-07-30
Increased default size for document search to 3

EricBorczuk:
The default `page-size` for document search used to be 1, so if you search without explicitly providing the parameter, you get 1 document back and a page-state object if there are more. This PR increases that default so as to not confuse new users.


dimas-b (approved):
There are 3 INT test failures with C* 4.0, otherwise LGTM :+1:

PR #1171, opened 2021-08-09
document `$not` in Swagger

ivansenic:
Left over from #1158.


dimas-b (approved):
Thanks for catching this @ivansenic :)

PR #1172, opened 2021-08-09
gRPC: wait for schema agreement asynchronously (fixes #1145)

### LINKED ISSUE #1145: gRPC: Waiting for schema agreement should be asynchronous

olim7t:
The most tricky part is that we need an executor to schedule the retries. Ideally, I would have liked to expose the feature directly on `Persistence`, but not all implementations use an executor internally, and for those that do it doesn't allow scheduled tasks. So the next best thing is to use the executor of the gRPC service.

### DIFF
--- a/grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
+++ b/grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
@@ -39,9 +39,9 @@ import org.slf4j.LoggerFactory;
 
 public class GrpcImpl {
   private static final Logger logger = LoggerFactory.getLogger(GrpcImpl.class);
-  private static final Integer EXECUTOR_SIZE =
-      Integer.getInteger(
-          "stargate.grpc.executor_size", Runtime.getRuntime().availableProcessors() * 2);
+  private static final Integer EXECUTOR_SIZE = Integer.getInteger("stargate.grpc.executor_size", 8);
+  private static final Integer SHUTDOWN_TIMEOUT_SECONDS =
+      Integer.getInteger("stargate.grpc.shutdown_timeout_seconds", 60);
 
   private final Server server;
   private final ScheduledExecutorService executor;
@@ -87,12 +87,18 @@ public class GrpcImpl {
 
   public void stop() {
     try {
-      server.shutdown().awaitTermination();
+      server.shutdown();
       // Since we provided our own executor, it's our responsibility to shut it down.
       // Note that we don't handle restarts because GrpcActivator never reuses an existing instance
       // (and that wouldn't work anyway, because Server doesn't support it either).
       executor.shutdown();
-      if (!executor.awaitTermination(1, TimeUnit.MINUTES)) {
+
+      long timeoutMillis = TimeUnit.SECONDS.toMillis(SHUTDOWN_TIMEOUT_SECONDS);
+      long start = System.currentTimeMillis();
+
+      if (!server.awaitTermination(timeoutMillis, TimeUnit.MILLISECONDS)
+          || !executor.awaitTermination(
+              timeoutMillis - (System.currentTimeMillis() - start), TimeUnit.MILLISECONDS)) {
         logger.warn("Timed out while waiting for executor shutdown");
       }
     } catch (InterruptedException e) {
--- a/grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
+++ b/grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
@@ -16,6 +16,7 @@
 package io.stargate.grpc.impl;
 
 import io.grpc.Server;
+import io.grpc.internal.GrpcUtil;
 import io.grpc.netty.shaded.io.grpc.netty.NettyServerBuilder;
 import io.micrometer.core.instrument.binder.grpc.MetricCollectingServerInterceptor;
 import io.stargate.auth.AuthenticationService;
@@ -30,13 +31,20 @@ import java.io.UncheckedIOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class GrpcImpl {
   private static final Logger logger = LoggerFactory.getLogger(GrpcImpl.class);
+  private static final Integer EXECUTOR_SIZE =
+      Integer.getInteger(
+          "stargate.grpc.executor_size", Runtime.getRuntime().availableProcessors() * 2);
 
   private final Server server;
+  private final ScheduledExecutorService executor;
 
   public GrpcImpl(
       Persistence persistence, Metrics metrics, AuthenticationService authenticationService) {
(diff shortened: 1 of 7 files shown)


grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
```diff
@@ -53,11 +61,15 @@ public GrpcImpl(
 
     int port = Integer.getInteger("stargate.grpc.port", 8090);
 
+    executor =
+        Executors.newScheduledThreadPool(
+            EXECUTOR_SIZE, GrpcUtil.getThreadFactory("grpc-stargate-executor", true));
     server =
         NettyServerBuilder.forAddress(new InetSocketAddress(listenAddress, port))
+            .executor(executor)
```
olim7t, line 69: Beyond this ticket, injecting our own executor is a good thing: if we don't, it defaults to `Executors.newCachedThreadPool`, which is unbounded.

grpc/src/main/java/io/stargate/grpc/service/QueryHandler.java
```diff
@@ -192,4 +201,30 @@ private Parameters makeParameters(QueryParameters parameters, Optional<ClientInf
...
+
+  private void waitForSchemaAgreement(
+      int remainingAttempts, CompletableFuture<Void> agreementFuture) {
+    if (persistence.isInSchemaAgreement()) {
+      agreementFuture.complete(null);
+      return;
+    }
+    if (remainingAttempts <= 1) {
+      agreementFuture.completeExceptionally(
+          new IllegalStateException(
```
mpenick, line 219: I think this ends being an `UNKNOWN` error from the user's perspective without any information. It could be nicer to use a `StatusException`.  `Status.DEADLINE_EXCEEDED.withDescriptions("Failed to reach schema agreement after...").toException()` ?

mpenick:
I might add a unit test where there's never schema agreement. Should be fairly straightforward, look in in the package `io.stargate.grpc.service` in the `test` directory.

grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
mpenick: `Executors.newScheduledThreadPool()` is also unbounded unless we explicitly set the maximum pool size.  

`ScheduledThreadPoolExecutor` inherits `ThreadPoolExecutor` which is the same implementation returned by `Executors.newCachedThreadPool()`.

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ScheduledThreadPoolExecutor.html

olim7t: It actually overrides that behavior. From the class-level javadoc in that same page:

> because it acts as a fixed-sized pool using `corePoolSize` threads and an unbounded queue, adjustments to `maximumPoolSize` have no useful effect

grpc/src/main/java/io/stargate/grpc/service/QueryHandler.java
olim7t: Good suggestion, will do 👍

grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
mpenick: Thanks, I missed that.

grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
```diff
@@ -28,13 +29,20 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class GrpcImpl {
   private static final Logger logger = LoggerFactory.getLogger(GrpcImpl.class);
+  private static final Integer EXECUTOR_SIZE =
+      Integer.getInteger(
+          "stargate.grpc.executor_size", Runtime.getRuntime().availableProcessors() * 2);
```
tomekl007, line 42: will this `Runtime.getRuntime().availableProcessors()` result in a sane number in the shared k8s environment?

grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
```diff
@@ -72,6 +84,13 @@ public void start() {
   public void stop() {
     try {
       server.shutdown().awaitTermination();
+      // Since we provided our own executor, it's our responsibility to shut it down.
+      // Note that we don't handle restarts because GrpcActivator never reuses an existing instance
+      // (and that wouldn't work anyway, because Server doesn't support it either).
+      executor.shutdown();
+      if (!executor.awaitTermination(1, TimeUnit.MINUTES)) {
```
tomekl007, line 91: the `executor.awaitTermination(1, TimeUnit.MINUT)` is blocking. 
It seems that in the worst case, it may slow down the shutdown of Stargate by one minute. Isn't it too much?

grpc/src/main/java/io/stargate/grpc/service/QueryHandler.java
```diff
@@ -192,4 +201,32 @@ private Parameters makeParameters(QueryParameters parameters, Optional<ClientInf
...
+              .withDescription(
+                  "Failed to reach schema agreement after "
+                      + (200 * Persistence.SCHEMA_AGREEMENT_WAIT_RETRIES)
+                      + " milliseconds.")
+              .asException());
+      return;
+    }
+    executor.schedule(
+        () -> waitForSchemaAgreement(remainingAttempts - 1, agreementFuture),
+        200,
```
tomekl007, line 229: shouldn't we make the interval time configurable? Or at least extract it to a constant and name it for better visibility of it?

grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
dougwettlaufer: Looks like it works in everything after 8u131
https://blogs.oracle.com/java-platform-group/java-se-support-for-docker-cpu-and-memory-limits

dougwettlaufer: Wait, actually might not apply until JDK 10
https://dev.to/zac_siegel/java-and-docker---memory-and-cpu-limits-3h77

grpc/src/main/java/io/stargate/grpc/service/QueryHandler.java
olim7t: It's already configurable, `Persistence.SCHEMA_AGREEMENT_WAIT_RETRIES` can be overridden with a system property.

grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
olim7t: I can always make it customizable with yet another system property...

We'll also need to address the fact that the previous line calls `server.shutdown().awaitTermination();` which can block indefinitely.

grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
olim7t: Good catch. If we can't trust that method, then the only alternative is a hard-coded number.

mpenick (approved):
Looks good. :+1:

PR #1173, opened 2021-08-10
gRPC: deprecate ResultSet.page_size (fixes #1161)

### LINKED ISSUE #1161: gRPC: Remove `page_size` from `ResultSet` type
It's redundant and not necessary.

olim7t:


grpc-proto/proto/query.proto
```diff
@@ -488,8 +488,8 @@ message ResultSet {
   // The rows of CQL data.
   repeated Row rows = 2;
 
-  // The number of rows.
-  google.protobuf.Int32Value page_size = 3;
+  // Deprecated: result sets don't include this field anymore. It can be inferred from rows.size.
+  google.protobuf.Int32Value page_size = 3 [deprecated = true];
```
olim7t, line 492: I'm not sure how strict we want to be with backward compatibility at this stage.

Alternatively, I can remove the field entirely, and maybe even adjust the index of the next field...

mpenick: No backward compatibility. We should just remove it.

mpenick (approved):

dougwettlaufer (approved):

PR #1174, opened 2021-08-10
gRPC: use a default page size (fixes #1156)

### LINKED ISSUE #1156: gRPC: use a default page size
If the payload does not specify a page size, we don't set anything when calling the Persistence API. This results in a non-paged query.
This is dangerous, we should set a default at the service level.

Don't forget to update the docs in `query.proto`.

olim7t:


mpenick (approved):
:+1:

grpc/src/test/java/io/stargate/grpc/service/QueryParametersTest.java
```diff
@@ -81,37 +81,40 @@ public void queryParameters(QueryParameters actual, Parameters expected) {
 
   public static Stream<Arguments> queryParameterValues() {
     return Stream.of(
-        arguments(cqlQueryParameters().build(), Parameters.builder().build()),
+        arguments(cqlQueryParameters().build(), Parameters.builder().pageSize(100).build()),
```
tatu-at-datastax, line 84: Should this refer to the constant (Service.DEFAULT_PAGE_SIZE) in case that changes?

dougwettlaufer (approved):

PR #1175, opened 2021-08-10
gRPC: set default consistency levels in the service (fixes #1157)

### LINKED ISSUE #1157: gRPC: set a default serial consistency level
If the payload does not specify a serial consistency level, we don't set anything when calling the Persistence API. According to the javadocs:
> a default will be used but is unspecified and might depend of the Persistence implementation

We should set a default at the service level, to provide a more reliable contract to gRPC clients.

Don't forget to update the docs in `query.proto`.

olim7t:
I've used the same defaults as the GraphQL API.

### DIFF
--- a/grpc-proto/proto/query.proto
+++ b/grpc-proto/proto/query.proto
@@ -177,7 +177,7 @@ message QueryParameters {
   google.protobuf.StringValue keyspace = 1;
 
   // The consistency level.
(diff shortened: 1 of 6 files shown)


mpenick (approved):
:+1:

dougwettlaufer (approved):

PR #1176, opened 2021-08-11
gRPC: propagate HTTP headers to the persistence layer

olim7t:
Opening as a draft because I still need to test manually against serverless.

### DIFF
--- a/grpc/src/main/java/io/stargate/grpc/service/interceptors/HeadersInterceptor.java
+++ b/grpc/src/main/java/io/stargate/grpc/service/interceptors/HeadersInterceptor.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.grpc.service.interceptors;
+
+import io.grpc.Context;
+import io.grpc.Contexts;
+import io.grpc.Metadata;
+import io.grpc.Metadata.Key;
+import io.grpc.ServerCall;
+import io.grpc.ServerCallHandler;
+import io.grpc.ServerInterceptor;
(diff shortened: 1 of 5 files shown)


grpc/src/main/java/io/stargate/grpc/service/interceptors/HeadersInterceptor.java
```diff
@@ -0,0 +1,30 @@
...
+import io.grpc.Metadata;
+import io.grpc.Metadata.Key;
+import io.grpc.ServerCall;
+import io.grpc.ServerCallHandler;
+import io.grpc.ServerInterceptor;
+import io.stargate.grpc.service.Service;
+import java.util.HashMap;
+import java.util.Map;
+
+public class HeadersInterceptor implements ServerInterceptor {
```
olim7t, line 14: I went with a separate component, but this could also be merged with `AuthenticationInterceptor` since it does similar work.

grpc/src/main/java/io/stargate/grpc/service/interceptors/HeadersInterceptor.java
```diff
@@ -0,0 +1,30 @@
...
+  @Override
+  public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
+      ServerCall<ReqT, RespT> call, Metadata headers, ServerCallHandler<ReqT, RespT> next) {
+    Map<String, String> stringHeaders = new HashMap<>();
+    for (String key : headers.keys()) {
+      if (key.endsWith("-bin")) {
```
tomekl007, line 20: why do we need this edge case?

olim7t: gRPC allows both text and binary headers. Binary header keys always end with `-bin`.

`Connection.setCustomProperties` only accept text headers (and we know the tenant ID will come in a text header).

olim7t:
Finally validated this against an actual serverless environment. This is good to go, pending a possible merge of the interceptors.

grpc/src/main/java/io/stargate/grpc/service/interceptors/HeadersInterceptor.java
dougwettlaufer: > gRPC allows both text and binary headers
TIL

dougwettlaufer (approved):

mpenick (approved):
Maybe overkill, but is there any value in adding a unit test where we pass in headers and validate them with a `Connection` mock?

olim7t:
> Maybe overkill, but is there any value in adding a unit test where we pass in headers and validate them with a `Connection` mock?

Done, the existing test infrastructure made it pretty straightforward.

mpenick (approved):

PR #1179, opened 2021-08-12
Fix #1178 by removing unused/unnecessary dep to `json-simple` from poms

### LINKED ISSUE #1178: Remove `json-simple` library dependency (unused)
Looks like both `health-checker` and `restapi` depend on `json-simple` library (https://mvnrepository.com/artifact/com.googlecode.json-simple/json-simple) via their `pom.xml`s, but I don't see any actual usage. It is likely that there either was some code earlier (since removed or converted to use Jackson).

Assuming I did not miss anything it seems best to remove this unnecessary dependency even if the library itself is quite small (but also unmaintained, last release in 2012).
Note that this should not have an effect on possible transitive dependencies (something else requiring `json-simple`)

I will provide a PR.

tatu-at-datastax:
As per title, simple dependency pruning.

### DIFF
--- a/health-checker/pom.xml
+++ b/health-checker/pom.xml
@@ -54,11 +54,6 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
-      <groupId>com.googlecode.json-simple</groupId>
-      <artifactId>json-simple</artifactId>
-      <version>1.1</version>
-    </dependency>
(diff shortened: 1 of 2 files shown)


ivansenic (approved):
Awesome catch.. I love cleanups :heart:

PR #1180, opened 2021-08-12
Mock peers with internal addresses if the client connects over an internal network

olim7t:
Not sure if the title is right.

TODO:
- [ ] confirm that the proxy port doesn't change between public & internal


persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/ClientStateWithPublicAddress.java
```diff
@@ -25,9 +27,18 @@ public ClientStateWithPublicAddress(
...
 
   public InetSocketAddress publicAddress() {
     return publicAddress;
   }
+
+  /**
+   * Whether {@link #publicAddress()} is "internal", meaning that the client connected via a local
+   * network rather than the public internet.
+   */
+  public boolean isInternal() {
```
olim7t, line 41: There's a bit of a terminology clash here because the class already used "public address", but that address can be public or private now.

Maybe it would be better to rename the class to `ClientStateWithServerAddress`, and this method to `isPrivate()`.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptor.java
```diff
@@ -71,11 +71,13 @@
...
   private final Resolver resolver;
   private final String proxyDnsName;
+  private final String internalProxyDnsName;
   private final int proxyPort;
   private final long resolveDelaySecs;
   private final List<EventListener> listeners = new CopyOnWriteArrayList<>();
   private final Map<InetAddress, Set<String>> tokensCache = new ConcurrentHashMap<>();
-  private volatile Set<InetAddress> peers = Collections.emptySet();
+  private volatile Set<InetAddress> externalPeers = Collections.emptySet();
+  private volatile Set<InetAddress> internalPeers = Collections.emptySet();
```
olim7t, line 80: As far as I can tell, this interceptor is shared across connections (it's set on `ClientState.cqlQueryHandler`, which is static).

We can't predict if connections are going to be internal or public (or both), so we have to handle both cases.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptor.java
```diff
@@ -187,34 +190,40 @@ public void register(EventListener listener) {
...
-          }
-          peers = resolved;
+      for (EventListener listener : listeners) {
+        for (InetAddress peer : added) {
+          listener.onJoinCluster(peer, proxyPort);
+          listener.onUp(peer, proxyPort);
+        }
+        for (InetAddress peer : removed) {
+          tokensCache.remove(peer);
+          listener.onLeaveCluster(peer, proxyPort);
```
olim7t, line 222: We are emitting events for both sets of addresses. I don't think we can do better, we don't know in which context each listener operates.

I believe this was already happening before this change, clients that didn't connect with the PROXY header would still receive events for proxied peers.

In the end it shouldn't matter because drivers discard events if they don't recognize the address.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptor.java
```diff
@@ -288,7 +297,7 @@ private ByteBuffer buildColumnValue(String name, InetAddress publicAddress) {
    * @return a list of random token calculated using the the public address as a seed.
    */
   private Set<String> getTokens(InetAddress publicAddress) {
-    if (peers.contains(publicAddress)) {
+    if (externalPeers.contains(publicAddress)) {
```
olim7t, line 300: Oh I missed that, need to check what I have to change here.

prabhatjha:
My first comment on Stargate PR so hopefully my comment makes sense.

@olim7t  May be PR title should be "Return private ip address for peers if client connects over private/internal network`

Can you add test case/s?

olim7t:
> May be PR title should be "Return private ip address for peers if client connects over private/internal network`

👍 

> Can you add test case/s?

That is complicated, even as an integration test, because it requires a very specific environment: a proxy in front of Stargate that adds the header, and public IPs.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/ClientStateWithPublicAddress.java
```diff
@@ -12,11 +12,13 @@
 public class ClientStateWithPublicAddress extends ClientState {
 
   private final InetSocketAddress publicAddress;
+  private final boolean internal;
```
tatu-at-datastax, line 15: Probably implied by type `boolean`, but maybe should be named `isInternal`?
(and/or add brief javadoc comment on its purpose)

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptor.java
```diff
@@ -159,22 +162,23 @@ public void initialize() {
           "Unable to intercept proxy protocol system query without a valid public address");
     }
     String tableName = selectStatement.table();
+    Set<InetAddress> currentPeers = clientState.isInternal() ? internalPeers : externalPeers;
```
mpenick, line 165: Instead of having an "internal" flag could we check the peers for membership? This is more generalized, and we could potential support more than two sets (internal, external). I'm thinking about something like supporting private link + VPC peering + public, etc.

olim7t: That would work too, but is there any risk that the peer sets could be out of sync (with the refresh interval being 10 seconds). We'd have to error out if we don't find the address anywhere.

The current implementation is immune to that because:
* system.peers = knownPeers - publicAddress
* system.local = publicAddress (does not check that it's present in knownPeers)

olim7t:
> Maybe it would be better to rename the class to `ClientStateWithDestinationAddress`, and this method to `isPrivate()`.

I went ahead with that renaming. "Destination address" is the term that the PROXY protocol uses.

Another problem is that there's already a method `ClientInfo.isInternal()`, and I was accidentally overriding it.

persistence-api/src/main/java/io/stargate/db/ClientInfo.java
```diff
@@ -83,7 +86,7 @@ public String toString() {
         + "remoteAddress="
         + remoteAddress
         + ", publicAddress="
```
dougwettlaufer, line 88: Should this be updated as well?

ci/test.sh
```diff
@@ -10,8 +10,8 @@ chown -R ubuntu:ubuntu /workspace/
 chown -R ubuntu:ubuntu /cache/
 
 # Add DNS entries for proxy protocol tests
-echo '127.0.1.11 stargate.local' >> /etc/hosts
-echo '127.0.1.12 stargate.local' >> /etc/hosts
+echo '127.0.1.11 internal-stargate.local' >> /etc/hosts
```
tomekl007, line 13: shouldn't we retain the previous entries as well? Could you elaborate more on this?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptor.java
```diff
@@ -187,45 +190,50 @@ public void register(EventListener listener) {
...
+  private void resolveAllPeers() {
     if (!Strings.isNullOrEmpty(proxyDnsName)) {
       try {
-        Set<InetAddress> resolved = resolver.resolve(proxyDnsName);
+        publicPeers = resolvePeers(proxyDnsName, publicPeers);
+        privatePeers = resolvePeers(internalProxyDnsName, privatePeers);
+      } catch (UnknownHostException e) {
+        throw new ServerError("Unable to resolve DNS for proxy protocol peers table", e);
+      }
+      scheduler.schedule(this::resolveAllPeers, resolveDelaySecs, TimeUnit.SECONDS);
```
tomekl007, line 202: From the code, I see that a typical value for `resolveDelaySecs` is `10 seconds`, is that true? 
Is it good enough in the Stargate context?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptor.java
tomekl007: Could you elaborate more when `In the end it shouldn't matter because drivers discard events if they don't recognize the address.` can happen?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptor.java
```diff
@@ -117,6 +114,7 @@ private ProxyProtocolQueryInterceptor(
       QueryInterceptor wrapped) {
     this.resolver = resolver;
     this.proxyDnsName = proxyDnsName;
+    this.internalProxyDnsName = "internal-" + proxyDnsName;
```
dougwettlaufer, line 117: We should pull this from a system property as well rather than requiring a naming convention.

prabhatjha:
@olim7t We went from no unit test to what seems like pretty serious test. Nice job. 💪🏼

ci/test.sh
olim7t: Our CI environment uses local IP addresses, so `ProxyProtocolQueryInterceptor` will resolve the peers using the "internal" DNS name.

persistence-api/src/main/java/io/stargate/db/ClientInfo.java
olim7t: Yes, good catch.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptor.java
olim7t: Yes it's 10 seconds. I didn't pick that value, it's what was already there.

@mpenick any insight on how it was chosen?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptor.java
olim7t: This interceptor is a global component shared across all connections (see [my other comment](https://github.com/stargate/stargate/pull/1180#discussion_r688140706)).

Unlike `interceptQuery()`, `resolveAllPeers()` is NOT executed in the context of a client query, it's a periodic task that runs in the background. So it must emit events for both private and public clients.

Let's say that node A goes down, its public address is 1.1.1.1 and its private address is 127.28.0.1. We emit DOWN events for both addresses.
A client that is connected to a public address sees public addresses in `system.peers`, so it will process the event for 1.1.1.1, but won't know what to do with 127.28.0.1. For a client connected to a private address it's the opposite.

In the Java driver, an unknown address would be handled [here](https://github.com/datastax/java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java#L351-L355).

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptor.java
olim7t: I was taking the path of least resistance because a new property will have more impact in downstream projects.

But actually I can have the best of both worlds by using the convention if the new property is absent. Will do.

dougwettlaufer:
Looks like a flaky test

```
Step #4 - "cassandra-4.0": INFO  [process-runner-stargate-0-0] 2021-08-20 00:59:04,394 ProcessRunner.java:162 - Stargate process 0/0 failed with exception: {}
Step #4 - "cassandra-4.0": org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1)
Step #4 - "cassandra-4.0":      at org.apache.commons.exec.DefaultExecutor.executeInternal(DefaultExecutor.java:404)
Step #4 - "cassandra-4.0":      at org.apache.commons.exec.DefaultExecutor.access$200(DefaultExecutor.java:48)
Step #4 - "cassandra-4.0":      at org.apache.commons.exec.DefaultExecutor$1.run(DefaultExecutor.java:200)
Step #4 - "cassandra-4.0":      at java.lang.Thread.run(Thread.java:748)
Step #4 - "cassandra-4.0": [ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 13.623 s <<< FAILURE! - in io.stargate.it.cql.SnappyCompressionTest
Step #4 - "cassandra-4.0": [ERROR] io.stargate.it.cql.SnappyCompressionTest  Time elapsed: 13.622 s  <<< ERROR!
Step #4 - "cassandra-4.0": java.lang.IllegalStateException: java.util.concurrent.ExecutionException: org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1)
Step #4 - "cassandra-4.0": Caused by: java.util.concurrent.ExecutionException: org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1)
Step #4 - "cassandra-4.0": Caused by: org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1)
Step #4 - "cassandra-4.0":
```

dougwettlaufer (approved):

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptor.java
mpenick: I'm sorry I don't have as much insight as I should for the value of the delay. I believe that was the existing value.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/interceptors/ProxyProtocolQueryInterceptor.java
mpenick: I agree it's probably not a big deal. There is a way to fix this now because there's a concept of event filtering in the cql module. We could pass down the destination address and filter the events based on matching that to the `ProxyInfo.attributeKey` channel attribute, but maybe it's not worth it.

tatu-at-datastax: Is there a way to add summary of above as comments in sources, to explain the logic and assumptions (unknown addresses will be dropped quietly)? Just as context for whoever gets to modify the code next.

PR #1186, opened 2021-08-17
closes #1185: docsapi to have it's own module in the metrics

### LINKED ISSUE #1185: Docs API should have it's own module in the http metric
Currently both Docs API and Rest API are having the same `module="restapi"` as they are actually in the same, `restapi` bundle. We should overwrite the default one here and set it properly to distinguish.

ivansenic:
@dougwettlaufer and I agreed to overwrite the default module tag set by the bundle, so that we can clearly separate docsapi requests in the metrics.

### DIFF
--- a/testing/src/main/java/io/stargate/it/http/MetricsTest.java
+++ b/testing/src/main/java/io/stargate/it/http/MetricsTest.java
@@ -109,6 +109,54 @@ public class MetricsTest extends BaseOsgiIntegrationTest {
   }
 
   @Test
+  public void docsApiHttpRequestMetrics() throws IOException {
+    // call the rest api path with target header
+    String path = String.format("%s:8082/v2/namespaces/some", host);
+    OkHttpClient client = new OkHttpClient().newBuilder().build();
+    Request request =
+        new Request.Builder()
+            .url(path)
+            .get()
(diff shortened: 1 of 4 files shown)


dougwettlaufer (approved):

testing/src/main/java/io/stargate/it/http/MetricsTest.java
```diff
@@ -108,6 +108,54 @@ public void restApiHttpRequestMetrics() throws IOException {
...
+
+    await()
+        .atMost(Duration.ofSeconds(5))
+        .untilAsserted(
+            () -> {
+              String result =
+                  RestUtils.get("", String.format("%s:8084/metrics", host), HttpStatus.SC_OK);
+
+              List<String> lines =
+                  Arrays.stream(result.split(System.getProperty("line.separator")))
```
tatu-at-datastax, line 133: (nitpick) Probably not an actual problem but relying on local line separate default is usually not a good idea; and since line separators are either `\n`, `\r` or `\r\n` might make sense to rather just split with regexp of `[\\r\\n]`.

But more to the point would probably make sense to encapsulate something like "findHttpHeaderLines(String response, String headerType)" at some point to make tests more compact.

PR #1187, opened 2021-08-17
Adding a PR template to the repo

dougwettlaufer:


prabhatjha (approved):

tatu-at-datastax (approved):

.github/pull_request_template.md
```diff
@@ -0,0 +1,15 @@
...
+2. Rebase your PR if it gets out of sync with main
+-->
+
+**What this PR does**:
+
+**Which issue(s) this PR fixes**:
+Fixes #<issue number>
+
+**Checklist**
+- [ ] Changes manually tested
```
tomekl007, line 13: I am not sure what these points mean. I think that there are some low-level changes that can be fully covered in unit and integration tests. Why should we also require manual tests in that case? Some low-level changes might be tough to test manually.

prabhatjha: @tomekl007 Some tests may be complicated to test as unit test for example network connectivity related tests over VPC. We would mark those PRs as manually tested.  Developers usually run their test manually anyway before PR is created so in those case it's just a reminder to do that.

tomekl007: yes, but this list looks like ALL conditions must be met, not some of them

PR #1189, opened 2021-08-19
Catch JsonProcessingException and return 400 status

### LINKED ISSUE #1110: Catch JsonParseException and return 400
Noticed this error specifically with updateRows but there's other places it can happen. 

```java
ERROR [dw-484 - PUT /v2/keyspaces/ks1/tbl1/af2603d2-8c03-11eb-a03f-0ada685e000b] 2021-07-15 12:11:04,885 RequestHandler.java:96 - Error when executing request
com.fasterxml.jackson.core.JsonParseException: Unrecognized token 'foo': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
 at [Source: (String)"{
    "description": "bar",
    "field1": foo
}"; line: 3, column: 18]
	at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:1840)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:722)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._reportInvalidToken(ReaderBasedJsonParser.java:2873)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._reportInvalidToken(ReaderBasedJsonParser.java:2851)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._matchToken2(ReaderBasedJsonParser.java:2648)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._matchToken(ReaderBasedJsonParser.java:2626)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._matchFalse(ReaderBasedJsonParser.java:2600)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser.nextFieldName(ReaderBasedJsonParser.java:965)
	at com.fasterxml.jackson.databind.deser.std.MapDeserializer._readAndBindStringKeyMap(MapDeserializer.java:512)
	at com.fasterxml.jackson.databind.deser.std.MapDeserializer.deserialize(MapDeserializer.java:364)
	at com.fasterxml.jackson.databind.deser.std.MapDeserializer.deserialize(MapDeserializer.java:29)
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4218)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3214)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3182)
	at io.stargate.web.resources.v2.RowsResource.modifyRow(RowsResource.java:606)
	at io.stargate.web.resources.v2.RowsResource.lambda$updateRows$5(RowsResource.java:463)
	at io.stargate.web.resources.RequestHandler.handle(RequestHandler.java:37)
	at io.stargate.web.resources.v2.RowsResource.updateRows(RowsResource.java:461)
```

We need to either go through and wrap `mapper.readValue(payload, Map.class)` occurrences with a try-catch or catch it at the top in `io.stargate.web.resources.RequestHandler#handle`. The only problem with catching it at the top and returning the 400 is there might be a chance we'd incorrectly return 400 when it's actually a 500.

dougwettlaufer:
**What this PR does**:
Bringing the catch logic over from docsapi to restapi so that if JSON input is malformed its caught and an expected 400 is returned instead of a 500.

Fixes #1110 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/restapi/src/main/java/io/stargate/web/resources/ResourceUtils.java
+++ b/restapi/src/main/java/io/stargate/web/resources/ResourceUtils.java
@@ -3,15 +3,21 @@ package io.stargate.web.resources;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectReader;
 import java.util.HashMap;
 import java.util.Map;
 
 public class ResourceUtils {
 
-  public static Map<String, Object> readJson(String payload, ObjectMapper mapper) {
+  private static final ObjectMapper mapper = new ObjectMapper();
+  private static final ObjectReader MAP_READER =
+      mapper.readerFor(
+          mapper.getTypeFactory().constructType(new TypeReference<HashMap<String, Object>>() {}));
+
+  public static Map<String, Object> readJson(String payload) {
     Map<String, Object> requestBody;
     try {
-      requestBody = mapper.readValue(payload, new TypeReference<HashMap<String, Object>>() {});
+      requestBody = MAP_READER.readValue(payload);
     } catch (JsonProcessingException e) {
       throw new IllegalArgumentException(
           String.format("Input provided is not valid json. %s", e.getMessage()));
--- a/restapi/src/main/java/io/stargate/web/resources/ResourceUtils.java
+++ b/restapi/src/main/java/io/stargate/web/resources/ResourceUtils.java
@@ -1,0 +1,21 @@
+package io.stargate.web.resources;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ResourceUtils {
+
+  public static Map<String, Object> readJson(String payload, ObjectMapper mapper) {
+    Map<String, Object> requestBody;
+    try {
+      requestBody = mapper.readValue(payload, new TypeReference<HashMap<String, Object>>() {});
+    } catch (JsonProcessingException e) {
+      throw new IllegalArgumentException(
+          String.format("Input provided is not valid json. %s", e.getMessage()));
+    }
+    return requestBody;
+  }
+}
--- a/restapi/src/main/java/io/stargate/web/resources/v2/RowsResource.java
+++ b/restapi/src/main/java/io/stargate/web/resources/v2/RowsResource.java
@@ -19,7 +19,6 @@ import static io.stargate.web.docsapi.resources.RequestToHeadersMapper.getAllHea
 
 import com.codahale.metrics.annotation.Timed;
 import com.datastax.oss.driver.shaded.guava.common.base.Strings;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import io.stargate.auth.Scope;
 import io.stargate.auth.SourceAPI;
 import io.stargate.auth.TypedKeyValue;
@@ -88,7 +87,7 @@ import org.apache.cassandra.stargate.db.ConsistencyLevel;
 public class RowsResource {
 
   @Inject private Db db;
-  private static final ObjectMapper mapper = new ObjectMapper();
+
   private final int DEFAULT_PAGE_SIZE = 100;
 
   @Timed
@@ -381,7 +380,7 @@ public class RowsResource {
           AuthenticatedDB authenticatedDB =
               db.getRestDataStoreForToken(token, getAllHeaders(request));
 
-          Map<String, Object> requestBody = ResourceUtils.readJson(payload, mapper);
+          Map<String, Object> requestBody = ResourceUtils.readJson(payload);
 
           Table table = authenticatedDB.getTable(keyspaceName, tableName);
 
@@ -602,7 +601,7 @@ public class RowsResource {
           .build();
     }
 
-    Map<String, Object> requestBody = ResourceUtils.readJson(payload, mapper);
+    Map<String, Object> requestBody = ResourceUtils.readJson(payload);
     List<ValueModifier> changes =
         requestBody.entrySet().stream()
             .map(e -> Converters.colToValue(e.getKey(), e.getValue(), tableMetadata))
@@ -698,7 +697,7 @@ public class RowsResource {
     }
 
     List<ColumnOrder> order = new ArrayList<>();
-    Map<String, Object> sortOrder = ResourceUtils.readJson(sort, mapper);
+    Map<String, Object> sortOrder = ResourceUtils.readJson(sort);
 
     for (Map.Entry<String, Object> entry : sortOrder.entrySet()) {
       Column.Order colOrder =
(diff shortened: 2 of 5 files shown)


restapi/src/main/java/io/stargate/web/resources/v2/RowsResource.java
```diff
@@ -380,8 +382,14 @@ public Response createRow(
...
               db.getRestDataStoreForToken(token, getAllHeaders(request));
 
-          @SuppressWarnings("unchecked")
-          Map<String, Object> requestBody = mapper.readValue(payload, Map.class);
+          Map<String, Object> requestBody;
+          try {
+            requestBody =
+                mapper.readValue(payload, new TypeReference<HashMap<String, Object>>() {});
+          } catch (JsonProcessingException e) {
+            throw new IllegalArgumentException(
```
tatu-at-datastax, line 390: As per my comment on the issue, it might make sense to further separate out exception types; `JsonParseException` subtype would be for low-level decoding/parsing.
Probably does not make much difference here since there isn't much chance of data-binding (json tokens -> Java objects) problems.

restapi/src/main/java/io/stargate/web/resources/v2/RowsResource.java
```diff
@@ -380,8 +382,14 @@ public Response createRow(
...
+                String.format("Input provided is not valid json. %s", e.getMessage()));
```
tatu-at-datastax, line 391: Aside from question of what to catch, what to throw, would it make sense to include type of exception caught here like:

```
String.format("(%s) Input provider not valid json. %s", e.getClass().getName(), e.getMessage());
```

just in case more investigation was needed? Or I guess alternatively use `e.toString()` which adds type as well (but also includes location information which may or may not be desirable).

restapi/src/main/java/io/stargate/web/resources/v2/RowsResource.java
```diff
@@ -602,8 +610,13 @@ private Response modifyRow(
           .build();
     }
 
-    @SuppressWarnings("unchecked")
-    Map<String, Object> requestBody = mapper.readValue(payload, Map.class);
+    Map<String, Object> requestBody;
+    try {
```
tatu-at-datastax, line 614: ... since this is an idiom used in a few places, might want to just extract into separate method that handles error catch/rethrow?

restapi/src/main/java/io/stargate/web/resources/v2/RowsResource.java
```diff
@@ -693,14 +706,19 @@ private Object getRows(
...
+  private List<ColumnOrder> buildSortOrder(String sort) {
     if (Strings.isNullOrEmpty(sort)) {
       return new ArrayList<>();
     }
 
     List<ColumnOrder> order = new ArrayList<>();
-    Map<String, String> sortOrder = mapper.readValue(sort, Map.class);
+    Map<String, String> sortOrder;
+    try {
+      sortOrder = mapper.readValue(sort, new TypeReference<HashMap<String, String>>() {});
```
tatu-at-datastax, line 717: Is it important to bind into `Map<String, String>` instead of `Map<String, Object>`? Not a big deal but does coerce numbers into Strings and would fail if values JSON Arrays, Objects.

dougwettlaufer: These will be purely strings for example `firstName ASC` or `username DESC`. If there is a number provided then it's not a big deal to handle as a string.

restapi/src/main/java/io/stargate/web/resources/v2/RowsResource.java
dougwettlaufer: Sure.

restapi/src/main/java/io/stargate/web/resources/v2/RowsResource.java
dougwettlaufer: I'm not sure if the type of exception will be that valuable for API consumers. They'll be more concerned with the "invalid json" message.

tatu-at-datastax: Right, I was more thinking of us (datastax admins, devs) looking at logs/failures, not end users (except if and when they need to relay that information to us).
Not sure if there is a way to retain more information for our consumption.

But it is probably true that error message itself is enough to backtrack into type if we need that on case-by-case basis (just cannot search for specific type).

restapi/src/main/java/io/stargate/web/resources/v2/RowsResource.java
tatu-at-datastax: I guess I was more thinking of whether it could just use that same `Map<String, Object>` mapping as other methods, if a helper method was extracted. Sounds like it should only ever get String values, so in happy path no difference.

dougwettlaufer: I ended up switching to `Map<String, Object>` so I could reuse the catch block.

restapi/src/main/java/io/stargate/web/resources/ResourceUtils.java
```diff
@@ -0,0 +1,21 @@
...
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ResourceUtils {
+
+  public static Map<String, Object> readJson(String payload, ObjectMapper mapper) {
```
tatu-at-datastax, line 11: Could now also move `ObjectMapper` as a static member here? That would allow another thing, see next comment....

restapi/src/main/java/io/stargate/web/resources/ResourceUtils.java
```diff
@@ -0,0 +1,21 @@
...
+    Map<String, Object> requestBody;
+    try {
+      requestBody = mapper.readValue(payload, new TypeReference<HashMap<String, Object>>() {});
```
tatu-at-datastax, line 14: While construction of `TypeReference` is cheap, resolving that to `JavaType` in Jackson has some overhead. Nothing drastic, but if we have an `ObjectMapper`, we could streamline things by pre-constructing type as well as creating `ObjectReader` (which pre-fetches deserializer to use):

```
ObjectReader MAP_READER = mapper.readerFor(mapper.constructType(new TypeReference<...>() { }));

....

Map<String, Object> requestBody = MAP_READER.readValue(payload);
```

I doubt this has measurable effect in grand scheme of things but is also not much more complicated way.

restapi/src/main/java/io/stargate/web/resources/v2/RowsResource.java
```diff
@@ -693,18 +692,17 @@ private Object getRows(
...
-  private List<ColumnOrder> buildSortOrder(String sort)
-      throws com.fasterxml.jackson.core.JsonProcessingException {
+  private List<ColumnOrder> buildSortOrder(String sort) {
     if (Strings.isNullOrEmpty(sort)) {
       return new ArrayList<>();
     }
 
     List<ColumnOrder> order = new ArrayList<>();
-    Map<String, String> sortOrder = mapper.readValue(sort, Map.class);
+    Map<String, Object> sortOrder = ResourceUtils.readJson(sort, mapper);
```
tomekl007, line 701: where the `IllegalArgumentException` thrown from this method is handled?

dougwettlaufer: Here
https://github.com/stargate/stargate/blob/master/restapi/src/main/java/io/stargate/web/resources/RequestHandler.java#L48

restapi/src/main/java/io/stargate/web/resources/ResourceUtils.java
dougwettlaufer: Well that's cool. I'll make the change

tatu-at-datastax (approved):

PR #1190, opened 2021-08-20
Retries for gRPC

tomekl007:
**What this PR does**:
Provides retry capabilities for the gRPC API. This is based on the production-proven Java-Driver retry policy (https://github.com/datastax/java-driver/tree/ef56d561d97adcae48e0e6e8807f334aedc0d783/core/src/main/java/com/datastax/oss/driver/api/core/retry).

It provides a way to retry:
- `ReadTimeoutException`
-  `WriteTimeoutException`

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested

todo:
- [x] plug `is_idempotent` implemented in https://github.com/stargate/stargate/pull/1194

### DIFF
--- a/grpc/src/main/java/io/stargate/grpc/retries/DefaultRetryPolicy.java
+++ b/grpc/src/main/java/io/stargate/grpc/retries/DefaultRetryPolicy.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.grpc.retries;
+
+import com.google.common.annotations.VisibleForTesting;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import net.jcip.annotations.ThreadSafe;
+import org.apache.cassandra.stargate.db.WriteType;
+import org.apache.cassandra.stargate.exceptions.ReadTimeoutException;
+import org.apache.cassandra.stargate.exceptions.UnavailableException;
+import org.apache.cassandra.stargate.exceptions.WriteTimeoutException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The default retry policy.
+ *
+ * <p>This is a very conservative implementation: it triggers a maximum of one retry per request,
+ * and only in cases that have a high chance of success (see the method javadocs for detailed
+ * explanations of each case).
+ */
+@ThreadSafe
+public class DefaultRetryPolicy implements RetryPolicy {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultRetryPolicy.class);
+
+  @VisibleForTesting
+  public static final String RETRYING_ON_READ_TIMEOUT =
+      "Retrying on read timeout (consistency: {}, required responses: {}, "
+          + "received responses: {}, data retrieved: {}, retries: {})";
+
+  @VisibleForTesting
+  public static final String RETRYING_ON_WRITE_TIMEOUT =
+      "Retrying on write timeout (consistency: {}, write type: {}, "
+          + "required acknowledgments: {}, received acknowledgments: {}, retries: {})";
+
+  @VisibleForTesting
+  public static final String RETRYING_ON_UNAVAILABLE =
+      "Retrying on unavailable exception (consistency: {}, "
+          + "required replica: {}, alive replica: {}, retries: {})";
+
+  public DefaultRetryPolicy() {}
+
+  /**
+   * {@inheritDoc}
+   *
+   * <p>This implementation triggers a maximum of one retry (to the same node), and only if enough
+   * replicas had responded to the read request but data was not retrieved amongst those. That
+   * usually means that enough replicas are alive to satisfy the consistency, but the coordinator
+   * picked a dead one for data retrieval, not having detected that replica as dead yet. The
+   * reasoning is that by the time we get the timeout, the dead replica will likely have been
+   * detected as dead and the retry has a high chance of success.
+   *
+   * <p>Otherwise, the exception is rethrown.
+   */
+  @Override
+  public RetryDecision onReadTimeout(@NonNull ReadTimeoutException rte, int retryCount) {
+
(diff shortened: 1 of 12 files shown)


grpc/src/main/java/io/stargate/grpc/retries/DefaultRetryPolicy.java
```diff
@@ -0,0 +1,135 @@
...
+   * WriteType.BATCH_LOG} write. The reasoning is that the coordinator tries to write the
+   * distributed batch log against a small subset of nodes in the local datacenter; a timeout
+   * usually means that none of these nodes were alive but the coordinator hadn't detected them as
+   * dead yet. By the time we get the timeout, the dead nodes will likely have been detected as
+   * dead, and the retry has thus a high chance of success.
+   *
+   * <p>Otherwise, the exception is rethrown.
+   */
+  @Override
+  public RetryDecision onWriteTimeout(@NonNull WriteTimeoutException wte, int retryCount) {
```
tomekl007, line 98: This needs to be removed if we don't have an idempotency info

grpc/src/main/java/io/stargate/grpc/service/MessageHandler.java
```diff
@@ -104,29 +108,72 @@ protected MessageHandler(
...
+  }
+
+  protected RetryDecision shouldRetry(Throwable throwable, int retryCount) {
+    Optional<PersistenceException> cause = unwrapCause(throwable);
+    if (!cause.isPresent()) {
+      return RetryDecision.RETHROW;
+    }
+    PersistenceException pe = cause.get();
+    switch (pe.code()) {
+      case UNAVAILABLE:
```
tomekl007, line 148: we can safely retry only in those two cases

grpc/src/main/java/io/stargate/grpc/retries/DefaultRetryPolicy.java
```diff
@@ -0,0 +1,135 @@
...
+import org.apache.cassandra.stargate.exceptions.ReadTimeoutException;
+import org.apache.cassandra.stargate.exceptions.UnavailableException;
+import org.apache.cassandra.stargate.exceptions.WriteTimeoutException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The default retry policy.
+ *
+ * <p>This is a very conservative implementation: it triggers a maximum of one retry per request,
```
tomekl007, line 31: Ported from java-driver and adapted. We may decide to make the retry number configurable instead of hard-coded 1

mpenick:
I don't think it's useful to retry `UnavailableException`. An immediate retry under that exception is likely to fail too. The driver retries the failure by moving to the next coordinator in hopes that it has access to all replicas which is not possible from within a coordinator.

```

**Comment by tomekl007** [comment]
> I don't think it's useful to retry `UnavailableException`. An immediate retry under that exception is likely to fail too. The driver retries the failure by moving to the next coordinator in hopes that it has access to all replicas, but this is not possible from within a coordinator.
> 
> ```
> The default policy triggers a maximum of one retry, to the next node in the query plan. The rationale is that the first coordinator might have been network-isolated from all other nodes (thinking they’re down), but still able to communicate with the client; in that case, retrying on the same node has almost no chance of success, but moving to the next node might solve the issue.
> ```

Yes, I was also thinking if it is worth doing. I will remove the retry for `Unavailable`.

**Comment by tatu-at-datastax** [comment]
One small thing, not directly related to this patch: I noticed this module also has `Service` class. That's technically fine of course, as it's scoped by Java package. But it might be cleaner overall to rename that as `GrpcService`?
I renamed `Service` class of cql module, after it became sometimes confusing to have multiple Service classes open on IDE.

**Review by mpenick — APPROVED** [review]
Looks good.

File: grpc/src/main/java/io/stargate/grpc/service/MessageHandler.java [hunk 4]
```diff
@@ -104,30 +108,94 @@ protected MessageHandler(
...
+        || throwable instanceof StatusRuntimeException) {
+      return Optional.empty();
+    } else if (throwable instanceof PersistenceException) {
+      return Optional.of((PersistenceException) throwable);
+    } else {
+      return Optional.empty();
+    }
+  }
+
+  public CompletionStage<Response> executeQuery() {
```
olim7t, line 192: This can be private.
Also would you mind moving it just below `executeWithRetry`? I think that would be easier to follow since it's the first method that it invokes.

grpc/src/main/java/io/stargate/grpc/service/MessageHandler.java
```diff
@@ -104,30 +108,94 @@ protected MessageHandler(
...
+    executeQuery()
+        .whenComplete(
+            (response, error) -> {
+              if (error != null) {
+                RetryDecision decision = shouldRetry(error, retryCount);
+                if (decision == RetryDecision.RETRY) {
+                  executeWithRetry(retryCount + 1);
+                } else if (decision == RetryDecision.RETHROW) {
+                  handleException(error);
+                }
```
olim7t, line 134: This can be written as a switch statement.
And I think we throw on the default case, just in case another constant gets added in the future.

olim7t (approved):
Minor remarks inline.

Surfacing the flag via the exceptions is a bit convoluted (unwrap methods), but I understand why it needs to be done this way, and I don't have a simpler suggestion.

tomekl007:
> One small thing, not directly related to this patch: I noticed this module has generically named `Service` class. That's technically fine of course, as it's scoped by Java package. But it might be cleaner overall to rename that as `GrpcService`?
> I renamed `Service` class of cql module, after it became sometimes confusing to have multiple Service classes open on IDE.

good idea, we can do it in a separate PR.

grpc/src/main/java/io/stargate/grpc/service/MessageHandler.java
tomekl007: done

grpc/src/main/java/io/stargate/grpc/service/MessageHandler.java
tomekl007: done

grpc/src/main/java/io/stargate/grpc/retries/RetryPolicy.java
```diff
@@ -0,0 +1,64 @@
...
+   * Whether to retry when the server replied with a {@code READ_TIMEOUT} error; this indicates a
+   * <b>server-side</b> timeout during a read query, i.e. some replicas did not reply to the
+   * coordinator in time.
+   *
+   * <p>{@link ReadTimeoutException#consistency} the requested consistency level. {@link
+   * ReadTimeoutException#blockFor} the minimum number of replica acknowledgements/responses that
+   * were required to fulfill the operation. {@link ReadTimeoutException#received} the number of
+   * replica that had acknowledged/responded to the operation before it failed. {@link
+   * ReadTimeoutException#dataPresent} whether the actual data was amongst the received replica
+   * responses.
```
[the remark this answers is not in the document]
tomekl007: good idea, done

PR #1194, opened 2021-08-23
Infer and propagate information about indepotency of gRPC queries

### LINKED ISSUE #1196: Infer and propagate information about indepotency of gRPC queries
To allow client-side retries for the gRPC client, we need to provide the caller with the info on whether a submitted query is idempotent or not.

The better solution in the context of UX would be to infer the idempotency from the query and return that info in the gRPC Response (this solution is more complex). The simpler solution is to require the caller to specify whether the query is idempotent or not. However, this is non-ideal because the rules for queries idempotency are complex:
https://docs.datastax.com/en/devapp/doc/devapp/driversQueryIdempotence.html#Non-idempotentexamples

Java-driver follows this approach:

```
By default, all DataStax Drivers consider queries to be non-idempotent. It is the user’s responsibility to mark queries as idempotent to leverage features such as retry and speculative execution
```

Summary of non-idempotent queries:

- Mutations with updates to collections/counters

- Mutations that use of `now()` and `uuid()` (or UDFs)

- LWTs

Please note that things like `writetime` and `ttl` are excluded from the definition of idempotency:

```
“Result of the application” means the outcome on the state of the database. Some functions like ttl() return a different thing to the user on every call, but they don’t mutate anything.
```

tomekl007:
**What this PR does**:
Returns the `bool isIdempotent` in the gRPC response to allow client-side retries.
It analyzes the CQL Statement, and return that query is non-idempotent in the following scenarios:
- List appends, prepends or element deletions.
- Counter increments / decrements.
- Mutations that use non-idempotent functions such as now() and uuid()
- LWTs
- Truncate, schema changes, and USE

also:
- Batches are idempotent if their underlying statements are all
idempotent.
- All reads are idempotent.

reference work:
https://github.com/riptano/apollo/pull/882

Fixes #1196

**Checklist**
- [x] Changes manually tested
- [x] Documentation added/updated: https://github.com/stargate/stargate/pull/1194/files?short_path=8111a16#diff-8111a16121446391e0059fcaf7ce26986ae941bca669b88ba40b1a8d220cf6fc

**todo:**
- [x] infer idempotency in for the DSE
- [x] infer idempotency in for the C* 3
- [x] infer idempotency in for the C* 4
- [x] handle append/prepend for collection: https://github.com/stargate/stargate/pull/1194/files#diff-a9aee57bc231ca64451f45d22bca51d10af55788b6f3fe4edf48d6cf346f566eR96
- [x] decide/handle `BatchStatement`
- [x] decide if we should treat: Truncate, schema changes, and USE as non-idempotent.
- [x] get the info if about idempotency during `decode()`: https://github.com/stargate/stargate/pull/1194/files#diff-85e0db2983367de622d1f158ed4ee09146dc9914bfac676de085440491528f00R378
- [x] plug the new `PreparedWithIdempotent` into gRPC `Response`
- [x] fix the DSE build: https://github.com/stargate/stargate/pull/1194/checks?check_run_id=3442885398
- [x] do not return `is_idempotent` to a caller, for more info see: https://github.com/stargate/stargate/pull/1194#discussion_r698605047


grpc/src/main/java/io/stargate/grpc/idempotency/IdempotencyAnalyzer.java
```diff
@@ -0,0 +1,61 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.grpc.idempotency;
+
+import io.stargate.proto.QueryOuterClass;
+import java.util.Arrays;
+import java.util.List;
+
+public class IdempotencyAnalyzer {
```
tomekl007, line 22: We may consider moving this to the `persistence-api` and propagating the `idempotency` info in the:
`public static class Prepared extends Result`
The `isIdempotent` analysis logic seems not gRPC specific, so other modules that are preparing using cql String may benefit from that. wdyt?

grpc/src/main/java/io/stargate/grpc/idempotency/IdempotencyAnalyzer.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  public static boolean isIdempotent(String query) {
+    // analyze only mutations
+    if (query.contains("SELECT")) {
+      return true;
+    }
+
+    // check if it is updating a Collection or Counter
+    if (query.contains("SET") && query.contains("+") && query.contains("WHERE")) {
```
tomekl007, line 43: I've tried to catch the 3 cases of non-idempotent queries without the need for parsing the whole CQL query.
I am not entirely sure if this is exclusive with all other possible idempotent queries.

grpc-proto/proto/query.proto
```diff
@@ -359,6 +359,9 @@ message Response {
   repeated string warnings = 2;
   // The tracing information, if it was requested for the query.
   Traces traces = 3;
+
+  // Denotes if the query is idempotent or not. If it is, it is safe to retry it client-side
+  bool isIdempotent = 5;
```
mpenick, line 364: nit: `is_idempotent` (to match the preferred casing of .proto files)

grpc/src/main/java/io/stargate/grpc/idempotency/IdempotencyAnalyzer.java
tomekl007: After some more analysis, I found that it may be hard to move the `isIdempotent` to the `Result.Prepared` class. 
The `toResultInternal` method is constructing the `Preapred` in the 
```
public static Result toResultInternal(
      ResultMessage resultMessage, org.apache.cassandra.transport.ProtocolVersion version)
``` 
method. I don't see a way to extract the query from the `resultMessage` to analyze whether it is idempotent or not.
We could propagate the `String query` up from:
1. `public CompletableFuture<Result.Prepared> prepare(String query, Parameters parameters)`
2. `private <T extends Result> CompletableFuture<T> executeRequestOnExecutor(
            Parameters parameters, long queryStartNanoTime, Supplier<Request> requestSupplier,
            @Nullable String query)`
3. ```
 Conversion.toResult(
                        (ResultMessage) response,
                        Conversion.toInternal(parameters.protocolVersion()),
                            query);```
4. ```
public static Result toResultInternal(
          ResultMessage resultMessage, org.apache.cassandra.transport.ProtocolVersion version,
          String query)
```
5. and finally:
```
return new Result.Prepared(
            Conversion.toExternal(prepared.statementId),
            null,
            toResultMetadata(prepared.resultMetadata, null),
            toPreparedMetadata(prepared.metadata.names, preparedStatement.partitionKeyBindIndexes),
                IdempotencyAnalyzer.isIdempotent(query)
                );
```
but it seems that it complicates the code of persistence too much, outweighing benefits.

The other solution would be to propagate the `Request` and it to `PrependMessage`:
```
PrepareMessage prepareMessage = (PrepareMessage) request;
```
as the `PrepareMessage` has a `query` object underneath. Unfortunately, it does not expose it in any way, so we would need to use the reflection to extract the `query` object. It  I am not sure if this use case justifies it.

File: grpc/src/main/java/io/stargate/grpc/idempotency/IdempotencyAnalyzer.java [hunk 4]
```diff
@@ -0,0 +1,82 @@
...
+      boolean idempotent = isIdempotent(query.getCql());
+      // if ANY query within a batch is non-idempotent, then the whole batch is non-idempotent
+      if (!idempotent) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  public static boolean isIdempotent(String query) {
```
tomekl007, line 39: As you can see, the implementation is convoluted and complex. I am not sure if we will be able to catch all cases. At this point, I think that we should consider the same option as java-driver: that is, the end-user providing the `is_idempotent` when issuing a query with a default = `false` same as the driver:
```
By default, all DataStax Drivers consider queries to be non-idempotent. It is the user's responsibility to mark queries as idempotent to leverage features such as retry and speculative execution.
```
We may override/set it to `true` for all `SELECT` queries.
@olim7t, what was the main reason why the java-driver didn't decide to infer it from a query? (I think that I know the answer - the complexity but I wanted to double-check)

olim7t: > what was the main reason why the java-driver didn't decide to infer it from a query? (I think that I know the answer - the complexity but I wanted to double-check)

Yes, the complexity, and unfortunately we're running into the same kind of issue here. I'm afraid simple text searches are not going to be robust enough. Here are two queries that are not idempotent, yet they enter the first `if` block below:
```
INSERT INTO tbl(id, v) VALUES (uuid(), 'SELECT');
INSERT INTO a_table_named_select (id) VALUES (uuid());
```

To reliably analyze a CQL query string we need to parse it, but doing that in high-level modules like gRPC introduces two issues:
* duplication of the parsing logic that already exists in the Cassandra code
* the CQL grammar can change depending on the backend

I agree that we should do this further down the stack, but even our persistence modules are not deep enough: `Connection.prepare` is still dealing with a raw string, after that it's handling it to the regular (i.e. not overridden in Stargate) server code.

So we'd need to patch Cassandra itself, there is already a ticket for that: [CASSANDRA-10813](https://issues.apache.org/jira/browse/CASSANDRA-10813).

olim7t: PS- maybe we can do it in our persistence modules. That would take care of the second issue (CQL grammar specific to backend).
The double parsing would still happen, but maybe it's not that much of an overhead.

tomekl007: Taking all of that into consideration, I think that at this stage of gRPC API evolution we should go with the simplest approach and let the caller provide this info (same as java-driver, here is the PR:
https://github.com/stargate/stargate/pull/1197)
please let me know wdyt

tatu-at-datastax: Makes sense to me, given all the complexities. Worth revisiting if and when lower layer(s) can provide necessary information, and not have to work on string/regexp heuristics that may or may not work.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/ResultMessage.java
```diff
@@ -375,7 +376,9 @@ public Result decode(ByteBuf body, ProtocolVersion version) {
       if (version.isGreaterThan(ProtocolVersion.V1))
         resultMetadata = RowsSubCodec.METADATA_CODEC.decode(body, version);
 
-      return new Result.Prepared(id, resultMetadataId, resultMetadata, metadata);
+      boolean isIdempotent = byteToBoolean(CBUtil.readBytes(body));
```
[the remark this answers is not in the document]
mpenick: Why does this add the idempotent flag to the native protocol?

grpc-proto/proto/query.proto
```diff
@@ -359,6 +359,9 @@ message Response {
   repeated string warnings = 2;
   // The tracing information, if it was requested for the query.
   Traces traces = 3;
+
+  // Denotes if the query is idempotent or not. If it is, it is safe to retry it client-side
+  bool is_idempotent = 5;
```
mpenick, line 364: I'm not sure how valuable adding an `is_idempotent` flag to the response is. A `Response` is the result of a successful execution and wouldn't require a retry. It more of value to add the the error response as a trailer.

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/ResultMessage.java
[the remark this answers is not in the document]
tomekl007: To propagate info about idempotency, we need to have it in the `Result.Prepared`.
Unfortunately, it means that a new instance of `Result.Prepared` needs to be propagated here: 
https://github.com/stargate/stargate/pull/1194/files#diff-85e0db2983367de622d1f158ed4ee09146dc9914bfac676de085440491528f00R381
So the info whether it `isIdempotent` or not needs to be extracted somehow.
Is it something that we should not do?
I don't see an alternative way of getting that info.
We can do:
```
MD5Digest id = MD5Digest.wrap(CBUtil.readBytes(body));
QueryHandler.Prepared prepared = QueryProcessor.instance.getPrepared(org.apache.cassandra.utils.MD5Digest.compute(id.bytes));
``` 
but it returns `QueryHandler.Prepared` NOT `ResultMessage.Prepared` (that we could cast to `PreparedWithIdempotent`).

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/idempotency/IdempotencyAnalyzer.java
```diff
@@ -0,0 +1,122 @@
...
+          return false;
+        }
+      }
+    }
+
+    // delete
+    if (statement instanceof DeleteStatement) {
+      for (ColumnDefinition c : ((DeleteStatement) statement).updatedColumns()) {
+        // on a List
+        if (c.type instanceof ListType) {
```
mpenick, line 115: Should this be `CollectionType` to handle all collections (including map and set)?

tomekl007 (commented):
GH problem

grpc-proto/proto/query.proto
tomekl007: good point, I will propagate this info in the:
```
private <T> Metadata makeTrailer(Metadata.Key<T> key, T value) {
    Metadata trailer = new Metadata();
    trailer.put(key, value);
    return trailer;
  }
```
I wonder if there is a use-case for having that in the `Response` as well, or should we remove it.

persistence-cassandra-3.11/src/main/java/io/stargate/db/cassandra/impl/idempotency/IdempotencyAnalyzer.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.cassandra.cql3.statements.TruncateStatement;
+import org.apache.cassandra.cql3.statements.UseStatement;
+import org.apache.cassandra.db.marshal.CounterColumnType;
+import org.apache.cassandra.db.marshal.ListType;
+
+public class IdempotencyAnalyzer {
+  private static final Set<Function> NON_IDEMPOTENT_FUNCTION;
+
+  static {
+    Set<Function> allFunctions = new HashSet<>();
```
mpenick, line 41: nit: Maybe make use `ImmutableSet` or wrap `Collections.unmodifiableSet()`?

tomekl007: good point. done.

grpc-proto/proto/query.proto
tomekl007: We decided that it is not safe to return that info to the client in any form.
The main problem with it is that the failure may happen before/during preparing logic - in that case, we will not have information about idempotency. The client-side retries will not be able to see the flag, as it will be not set. It will result in a convoluted logic, where it is safe to retry only if the `is_idempotent` is present and is `true`.
We will use is_idempotent only for the internal server-side retires.  (cc @mpenick)

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/ResultMessage.java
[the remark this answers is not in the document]
tomekl007: We decided that it is safe to return hard-coded `false` here because it is not used. 
done in: https://github.com/stargate/stargate/pull/1194/commits/8eddd4aadc9386ae9a244cb9cc2770bd76fde73b

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateQueryHandler.java
```diff
@@ -125,7 +127,15 @@ public void register(QueryInterceptor interceptor) {
   @Override
   public Single<ResultMessage.Prepared> prepare(
       String query, QueryState queryState, Map<String, ByteBuffer> customPayload) {
-    return QueryProcessor.instance.prepare(query, queryState, customPayload);
+    return QueryProcessor.instance
+        .prepare(query, queryState, customPayload)
+        .map(
```
dougwettlaufer, line 132: Why is this different from the other two?

docs/idempotency/README.md
```diff
@@ -0,0 +1,28 @@
...
+- [IdempotencyAnalyzer C* 3.11]
+- [IdempotencyAnalyzer C* 4.0]
+- [IdempotencyAnalyzer DSE]
+
+Its responsibility is to analyze `CqlStatement`s and infer whether the query is idempotent or not.
+All prepared queries are analyzed.
+If yes, it is safe to retry such a query on the client or/and server-side. 
+If it is not, the query should not be retried because it may result in an inconsistent state of a Database.
+
+## Idempotency rules
```
dougwettlaufer, line 13: Shouldn't this be javadoc?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateQueryHandler.java
tomekl007: because it return `Single<ResultMessage.Prepared>` and not `ResultMessage.Prepared`

docs/idempotency/README.md
tomekl007: why? I think that we should have more docs in `readme.md` form that contains consolidated info about this feature. I can also add Javadoc if you think it's needed.

grpc-proto/proto/query.proto
tomekl007: done in: https://github.com/stargate/stargate/pull/1194/commits/839d27cbf678d747146b5957577667e8fe171d4b

docs/idempotency/README.md
dougwettlaufer: Because it's easier to read documentation next to the code you're looking at rather than jumping between code and external docs. Plus the closer docs are to code the more likely it is to stay up to date. Not specific to your PR but in general we're lacking on javadoc which makes it difficult for those not immersed in the project or specific feature to understand what's going on.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateQueryHandler.java
dougwettlaufer: Oh, right.

dougwettlaufer (approved):
Code looks good but I would suggest moving some of the documentation closer to the actual code.

mpenick (approved):

tomekl007:
> Code looks good but I would suggest moving some of the documentation closer to the actual code.

But in that case, we will also have documentation duplicated in 3 places (each `IdempotencyAnalyzer`).
In the current solution when it is hoisted up in the project structure, it does not need to be duplicated. @dougwettlaufer wdyt?

tatu-at-datastax:
> > Code looks good but I would suggest moving some of the documentation closer to the actual code.
> 
> But in that case, we will also have documentation duplicated in 3 places (each `IdempotencyAnalyzer`).
> In the current solution when it is hoisted up in the project structure, it does not need to be duplicated. @dougwettlaufer wdyt?

I think that while DRY is often useful, there are cases where some duplication makes sense, to support different use cases.
When following code paths Javadocs are much more relevant than READMEs, but I am sure there are opposite cases as well. It may make sense to keep one description brief, refer to the other, for more complicate explanations? (probably easier from README -> javadoc direction, at least for class javadoc).

tomekl007:
> > > Code looks good but I would suggest moving some of the documentation closer to the actual code.
> > 
> > 
> > But in that case, we will also have documentation duplicated in 3 places (each `IdempotencyAnalyzer`).
> > In the current solution when it is hoisted up in the project structure, it does not need to be duplicated. @dougwettlaufer wdyt?
> 
> I think that while DRY is often useful, there are cases where some duplication makes sense, to support different use cases.
> When following code paths Javadocs are much more relevant than READMEs, but I am sure there are opposite cases as well. It may make sense to keep one description brief, refer to the other, for more complicate explanations? (probably easier from README -> javadoc direction, at least for class javadoc).

Ok, I moved the docs closer to code: https://github.com/stargate/stargate/pull/1194/commits/9bc4aa0312e6c1f996344655dd8985ee296764ed

PR #1199, opened 2021-08-25
Exclude azure-storage-blob dependency from DSE dependencies

dougwettlaufer:
The current version of DSE that we're using is 6.8.13 and it depends on
azure-storage-blob 12.4.0 which has been removed from maven central. Per the GH
issue (https://github.com/Azure/azure-sdk-for-java/issues/23562) it looks like
they will not be adding it back which means until DSE is updated to the latest
version we won't be able to build. So instead of waiting for a newer DSE dependency
we'll exclude azure-storage-blob in order to fix our builds since Stargate is not
making use of any of the DSE functionality that requires talking to the Azure Blob Store.

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/persistence-dse-6.8/pom.xml
+++ b/persistence-dse-6.8/pom.xml
@@ -79,6 +79,10 @@
           <groupId>dnsjava</groupId>
           <artifactId>dnsjava</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>com.azure</groupId>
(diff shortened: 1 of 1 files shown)


tatu-at-datastax (approved):
Would it make sense to add an xml comment next to first exclusion to indicate the reason (link to root cause)?

PR #1202, opened 2021-08-26
GraphQL: Add developer docs

olim7t:
**What this PR does**:
Add docs for new developers learning about the GraphQL module.

N/A

**Checklist**


graphqlapi/developer_docs.md
```diff
@@ -0,0 +1,299 @@
...
+name indicates, it's implemented with
+[Dropwizard](https://www.dropwizard.io/en/latest/).
+
+## GraphQL Java primer
+
+We rely extensively on [GraphQL Java](https://www.graphql-java.com/). Before delving further into
+the Stargate code, it can be helpful to have a basic understanding of that library:
+
+```java
+Random random = new Random();
```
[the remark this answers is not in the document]
olim7t: It's effectively final, the compiler is able to infer that it's never reassigned.

tomekl007 (approved):

ashryanbeats:
> Should this be combined with the existing README in the directory?

Have we asked Mark to weigh in as well? He is thinking through contributor experience at the moment.

olim7t:
The existing README is more of a user documentation. TBH I'd rather remove it because it's a duplicate of what's available on [stargate.io](https://stargate.io/docs/stargate/1.0/developers-guide/graphql.html), maintaining the same information in two different places is inefficient. If we go that route, then the developer doc could replace it.

prabhatjha:
IMO, this kind of document should remain in their respective folder. Let's have these done for each of the remaining API and then we can update project level README with the links.

dougwettlaufer:
> IMO, this kind of document should remain in their respective folder. Let's have these done for each of the remaining API and then we can update project level README with the links.

Right, graphql already has a module level readme in addition to this doc. I agree with @olim7t that we should delete the current readme. Then we can rename the file in this PR to README.md.

prabhatjha:
Let's do that.

graphqlapi/developer_docs.md
```diff
@@ -0,0 +1,299 @@
+# Stargate GraphQL -- developer docs
+
+This module builds an OSGi bundle that exposes Stargate's GraphQL services over HTTP.
+
+Here is a brief functional overview of those services (for more details, refer to the [Stargate
```
dougwettlaufer, line 5: Neat, I had no idea markdown supported links like that.

graphqlapi/developer_docs.md
```diff
@@ -0,0 +1,299 @@
...
+```
+
+[StargateGraphqlContext] (via its inner class `BatchContext`) provides coordination services that
+allows independent fetchers to accumulate queries, and track which is the last one that must execute
+the batch.
+
+In the CQL-first code, [MutationFetcher] and [BulkMutationFetcher] are equipped to detect the
+directive and use the batch context. In the GraphQL-first code, there is a single
+[MutationFetcher][graphqlfirst.MutationFetcher] parent class that handles both regular and bulk
+operations. The GraphQL-first design is cleaner IMHO, so it would probably be a good idea to
```
dougwettlaufer, line 250: Should we create an issue and just remove this sentence?

dougwettlaufer (approved):

graphqlapi/developer_docs.md
olim7t: #1207

mpenick (approved):

graphqlapi/developer_docs.md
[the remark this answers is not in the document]
tatu-at-datastax: Ok thanks. I remember having had to add final in other cases so just wanted to double-check.

PR #1206, opened 2021-08-30
GraphQL CQL-first: use code registry to inject data fetchers

### LINKED ISSUE #1201: Refactor CQL-first schema builders to avoid deprecated method
`DdlSchemaBuilder` and `DmlSchemaBuilder` use `GraphQLFieldDefinition.Builder.dataFetcher()` to specify each data fetchers. That method has been deprecated in favor of `GraphQLSchema.Builder.codeRegistry()`, it would be cleaner to adapt our code to avoid the deprecation warnings.

For example, see how it's done in `AdminSchemaBuilder.build()`.

olim7t:
**What this PR does**:

Use a `GraphQLCodeRegistry` to inject the data fetchers in the schemas, rather than the deprecated `GraphQLFieldDefinition.Builder#dataFetcher`.

I also used the opportunity to refactor the DDL schema to only use static fields. This is a cleaner approach IMHO.
(We can't do it for DML because the schema is dynamic.)

Fixes #1201

**Checklist**
- [x] Changes manually tested
- Documentation: N/A


olim7t:
Moving this to draft because it's low priority, and given that we have 3 versions of the GraphQL API in flight at the moment, it's more of a distraction.

This can be revisited once things have stabilized and only the Quarkus version of GraphQL is left.

jeffreyscarpenter:
Is there something here that we should be trying to port to v2 @olim7t?

PR #1211, opened 2021-09-02
Fix for #1210: fix bundle name of `cql` module, Server->CqlServer, javadoc fixes

### LINKED ISSUE #1210: Small naming issues, use of deprecated options in CQL module
Looking at `cql` module, I am finding some small issues which are not bugs per se, but sub-optimal or misleading naming. There is also some use of deprecated functionality from 3rd party libraries like Netty.
For example:

* Bundle name is set as `RestAPI` (copy+paste?)
* Some types use very generic naming (like `Server`): seems some cases should simple `Cql` prefix (`CqlServer`) would be helpful when looking at things in IDE
    * note: earlier I thought `Config` was used but that's from Cassandra, cannot rename
* Some deprecated Netty constructs:
    * Slf4JLoggerFactory deprecated (not sure if we can use "instance()" method as suggested)
    * `Server` has ChannelOption.WRITE_BUFFER_HIGH_WATER_MARK and WRITE_BUFFER_LOW_WATER_MARK (deprecated names, use replacements)

I can provide a PR for basic cleanup here.

tatu-at-datastax:
**What this PR does**:

* Fixes duplicate bundle name: cql module "reused" name "RestAPI"
* Change generic cql module `Server` into `CqlServer` (code readability)
* Smaller fixes to javadoc wording ("be be" -> "be")

Fixes #1210

**Checklist**
- [ ] Changes manually tested -- no functional changes, will rely on CI
- [ ] Automated Tests added/updated -- no functional changes, will rely on CI
- [x] Documentation added/updated (although would be good to have README for cql)


cql/src/main/java/org/apache/cassandra/stargate/transport/internal/CqlServer.java
```diff
@@ -21,13 +21,7 @@
...
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
-import io.netty.channel.Channel;
-import io.netty.channel.ChannelFuture;
-import io.netty.channel.ChannelHandlerContext;
-import io.netty.channel.ChannelInitializer;
-import io.netty.channel.ChannelOption;
-import io.netty.channel.ChannelPipeline;
-import io.netty.channel.EventLoopGroup;
+import io.netty.channel.*;
```
ivansenic, line 24: Seems that IntelliJ is collapsing the imports.. We agreed at some point to not have the collapsed imports.. Can you increase the number of imports to use `*` to something very high (it's in Editor > Code Style > Java, then Imports tab)..

ivansenic (approved):
Seems good :)

cql/src/main/java/org/apache/cassandra/stargate/transport/internal/CqlServer.java
tatu-at-datastax: I was wondering about that; I think many IDEs default to collapse at 5 or so. I'll change my settings.

tatu-at-datastax: Reverted import change as suggested.

tatu-at-datastax:
Thank you @ivansenic! Will wait for the new CI run after imports now.

dougwettlaufer (approved):

Issue #1184, opened 2021-08-16
Add support for merging AND queries


ivansenic:
With support for the merging OR queries, we have an opportunity to implement merging AND queries as well. This can have a huge performance impact.

Right now the query like `{"price":{"$gt":0}, "key":{"$lt":100}}` will execute one filter op first and then for each candidate given, make sure that the second filter op is fulfilled. However, for the second filter we execute a query per candidate. Meaning that we end up in 1 + N Cassandra queries, where N is number of candidates needed to be checked. This is especially problematic when first filter matches a lot of documents, and second one a small number.

We could execute both queries as there are no candidates and then join the results. We need to evaluate if there is anything in the RxJava that can help here. Alternatively we could:
 1. Create our own operator
 2. Use existing support for the merging OR and then do the in memory evaluation

Note that this is only possible for filter operations that have a Cassandra predicate.

Current workaround for users can be to prioritize filter ops that match smaller amount of elements by having a `$selectivity` parameter.

Before starting this, we need perf test scenarios completed, so that we can evaluate the improvement in different situations.

ivansenic:
Easily implemented with https://smallrye.io/smallrye-mutiny/1.6.0/guides/combining-items/#combining-the-latest-items-of-multis

ivansenic:
Possible algorithm:

* each emitted item get enriched with the query index
* do an ordered merge based on the value and index
* then scan and emit items so that the ones having `ignore` false are existing in all streams
* filter out ignored elements

```
    public static void main(String[] args)
    {
        int totalFlows = 4;
        Flowable<IndexNumber> f1 = Multi.createFrom().items(0, 1, 2, 4, 7, 9).map(i -> new IndexNumber(i, 0)).convert().with(MultiRx3Converters.toFlowable());
        Flowable<IndexNumber> f2 = Multi.createFrom().items(3, 4, 5, 7).map(i -> new IndexNumber(i, 1)).convert().with(MultiRx3Converters.toFlowable());
        Flowable<IndexNumber> f3 = Multi.createFrom().items(0, 2, 3, 4, 5, 7, 9, 12).map(i -> new IndexNumber(i, 2)).convert().with(MultiRx3Converters.toFlowable());
        Flowable<IndexNumber> f4 = Multi.createFrom().items(4, 7, 9).map(i -> new IndexNumber(i, 3)).convert().with(MultiRx3Converters.toFlowable());

[...]

PR #1218, opened 2021-09-03
[WIP] closes #1183: exponential storage page size and appropriate sizes in all resolvers and fetchers

### LINKED ISSUE #1183: Grow document search page size exponentially
Currently the document API search page is set to 1000 and this is having an negative performance influence to the search. The max document page size is 20 in our API, thus making an initial search for 1000 matching ones when having a persistence filter operation is madness. We should:

* start from the requested document page size and then grow exponentially or use exponential back-off algorithm..
* as page size can also be one, declare a minimum (f.e. 10), but also have maximum safeguard
* keep high number (like 1000) for any population queries as there we need all the rows, thus it's fine

This can be implemented directly in the `QueryExecutor`, where we would replace `pageSize` with `initalPageSize`. In the `execute` when there is a need to fetch the next page, calculate the value from the initial page size and the page being fetched.

In very quick perf test, there was a gain of 3x when search page size was fixed to 40 (double requested document page size), in comparison to 1000. My proposal would be to fetch db page sizes like 20, 40, 80, 160 (or adapting to backoff algorithm)..

ivansenic:
**What this PR does**:

Move away from the crazy storage page size of `1000` when searching for the documents. This PR adds support to increase the page size in the `QueryExecutor` in the exponential fashion. 

Idea is to be able to start fetching matching rows from a reasonably small value, and then increase as we need more. The starting point is usually the requested docs size (increased by 1) as we want to make sure that we don't fetch too many non-needed documents.

For the full searches, we adapted and start with a page size equaling to `16 * numberOfDocuments`. It's because there we fetch whole documents, so 16 is taken as the dummy number representing average number of leafs in each document. I am open for suggestions and adaptations of this number, or any other ideas. Bottom line is again to avoid fetching too many rows unnecessary from the data store.

For population or queries when we need complete doc or sub-docs, we still go for the max storage page size as we want to take all in one hop to the DB.

**This is still a draft and requires some checks and possibly refactoring to:**

* See if we can avoid the +1 increase of the requested doc size in the document resolver (also min 1 in the Or resolver). This has to be tested against C and DSE, cause DSE could have some advantages here.
* Check and confirm performance is way better now

Fixes #1183

**Checklist**
- [ ] Changes manually tested
- [ ] Quick performance tests
- [ ] More Java docs and clarifications


ivansenic:
Just did some search performance tests wrt this PR. Seems that this fix will give us massive improvements as expected. I am gonna wrap up and set this up for review.

**Setup**
* Local test
* Stargate single node, with DSE single node on same machine
* 60 nosqlbench threads

**Results**:
 
image
 
**Observations**:
* As expected huge speed gains in all cases where query storage page size was unnecessary high. 
* Small speed loss for the `Singe in-memory eval` can be ignored as this condition matches all documents, thus previous behavior wins as it's needing one query, vs two now. The reason we need two now is that we are guessing that there are 16 fields per document, and in my test there is 25. Thus, we need two hops here to get the document. In case of smaller docs, we would be faster.
* ANDs are slower than ORs because:
   * they select lower amount of documents
   * they suffer from N+1 described here #1184 
* Test design has to be improved to select same amount of documents in each case, only then can we compare cases between each other. This is quite hard with nosqlbench.

ivansenic:
@EricBorczuk have a look :arrow_up:

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
```diff
@@ -186,16 +221,29 @@ public QueryExecutor(DataStore dataStore) {
...
     if (pagingState != null && pagingState.remaining() == 0) {
       return Flowable.empty();
     }
 
+    AtomicInteger effectivePageSize = new AtomicInteger(pageSize);
     return fetchPage(query, pageSize, pagingState)
         .compose( // Expand BREADTH_FIRST to reduce the number of "proactive" page requests
             FlowableTransformers.expand(
-                rs -> fetchNext(rs, pageSize, query), ExpandStrategy.BREADTH_FIRST, 1));
+                rs -> {
```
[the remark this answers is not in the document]
ivansenic: Very good idea, I created a ticket for this as I guess this requires a lot of wiring and explanations. But again, we should do this as soon as possible :+1:

#1228

restapi/src/main/java/io/stargate/web/docsapi/service/DocsApiConfiguration.java
```diff
@@ -2,21 +2,40 @@
...
 
-  default int getSearchPageSize() {
-    return DocumentDB.SEARCH_PAGE_SIZE;
+  /**
+   * @return Returns approximate storage page size to use when querying database, based on the
+   *     amount of documents we are searching for. We consider that in average documents have 16
+   *     fields.
+   */
+  default int getApproximateStoragePageSize(int numberOfDocuments) {
+    // TODO Should 16 be configurable
```
EricBorczuk, line 22: Yeah, maybe this should be configurable - I've already seen use-cases where real clients have ~50 fields per doc; but I also agree that maybe we should feel this out first, it is possible that 16 will work just fine

restapi/src/main/java/io/stargate/web/docsapi/service/QueryExecutor.java
```diff
@@ -186,16 +221,29 @@ public QueryExecutor(DataStore dataStore) {
...
+                  // in case we have the exponent page size, increase the current value by itself
+                  // but ensure we are never over maximum
+                  int nextPageSize =
+                      exponentPageSize
+                          ? effectivePageSize.updateAndGet(
+                              x -> Math.min(x * 2, DocumentDB.MAX_STORAGE_PAGE_SIZE))
```
EricBorczuk, line 241: :+1:

EricBorczuk (approved):
This looks good! I only had one comment about the name of the `document_search_page_size` parameter and backwards compatibility

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -49,8 +44,8 @@
   public static final int MAX_PAGE_SIZE = 20;
   public static final Integer MAX_DEPTH = Integer.getInteger("stargate.document_max_depth", 64);
   private Boolean useLoggedBatches;
-  public static final Integer SEARCH_PAGE_SIZE =
-      Integer.getInteger("stargate.document_search_page_size", 1000);
+  public static final Integer MAX_STORAGE_PAGE_SIZE =
+      Integer.getInteger("stargate.document_max_storage_page_size", 1000);
```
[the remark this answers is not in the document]
ivansenic: Let's keep the old value here.. It's actually an important setting..

restapi/src/main/java/io/stargate/web/docsapi/service/DocsApiConfiguration.java
ivansenic: It's hard to make it configurable as in different collections you would have different documents.. Let's keep it like this for now..

PR #1225, opened 2021-09-07
Arrays can now be read in every case, even at the root of the document

### LINKED ISSUE #1224: Trying to read a document that is an array results in error
When writing a document that is an array, such as [1, 2, 3] - no error occurs. However upon reading that document a 500 occurs every time. This is a regression as this used to work in previous versions.

EricBorczuk:
**What this PR does**:
Fixes a regression that was recently encountered, where root level array data could be written but could not be read back out.  When reading back out a 500 would occur consistently.

Fixes #1224

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverter.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverter.java
@@ -50,6 +50,24 @@ public class JsonConverter {
     return convertToJsonDoc(rows, collector, writeAllPathsAsObjects, numericBooleans, maxDepth);
   }
 
+  /**
+   * Takes a List of rows from a documents table, iterating over it and constructing a JSON object
+   * (or array) that is represented by those rows. Note: Because the system avoids
+   * read-before-write, it is possible for the same "path" in a document to contain two different
+   * sets of data from two different versions of the document. When this is encountered during the
+   * conversion, the version that has the latest writetime is accepted, and the older version is
+   * collected in the DeadLeafCollector to be purged in the background.
+   *
+   * @param rows Rows from the database
+   * @param collector a DeadLeafCollector implementation, to handle deletion of old and conflicting
+   *     data
+   * @param writeAllPathsAsObjects Instead of writing arrays such as [1, 2], write an object such as
+   *     {"0": 1, "1": 2}
+   * @param numericBooleans If these rows do not support boolean values and are using tinyint
+   *     instead
+   * @param maxDepth The rows' max depth (i.e. the highest pN value)
+   * @return the JsonNode represented by converting the @param rows
+   */
   private JsonNode convertToJsonDoc(
       List<Row> rows,
       DeadLeafCollector collector,
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverter.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverter.java
@@ -121,6 +121,11 @@ public class JsonConverter {
         if (isArray && !writeAllPathsAsObjects) {
           int index = Integer.parseInt(p.substring(1, p.length() - 1));
 
+          boolean shouldBeArray = isArray && !ref.isArray();
+          if (i == 0 && shouldBeArray) {
+            doc = mapper.createArrayNode();
+            ref = doc;
+          }
           ArrayNode arrayRef = (ArrayNode) ref;
 
           int currentSize = arrayRef.size();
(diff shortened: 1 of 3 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverter.java
```diff
@@ -50,6 +50,24 @@ public JsonNode convertToJsonDoc(
     return convertToJsonDoc(rows, collector, writeAllPathsAsObjects, numericBooleans, maxDepth);
   }
 
+  /**
```
tatu-at-datastax, line 53: Really appreciate a good Javadoc here, +1!

tatu-at-datastax:
One beginner question: are root-level arrays expected valid content? I assume so, just wanted to verify as requiring (JSON) Object at root level is something databases could legitimately do.

EricBorczuk:
So I remembered that initially root-level arrays were allowed, and suspected that at some point things regressed, so I looked back in git history; and sure enough there was a point where the code got refactored and the support for it sort of got dropped (and there was no test to cover it)

At any rate, I think it is legitimate to allow root-level docs - search will still work with them (you could search using the array path brackets like `[0].whatever`), and since it's valid JSON I think it is reasonable to allow them as documents!

tatu-at-datastax:
@EricBorczuk this is probably something Lorene could maybe help with once she is back?

restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverter.java
```diff
@@ -121,6 +139,11 @@ private JsonNode convertToJsonDoc(
         if (isArray && !writeAllPathsAsObjects) {
           int index = Integer.parseInt(p.substring(1, p.length() - 1));
 
+          boolean shouldBeArray = isArray && !ref.isArray();
+          if (i == 0 && shouldBeArray) {
```
tomekl007, line 143: why onlt if `i == 0`?

ivansenic: Good point, seems that we would end up in the cast exception if `i` is not `0` and it should be array?

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -535,6 +535,15 @@ public void testPutNullsAndEmpties() throws IOException {
         .isEqualTo(wrapResponse(OBJECT_MAPPER.createArrayNode(), "4", null));
   }
 
+  @Test
+  public void testPutSimpleArray() throws IOException {
```
ivansenic, line 539: should we consider adding other edge cases here? 

* I guess normal string is a valid JSON, should we test with string, blank string empty string?
* same with numbers, booleans
* empty documents (I guess this is not allowed at the moment)

restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverter.java
tatu-at-datastax: If entries are sorted first, wouldn't we always get `[0]` entry first? Unless that was filtered I guess?
Then again there should be no harm in checking it for further entries either way.

EricBorczuk: So I took a look at this code, and yes - for completeness there should be a clause to deal with converting an object to an array when it is not at the root - I added that.

However, because of how row deletions occur before inserts in the API, there isn't a real situation at the moment that causes an object to have to be converted to an array - which is why this issue has never been hit. At any rate, I added a test that should cover this behavior, as well, in case something changes in the future to how we delete data before writing.

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
EricBorczuk: These are all disallowed, and tested by [testInvalidPuts](https://github.com/stargate/stargate/blob/bc9b7f509fb498044353e5d6ff63e39e2d720936/testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java#L665)

restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverter.java
tatu-at-datastax: ahhh. My comment was in error; I did not pay attention and was thinking of 0 as the first element, as opposed to the first path segment.

ivansenic (approved):
Seems good..

PR #1230, opened 2021-09-08
(WIP) Fix #1229: add managed dependency version (29.0-jre) for Guava

### LINKED ISSUE #1229: Replace direct Guava usage by Stargate (use shaded everywhere applicable) to reduce Guava conflict w/ Cassandra backend

tatu-at-datastax:
**What this PR does**:

Adds managed dependency version `29.0-jre` for Guava, removes explicit version definitions from poms

Fixes #1299

**Checklist**
- [n/a] Changes manually tested -- dependency upgrade tested via existing unit, integ, func tests
- [n/a] Automated Tests added/updated -- same as above
- [n/a] Documentation added/updated -- should not change visible behavior.
-


tatu-at-datastax:
Ready for review (CI passed).

ivansenic (changes requested):
It would really be great if we could fix now also the apparently wrong imports of the shadowed guava.. I think every non-persistence module should use guava directly.. If the dependency is not in the pom explicitly we should add it..

Futhermore, maybe we can quickly investigate how can we ban this in future. The shadowed guava should not be on the compile path for those modules.. I have seen other deps to be also available in the compile, but then at run-time not.. I am pritty sure I hit that with `MutableInt (Commons Lang 2.6 API)`, but when using same from `lang3` it works..

@tatu-at-datastax adding request changes, but we can also split in another PR if you like it better..

ivansenic:
@tatu-at-datastax also see Dependabot alert about the Guava version..

tatu-at-datastax:
@ivansenic Ah yes. I was curious about shadowed guava and +1 for removing that if possible. I think I'd like to do that in a separate issue/PR.

Also agree with explicit dependency: I'll see if I can find usages, add those explicitly.
Guava as a dependency can be pretty gnarly.

tatu-at-datastax:
Oh and Dependabot. Yeah, maybe it makes sense to boost default version to even higher -- if I recall, that CVE is not often actually relevant, but all security tools flag it so it's just easier to upgrade. We should be able to do that.

ivansenic (approved):
Changing to approved as we agreed to move suggestions to other PRs...

dougwettlaufer (approved):

tatu-at-datastax:
@ivansenic @dougwettlaufer I will do some of simple changes here still, moving back to WIP. Apologies for confusion. Will let you know once this is ready

config-store-yaml/pom.xml
```diff
@@ -41,7 +35,7 @@
       <dependency>
         <groupId>com.fasterxml.jackson.dataformat</groupId>
         <artifactId>jackson-dataformat-yaml</artifactId>
-        <version>2.11.3</version>
+        <version>2.12.5</version>
```
tatu-at-datastax, line 38: In near future, I'll probably try to add `dependencyManagement` for Jackson (similar to Guava here); challenges are similar as Stargate can use newer/-est versions just fine but C*3.11/4.0/DSE depend on older versions (although I am working on trying to get Jackson update for Cassandra too, concurrently).

grpc/pom.xml
```diff
@@ -52,6 +52,10 @@
       <version>${grpc.version}</version>
     </dependency>
     <!-- Other -->
+    <dependency>
```
tatu-at-datastax, line 55: Has direct usage, adding dependency.

health-checker/pom.xml
```diff
@@ -101,11 +95,6 @@
       <groupId>io.dropwizard</groupId>
       <artifactId>dropwizard-logging</artifactId>
     </dependency>
-    <dependency>
```
tatu-at-datastax, line 104: Removed minimal usage (see below), and hence dependency and exclusions as well.

persistence-test/pom.xml
```diff
@@ -52,9 +52,12 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
+      <!-- 09-Sep-2021, tatu: Cassandra 3.11/4.0 backends depend on Jackson 2.9;
```
tatu-at-datastax, line 55: Could probably have worked around this by forcing `jackson-core` and `jackson-databind` versions to 2.12.5, but due to compatibility concerns with backends decided to limit scope here, only boosting to 2.9.10 that C*/DSE use.

pom.xml
```diff
@@ -38,6 +38,7 @@
     <commons-io.version>2.7</commons-io.version>
     <immutables.version>2.5.6</immutables.version>
     <grpc.version>1.35.0</grpc.version>
+    <guava.version>30.0-jre</guava.version>
```
tatu-at-datastax, line 41: As per Dependabot warning about Guava version <= 29.0

tatu-at-datastax:
@ivansenic @dougwettlaufer I did bit more work as suggested, added notes on changes -- please let me know if there's anything else to address here. I will create a new issue for removing/reducing use of shaded guava dependency for follow-up work (and maybe one for Jackson deps too)

cql/pom.xml
```diff
@@ -63,6 +63,10 @@
       <artifactId>caffeine</artifactId>
       <version>2.8.5</version>
     </dependency>
+    <dependency>
```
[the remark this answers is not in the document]
ivansenic: OK wait, but are you sure there is no transitive deps here of guava, meaning that something else was using Guava but now we override the version? I was thinking to add dep only where shadowed Guava was used and we wanted to move away from it.. Let's sync on this quickly..

health-checker/src/main/java/io/stargate/health/CheckerResource.java
```diff
@@ -40,7 +39,7 @@ public CheckerResource() {
     // for production use with dependency injection
   }
 
-  @VisibleForTesting
+  //  @VisibleForTesting // the only use of Guava from this module
```
ivansenic, line 42: Delete the line?

grpc/pom.xml
ivansenic: Here is the same, some gRPC dep was introducing guava and we now overwrote the version most likely

ivansenic (changes requested):
I have some worries about this.. Let's sync please..

health-checker/src/main/java/io/stargate/health/CheckerResource.java
tatu-at-datastax: > Delete the line?

Sure, will do.

tatu-at-datastax:
@ivansenic I added explicit Guava dependency in cases where module itself depends on guava (grepped for imports), removed if not. So did not leave dependency for transitive usage.

ivansenic:
@tatu-at-datastax What's the state with this and what path are we taking at the end.. Maybe we can talk shortly today?

tatu-at-datastax:
@ivansenic yes, was about to add a note. Let's sync up today.

My plan was to remove usage of un-shaded Guava, along with reducing usage in general where applicable.

cql/pom.xml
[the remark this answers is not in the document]
tatu-at-datastax: Changed inclusion to shaded version.

grpc/pom.xml
tatu-at-datastax: Changed the approach to avoid this issue.

tatu-at-datastax:
Closing this in favor of #1253 (although going through similar issues over there -- but finally making progress).

PR #1232, opened 2021-09-08
Fix #1231: improve WARN message for `MeterRegistryConfiguration` wrt invalid config value

### LINKED ISSUE #1231: Reduce noise by `MeterRegistryConfiguration` validation wrt invalid Double values
When running integration tests (to verify fix for #1229) I noticed that use of invalid expected percentile value (should be [0.0, 1.0]) not only adds WARN message but also includes stack trace. Since stack trace is for `Double.parseDouble()`, it contains no useful information as it just confirms value is not valid. Warning itself also does not quote value used which can be confusing for some values (most notably for empty String).

tatu-at-datastax:
**What this PR does**:

Improves warning message for config value validation by `MeterRegistryConfiguration` so that:

1. Invalid value is added in quotes (helps with things like empty Strings, non-numbers)
2. Removes inclusion of stack trace: in this particular case failure is from `Double.parseDouble()`, details of which do not seem useful (but can clog output logs) -- we already indicate that value is invalid and show that value.

Fixes #12311

**Checklist**
- [-] Changes manually tested -- covered by existing unit tests
- [-] Automated Tests added/updated -- covered by existing unit tests
- [-] Documentation added/updated -- does not change observed behavior from user perspective

### DIFF
--- a/core/src/main/java/io/stargate/core/metrics/impl/MeterRegistryConfiguration.java
+++ b/core/src/main/java/io/stargate/core/metrics/impl/MeterRegistryConfiguration.java
@@ -54,24 +54,15 @@ public final class MeterRegistryConfiguration {
                   p -> {
                     try {
                       double percentile = Double.parseDouble(p);
-                      if (percentile < 0
-                          || percentile > 1
-                          || Double.isNaN(percentile)
-                          || Double.isInfinite(percentile)) {
-                        logger.warn(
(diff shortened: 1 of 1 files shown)


tatu-at-datastax:
Ready for review.

ivansenic (approved):
Ship it :+1:

PR #1234, opened 2021-09-08
Fix #1227: replace gson-backed j[son]surfer with jackson-backed one

### LINKED ISSUE #1227: Replace "jsurfer-gson" with "jsurfer-jackson" (version 1.6.2)
While Stargate uses Jackson for all of its json handling (and YAML config access too), there is also a little bit of GSON usage for `jsurfer` (JSON Surfer -> https://github.com/jsurfer/JsonSurfer). JSurfer does have multiple backends, however, including one for `json-simple` as well as Jackson (https://mvnrepository.com/artifact/com.github.jsurfer/jsurfer-jackson).

It would make sense to unify handling here, for a couple of reasons:

1. Doing this would allow us to drop one more library as a dependency (Gson does not have huge dependencies but still)
2. Jackson is typically a bit faster than GSON for most operations so this could improve performance for parsing
3. When using Jackson for all processing we can avoid conversions between Jackson `JsonNode` and GSON equivalent (JsonValue?) -- if so, this would improve performance and reduce garbage generation.

Note: we should use version `1.6.2` as it has one important fix (https://github.com/jsurfer/JsonSurfer/issues/70) for the jackson backend.

tatu-at-datastax:
Change (as explained in #1227) is to unify JSON handling: Jackson used for everything else, gson only here.

**What this PR does**:

Replaces `jsurfer-gson` (JSON Surfer using GSON for parsing) with `jsurfer-jackson` (JSON Surfer using Jackson for parsing), to unify underlying JSON library.

Fixes #1227

**Checklist**
- [-] Changes manually tested: Automated test coverage exists, no need to manually test.
- [-] Automated Tests added/updated: Existing test suite seems to cover all aspects; no functional changes
- [-] Documentation added/updated: Use of JsonSurfer implementation detail, not documented (nor part of API)

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
@@ -5,13 +5,18 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.fasterxml.jackson.databind.node.ValueNode;
 import com.github.fge.jsonschema.core.exceptions.ProcessingException;
 import com.google.common.base.Splitter;
+/*
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonNull;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
+*/
 import io.stargate.auth.UnauthorizedException;
 import io.stargate.web.docsapi.dao.DocumentDB;
 import io.stargate.web.docsapi.exception.ErrorCode;
@@ -34,7 +39,7 @@ import javax.inject.Inject;
 import javax.ws.rs.core.PathSegment;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.jsfr.json.JsonSurfer;
-import org.jsfr.json.JsonSurferGson;
+import org.jsfr.json.JsonSurferJackson;
 import org.jsfr.json.compiler.JsonPathCompiler;
 import org.jsfr.json.path.JsonPath;
 import org.jsfr.json.path.PathOperator;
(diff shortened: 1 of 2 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -124,7 +116,8 @@ private boolean isEmptyArray(Object v) {
           .configBuilder()
           .bind(
               "$..*",
-              (v, parsingContext) -> {
+              (v0, parsingContext) -> {
+                final JsonNode v = (JsonNode) v0;
```
ivansenic, line 120: Is this cast here safe? Can it be anything else?

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -5,13 +5,9 @@
 import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ValueNode;
 import com.github.fge.jsonschema.core.exceptions.ProcessingException;
 import com.google.common.base.Splitter;
-import com.google.gson.JsonArray;
```
ivansenic, line 10: can we also get rid of the dependency or was this transitive of `jsurfer-gson`?

ivansenic (commented):
I just noticed this is WIP.. Looks good.. Any further optimizations to get clear code in this class are appreciated..

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
tatu-at-datastax: As far as I understand it is safe and the only reason arg comes as `Object` is so that JsonSurfer API is identical across providers -- not sure why it does not use generics instead, but this way json-simple/gson/jackson json value types are opaque. I am guessing some JSON libraries might not have a common base type above `java.lang.Object` (I think org.json and json-lib might be like this).

Jackson specifically always produces `JsonNode` and no other values are possible. I am pretty sure similarly Gson only produced `JsonElement`s.

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
tatu-at-datastax: Gson was only transitively included via `jsurfer-gson`: I checked that bundles had it before change, but not after.
(Grpc has Gson as a separate dependency for grpc api).

tatu-at-datastax:
@ivansenic WIP was my bad, forgot to remove. This is ready for review. Thank you for comments so far.

And yes, I would love to try to clean up the logic here when I have some time. I would like to understand if JsonSurfer really is needed, for one, and beyond that try to simplify processing. It is possible there could be optimizations when handling is simplified: for example wrt concatenation of names.

dougwettlaufer (approved):

PR #1236, opened 2021-09-09
Reference example for gRPC java client

tomekl007:
**What this PR does**:
Adds java examples and documentation to the `README.md` in the gRPC module.

**Checklist**


grpc/README.md
```diff
@@ -1,5 +1,225 @@
...
+For the development purpose, this should be good enough:
+```shell script
+curl -X POST localhost:8081/v1/auth/token/generate --header "Content-Type: application/json" --data '{"key":"cassandra","secret":"cassandra"}'
+```
+assuming that Stargate is running on the `localhost:8081`  
+
+Once we have that, we can connect to the gRPC API. Firstly, we need to generate the `Channel` that is used to perform connection:
+```java
+public ManagedChannel createChannel(String host, int port) {
+    return ManagedChannelBuilder.forAddress(host, port).usePlaintext().build();
```
ivansenic, line 53: maybe add a small note that `usePlaintext()` should only be used for testing and non-sensitive data? Actually removing it could be even better?

tomekl007: @mpenick what's the recommended way for clients to create this channel?
When I drop `usePlaintext()`, I got:
```
Exception in thread "main" io.grpc.StatusRuntimeException: UNAVAILABLE: io exception
Channel Pipeline: [SslHandler#0, ProtocolNegotiators$ClientTlsHandler#0, WriteBufferingAndExceptionHandler#0, DefaultChannelPipeline$TailContext#0]
	at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:262)
	at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:243)
	at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:156)
	at io.stargate.proto.StargateGrpc$StargateBlockingStub.executeQuery(StargateGrpc.java:236)
	at GrpcClient.executeQuery(GrpcClient.java:27)
	at GrpcClient.main(GrpcClient.java:61)
```
locally

mpenick: Ivan is correct, `usePlaintext()` should only be used for development and testing. When using in a production environment it should use a load balancer that terminates TLS or the gRPC needs to support TLS (which it currently doesn't).

tomekl007: Do you have an example from the client perspective how this should be configured?

tomekl007: I mentioned it on:
https://github.com/stargate/stargate/pull/1236/commits/665ce94e4bb88f0d415fd113c93a9e491fbfc9f2

grpc/README.md
```diff
@@ -1,5 +1,227 @@
...
+    value: "some_binary_data"
+  }
+}
+```    
+The value contains the binary data, that we can deserialize. 
+Firstly, we need to unwrap the `ResultSet`:
+```java
+ResultSet rs = response.getResultSet().getData().unpack(QueryOuterClass.ResultSet.class);
+
+rs.getRows(0).getValues(0).getString(); // it will return value for k = "a"
```
ivansenic, line 131: what about if we replace this with the example of get the row list and iterating over, some kind of example that is not so bounded on the type of the query, would this be possible? starter programmers could like this better I guess..

ivansenic (approved):
For me this looks good, consider I never created a Java gRPC client in my life.. Maybe somebody with more experience should also look and add comments.. But for a newbie it should be fine..

grpc/README.md
```diff
@@ -1,5 +1,227 @@
...
+import io.stargate.grpc.StargateBearerToken;
+
+StargateGrpc.StargateBlockingStub blockingStub = StargateGrpc.newBlockingStub(channel)
+                                                    .withCallCredentials(new StargateBearerToken("token-value"))
+                                                    .withDeadlineAfter(5, TimeUnit.SECONDS);
+```
+
+The second one is async (non-blocking):
+```java
+StargateGrpc.StargateStub = = StargateGrpc.newStub(channel)
```
mpenick, line 77: `= =`

grpc/README.md
```diff
@@ -1,5 +1,227 @@
 # gRPC Service
 
+## gRPC Java Client 
+
+### Setup
+
+Let's see how to configure a java project to use the Stargate gRPC API.
+The generated code based on protobuf files (`query.proto` and `stargate.proto`) is shipped with the `grpc-proto` dependency.
+
+To see a guide how it compiles the Java code see: [gRPC setup project dependencies].
```
dougwettlaufer, line 10:
```suggestion
To see a guide how the Java code is compiled from the proto files see: [gRPC setup project dependencies]. This is purely background information and not required for using the client discussed here.
```

When I first clicked the link I was a little confused about how it applies to the client.

grpc/README.md
```diff
@@ -1,5 +1,227 @@
...
+The last missing piece is to add a functional channel service provider. We pick netty:
+```xml
+<dependencies>
+    <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-netty-shaded</artifactId>
+            <version>1.40.1</version>
+    </dependency>
+</dependencies>
+```
```
dougwettlaufer, line 31:
```suggestion
In your client application, you only need to add two dependencies which are the client and a functional channel service provider (we pick netty here).

```xml
<dependencies>
    <dependency>
        <version>1.0.32</version>
    </dependency>
    <dependency>
            <groupId>io.grpc</groupId>
            <artifactId>grpc-netty-shaded</artifactId>
            <version>1.40.1</version>
    </dependency>
</dependencies>
```
```

File: grpc/README.md [hunk 6]
```diff
@@ -1,5 +1,227 @@
...
+Once we have that, we can connect to the gRPC API. Firstly, we need to generate the `Channel` that is used to perform connection:
+```java
+public ManagedChannel createChannel(String host, int port) {
+    return ManagedChannelBuilder.forAddress(host, port).usePlaintext().build();
+}
+```
+Please note that `usePlaintext()` should only be used for development and testing.
+When using in a production environment it should use a load balancer that terminates TLS.
+
+For local development of stargate, it will be:
```
dougwettlaufer, line 59:
```suggestion
For local development of Stargate, it will be:
```

grpc/README.md
```diff
@@ -1,5 +1,227 @@
...
+The value contains the binary data, that we can deserialize. 
+Firstly, we need to unwrap the `ResultSet`:
+```java
+ResultSet rs = response.getResultSet().getData().unpack(QueryOuterClass.ResultSet.class);
+
+rs.getRows(0).getValues(0).getString(); // it will return value for k = "a"
+rs.getRows(0).getValues(1).getInt(); // it will return value for v = 1             
+```
+We can get all rows `getRowsList()` and iterate over the result 
+OR get the specific row using its index and passing it to the `getRows(int index)` method. We picked the formed approach.
```
dougwettlaufer, line 135:
```suggestion
OR get the specific row using its index and passing it to the `getRows(int index)` method. We picked the latter approach in the example above.
```

grpc/README.md
```diff
@@ -1,5 +1,227 @@
...
+Our retrieval query (`SELECT k, v FROM test"`) stated explicitly which columns should be retrieved.
```
dougwettlaufer, line 136:
```suggestion
Our retrieval query (`SELECT k, v FROM ks.test"`) stated explicitly which columns should be retrieved.
```

grpc/README.md
```diff
@@ -1,5 +1,227 @@
...
+Thanks to that, we can safely get the values using their positions via the `getValues()` method.
+The `getString()` and `getInt()` perform deserialization of data. The API provides utility methods for deserialization for more types.
```
dougwettlaufer, line 138:
```suggestion
The `getString()` and `getInt()` perform deserialization of data. These methods were used because we knew the underlying type of the corresponding columns. The API provides utility methods for deserialization for more types as well.
```

Clarifying for the "Well why did you use getString() here instead of X?" question.

grpc/README.md
```diff
@@ -1,5 +1,227 @@
...
+                blockingStub.executeBatch(
+                        QueryOuterClass.Batch.newBuilder()
+                                .addQueries(QueryOuterClass.BatchQuery.newBuilder().setCql("INSERT INTO ks.test (k, v) VALUES ('a', 1)").build())
+                                .addQueries(
+                                        QueryOuterClass.BatchQuery.newBuilder().setCql("INSERT INTO ks.test (k, v) VALUES ('b', 2)").build())
+                                .build());
+
+```
+
+It takes the `Batch` as an argument. A Batch can contain N queries. We are adding two queries via `addQueries` method.
```
dougwettlaufer, line 155:
```suggestion
It takes the `Batch` as an argument. A Batch can contain N queries of type INSERT, UPDATE, or DELETE. We are adding two queries via `addQueries` method.
```

Might also want to consider linking to the batch docs somewhere
https://docs.datastax.com/en/dse/6.0/cql/cql/cql_reference/cql_commands/cqlBatch.html

grpc/README.md
```diff
@@ -1,5 +1,227 @@
...
+               }
+           }
+           @Override
+           public void onError(Throwable throwable) {
+               System.out.println("Error: " + throwable);
+           }
+           @Override
+           public void onCompleted() {
+               System.out.println("completed");
+           }
```
dougwettlaufer, line 182: Maybe pseudo code as comments or suggestions on what should be done here?

grpc/README.md
```diff
@@ -1,5 +1,227 @@
...
+```
+Please note that this is a very simplified version only for demonstration purposes and should not be used on production.
+
+Once we have the Observer, we can pass it to the `executeQuery` method on the async stub:
+```java
+stub.executeQuery(QueryOuterClass.Query.newBuilder().setCql("SELECT k, v FROM ks.test").build(), streamObserver);
+```
+This query will return immediately because it is non-blocking. 
+If your program (or test) is progressing to the end, you may not be able to see the results. 
+Your program may exist before the will arrives.
```
dougwettlaufer, line 193:
```suggestion
Your program may exit before the data arrives. 
```

mpenick (commented):
I was more expecting a runnable project with examples in addition to documentation. Like CQL driver examples.

tomekl007:
> I was more expecting a runnable project with examples in addition to documentation. Like CQL driver examples.

@mpenick I have my local project but on a separate repo. Do we want to add a new module in the `Stargate` repo called `grpc-examples`? If yes, I can move it here.

grpc/README.md
tomekl007: done

mpenick:
`grpc-examples` sounds reasonable to me. Anyone else have an opinion?

tomekl007:
> `grpc-examples` sounds reasonable to me. Anyone else have an opinion?

Ok, I created a new module with runnable examples and docs. Please take a look.

grpc-examples/README.md
```diff
@@ -0,0 +1,236 @@
...
+Once we have the Observer, we can pass it to the `executeQuery` method on the async stub:
+```java
+stub.executeQuery(QueryOuterClass.Query.newBuilder().setCql("SELECT k, v FROM ks.test").build(), streamObserver);
+```
+This query will return immediately because it is non-blocking. 
+If your program (or test) is progressing to the end, you may not be able to see the results. 
+Your program may exit before the data arrives. 
+After some time, when the data arrives, the `streamObserver` will be called.
+
+The output of our program will look like this:
```
tatu-at-datastax, line 205: Looks like we are calling `toString()` on opaque results: is this format documented/defined? It might also be good to know Java type of the result since that is needed to take any action (although for debugging String below is fine). From other documentation I think the default result type is CQL-specific.

PR #1238, opened 2021-09-10
Report OverloadedException as a 429 in HTTP APIs (fixes #1233)

### LINKED ISSUE #1233: Catch OverloadedException in HTTP services and return 429
In the event rate limiting is used and an `org.apache.cassandra.exceptions.OverloadedException` is thrown it's currently not being handled and bubbling up as a 500. Instead this exception should be caught and a [429](https://httpstatuses.com/429) should be returned instead. This is most prevalent in the Docs API due to the number of queries it makes per request but any of our services could have the same problem. 

```java
ERROR [jersey-server-managed-async-executor-7] 2021-09-08 17:48:21,780 ErrorHandler.java:59 - Error when executing request
 java.util.concurrent.CompletionException: org.apache.cassandra.exceptions.OverloadedException: Rate limit reached
 	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:331)
 	at java.base/java.util.concurrent.CompletableFuture.uniApplyNow(CompletableFuture.java:670)
 	at java.base/java.util.concurrent.CompletableFuture.uniApplyStage(CompletableFuture.java:658)
 	at java.base/java.util.concurrent.CompletableFuture.thenApply(CompletableFuture.java:2094)
 	at io.stargate.db.datastore.PersistenceBackedDataStore.batch(PersistenceBackedDataStore.java:190)
 	at io.stargate.db.datastore.PersistenceBackedDataStore.batch(PersistenceBackedDataStore.java:179)
 	at io.stargate.db.query.AsyncQueryExecutor.batch(AsyncQueryExecutor.java:104)
 	at io.stargate.db.query.AsyncQueryExecutor.batch(AsyncQueryExecutor.java:82)
 	at io.stargate.web.docsapi.dao.DocumentDB.executeBatch(DocumentDB.java:391)
 	at io.stargate.web.docsapi.dao.DocumentDB.deleteThenInsertBatch(DocumentDB.java:612)
 	at io.stargate.web.docsapi.service.DocumentService.putAtPath(DocumentService.java:526)
 	at io.stargate.web.docsapi.resources.DocumentResourceV2.lambda$postDoc$0(DocumentResourceV2.java:147)
 	at io.stargate.web.docsapi.resources.DocumentResourceV2.handle(DocumentResourceV2.java:624)
 	at io.stargate.web.docsapi.resources.DocumentResourceV2.postDoc(DocumentResourceV2.java:137)
```

olim7t:
**What this PR does**:
Handle Cassandra's `OverloadedException` specially in HTTP APIs in order to return a 429 response code.

Fixes #1233

**Checklist**
- [x] Changes manually tested
- [ ] ~Documentation added/updated~

### DIFF
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/CassandraFetcherExceptionHandler.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/CassandraFetcherExceptionHandler.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.graphql.schema;
+
+import graphql.execution.DataFetcherExceptionHandlerParameters;
+import graphql.execution.DataFetcherExceptionHandlerResult;
+import graphql.execution.SimpleDataFetcherExceptionHandler;
+import io.stargate.graphql.web.StargateGraphqlContext;
+import org.apache.cassandra.stargate.exceptions.OverloadedException;
+
+public class CassandraFetcherExceptionHandler extends SimpleDataFetcherExceptionHandler {
+
+  public static CassandraFetcherExceptionHandler INSTANCE = new CassandraFetcherExceptionHandler();
+
+  @Override
+  public DataFetcherExceptionHandlerResult onException(
+      DataFetcherExceptionHandlerParameters handlerParameters) {
+    if (handlerParameters.getException() instanceof OverloadedException) {
+      StargateGraphqlContext context = handlerParameters.getDataFetchingEnvironment().getContext();
+      context.setOverloaded();
+    }
+
+    return super.onException(handlerParameters);
+  }
+}
--- a/graphqlapi/src/main/java/io/stargate/graphql/web/StargateGraphqlContext.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/web/StargateGraphqlContext.java
@@ -35,6 +35,7 @@ import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicReference;
 import javax.servlet.http.HttpServletRequest;
+import org.apache.cassandra.stargate.exceptions.OverloadedException;
 
 public class StargateGraphqlContext implements HTTPRequestHeaders {
 
@@ -96,6 +97,10 @@ public class StargateGraphqlContext implements HTTPRequestHeaders {
     return graphqlCache;
   }
 
+  /**
+   * Records the fact that at least one CQL query in the current execution failed with {@link
+   * OverloadedException}. This will be translated into an HTTP 429 error at the resource layer.
+   */
   public void setOverloaded() {
     this.overloaded = true;
   }
--- a/graphqlapi/src/main/java/io/stargate/graphql/web/StargateGraphqlContext.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/web/StargateGraphqlContext.java
@@ -53,6 +53,8 @@ public class StargateGraphqlContext implements HTTPRequestHeaders {
   // For more information.
   private final BatchContext batchContext = new BatchContext();
 
+  private volatile boolean overloaded;
+
   public StargateGraphqlContext(
       HttpServletRequest request,
       AuthorizationService authorizationService,
@@ -92,6 +94,14 @@ public class StargateGraphqlContext implements HTTPRequestHeaders {
 
   public GraphqlCache getGraphqlCache() {
     return graphqlCache;
+  }
+
+  public void setOverloaded() {
+    this.overloaded = true;
+  }
+
+  public boolean isOverloaded() {
+    return overloaded;
   }
 
   /**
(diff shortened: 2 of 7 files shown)


graphqlapi/src/main/java/io/stargate/graphql/schema/CassandraFetcherExceptionHandler.java
```diff
@@ -0,0 +1,38 @@
...
+public class CassandraFetcherExceptionHandler extends SimpleDataFetcherExceptionHandler {
+
+  public static CassandraFetcherExceptionHandler INSTANCE = new CassandraFetcherExceptionHandler();
+
+  @Override
+  public DataFetcherExceptionHandlerResult onException(
+      DataFetcherExceptionHandlerParameters handlerParameters) {
+    if (handlerParameters.getException() instanceof OverloadedException) {
+      StargateGraphqlContext context = handlerParameters.getDataFetchingEnvironment().getContext();
+      context.setOverloaded();
```
olim7t, line 33: `GraphqlResourceBase` checks this flag before completing the response.
This is the only way I have found to propagate the information. If I rethrow, the GraphQL library has a special catch block that will still wrap it into a GraphQL error.

restapi/src/main/java/io/stargate/web/resources/RequestHandler.java
```diff
@@ -45,6 +46,13 @@ public static Response handle(Callable<Response> action) {
                   "Resource not found: " + nfe.getMessage(),
                   Response.Status.NOT_FOUND.getStatusCode()))
           .build();
+    } catch (OverloadedException e) {
+      logger.info("Overloaded", e);
```
olim7t, line 50: Duplicated to mimic the existing code, but logging errors at info feels weird.

olim7t (commented):
This is hard to test since we would have to reproduce the overload error.

I've checked GraphQL manually by temporarily modifying the fetcher code to throw the exception, I'll do the same for the other APIs.

restapi/src/main/java/io/stargate/web/resources/RequestHandler.java
ivansenic: True.. Just imagine the size of the log once we are hitting the `OverloadedException`.. Let's log only the important stuff, something we don't know why is it happening, or?

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlResourceBase.java
```diff
@@ -341,11 +341,21 @@ protected static void executeAsync(
...
     graphql
         .executeAsync(input)
-        .thenApply(result -> asyncResponse.resume(result.toSpecification()))
+        .thenApply(
+            result -> {
+              StargateGraphqlContext context = (StargateGraphqlContext) input.getContext();
+              if (context.isOverloaded()) {
+                replyWithGraphqlError(
+                    Status.TOO_MANY_REQUESTS, "Database is overloaded", asyncResponse);
+                return true;
```
ivansenic, line 350: what's the meaning of this `true` here? shouldn't you return `boolean` as the result of the `replyWithGraphqlError`?

ivansenic (approved):
Looks good, have a look on the comments..

restapi/src/main/java/io/stargate/web/docsapi/resources/error/ErrorHandler.java
```diff
@@ -48,6 +49,12 @@ public Response apply(Throwable throwable) {
                   "Role unauthorized for operation: " + throwable.getMessage(),
                   Response.Status.UNAUTHORIZED.getStatusCode()))
           .build();
+    } else if (throwable instanceof OverloadedException) {
```
tomekl007, line 52: should this be handled for gRPC as well?

restapi/src/main/java/io/stargate/web/resources/RequestHandler.java
olim7t: You're right, this is definitely a bad idea, removing it.

graphqlapi/src/main/java/io/stargate/graphql/web/resources/GraphqlResourceBase.java
olim7t: It's just to respect the return type of the lambda, we don't do anything with the returned value. This was kind of hidden previously when the lambda was just calling `asyncResponse.resume`.

I'm changing the return type to `Void` to clarify things.

restapi/src/main/java/io/stargate/web/docsapi/resources/error/ErrorHandler.java
olim7t: This PR is about HTTP APIs.

But yes, gRPC should also handle `Overloaded` in a protocol-specific way (most likely an error status).

tatu-at-datastax:
Quick note: this will help with #1279 a bit too -- will probably want to add to exception handler being added in this PR.

dougwettlaufer (approved):

graphqlapi/src/main/java/io/stargate/graphql/web/StargateGraphqlContext.java
```diff
@@ -53,6 +53,8 @@
   // For more information.
   private final BatchContext batchContext = new BatchContext();
 
+  private volatile boolean overloaded;
```
tatu-at-datastax, line 56: Maybe add brief Javadoc here (or on setter) mentioning reason; flag to indicate that overload fail (429) was received during fetching.

ivansenic: I can not find this new `volatile` anywhere, is this removed or what s going on?

olim7t: Volatile is enough, we just need to guarantee that if the value was changed, the change will be visible.

We don't need synchronization because:
* the only thing writers can do is set the value to `true`. So even if this happens multiple times, they don't race with each other. (PS: it can happen multiple times, if a GraphQL payload contains multiple operations and more than one `OverloadedException` occurs)
* we read back the value after the execution of the payload (`GraphqlResourceBase`). All exception handlers are guaranteed to have run at that point, since their output is part of the response.

PR #1239, opened 2021-09-10
use cache in AuthnTableBasedService

tomekl007:
**What this PR does**:

Introduces cache in the `AuthnTableBasedService` so that token must not be validated against the data store each time.

Fixes performance for all the calls that end-up on this path. issue #1251

### DIFF
--- a/auth-table-based-service/src/main/java/io/stargate/auth/table/AuthnTableBasedService.java
+++ b/auth-table-based-service/src/main/java/io/stargate/auth/table/AuthnTableBasedService.java
@@ -16,8 +16,8 @@
 package io.stargate.auth.table;
 
 import com.datastax.oss.driver.shaded.guava.common.base.Strings;
-import com.datastax.oss.driver.shaded.guava.common.cache.Cache;
-import com.datastax.oss.driver.shaded.guava.common.cache.CacheBuilder;
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 import io.stargate.auth.AuthenticationService;
 import io.stargate.auth.AuthenticationSubject;
 import io.stargate.auth.UnauthorizedException;
@@ -59,7 +59,7 @@ public class AuthnTableBasedService implements AuthenticationService {
       Boolean.parseBoolean(System.getProperty("stargate.auth_tablebased_init", "true"));
 
   private final Cache<String, AuthenticationSubject> tokenCache =
-      CacheBuilder.newBuilder()
+      Caffeine.newBuilder()
           .expireAfterWrite(Duration.ofSeconds(CACHE_TTL_SECONDS))
           .maximumSize(CACHE_MAX_SIZE)
           .build();
@@ -248,8 +248,16 @@ public class AuthnTableBasedService implements AuthenticationService {
 
     // otherwise, look in the cache and optionally fetch if missing
     try {
-      return tokenCache.get(token, () -> getAuthenticationSubject(token));
-    } catch (ExecutionException e) {
+      return tokenCache.get(
+          token,
+          v -> {
+            try {
+              return getAuthenticationSubject(v);
+            } catch (UnauthorizedException e) {
+              throw new RuntimeException(e);
+            }
+          });
+    } catch (RuntimeException e) {
       // properly inspect the cause of the execution exception
       // and re-throw if UnauthorizedException or RuntimeException
       // otherwise wrap in the RuntimeException
@@ -259,7 +267,7 @@ public class AuthnTableBasedService implements AuthenticationService {
       } else if (cause instanceof RuntimeException) {
         throw (RuntimeException) cause;
       } else {
-        throw new RuntimeException(e);
+        throw e;
       }
     }
   }
--- a/auth-table-based-service/src/main/java/io/stargate/auth/table/AuthnTableBasedService.java
+++ b/auth-table-based-service/src/main/java/io/stargate/auth/table/AuthnTableBasedService.java
@@ -35,7 +35,6 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.Map;
 import java.util.UUID;
-import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import org.apache.cassandra.stargate.db.ConsistencyLevel;
 import org.mindrot.jbcrypt.BCrypt;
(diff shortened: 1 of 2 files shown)


auth-table-based-service/src/main/java/io/stargate/auth/table/AuthnTableBasedService.java
```diff
@@ -230,54 +237,64 @@ public AuthenticationSubject validateToken(String token) throws UnauthorizedExce
...
+                      int timestamp = row.getInt("created_timestamp");
+                      username = row.getString("username");
+
+                      dataStore
+                          .queryBuilder()
+                          .update(AUTH_KEYSPACE, AUTH_TABLE)
+                          .ttl(tokenTTL)
+                          .value("username", username)
+                          .value("created_timestamp", timestamp)
+                          .where("auth_token", Predicate.EQ, UUID.fromString(token))
```
tatu-at-datastax, line 284: Could/should we not pass already decoded `uuid` here, instead of re-decoding?

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthnTableBasedService.java
```diff
@@ -230,54 +237,64 @@ public AuthenticationSubject validateToken(String token) throws UnauthorizedExce
...
+      return (AuthenticationSubject)
+          tokenCache.get(
+              token,
+              (Callable<AuthenticationSubject>)
+                  () -> {
+                    UUID uuid;
+                    try {
+                      uuid = UUID.fromString(token);
+                    } catch (IllegalArgumentException exception) {
+                      throw new UnauthorizedException("authorization failed - bad token");
```
tatu-at-datastax, line 250: Might want to add the invalid value here? ("bad token \"%s\", token)

dougwettlaufer: Probably not a good idea to be logging something like that.

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthnTableBasedService.java
```diff
@@ -51,6 +55,9 @@
   private static final boolean shouldInitializeAuthKeyspace =
       Boolean.parseBoolean(System.getProperty("stargate.auth_tablebased_init", "true"));
 
+  private final Cache<Object, Object> tokenCache =
```
ivansenic, line 58: * Why are we not having properly typed cache?
* Why are we recording stats here?

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthnTableBasedService.java
```diff
@@ -230,54 +237,64 @@ public AuthenticationSubject validateToken(String token) throws UnauthorizedExce
...
-    } catch (InterruptedException | ExecutionException e) {
-      logger.error("Failed to validate token", e);
+      return (AuthenticationSubject)
+          tokenCache.get(
+              token,
+              (Callable<AuthenticationSubject>)
+                  () -> {
+                    UUID uuid;
+                    try {
+                      uuid = UUID.fromString(token);
```
ivansenic, line 248: Is this conversion really necessary or is it there only to confirm token is UUID? Should we use regex matching better?

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthnTableBasedService.java
ivansenic: Fixed in 4e0e19c68b3adf98849671dae4c498a3ffa02c44

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthnTableBasedService.java
ivansenic: Yep no logging..

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthnTableBasedService.java
ivansenic: Fixed in 4e0e19c68b3adf98849671dae4c498a3ffa02c44

ivansenic:
I made updates to the PR, fixing most of the stuff as @tomekl007 is on holiday:

* Fixed comments
* Fixed types
* Update TTL in async way
* Fixed tests

mpenick:
I'm wondering if this is something we want to pursue? The REST/DocDB cache other things along with the authentication. In gRPC, I'm using a similar approach and caching the whole `Connection` object (which is also expensive) instead of just the `AuthenticationSubject` https://github.com/stargate/stargate/commit/6818f6a51a1432fb2079730c73b60db6d9b993aa

ivansenic:
> I'm wondering if this is something we want to pursue? The REST/DocDB cache other things along with the authentication token. In gRPC, I'm using a similar approach and caching the whole `Connection` object (which is also expensive) instead of just the `AuthenticationSubject` [6818f6a](https://github.com/stargate/stargate/commit/6818f6a51a1432fb2079730c73b60db6d9b993aa)

I think from this module perspective is good that we have caching here.. It's even better that other modules cache more high-level things (like `Connection` or `DocumentDB`).. It's win win for me..

auth-table-based-service/src/main/java/io/stargate/auth/table/AuthnTableBasedService.java
```diff
@@ -226,18 +233,39 @@ protected static boolean checkpw(String password, String hash) {
...
+    try {
+      return tokenCache.get(token, () -> getAuthenticationSubject(token));
+    } catch (ExecutionException e) {
+      // properly inspect the cause of the execution exception
+      // and re-throw if UnauthorizedException or RuntimeException
+      // otherwise wrap in the RuntimeException
+      Throwable cause = e.getCause();
+      if (cause instanceof UnauthorizedException) {
+        throw (UnauthorizedException) cause;
+      } else if (cause instanceof RuntimeException) {
```
dougwettlaufer, line 251: Does the cache wrap everything in `ExecutionException`?

ivansenic: Yep and this was confirmed in tests that were failing with 500 instead of unauthorized before I added this..

ivansenic (approved):
I am up for shipping this :+1:

tomekl007:
> I am up for shipping this 👍

I also think that we may benefit from caching at this layer. Firstly, all the APIs that need to validate the token, will benefit from it. Secondly, assuming that someone from the OS community (e.g. when adding a new module) will want to integrate with our `Auth`, they will also benefit from this optimization. @mpenick @dougwettlaufer wdyt?

dougwettlaufer (approved):
Not going to block this but I would prefer that we just cache at the top layer. First, this caching will not apply for other implementation like JWT based or anything custom. Second, this caching will be redundant since we already need to cache the connection in REST and gRPC to improve performance in other aspects.

PR #1240, opened 2021-09-10
Add new endpoints for "built-in" functions, add two built in funcs $push and $pop

### LINKED ISSUE #854: Add ability to "push" to an array path in a document
Similar to javascript's Array.proto.push(), add some functionality to be able to push a value onto an array path.  There are a few things to consider here:

1) The syntax/URL for such an action should be consistent with what exists already, i.e. you should still supply a document path in the URL
2) read-before-write should be avoided, or if it's absolutely necessary, the push functionality should be decoupled from other write operations
3) Needs to error out if a push is attempted on a non-array

EricBorczuk:
**What this PR does**:
Adds the ability to push to or pop from an array in the documents API.

Fixes #854

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/models/BuiltInApiFunction.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/models/BuiltInApiFunction.java
@@ -1,6 +1,8 @@
 package io.stargate.web.docsapi.models;
 
 import com.fasterxml.jackson.annotation.JsonFormat;
+import io.stargate.web.docsapi.exception.ErrorCode;
+import io.stargate.web.docsapi.exception.ErrorCodeRuntimeException;
 import org.codehaus.jackson.annotate.JsonProperty;
 
 @JsonFormat(shape = JsonFormat.Shape.OBJECT)
@@ -29,6 +31,8 @@ public enum BuiltInApiFunction {
         return apiFunc;
       }
     }
-    throw new IllegalArgumentException("No BuiltInApiFunction found for name: " + name);
+    throw new ErrorCodeRuntimeException(
+        ErrorCode.DOCS_API_INVALID_BUILTIN_FUNCTION,
+        "No BuiltInApiFunction found for name: " + name);
   }
 }
--- a/restapi/src/main/java/io/stargate/web/docsapi/models/BuiltInApiFunction.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/models/BuiltInApiFunction.java
@@ -1,0 +1,34 @@
+package io.stargate.web.docsapi.models;
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+import org.codehaus.jackson.annotate.JsonProperty;
+
+@JsonFormat(shape = JsonFormat.Shape.OBJECT)
+public enum BuiltInApiFunction {
+  ARRAY_PUSH("$push", "Appends data to the end of an array"),
+  ARRAY_POP("$pop", "Removes data from the end of an array, returning it");
+
+  @JsonProperty("name")
+  public String name;
+
+  @JsonProperty("description")
+  public String description;
+
+  public boolean requiresValue() {
+    return this == ARRAY_PUSH;
+  }
+
+  BuiltInApiFunction(String name, String description) {
+    this.name = name;
+    this.description = description;
+  }
+
+  public static BuiltInApiFunction fromName(String name) {
+    for (BuiltInApiFunction apiFunc : BuiltInApiFunction.values()) {
+      if (apiFunc.name.equals(name)) {
+        return apiFunc;
+      }
+    }
+    throw new IllegalArgumentException("No BuiltInApiFunction found for name: " + name);
+  }
+}
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
@@ -2,7 +2,6 @@ package io.stargate.web.docsapi.resources;
 
 import static io.stargate.web.docsapi.resources.RequestToHeadersMapper.getAllHeaders;
 
-import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
 import edu.umd.cs.findbugs.annotations.NonNull;
@@ -10,7 +9,6 @@ import io.stargate.web.docsapi.dao.DocumentDB;
 import io.stargate.web.docsapi.examples.WriteDocResponse;
 import io.stargate.web.docsapi.models.DocumentResponseWrapper;
 import io.stargate.web.docsapi.models.MultiDocsResponse;
-import io.stargate.web.docsapi.models.dto.ExecuteBuiltInFunction;
 import io.stargate.web.docsapi.resources.error.ErrorHandler;
 import io.stargate.web.docsapi.service.DocsApiConfiguration;
 import io.stargate.web.docsapi.service.DocsSchemaChecker;
@@ -34,7 +32,6 @@ import java.util.UUID;
 import java.util.concurrent.Callable;
 import javax.inject.Inject;
 import javax.servlet.http.HttpServletRequest;
-import javax.validation.Valid;
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotNull;
 import javax.ws.rs.Consumes;
@@ -307,74 +304,6 @@ public class DocumentResourceV2 {
               .entity(
                   mapper.writeValueAsString(
                       new DocumentResponseWrapper<>(id, null, null, context.toProfile())))
-              .build();
-        });
-  }
-
-  @POST
-  @ManagedAsync
-  @ApiOperation(
-      value =
-          "Execute a built-in function (e.g. $push and $pop) against this document. Performance may vary.")
-  @ApiResponses(
-      value = {
-        @ApiResponse(code = 200, message = "OK", response = WriteDocResponse.class),
-        @ApiResponse(code = 400, message = "Bad request", response = Error.class),
-        @ApiResponse(code = 401, message = "Unauthorized", response = Error.class),
-        @ApiResponse(code = 403, message = "Forbidden", response = Error.class),
-        @ApiResponse(code = 422, message = "Unprocessable entity", response = Error.class),
-        @ApiResponse(code = 500, message = "Internal Server Error", response = Error.class)
-      })
-  @Path("collections/{collection-id}/{document-id}/{document-path: .*}/function")
-  @Consumes({MediaType.APPLICATION_JSON})
-  @Produces(MediaType.APPLICATION_JSON)
-  public Response executeBuiltInFunction(
-      @Context HttpHeaders headers,
-      @Context UriInfo ui,
-      @ApiParam(
-              value =
-                  "The token returned from the authorization endpoint. Use this token in each request.",
-              required = true)
-          @HeaderParam("X-Cassandra-Token")
-          String authToken,
-      @ApiParam(value = "the namespace that the collection is in", required = true)
-          @PathParam("namespace-id")
-          String namespace,
-      @ApiParam(value = "the name of the collection", required = true) @PathParam("collection-id")
-          String collection,
-      @ApiParam(value = "the name of the document", required = true) @PathParam("document-id")
-          String id,
-      @ApiParam(
-              value = "the path in the JSON that you want to execute the operation on",
-              required = true)
-          @PathParam("document-path")
-          List<PathSegment> path,
-      @ApiParam(value = "The operation to perform", required = true) @Valid
-          ExecuteBuiltInFunction payload,
-      @ApiParam(
-              value = "Whether to include profiling information in the response (advanced)",
-              defaultValue = "false")
-          @QueryParam("profile")
-          Boolean profile,
-      @Context HttpServletRequest request) {
-    return handle(
-        () -> {
-          ExecutionContext context = ExecutionContext.create(profile);
-          JsonNode result =
-              documentService.executeBuiltInFunction(
-                  authToken,
-                  namespace,
-                  collection,
-                  id,
-                  payload,
-                  path,
-                  dbFactory,
-                  context,
-                  getAllHeaders(request));
-          return Response.ok()
-              .entity(
-                  mapper.writeValueAsString(
-                      new DocumentResponseWrapper<>(id, null, result, context.toProfile())))
               .build();
         });
   }
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
@@ -2,6 +2,7 @@ package io.stargate.web.docsapi.resources;
 
 import static io.stargate.web.docsapi.resources.RequestToHeadersMapper.getAllHeaders;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
 import edu.umd.cs.findbugs.annotations.NonNull;
@@ -9,6 +10,7 @@ import io.stargate.web.docsapi.dao.DocumentDB;
 import io.stargate.web.docsapi.examples.WriteDocResponse;
 import io.stargate.web.docsapi.models.DocumentResponseWrapper;
 import io.stargate.web.docsapi.models.MultiDocsResponse;
+import io.stargate.web.docsapi.models.dto.ExecuteBuiltInFunction;
 import io.stargate.web.docsapi.resources.error.ErrorHandler;
 import io.stargate.web.docsapi.service.DocsApiConfiguration;
 import io.stargate.web.docsapi.service.DocsSchemaChecker;
@@ -32,6 +34,7 @@ import java.util.UUID;
 import java.util.concurrent.Callable;
 import javax.inject.Inject;
 import javax.servlet.http.HttpServletRequest;
+import javax.validation.Valid;
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotNull;
 import javax.ws.rs.Consumes;
@@ -304,6 +307,71 @@ public class DocumentResourceV2 {
               .entity(
                   mapper.writeValueAsString(
                       new DocumentResponseWrapper<>(id, null, null, context.toProfile())))
+              .build();
+        });
+  }
+
+  @POST
+  @ManagedAsync
+  @ApiOperation(
+      value =
+          "Execute a built-in function (e.g. $push and $pop) against this document. Performance may vary.")
+  @ApiResponses(
+      value = {
+        @ApiResponse(code = 200, message = "OK", response = WriteDocResponse.class),
+        @ApiResponse(code = 400, message = "Bad request", response = Error.class),
+        @ApiResponse(code = 401, message = "Unauthorized", response = Error.class),
+        @ApiResponse(code = 403, message = "Forbidden", response = Error.class),
+        @ApiResponse(code = 422, message = "Unprocessable entity", response = Error.class),
+        @ApiResponse(code = 500, message = "Internal Server Error", response = Error.class)
+      })
+  @Path("collections/{collection-id}/{document-id}/{document-path: .*}/function")
+  @Consumes({MediaType.APPLICATION_JSON})
+  @Produces(MediaType.APPLICATION_JSON)
+  public Response executeBuiltInFunction(
+      @Context HttpHeaders headers,
+      @Context UriInfo ui,
+      @ApiParam(
+              value =
+                  "The token returned from the authorization endpoint. Use this token in each request.",
+              required = true)
+          @HeaderParam("X-Cassandra-Token")
+          String authToken,
+      @ApiParam(value = "the namespace that the collection is in", required = true)
+          @PathParam("namespace-id")
+          String namespace,
+      @ApiParam(value = "the name of the collection", required = true) @PathParam("collection-id")
+          String collection,
+      @ApiParam(value = "the name of the document", required = true) @PathParam("document-id")
+          String id,
+      @ApiParam(
+              value = "the path in the JSON that you want to execute the operation on",
+              required = true)
+          @PathParam("document-path")
+          List<PathSegment> path,
+      @ApiParam(value = "The operation to perform", required = true)
+          @NotNull(message = "payload not provided")
+          @NotBlank(message = "payload must not be empty")
+          @Valid
+          ExecuteBuiltInFunction payload,
+      @Context HttpServletRequest request) {
+    return handle(
+        () -> {
+          JsonNode result =
+              documentService.executeBuiltInFunction(
+                  authToken,
+                  namespace,
+                  collection,
+                  id,
+                  payload,
+                  path,
+                  dbFactory,
+                  getAllHeaders(request));
+          return Response.ok()
+              .entity(
+                  mapper.writeValueAsString(
+                      new DocumentResponseWrapper<>(
+                          id, null, result, ExecutionContext.NOOP_CONTEXT.toProfile())))
               .build();
         });
   }
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
@@ -349,14 +349,17 @@ public class DocumentResourceV2 {
               required = true)
           @PathParam("document-path")
           List<PathSegment> path,
-      @ApiParam(value = "The operation to perform", required = true)
-          @NotNull(message = "payload not provided")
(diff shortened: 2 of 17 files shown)


restapi/src/main/java/io/stargate/web/docsapi/resources/NamespacesResource.java
```diff
@@ -112,6 +116,34 @@ public Response getAllNamespaces(
...
+  public Response getBuiltInFunctions(
+      @ApiParam(
+              value =
+                  "The token returned from the authorization endpoint. Use this token in each request.",
+              required = true)
+          @HeaderParam("X-Cassandra-Token")
+          String token,
+      @ApiParam(value = "The namespace to use for the request.", required = true)
+          @PathParam("namespace-id")
+          final String namespaceName,
```
ivansenic, line 139: validation missing?

restapi/src/main/java/io/stargate/web/docsapi/resources/NamespacesResource.java
```diff
@@ -112,6 +116,34 @@ public Response getAllNamespaces(
         });
   }
 
+  @GET
+  @ManagedAsync
```
ivansenic, line 120: Not sure if this is needed?

restapi/src/main/java/io/stargate/web/docsapi/resources/NamespacesResource.java
```diff
@@ -112,6 +116,34 @@ public Response getAllNamespaces(
...
+                  "The token returned from the authorization endpoint. Use this token in each request.",
+              required = true)
+          @HeaderParam("X-Cassandra-Token")
+          String token,
+      @ApiParam(value = "The namespace to use for the request.", required = true)
+          @PathParam("namespace-id")
+          final String namespaceName,
+      @Context HttpServletRequest request) {
+    List<BuiltInApiFunction> functions = Arrays.asList(BuiltInApiFunction.values());
+    return Response.status(Response.Status.OK)
```
ivansenic, line 142: should we simply have a static ref to this response?

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -308,6 +311,74 @@ public Response putDoc(
...
+        @ApiResponse(code = 400, message = "Bad request", response = Error.class),
+        @ApiResponse(code = 401, message = "Unauthorized", response = Error.class),
+        @ApiResponse(code = 403, message = "Forbidden", response = Error.class),
+        @ApiResponse(code = 422, message = "Unprocessable entity", response = Error.class),
+        @ApiResponse(code = 500, message = "Internal Server Error", response = Error.class)
+      })
+  @Path("collections/{collection-id}/{document-id}/{document-path: .*}/function")
+  @Consumes({MediaType.APPLICATION_JSON})
+  @Produces(MediaType.APPLICATION_JSON)
+  public Response executeBuiltInFunction(
```
ivansenic, line 331: validation for parameters missing?

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -308,6 +311,74 @@ public Response putDoc(
...
+              value = "Whether to include profiling information in the response (advanced)",
+              defaultValue = "false")
+          @QueryParam("profile")
+          Boolean profile,
+      @Context HttpServletRequest request) {
+    return handle(
+        () -> {
+          ExecutionContext context = ExecutionContext.create(profile);
+          JsonNode result =
+              documentService.executeBuiltInFunction(
```
ivansenic, line 364: should we try to make this reactive? we kind of wanted to transit to the reactive way of processing?

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -47,19 +51,22 @@
   private final ObjectMapper mapper;
   private final DocsSchemaChecker schemaChecker;
   private final JsonSchemaHandler jsonSchemaHandler;
+  private final ReactiveDocumentService reactiveDocumentService;
```
ivansenic, line 54: ohh no, let's not bind these two as dependencies.. as said it would be better to implement everything in the `ReactiveDocumentService` and `ReactiveDocumentResourceV2`

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -489,12 +490,125 @@ public ReactiveDocumentService(
...
+              } else if (value instanceof Integer) {
+                arrayData.add(Long.valueOf((Integer) value));
+              } else if (value instanceof Long) {
+                arrayData.add((Long) value);
+              } else {
+                arrayData.add((Double) value);
+              }
+              return arrayData;
+            })
+        .blockingGet();
```
ivansenic, line 548: no, please return reactive type, it's gonna be a mess again..

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -489,12 +490,125 @@ public ReactiveDocumentService(
...
+                finalValue = objectMapper.treeToValue(value, Map.class);
+              } else if (value.isArray()) {
+                finalValue = objectMapper.treeToValue(value, List.class);
+              } else {
+                finalValue = null;
+              }
+              arrayData.remove(arrayData.size() - 1);
+              return Pair.with(arrayData, finalValue);
+            })
+        .blockingGet();
```
ivansenic, line 609: same here..

restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java
```diff
@@ -448,6 +455,77 @@ private DocumentDB maybeCreateTableAndIndexes(
     return idsWritten;
   }
 
+  public JsonNode executeBuiltInFunction(
```
ivansenic, line 458: move to reactive doc service and make reactive, should be easier than mixing the two

restapi/src/test/java/io/stargate/web/docsapi/resources/DocumentResourceV2Test.java
```diff
@@ -214,4 +215,33 @@ public void deleteDocPath() {
 
     assertThat(r.getStatus()).isEqualTo(204);
   }
+
+  @Test
+  public void executeBuiltInFunction() throws JsonProcessingException {
```
ivansenic, line 220: is this really testing anything?

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -3259,6 +3259,244 @@ public void testSearchFullDocWithNestedFields() throws IOException {
                 "{\"a\": {\"value\": 1},\"b\": {\"value\": 2}, \"bb\": {\"value\": 4}}"));
   }
 
+  @Test
+  public void testErrorCasesPushPopFunction() throws IOException {
```
ivansenic, line 3263: test cases for missing function, missing value, etc

ivansenic (changes requested):
IMO this has to be implemented in reactive style.. Everything else is just gonna make more mess..

@EricBorczuk I am open for some pair programming if you need help..

restapi/src/main/java/io/stargate/web/docsapi/resources/NamespacesResource.java
EricBorczuk: All the rest of the endpoints in this file are missing any validation (what validation would be needed?)

ivansenic: My bad, `namespaceName` is the path param, no validation needed there.. Will resolve..

restapi/src/main/java/io/stargate/web/docsapi/resources/ReactiveDocumentResourceV2.java
```diff
@@ -340,6 +348,80 @@ public void searchDoc(
...
+              defaultValue = "false")
+          @QueryParam("profile")
+          Boolean profile,
+      @ApiParam(value = "Unwrap results", defaultValue = "false") @QueryParam("raw") Boolean raw,
+      @Context HttpServletRequest request,
+      @Suspended AsyncResponse asyncResponse) {
+    Single.fromCallable(
+            () -> {
+              DocumentDB db = dbFactory.getDocDataStoreForToken(authToken, getAllHeaders(request));
+              schemaChecker.checkValidity(namespace, collection, db);
```
ivansenic, line 404: let's make a private method to encapsulate those two, seems it's use everywhere..

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -489,12 +498,238 @@ public ReactiveDocumentService(
...
+   * @param db
+   * @param namespace
+   * @param collection
+   * @param documentId
+   * @param processedPath
+   * @param value
+   * @param context
+   * @return a JSON representation of the array after pushing the new value
+   */
+  public Maybe<JsonNode> getArrayAfterPush(
```
ivansenic, line 521: private now?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -489,12 +498,238 @@ public ReactiveDocumentService(
   }
 
   // we need to transform the stuff to support array elements
-  private List<String> processSubDocumentPath(List<String> subDocumentPath) {
+  public List<String> processSubDocumentPath(List<String> subDocumentPath) {
```
ivansenic, line 501: back to private

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -489,12 +498,238 @@ public ReactiveDocumentService(
...
+        .map(
+            array -> {
+              JsonNode data = array.getData();
+              if (!data.isArray()) {
+                throw new ErrorCodeRuntimeException(
+                    ErrorCode.DOCS_API_SEARCH_ARRAY_PATH_INVALID,
+                    "The path provided to push to has no array");
+              }
+              ArrayNode arrayData = (ArrayNode) data;
+              if (value == null) {
```
ivansenic, line 539: isn't there anything available to overcome this crazy if/else sequence? @tatu-at-datastax suggestions?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -489,12 +498,238 @@ public ReactiveDocumentService(
...
+   * @param db
+   * @param namespace
+   * @param collection
+   * @param documentId
+   * @param processedPath
+   * @param context
+   * @return a Pair containing the JSON representation of the array after popping the value and the
+   *     value itself
+   */
+  public Maybe<Pair<ArrayNode, Object>> getArrayAndValueAfterPop(
```
ivansenic, line 572: private

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -489,12 +498,238 @@ public ReactiveDocumentService(
...
+              if (data == null || !data.isArray()) {
+                throw new ErrorCodeRuntimeException(
+                    ErrorCode.DOCS_API_SEARCH_ARRAY_PATH_INVALID,
+                    "The path provided to pop from has no array");
+              }
+              ArrayNode arrayData = (ArrayNode) data;
+              if (arrayData.size() == 0) {
+                throw new ErrorCodeRuntimeException(ErrorCode.DOCS_API_ARRAY_POP_OUT_OF_BOUNDS);
+              }
+              JsonNode value = arrayData.get(arrayData.size() - 1);
```
ivansenic, line 592: why don't we call remove the `remove` here, it will return back the removed `JsonNode`?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -489,12 +498,238 @@ public ReactiveDocumentService(
...
+
+  public Maybe<DocumentResponseWrapper<? extends JsonNode>> executeBuiltInFunction(
+      DocumentDB db,
+      String keyspace,
+      String collection,
+      String id,
+      ExecuteBuiltInFunction funcPayload,
+      List<PathSegment> path,
+      ExecutionContext context)
+      throws UnauthorizedException {
```
ivansenic, line 626: where are we actually doing the authorization here?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -489,12 +498,238 @@ public ReactiveDocumentService(
...
+      String keyspace,
+      String collection,
+      String id,
+      Object valueToPush,
+      List<String> pathString,
+      ExecutionContext context) {
+    return getArrayAfterPush(db, keyspace, collection, id, pathString, valueToPush, context)
+        .map(
+            jsonArray -> {
+              if (jsonArray == null) {
```
ivansenic, line 660: I don't think mapper will ever pass a null here?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -489,12 +498,238 @@ public ReactiveDocumentService(
...
+              db.deleteThenInsertBatch(
+                  keyspace,
+                  collection,
+                  id,
+                  bindParams,
+                  processedPath,
+                  timeSource.currentTimeMicros(),
+                  context.nested("ASYNC INSERT"));
+              return jsonArray;
+            });
```
ivansenic, line 688: seems that this is same as lines 706-730, can we refactor into a mapper method?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -489,12 +498,238 @@ public ReactiveDocumentService(
...
+                finalValue = objectMapper.treeToValue(value, List.class);
+              } else {
+                finalValue = null;
+              }
+              arrayData.remove(arrayData.size() - 1);
+              return Pair.with(arrayData, finalValue);
+            });
+  }
+
+  public Maybe<DocumentResponseWrapper<? extends JsonNode>> executeBuiltInFunction(
```
ivansenic, line 618: can you move this up where all the public methods of this class are and add Java docs please

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
```diff
@@ -122,6 +136,288 @@ private static String convertSingleArrayPath(String path) {
...
+   * @param key The name of the document that will be written
+   * @param payload a JSON object, or a URL-encoded form with the relevant data in it
+   * @param patching If this payload meant to be part of a PATCH request (this causes a small amount
+   *     of extra validation if true)
+   * @param isJson if the request had a content type of application/json, else it will be
+   *     interpreted as a URL encoded form
+   * @return The full bind variable list for the subsequent inserts, and all first-level keys, as an
+   *     ImmutablePair.
+   */
+  public static ImmutablePair<List<Object[]>, List<String>> shredPayload(
```
ivansenic, line 162: this needs proper unit tests imo, the complexity of that path is so huge and we have no single unit tests, I would invest here in order to make sure this is good.. Also not sure if we should extract this to a separate shredding class, we are again getting huge classes..

In addition, we should start optimizing the code, we need to add more comments, split these huge methods into smaller ones, don't expose constants like `"dbl_value"` as plain text, etc..

ivansenic (changes requested):
Added some other concerns and suggestions.. Seems auth is not taken care of?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
EricBorczuk: both `handlePush` and `handlePop` call `getDocument` which calls `getDocumentInternal` (and does the authentication piece)

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
EricBorczuk: Yea I forgot to remove this, it will never be null

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
EricBorczuk: I'll look into this, there's definitely just a method that I don't know about on the mapper or something

restapi/src/main/java/io/stargate/web/docsapi/models/BuiltInApiFunction.java
```diff
@@ -0,0 +1,38 @@
+package io.stargate.web.docsapi.models;
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+import io.stargate.web.docsapi.exception.ErrorCode;
+import io.stargate.web.docsapi.exception.ErrorCodeRuntimeException;
+import org.codehaus.jackson.annotate.JsonProperty;
+
+@JsonFormat(shape = JsonFormat.Shape.OBJECT)
```
tatu-at-datastax, line 8: As long as we only serialize (write) these enums as JSON Objects this is fine; just need to know that deserialization is not supported.

restapi/src/main/java/io/stargate/web/docsapi/resources/ReactiveDocumentResourceV2.java
```diff
@@ -340,6 +348,80 @@ public void searchDoc(
                 asyncResponse, ErrorHandler.EXCEPTION_TO_RESPONSE));
   }
 
+  @POST
+  @ManagedAsync
+  @ApiOperation(
+      value =
+          "Execute a built-in function (e.g. $push and $pop) against this document. Performance may vary.")
```
tatu-at-datastax, line 355: Not sure how terminology is used elsewhere, but wouldn't operation be against a value within document or documents? Value would be the whole document only if document consists of Array value.

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
tatu-at-datastax: Unfortunately I don't think there is anything super-convenient, and would probably prefer explicit traversal of Map/List over use of `valueToTree()` just to be fully explicit (plus there's a bit of overhead with conversion as logically that serializes value as tokens, reads back -- nothing too bad but maybe not worth it here) so there is nothing funny going on (in particular, ensure they are all "simple" values, Numbers, Strings, Booleans, Maps, Lists, no POJOs).

But I would extract this into a helper method, with recursive calls for `Map`, `List`.

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -489,12 +498,238 @@ public ReactiveDocumentService(
...
+                    ErrorCode.DOCS_API_SEARCH_ARRAY_PATH_INVALID,
+                    "The path provided to pop from has no array");
+              }
+              ArrayNode arrayData = (ArrayNode) data;
+              if (arrayData.size() == 0) {
+                throw new ErrorCodeRuntimeException(ErrorCode.DOCS_API_ARRAY_POP_OUT_OF_BOUNDS);
+              }
+              JsonNode value = arrayData.get(arrayData.size() - 1);
+              Object finalValue;
+              if (value.isDouble() || value.canConvertToLong()) {
```
tatu-at-datastax, line 594: I think a comment would be warranted here, to explain what the logic here is for (avoid over-/underflow?)

I have some concerns about our use of `double` for everything (with no integer type) but that's much bigger question & maybe too late (we do Javascript style here :) ); both wrt correctness and efficiency.

restapi/src/main/java/io/stargate/web/docsapi/models/BuiltInApiFunction.java
EricBorczuk: This is actually only being serialized at the moment, for the GET built in functions endpoint

When the "execute function" endpoint is requested, the user sends an `operation` and potentially a `value`, and the String value of `operation` is checked against this enum - but it is not actually used in deserialization

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
EricBorczuk: I agree - but I also think this would be a pretty sizeable chunk of work, and actually this feature is something that a customer was looking to have soon

In the interest of expedience for them, could I make a follow-up ticket to do this work and merge it as-is to get the feature in faster for them?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
EricBorczuk: `arrayNode.insertPOJO(...)` seems to have worked here... is that the wrong way to do this though?

EricBorczuk: Ah I see what you wrote about `no POJOs` - I'll revert this

tatu-at-datastax: Interesting idea! `insertPOJO` will add "opaque" value which will be serialized just fine for normal usage, as if serialized on its own. So yes, it should work; if there were issues a simple test would catch it. That methods just adds "POJONode" wrapper which should be fine. It'd be awkward if `JsonNode` was processed etc but when constructing payload seems like a reasonable choice.
Good thinking!

restapi/src/main/java/io/stargate/web/docsapi/models/BuiltInApiFunction.java
tatu-at-datastax: Ok. And fwtw, we could make it deserialize if we had to, possibly with `@JsonCreator` or if worst came to worst, custom deserializer. So it is not that it can't be done but simply that shape-changing version would be one-way street.

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
EricBorczuk: I started work on this on a separate branch - will fast-follow after merging this, if that's alright

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
tatu-at-datastax: And just realized my "no POJOs" simply meant that if converting from Lists/Maps, it's simplest if there were no other types: but `insertPOJO` / `addPOJO` (for ArrayNode) are perfectly legit for those too. And `Map`s and `List`s etc; anything Jackson can serialize as JSON value is fine when `JsonNode` structure is just for that purpose.

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
ivansenic: but getDocument is a read op? these are write ones?

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
ivansenic: perfect

ivansenic:
@EricBorczuk can you resolve all the comments that are fixed and let me know when I can review the stuff..

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
EricBorczuk: After reading, it calls `DocumentDB#deleteThenInsertBatch` which does write auth

EricBorczuk:
@ivansenic done! I think there are only a couple of conversations that are still relevant :)

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
ivansenic: OK can we at least create a specific class for shredding and create a ticket to add tests and refactor please? 

Also can we make shred methods in non-static and make the class `@Inject DocsApiConfiguration docsApiConfiguration` so that we can expose methods that take defaults for the `docsApiConfiguration.getMaxDepth() & docsApiConfiguration.getMaxArrayLength()`? This way we don't need to have `DocsApiConfiguration` in the `ReactiveDocumentService` and any users of shred service as well?

ivansenic (approved):
Approved it, but I have one last request :pray:

PR #1241, opened 2021-09-13
Add a gRPC interceptor that caches `Connection`

mpenick:
Profiling found that validating the token using the authentication
service and constructing a new `Connection` object was consuming a
significant percentage of a request's time.

This adds a new interceptor called `NewConnectionInterceptor` which
combines the functionality of the authentication, headers, and remote
address interceptor and caches the resulting `Connection` object

**Checklist**
- [X] Changes manually tested
- ~[ ] Documentation added/updated~


grpc/src/main/java/io/stargate/grpc/service/interceptors/NewConnectionInterceptor.java
```diff
@@ -0,0 +1,128 @@
...
+  private static Map<String, String> convertAndFilterHeaders(Metadata headers) {
+    Map<String, String> stringHeaders = new HashMap<>();
+    for (String key : headers.keys()) {
+      // Ignore gRPC specific (in addition to binary) headers because they're not used by Connection
+      // and some of the values contain unique values for each request which prevents effective
+      // caching.
+      if (key.endsWith("-bin") || key.startsWith("grpc-")) {
+        continue;
+      }
+      String value = headers.get(Key.of(key, Metadata.ASCII_STRING_MARSHALLER));
```
ivansenic, line 121: I know this is from old code, but isn't there an option to iterate key/value pairs instead of pulling the value here?

grpc/src/main/java/io/stargate/grpc/service/interceptors/NewConnectionInterceptor.java
```diff
@@ -0,0 +1,128 @@
...
+              .headers(convertAndFilterHeaders(headers))
+              .remoteAddress(call.getAttributes().get(Grpc.TRANSPORT_ATTR_REMOTE_ADDR))
+              .build();
+
+      Connection connection = connectionCache.get(info);
+
+      Context context = Context.current();
+      context = context.withValue(GrpcService.CONNECTION_KEY, connection);
+      return Contexts.interceptCall(context, call, headers, next);
+    } catch (Exception e) {
```
ivansenic, line 82: can some other exception be thrown here, so we report falsely the `UNAUTHENTICATED`? Maybe we can split the catch into multiple ones with correct mapping?

grpc/src/main/java/io/stargate/grpc/service/interceptors/NewConnectionInterceptor.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  private Connection newConnection(RequestInfo info) throws UnauthorizedException {
+    AuthenticationSubject authenticationSubject = authenticationService.validateToken(info.token());
+
+    AuthenticatedUser user = authenticationSubject.asUser();
+    Connection connection;
+    if (!user.isFromExternalAuth()) {
+      SocketAddress remoteAddress = info.remoteAddress();
+      InetSocketAddress inetSocketAddress = DUMMY_ADDRESS;
```
ivansenic, line 96: what's the deal with `DUMMY_ADDRESS` :smile: ?

ivansenic (approved):
Looks good, did not look into tests in detail..

ivansenic (commented):
Ah tests are failing, need to revoke the approval :cry:

ivansenic (commented):
OK I need to explicitly request changes to remove my approval.. @mpenick sorry for the confusion here..

grpc/src/main/java/io/stargate/grpc/service/interceptors/NewConnectionInterceptor.java
mpenick: It looks like `Metadata` doesn't implement `Map<K,V>` and doesn't have that funcationality.

grpc/src/main/java/io/stargate/grpc/service/interceptors/NewConnectionInterceptor.java
mpenick: This is a best effort attempt to put the correct remote address in the `ClientState` object. If the remote address is not an `InetSocketAddress` I just put a dummy value in there instead of erroring out for something that's not critical. It will almost never not be a `InetSocketAddress` though.

mpenick:
@ivansenic Thanks for the review. I fixed the tests. It's because we're caching now, the `ClientState` can hold keyspace state between requests which was not possible before. Instead of using the response type `SetKeyspace` the code has to actually prevent the possibility of `USE <keyspace>` statements from being executed.

grpc/src/main/java/io/stargate/grpc/service/interceptors/NewConnectionInterceptor.java
```diff
@@ -0,0 +1,137 @@
...
+
+  public static final Metadata.Key<String> TOKEN_KEY =
+      Metadata.Key.of("X-Cassandra-Token", Metadata.ASCII_STRING_MARSHALLER);
+
+  private static final InetSocketAddress DUMMY_ADDRESS = new InetSocketAddress(9042);
+
+  private final Persistence persistence;
+  private final AuthenticationService authenticationService;
+  private final LoadingCache<RequestInfo, Connection> connectionCache =
+      Caffeine.newBuilder().expireAfterWrite(Duration.ofMinutes(1)).build(this::newConnection);
```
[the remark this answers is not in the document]
tomekl007: Shouldn't we limit the size of this cache? Right now is unbounded, and in a hypothetical case when there is really high traffic, this cache may grow a lot, resulting in OOM.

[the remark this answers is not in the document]
mpenick: > Can we expose a Gauge (Metric) that measures and reports the size of this cache?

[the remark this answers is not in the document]
mpenick: > Can we expose a Gauge (Metric) that measures and reports the size of this cache?

There are other caches that could benefit from a custom `StatsCounter` that uses our metrics setup. We should add this in another issue.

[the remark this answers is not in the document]
mpenick: Added https://github.com/stargate/stargate/issues/1249

tomekl007 (approved):

PR #1250, opened 2021-09-14
CQL: Expose backpressure options and set sensible defaults

olim7t:
**What this PR does**:

Fix the initialization of the CQL module to correctly configure the built-in backpressure mechanism.

Also expose the options to limit the number of connections. They're unset by default, but we might want to set them in Astra in the future.

N/A

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/cql/src/main/java/io/stargate/cql/CqlActivator.java
+++ b/cql/src/main/java/io/stargate/cql/CqlActivator.java
@@ -99,6 +99,19 @@ public class CqlActivator extends BaseActivator {
       c.rpc_address = listenAddress;
       c.native_transport_port = cqlPort;
 
+      c.native_transport_max_concurrent_connections =
+          Long.getLong("stargate.cql.native_transport_max_concurrent_connections", -1);
+      c.native_transport_max_concurrent_connections_per_ip =
+          Long.getLong("stargate.cql.native_transport_max_concurrent_connections_per_ip", -1);
+      c.native_transport_max_concurrent_requests_in_bytes =
+          Long.getLong(
+              "stargate.cql.native_transport_max_concurrent_requests_in_bytes",
+              Runtime.getRuntime().maxMemory() / 10);
+      c.native_transport_max_concurrent_requests_in_bytes_per_ip =
+          Long.getLong(
+              "stargate.cql.native_transport_max_concurrent_requests_in_bytes_per_ip",
+              Runtime.getRuntime().maxMemory() / 40);
+
       return c;
     } catch (UnknownHostException e) {
       throw new RuntimeException(e);


cql/src/main/java/io/stargate/cql/CqlActivator.java
```diff
@@ -99,6 +99,19 @@ private static Config makeConfig() {
...
       c.native_transport_port = cqlPort;
 
+      c.native_transport_max_concurrent_connections =
+          Long.getLong("stargate.cql.native_transport_max_concurrent_connections", -1);
+      c.native_transport_max_concurrent_connections_per_ip =
+          Long.getLong("stargate.cql.native_transport_max_concurrent_connections_per_ip", -1);
+      c.native_transport_max_concurrent_requests_in_bytes =
+          Long.getLong(
+              "stargate.cql.native_transport_max_concurrent_requests_in_bytes",
+              Runtime.getRuntime().maxMemory() / 10);
```
olim7t, line 109: This is what OSS Cassandra uses ([source](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/config/DatabaseDescriptor.java#L534-L542)). We might want to lower them in Stargate since there are potentially other workloads running in the same VM.

cql/src/main/java/io/stargate/cql/CqlActivator.java
```diff
@@ -99,6 +99,19 @@ private static Config makeConfig() {
       c.rpc_address = listenAddress;
       c.native_transport_port = cqlPort;
 
+      c.native_transport_max_concurrent_connections =
+          Long.getLong("stargate.cql.native_transport_max_concurrent_connections", -1);
```
tomekl007, line 103: do we want it to be unbounded? Wat value those had before?
(this and `stargate.cql.native_transport_max_concurrent_connections_per_ip"`)

Is it possible to add an Integration Test for it?

cql/src/main/java/io/stargate/cql/CqlActivator.java
```diff
@@ -99,6 +99,19 @@ private static Config makeConfig() {
...
+      c.native_transport_max_concurrent_connections_per_ip =
+          Long.getLong("stargate.cql.native_transport_max_concurrent_connections_per_ip", -1);
+      c.native_transport_max_concurrent_requests_in_bytes =
+          Long.getLong(
+              "stargate.cql.native_transport_max_concurrent_requests_in_bytes",
+              Runtime.getRuntime().maxMemory() / 10);
+      c.native_transport_max_concurrent_requests_in_bytes_per_ip =
```
tomekl007, line 110: what does it mean in our context (per `ip`)? How this is handled?

cql/src/main/java/io/stargate/cql/CqlActivator.java
ivansenic: I guess these options are deprecated now or?

cql/src/main/java/io/stargate/cql/CqlActivator.java
ivansenic: Good point, but the more I think about how much to lower it the more I think we should go with the Cassandra defaults.. Although 10% seems a lot, this is just a max and it's not like occupy this amount all the time.. Not sure, if you decide to decrease, then I would say max by factor 2..

cql/src/main/java/io/stargate/cql/CqlActivator.java
ivansenic: Per client IP I guess..

ivansenic (approved):
Seems good imo..

cql/src/main/java/io/stargate/cql/CqlActivator.java
tomekl007: Yeah, it is described in: https://cassandra.apache.org/blog/Improving-Apache-Cassandras-Front-Door-and-Backpressure.html

tjake (commented):
Can we also add the disable / enable flag?

cql/src/main/java/io/stargate/cql/CqlActivator.java
olim7t: It was unbounded in Stargate before this PR, and it's unbounded by default in OSS Cassandra.

olim7t:
> Can we also add the disable / enable flag?

Do you mean `Config.back_pressure_enabled`? It applies to inter-node backpressure (`BackpressureStrategy`), not the native transport one. Also it's deprecated and not referenced anywhere in the code.

I don't see any other mechanism to disable backpressure, but setting the two `max_concurrent_requests_in_bytes` options to `Long.MAX_VALUE` would be a way to achieve that.

tjake:
There is a flag https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/config/Config.java#L196

`native_transport_rate_limiting_enabled`

olim7t:
> `native_transport_rate_limiting_enabled`

That one is from a very recent ticket: [CASSANDRA-16663](https://issues.apache.org/jira/browse/CASSANDRA-16663), scheduled for 4.1.
I missed it because I was looking at the 4.0.0 sources.

It's a new rate-based mechanism (`NonBlockingRateLimiter`) that coexists with the existing bytes-in-flight limits (`max_concurrent_requests_in_bytes`). But unsetting this flag only disables the rate-base limit, not the in-flight based limit.

We haven't backported the rate-based limit in Stargate yet, but I will look into it.

tjake:
Ok, we can do that in a separate ticket

cql/src/main/java/io/stargate/cql/CqlActivator.java
tatu-at-datastax: As the general practice it seems that unbounded is not a good choice? I realize that we do have to worry about backwards compatibility but maybe some high enough but bounded value would be better?

PR #1253, opened 2021-09-15
Remove use of unshaded guava (fix #1229)

### LINKED ISSUE #1229: Replace direct Guava usage by Stargate (use shaded everywhere applicable) to reduce Guava conflict w/ Cassandra backend

tatu-at-datastax:
**What this PR does**:

Replaces uses of plain Guava with either shaded version or (in some cases) JDK 8 methods: the goal being to avoid version conflict with Guava that persistence implementations (esp. Cassandra 3.11) bring in .

Direct non-shaded Guava usage replaced with shaded variant from all modules except following:

* Persistence backends (3.11, 4.0, dse): these are affected by underlying implementations (but may be tackled next)
* gRPC: module changing rapidly, will submit separate PR

In some cases replaces bit of Guava usage with JDK 8 equivalents (only where usage was very light)

Fixes #1229 (partially, will need follow-up PRs as indicated above)

**Checklist**
- [n/a] Changes manually tested - No functionality changes relying on automated tests
- [n/a] Automated Tests added/updated - No functional changes so existing test coverage should be sufficient
- [na/] Documentation added/updated - as per above, no functional changes, use of Guava impl detail.


tatu-at-datastax:
Ok, finally working around the initial `BundleException`, as well as following `NoClassDefFoundError`s. Findings:

* For some reason, adding OSGI-Import for shaded guava causes problems: it should NOT typically be imported (left imports for persistence components for now as they were there but may need to remove). Removal here resolved `BundleException`s
* Actual dependency is definitely needed (without it, an auth component gave `NoClassDefFoundError`

I am guessing that there's some "interesting" interaction between OSGi declarations (across Bundles) and inclusion of jars within bundles, and those need to align. Shaded guava seems very much like private dependency so in that sense it seems somewhat logical.
But I wish I fully understood workings; and that there was a quicker way to find issues during unit test phase somehow (basic Bundle load,  not even running anything).

restapi/pom.xml
```diff
@@ -253,7 +253,6 @@
                             com.codahale.metrics,
                             com.codahale.metrics.*,
                             io.micrometer.core.*,
-                            com.datastax.oss.driver.*,
```
tatu-at-datastax, line 256: Whops. No, this is not the guava shade, needs to go back.

tatu-at-datastax:
@ivansenic Finally all tests passed so ready for review once last round passes (I removed a set of what I believe are unnecessary exclusions). As per updated description, left `grpc` for a follow-up PR (since it's being worked on by others concurrently); and persistence backends are another open question as they might be trickier.

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthnJwtService.java
```diff
@@ -69,7 +68,7 @@ public String createToken(String key, Map<String, String> headers) {
    */
   @Override
   public AuthenticationSubject validateToken(String token) throws UnauthorizedException {
-    if (Strings.isNullOrEmpty(token)) {
+    if ((token == null) || token.isEmpty()) {
```
tatu-at-datastax, line 71: Normally I'm all for using helper methods, but here replacement allows us to get rid of the whole Guava dep (shaded or not). Ditto for isolated cases of removing "@VisibleForTesting".

auth-jwt-service/src/main/java/io/stargate/auth/jwt/AuthzJwtService.java
```diff
@@ -75,7 +74,6 @@ public ResultSet authorizedDataRead(
     return result.withRowInspector(row -> hasCorrectClaims(stargateClaims, row));
   }
 
-  @VisibleForTesting
```
tatu-at-datastax, line 78: As per above only removed to get rid of the whole guava dependency for module.

auth-table-based-service/pom.xml
```diff
@@ -96,7 +95,6 @@
               org.apache.cassandra.stargate,
               org.apache.cassandra.stargate.*,
               org.javatuples,
-              com.datastax.oss.driver.shaded.guava.*,
```
tatu-at-datastax, line 99: For some reason this has to be removed; I think it would try to "pull-in" dependency from other modules instead of using a private copy. If we wanted to, we could alternatively Export shaded Guava from some other module, I suspect, but it took a while to get everything to work this way.

health-checker/src/main/java/io/stargate/health/CheckerResource.java
```diff
@@ -6,7 +6,6 @@
 import com.codahale.metrics.health.HealthCheck;
 import com.codahale.metrics.health.HealthCheck.Result;
 import com.codahale.metrics.health.HealthCheckRegistry;
-import com.google.common.annotations.VisibleForTesting;
```
tatu-at-datastax, line 9: As with earlier auth-modules, Guava usage in `health-checker` minimal so decided to pull out/replace this particular usage.

restapi/pom.xml
```diff
@@ -253,6 +247,7 @@
                             com.codahale.metrics,
                             com.codahale.metrics.*,
                             io.micrometer.core.*,
+                            !com.datastax.oss.driver.shaded.guava.*,
```
tatu-at-datastax, line 250: We do need the driver code, but nothing exports shaded guava, and hence this is needed (OSGi error otherwise). Unfortunate that Guava shade is under some package as main driver.

auth-table-based-service/pom.xml
ivansenic: OK I investigated here a bit more, I am kinda afraid this removal of the `com.datastax.oss.driver.shaded.guava.*,` and if it is good.. 

Bottom line is this module depends on the `core`, `authnz` and `persistance-api` modules.. And it imports set of different classes from these modules..
```
io.stargate.core.*,
io.stargate.auth,
io.stargate.db.*,
org.apache.cassandra.stargate.*,
```

Now what happens if some of the modules above depend on the (1) shaded guava and (2) use shaded guava in some of the classes above.. I think then when this module tries to uses that classes, we would end up in the linkage errors.. Because the guava classes would be loaded with different class loader?

To overcome this, imo we should:
* remove the explicit dependency to the shaded guava (and thus depend on the version that `persistence-api` is providing)
* return the imported packages..

We need to do this for every module that depends on the `persistence-api`.. For other modules that don't, and thus don't have the shaded guava on their compile path we can declare explicitly..

ivansenic (commented):
Tatu looks great, I am really thankful for taking the effort to solve this.. 

I have only one important OSGi thing about the importing of classes, let's just try to see what up with that.. Otherwise I would approve it..

config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreYaml.java
```diff
@@ -16,16 +16,16 @@
 package io.stargate.config.store.yaml;
 
 import com.codahale.metrics.MetricRegistry;
+import com.datastax.oss.driver.shaded.guava.common.annotations.VisibleForTesting;
+import com.datastax.oss.driver.shaded.guava.common.base.Ticker;
+import com.datastax.oss.driver.shaded.guava.common.cache.CacheBuilder;
+import com.datastax.oss.driver.shaded.guava.common.cache.CacheLoader;
```
dougwettlaufer, line 22: @tatu-at-datastax should we open an issue to replace this with caffeine?

auth-api/pom.xml
```diff
@@ -45,10 +45,6 @@
       <version>${project.version}</version>
       <scope>provided</scope>
       <exclusions>
-        <exclusion>
```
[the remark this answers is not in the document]
tatu-at-datastax: After changes it would not have a dependency. Although I did not check dependency graph; maybe I should?

config-store-yaml/pom.xml
```diff
@@ -64,9 +58,9 @@
         <scope>test</scope>
       </dependency>
       <dependency>
-        <groupId>com.google.guava</groupId>
-        <artifactId>guava</artifactId>
-        <version>29.0-jre</version>
+        <groupId>com.datastax.oss</groupId>
+        <artifactId>java-driver-shaded-guava</artifactId>
+        <version>25.1-jre</version>
```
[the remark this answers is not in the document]
tatu-at-datastax: Ah yes, I forgot this part about dependencyManagement; will add. Not sure about adding dependency itself on parent; that's a possibility I guess.
Not 100% sure how it'd interact with OSGi; perhaps all modules should then add OSGi import.

config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreYaml.java
tatu-at-datastax: Good point, yes.

tatu-at-datastax: Created: https://github.com/stargate/stargate/issues/1255

tatu-at-datastax:
All still working after making changes suggested by @ivansenic -- ready for review again.

auth-table-based-service/pom.xml
ivansenic: OK but here, we did not revert the import? and guava is in the `test` scope?

config-store-yaml/pom.xml
```diff
@@ -49,6 +43,11 @@
         <version>3.17.2</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
+        <groupId>com.datastax.oss</groupId>
+        <artifactId>java-driver-shaded-guava</artifactId>
+	<scope>provided</scope>
```
ivansenic, line 49: bad formatting?

cql/pom.xml
```diff
@@ -128,7 +133,7 @@
                             org.javatuples,
                             com.codahale.metrics,
                             io.micrometer.core.*,
-                            com.datastax.oss.driver.shaded.guava.*,
+                            com.datastax.oss.driver.shaded.guava.common.*,
```
ivansenic, line 136: so what's with the `.common` extra?

graphqlapi/pom.xml
```diff
@@ -42,9 +42,8 @@
       <version>${dropwizard.metrics.version}</version>
     </dependency>
     <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-      <version>29.0-jre</version>
+      <groupId>com.datastax.oss</groupId>
+      <artifactId>java-driver-shaded-guava</artifactId>
```
ivansenic, line 46: here the scope is missing?

persistence-api/pom.xml
```diff
@@ -60,7 +60,6 @@
     <dependency>
       <groupId>com.datastax.oss</groupId>
       <artifactId>java-driver-shaded-guava</artifactId>
-      <version>25.1-jre</version>
```
ivansenic, line 63: no scope?

restapi/pom.xml
```diff
@@ -112,9 +106,8 @@
       <artifactId>dropwizard-logging</artifactId>
     </dependency>
     <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-      <version>27.0-jre</version>
+      <groupId>com.datastax.oss</groupId>
+      <artifactId>java-driver-shaded-guava</artifactId>
```
ivansenic, line 110: no scope

ivansenic (changes requested):
+1 for all the changes in the persistence modules.. 

However, the dependency scope is not consistent now, I saw `default`, `provided` and `test` in different modules.. Also OSGi import of the classes is not returned to all the modules that depend on the persistence-api..

I am requesting changes only to understand what's the strategy we follow here, seems not consistent.

auth-table-based-service/pom.xml
tatu-at-datastax: Correct: Guava (shaded) is only needed by test code for this module. As to Import, as far as I understand it this is not needed for (unit) test code, only for runtime; this is why I did not add it back.

cql/pom.xml
tatu-at-datastax: I saw an issue earlier for odd import of packages under "shaded.guava." (for something like `...guava.appengine.api`), and everything we need is under "shaded.guava.common.". So it is just for minimal inclusion -- seems like we should use this limited inclusion everywhere. Would make sense to be consistent.

graphqlapi/pom.xml
tatu-at-datastax: good catch, will add.

config-store-yaml/pom.xml
tatu-at-datastax: Accidental tab it seems, will fix.

persistence-api/pom.xml
tatu-at-datastax: Need to add. Curious if "provided" works here or not, will try (that is, does anything need to expose as "compile")

restapi/pom.xml
tatu-at-datastax: added "provided.

restapi/pom.xml
tatu-at-datastax: Note: was removed by a further commit.

config-store-yaml/pom.xml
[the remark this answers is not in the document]
tatu-at-datastax: Added main-level managed version.

config-store-yaml/pom.xml
tatu-at-datastax: And on scope/import: as per our out-of-band discussion, only test dependency here for which no Imports needed.

tatu-at-datastax:
More adds due to @ivansenic's feedback:

* Add missing scopes ("provided") 

Otherwise should be good to go after CI builds pass.

tatu-at-datastax:
> +1 for all the changes in the persistence modules..
> 
> However, the dependency scope is not consistent now, I saw `default`, `provided` and `test` in different modules.. Also OSGi import of the classes is not returned to all the modules that depend on the persistence-api..
> 
> I am requesting changes only to understand what's the strategy we follow here, seems not consistent.

Made the changes requested here, had discussions, so I think we are good once CI passes.

auth-table-based-service/pom.xml
ivansenic: I would still return the import.. Why? because tomorrow somebody can use a shaded guava in this module and he will forget to return the import! Plus, we would have the same approach in all the modules.. And this should be applied to all modules imo..

ivansenic (approved):
I am approving, but would still like to have those OSGi imports back at every module that depends on `persistence-api`.. @tatu-at-datastax Wdty?

tatu-at-datastax:
> I am approving, but would still like to have those OSGi imports back at every module that depends on `persistence-api`.. @tatu-at-datastax Wdty?

As I said, I do not think (unit) test-dependencies need these: for runtime I agree.
I thought I had made changes for this to be the case. Are there modules that I missed?
Here are modules as I can see them (looking at poms that refer to "shaded" in some way):

* `auth-table-based-service/`: test-only (no import)
* `config-store-yaml`: runtime, has Import
* `cql`: runtime, has Import
* `graphqlapi`: runtime, has Import
* `grpc`: not yet changed in this PR, wait until next
* `persistence-api`:runtime, has Import
* Persistence impls: missing dependency, has imports
* `restapi`: runtime, has Import
* `testing`: runtime, but Does not have OSGi Imports

So, one missing part I can see now are persistence impls: they are missing dependency but have Imports.
Is this what you were thinking of as missing?
I also didn't tough `grpc` since I thought merging that might be better done separately (try to avoid merge conflicts right now).

tatu-at-datastax:
As per our discussions, changes to make before merge:

* Since change from `test` dep to `provided` ALSO requires OSGi Import add, add comments in relevant `pom.xml`(s) (auth-table-based-service/)
* Add missing "provided" dependency for persistence implementations, for sake of completeness, documentation.
* Leave grpc change for another PR (to follow immediately after getting this one merged).

config-store-yaml/pom.xml
```diff
@@ -49,6 +43,11 @@
         <version>3.17.2</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
+        <groupId>com.datastax.oss</groupId>
+        <artifactId>java-driver-shaded-guava</artifactId>
+        <scope>provided</scope>
```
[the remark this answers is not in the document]
tatu-at-datastax: Above discussed off-line: there are two (or more) ways to do this & question is how explicit we want to be.
To address first issue, a comment added to pom.xml; for second one, inclusion is mostly for visibility (this module does have usage) -- it is not strictly necessary as transitive dependency would indeed work with OSGi import.

Resolving, but we probably really need to document Guava guidelines in READMEs or so.

tatu-at-datastax:
Note: merged second part (#1260) to actually complete this task.

Now the one and only dependency to un-shaded Guava is from `persistence-cassandra-4.0`, class `StargateSystemKeyspace.java`, which is due to underlying Cassandra leaking one type (`ListenableFuture`).

PR #1256, opened 2021-09-20
Use `bytes` types to encode UUID types

### LINKED ISSUE #1248: Change representation of `Uuid` type to use `bytes`
`Uuid` currently uses a `uint64` to represent the most and least significant bytes of a UUID. It's been suggested that a better representation for most would be a simple `bytes` type.

Current representation:
```proto
// A UUID value (for CQL types uuid and timeuuid).
message Uuid {
  // TODO: Is this going to have endian problems representing it this way?
  fixed64 msb = 1;
  fixed64 lsb = 2;
}

message Value {
  message Null {}
  message Unset {}
  oneof inner {
     
   // ...

    // CQL types: blob, inet, custom
    bytes bytes = 8;

    // CQL types: uuid, timeuuid
    Uuid uuid = 9;

  }
}
```

To?

```proto
message Value {
  message Null {}
  message Unset {}
  oneof inner {
     
   // ...

    // CQL types: blob, inet, custom
    bytes bytes = 8;

    // CQL types: uuid, timeuuid
    bytes uuid = 9; // Always 16 bytes

  }
}
```

mpenick:
**What this PR does**:

Moves `Uuid` type from using `fixed64` for the MSB/LSB elements to using the raw bytes encoding.

Also, made `Inet` its own distinct type so that it can inferred from the value without extra metadata and it is more consistent with other types.

Fixes #1248

**Checklist**
- [X] Changes manually tested


dougwettlaufer (approved):

grpc-proto/proto/query.proto
```diff
@@ -57,9 +57,11 @@ message UdtValue {
 
 // A UUID value (for CQL types uuid and timeuuid).
 message Uuid {
-  // TODO: Is this going to have endian problems representing it this way?
-  fixed64 msb = 1;
-  fixed64 lsb = 2;
+  bytes value = 1;
```
tatu-at-datastax, line 60: Does protobuf have a way to indicate byte sequence of fixed length? I should remember this but I don't -- I assume there isn't but thought to ask just in case; we'd want to verify it is 16 bytes if it was possible.

grpc-proto/proto/query.proto
```diff
@@ -57,9 +57,11 @@ message UdtValue {
...
+}
+
+message Inet {
```
tatu-at-datastax, line 63: Would it make sense to add a comment that it's 4 or 16 bytes (ipv4 vs 6)?

grpc-proto/src/main/java/io/stargate/grpc/Values.java
```diff
@@ -114,11 +118,11 @@ public static Value of(BigDecimal value) {
   }
 
   public static Value of(UUID value) {
+    ByteBuffer bytes = ByteBuffer.allocate(16);
+    bytes.putLong(0, value.getMostSignificantBits());
```
tatu-at-datastax, line 122: As per earlier comments, is there a way to verify that byte-ordering won't be problematic here? Or does JDK's `UUID` guarantee Big-endian handling.
I know `ByteBuffer`s byte order may be changed but not sure if that could be used for testing correctness since ordering with `ByteBuffer.allocate()` would default to whatever platform default is.

grpc-proto/proto/query.proto
```diff
@@ -99,36 +101,39 @@ message Value {
     // CQL types: ascii, varchar, text
     string string = 7;
 
-    // CQL types: blob, inet, custom
+    // CQL types: blob, custom
     bytes bytes = 8;
 
+    // CQL types: inet
+    Inet inet = 9;
```
tomekl007, line 108: why do we add it in the middle (instead of at the end)?

grpc-proto/src/main/java/io/stargate/grpc/Values.java
tomekl007: Can we make it explicit, instead of relying on `default`?

grpc-proto/proto/query.proto
mpenick: Good thought. Unfortunately, no. I wish it was possible though.

grpc-proto/proto/query.proto
mpenick: Good idea. thanks

grpc-proto/proto/query.proto
mpenick: We should favor types that are more likely to be used in the first "1 thorugh 15" field numbers: https://developers.google.com/protocol-buffers/docs/proto3#assigning_field_numbers

Also, it's okay to break the contract currently.

grpc-proto/src/main/java/io/stargate/grpc/Values.java
mpenick: > Can we make it explicit, instead of relying on default?

From: https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html

> The initial order of a byte buffer is always BIG_ENDIAN.

PR #1260, opened 2021-09-21
Second part to fully resolve #1229: convert grpc module too

### LINKED ISSUE #1229: Replace direct Guava usage by Stargate (use shaded everywhere applicable) to reduce Guava conflict w/ Cassandra backend

tatu-at-datastax:
**What this PR does**:

Fixes #1229

**Checklist**
- [-] Changes manually tested -- automated test suite should cover since issues would be mostly for startup (osgi)
- [-] Automated Tests added/updated -- no functional changes, existing IT/CI should be sufficient
- [-] Documentation added/updated -- no functional changes


restapi/src/main/java/io/stargate/web/docsapi/service/DocsShredder.java
```diff
@@ -1,8 +1,8 @@
 package io.stargate.web.docsapi.service;
 
+import com.datastax.oss.driver.shaded.guava.common.base.Splitter;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ValueNode;
-import com.google.common.base.Splitter;
```
ivansenic, line 5: Was Eric faster in merging the shredding than you? Can we ensure once we merge this that it's top of the master and there are no direct usages of guava?

ivansenic (commented):
Looks good so far..

restapi/src/main/java/io/stargate/web/docsapi/service/DocsShredder.java
tatu-at-datastax: I am not 100% sure how this came about. I will make sure to merge from master and check that the only Guava usage is from `persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/StargateSystemKeyspace.java` (which is due to Cassandra leaking `ListenableFuture`)

tatu-at-datastax:
Should be good, now with DSE persistence too. Waiting for CI to complete; will keep in-sync with master as well.

dougwettlaufer (approved):

tomekl007 (approved):

ivansenic (approved):
Perfect..

PR #1262, opened 2021-09-21
eliminating conversion of column names to lower case

### LINKED ISSUE #1208: Not able to retrieve row via REST with uppercase column name
Given a table created as

```sh
curl -L -X POST 'http://localhost:8082/v2/schemas/keyspaces/testks/tables' \
-H 'X-Cassandra-Token: <TOKEN>' \
--data-raw '{
    "name": "foo",
    "columnDefinitions": [
        {
            "name": "ID",
            "typeDefinition": "uuid"
        },
        {
            "name": "Firstname",
            "typeDefinition": "text"
        },
        {
            "name": "Lastname",
            "typeDefinition": "text"
        }
    ],
    "primaryKey": {
        "partitionKey": [
            "ID"
        ]
    },
    "ifNotExists": true
}'
```

I inserted a row with 

```sh
curl -L -X POST 'http://localhost:8082/v2/keyspaces/testks/foo' \
-H 'X-Cassandra-Token: <TOKEN>' \
--data-raw '{
    "ID": "af2603d2-8c03-11eb-a03f-0ada685e0000",
    "Firstname": "John",
    "Lastname": "Doe"
}'
```

But when I tried to retrieve the row it failed with the following error

```sh
curl -L -X GET 'http://localhost:8082/v2/keyspaces/testks/foo/af2603d2-8c03-11eb-a03f-0ada685e0000' \
-H 'X-Cassandra-Token: <TOKEN>'
{"description":"Bad request: Cannot find column id in table testks.foo","code":400}
```

It looks like the problem is we're always lowercasing the name here instead of preserving
https://github.com/stargate/stargate/blob/master/restapi/src/main/java/io/stargate/web/resources/Converters.java#L186

jeffreyscarpenter:
Fixes #1208
**Checklist**
- [x] Changes manually tested


restapi/src/main/java/io/stargate/web/resources/Converters.java
```diff
@@ -183,7 +183,7 @@ public static BuiltCondition idToWhere(String val, String name, Table tableData)
       value = toCqlValue(type, val);
     }
 
-    return BuiltCondition.of(name.toLowerCase(), Predicate.EQ, value);
+    return BuiltCondition.of(name, Predicate.EQ, value);
```
dougwettlaufer, line 186: Can we add some integration tests for this?

ivansenic: Furthermore, is this backward compatible or would we break somebody? Also, are there other places where `.toLowerCase()` is used?

ivansenic (approved):
+1 for the INT test, seems CI fails due to the formatting?

restapi/src/main/java/io/stargate/web/resources/Converters.java
jeffreyscarpenter: I searched and PR fixes two places where `.toLowerCase()` is used.

I added an integration test that tests creating and accessing tables with mixed case table name and column names.

PR #1268, opened 2021-09-22
Remove DEBUG log entries from "ExternalStorage" buffering (fix #1265)

### LINKED ISSUE #1265: Reduce CI test noise by filtering out DEBUG entries from output by `ExternalStorage`
Currently `ExternalStorage` buffers up all log output from the backend test cluster during Integration Test runs on CI.
This is useful but produces a lot of output (up to 35% of total logging). Of buffered (and output) logging there is one subset that seems like it should be ignored: DEBUG level entries. Since Stargate tests themselves use INFO as the base level for most things, it seems reasonable to filter out external DEBUG log entries.

From quick look it seems that:

* Total log lines from CI -- about 250k (60-70 meg size)
* Log lines via `ExternalStorage` -- about 85k
* DEBUG level log lines by `ExternalStorage` -- about 33k

So this could trim output by bit over 10%, without losing generally useful information (plus we can always tweak the logic if it appears we do want more).

tatu-at-datastax:
**What this PR does**:

PR will remove DEBUG-level entries from being propagated by `ExternalStorage`: this should reduce CI verbosity by over 10% of lines.

Fixes #1265

**Checklist**
- [x] Changes manually tested
- [ ] Automated Tests added/updated -- affects automated integration tests but cannot be meta-tested
- [ ] Documentation added/updated -- not aware of documentation about our IT/CI logging

### DIFF
--- a/persistence-test/src/main/java/io/stargate/it/storage/ExternalStorage.java
+++ b/persistence-test/src/main/java/io/stargate/it/storage/ExternalStorage.java
@@ -322,7 +322,15 @@ public class ExternalStorage extends ExternalResource<ClusterSpec, ExternalStora
             new LogOutputStream() {
               @Override
               protected void processLine(String line, int logLevel) {
-                LOG.info("storage log: {}>> {}", relPath, line);
+                //                LOG.info("storage log: {}>> {}", relPath, line);
+                // !!! TESTING ONLY:
+                boolean isDebug = line.startsWith("DEBUG: ");
+                LOG.info(
+                    "storage log: [logLevel: {}, debug? {}] {}>> {}",
+                    logLevel,
+                    isDebug,
+                    relPath,
+                    line);
               }
             }) {
           FileUtils.copyFile(file, dumper);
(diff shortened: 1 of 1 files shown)


tatu-at-datastax:
Note: committing initial test code since that is the simplest way to see how CI test run logging works. Not to be included as-is, but I was unable to figure out if `logLevel` refers to actual level constants (and if so, be useful), or is just some arbitrary number and need to use heuristics (more likely).

tatu-at-datastax:
Should be ready but will manually verify that output has changed (added summary statement at the end that should indicate number of lines included/filtered, as well as log files processed).

tatu-at-datastax:
After test runs all complete, ready for review.

ivansenic (approved):
OK not sure why we need the atomic here, but it does not matter, it's just a test class :)

mpenick:
@ivansenic  The values need to be final to be captured by `processLine` which requires some sort of reference-based integer.

mpenick (approved):

ivansenic:
> @ivansenic The values need to be final to be captured by `processLine` which requires some sort of reference-based integer.

ahhh `MutableInt` to the rescue.. :rocket:

tatu-at-datastax:
Yeah, as per @mpenick it was just to make counters mutable from within :)

Issue #1235, opened 2021-09-09
Refactor `DocumentDB`


ivansenic:
The `DocumentDB` class currently mixes up few concerns that should be split, thus making the class more readable and better testable. Tasks:

* no docs API constant should be hard-coded, rename `QueryConstants` to `DocsApiConstants` and replace all `text_value`, `dbl_value`, `bool_value`, & `leaf` to constant access (across the code base)
* move other constants to the `DocsApiConstants` (`GLOB_VALUE`, `GLOB_ARRAY_VALUE`, `ROOT_DOC_MARKER`, etc)
* extract database structure to a Immutable class with cached methods (move `allColumns`, `allColumnNames`and other static values to the methods in this immutable class) + add tests for that class
* move all properties to a `DocsApiProperties` or `DocsApiConfiguration`

ivansenic:
Done with moving to V2.

PR #1259, opened 2021-09-21
shredJson() refactor and testing

EricBorczuk:
**What this PR does**:
Refactors shredJson() into more logical pieces; marks `shredForm` as deprecated (it should be deleted after we move away from YCSB testing and it is not documented in the API), and added testing for `shredJson`.

**Which issue(s) this PR fixes**:

**Checklist**
- [x] Changes manually tested
- [x] Automated Tests added/updated

PR #1270, opened 2021-09-23
Separate constants from DocumentDB into DocsApiConstants

EricBorczuk:
**What this PR does**:
This PR refactors a bunch of the constants from DocumentDB into a combination of DocsApiConstants (Renamed from QueryConstants) and DocsApiConfiguration. As a result of this, a few things have changed:
1) In a bunch of places, DocsApiConfiguration or its methods gets used instead of DocumentDB
2) `allColumns` and its derivatives were consolidated into one place. Before, there were two separate places for generating the columns for writing and reading, not anymore - but this also means that the column order expectation was changed for reading (in other words, writes require a certain order of columns for binding params, while selecting didn't, but now they are both the same)
3) newBindMap is no longer a part of DocumentDB

The only thing missing from this PR w.r.t the issue description is an Immutable class that caches calls to `allColumns` and its sibling functions; I could add that in this PR but it was getting kind of large and I think it's reasonable to review in this state.

Fixes https://github.com/stargate/stargate/issues/1235

**Checklist**
- [x] Changes manually tested


ivansenic:
@EricBorczuk The changes are rather large and include changes of the signatures in the search query builder, can we discuss first what s being done and why, it would be easier for me to review this monster?

ivansenic (changes requested):
See my comment above..

tatu-at-datastax:
It would be great to get this (and the other related refactoring) merged -- of course considering open issues etc -- since V2 prototype work will be focused on `restapi` module, and unfortunately Docs API is bundled with it.
Current thinking is that v2 work will happen in a branch, requiring steady merging from `master`. That's fine for fixes, feature changes, but can sometimes be more hassle with bigger refactorings.

EricBorczuk:
@tatu-at-datastax Agreed - for this PR, I just incorporated the requested changes; I believe tests will pass in its current state. Its base https://github.com/stargate/stargate/pull/1259 is actually ready to merge after I get an approval. So I think I can get these out in the next day or so!

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -314,9 +255,13 @@ public void tableExists(String keyspaceName, String tableName) {
     }
   }
 
+  private List<String> allColumnNames() {
+    return Arrays.asList(DocsApiConstants.ALL_COLUMNS_NAMES.apply(config.getMaxDepth()));
```
ivansenic, line 259: I was thinking that such things we can extract into immutable class as I specified in the #1235. Since, this is too big maybe we can keep that issue open and just do that on top of this PR. Wdyt?

restapi/src/main/java/io/stargate/web/docsapi/service/DocsApiConfiguration.java
```diff
@@ -1,41 +1,42 @@
...
 public interface DocsApiConfiguration {
 
   /** Default config instance */
   DocsApiConfiguration DEFAULT = new DocsApiConfiguration() {};
 
   /** @return Returns max allowed document page size that user can request. */
-  default int getMaxPageSize() {
-    return DocumentDB.MAX_PAGE_SIZE;
-  }
+  int MAX_PAGE_SIZE = 20;
```
ivansenic, line 10: why was this moved to static field instead of the function? the idea was that with the function you can override it in tests with a mocked value. This way we can not?

I would also be fine if you move this to the constants class?

restapi/src/main/java/io/stargate/web/docsapi/service/DocsApiConfiguration.java
```diff
@@ -1,41 +1,42 @@
...
-    return Math.min(numberOfDocuments * 16, DocumentDB.MAX_STORAGE_PAGE_SIZE);
+    return Math.min(numberOfDocuments * 16, getMaxStoragePageSize());
   }
 
   /**
    * @return The maximum storage page size that should be allowed when doing any doc API queries.
    */
   default int getMaxStoragePageSize() {
-    return DocumentDB.MAX_STORAGE_PAGE_SIZE;
+    return Integer.getInteger("stargate.document_search_page_size", 1000);
```
ivansenic, line 25: should we extract those to the fields and return the field here? this way we are always parsing the strings, but sure why would we have that overhead?

Also fine to move to constants or some properties class

restapi/src/main/java/io/stargate/web/docsapi/resources/ReactiveDocumentResourceV2.java
```diff
@@ -297,16 +297,11 @@ public void searchDoc(
...
-                      + DocumentDB.MAX_PAGE_SIZE
-                      + ", default 3",
+              value = "the max number of documents to return, max 20, default 3",
               defaultValue = "3")
           @QueryParam("page-size")
           @Min(value = 1, message = "the minimum number of documents to return is one")
-          @Max(
-              value = DocumentDB.MAX_PAGE_SIZE,
-              message = "the max number of documents to return is " + DocumentDB.MAX_PAGE_SIZE)
+          @Max(value = 20, message = "the max number of documents to return is 20")
```
ivansenic, line 304: I would still keep the static fields in the constants and use for this purpose.

restapi/src/main/java/io/stargate/web/docsapi/resources/CollectionsResource.java
```diff
@@ -303,7 +304,8 @@ public Response upgradeCollection(
                   new DocumentDB(
                       authenticatedDB.getDataStore(),
                       authenticatedDB.getAuthenticationSubject(),
-                      db.getAuthorizationService()),
+                      db.getAuthorizationService(),
+                      DocsApiConfiguration.DEFAULT),
```
ivansenic, line 308: inject instance. not default please

restapi/src/main/java/io/stargate/web/docsapi/exception/ErrorCode.java
```diff
@@ -57,7 +57,7 @@
       Response.Status.BAD_REQUEST,
       String.format(
           "The parameter `page-size` is limited to %d.",
-          DocsApiConfiguration.DEFAULT.getMaxPageSize())),
+          DocsApiConfiguration.DEFAULT.MAX_PAGE_SIZE)),
```
ivansenic, line 60: revert to method?

restapi/src/main/java/io/stargate/web/docsapi/service/query/DocsApiConstants.java
```diff
@@ -0,0 +1,104 @@
...
+package io.stargate.web.docsapi.service.query;
+
+import io.stargate.db.schema.Column;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+/** Constants needed in the query. */
+public interface DocsApiConstants {
```
ivansenic, line 26: OK now we need to add Java doc here for everything so that readers can understand what each value or function does!

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
```diff
@@ -55,13 +57,13 @@ private DocsApiUtils() {}
    * @param path single filter or field path
    * @return Converted to represent expected path in DB, keeping the segmentation.
    */
-  public static String convertArrayPath(String path) {
+  public static String convertArrayPath(String path, int maxArrayLength) {
```
ivansenic, line 60: do you think we should provide alternatives to the extended methods here that take the configuration as parameter?

```
convertArrayPath(String path, DocsApiConfiguraiton config) {
  return convertArrayPath(path, config.get..);
}
```

restapi/src/main/java/io/stargate/web/resources/Db.java
```diff
@@ -92,7 +93,8 @@ private DocumentDB getDocDataStoreForTokenInternal(TokenAndHeaders tokenAndHeade
     return new DocumentDB(
         authenticatedDB.getDataStore(),
         authenticatedDB.getAuthenticationSubject(),
-        getAuthorizationService());
+        getAuthorizationService(),
+        DocsApiConfiguration.DEFAULT);
```
ivansenic, line 97: inject, not static

ivansenic (changes requested):
This look really good now.. Tests are failing so I set this to request changes.. I was lazy to look to the tests atm as they are not green, but I assume tests will not be a problem.. 

I left few comments as I was going through, I would like to know what do you think about that as well..

ivansenic:
> It would be great to get this (and the other related refactoring) merged -- of course considering open issues etc -- since V2 prototype work will be focused on `restapi` module, and unfortunately Docs API is bundled with it. Current thinking is that v2 work will happen in a branch, requiring steady merging from `master`. That's fine for fixes, feature changes, but can sometimes be more hassle with bigger refactorings.

@tatu-at-datastax are you sure V2 will tackle both rest and docs, I know they in the same module now, but I guess they will be completely separated in V2?

tatu-at-datastax:
@ivansenic We don't fully know details of V2 here wrt whether one service would handle both REST and Document API but ideally I think they would be separate. But even if they were left co-located I think separating them within module makes sense -- some parts can remain shared, as long as they are cleanly separate so that if we wanted to we could create separate "restdoc-common" package, if that makes sense.

It may get bit tricky with constants, but I would probably prefer RestApiConstants vs DocumentApiConstants for everything that can be separated.

restapi/src/test/java/io/stargate/web/docsapi/dao/DocumentDBTest.java
```diff
@@ -12,7 +12,7 @@
 import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.mock;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.ImmutableList;
```
ivansenic, line 15: not quite sure how this works now, but I think @tatu-at-datastax removed all direct guava usages and just depended on the shadowed version.. Would you be up using `Arrays.asList()` or `Collections.singeltonList()` just to avoid again the same old problem?

restapi/src/test/java/io/stargate/web/docsapi/service/QueryExecutorTest.java
```diff
@@ -50,6 +50,7 @@
 import org.junit.jupiter.params.provider.CsvSource;
 
 class QueryExecutorTest extends AbstractDataStoreTest {
+  private DocsApiConfiguration config = DocsApiConfiguration.DEFAULT;
```
ivansenic, line 53: For proper unit tests, we should use the mocked version, mock depending methods (I think this test has none) and then verify component is using values the mocked values..

restapi/src/test/java/io/stargate/web/docsapi/service/ReactiveDocumentServiceTest.java
```diff
@@ -85,6 +85,8 @@
 
   ObjectMapper objectMapper = new ObjectMapper();
 
+  DocsApiConfiguration config = DocsApiConfiguration.DEFAULT;
```
ivansenic, line 88: Mocked version please..

restapi/src/test/java/io/stargate/web/docsapi/service/query/ExpressionParserIntTest.java
```diff
@@ -59,9 +60,11 @@
 
   ObjectMapper mapper = new ObjectMapper();
 
+  DocsApiConfiguration configuration = DocsApiConfiguration.DEFAULT;
```
ivansenic, line 63: mocked version please, verify proper values if needed (not sure)

restapi/src/test/java/io/stargate/web/docsapi/service/query/search/db/impl/DocumentsSearchQueryBuilderTest.java
```diff
@@ -45,6 +46,7 @@
   private static final DocsApiTestSchemaProvider SCHEMA_PROVIDER = new DocsApiTestSchemaProvider(4);
   private static final String KEYSPACE_NAME = SCHEMA_PROVIDER.getKeyspace().name();
   private static final String COLLECTION_NAME = SCHEMA_PROVIDER.getTable().name();
+  private static final DocsApiConfiguration config = DocsApiConfiguration.DEFAULT;
```
ivansenic, line 49: mocked version, as we have in the other SearchQueryBuilderTests

restapi/src/test/java/io/stargate/web/docsapi/service/query/search/db/impl/FilterExpressionSearchQueryBuilderTest.java
```diff
@@ -48,6 +49,7 @@
   private static final DocsApiTestSchemaProvider SCHEMA_PROVIDER = new DocsApiTestSchemaProvider(4);
   private static final String KEYSPACE_NAME = SCHEMA_PROVIDER.getKeyspace().name();
   private static final String COLLECTION_NAME = SCHEMA_PROVIDER.getTable().name();
+  private static final DocsApiConfiguration config = DocsApiConfiguration.DEFAULT;
```
ivansenic, line 52: mocker version

restapi/src/test/java/io/stargate/web/docsapi/service/query/search/db/impl/FilterPathSearchQueryBuilderTest.java
```diff
@@ -47,6 +47,7 @@
   private static final DocsApiTestSchemaProvider SCHEMA_PROVIDER = new DocsApiTestSchemaProvider(4);
   private static final String KEYSPACE_NAME = SCHEMA_PROVIDER.getKeyspace().name();
   private static final String COLLECTION_NAME = SCHEMA_PROVIDER.getTable().name();
+  private static final DocsApiConfiguration config = DocsApiConfiguration.DEFAULT;
```
ivansenic, line 50: mocked version

restapi/src/test/java/io/stargate/web/docsapi/service/query/search/db/impl/FullSearchQueryBuilderTest.java
```diff
@@ -32,6 +33,7 @@
   private static final DocsApiTestSchemaProvider SCHEMA_PROVIDER = new DocsApiTestSchemaProvider(0);
   private static final String KEYSPACE_NAME = SCHEMA_PROVIDER.getKeyspace().name();
   private static final String COLLECTION_NAME = SCHEMA_PROVIDER.getTable().name();
+  private static final DocsApiConfiguration config = DocsApiConfiguration.DEFAULT;
```
ivansenic, line 36: not used?

restapi/src/test/java/io/stargate/web/docsapi/service/query/search/db/impl/PathSearchQueryBuilderTest.java
```diff
@@ -38,6 +39,7 @@
   private static final DocsApiTestSchemaProvider SCHEMA_PROVIDER = new DocsApiTestSchemaProvider(4);
   private static final String KEYSPACE_NAME = SCHEMA_PROVIDER.getKeyspace().name();
   private static final String COLLECTION_NAME = SCHEMA_PROVIDER.getTable().name();
+  private static final DocsApiConfiguration config = DocsApiConfiguration.DEFAULT;
```
ivansenic, line 42: not used?

restapi/src/test/java/io/stargate/web/docsapi/service/query/search/db/impl/PopulateSearchQueryBuilderTest.java
```diff
@@ -32,6 +33,7 @@
   private static final DocsApiTestSchemaProvider SCHEMA_PROVIDER = new DocsApiTestSchemaProvider(0);
   private static final String KEYSPACE_NAME = SCHEMA_PROVIDER.getKeyspace().name();
   private static final String COLLECTION_NAME = SCHEMA_PROVIDER.getTable().name();
+  private static final DocsApiConfiguration config = DocsApiConfiguration.DEFAULT;
```
ivansenic, line 36: not used?

restapi/src/test/java/io/stargate/web/docsapi/service/query/search/db/impl/SubDocumentSearchQueryBuilderTest.java
```diff
@@ -37,6 +38,7 @@
   private static final DocsApiTestSchemaProvider SCHEMA_PROVIDER = new DocsApiTestSchemaProvider(4);
   private static final String KEYSPACE_NAME = SCHEMA_PROVIDER.getKeyspace().name();
   private static final String COLLECTION_NAME = SCHEMA_PROVIDER.getTable().name();
+  private static final DocsApiConfiguration config = DocsApiConfiguration.DEFAULT;
```
ivansenic, line 41: not used?

restapi/src/test/java/io/stargate/web/docsapi/service/query/search/resolver/BaseResolverTest.java
```diff
@@ -55,14 +56,16 @@
 
   @Mock DocumentsResolver candidatesResolver;
 
+  @Mock DocsApiConfiguration configuration;
```
ivansenic, line 59: :rocket: :guitar:

restapi/src/test/java/io/stargate/web/docsapi/service/query/search/resolver/filter/impl/InMemoryCandidatesFilterTest.java
```diff
@@ -120,12 +123,13 @@ public void fixedPath() throws Exception {
 
       withQuery(
           TABLE,
-          "SELECT key, leaf, text_value, dbl_value, bool_value, p0, p1, WRITETIME(leaf) FROM %s WHERE p0 = ? AND leaf = ? AND p1 = ? AND key = ? LIMIT ? ALLOW FILTERING");
+          "SELECT key, p0, p1, leaf, text_value, dbl_value, bool_value, WRITETIME(leaf) FROM %s WHERE p0 = ? AND leaf = ? AND p1 = ? AND key = ? LIMIT ? ALLOW FILTERING");
```
ivansenic, line 126: why is this changed now? any ideas?

restapi/src/test/java/io/stargate/web/docsapi/service/util/DocsApiUtilsTest.java
```diff
@@ -56,42 +56,42 @@
     public void happyPath() {
       int index = RandomUtils.nextInt(100, 999);
 
-      String result = DocsApiUtils.convertArrayPath(String.format("[%d]", index));
+      String result = DocsApiUtils.convertArrayPath(String.format("[%d]", index), 999999);
```
ivansenic, line 59: can you just add one test that simulates array index higher that given max? and verify correct error being returned?

ivansenic (approved):
Approving as it's awesome.. Some minor testing fixes and additions needed.. Also rebase need due to the #1283

restapi/src/test/java/io/stargate/web/docsapi/service/query/search/resolver/filter/impl/InMemoryCandidatesFilterTest.java
EricBorczuk: This is changed because now the search path and the write path use the same function for generating column names. The search path historically didn't care about the order of these columns, but for parameter binding purposes, the write side expects the columns to always be in the order `key, p0, ...pN, leaf, text_value, dbl_value, bool_value`

ivansenic:
@EricBorczuk did you address the comments above before merging? if not I can create a quick followup?

PR #1272, opened 2021-09-23
(WIP) Fix #1271: remove inclusion of stack trace for logging

### LINKED ISSUE #1271: Remove stack trace from logging of "known exceptions" in `RequestHandler`, `AuthnJwtService`
There are a few cases in which some of APIs both log an exception and throw it: this is usually frowned upon, but may make sense for gathering some information. In such cases it is, however, usually best not to always include stack trace in logging -- especially when exception type and message indicate actual reason.
This is the case for a few Authnz handlers, as well as code that resolves C* resources such as keyspaces and tables.

I'll do a PR that will remove places where stack trace looks unnecessary, based on existing CI log inspection. This should further improve readability of the log output by removing more fluff.

tatu-at-datastax:
**What this PR does**:

Main thing is to remove inclusion of stack trace for selected cases of logging, where trace adds no valuable information (where we know from context, exception type and message all pertitent info).

Also removes one extra Maven dependency that mvn build complained about.

Fixes #1271

**Checklist**
- [x] Changes manually tested -- check out CI logs to verify (compared to earlier)

### DIFF
--- a/persistence-test/src/main/java/io/stargate/it/storage/ExternalStorage.java
+++ b/persistence-test/src/main/java/io/stargate/it/storage/ExternalStorage.java
@@ -148,7 +148,7 @@ public class ExternalStorage extends ExternalResource<ClusterSpec, ExternalStora
 
     private final UUID id = UUID.randomUUID();
     private final ClusterSpec spec;
-    private final AtomicBoolean errorInTest = new AtomicBoolean();
+    private final AtomicInteger errorsInTest = new AtomicInteger(0);
 
     protected Cluster(ClusterSpec spec) {
       this.spec = spec;
@@ -164,11 +164,11 @@ public class ExternalStorage extends ExternalResource<ClusterSpec, ExternalStora
     public abstract String infoForTestLog();
 
     private void markError() {
-      errorInTest.set(true);
+      errorsInTest.incrementAndGet();
     }
 
-    public boolean errorsDetected() {
-      return errorInTest.get();
+    public int errorsDetected() {
+      return errorsInTest.get();
     }
   }
 
@@ -293,7 +293,7 @@ public class ExternalStorage extends ExternalResource<ClusterSpec, ExternalStora
           }
         } catch (Exception e) {
           // This should not affect test result validity, hence logging as WARN
-          LOG.warn("Exception during CCM cluster shutdown: {}", e, e);
+          LOG.warn("Exception during CCM cluster shutdown: " + e.getMessage(), e);
         }
       }
     }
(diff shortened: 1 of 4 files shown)


restapi/src/main/java/io/stargate/web/resources/RequestHandler.java
```diff
@@ -38,29 +38,31 @@ public static Response handle(Callable<Response> action) {
...
-      logger.info("Bad request", iae);
+      logger.info("Bad request (IllegalArgumentException->BAD_REQUEST): {}", iae.getMessage());
       return Response.status(Response.Status.BAD_REQUEST)
           .entity(
               new Error(
                   "Bad request: " + iae.getMessage(), Response.Status.BAD_REQUEST.getStatusCode()))
           .build();
     } catch (InvalidRequestException ire) {
-      logger.info("Bad request", ire);
+      logger.info("Bad request (InvalidRequestException->BAD_REQUEST): {}", ire.getMessage());
```
dougwettlaufer, line 56: Isn’t the `BAD_REQUEST` redundant?

dougwettlaufer (approved):
Minor nitpick on log message but looks good otherwise

restapi/src/main/java/io/stargate/web/resources/RequestHandler.java
tatu-at-datastax: It is; I am not sure what'd be most meaningful combination here. Intent is/was to indicate Exception gotten, HTTP response given, and do that consistently. So I am definitely open to a better scheme.

persistence-test/src/main/java/io/stargate/it/storage/ExternalStorage.java
```diff
@@ -293,7 +293,7 @@ public void stop() {
           }
         } catch (Exception e) {
           // This should not affect test result validity, hence logging as WARN
-          LOG.warn("Exception during CCM cluster shutdown: {}", e, e);
+          LOG.warn("Exception during CCM cluster shutdown: " + e.getMessage(), e);
```
tatu-at-datastax, line 296: Would have failed to do what is expected: overload was taking both `e`s as substitutions (but intent was to include stack trace).

restapi/src/main/java/io/stargate/web/resources/RequestHandler.java
tatu-at-datastax: Will merge as-is but can easily improve as a follow-up if we can figure out better scheme here.

PR #1275, opened 2021-09-24
Fix #1273: add explicit handling of Assume-sourced exceptions for Integ tests

### LINKED ISSUE #1273: Add handling for skipped tests (via "assume" vs assert) for `ExternalStorage` to avoid false errors
Helper class `ExternalStorage` of "persistence-test" dumps out sizable external logs in cases it thinks there are failing tests. Unfortunately this currently includes skipped (NOT failing) tests like cases where Cassandra-4.0 backend has to skip a test with log such as:

```
Step #4 - "cassandra-4.0": org.opentest4j.TestAbortedException: assumption was not met due to: [Disabled because it is currently not possible to enable SAI indexes on a Cassandra 4 backend] 
```

This in turn results in:

```
Step #4 - "cassandra-4.0": WARN  [main] 2021-09-23 20:36:41,974 ExternalStorage.java:313 - Dumping storage cluster (Cassandra 4.0) logs due to 1 test failures.
```

even though actually no test failed at all. This can be avoided by changes in method `handleTestExecutionException()` of `ExternalStorage` to consider two special exceptions:

* `org.junit.AssumptionViolatedException` 
* `org.opentest4j.TestAbortedException`

as indicated skipped, not failing test.

tatu-at-datastax:
**What this PR does**:

Adds explicit handling of 2 kinds of exceptions junit throws to implement "Assume" functionality (ability to dynamically skip tests) -- here used to ignore non-applicable tests for some persistence backends.
Before change these are considered test failures wrt dumping out `ExternalStorage` buffered logs; after change they are not considered test failures any more. Although there are only 12 of such cases for C*4.0 backend, they are responsible for probably 30,000 more log lines.

Fixes #1273

**Checklist**
- [x] Changes manually tested - will verify changes to logging
- [ ] Automated Tests added/updated - yes, tests were changed (although not in the sense asked here maybe)
- [ ] Documentation added/updated - no documentation about ITs.

### DIFF
--- a/persistence-test/src/main/java/io/stargate/it/storage/ExternalStorage.java
+++ b/persistence-test/src/main/java/io/stargate/it/storage/ExternalStorage.java
@@ -47,6 +47,7 @@ import org.junit.jupiter.api.extension.ParameterContext;
 import org.junit.jupiter.api.extension.ParameterResolutionException;
 import org.junit.jupiter.api.extension.ParameterResolver;
 import org.junit.jupiter.api.extension.TestExecutionExceptionHandler;
+import org.opentest4j.TestAbortedException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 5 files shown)


tatu-at-datastax:
Seems to work as expected; numbers:

Before changes, C*4.0 integration tests:

* 50.9 MB
* 159,000 lines

after changes

* 31.4 MB
* 108,000 lines

and now we have log lines like:

```
Step #4 - "cassandra-4.0": INFO  [main] 2021-09-24 18:37:46,590 ExternalStorage.java:332 - Skipping dumping storage cluster (Cassandra 4.0) logs: no test failures (7 tests skipped).
```

to indicate cases that would formerly have dumped boatloads of external logs due to assumed failures.

persistence-test/src/main/java/io/stargate/it/storage/ExternalStorage.java
```diff
@@ -138,8 +139,21 @@ public void beforeTestExecution(ExtensionContext context) {
   @Override
   public void handleTestExecutionException(ExtensionContext context, Throwable throwable)
       throws Throwable {
-    LOG.warn("Error in {}", context.getUniqueId(), throwable);
-    getResource(context).ifPresent(Cluster::markError);
+    // 24-Sep-2021, tatu: When using "assume()" methods for dynamic disabling/skipping
+    //    of tests, we still get an exception -- but those are not fails, need to filter out
+    //    so that we do not unnecessarily dump output logs
+    if ((throwable instanceof TestAbortedException) // opentest4j
+        || "AssumptionViolatedException".equals(throwable.getClass().getSimpleName()) // junit
```
tatu-at-datastax, line 146: This may look odd but module does not have dependency on `junit`, and I am bit worried about OSGi providing different version of the class. Hence chose to check the name, looks explicit enough and works.

dougwettlaufer (approved):

tomekl007 (approved):

PR #1285, opened 2021-09-28
Fix #1284: rename v1 endpoint classes, Server->RestApiServer

### LINKED ISSUE #1284: Refactor `restapi` to separate legacy REST v1 endpoints from v2 ones
Minor cleanup: currently REST API `v2` endpoints are under separate package but v1 ones are under `resources`.
Would be cleaner to have separation; change is not visible to users but can be confusing to new developers.

tatu-at-datastax:
**What this PR does**:

Minor rename patch to improve readability of `restapi` (which contains REST and Documents APIs):

* Move `v1` resources under specific `v1/` package similar to `v2`
* Rename generic `Server` as `RestApiServer`; `WebImpl` as `RestApiRunner`

Fixes #1284

**Checklist**
- [x] Changes manually tested -- verified by CI that things still work (no functional changes)
- [x] Documentation added/updated -- one reference to renamed `RestApiServer`

### DIFF
--- a/restapi/src/main/java/io/stargate/web/RestApiActivator.java
+++ b/restapi/src/main/java/io/stargate/web/RestApiActivator.java
@@ -21,7 +21,7 @@ import io.stargate.core.activator.BaseActivator;
 import io.stargate.core.metrics.api.HttpMetricsTagProvider;
 import io.stargate.core.metrics.api.Metrics;
 import io.stargate.db.datastore.DataStoreFactory;
-import io.stargate.web.impl.WebImpl;
+import io.stargate.web.impl.RestApiRunner;
 import java.util.Arrays;
 import java.util.List;
 import org.slf4j.Logger;
@@ -32,7 +32,7 @@ public class RestApiActivator extends BaseActivator {
 
   public static final String MODULE_NAME = "restapi";
   private static final Logger logger = LoggerFactory.getLogger(RestApiActivator.class);
-  private final WebImpl web = new WebImpl();
+  private final RestApiRunner runner = new RestApiRunner();
   private final ServicePointer<AuthenticationService> authenticationService =
       ServicePointer.create(
           AuthenticationService.class,
@@ -53,16 +53,17 @@ public class RestApiActivator extends BaseActivator {
 
   @Override
   protected ServiceAndProperties createService() {
-    web.setAuthenticationService(authenticationService.get());
-    web.setMetrics(metrics.get());
-    web.setHttpMetricsTagProvider(httpTagProvider.get());
-    web.setAuthorizationService(authorizationService.get());
-    web.setDataStoreFactory(dataStoreFactory.get());
+    runner.setAuthenticationService(authenticationService.get());
+    runner.setMetrics(metrics.get());
+    runner.setHttpMetricsTagProvider(httpTagProvider.get());
+    runner.setAuthorizationService(authorizationService.get());
+    runner.setDataStoreFactory(dataStoreFactory.get());
     try {
-      this.web.start();
+      runner.start();
     } catch (Exception e) {
-      logger.error("Failed", e);
+      logger.error("Running RestApiRunner Failed", e);
     }
+    // Shouldn't we return something to avoid add logging for "service null"?
     return null;
   }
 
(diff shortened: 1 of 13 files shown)


restapi/src/main/java/io/stargate/web/impl/RestApiRunner.java
```diff
@@ -21,18 +21,14 @@
...
-public class WebImpl {
+public class RestApiRunner {
 
   private AuthenticationService authenticationService;
   private AuthorizationService authorizationService;
   private Metrics metrics;
   private HttpMetricsTagProvider httpMetricsTagProvider;
   private DataStoreFactory dataStoreFactory;
 
-  public AuthenticationService getAuthenticationService() {
```
tatu-at-datastax, line 32: As far as I know, no getters are needed. Could also just remove setters as well, pass via constructor.

restapi/src/main/java/io/stargate/web/impl/RestApiServer.java
```diff
@@ -136,16 +137,22 @@ protected void configure() {
                 bind(environment.getObjectMapper()).to(ObjectMapper.class);
               }
             });
+
```
tatu-at-datastax, line 140: Just reordered endpoints by grouping, then alphabetic, no changes to what is registered.

restapi/src/main/java/io/stargate/web/impl/RestApiServer.java
```diff
@@ -156,6 +163,7 @@ protected void configure() {
     environment.jersey().register(CollectionsResource.class);
     environment.jersey().register(NamespacesResource.class);
 
+    // Swagger endpoints
     environment.jersey().register(ApiListingResource.class);
     environment.jersey().register(SwaggerSerializers.class);
```
tatu-at-datastax, line 168: This could very likely be a singleton too, if it matters.

restapi/src/main/java/io/stargate/web/swagger/SwaggerUIResource.java
```diff
@@ -90,7 +92,7 @@ public Response get(
       token =
           Strings.isNullOrEmpty(bearerToken)
               ? bearerToken
-              : bearerToken.replaceFirst("^Bearer\\s", "");
+              : bearerTokenPattern.matcher(bearerToken).replaceFirst("");
```
tatu-at-datastax, line 95: Minor performance optimization since it's true regexp and seems like it might get called quite regularly.

restapi/src/main/java/io/stargate/web/RestApiActivator.java
```diff
@@ -53,16 +53,17 @@ public RestApiActivator() {
...
+    runner.setAuthorizationService(authorizationService.get());
+    runner.setDataStoreFactory(dataStoreFactory.get());
     try {
-      this.web.start();
+      runner.start();
     } catch (Exception e) {
-      logger.error("Failed", e);
+      logger.error("Running RestApiRunner Failed", e);
     }
+    // Shouldn't we return something to avoid add logging for "service null"?
```
ivansenic, line 66: this looks odd, but it is specified in the java docs.. It's even more confusing that `BaseActivator` does not check for returned `null` here and creates a list with a null value:

```
  protected List<ServiceAndProperties> createServices() {
    return Collections.singletonList(createService());
  }
```

So I would surely fix that `BaseActivator` stuff and then you can also just rename this method to createServices() and return empty list if you feel this is cleaner.

restapi/src/main/java/io/stargate/web/impl/RestApiRunner.java
ivansenic: Or even go for the immutable class..

restapi/src/main/java/io/stargate/web/impl/RestApiServer.java
ivansenic: Yep, I will do this on top. Can you move Swagger resource to this group if you are grouping already?

ivansenic (approved):
I love refactoring..

restapi/src/main/java/io/stargate/web/RestApiActivator.java
tatu-at-datastax: Ok. I'll keep this in mind and return to it later since I suspect same pattern used elsewhere.

tatu-at-datastax:
Yes, `RestApiRunner` could def become immutable altogether. I think I should probably make a pass across services to see if same applies elsewhere.

PR #1288, opened 2021-09-30
Fix #1287: clean up division between `DocumentDB` and `AuthenticatedDB`

### LINKED ISSUE #1287: Refactoring of REST/Doc Api: clean separation of `AuthenticatedDB` vs `DocumentDB`
Currently `restapi` module hosts bth REST an Document APIs: however, code is partly split so that newer Document API resides in package `io.stargate.docsapi`.
Access to underlying `DataStore` is mostly split so that:

* `Db` is used to create actual auth'ed instances: `AuthenticatedDB`(not optimal name)  for REST access and `DocumentDB` for DocumentAPI`

There is, however, some residual usage of `AuthenticatedDB` by Documents API: this seems easy enough to solve.

Going forward it will probably make sense to consider renaming `Db` as `DbProvider` (or `DbFactory`); and maybe even split into REST/Document parts: eventual goal being to split the module in 2 separate services similar to how graphql, cql and grpc APIs work.

But first things first: let's clean up existing split.

tatu-at-datastax:
**What this PR does**:

As per linked issue, the goal here is to improve separation between logically distinct REST and Document APIs.
There is also smaller clean up wrt abstractions themselves, construction of DocumentDB, AuthenticatedDB by `Db` class.

Fixes #1287

**Checklist**
- [x] Changes manually tested -- Should not change functionality, relying on CI (plus local sanity check)

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
@@ -133,6 +133,10 @@ public class DocumentDB {
     return authenticationSubject;
   }
 
+  public boolean supportsSAI() {
+    return dataStore.supportsSAI();
+  }
+
   public boolean treatBooleansAsNumeric() {
     return !dataStore.supportsSecondaryIndex();
   }
@@ -141,7 +145,7 @@ public class DocumentDB {
     return allColumns;
   }
 
-  public QueryBuilder builder() {
+  public QueryBuilder queryBuilder() {
     return dataStore.queryBuilder();
   }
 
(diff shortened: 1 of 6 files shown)


ivansenic (approved):
:+1:

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -141,16 +145,29 @@ public boolean treatBooleansAsNumeric() {
...
+  public QueryBuilder queryBuilder() {
     return dataStore.queryBuilder();
   }
 
   public Schema schema() {
     return dataStore.schema();
   }
 
+  public Table getTable(String keyspaceName, String table) {
+    Keyspace keyspace = getKeyspace(keyspaceName);
```
tomekl007, line 157: why not `dataStore.getTable()`?

tomekl007 (approved):

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
tatu-at-datastax: I can do that too; just copied code. Will change, no real reason (and some consider it bad practice for public API methods to call other public API methods).

tatu-at-datastax: Oh. Actually, no; `DataStore` has just accessor for `schema`; and `schema` for `Keyspace`.
So there's no convenience method. I could add method there if that makes sense but I think for this PR I'll leave it as-is.

PR #1292, opened 2021-10-01
Adapt gRPC example to work with Astra

tomekl007:
**What this PR does**:
Adapts the `GrpcClientExecuteQueryAstra` to allow user to interact with the gRPC API on the astra cluster (not only local `Staragate`)

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/grpc-examples/src/main/java/io/stargate/grpcexample/GrpcClientExecuteQuery.java
+++ b/grpc-examples/src/main/java/io/stargate/grpcexample/GrpcClientExecuteQuery.java
@@ -52,7 +52,7 @@ public class GrpcClientExecuteQuery {
   //  "{$ASTRA_DB_ID}-{ASTRA_DB_REGION}.apps.astra.datastax.com";
 
   // change to false when running on Astra
-  private static final boolean LOCAL_RUN = false;
+  private static final boolean LOCAL_RUN = true;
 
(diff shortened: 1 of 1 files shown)


ivansenic (approved):

dougwettlaufer (changes requested):
We don't need something Astra specific here. As discussed offline let's handle this somewhere else.

tomekl007:
> We don't need something Astra specific here. As discussed offline let's handle this somewhere else.

Could you elaborate more? Where is the `somewhere else`?

grpc-examples/src/main/java/io/stargate/grpcexample/GrpcClientExecuteQuery.java
```diff
@@ -26,15 +26,39 @@
...
+ * </ol>
  */
 public class GrpcClientExecuteQuery {
 
   private final StargateGrpc.StargateBlockingStub blockingStub;
   private final StargateGrpc.StargateStub asyncStub;
 
   private static final String STARGATE_TOKEN = ""; // you need to set the token
-  private static final String STARGATE_HOST = "localhost";
+  private static final String STARGATE_HOST = "localhost";;
```
tatu-at-datastax, line 49: Super tiny but there's an extra semicolon :)

Issue #1290, opened 2021-09-30
`json-schema` Endpoint: Single Quotes in the "description" field of a JSON Schema property cause errors.


kidrecursive:
When creating a JSON schema that looks like this: 
```json
{

  "$id": "https://example.com/person.schema.json",
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "title": "Person",
  "type": "object",
  "properties": {
    "firstName": {
      "type": "string",
      "description": "The person's first name."
    },
    "lastName": {
      "type": "string",
      "description": "The person's last name."
    },
    "age": {
      "description": "Age in years which must be equal to or greater than zero.",
      "type": "integer",
      "minimum": 0
    }
  }
}
```
The API returns a 500. When escaping the single quote, the API returns a 400. The request was issued via the Swagger UI.

dougwettlaufer:
@kidrecursive do you have the response message from the 500 and 400?

kidrecursive:
Sure thing:

400 
```json
{
  "description": "Malformed JSON schema provided.",
  "code": 400
}
```

500
```
{
  "description": "Server error: org.apache.cassandra.stargate.exceptions.SyntaxException: line 1:262 mismatched input 's' expecting EOF (....crw_test WITH comment = '{\"schema\":{\"$id\":\"https://example.com/person.schema.json\",\"$schema\":\"https://json-schema.org/draft/2020-12/schema\",\"title\":\"Person\",\"type\":\"object\",\"properties\":{\"firstName\":{\"type\":\"string\",\"description\":\"The person'[s]...)",
  "code": 500
}
```

tatu-at-datastax:
Sounds like there might be two issues: first, failure should be 4xx not 500 (for the first case) if there's an issue with user-provided input. And second, something about escaping at CQL level may be off (error does not come from JSON parser I think, but from underlying CQL statement).

ivansenic:
This is a Swagger issue, just tested with your schema and command like:

```
curl -X PUT -d @schema.json -H 'content-type: application/json' -H 'X-Cassandra-Token: ..' "http://127.0.0.4:8082/v2/namespaces/ns/collections/test/json-schema"
```

kidrecursive:
Heya, getting the same thing from `cURL`, `Postman` and `@astrajs`, against Astra and Stargate (v1.0.34)

ivansenic:
@kidrecursive OK I am investigating..

ivansenic:
[...]

PR #1294, opened 2021-10-01
closes #1293: correctly quote comment in create and alter table queries

### LINKED ISSUE #1293: `QueryBuilderImpl` adds comment without preprocessing with StringCodes
The query builder adds the comment for the create and alter table without preprocessing it with the `StringCodec`. This results in strings like `This is Mark's comment.` being converted to statements:

```
ALTER TABLE keyspace.table WITH comment = 'This is Mark's comment.';
```

which will effectively be refused by Cassandra as having syntax errors (single quote in the string value not escaped).

This currently effects the Documents API JSON schema insertion, if the schema contains fields that have single quotes.

ivansenic:
**What this PR does**:

Correctly quotes the comment that might be used while building create and alter table queries.

It's kinda insane we don't have any tests for create & alter queries, keyspaces, tables, etc.. We need to add this and I will create a ticket.

Fixes #1293 

**Checklist**
- [x] Changes manually tested
- ~[ ] Documentation added/updated~

### DIFF
--- a/persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
+++ b/persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
@@ -21,6 +21,7 @@ import static java.lang.String.format;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
 
+import com.datastax.oss.driver.internal.core.util.Strings;
 import com.datastax.oss.driver.shaded.guava.common.collect.Sets;
 import com.github.misberner.apcommons.util.AFModifier;
 import com.github.misberner.duzzt.annotations.DSLAction;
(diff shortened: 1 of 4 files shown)


tomekl007 (approved):

dougwettlaufer:
@ivansenic does this resolve #1290 as well or is there more work for that one?

tatu-at-datastax:
Looks good. Would it be possible to add IT to ensure that quoted comment gets stored properly, in addition to unit test of `QueryBuilder`?

tatu-at-datastax (approved):

ivansenic:
@dougwettlaufer yep, closed the #1290 as well..
@tatu-at-datastax there's also an INT test update, see the update in the `schema.json` file..

tatu-at-datastax:
@ivansenic Ah. Missed that, thanks!

PR #1298, opened 2021-10-01
Reorganize grpc-examples directory

dougwettlaufer:
Reorganize the grpc-examples directory to handle other languages.

**Checklist**
- ~[ ] Changes manually tested~
- ~[ ] Automated Tests added/updated~


tatu-at-datastax (approved):

pom.xml
```diff
@@ -181,7 +181,6 @@
         <module>grpc-proto</module>
         <module>grpc</module>
         <module>metrics-jersey</module>
-        <module>grpc-examples</module>
```
dougwettlaufer, line 184: Removing this from the parent pom since I'm not sure why we need to build examples along with the rest of the project.

tatu-at-datastax: If there is a way to keep it in some setting (profile), one benefit there'd be would be to ensure examples still compile?
I agree that packaging and releasing does not make much sense, not sure how to achieve just compilation (and if any tests exist, running them).

grpc-examples/golang/README.md
```diff
@@ -0,0 +1 @@
+# gRPC Rust Client
```
tomekl007, line 1: how feasible it is to move the rust examples here? I think it is not feasible. Shouldn't it live in a dedicated repo (similar to go, node)?

mpenick:
I think it makes sense to move all the examples to a new repo e.g. `stargate-grpc-examples` vesus putting them here. wdyt?

PR #1300, opened 2021-10-01
Provide a flag to disable watching the bundles directory

### LINKED ISSUE #1286: Stargate fails to start because of too many open files
I am running Stargate 1.0.31 in Kubernetes with K8ssandra. The Stargate image used is stargateio/stargate-3_11:v1.0.31. In one of our automated tests we have seen Stargate fail to start a few times with this a resource limit error like this:

```console
INFO  [main] 2021-09-28 21:25:10,597 AbstractConnector.java:331 - Started Server@29862f53{HTTP/1.1, (http/1.1)}{0.0.0.0:8082}
INFO  [main] 2021-09-28 21:25:10,598 Server.java:415 - Started @80358ms
INFO  [main] 2021-09-28 21:25:10,599 BaseActivator.java:185 - Started restapi
Finished starting bundles.
Unexpected error: java.io.IOException: User limit of inotify instances reached or too many open files
java.lang.RuntimeException: java.io.IOException: User limit of inotify instances reached or too many open files
        at io.stargate.starter.Starter.watchJarDirectory(Starter.java:539)
        at io.stargate.starter.Starter.start(Starter.java:441)
        at io.stargate.starter.Starter.cli(Starter.java:619)
        at io.stargate.starter.Starter.main(Starter.java:660)
Caused by: java.io.IOException: User limit of inotify instances reached or too many open files
        at sun.nio.fs.LinuxWatchService.<init>(LinuxWatchService.java:64)
        at sun.nio.fs.LinuxFileSystem.newWatchService(LinuxFileSystem.java:47)
        at io.stargate.starter.Starter.watchJarDirectory(Starter.java:526)
        ... 3 more
```

This is in a CI environment with limited cpu/memory resources. The test is running in the free tier runner in GitHubActions. The runner vm has 2 cpus and 7 GB memory. The particular test in which this failed had already deployed two Cassandra nodes and one Stargate node. This failure is from the second Stargate node.

I believe the open file limit on the vm is set to 65536. I don't think I am able to increase it. Maybe the solution is to run my tests in an environment with more resources, but it would be nice if Stargate could less demanding especially considering this happens on startup.

dougwettlaufer:
**What this PR does**:
This change adds a flag to optionally disable watching the bundle directory that allows for hot-reload of new jars and it will log an error instead of throwing an exception in the event there is a failure while watching the directory.

Fixes #1286

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/stargate-starter/src/main/java/io/stargate/starter/Starter.java
+++ b/stargate-starter/src/main/java/io/stargate/starter/Starter.java
@@ -233,6 +233,14 @@ public class Starter {
       name = {"--disable-mbean-registration", "Whether the mbean registration should be disabled"})
   protected boolean disableMBeanRegistration = false;
 
+  @Order(value = 21)
+  @Option(
+      name = {
+        "--disable-bundles-watch",
+        "Whether watching the bundle directory for new jars load should be disabled"
+      })
+  protected boolean disableBundlesWatch = false;
+
   @Order(value = 1000)
   @Option(
       name = "--nodetool",
@@ -438,7 +446,9 @@ public class Starter {
 
     System.out.println(STARTED_MESSAGE);
 
-    if (watchBundles) watchJarDirectory(JAR_DIRECTORY);
+    if (watchBundles && !disableBundlesWatch) {
+      watchJarDirectory(JAR_DIRECTORY);
+    }
   }
 
   protected Map<String, String> felixConfig() {
@@ -536,7 +546,9 @@ public class Starter {
         poll = pollEvents(watchService);
       }
     } catch (IOException | ClosedWatchServiceException | InterruptedException e) {
-      throw new RuntimeException(e);
+      // Since most deployments will occur in the cloud its not worth breaking the start up to
+      // support hot-reload
+      System.err.printf("Jars will not be watched due to unexpected error: %s%n", e.getMessage());
     }
   }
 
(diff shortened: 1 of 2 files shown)


stargate-starter/src/main/java/io/stargate/starter/Starter.java
```diff
@@ -536,7 +546,9 @@ private void watchJarDirectory(String directoryPath) throws BundleException {
         poll = pollEvents(watchService);
       }
     } catch (IOException | ClosedWatchServiceException | InterruptedException e) {
-      throw new RuntimeException(e);
+      // Since most deployments will occur in the cloud its not worth breaking the start up to
+      // support hot-reload
+      System.err.printf("Jars will not be watched due to unexpected error: %s%n", e.getMessage());
```
tatu-at-datastax, line 551: Since this is not via Logger system (probably since it's not configured yet), would it make sense to start with `ERROR:` prefix?

tatu-at-datastax (approved):

ivansenic (approved):
Exactly what I had in mind...

stargate-starter/src/main/java/io/stargate/starter/Starter.java
dougwettlaufer: https://github.com/stargate/stargate/issues/1303

PR #1302, opened 2021-10-02
Fix #1301: encapsulate `AuthenticatedDB` better wrt DataStore

### LINKED ISSUE #1301: Improve encapsulation of `AuthenticatedDB` (hide DataStore better)
(note: follow up to #1287)

Currently REST API has `AuthenticatedDB` to hide/bundle some aspects of access to data, but abstraction is bit leaky.
It is both used as "entity container" (so other things just ask for entities it has) and also ignored (some code just calls `Db` to access things AuthenticatedDB has).
Document API has its alternative (`DocumentDB`) which handles this a bit better.

So the next step is to refactor things to use better encapsulation to make it a more useful abstraction.

tatu-at-datastax:
**What this PR does**:

Follows up on #1287 to further improve `AuthenticatedDB` abstraction; try to remove direct use of `Db` for anything other than creating `AuthenticatedDB` instances, and simplify call patterns to avoid using unnecessary chains like `getDataStore().queryBuilder()`. In other words make `AuthenticatedDB` more similar to `DocumentDB`, actually useful abstraction on top of DataStore.

Note: will likely follow-up with renaming of `Db`, as well as reconsidering how `getAuthorizationService()` / `getAuthenticationSubject()` should be handled.

Fixes #1301

**Checklist**

- [x] Changes manually tested - no functional change so IT/CI should be enough

### DIFF
--- a/restapi/src/main/java/io/stargate/web/resources/AuthenticatedDB.java
+++ b/restapi/src/main/java/io/stargate/web/resources/AuthenticatedDB.java
@@ -60,6 +60,15 @@ public class AuthenticatedDB {
     return authenticationSubject;
   }
 
+  /**
+   * Method for trying to find and return all tables for given keyspace. Keyspace must exist for
+   * call to work; otherise {@link NotFoundException} will be thrown
+   *
+   * @param keyspaceName Name of keyspace to look for tables (must exist)
+   * @return A collection that contains all tables for given keyspace
+   * @throws NotFoundException If no keyspace with given keyspace exists in the underlying data
+   *     store
+   */
   public Collection<Table> getTables(String keyspaceName) {
     Keyspace keyspace = getKeyspace(keyspaceName);
     if (keyspace == null) {
@@ -69,6 +78,16 @@ public class AuthenticatedDB {
     return keyspace.tables();
   }
 
+  /**
+   * Method for trying to find specific table that exists in given keyspace. Keyspace must exist for
+   * call to work; otherise {@link NotFoundException} will be thrown
+   *
+   * @param keyspaceName Name of keyspace to look for tables (must exist)
+   * @param table Name of table to look for (must exist)
+   * @return Metadata for Table requested
+   * @throws NotFoundException If no keyspace with given keyspace exists in the underlying data
+   *     store, or if no table with specified name exists within that keyspace.
+   */
   public Table getTable(String keyspaceName, String table) {
     Keyspace keyspace = getKeyspace(keyspaceName);
     if (keyspace == null) {
@@ -82,19 +101,34 @@ public class AuthenticatedDB {
     return tableMetadata;
   }
 
+  /**
+   * Method for finding and returning metadata for all keyspaces for the underlying data store.
+   *
+   * @return A set of metadata for all keyspaces the underlying data store has.
+   */
   public Set<Keyspace> getKeyspaces() {
     return dataStore.schema().keyspaces();
   }
 
+  /**
+   * Method for trying to find and return metadata for given keyspace, if one exists; if none,
+   * {@code null} is returned.
+   *
+   * @param keyspaceName Name of keyspace to look for
+   * @return Metadata for keyspace requested if one exists; {@code null} otherwise.
+   */
   public Keyspace getKeyspace(String keyspaceName) {
     return dataStore.schema().keyspace(keyspaceName);
   }
 
   /**
-   * Retrieve user defined types definitions for a keyspace.
+   * Retrieve user defined types definitions for a keyspace. Keyspace must exist, otherwise a {@link
+   * NotFoundException} is thrown.
    *
    * @param keyspaceName existing keyspace name
-   * @return a collection of user defined types definitions
+   * @return a collection of user defined types definitions within specified keyspace
+   * @throws NotFoundException If no keyspace with given keyspace exists in the underlying data
+   *     store
    */
   public Collection<UserDefinedType> getTypes(String keyspaceName) {
     Keyspace keyspace = getKeyspace(keyspaceName);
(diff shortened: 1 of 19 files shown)


ivansenic (approved):
Looks good to me..

restapi/src/main/java/io/stargate/web/resources/AuthenticatedDB.java
```diff
@@ -16,22 +16,48 @@
...
+
+  public AuthorizationService getAuthorizationService() {
+    return authorizationService;
+  }
+
+  public AuthenticationSubject getAuthenticationSubject() {
+    return authenticationSubject;
   }
 
   public Collection<Table> getTables(String keyspaceName) {
```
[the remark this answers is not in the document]
tatu-at-datastax: Yes, good point. Was thinking about that wrt null checks but forgot. Will add.

restapi/src/main/java/io/stargate/web/resources/Converters.java
```diff
@@ -73,15 +72,11 @@
 
 public class Converters {
```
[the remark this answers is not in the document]
tatu-at-datastax: Possibly. I'd probably want to first wait for actual use case before making things pluggable, but I agree that this is bit awkward. There are other odd cases that I was wondering about as well: `RequestHandler`, for example, is not at all what name implies -- it is a class with static utility method(s) and not a handler.

Name here is too broad as well and definitely suggests something more (and less) than what it does.

restapi/src/main/java/io/stargate/web/resources/AuthenticatedDB.java
```diff
@@ -106,4 +124,21 @@ public UserDefinedType getType(String keyspaceName, String typeName) {
...
+
+  public QueryBuilder queryBuilder() {
+    return dataStore.queryBuilder();
+  }
+
+  public Schema schema() {
+    return dataStore.schema();
+  }
+
+  public CompletableFuture<ResultSet> execute(
```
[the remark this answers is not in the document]
tatu-at-datastax: I agree that these are sort of awkward, and I was wondering about same thing. Especially for `AuthenticatedDB`, less so for `DocumentDB`.
My main concern is that there should ideally not be multiple partially overlapping layers, unless there is some clear reason for why one layer does not encapsulate another. So the design question here is sort of like... how do we combine Authnz aspects with `DataStore`?

Having said that, if it's ok I think I'll first do this refactoring to reduce calls to `DataStore`, and then we can figure out a proper abstraction -- this because the main goal here is to separate out Docs API and REST API access.
That is needed for stargate v2 when these APIs will be separate codebases, possibly deployed separately.
I know that this does not mandate full hiding of `DataStore`, but it helps show exact data access that REST API in particular needs (which has to be re-routed to go over gRPC [or, if we preferred it, CQL protocol]).

Last thing: `QueryExecutor` is interesting... it seems that this encapsulates "extra" features Document API needs above and beyond `DataStore`. I will need to understand it better for v2 work.

restapi/src/main/java/io/stargate/web/resources/Converters.java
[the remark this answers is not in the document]
ivansenic: Yea I saw that annotation removed from `RequestHandler`.. Lot of mixup here, basically we should avoid static stuff as much as possible, because it makes unit tests way more easier and clearer..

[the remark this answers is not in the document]
tatu-at-datastax: Yea, was not sure where that annotation came about. Maybe it was non-static stuff earlier or something.
Definitely true wrt testing especially for code related to request/response handling.

restapi/src/main/java/io/stargate/web/resources/Converters.java
```diff
@@ -131,7 +126,7 @@ static Object toJsonValue(final Object cqlValue) {
     if (cqlValue instanceof InetAddress) {
       return ((InetAddress) cqlValue).getHostAddress();
     }
-    if (cqlValue instanceof List || cqlValue instanceof Set) {
+    if (cqlValue instanceof Collection) {
```
dougwettlaufer, line 129: I don't think we want to catch maps or tuples here.

restapi/src/main/java/io/stargate/web/resources/Converters.java
```diff
@@ -73,15 +72,11 @@
...
 public class Converters {
 
-  private static final ObjectMapper mapper = new ObjectMapper();
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
   private static final Pattern UNQUOTED_IDENTIFIER = Pattern.compile("[a-z][a-z0-9_]*");
   private static final Pattern PATTERN_DOUBLE_QUOTE = Pattern.compile("\"", Pattern.LITERAL);
   private static final String ESCAPED_DOUBLE_QUOTE = Matcher.quoteReplacement("\"\"");
 
-  static {
-    mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
dougwettlaufer, line 82: Why is this being removed?

restapi/src/main/java/io/stargate/web/resources/Converters.java
tatu-at-datastax: Maps are not Collections, will check Tuples.

tatu-at-datastax: `TupleValue`s do not implement `Collection` (or subtype) either. Will check offline if I should change this back.

restapi/src/main/java/io/stargate/web/resources/Converters.java
tatu-at-datastax: Should have added a comment: it is not needed here, mapper only used for writing JSON and this setting only matters for reading.

restapi/src/main/java/io/stargate/web/resources/Converters.java
dougwettlaufer: My bad, I think I got mixed up with the `isCollection` check.

dougwettlaufer (approved):

PR #1304, opened 2021-10-04
Enforce DSE guardrails when using external auth

olim7t:
**What this PR does**:

Introduce `ExternallyAuthenticatedUserRole`, a custom `UserRolesAndPermissions` subclass, in order to ensure that DSE guardrails are still enforced when external authentication is used.
Without this fix, `DsePersistence.newQueryState()` converts external users into the SYSTEM role, which is not subjected to guardrails.

N/A

**Checklist**
- [x] Changes manually tested
- [x] Automated Tests added/updated -- in a private project where external auth is covered
- ~~Documentation added/updated~~ N/A


dougwettlaufer (approved):

mpenick (approved):

tatu-at-datastax:
I don't know the system well enough to know, but I was wondering if existence of other `ClientState` implementations (like CNDB's `ClientStateWithTenantId`) would be problematic wrt sub-classing here?
(if that was a problem, would probably need to use delegation approach instead).

olim7t:
My changes break the `USE` statement (see failure in `DsePersistenceIT.testWarnings()`, I don't understand why yet.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -440,7 +440,10 @@ private DseConnection() {
     @SuppressWarnings("RxReturnValueIgnored")
     private DseConnection(@Nullable ClientInfo clientInfo, ClientState clientState) {
       super(clientInfo);
-      this.clientState = clientState;
+      this.clientState =
+          clientState instanceof StargateClientState
```
ivansenic, line 444: how about doing this in the static `StargateClientState.from(ClientState cs)` and hide the constructor?

ivansenic (commented):
Seems OK, but there is no unit nor INT test changes in this PR.. Do we have a unit test for the `DsePersistence`? Or can we simulate external auth in the INT tests? Having this merged without any tests feels wrong imo, this is quite a crucial part of the persistence..

olim7t:
> wondering if existence of other `ClientState` implementations [...] would be problematic wrt sub-classing here?

We have two other subclasses in this module already, and we do unchecked casts to them. It might be safer to merge everything into a single class, I'll take a look at it.

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -440,14 +440,14 @@ private DseConnection() {
     @SuppressWarnings("RxReturnValueIgnored")
     private DseConnection(@Nullable ClientInfo clientInfo, ClientState clientState) {
       super(clientInfo);
-      this.clientState = clientState;
+      this.clientState = ClientStateWithIsExternal.from(clientState);
```
olim7t, line 443: I confirmed that this breaks the interceptors that rely on `instanceof` checks. Pushing another commit to fix that.

olim7t:
Rebased on latest master, in particular #1180. Sorry I had to squash history to simplify conflict resolution.

PR #1305, opened 2021-10-04
Add value conversion methods for complex types

mpenick:
Add methods to convert from `Value` to Java types.

Also, add value conversion methods for less complex types for
completeness.

This doesn't include collections, UDTs or tuples. That's a PR in itself.

**Checklist**
- [X ] Changes manually tested

### DIFF
--- a/grpc-proto/src/main/java/io/stargate/grpc/Values.java
+++ b/grpc-proto/src/main/java/io/stargate/grpc/Values.java
@@ -22,11 +22,13 @@ import io.stargate.proto.QueryOuterClass.Inet;
 import io.stargate.proto.QueryOuterClass.UdtValue;
 import io.stargate.proto.QueryOuterClass.Uuid;
 import io.stargate.proto.QueryOuterClass.Value;
+import io.stargate.proto.QueryOuterClass.Value.InnerCase;
 import io.stargate.proto.QueryOuterClass.Value.Null;
 import io.stargate.proto.QueryOuterClass.Value.Unset;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
 import java.time.LocalDate;
 import java.time.LocalTime;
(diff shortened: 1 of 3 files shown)


grpc-proto/src/test/java/ValuesTest.java
```diff
@@ -0,0 +1,323 @@
...
+import java.nio.ByteBuffer;
+import java.time.LocalDate;
+import java.time.LocalTime;
+import java.util.UUID;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+public class ValuesTest {
+
+  @Nested
```
tatu-at-datastax, line 18: Quick question on this: I have seen use of `@Nested`, but was curious as to its benefits -- is it just for improved grouping of test results, or such? Nothing against the use just want to understand benefits over either including tests directly in main test class, or as separate main-level test cases.

grpc-proto/src/test/java/ValuesTest.java
```diff
@@ -0,0 +1,323 @@
...
+  class IntTest {
+
+    @Test
+    public void encodeDecode() {
+      int expected = Integer.MAX_VALUE;
+      assertThat(Values.int_(Values.of(expected))).isEqualTo(expected);
+    }
+
+    @Test
+    public void invalid() {
```
tatu-at-datastax, line 47: Would it also be possible to test value overflow, since it is checked by converter?

grpc-proto/src/main/java/io/stargate/grpc/Values.java
```diff
@@ -163,4 +165,143 @@ public static Value of(Map<Value, Value> fields) {
...
+    }
+    return value.getBoolean();
+  }
+
+  public static int int_(Value value) {
+    if (value.getInnerCase() != InnerCase.INT) {
+      throw new IllegalArgumentException("Expected int value");
+    }
+    int intValue = (int) value.getInt();
+    if (intValue != value.getInt()) {
```
tatu-at-datastax, line 181: Would this actually fail? As opposed to using same coercion here as in above line.
(I did add a note to test this case to make sure)

grpc-proto/src/main/java/io/stargate/grpc/Values.java
```diff
@@ -163,4 +165,143 @@ public static Value of(Map<Value, Value> fields) {
...
+
+  public static float float_(Value value) {
+    if (value.getInnerCase() != InnerCase.FLOAT) {
+      throw new IllegalArgumentException("Expected float value");
+    }
+    return value.getFloat();
+  }
+
+  public static double double_(Value value) {
+    if (value.getInnerCase() != InnerCase.DOUBLE) {
```
tatu-at-datastax, line 227: This is probably correct but just want to verify: should it not coerce from `FLOAT` to `DOUBLE`?

grpc-proto/src/main/java/io/stargate/grpc/Values.java
```diff
@@ -163,4 +165,143 @@ public static Value of(Map<Value, Value> fields) {
...
+      throw new IllegalArgumentException("Expected double value");
+    }
+    return value.getDouble();
+  }
+
+  public static ByteBuffer byteBuffer(Value value) {
+    if (value.getInnerCase() != InnerCase.BYTES) {
+      throw new IllegalArgumentException("Expected bytes value");
+    }
+    ByteBuffer bytes = ByteBuffer.allocate(value.getBytes().size());
```
tatu-at-datastax, line 237: Does `getBytes()` always return a new copy? If so, would be good to use temporary variable to hold on to that copy to avoid extra copy.
(if method returns raw value instance, does not matter)

grpc-proto/src/main/java/io/stargate/grpc/Values.java
```diff
@@ -163,4 +165,143 @@ public static Value of(Map<Value, Value> fields) {
...
+  }
+
+  public static UUID uuid(Value value) {
+    if (value.getInnerCase() != InnerCase.UUID) {
+      throw new IllegalArgumentException("Expected uuid value");
+    }
+    ByteBuffer bytes = ByteBuffer.allocate(16);
+    value.getUuid().getValue().copyTo(bytes);
+    bytes.flip();
+    return new UUID(bytes.getLong(bytes.position()), bytes.getLong(bytes.position() + 8));
```
tatu-at-datastax, line 263: Could just use positions `0` and `8`, since we allocated `ByteBuffer`.
That is, unless we could avoid copy altogether if `value` exposes `ByteBuffer` or allows read of 2 `long`s we need
(java-uuid-generator package would also have way to constructor `UUID` directly from `byte[]` but probably don't want to add a new dependency just for that).

grpc-proto/src/main/java/io/stargate/grpc/Values.java
```diff
@@ -163,4 +165,143 @@ public static Value of(Map<Value, Value> fields) {
...
+  }
+
+  public static InetAddress inet(Value value) {
+    if (value.getInnerCase() != InnerCase.INET) {
+      throw new IllegalArgumentException("Expected inet value");
+    }
```
tatu-at-datastax, line 269: Should we explicitly check the length (4 or 16 bytes) here?

grpc-proto/src/test/java/ValuesTest.java
mpenick: done

grpc-proto/src/test/java/ValuesTest.java
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    @Test
+    public void encodeDecodeV6() throws UnknownHostException {
+      InetAddress expected = InetAddress.getByName("2001:db8::8a2e:370:7334");
+      assertThat(Values.inet(Values.of(expected))).isEqualTo(expected);
+    }
+
+    @Test
+    public void invalid() {
```
tatu-at-datastax, line 290: Should probably also test for lengths other than 4 or 16 bytes (maybe take legit Ipv4/v6 value, append one more byte).

grpc-proto/src/main/java/io/stargate/grpc/Values.java
mpenick: it returns a `long`

grpc-proto/src/main/java/io/stargate/grpc/Values.java
mpenick: it doesn't return a copy.

grpc-proto/src/main/java/io/stargate/grpc/Values.java
tatu-at-datastax: Ok then that makes sense (even if odd naming convention :) ).

grpc-proto/src/main/java/io/stargate/grpc/Values.java
tatu-at-datastax: Ok then that's fine.

grpc-proto/src/main/java/io/stargate/grpc/Values.java
mpenick: done

grpc-proto/src/test/java/ValuesTest.java
mpenick: done

grpc-proto/src/main/java/io/stargate/grpc/Values.java
mpenick: good catch. I wish that that `UUID(byte[])` constructor was public. Switch to using absolute positions.

grpc-proto/src/main/java/io/stargate/grpc/Values.java
mpenick: yeah, that's a good thought. I think I would like to avoid coercion mainly because it get murky fast with other types of conversions like integer types can be stored as floating point types and floating point types (and other types) can be represented as a string, etc.

tatu-at-datastax: Yes that makes sense. And if there is need this could be relaxed at a later point.

grpc-proto/src/main/java/io/stargate/grpc/Values.java
tatu-at-datastax: Right. Seems odd that constructor (or factory method) is not provided.

tatu-at-datastax (approved):

grpc-proto/src/main/java/io/stargate/grpc/Values.java
```diff
@@ -163,4 +165,150 @@ public static Value of(Map<Value, Value> fields) {
         .setCollection(Collection.newBuilder().addAllElements(values).build())
         .build();
   }
+
+  public static boolean bool(Value value) {
+    if (value.getInnerCase() != InnerCase.BOOLEAN) {
+      throw new IllegalArgumentException("Expected boolean value");
```
ivansenic, line 171: should we enrich all exception messages with the value that was provided, it might help the user?

ivansenic (approved):

grpc-proto/src/main/java/io/stargate/grpc/Values.java
mpenick: good suggestion: https://github.com/stargate/stargate/pull/1305/commits/12fc903cccb14e4f78cae69e17dccac92eb4c73a

grpc-proto/src/main/java/io/stargate/grpc/Values.java
```diff
@@ -163,4 +165,141 @@ public static Value of(Map<Value, Value> fields) {
...
+      throw new IllegalArgumentException(
+          String.format("Valid range for int is %d to %d", Integer.MIN_VALUE, Integer.MAX_VALUE));
+    }
+    return intValue;
+  }
+
+  public static long bigint(Value value) {
+    checkInnerCase(value, InnerCase.INT);
+
+    return value.getInt();
```
tomekl007, line 189: shouldn't be `value.getVarint()`?

grpc-proto/src/main/java/io/stargate/grpc/Values.java
```diff
@@ -163,4 +165,141 @@ public static Value of(Map<Value, Value> fields) {
...
+
+    return new BigDecimal(
+        new BigInteger(value.getDecimal().getValue().toByteArray()), value.getDecimal().getScale());
+  }
+
+  public static LocalDate date(Value value) {
+    checkInnerCase(value, InnerCase.DATE);
+
+    int unsigned = value.getDate();
+    int signed = unsigned + Integer.MIN_VALUE;
```
tomekl007, line 289: just for curiosity, why do we need `+ Integer.MIN_VALUE` and `- Integer.MIN_VALUE;` for another way around? Can't we drop this addition/substraction?

grpc-proto/src/main/java/io/stargate/grpc/Values.java
mpenick: `.getVarint()` is for `varint` not `bigint`.

grpc-proto/src/main/java/io/stargate/grpc/Values.java
mpenick: This is the representation of date:

```
    // CQL types: date
    // An unsigned integer representing days with Unix epoch (January, 1 1970) at 2^31.
    // Examples:
    // 0:    -5877641-06-23
    // 2^31: 1970-1-1
    // 2^32: 5881580-07-11
    uint32 date = 11;
```

PR #1307, opened 2021-10-04
Compile path call is removed, unit testing passes

EricBorczuk:
**What this PR does**:
Removes a call to JsonPathCompiler#compile, which seems to take a long time especially for documents with many paths and deep nesting.

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested


EricBorczuk:
I will post some data about the before/after tomorrow AM

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
```diff
@@ -233,6 +233,17 @@ public static String convertJsonToBracketedPath(String path) {
     return newPath.toString();
   }
 
+  public static String[] bracketedPathAsArray(String bracketedPath) {
+    String[] split = bracketedPath.split("\\]\\[");
```
ivansenic, line 237: can we have a compiled pattern for this split? I think this will not take fast path:

```
        /* fastpath if the regex is a
         (1)one-char String and this character is not one of the
            RegEx's meta characters ".$|()[{^?*+\\", or
         (2)two-char String and the first char is the backslash and
            the second is not the ascii digit or ascii letter.
         */
```

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
```diff
@@ -233,6 +233,17 @@ public static String convertJsonToBracketedPath(String path) {
     return newPath.toString();
   }
 
+  public static String[] bracketedPathAsArray(String bracketedPath) {
```
ivansenic, line 236: add java docs

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
```diff
@@ -233,6 +233,17 @@ public static String convertJsonToBracketedPath(String path) {
...
+    String[] split = bracketedPath.split("\\]\\[");
+    if (split.length > 0) {
+      split[0] = split[0].substring(2);
```
ivansenic, line 239: should we immediately do a split on the `bracketedPath.substring(2, bracketedPath.length -1)` and thus avoid this gymnastics here?

restapi/src/main/java/io/stargate/web/docsapi/service/DocsShredder.java
```diff
@@ -177,42 +173,40 @@ private Object convertToBackendBooleanValue(boolean value, boolean numericBoolea
...
+    String[] bracketedPathArray = DocsApiUtils.bracketedPathAsArray(bracketedPath);
+    for (int j = 0; j < bracketedPathArray.length; j++) {
       if (i >= config.getMaxDepth()) {
         throw new ErrorCodeRuntimeException(ErrorCode.DOCS_API_GENERAL_DEPTH_EXCEEDED);
       }
 
-      PathOperator op = it.next();
-      String pathValue = op.toString();
+      String pathValue = bracketedPathArray[j];
+      boolean isArrayValue = !pathValue.startsWith("'");
```
ivansenic, line 192: use `charAt(0) !=`?

restapi/src/main/java/io/stargate/web/docsapi/service/DocsShredder.java
```diff
@@ -177,42 +173,40 @@ private Object convertToBackendBooleanValue(boolean value, boolean numericBoolea
...
-        }
-        if (isAtTopLevel(i, path.size())) {
-          firstLevelKeys.add(innerPath);
-        }
-        String convertedPath = convertPathValueForArrays(innerPath, op);
-        unboundPaths[i++] = convertedPath;
-        leaf = convertedPath;
+      // pathValue always starts and ends with a square brace because of
+      // DocsApiUtils#convertJsonToBracketedPath
+      String innerPath = DocsApiUtils.convertEscapedCharacters(pathValue);
```
ivansenic, line 196: can we not do this for arrays, I guess there is no need? Should be optimized by a compiler though.. Ignore..

ivansenic (approved):
Esoooooo, great stuff.. See the comments for some extra speed.. Any idea on the benefits?

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
EricBorczuk: The idea here is that the input string could be something like `$['a'][0]['b']` and I want an array with values ['a', 0, 'b']` from it.

So the initial split gets me a weird looking array with `["$['a'", "0", "'b']"]`. Then I have to clean up the first and last elements of the array by chopping off the first two characters of the first element `$['a'` and the final character of the last element `b'`. So I think the two steps here might be necessary

EricBorczuk:
I used a sample JSON of ~15kB size, from the "Sample bitcoin transaction" API

On master, for a 150 second sample:
6.6% of CPU time was spent on JsonPathCompiler#compile.
p50 response time was 502ms.

On this branch, JsonPathCompiler#compile was removed so the new function bracketedPathAsArray was measured instead. For a ~200 second sample:
0.1% of CPU time was spent on bracketedPathAsArray
p50 response time was 472ms.

So there was a bit of an improvement, and there's definitely more to be made in the near future!

EricBorczuk:
After commit `f2e0beb`, average response time dropped to 381ms! This is because a regex was being recompiled for every single path value in every row 🤦

PR #1310, opened 2021-10-04
(WIP) Fix #1309: split `Db` into separate factories, rename data access types

### LINKED ISSUE #1309: Split, rename existing `Db` factory into `RestDBFactory`, `DocDBFactory`
Currently entity named `Db` works as a factory for data-access helpers `DocumentDB` (doc-api) and `AuthenticatedDB` (REST api). It should be split into 2 replacements, along with bit of supporting renaming, for eventual goal of cleanly splitting REST and Document APIs from each other.

tatu-at-datastax:
(Work in Progress)

**What this PR does**:

Continuation of splitting of REST/Document APIs, improve naming.

Fixes #1309

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/restapi/src/main/java/io/stargate/web/impl/RestApiServer.java
+++ b/restapi/src/main/java/io/stargate/web/impl/RestApiServer.java
@@ -51,6 +51,7 @@ import io.stargate.web.resources.v2.schemas.IndexesResource;
 import io.stargate.web.resources.v2.schemas.KeyspacesResource;
 import io.stargate.web.resources.v2.schemas.TablesResource;
 import io.stargate.web.resources.v2.schemas.UserDefinedTypesResource;
+import io.stargate.web.restapi.dao.RestDBAccessFactory;
 import io.stargate.web.swagger.SwaggerUIResource;
 import io.stargate.web.validation.ViolationExceptionMapper;
 import io.swagger.config.ScannerFactory;
@@ -116,12 +117,9 @@ public class RestApiServer extends Application<ApplicationConfiguration> {
   public void run(
       final ApplicationConfiguration applicationConfiguration, final Environment environment)
       throws IOException {
+    environment.jersey().register(new ViolationExceptionMapper());
     final Db db =
         new Db(authenticationService, authorizationService, dataStoreFactory, docsApiConf);
-
-    configureObjectMapper(environment.getObjectMapper());
-
-    environment.jersey().register(new ViolationExceptionMapper());
     environment
         .jersey()
         .register(
@@ -131,13 +129,26 @@ public class RestApiServer extends Application<ApplicationConfiguration> {
                 bind(db).to(Db.class);
               }
             });
+    final RestDBAccessFactory restDBFactory =
+        new RestDBAccessFactory(authenticationService, authorizationService, dataStoreFactory);
     environment
         .jersey()
         .register(
             new AbstractBinder() {
               @Override
               protected void configure() {
-                bind(environment.getObjectMapper()).to(ObjectMapper.class);
+                bind(restDBFactory).to(RestDBAccessFactory.class);
+              }
+            });
+
+    final ObjectMapper objectMapper = configureObjectMapper(environment.getObjectMapper());
+    environment
+        .jersey()
+        .register(
+            new AbstractBinder() {
+              @Override
+              protected void configure() {
+                bind(objectMapper).to(ObjectMapper.class);
               }
             });
 
@@ -200,9 +211,10 @@ public class RestApiServer extends Application<ApplicationConfiguration> {
     environment.jersey().property(ServerProperties.RESPONSE_SET_STATUS_OVER_SEND_ERROR, true);
   }
 
-  public static void configureObjectMapper(ObjectMapper objectMapper) {
+  public static ObjectMapper configureObjectMapper(ObjectMapper objectMapper) {
     objectMapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
     objectMapper.registerModule(new JavaTimeModule());
+    return objectMapper;
   }
 
   @Override
--- a/restapi/src/main/java/io/stargate/web/restapi/dao/RestDBAccess.java
+++ b/restapi/src/main/java/io/stargate/web/restapi/dao/RestDBAccess.java
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.stargate.web.resources;
+package io.stargate.web.restapi.dao;
 
 import io.stargate.auth.AuthenticationSubject;
 import io.stargate.auth.AuthorizationService;
@@ -34,16 +34,15 @@ import javax.ws.rs.NotFoundException;
 import org.apache.cassandra.stargate.db.ConsistencyLevel;
 
 /**
- * Data store abstraction used by Rest API: encapsulates authentication aspects along with
+ * Data access abstraction used by Rest API: encapsulates authentication aspects along with
  * underlying actual {@link DataStore}, offers some more convenience access.
  */
-public class AuthenticatedDB {
-
+public class RestDBAccess {
   private final DataStore dataStore;
   private final AuthenticationSubject authenticationSubject;
   private final AuthorizationService authorizationService;
 
-  public AuthenticatedDB(
+  public RestDBAccess(
       DataStore dataStore,
       AuthenticationSubject authenticationSubject,
       AuthorizationService authorizationService) {
--- a/restapi/src/main/java/io/stargate/web/restapi/dao/RestDBAccessFactory.java
+++ b/restapi/src/main/java/io/stargate/web/restapi/dao/RestDBAccessFactory.java
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.stargate.web.resources;
+package io.stargate.web.restapi.dao;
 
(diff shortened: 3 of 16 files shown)


restapi/src/main/java/io/stargate/web/impl/RestApiServer.java
```diff
@@ -51,6 +51,7 @@
 import io.stargate.web.resources.v2.schemas.KeyspacesResource;
 import io.stargate.web.resources.v2.schemas.TablesResource;
 import io.stargate.web.resources.v2.schemas.UserDefinedTypesResource;
+import io.stargate.web.restapi.dao.RestDBAccessFactory;
```
tatu-at-datastax, line 54: Part of former `Db` used for creating "data store" for REST API.

restapi/src/main/java/io/stargate/web/resources/Db.java
```diff
@@ -33,8 +33,6 @@
 import java.util.stream.Collectors;
 
 public class Db {
-
-  private final DataStore dataStore;
```
tatu-at-datastax, line 37: `DataStore` was no longer used for anything before this PR: factory used when creating new instances. So just removing as unused.

restapi/src/main/java/io/stargate/web/resources/Db.java
```diff
@@ -44,12 +42,6 @@
           .expireAfterWrite(Duration.ofMinutes(1))
           .build(this::getDocDataStoreForTokenInternal);
 
-  private final LoadingCache<TokenAndHeaders, AuthenticatedDB> restTokensToDataStore =
```
tatu-at-datastax, line 47: Moved to the new `RestDBAccessFactory`

restapi/src/main/java/io/stargate/web/docsapi/rx/RxUtils.java
```diff
@@ -15,7 +15,7 @@
  *
  */
 
-package io.stargate.web.rx;
+package io.stargate.web.docsapi.rx;
```
tatu-at-datastax, line 18: Moved to new package since it was only used by Doc API (to help eventual split)

restapi/src/main/java/io/stargate/web/restapi/dao/RestDBAccess.java
```diff
@@ -175,6 +173,50 @@ public Schema schema() {
     return dataStore.schema();
   }
 
+  public void authorizeDataRead(String keyspace, String table, SourceAPI sourceAPI)
```
tatu-at-datastax, line 176: Added pass-through methods to encapsulate `authenticationSubject` (instead of giving accessor), to slightly simplify call patterns from resources.

restapi/src/main/java/io/stargate/web/restapi/dao/RestDBAccessFactory.java
```diff
@@ -0,0 +1,123 @@
...
+import io.stargate.db.datastore.DataStore;
+import io.stargate.db.datastore.DataStoreFactory;
+import io.stargate.db.datastore.DataStoreOptions;
+import java.time.Duration;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.CompletionException;
+import java.util.stream.Collectors;
+
+public class RestDBAccessFactory {
```
tatu-at-datastax, line 33: Extracted pieces from `Db` related to creating accessor for Rest API to use.

restapi/src/main/java/io/stargate/web/docsapi/rx/RxUtils.java
ivansenic: Yes, but I hope this kind of utilities will be needed in all new stuff, as hopefully we are going reactive by default?

restapi/src/main/java/io/stargate/web/impl/RestApiServer.java
```diff
@@ -131,13 +129,26 @@ protected void configure() {
...
             });
+    final RestDBAccessFactory restDBFactory =
+        new RestDBAccessFactory(authenticationService, authorizationService, dataStoreFactory);
+    environment
+        .jersey()
+        .register(
+            new AbstractBinder() {
+              @Override
+              protected void configure() {
+                bind(restDBFactory).to(RestDBAccessFactory.class);
```
ivansenic, line 140: is there any chance we can group the stuff in one `AbstractBinder` instead of creating a new one for each registration here.. It's just easier to read at the end? We also extracted docs stuff into an own binder..

restapi/src/main/java/io/stargate/web/restapi/dao/RestDBAccessFactory.java
ivansenic, line 33: Java docs, unit tests missing?

restapi/src/main/java/io/stargate/web/restapi/dao/RestDBAccessFactory.java
```diff
@@ -0,0 +1,123 @@
...
+      AuthenticationSubject authenticationSubject, TokenAndHeaders tokenAndHeaders) {
+    return dataStoreFactory.create(
+        authenticationSubject.asUser(),
+        DataStoreOptions.builder()
+            .alwaysPrepareQueries(true)
+            .putAllCustomProperties(tokenAndHeaders.headers)
+            .build());
+  }
+
+  static class TokenAndHeaders {
```
ivansenic, line 90: we have this duplicated now? extract, abstract class?

restapi/src/main/java/io/stargate/web/restapi/dao/RestDBAccess.java
ivansenic: Again, seems we are complicating more than simplifying..

ivansenic (commented):
Not sure what are we really gaining with this, except more complexity..

restapi/src/main/java/io/stargate/web/docsapi/rx/RxUtils.java
tatu-at-datastax: If so can definitely move into shareable component -- there are some other similar things.
So what would be a good place, something like `stargate-service-common`? Or would any of existing modules work?

On very short term I just want to avoid things shared by only 2 things (REST and Doc APIs), so this move is just to note that as-is this is only used by Doc API.

restapi/src/main/java/io/stargate/web/impl/RestApiServer.java
tatu-at-datastax: I was following an existing pattern, no reason for not combining. Will change this.

restapi/src/main/java/io/stargate/web/restapi/dao/RestDBAccessFactory.java
tatu-at-datastax: Yes, I will add Javadocs.
As to tests: since this is part moved from `Db`, existing tests should exercise functionality?
Are there specific aspects that may be missing coverage?

restapi/src/main/java/io/stargate/web/restapi/dao/RestDBAccessFactory.java
tatu-at-datastax: If we are to split Document and REST APIs, they should not share anything just amongst each other: either there is bit of duplication, or we extract common implementation into something that may be shared by other services, I think.
This particular inner class is not exposed outside at all so we could go either way.

My thinking here would be to first use bit of duplication, but then extract reusable class into some common component. I just don't know where -- worth a discussion?

tatu-at-datastax:
@ivansenic @dougwettlaufer I understand why routing of Query execution, Auth checks via renamed `RestDBAccess` appear as unnecessary complexity (perhaps it is?).

My thinking here is that by doing this, all REST operations will only use API of that one class and need not expose either `DataStore`, `AuthenticationService`, `AuthenticationSubject` or `AuthorizationService`.
There are a few added pass-through methods which for `DataStore` do not add much convenience for now -- but do limit surface API developer sees which to me is a plus -- but for access checks remove need to use 2 getters.
There is potential need for adding more pass-through methods but likelihood of significant extra work seems low.

And from Resource method side, code to me is bit cleaner. But that's matter of taste, since there is the added encapsulation.

So in the end I agree that this is an arbitrary choice. 

I am open to undoing of `getDataStore()` as well as 2 accessors for Auth aspects.

If adding `getDataStore()`, resource classes could then access it for any operations: whether 2 one-line `execute()` convenience methods should be dropped I don't know. To me they simplify call patterns.

restapi/src/main/java/io/stargate/web/docsapi/rx/RxUtils.java
tatu-at-datastax: @dougwettlaufer pointed me to: https://github.com/stargate/stargate/issues/239 -- so if that existed, would probably belong there?

restapi/src/main/java/io/stargate/web/impl/RestApiServer.java
tatu-at-datastax: Combined -- not sure what'd be the cleanest way wrt grouping (REST vs Doc API), but at least less verbose now.

restapi/src/main/java/io/stargate/web/docsapi/rx/RxUtils.java
tatu-at-datastax: Will consider as part of more general "how to reuse code".

dougwettlaufer (approved):

PR #1311, opened 2021-10-05
add request counter metrics possibility

ivansenic:
**What this PR does**:
Adds ability to count HTTP Jersey requests, in addition to metering them. By default, counting is focusing on the exceptions. Both metering and counting are enabled by default, but they are fully configurable, so they can be turned off if needed. The reason for adding counting as well is that the cardinality of metrics wrt different tags.

I added the `CounterApplicationEventListener` that basically increases the counter on each finished request and consults the tag provider for any extra tags.

This PR also had to deal with the small refactoring of the `metrics-jersey` module, so that this update is possible: 

* in the `io.stargate.metrics.jersey.tags` now we have dedicated implementations for different tag providers, breaking what was before in a single class
* the `CompositeJerseyTagsProviderTest` has been added so different providers can be mixed up
* the `MetricsBinder` registers needed application listeners based on the passed configuration

I am creating additional PR on top of this for:
* adding `README.MD` in the module where I would list all configuration possibilities
* extracting logic from `DefaultHttpMetricsTagProvider` to dedicated tag provider, as this is easily possible now (will provide `NoopHttpMetricsTagProvider`)

-

**Checklist**
- [x] Changes manually tested
- [ ] :x:  Documentation added/updated (in the follow-up PR)


metrics-jersey/src/main/java/io/stargate/metrics/jersey/MetricsBinder.java
```diff
@@ -0,0 +1,190 @@
...
+      MetricsApplicationEventListener listener =
+          new MetricsApplicationEventListener(
+              metrics.getMeterRegistry(),
+              meterTagsProvider,
+              StargateMetricConstants.METRIC_HTTP_SERVER_REQUESTS,
+              true);
+      jersey.register(listener);
+    }
+
+    if (counterListenerConfig.isEnabled()) {
```
tomekl007, line 131: what's the easiest way to disable it?

metrics-jersey/src/main/java/io/stargate/metrics/jersey/MetricsBinder.java
```diff
@@ -0,0 +1,190 @@
...
+      CounterApplicationEventListener listener =
+          new CounterApplicationEventListener(
+              metrics.getMeterRegistry(),
+              counterTagsProvider,
+              StargateMetricConstants.METRIC_HTTP_SERVER_REQUESTS_COUNTER);
+      jersey.register(listener);
+    }
+  }
+
+  private static JerseyTagsProvider getMeterTagsProvider(
```
tomekl007, line 143: what's the difference between this and `getCounterTagsProvider`? Can't we have one method for it?

metrics-jersey/src/main/java/io/stargate/metrics/jersey/listener/CounterApplicationEventListener.java
```diff
@@ -0,0 +1,65 @@
...
+      MeterRegistry meterRegistry, JerseyTagsProvider tagsProvider, String metricName) {
+    this.meterRegistry = meterRegistry;
+    this.tagsProvider = tagsProvider;
+    this.metricName = metricName;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public void onEvent(ApplicationEvent event) {
+    // ignored
```
tomekl007, line 57: why ignored? shouldn't this propagate that event to `CounterRequestEventListener.onEvent`?

metrics-jersey/src/main/java/io/stargate/metrics/jersey/config/SystemPropsMetricsListenerConfig.java
```diff
@@ -0,0 +1,60 @@
...
+   * @param propertyPrefix Prefix of the system properties to use.
+   */
+  public SystemPropsMetricsListenerConfig(String propertyPrefix) {
+    this.propertyPrefix = propertyPrefix;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public boolean isEnabled() {
+    String value = System.getProperty(propertyWithPrefix("enabled"), "true");
```
tomekl007, line 46: I wonder if this is a good granularity for this setting. Is it possible that we would like to disable all counter/meter metrics? If yes, how easy it will be to do using this mechanism?

metrics-jersey/src/main/java/io/stargate/metrics/jersey/tags/CompositeJerseyTagsProvider.java
```diff
@@ -0,0 +1,64 @@
...
+    for (JerseyTagsProvider provider : delegateProviders) {
+      tags = tags.and(provider.httpRequestTags(event));
+    }
+
+    return tags;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public Iterable<Tag> httpLongRequestTags(RequestEvent event) {
```
tomekl007, line 55: why do we need both `httpLongRequestTags` and `httpRequestTags`?

PR #1315, opened 2021-10-06
Add ability to close/reject persistent connections

mpenick:
This adds the ability to close or reject CQL connections based on
whether a instance of `Persistence` is currently able to handle the
client's requests. Metadata properties are used to identify the
particular client.

**Checklist**
- [X] Changes manually tested
- [ ] ~Automated Tests added/updated~
I was unable to think of a way to add an automated test. Ideas welcome.
- [ ] ~Documentation added/updated~


mpenick:
@sbtourist Yeah, both the race and some driver rely on the server to close the connection even for remove events... So it's a necessary patch.

Issue #1303, opened 2021-10-04
Add helper method for logging in Starter.java


dougwettlaufer:
In the starter class we `System.out` for "logging" since logback hasn't been enabled yet in this context. This is fine except it makes it a little difficult to grok for Starter related logs. 

Add a helper method to prepend levels to the messages to make it easier to identify errors versus info.

shashankbrgowda:
Hi, Can I work on this issue ?

ivansenic:
@shashankbrgowda Hi, we already have a PR for this issue in review: https://github.com/stargate/stargate/pull/1348.

dougwettlaufer:
@Lukkaeuss are you still working this issue?

Lukkaeuss:
> @Lukkaeuss are you still working this issue?

Hello, yes im still working on it. I expect to do PR in upcoming week if that's ok with you.

PR #1348, opened 2021-10-18
Add helper method for logging in Starter.java #1303

Lukkaeuss:


tatu-at-datastax:
@Lukkaeus in case you were still interested, this seems like a good improvement if concerns were addressed?

jeffreyscarpenter:
Opened #2418 using a patch so original author's work is preserved, then added another commit to address the unresolved feedback on this PR. Closing.

PR #1361, opened 2021-10-21
Fix #1255: replace Guava cache with Caffeine in config-store-yaml

### LINKED ISSUE #1255: Change `config-store-yaml` to use Caffeine cache instead of Guava
Other cache implementations have been changed to use Caffeine, should do the same for `config-store-yaml`.

tatu-at-datastax:
**What this PR does**:

Removes (shaded) Guava dependency from `config-store-yaml` by using Caffeine cache instead of Guava one.
Specifies Caffeine version in parent pom as managed, instead of being included in separate poms.

Also adds Caffeine as a "shared" dependency (from parent pom) which allows for avoiding inclusion of private copies by sub-modules -- this change only affects `config-store-yaml` but a followup PR can then remove copies from `restapi`, `cql` and `grpc` modules.

**NOTE** Right now parent inclusion leads to excessive copies -- need to figure out how to avoid that. I **think** the solution is to add non-provided dependency to `core`, make it export Caffeine to others

Fixes #1255

**Checklist**
- [x] Changes manually tested - relies on existing test suite


tatu-at-datastax:
Apparently this module is not really even being used. Still, getting rid of (shaded) Guava dependency seems useful on short term, before Stargate v2.

PR #1366, opened 2021-10-25
Return UNAVAILABLE for unhandled clients

mpenick:
**What this PR does**:

Handles `UnhandleClientException` by returning the `UNAVAILABLE` status. Currently, the `UNKNOWN` status is being returned. This is an interim fix which allows clients to at least attempt to retry when a client becomes unhandled by the persistence layer.

**Checklist**
- [X] Changes manually tested


tomekl007:
I’ve tried the retry approach when experimenting with nb driver.
The problem with providing `serviceConfig` map is that the `ManagedChannelBuilder` provides a very coarse-grained API. So at the Client-side, we need to do something like:
```
     Map<String, Map<String, ?> > serviceConfig = new HashMap<>();
        Map<String, Object> retryConfig = new HashMap<>();
        retryConfig.put("retryableStatusCodes", Arrays.asList("UNAVAILABLE"));
        retryConfig.put("maxAttempts", Arrays.asList("UNAVAILABLE"));
        Map<String, Object> methodConfig = new HashMap<>();
        methodConfig.put("retryPolicy", retryConfig);
        serviceConfig.put("methodConfig", retryConfig);


       ManagedChannelBuilder.forAddress(host, port)
                    .directExecutor()
                    .enableRetry()
                    .defaultServiceConfig(serviceConfig)
                    .build();
```
I am not sure how does it look like in other languages. 

The other question/concern. Are we sure that every time the `UNAVAILABLE` exception is thrown, it is thrown for an idempotent query and it is safe to retry? Maybe instead of propagating the `UNAVAILABLE` in this case, we should create a custom error code that denotes this tenant_id migration and is always safe to retry?

tomekl007:
> Agreed, but it's better than returning `UNKNOWN` and is an interim fix. `UNAVAILABLE` is a truthful error, the coordinator is unavailable to the client. We still need to handle the exception in some way.

So I understand that this issue is only about translating that exception into `UNAVAILABLE`. Should we create a follow-up ticket with retry support for clients?

tomekl007 (approved):

PR #1386, opened 2021-11-01
(WIP) SGv2: use `QueryBuilder` for "getAllRows()"

tatu-at-datastax:
**What this PR does**:

Minor change to use property `QueryBuilder` for constructing query for "getAllRows()"

No issue filed

**Checklist**
- [x] Changes manually tested -- passes same integration tests as before.

### DIFF
--- a/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResource.java
+++ b/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResource.java
@@ -4,7 +4,9 @@ import com.codahale.metrics.annotation.Timed;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.BytesValue;
 import com.google.protobuf.Int32Value;
-import com.google.protobuf.InvalidProtocolBufferException;
+import io.stargate.db.query.builder.BuiltQuery;
+import io.stargate.db.query.builder.QueryBuilder;
+import io.stargate.db.schema.Column;
 import io.stargate.grpc.StargateBearerToken;
 import io.stargate.proto.QueryOuterClass;
 import io.stargate.proto.StargateGrpc;
@@ -13,15 +15,19 @@ import io.stargate.sgv2.restsvc.grpc.ExtProtoValueConverters;
 import io.stargate.sgv2.restsvc.impl.GrpcClientFactory;
 import io.stargate.sgv2.restsvc.models.RestServiceError;
 import io.stargate.sgv2.restsvc.models.Sgv2RowsResponse;
+import io.stargate.sgv2.restsvc.util.SchemaNoPersistence;
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
 import io.swagger.annotations.ApiResponse;
 import io.swagger.annotations.ApiResponses;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Base64;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import javax.servlet.http.HttpServletRequest;
@@ -96,12 +102,9 @@ public class Sgv2RowsResource {
           final boolean raw,
       @ApiParam(value = "Keys to sort by") @QueryParam("sort") final String sort,
       @Context HttpServletRequest request) {
-    if (isStringEmpty(fields)) {
-      fields = "*";
-    } else {
-      fields = removeSpaces(fields);
-    }
-    final String cql = String.format("SELECT %s from %s.%s", fields, keyspaceName, tableName);
+
+    final String cql = buildSelectAllQuery(keyspaceName, tableName, fields);
+
     logger.info("Calling gRPC method: try to call backend with CQL of '{}'", cql);
 
     StargateGrpc.StargateBlockingStub blockingStub =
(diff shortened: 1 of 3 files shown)


sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResource.java
```diff
@@ -139,6 +142,40 @@
     return javax.ws.rs.core.Response.status(Response.Status.OK).entity(response).build();
   }
 
+  private String buildSelectAllQuery(String keyspaceName, String tableName, String fields) {
+    List<Column> columns = fieldsToColumns(fields);
+    BuiltQuery<?> qb =
+        new QueryBuilder(new SchemaNoPersistence(), null, null)
```
ivansenic, line 148: So this constructor is so that we don't reference any existing schema right?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/util/SchemaNoPersistence.java
```diff
@@ -0,0 +1,35 @@
...
+
+  @Override
+  public Set<Keyspace> keyspaces() {
+    return Collections.emptySet();
+  }
+
+  @Override
+  public Keyspace keyspace(String name) {
+    if (name == null) {
+      return ANONYMOUS;
```
ivansenic, line 27: what's the deal with anonymous?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/util/SchemaNoPersistence.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+    return ImmutableKeyspace.builder().name(name).build();
+  }
+
+  public List<String> keyspaceNames() {
+    return keyspaces().stream().map(k -> k.name()).sorted().collect(Collectors.toList());
```
ivansenic, line 33: Is this implemented like this as it's expected that the `keyspaces()` could be overriden? Also it feels a bit strange that `keyspaces` and `keyspaceNames` would return empty stuff, but if you reference a keyspace by name in the `keyspaceNames`, then you would get the non-null result?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResource.java
tatu-at-datastax: Correct: new api/service instances do not have access to true Schema.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/util/SchemaNoPersistence.java
tatu-at-datastax: Copied from original `Schema`: not sure what the intent there is. Could try just leaving it out, or create dynamically.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/util/SchemaNoPersistence.java
tatu-at-datastax: This was copy-pasted from `Schema`: I'll go ahead and simply return empty List instead, makes more sense in this case. Thanks!

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/util/SchemaNoPersistence.java
ivansenic: No idea, someone else should comment here I guess..

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/util/SchemaNoPersistence.java
ivansenic: What about the second point I mentioned?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/util/SchemaNoPersistence.java
mpenick: Does it mean a query without a keyspace? `SELECT * FROM tableName` vs `SELECT * FROM keyspaceName.tableName`? I'm not sure that even makes sense in the context of the REST API where the resource requires a valid keyspace.

tatu-at-datastax: How about I'll remove it from here since all usage here should be for proper keyspace and `null` should be invalid.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/util/SchemaNoPersistence.java
tatu-at-datastax: From modeling perspective this is not right, I agree; but at this point this is the way to be able to use Stargate `QueryBuilder`. I am not sure we necessarily should -- there is an alternative from Cassandra Java driver -- but at this point we are still prototyping.
If we decide to try to make this QB work better, be used, need to figure out how to refactor it in a way that does not rely on `Schema`.

tatu-at-datastax:
Quick note: found out that this change DOES NOT actually work -- I get a failure due to missing table because `Keyspace` instance `Schema` creates is "empty", so attempts to get `Table` fails. So I'd need to create sub-class there as well.

Not sure this is the road to take, but will investigate a little bit more before giving up.

tatu-at-datastax:
Came to conclusion that this will not work by mocking/stubbing; closing PR.

PR #1389, opened 2021-11-02
Add schema agreement check to persistence connection

mpenick:
This adds a more granular level of checking for schema agreement. If the
underlying persistence implementation supports per-client schema this
allows for other custom, client-level properties to be accounted for
when checking the schema.

**Checklist**
- [x] Changes manually tested
- [X] Documentation added/updated (added javadoc comments)

### DIFF
--- a/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/SchemaAgreement.java
+++ b/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/SchemaAgreement.java
@@ -24,14 +24,14 @@ public class SchemaAgreement {
   private static final ScheduledThreadPoolExecutor EXECUTOR = new ScheduledThreadPoolExecutor(1);
 
   public static CompletableFuture<? extends Result> maybeWaitForAgreement(
-      CompletableFuture<? extends Result> future, Persistence persistence) {
+      CompletableFuture<? extends Result> future, Persistence.Connection connection) {
     return future.thenCompose(
         (result) -> {
           CompletableFuture<? extends Result> resultFuture = future;
           if (result.kind == Result.Kind.SchemaChange) {
             CompletableFuture<Result> agreementFuture = new CompletableFuture<Result>();
             EXECUTOR.schedule(
-                new SchemaAgreement.Handler(agreementFuture, result, persistence),
+                new SchemaAgreement.Handler(agreementFuture, result, connection),
                 SCHEMA_AGREEMENT_RETRIES_INTERVAL_MILLIS,
                 TimeUnit.MILLISECONDS);
             resultFuture = agreementFuture;
@@ -42,20 +42,20 @@ public class SchemaAgreement {
 
   private static class Handler implements Runnable {
     private final CompletableFuture<Result> future;
-    private final Persistence persistence;
+    private final Persistence.Connection connection;
     private final Result result;
     private int count;
 
-    Handler(CompletableFuture<Result> future, Result result, Persistence persistence) {
+    Handler(CompletableFuture<Result> future, Result result, Persistence.Connection connection) {
       this.future = future;
-      this.persistence = persistence;
+      this.connection = connection;
       this.result = result;
       this.count = SCHEMA_AGREEMENT_WAIT_RETRIES;
     }
 
     @Override
     public void run() {
-      if (persistence.isInSchemaAgreement()) {
+      if (connection.isInSchemaAgreement()) {
         future.complete(result);
       } else if (count > 0) {
         count--;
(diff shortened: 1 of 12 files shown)


cql/src/main/java/org/apache/cassandra/stargate/transport/internal/SchemaAgreement.java
```diff
@@ -24,14 +24,14 @@
   private static final ScheduledThreadPoolExecutor EXECUTOR = new ScheduledThreadPoolExecutor(1);
 
   public static CompletableFuture<? extends Result> maybeWaitForAgreement(
```
tatu-at-datastax, line 26: Technically this would be API change; should there be a deprecated variant with old signature, calling new one with `persistence.connection()`? I don't know if anyone/anything outside Stargate calls this method, if not no need to bother I guess.

In future we probably need to consider API changes more carefully and add javadoc tags ("@since [version]").

mpenick: `maybeWaitForAgreement()` is internally implementation only.

persistence-api/src/main/java/io/stargate/db/Persistence.java
```diff
@@ -252,5 +233,16 @@ default void setCustomProperties(Map<String, String> customProperties) {}
...
+
+    /**
+     * Checks whether this coordinator in schema agreement with the other nodes in the cluster. This
+     * is the same as {@code Persistence#isInSchemaAgreement()}, except that it *may* consider other
+     * connection specific properties to determine whether schema is in agreement.
+     *
+     * @return true if schema is agreement; otherwise false.
+     */
+    default boolean isInSchemaAgreement() {
+      return persistence().isInSchemaAgreement();
```
ivansenic, line 245: this would actually guarantee the same old behavior by default here.. But any `Connection` impl can provide it's own version.. Nice :+1:

persistence-api/src/main/java/io/stargate/db/Persistence.java
```diff
@@ -34,10 +32,7 @@
  * higher level {@link DataStore} API which wraps an instance of this interface) to query the
  * underlying store, and thus the one interface that persistence extensions must implement.
  */
-public interface Persistence {
-
-  int SCHEMA_AGREEMENT_WAIT_RETRIES =
-      Integer.getInteger("stargate.persistence.schema.agreement.wait.retries", 900);
+public interface Persistence extends SchemaAgreementChecker {
```
ivansenic, line 35: I guess you should now remove the declaration of the `boolean isInSchemaAgreement();` as it's now part of the `SchemaAgreementChecker` interface?

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
```diff
@@ -1666,7 +1666,7 @@ public void testWarnings() throws ExecutionException, InterruptedException {
     execute(conn, "USE " + ks.cqlName());
     execute(conn, "CREATE TABLE t (k int PRIMARY KEY)");
 
-    persistence().waitForSchemaAgreement();
+    conn.waitForSchemaAgreement();
```
ivansenic, line 1669: are we sure these 3 places are the only ones where we need to switch from `Persistence#waitForSchemaAgreement()` to `Connection#waitForSchemaAgreement`? 

Furthermore, is there a need now that `Persistence` also implements `SchemaAgreementChecker`? I guess yes because we also need to check if we are in schema agreement without actually having the connection, or? I wanted to double check, as if we can remove it we would be sure that all code has transitioned to the new way by a simple compile check..

ivansenic (approved):
Approving, but have a look on the comments please..

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
mpenick: Updated the `DataStore`s which are the only users of the `waitForSchemaAgreement()` other than a test. This is used by the HTTP services (REST, GraphQL). CQL and gRPC use `Connection#isInSchemaAgreement()` directly.

persistence-api/src/main/java/io/stargate/db/Persistence.java
mpenick: done.

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
tomekl007: @mpenick one question regarding the direct usage of `Connection#isInSchemaAgreement()` in gRPC. Why it is used only in `QueryHandler`, but we don't need to use it in the `BatchHandler`? Can you elaborate more on this?

tomekl007 (approved):

mpenick:
@tomekl007 good question. It's because DDL queries cannot be done in batches.

PR #1394, opened 2021-11-03
Add "createKeyspace()" and "get[All]Keyspace[s]" schema REST endpoints for v2.0 prototype

tatu-at-datastax:
**What this PR does**:

Will (once complete) add support for Stargate V2 endpoint for creating keyspace(s), as well as accessing one or all.

Not filed for incremental progress towards REST SGv2 prototype.

**Checklist**
- [x] Changes manually tested -- uses existing IT

### DIFF
--- a/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/models/Sgv2Keyspace.java
+++ b/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/models/Sgv2Keyspace.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.sgv2.restsvc.models;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModelProperty;
+import java.util.List;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class Sgv2Keyspace {
+  @JsonProperty("name")
+  private String name;
+
+  @JsonProperty("datacenters")
+  private List<Datacenter> datacenters;
+
+  @ApiModelProperty(
+      value =
+          "The datacenters within a keyspace. Only applies for those keyspaces created with NetworkTopologyStrategy.")
+  public List<Datacenter> getDatacenters() {
+    return datacenters;
+  }
+
+  public void setDatacenters(List<Datacenter> datacenters) {
+    this.datacenters = datacenters;
+  }
+
+  @ApiModelProperty(required = true, value = "The name of the keyspace.")
+  @JsonProperty("name")
+  public String getName() {
+    return name;
+  }
+
+  public Sgv2Keyspace setName(String name) {
+    this.name = name;
+    return this;
+  }
+
+  @JsonCreator
+  public Sgv2Keyspace(
+      @JsonProperty("name") String name,
+      @JsonProperty("datacenters") List<Datacenter> datacenters) {
+    this.name = name;
+    this.datacenters = datacenters;
+  }
+
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  public static class Datacenter {
+    private String name;
+    private int replicas;
+
+    @ApiModelProperty(required = true, value = "The name of the datacenter.")
+    public String getName() {
+      return name;
+    }
+
+    public Datacenter setName(String name) {
+      this.name = name;
+      return this;
+    }
+
+    @ApiModelProperty(
+        required = true,
+        value =
+            "The number of replicas in the datacenter. In other words, the number of copies of each row in the datacenter.")
+    public int getReplicas() {
+      return replicas;
+    }
+
+    public Datacenter setReplicas(int replicas) {
+      this.replicas = replicas;
+      return this;
+    }
+
+    @JsonCreator
+    public Datacenter(@JsonProperty("name") String name, @JsonProperty("replicas") int replicas) {
+      this.name = name;
+      this.replicas = replicas;
+    }
+  }
+}
--- a/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/models/Sgv2Keyspace.java
+++ b/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/models/Sgv2Keyspace.java
@@ -19,15 +19,25 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import io.swagger.annotations.ApiModelProperty;
+import java.util.ArrayList;
 import java.util.List;
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public class Sgv2Keyspace {
-  @JsonProperty("name")
-  private String name;
+  protected String name;
 
-  @JsonProperty("datacenters")
-  private List<Datacenter> datacenters;
+  protected List<Datacenter> datacenters;
+
+  protected Sgv2Keyspace() {} // for deserializer
+
+  public Sgv2Keyspace(String name) {
+    this.name = name;
+  }
+
+  @ApiModelProperty(required = true, value = "The name of the keyspace.")
+  public String getName() {
+    return name;
+  }
 
   @ApiModelProperty(
       value =
@@ -36,33 +46,23 @@ public class Sgv2Keyspace {
     return datacenters;
   }
 
-  public void setDatacenters(List<Datacenter> datacenters) {
-    this.datacenters = datacenters;
-  }
-
-  @ApiModelProperty(required = true, value = "The name of the keyspace.")
-  @JsonProperty("name")
-  public String getName() {
-    return name;
-  }
-
-  public Sgv2Keyspace setName(String name) {
-    this.name = name;
-    return this;
-  }
-
-  @JsonCreator
-  public Sgv2Keyspace(
-      @JsonProperty("name") String name,
-      @JsonProperty("datacenters") List<Datacenter> datacenters) {
-    this.name = name;
-    this.datacenters = datacenters;
+  public void addDatacenter(String name, int replicas) {
+    if (datacenters == null) {
+      datacenters = new ArrayList<>();
+    }
+    datacenters.add(new Datacenter(name, replicas));
   }
 
   @JsonInclude(JsonInclude.Include.NON_NULL)
   public static class Datacenter {
     private String name;
     private int replicas;
+
+    @JsonCreator
+    public Datacenter(@JsonProperty("name") String name, @JsonProperty("replicas") int replicas) {
+      this.name = name;
+      this.replicas = replicas;
+    }
 
     @ApiModelProperty(required = true, value = "The name of the datacenter.")
     public String getName() {
@@ -85,12 +85,6 @@ public class Sgv2Keyspace {
     public Datacenter setReplicas(int replicas) {
       this.replicas = replicas;
       return this;
-    }
-
-    @JsonCreator
-    public Datacenter(@JsonProperty("name") String name, @JsonProperty("replicas") int replicas) {
-      this.name = name;
-      this.replicas = replicas;
     }
   }
 }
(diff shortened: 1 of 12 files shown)


sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/models/Sgv2Keyspace.java
```diff
@@ -0,0 +1,96 @@
...
+package io.stargate.sgv2.restsvc.models;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModelProperty;
+import java.util.List;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class Sgv2Keyspace {
```
tatu-at-datastax, line 25: Not yet needed for CREATE KEYSPACE but is the type returned with get keyspace/keyspaces.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/models/Sgv2Keyspace.java
```diff
@@ -0,0 +1,96 @@
...
+ */
+package io.stargate.sgv2.restsvc.models;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModelProperty;
+import java.util.List;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
ivansenic, line 24: This looks like so much of the boilerplate code.. Why aren't we already moving to some other framework where these would be so easier (spring, micronaut, whatever).. If we are refactoring already, I would expect that there we have:

* some kind of the lombok support
* moved to OpenAPI and Swagger 3.0

Sorry but I don't understand why we opted in having same issues as before and actually not improving anything.. This first module should imo set standards and work on important updates wrt to V1.. It's gonna be really hard to move away from this if we do that later on..

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/models/Sgv2Keyspace.java
```diff
@@ -0,0 +1,96 @@
...
+public class Sgv2Keyspace {
+  @JsonProperty("name")
+  private String name;
```
ivansenic, line 27: This also has no validations? If this is used in the posy body I would be able to pass null here.. See docs api for the validation support..

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResource.java
```diff
@@ -0,0 +1,246 @@
...
+    consumes = MediaType.APPLICATION_JSON,
+    tags = {"schemas"})
+@Path("/v2/schemas/keyspaces")
+@Produces(MediaType.APPLICATION_JSON)
+@Singleton
+public class Sgv2KeyspacesResource {
+  // Singleton resource so no need to be static
+  private final Logger logger = LoggerFactory.getLogger(getClass());
+
+  private static final SchemaBuilderHelper schemaBuilder =
```
ivansenic, line 64: shouldn't `SchemaBuilderHelper` also be injected? why static?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResource.java
```diff
@@ -0,0 +1,246 @@
...
+      new SchemaBuilderHelper(new JsonMapper());
+
+  /** Entity used to connect to backend gRPC service. */
+  @Inject private GrpcClientFactory grpcFactory;
+
+  @Timed
```
ivansenic, line 70: Why `@Timed`?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResource.java
```diff
@@ -0,0 +1,246 @@
...
+  @ApiResponses(
+      value = {
+        @ApiResponse(code = 200, message = "OK", response = Sgv2Keyspace.class),
+        @ApiResponse(code = 401, message = "Unauthorized", response = RestServiceError.class),
+        @ApiResponse(
+            code = 500,
+            message = "Internal server error",
+            response = RestServiceError.class)
+      })
+  public Response getAllKeyspaces(
```
ivansenic, line 86: Why aren't we moving all endpoints to the async/reactive implementations? Again, I am really worried about the approach we took.

ivansenic (commented):
I am worried about the approach we took for this new module.. We are not improving anything, seems like a copy/paste? I would be up in discussing more..

jeffreyscarpenter:
> I am worried about the approach we took for this new module.. We are not improving anything, seems like a copy/paste? I would be up in discussing more..
For v2 we are checking in partial work on branches and using PRs to the v2.0.0 to get visibility into each other's work. Some of this work gets merged into v2.0.0, some does not. Tatu can confirm, but I don't think the intent is that this is ready to go into v2 yet.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResource.java
jeffreyscarpenter: We will get there, just working a bit at a time as we get things started.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/models/Sgv2Keyspace.java
jeffreyscarpenter: Thanks for the items above on specific expectations, those are helpful. Please let us know if there are more.

dougwettlaufer: @ivansenic I was thinking this would be more incremental. First pass there would be minimal changes to the resources aside from having a more proper abstraction between resource and service (I should not have put as much logic in the resource as I did, instead they should have called a service class). Then once we get everything working we'd refactor to a newer framework.

So the first round of reviews would be pushing down logic. Second round would be replacing persistence calls for gRPC calls. Next we could rip out OSGI and make the service stand alone. And finally we'd work in a new framework. Those last two steps could be combined if we want. I'm not a fan of rewriting everything at once since it make it harder for review and when you're debugging you don't know which of the moving parts actually broke things.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResource.java
dougwettlaufer: Dropwizard annotation to get timing metrics.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResource.java
```diff
@@ -0,0 +1,246 @@
+/*
```
dougwettlaufer, line 1: How much of this is new and how much is copy-paste? This seems like something we can do in place in `io.stargate.web.restapi.resources.v2.schemas.KeyspacesResource`

dougwettlaufer: If you want to keep tests green then just change swap out the DataStore for gRPC on a case by case basis rather than running two services in parallel.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/models/Sgv2Keyspace.java
ivansenic: Yea I feel that I would favor more the bottom-up approach here, than the top-down.. I do understand the plan, but I feel this last step you mentioned would be then a refactoring on it's own again. I mean I don't have anything against current plan and don't want to stop further actions.. I just see so much potential for improvement and modernizing everything.. And also creating like a standardized way of doing things, even the simple things like package structures f.e.. Anyway, no need to discuss more..

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResource.java
ivansenic: We should go away from Dropwizard metrics for the endpoints.. Even now we would have the micrometer support we built.. Plus, I don't know if this has any effect..

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResource.java
tatu-at-datastax: I have been going back and forth wrt trying to mix-and-match old and new types; both approaches have (IMO) benefits. Copy-pasting -- for now, not necessarily in the end -- has the benefit of using optimal approach, minimizing dependencies. To me attempting to mix-and-match so far has seemed to pull in unwanted dependencies and significantly complicate progress.

But we can definitely discuss this further; this is very much WIP.

tatu-at-datastax:
@ivansenic I am not sure what the specific concern is: yes, there is copy-paste for reimplementation. This is to try to minimize, if possible, dependencies to v1 implementation and keep v2 minimal. V1 does not try to isolate things very hard, which for many types means that one type pulls two others and so forth. This is not the case for all types (Keyspaces type could be reused if that really matters -- but that's bit of an outlier).

When getting closer to something we could actually use for v2 it is not too difficult to try to retrofit things.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResource.java
tatu-at-datastax: It is just an internal stateless helper type: no benefit from DI or stateful instance.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResource.java
tatu-at-datastax: Yes, simply copied. I was wondering about moving away from JAX-RS (DropWizard is a thin layer: It's Jersey's implementation of JAX-RS spec) to something else but:

1. Main focus is to test out functionality, not to change platform (as far as I understand)
2. Many/most new Java service frameworks still use/support JAX-RS annotations.

So. I am totally fine moving to something else -- just not quite yet.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResource.java
tatu-at-datastax: @ivansenic Please do not consider this deployable production version. We are prototyping and trying to find the simplest functionally working version. Once we get further we will -- and need to -- revisit implementation details. Right now it seems challenging enough to convert v1 functionality snippets to simplest (blocking) alternative.

Perhaps it'd make sense to start a document outlining everything to revisit, however?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResource.java
tatu-at-datastax: And as to copy-paste: only things that can be used without modifications, and that do not bring other dependencies, are reused.

Resource classes in particular cannot be reused: some value types conceivably could.

But as to resources: this is what I was trying to suggest earlier -- signatures, annotations are potentially reusable, and could be extracted into interfaces (or abstract classes), implemented by v1 and v2 separately.
This is possible. But it is not clear it'd be worth the complexity for reader caused by interface/impl division of every Resource class.

tatu-at-datastax:
After adding "get keyspace(s)" implementations, merging to v2.0.0: not because comments wouldn't be valid (for future discussions & likely changes) but because right now PRs are more FYIs and since I need to share code with @jeffreyscarpenter (and vice versa). There obviously needs to be walk-throughs and bigger code reviews down the line as a consequence -- idea is not to avoid reviews but do them in a more efficient manner when code itself is more reviewable (less of prototype).

Improvement suggestions are obviously welcome & appreciated regardless. My intention is not to ignore it.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResource.java
ivansenic: unit tests?

tatu-at-datastax: In that case just get the static instance to test I would think.
(although testing itself can be bit challenging wrt creation of gRPC types).

PR #1398, opened 2021-11-10
Stargate v2 - draft PR for updates to gRPC API

jeffreyscarpenter:
Draft PR for adding schema operations to gRPC protos.

See #1401 for additional detail.

### DIFF
--- a/grpc-proto/proto/query.proto
+++ b/grpc-proto/proto/query.proto
@@ -311,6 +311,7 @@ message Traces {
   repeated Event events = 4;
 }
 
+// TODO: does this need to include indexes and materialized views?
 // If a Query message is a DDL statement, this will be included in the Response to describe the
 // how the CQL schema was impacted.
 message SchemaChange {
--- a/grpc-proto/proto/schema.proto
+++ b/grpc-proto/proto/schema.proto
@@ -1,0 +1,141 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+syntax = "proto3";
+
+import "google/protobuf/wrappers.proto";
+
+option java_package = "io.stargate.proto";
+option go_package = "stargate/proto";
+
+package stargate;
+
+// The type of a column
+enum CqlType {
+  INT = 0x00;
+  FLOAT = 0x01;
+  DOUBLE = 0x02;
+  BOOLEAN = 0x03;
+  STRING = 0x04;
+  BYTES = 0x05;
+  INET = 0x06;
+  UUID = 0x07;
+  DATE = 0x08;
+  TIME = 0x09;
+  COLLECTION = 0x0A;
+  VARINT = 0x0B;
+  DECIMAL = 0x0C;
+  UDT = 0x0D;
+}
+
+enum CqlCollectionType {
+  SET = 0x01;
+  LIST = 0x02;
+  MAP = 0x03;
+}
+
+// The type of a column
+enum CqlColumnOrder {
+  ASC = 0x00;
+  DESC = 0x01;
+}
+
+// A CQL Column
+message CqlColumn {
+
+  string name = 1;
+
+  CqlType type = 2;
+
+  CqlCollectionType collectionType = 3;
+
+  string udtType = 4;
+
+  bool isPartitionKey = 5;
+
+  uint32 clusteringColumnOrder = 6;
+
+  bool frozen = 7;
+
+  bool static = 8;
+}
+
+message CqlTableOptions {
+
+}
+
+// USER DEFINED TYPES
+
+message CqlUdtField {
+  string name = 1;
+
+  CqlType type = 2;
+}
+
+message CqlUserDefinedType {
+  string name = 1;
+
+  repeated CqlUdtField type = 2;
+}
+
+// TABLES
+
+message CqlTable {
+  repeated CqlColumn columns = 1;
+
+  CqlTableOptions tableOptions = 2;
+}
+
+// INDEXES
+
+message CqlIndex {
+
+}
+
+// MATERIALIZED VIEWS
+
+message CqlMaterializedView {
+
+}
+
+// KEYSPACES
+
+message CqlDatacenterReplication {
+  string datacenterName = 1;
+
+  uint32 replicationFactor = 2;
+}
+
+message CqlKeyspace {
+
+  string name = 1;
+
+  repeated CqlDatacenterReplication replication = 2;
+
+  bool durableWrites = 3;
+
+  repeated CqlUserDefinedType types = 4;
+
+  repeated CqlTable tables = 5;
+
+  repeated CqlIndex indexes = 6;
+
+  repeated CqlMaterializedView materializedViews = 7;
+}
+
+
+// questions
+// should we provide distinct operations for creating schema?
+// should we reuse/extend the SchemaChange type? it is not fully featured enough to include a full representation of a table
--- a/grpc-proto/proto/schema.proto
+++ b/grpc-proto/proto/schema.proto
@@ -25,23 +25,34 @@ package stargate;
 // TYPES
 
 enum CqlType {
-  INT = 0x00;
-  FLOAT = 0x01;
-  DOUBLE = 0x02;
+  ASCII = 0x00;
+  BIGINT = 0x01;
+  BLOB = 0x02;
   BOOLEAN = 0x03;
-  STRING = 0x04;
-  BYTES = 0x05;
-  INET = 0x06;
-  UUID = 0x07;
-  DATE = 0x08;
-  TIME = 0x09;
-  VARINT = 0x0A;
-  DECIMAL = 0x0B;
-  SET = 0x0C;
-  LIST = 0x0D;
+  COUNTER = 0x04;
+  DATE = 0x05;
+  DECIMAL = 0x06;
+  DOUBLE = 0x07;
+  DURATION = 0x08;
+  FLOAT = 0x09;
+  INET = 0x0A;
+  INT = 0x0B;
+  LIST = 0x0C;
+  LINESTRING = 0x0D;
   MAP = 0x0E;
-  UDT = 0x0F;
-  TUPLE = 0x10;
+  POINT = 0x0F;
+  POLYGON = 0x10;
+  SMALLINT = 0x11;
+  TEXT = 0x12;
+  TIME = 0x13;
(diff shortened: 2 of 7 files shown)


grpc-proto/proto/stargate.proto
```diff
@@ -29,5 +30,11 @@ service Stargate {
 
   // Executes a batch of CQL queries.
   rpc ExecuteBatch(Batch) returns (Response) {}
+
+  // Schema creation
+  rpc ExecuteSchemaQuery(CqlSchemaQuery) returns (Response) {}
```
tatu-at-datastax, line 35: I know Cassandra calls all kinds of operations "queries", so maybe we want to be compatible with that.
But outside C*, queries are for reading/requesting things, not modifications. So wondering if semantically it'd be better to use "SchemaOperation" / "SchemaChange" or something like that?

jeffreyscarpenter: Good thought, I'll change to `SchemaOperation`

tatu-at-datastax:
Looks quite compact, nice!

I have some questions around things:

1. Granularity of access: most of our initial access only needs a small subset of metadata (for keyspaces, just name & replication status, for example; for tables info on just a single one) -- unless we immediately jump into caching challenge and fetch "full schema" and extract pieces. Alternatively could have request parameters to indicate what info to include (if asking keyspaces, whether to include tables and so on).
2. Tenant id add/removal: there is code to handle possible (tenant) prefix; is handling needed on calling side (client) as well or not?
3. Access limits: since this is part of public API, do we need to worry about different access caller has for schema information vs. access to data itself? (apparently one could be able to, say, INSERT, but not DESCRIBE?)

jeffreyscarpenter:
> Looks quite compact, nice!
> 
> I have some questions around things:
> 
> 1. Granularity of access: most of our initial access only needs a small subset of metadata (for keyspaces, just name & replication status, for example; for tables info on just a single one) -- unless we immediately jump into caching challenge and fetch "full schema" and extract pieces. Alternatively could have request parameters to indicate what info to include (if asking keyspaces, whether to include tables and so on).
I was thinking to defer the lower granularity access until we get more clarity on the caching need, in which case I was assuming we'd just be pushing full schema to clients.

> 2. Tenant id add/removal: there is code to handle possible (tenant) prefix; is handling needed on calling side (client) as well or not?
My assumption was that passing along headers in the client was sufficient, perhaps others know with more certainty.

> 3. Access limits: since this is part of public API, do we need to worry about different access caller has for schema information vs. access to data itself? (apparently one could be able to, say, INSERT, but not DESCRIBE?) 
Great question. Looks like there is a DESCRIBE permission, guess we would need to check the current user has that permission for the requested database?

grpc-proto/proto/schema.proto
```diff
@@ -0,0 +1,208 @@
...
+  bool isStatic = 8;
+}
+
+// TABLES
+
+// TODO ok to make it this simplistic?
+message CqlTableOption {
+
+  string name = 1;
+  string value = 2;
```
olim7t, line 131: I'm a bit torn on this.

According to the CQL spec, option values can be:
- simple constant literals: string ,integer, float, boolean, duration, uuid, hex number
- map literals

If we go with simple strings, we can just concatenate them in the generated CQL string. But that means that string constants will have to be single-quoted, which is not super user-friendly.

Note that this is kind of a moot point, because we barely use table options in the current implementations: REST only uses `default_time_to_live` for table creation (see `TableOptions`). GraphQL does not support options at all.

And on the way out (describe), the `io.stargate.db.schema.Table` representation returned by our persistence layer does not even contain options.

olim7t (commented):
First round of comments on the protos.

grpc-proto/proto/schema.proto
```diff
@@ -0,0 +1,211 @@
...
+}
+
+message CqlTable {
+
+  string name = 1;
+
+  repeated CqlColumn columns = 2;
+
+  repeated CqlTableOption tableOptions = 3;
```
mpenick, line 140: Should we expose table options? This is pretty low-level for this type of API.

grpc-proto/proto/stargate.proto
```diff
@@ -29,5 +30,11 @@ service Stargate {
 
   // Executes a batch of CQL queries.
   rpc ExecuteBatch(Batch) returns (Response) {}
+
+  // Schema creation
+  rpc ExecuteSchemaOperation(CqlSchemaOperation) returns (Response) {}
```
mpenick, line 35: What do you think about specific RPC calls for creating each type of schema?

```grpc
rpc CreateKeyspace(...) returns (...) {}
rpc CreateTable(...) returns (...) {}
rpc CreateType(...) return (...) {}
```

grpc-proto/proto/schema.proto
```diff
@@ -0,0 +1,211 @@
...
+
+  repeated CqlMaterializedView materializedViews = 5;
+}
+
+// SCHEMA CREATION
+
+message CqlSchemaOperation {
+  oneof schema {
+
+    CqlTable table = 1;
```
mpenick, line 199: This might be missing a create keyspace operation.

grpc-proto/proto/schema.proto
mpenick: Never mind, if we allow creating tables we need all the information. :)

olim7t: The REST create table operation allows a custom `default_time_to_live`.

grpc-proto/proto/query.proto
```diff
@@ -311,6 +311,7 @@ message Traces {
   repeated Event events = 4;
 }
 
+// TODO: does this need to include indexes and materialized views?
```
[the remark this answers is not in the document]
jeffreyscarpenter: I'm not sure what the suggestion is?

grpc-proto/proto/schema.proto
```diff
@@ -0,0 +1,208 @@
...
+  //     CqlComplexColumnType(UDT, address, true)
+  // for TUPLE, this will have as many elements as are referenced
+  //   example: values <tuple<int, text, float>> -> CqlComplexColumnType(INT, null, false),
+  //     CqlComplexColumnType(STRING, null, false), CqlComplexColumnType(FLOAT, null, false)
+  repeated CqlComplexColumnType complexColumnTypes = 4;
+
+  bool isPartitionKey = 5;
+
+  // only used for clustering columns
+  uint32 clusteringColumnOrder = 6;
```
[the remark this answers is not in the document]
jeffreyscarpenter: no, this is not about order by. This is what order the column appears in the clustering column list (if any)

grpc-proto/proto/schema.proto
jeffreyscarpenter: I'm thinking of changing this to map<string, string>

grpc-proto/proto/schema.proto
jeffreyscarpenter: I'm thinking of changing this to map<string, string>. 

And I am wondering if we will end up needing to make changes to the persistence layer to get the full options back.

grpc-proto/proto/stargate.proto
jeffreyscarpenter: that would be simpler in some ways. I was trying to emulate the style of ExecuteQuery/ExecuteBatch, but am fine breaking them out

grpc-proto/proto/schema.proto
jeffreyscarpenter: will add

grpc-proto/proto/schema.proto
```diff
@@ -0,0 +1,208 @@
...
+import "google/protobuf/wrappers.proto";
+
+option java_package = "io.stargate.proto";
+option go_package = "stargate/proto";
+
+package stargate;
+
+// TYPES
+
+enum CqlType {
```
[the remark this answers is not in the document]
jeffreyscarpenter: Ok, will look at extending these existing types and reusing.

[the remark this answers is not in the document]
jeffreyscarpenter: The challenge is that these types are designed for giving clients just enough information to be able to interpret the structure of Rows. They don't include important information such as whether a column is a partition key, clustering key, static, frozen, etc. That makes things more complicated.

grpc-proto/proto/query.proto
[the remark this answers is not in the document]
olim7t: Sorry, I meant that this is how the built-in Cassandra events (surfaced to us via `Persistence.registerEventListener`) work. The suggestion is to simply forward them as-is.

grpc-proto/proto/schema.proto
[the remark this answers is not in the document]
olim7t: We can extend them with optional fields.

tatu-at-datastax:
> > Looks quite compact, nice!
> > I have some questions around things:
> > 
> > 1. Granularity of access: most of our initial access only needs a small subset of metadata (for keyspaces, just name & replication status, for example; for tables info on just a single one) -- unless we immediately jump into caching challenge and fetch "full schema" and extract pieces. Alternatively could have request parameters to indicate what info to include (if asking keyspaces, whether to include tables and so on).
> 
> I was thinking to defer the lower granularity access until we get more clarity on the caching need, in which case I was assuming we'd just be pushing full schema to clients.

Lower as-in... more granular (specific pieces)?

> 
> > 1. Tenant id add/removal: there is code to handle possible (tenant) prefix; is handling needed on calling side (client) as well or not?
> 
> My assumption was that passing along headers in the client was sufficient, perhaps others know with more certainty.

As long as we cache nothing, yes; but once we start caching on API side, yes -- otherwise keyspace names will collide (and everything else). Since (afaik) its Persistence that currently handles name decoration/undecoration.

> 
> > 1. Access limits: since this is part of public API, do we need to worry about different access caller has for schema information vs. access to data itself? (apparently one could be able to, say, INSERT, but not DESCRIBE?)
> 
> Great question. Looks like there is a DESCRIBE permission, guess we would need to check the current user has that permission for the requested database?

This is only part of the problem: yes, when explicitly asking for schema, users rights need to be checked -- but NOT when Stargate itself must know schema for, say, GET-by-primary-key.

grpc-proto/proto/schema.proto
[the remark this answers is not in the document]
olim7t: OK I see. Don't we need this for partition key columns too then?

Another approach would be to return separate groups of columns in `CqlTable` (similar to what we do in the Persistence API):
```
message CqlTable {
  ...
  repeated CqlColumn partitionKeyColumns = 2;
  repeated CqlColumn clusteringColumns = 3;
  repeated CqlColumn regularAndStaticColumns = 4;
  ...
}
```

`repeated` guarantees a consistent order.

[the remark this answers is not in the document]
jeffreyscarpenter: that is interesting. if I broke apart `regularAndStaticColumns` then I could also omit the `CqlColumn.isStatic` attribute.

grpc-proto/proto/stargate.proto
```diff
@@ -29,5 +30,14 @@ service Stargate {
...
+
+  // Similar to CQL "DESCRIBE KEYSPACE"
+  rpc DescribeKeyspace(DescribeKeyspaceQuery) returns (CqlKeyspaceDescribe) {}
+
+  // Similar to CQL "DESCRIBE KEYSPACE"
```
tatu-at-datastax, line 40: "DESCRIBE TABLE"

PR #1402, opened 2021-11-10
Implement "addRow()" for SGv2 prototype

### LINKED ISSUE #1399: Implement "addRow()" for SGv2 prototype
As part of incremental work for prototyping Stargate V2 REST api, we need ability to insert rows (currently implemented: create keyspace, get all rows).

Work is also necessary for investigating whether there is typing mismatch between less strict incoming JSON, and internal gRPC bridge interface we use; and if so, how to work around or solve the problem.

tatu-at-datastax:
**What this PR does**:

Adds support for INSERT/UPSERT of a single row; "addRows()" REST/v2 resource.

Also improves mapping of gRPC errors by SGv2 prototype.

Fixes #1399

**Checklist**
- [x] Changes manually tested - Uses existing Integration Test suite; changes improve pass rate (only minority pass due to incomplete implementation, so far)

### DIFF
--- a/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RequestHandler.java
+++ b/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RequestHandler.java
@@ -16,6 +16,8 @@
 package io.stargate.sgv2.restsvc.resources;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import io.grpc.Status;
+import io.grpc.StatusRuntimeException;
 import io.stargate.sgv2.restsvc.models.RestServiceError;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
@@ -30,8 +32,55 @@ public class Sgv2RequestHandler {
   public static Response handle(Callable<Response> action) {
     try {
       return action.call();
-    } catch (NotFoundException nfe) {
-      logger.info("Resource not found (NotFoundException->NOT_FOUND): {}", nfe.getMessage());
+    } catch (StatusRuntimeException grpcE) { // from gRPC
+      Status.Code sc = grpcE.getStatus().getCode();
+      switch (sc) {
+          // Start with codes we know how to handle
+        case INVALID_ARGUMENT:
+          return grpcResponse(
+              sc,
+              Response.Status.BAD_REQUEST,
+              "Invalid argument for gRPC operation",
+              grpcE.getMessage());
+        case NOT_FOUND:
+          return grpcResponse(
+              sc, Response.Status.NOT_FOUND, "Unknown gRPC operation", grpcE.getMessage());
+        case PERMISSION_DENIED:
+          return grpcResponse(
+              sc, Response.Status.UNAUTHORIZED, "Unauthorized gRPC operation", grpcE.getMessage());
+        case UNAUTHENTICATED:
+          return grpcResponse(
+              sc,
+              Response.Status.UNAUTHORIZED,
+              "Unauthenticated gRPC operation",
+              grpcE.getMessage());
+        case UNIMPLEMENTED:
+          return grpcResponse(
+              sc,
+              Response.Status.NOT_IMPLEMENTED,
+              "Unimplemented gRPC operation",
+              grpcE.getMessage());
+
+          // And then codes we ... don't know how to handle
+        case OK: // huh?
+
+        case CANCELLED:
+        case UNKNOWN:
+        case DEADLINE_EXCEEDED:
+        case ALREADY_EXISTS:
+        case RESOURCE_EXHAUSTED:
+        case FAILED_PRECONDITION:
+        case ABORTED:
+        case OUT_OF_RANGE:
+        case INTERNAL:
+        case UNAVAILABLE:
+        case DATA_LOSS:
+          break;
+      }
+      return grpcResponse(
+          sc, Response.Status.INTERNAL_SERVER_ERROR, "Unhandled gRPC failure", grpcE.getMessage());
+
+    } catch (NotFoundException nfe) { // too common, don't log
       return Response.status(Response.Status.NOT_FOUND)
           .entity(
               new RestServiceError(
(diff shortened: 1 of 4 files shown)


sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RequestHandler.java
```diff
@@ -0,0 +1,138 @@
...
+              "Unauthenticated gRPC operation",
+              grpcE.getMessage());
+        case UNIMPLEMENTED:
+          return grpcResponse(
+              sc,
+              Response.Status.NOT_IMPLEMENTED,
+              "Unimplemented gRPC operation",
+              grpcE.getMessage());
+
+          // And then codes we ... don't know how to handle
```
ivansenic, line 64: https://developers.google.com/maps-booking/reference/grpc-api/status_codes
https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md

This should help..

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResource.java
```diff
@@ -57,6 +60,65 @@
...
+      @ApiParam(value = "Keys to sort by") @QueryParam("sort") final String sort,
+      @Context HttpServletRequest request) {
+    if (isAuthTokenInvalid(token)) {
+      return invalidTokenFailure();
+    }
+    // 10-Nov-2021, tatu: Can not implement quite yet due to lack of schema access to bind
+    //     "primaryKey" segments to actual columns.
+    return Sgv2RequestHandler.handle(
+        () -> {
+          return javax.ws.rs.core.Response.status(Response.Status.NOT_IMPLEMENTED).build();
```
ivansenic, line 118: class already imported no need for FQN?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResource.java
```diff
@@ -117,27 +181,96 @@
...
+
+          String pageStateStr = null;
+          BytesValue pagingStateOut = rs.getPagingState();
+          if (pagingStateOut.isInitialized()) {
+            ByteString rawPS = pagingStateOut.getValue();
+            if (!rawPS.isEmpty()) {
+              byte[] b = rawPS.toByteArray();
+              pageStateStr = Base64.getEncoder().encodeToString(b);
+            }
+          }
```
ivansenic, line 201: should we create a utility for this, could be used in lot of places? Actually we should have this already in the old code for sure?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResource.java
```diff
@@ -117,27 +181,96 @@
...
+    if (isAuthTokenInvalid(token)) {
+      return invalidTokenFailure();
     }
+    final String cql = buildAddRowCQL(keyspaceName, tableName, payload);
+    logger.info("createRow(): try to call backend with CQL of '{}'", cql);
 
-    Sgv2RowsResponse response = new Sgv2RowsResponse(count, pageStateStr, convertRows(rs));
-    return javax.ws.rs.core.Response.status(Response.Status.OK).entity(response).build();
+    StargateGrpc.StargateBlockingStub blockingStub =
+        grpcFactory.constructBlockingStub().withCallCredentials(new StargateBearerToken(token));
```
ivansenic, line 254: how much is this costing and should this be in some kind of cache?

ivansenic (commented):
Just run through it.. Left some comments..

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RequestHandler.java
tatu-at-datastax: Thank you!

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResource.java
tatu-at-datastax: Agreed needs to be extracted. The challenge with old code tends to be dependencies, i.e. where exactly it is used; I can see how easily it could be reused.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResource.java
tatu-at-datastax: That would probably be one of the first optimization once we can measure performance, when we have end-to-end functionality (all operations for at least one successful test). Same goes for async handling.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResource.java
ivansenic: just for reference I was pointing to this -> `io.stargate.core.util.ByteBufferUtils#toBase64`

tatu-at-datastax: I'm thinking of maybe adding `byte[]` taking argument; and should really optimize this:

```
    if (base64.chars().anyMatch(c -> c == '/' || c == '+' || c == ' ')) {
```

since that is painfully slow way to check anything (I investigated it a bit earlier https://cowtowncoder.medium.com/measuring-string-indexofany-string-performance-java-fecb9eb473fa -- could use "commons-lang" variant or switch.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResource.java
tatu-at-datastax: Changed to non-FQN.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResource.java
tatu-at-datastax: Filed #1404.

ivansenic: where do we have this actually?

tatu-at-datastax: `ByteBufferUtils`

PR #1403, opened 2021-11-11
gRPC: Add operation to stream schema changes

olim7t:
**What this PR does**:
Add gRPC operation to stream schema changes.

N/A

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/grpc-proto/proto/stargate.proto
+++ b/grpc-proto/proto/stargate.proto
@@ -42,6 +42,6 @@ service Stargate {
   rpc DescribeTable(DescribeTableQuery) returns (CqlTable) {}
 
   // Requests notification of schema changes.
-  rpc getSchemaNotifications(GetSchemaNotificationsParams) returns (stream SchemaNotification) {}
+  rpc GetSchemaNotifications(GetSchemaNotificationsParams) returns (stream SchemaNotification) {}
 }
 
--- a/grpc-proto/proto/stargate.proto
+++ b/grpc-proto/proto/stargate.proto
@@ -42,6 +42,6 @@ service Stargate {
   rpc DescribeTable(DescribeTableQuery) returns (CqlTable) {}
 
   // Requests notification of schema changes.
-  rpc getSchemaChanges(GetSchemaChangeParams) returns (stream SchemaChange) {}
+  rpc getSchemaNotifications(GetSchemaNotificationsParams) returns (stream SchemaNotification) {}
 }
 
--- a/grpc-proto/proto/stargate.proto
+++ b/grpc-proto/proto/stargate.proto
@@ -20,7 +20,6 @@ option go_package = "stargate/proto";
 
 package stargate;
 
-import "google/protobuf/empty.proto";
 import "query.proto";
 import "schema.proto";
 
@@ -43,6 +42,6 @@ service Stargate {
   rpc DescribeTable(DescribeTableQuery) returns (CqlTable) {}
 
   // Requests notification of schema changes.
-  rpc getSchemaChanges(google.protobuf.Empty) returns (stream SchemaChange) {}
+  rpc getSchemaChanges(GetSchemaChangeParams) returns (stream SchemaChange) {}
 }
 
--- a/grpc-proto/proto/stargate.proto
+++ b/grpc-proto/proto/stargate.proto
@@ -20,6 +20,7 @@ option go_package = "stargate/proto";
 
 package stargate;
 
+import "google/protobuf/empty.proto";
 import "query.proto";
 import "schema.proto";
 
@@ -40,5 +41,8 @@ service Stargate {
 
   // Similar to CQL "DESCRIBE KEYSPACE"
   rpc DescribeTable(DescribeTableQuery) returns (CqlTable) {}
+
+  // Requests notification of schema changes.
+  rpc getSchemaChanges(google.protobuf.Empty) returns (stream SchemaChange) {}
 }
 
--- a/grpc/src/main/java/io/stargate/grpc/service/GrpcService.java
+++ b/grpc/src/main/java/io/stargate/grpc/service/GrpcService.java
@@ -21,7 +21,6 @@ import io.stargate.db.Persistence;
 import io.stargate.db.Persistence.Connection;
 import io.stargate.db.Result;
 import io.stargate.db.query.TypedValue;
-import io.stargate.proto.QueryOuterClass;
 import io.stargate.proto.QueryOuterClass.Batch;
 import io.stargate.proto.QueryOuterClass.Query;
 import io.stargate.proto.QueryOuterClass.Response;
@@ -31,6 +30,8 @@ import io.stargate.proto.Schema.CqlTable;
 import io.stargate.proto.Schema.CqlTableCreate;
 import io.stargate.proto.Schema.DescribeKeyspaceQuery;
 import io.stargate.proto.Schema.DescribeTableQuery;
+import io.stargate.proto.Schema.GetSchemaNotificationsParams;
+import io.stargate.proto.Schema.SchemaNotification;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.ScheduledExecutorService;
@@ -110,10 +111,9 @@ public class GrpcService extends io.stargate.proto.StargateGrpc.StargateImplBase
   }
 
   @Override
-  public void getSchemaChanges(
-      QueryOuterClass.GetSchemaChangeParams ignored,
-      StreamObserver<QueryOuterClass.SchemaChange> responseObserver) {
-    new SchemaChangesHandler(persistence, responseObserver).handle();
+  public void getSchemaNotifications(
+      GetSchemaNotificationsParams request, StreamObserver<SchemaNotification> responseObserver) {
+    new SchemaNotificationsHandler(persistence, responseObserver).handle();
   }
 
   static class ResponseAndTraceId {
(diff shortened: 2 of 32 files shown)


grpc-proto/proto/stargate.proto
```diff
@@ -29,5 +30,8 @@ service Stargate {
 
   // Executes a batch of CQL queries.
   rpc ExecuteBatch(Batch) returns (Response) {}
+
+  // Requests notification of schema changes.
+  rpc getSchemaChanges(google.protobuf.Empty) returns (stream SchemaChange) {}
```
olim7t, line 35: We might want to use a custom message here. Even if it's still empty now, that leaves us a way to add parameters in the future if needed.

testing/src/main/java/io/stargate/it/grpc/SchemaChangesTest.java
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  @Test
+  @DisplayName("Should receive table changes")
+  public void tableChangesTest(CqlSession session, @TestKeyspace CqlIdentifier keyspace) {
+    SchemaChangeAccumulator changes = new SchemaChangeAccumulator();
+    asyncStub.getSchemaChanges(Empty.newBuilder().build(), changes);
+
+    // TODO why does this not generate an event?
+    session.execute("CREATE TABLE foo(k int PRIMARY KEY)");
```
olim7t, line 49: This is weird. I've tried with UDTs as well, and same thing, CREATE does not generate a notification but DROP does.

olim7t (commented):
Work in progress, I'm still experimenting with the idea.

The server never completes the request, it's up to the client to cancel it when it's done.

testing/src/main/java/io/stargate/it/grpc/SchemaChangesTest.java
ivansenic: because the listener is not yet registered once you do the `execute`, you call the async api with the GRPC and return thus immediately, it's not guaranteed that the execute will be done after the `persistence.registerEventListener(this);` is called in the `SchemaChangesHandler`

jeffreyscarpenter:
Should we target this for v2.0.0 branch instead of master?

testing/src/main/java/io/stargate/it/grpc/SchemaChangesTest.java
olim7t: Good catch! Pausing after line 46 does indeed fix the issue.

olim7t:
> Should we target this for v2.0.0 branch instead of master?

It can't go on either side. But since we've been merging master->v2.0.0 so far, I think it's better to target master for things that are going to be on both sides.

jeffreyscarpenter:
> > Should we target this for v2.0.0 branch instead of master?
> 
> It can't go on either side. But since we've been merging master->v2.0.0 so far, I think it's better to target master for things that are going to be on both sides.

I thought we had not decided for sure on adding schema operations to public gRPC yet, which is why we were doing this schema work in v2.0.0 first.

grpc-proto/proto/stargate.proto
olim7t: Done.

grpc/src/main/java/io/stargate/grpc/service/GrpcService.java
```diff
@@ -70,6 +72,13 @@ public void executeBatch(Batch batch, StreamObserver<Response> responseObserver)
     new BatchHandler(batch, CONNECTION_KEY.get(), persistence, responseObserver).handle();
   }
 
+  @Override
+  public void getSchemaChanges(
+      Empty request, StreamObserver<QueryOuterClass.SchemaChange> responseObserver) {
+    // TODO we're not using the connection, modify the interceptor so that it doesn't create it
```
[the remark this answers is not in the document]
olim7t: I'm going to take a closer look at this.
Event listeners are registered with the Persistence directly, there's no need to open a connection.

If tenants is what you have in mind, I'm not sure if we need to filter here. Event handling is generally a VM-wide concern, so it spans multiple tenant contexts. Therefore what we need is an "internal" view that notifies for all tenants. At least that's how it works in the GraphQL cache, our only existing use case so far.

grpc/src/main/java/io/stargate/grpc/service/interceptors/NewConnectionInterceptor.java
```diff
@@ -83,13 +83,14 @@ public NewConnectionInterceptor(
...
               .remoteAddress(call.getAttributes().get(Grpc.TRANSPORT_ATTR_REMOTE_ADDR))
               .build();
+      Context context = Context.current().withValue(GrpcService.HEADERS_KEY, stringHeaders);
 
-      Connection connection = connectionCache.get(info);
+      if (!"stargate.Stargate/getSchemaChanges"
+          .equals(call.getMethodDescriptor().getFullMethodName())) {
+        Connection connection = connectionCache.get(info);
+        context = context.withValue(GrpcService.CONNECTION_KEY, connection);
+      }
```
olim7t, line 92: This is not super elegant...
Maybe creating the connection doesn't matter that much, since connections are cached and the same client is likely to invoke other methods than this one.

grpc/src/main/java/io/stargate/grpc/service/interceptors/NewConnectionInterceptor.java
```diff
@@ -83,13 +83,14 @@ public NewConnectionInterceptor(
               .headers(stringHeaders)
               .remoteAddress(call.getAttributes().get(Grpc.TRANSPORT_ATTR_REMOTE_ADDR))
               .build();
+      Context context = Context.current().withValue(GrpcService.HEADERS_KEY, stringHeaders);
 
-      Connection connection = connectionCache.get(info);
+      if (!"stargate.Stargate/getSchemaChanges"
```
tomekl007, line 88: why do we need this special condition?

olim7t: That operation interacts with the `Persistence` directly, it doesn't need the `Connnection`. So the idea was to avoid creating it.
But maybe that's overkill (see my other comment).

jeffreyscarpenter (approved):

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/InitReplayingStreamObserver.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * A stream observer that starts its life accumulating incoming elements; it only starts streaming
+ * them (via {@link #onNextReady(Object)} once {@link #markReady()} is called.
+ */
+public abstract class InitReplayingStreamObserver<T> implements StreamObserver<T> {
+
+  // This is based on the "queue-drain" pattern
+  // https://github.com/ReactiveX/RxJava/wiki/Implementing-custom-operators-(draft)#the-queue-drain-approach
+  private final Queue<T> elements = new ConcurrentLinkedQueue<>();
+  private final AtomicReference<ObserverState> state = new AtomicReference<>(ObserverState.INITIAL);
```
olim7t, line 32: This is complicated but I can't think of a simpler solution.

I suspect calls to `onNext()` are always happening from the same Netty event loop, but `markReady()` definitely isn't, so we need proper synchronization.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/schema/SchemaCache.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    @Override
+    public void onCompleted() {
+      LOG.error("Notification stream completed unexpectedly, trying to reconnect");
+      tryReconnect();
+    }
+
+    private void tryReconnect() {
+      refreshKeyspaces(false)
```
olim7t, line 167: Unlike `init()`, we don't want to block here, we're on a gRPC thread.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/schema/SchemaCache.java
```diff
@@ -0,0 +1,178 @@
...
+          .exceptionally(
+              t -> {
+                LOG.error(
+                    "Error while trying to reconnect to notification stream, "
+                        + "the schema cache will stop updating",
+                    t);
+                return null;
```
olim7t, line 174: I'm not sure what else to do here. Maybe some kind of periodic retry, but we don't have access to an executor.

grpc/src/main/java/io/stargate/grpc/service/GrpcService.java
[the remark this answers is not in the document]
tatu-at-datastax: We probably cannot really afford listener-per-tenant so it should be global one.

One possible solution is that update would only be used to invalidate entry (regardless of tenant), but lookups would be on behalf of tenant when needed (not part of notification handling).

tatu-at-datastax:
I think I'd like to get this merged. The only (?) question I have is wrt name handling of keyspaces, tenant ids, but that probably needs to be discussed outside this PR.

olim7t:
> The only (?) question I have is wrt name handling of keyspaces, tenant ids, but that probably needs to be discussed outside this PR.

Agreed. The current implementation will return "decorated" keyspace names in a multi-tenant environment, we'll need to decide if we process them here or leave that to the caller.

tatu-at-datastax:
> > The only (?) question I have is wrt name handling of keyspaces, tenant ids, but that probably needs to be discussed outside this PR.
> 
> Agreed. The current implementation will return "decorated" keyspace names in a multi-tenant environment, we'll need to decide if we process them here or leave that to the caller.

I like the idea of scoped identifier where unmodified `name` is returned along with string `tenant id`.
Probably best done with follow-up work so it does not block merging, and since we don't quite know how this would interact with multi-tenant backends.

PR #1405, opened 2021-11-12
Fix #1404: add/expose "byte[]" methods; remove legacy (broken) base64 handling workaround

### LINKED ISSUE #1404: Add `byte[]` taking variant(s) in `ByteBufferUtils` for Base64 handling
Since some use cases (esp. ones to do with gRPC messages, see #1402) need to work on `byte[]`, not `ByteBuffer`; and since `ByteBufferUtils` already internally converts things to `byte[]` as necessary, it'd make sense to expose that functionality.

In addition, use of Stream operations for character checking, a la:

```
if (base64.chars().anyMatch(c -> c == '/' || c == '+' || c == ' ')) {
``` 

is rather inefficient (see https://cowtowncoder.medium.com/measuring-string-indexofany-string-performance-java-fecb9eb473fa ), it'd make sense to optimize this case.

tatu-at-datastax:
**What this PR does**:

* Adds and exposes 2 methods that accept `byte[]` (over just `ByteBuffer`) since functionality exists and is useful for use cases that do not use `ByteBuffer`s
* Optimizes check for legacy URLs (used heavily for PagingState handling) (see issue for details -- significant perf penalty for existing Streams-based impl)

Fixes #1404 

**Checklist**
- [x] Changes manually tested -- verified that commenting out parts of new check would fail tests
- [x] Automated Tests added/updated -- relies on existing unit tests

### DIFF
--- a/core/src/main/java/io/stargate/core/util/ByteBufferUtils.java
+++ b/core/src/main/java/io/stargate/core/util/ByteBufferUtils.java
@@ -44,30 +44,15 @@ public class ByteBufferUtils {
   }
 
   public static ByteBuffer fromBase64UrlParam(String base64) {
-    // TODO: remove support for legacy use cases when they are no longer relevant
-    if (isLegacyEncodedBase64String(base64)) {
-      // Fix legacy strings that got broken by decoding `+` at HTTP level
-      base64 = base64.replace(' ', '+');
-      // Use the decoder compatible with the encoder previously used for URL params
-      return ByteBuffer.wrap(Base64.getDecoder().decode(base64));
-    }
-
+    // As per PR #1405 (and earlier issue #1041), special handling no longer needed
+    /*    if (base64.chars().anyMatch(c -> c == '/' || c == '+' || c == ' ')) {
+    Fix legacy strings that got broken by decoding `+` at HTTP level
+         base64 = base64.replace(' ', '+');
+         // Use the decoder compatible with the encoder previously used for URL params
+         return ByteBuffer.wrap(Base64.getDecoder().decode(base64));
+       }
+    */
     return ByteBuffer.wrap(Base64.getUrlDecoder().decode(base64));
-  }
-
-  private static boolean isLegacyEncodedBase64String(String base64) {
-    // From
-    // https://cowtowncoder.medium.com/measuring-string-indexofany-string-performance-java-fecb9eb473fa
-    // use the fastest (and simple) method here (since commons-lang3 not yet a dependency)
-    for (int i = 0, len = base64.length(); i < len; ++i) {
-      switch (base64.charAt(i)) {
-        case '/':
-        case '+':
-        case ' ':
-          return true;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/stargate/core/util/ByteBufferUtils.java
```diff
@@ -47,6 +55,21 @@ public static ByteBuffer fromBase64UrlParam(String base64) {
     return ByteBuffer.wrap(Base64.getUrlDecoder().decode(base64));
   }
 
+  private static boolean isLegacyEncodedBase64String(String base64) {
+    // From
+    // https://cowtowncoder.medium.com/measuring-string-indexofany-string-performance-java-fecb9eb473fa
```
tomekl007, line 60: since we are re-implementing it, we should add a unit test for this new implementation.

core/src/main/java/io/stargate/core/util/ByteBufferUtils.java
```diff
@@ -24,20 +24,28 @@
...
-    return Base64.getUrlEncoder().encodeToString(getArray(buffer));
+    return toBase64ForUrl(getArray(buffer));
+  }
+
+  public static String toBase64ForUrl(byte[] bytes) {
+    return Base64.getUrlEncoder().encodeToString(bytes);
   }
 
   public static ByteBuffer fromBase64UrlParam(String base64) {
     // TODO: remove support for legacy use cases when they are no longer relevant
```
dougwettlaufer, line 47: We can take care of this TODO now. The legacy case is no longer needed. It just needed to handle both while we were rolling out the change

https://github.com/stargate/stargate/pull/1041#discussion_r654583476

core/src/main/java/io/stargate/core/util/ByteBufferUtils.java
tatu-at-datastax: The logic is covered by existing unit tests; I verified that the legacy case is covered as well as regular one.

core/src/main/java/io/stargate/core/util/ByteBufferUtils.java
tatu-at-datastax: Ah. In that case I will simplify this and remove specialized handling!

tatu-at-datastax:
@dougwettlaufer @tomekl007 As per Doug's comment, disabled special handling. Not sure whether to remove all traces or leave comments (as in PR now); I'm fine either way.

tatu-at-datastax:
@dougwettlaufer any concerns on merging without legacy handling?

dougwettlaufer:
> @dougwettlaufer any concerns on merging without legacy handling?

Works for me. Although my vote would be to delete rather than comment but I'm not blocking this PR.

tatu-at-datastax:
@dougwettlaufer Sounds good: I'll try to remember to delete these after deployment.

PR #1407, opened 2021-11-13
WIP: use placeholder-INSERT for "addRow()" instead of as-JSON

tatu-at-datastax:
**What this PR does**:

Changes kind of CQL query send to gRPC for adding row: used "as JSON" binding; now slices values from JSON payload into specific columns.
Intent was to improve handling -- JSON payload only works for scalars, but not Lists, Maps etc -- but turns out it won't really improve situation much immediately as specific values do not convert (UUID not coerced from String, unlike with JSON).

But we can work on conversion on gRPC side most likely; or, if and when we get schema-awareness use that.

No issue filed.

**Checklist**
- [x] Changes manually tested -- use existing Integration tests (for v2.0.0)

### DIFF
--- a/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/JsonToProtoValueConverter.java
+++ b/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/JsonToProtoValueConverter.java
@@ -2,14 +2,17 @@ package io.stargate.sgv2.restsvc.grpc;
 
 import io.stargate.grpc.Values;
 import io.stargate.proto.QueryOuterClass;
+import java.math.BigDecimal;
 import java.math.BigInteger;
 
+/**
+ * Helper class for converting "natural" Java object types (such as ones bound by Jackson when
+ * target is {@code java.lang.Object}) into matching gRPC {@link QueryOuterClass.Value} types.
+ *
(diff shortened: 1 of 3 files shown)


sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/JsonToProtoValueConverter.java
```diff
@@ -0,0 +1,38 @@
...
+    if (javaValue instanceof Number) {
+      if (javaValue instanceof Long) {
+        return Values.of((Long) javaValue);
+      }
+      if (javaValue instanceof Integer) {
+        return Values.of((Integer) javaValue);
+      }
+      if (javaValue instanceof BigInteger) {
+        return Values.of((BigInteger) javaValue);
+      }
```
tomekl007, line 31: what about `Float`?

tatu-at-datastax: Yes, for sake of completeness I'll need to add `Double`, `Float` and `BigDecimal` too. thanks!

tatu-at-datastax:
While not optimal for now, improvement and needed as a base for schema-based handling so merging as is (with minor improvements added as suggested by Tomasz).

PR #1411, opened 2021-11-16
Implement gRPC createTable operation

olim7t:


grpc/pom.xml
```diff
@@ -50,6 +50,11 @@
       <version>${grpc.version}</version>
     </dependency>
     <!-- Other -->
+    <dependency>
+      <groupId>com.datastax.oss</groupId>
+      <artifactId>java-driver-query-builder</artifactId>
+      <version>4.13.0</version>
+    </dependency>
```
olim7t, line 57: I don't like this. That query builder is tied to the driver, in fact it pulls a dependency to it.
I'm going to try to use our internal query builder instead.

grpc/src/main/java/io/stargate/grpc/service/CreateTableHandler.java
```diff
@@ -0,0 +1,225 @@
...
+          (DefaultCreateTable)
+              builder.withColumn(columnProto.getName(), convertType(columnProto.getType()));
+    }
+    for (Map.Entry<String, Schema.ColumnOrderBy> entry :
+        tableProto.getClusteringOrdersMap().entrySet()) {
+      builder =
+          (DefaultCreateTable)
+              builder.withClusteringOrder(entry.getKey(), convertOrder(entry.getValue()));
+    }
+    // TODO handle table options
```
olim7t, line 116: The driver's query builder uses strongly-typed options. If we pass in strings, they will get quoted.

grpc/src/main/java/io/stargate/grpc/service/CreateTableHandler.java
```diff
@@ -0,0 +1,245 @@
...
+          ImmutableColumn.create(
+              columnProto.getName(), Column.Kind.Regular, convertType(columnProto.getType())));
+    }
+
+    return queryBuilder
+        .create()
+        .table(keyspaceName, tableProto.getName())
+        .ifNotExists(createProto.getIfNotExists())
+        .column(columns)
+        // TODO handle table options
```
olim7t, line 122: If we want to support generic `string => string` options, we'll need to add that capability to `QueryBuilder`.

But if we aim for a similar API as v1, the only option that's exposed so far is the default TTL, so we could go with that and just have a `defaultTimeToLIve` parameter in the gRPC message.

grpc-proto/proto/schema.proto
```diff
@@ -49,12 +49,13 @@ message CqlTable {
...
   repeated ColumnSpec staticColumns = 5;
   repeated ColumnSpec columns = 6;
-  map <string, ColumnOrderBy> clusteringKeyColumnSortOrders = 7;
+  map <string, ColumnOrderBy> clusteringOrders = 7;
 }
 
 message CqlTableCreate {
-  CqlTable table = 1;
-  bool ifNotExists = 2;
+  string keyspaceName = 1;
```
tatu-at-datastax, line 56: This is backwards-incompatible change (changing of field ids) and we probably do not/cannot make it.
So maybe add `keyspaceName` as `3` (the usual way to add things)?

grpc/src/main/java/io/stargate/grpc/service/CreateTableHandler.java
jeffreyscarpenter: I think we should treat table options (and keyspace options) as map<string, string> on the gRPC API so that the values can be obtained for the sake of completeness, in case there is a need to allow API consumers to obtain the full definitions of keyspaces and tables. 

However, I doubt there are many cases where the API clients will need to process the actual values, so I'm hesitant to put the effort into modeling them in the gRPC API. This seemed like a good compromise approach to me.

grpc-proto/proto/schema.proto
jeffreyscarpenter: This is a proto definition that has only ever existed on this branch and has never been released. I don't think we need to be that strict in this case.

olim7t: We just added that message in v2, I don't think we need to preserve backward compatibility yet.

jeffreyscarpenter (approved):
No specific comments on source files, but a question. Currently the gRPC `executeQuery` operation supports execution of DDL statements. I assume we are going to continue to allow that, but wondering if we should be routing those through the various subclasses of `DdlMessageHandler` to ensure they are handled consistently?

grpc-proto/proto/schema.proto
tatu-at-datastax: Ah... Ok sorry, did not realize it was change to what we added. That's fine then.

PR #1413, opened 2021-11-17
Adding handling of indexes and materialized views to gRPC describe operations

jeffreyscarpenter:


grpc-proto/proto/schema.proto
```diff
@@ -40,6 +38,27 @@ enum ColumnOrderBy {
...
+  string columnName = 2;
+  bool custom = 3;
+  string customType = 4;
+  map<string, string> options = 5;
+}
+
+// MATERIALIZED VIEWS
+
+message CqlMaterializedView {
+  string name = 1;
```
olim7t, line 54: Some other metadata from `system_schema.views` that might be worth exposing here:
- name of the base table
- the custom WHERE clause

I noticed that the persistence API does not expose them so we'd have to add them there too.

grpc-proto/proto/schema.proto
```diff
@@ -40,6 +38,27 @@ enum ColumnOrderBy {
   DESC = 0x01;
 }
 
+// INDEXES
+
+message CqlIndex {
+  string name = 1;
+  string columnName = 2;
+  bool custom = 3;
+  string customType = 4;
```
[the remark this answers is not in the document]
tatu-at-datastax: Is this really necessary? Leaving String value out should be fine and value would be left as `null` (and there is also method for checking if value was set).
I am not very familiar with use "XxxValue" although have seen it in places so I assume there are legit reasons for using it in some cases.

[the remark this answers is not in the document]
olim7t: `string` produces an empty string when the value is missing, whereas `StringValue` has the notion of "absent".

https://stackoverflow.com/questions/51707877/what-is-the-point-of-google-protobuf-stringvalue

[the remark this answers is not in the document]
tatu-at-datastax: I think that only applies to Go tho -- in Java, `String`s won't be "" but plain `null`. But since Go has no concept of `null` (at least not as universal value like Java) some default value must be used, hence "" on that platform.
Even with Go I think there is access for "isXxxSet()" if someone does have to distinguish Not-passed (absent) from "".

But I guess it's a valid point as we are not counting on only Java being used for implementation and might be more convenient for some languages. Even if possible extra wrapping for Java.

grpc-proto/proto/schema.proto
```diff
@@ -25,8 +25,6 @@ package stargate;
 
 // TYPES
 
-// TABLES
-
 message UserDefinedTypeCreate {
   string keyspaceName = 1;
```
olim7t, line 29: I missed it in earlier reviews, but gRPC conventions are snake case for fields, as in `keyspace_name`. That's what we've been doing in `query.proto` so we should stay consistent.

grpc-proto/proto/schema.proto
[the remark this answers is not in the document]
olim7t: > in Java, `String`s won't be "" but plain `null`

Are you sure? I haven't tried decoding a message, but `newBuilder.build()` gives me an empty string, so I assumed that was the default.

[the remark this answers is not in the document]
tatu-at-datastax: I was thinking more the case of receiving side where I think undefined would be received as `null`, and for Proto v2 definitely was. But I must admit I do not know how builder does that so you may be right. Proto v3 seems to have changed a lot.

grpc-proto/proto/schema.proto
tatu-at-datastax: Agreed.

jeffreyscarpenter: will do

grpc-proto/proto/schema.proto
jeffreyscarpenter: with the current design, you can only access the MVs from the base table, so you can get the base table name from there.

For custom WHERE clause, this falls into the category of whether we are trying to provide the equivalent of the cqlsh `DESCRIBE` such that the returned schema has the full information required to recreate the schema.

grpc-proto/proto/schema.proto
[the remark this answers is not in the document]
jeffreyscarpenter: fixed

grpc-proto/proto/schema.proto
jeffreyscarpenter: fixed

grpc-proto/proto/schema.proto
olim7t: > whether we are trying to provide the equivalent of the cqlsh `DESCRIBE` such that the returned schema has the full information required to recreate the schema.

Right, that hasn't been a use case for our persistence API so for, so it's not top priority.

PR #1416, opened 2021-11-18
closes #1220: fixed document store when having empty string json keys

### LINKED ISSUE #1220: DocsApi allows for creating json with value and empty key
Was able to create a new document like the following

```json
{"key1":"foo","":"Goofy","key2":"bar"}
``` 

Then when trying to retrieve all documents in the collection a 500 was returned.

ivansenic:
**What this PR does**:
Fixes the #1220, the empty key is valid in JSON. However, we were not storing this correctly and we had `leaf=null`, which would lead to the errors when getting the document. The `JsonConverter` also needed to be updated as it was not counting for this situation. `JsonConverterTest` was also wrong, as leaves were not correctly set in two rows.

Note that atm search with wildcard will not match the empty keys. This will be fixed in a separate issue and PR. Problem is that once we have a wildcard we are doing a search with `pX > ''` which in this case would not match the row with empty string key.

**@EricBorczuk Although this looks small, please let me know if there are any other implications as this is heavily affecting the current doc saving structure. Maybe I should have gone for the empty key marker better?**

Fixes #1220

**Checklist**
- [x] Changes manually tested
- [ ] Create follow-up issue  
- [x] ~Documentation added/updated~

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/DocsShredder.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/DocsShredder.java
@@ -12,6 +12,7 @@ import io.stargate.web.docsapi.service.util.DocsApiUtils;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import javax.inject.Inject;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.jsfr.json.JsonSurfer;
@@ -173,10 +174,15 @@ public class DocsShredder {
       String key,
       boolean patching,
       boolean numericBooleans) {
+    // the converter can not handle empty field names, thus add manually
     String bracketedPath = DocsApiUtils.convertJsonToBracketedPath(parsingContext.getJsonPath());
-    Map<String, Object> bindMap = DocsApiUtils.newBindMap(path, config.getMaxDepth());
+    if (Objects.equals("", parsingContext.getCurrentFieldName())) {
+      bracketedPath += "['']";
+    }
 
+    Map<String, Object> bindMap = DocsApiUtils.newBindMap(path, config.getMaxDepth());
     bindMap.put("key", key);
+
     String leaf = null;
     int i = path.size();
 
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverter.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/JsonConverter.java
@@ -20,6 +20,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import javax.inject.Inject;
 
 public class JsonConverter {
@@ -97,7 +98,7 @@ public class JsonConverter {
       for (int i = 0; i < maxDepth; i++) {
         String p = row.getString("p" + i);
         String nextP = i < maxDepth - 1 ? row.getString("p" + (i + 1)) : "";
-        boolean endOfPath = nextP.equals("");
+        boolean endOfPath = nextP.equals("") && Objects.equals(p, rowLeaf);
         boolean isArray = p.startsWith("[");
         boolean nextIsArray = nextP.startsWith("[");
 
--- a/restapi/src/test/java/io/stargate/web/docsapi/service/JsonConverterTest.java
+++ b/restapi/src/test/java/io/stargate/web/docsapi/service/JsonConverterTest.java
@@ -453,7 +453,7 @@ public class JsonConverterTest {
     data2.put("p2", "");
     data2.put("p3", "");
     data2.put("text_value", "initial");
-    data2.put("leaf", "a");
+    data2.put("leaf", "[0]");
 
     Map<String, Object> data3 = new HashMap<>();
 
@@ -464,7 +464,7 @@ public class JsonConverterTest {
     data3.put("p2", "[0]");
     data3.put("dbl_value", 1.23);
     data3.put("p3", "");
-    data3.put("leaf", "a");
+    data3.put("leaf", "[0]");
 
     Map<String, Object> data4 = new HashMap<>();
 
--- a/testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
+++ b/testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
@@ -760,6 +760,17 @@ public abstract class BaseDocumentApiV2Test extends BaseIntegrationTest {
   }
 
   @Test
+  public void testJsonEmptyKey() throws IOException {
+    String json = "{\"\": \"value\", \"nested\": {\"\": \"value\"}}";
+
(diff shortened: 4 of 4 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/DocsShredder.java
```diff
@@ -173,10 +174,15 @@ private Object convertToBackendBooleanValue(boolean value, boolean numericBoolea
       String key,
       boolean patching,
       boolean numericBooleans) {
+    // the converter can not handle empty field names, thus add manually
     String bracketedPath = DocsApiUtils.convertJsonToBracketedPath(parsingContext.getJsonPath());
```
EricBorczuk, line 178: What happens if there is an empty string that isn't at the leaf-most field name? Such as `{"": {"a": "b"}}`?

It seems that line 180 handles the leaf-most fields but not empty strings that are elsewhere in the document, is that handled properly by `convertJsonToBracketedPath`? (also, I believe there are unit tests for `convertJsonToBracketedPath`, if you could add a test for this)

EricBorczuk (changes requested):
Overall I think this looks good! I asked one question about `convertJsonToBracketedPath`, I think there's a case that isn't covered?

At any rate, I think we might run into issues representing empty-string here because of ambiguity in representation: the value for both "there is a field here that is empty-string" and "there is no field here" are both `''` and that is likely going to be a problem if/when you want to have wildcards include the empty string, etc. Also I believe it is not allowed to set the `pX` values to `null` because they are in the clustering keys/PK. One option might be to represent a true empty string as some other value, like a UUID (I did this with value nodes that are empty object and empty array) - that might work to disambiguate, but there'd be some logic that has to be written to convert empty-string to that UUID as you write a Doc and vice-versa as you read it back out.

To take a step back - should we definitely be supporting this? It technically abides by the JSON spec to have empty string fields, but in practice will people be using this feature (other than by mistake?) My gut is that most of the time if the API rejected a request with an empty string field, the user would be thankful for it...

restapi/src/main/java/io/stargate/web/docsapi/service/DocsShredder.java
ivansenic: True I guess this is not covered, will check on Monday. :+1:

ivansenic: @EricBorczuk I tried fixing the above, however, it's actually impossible with the current approach using surfer.. It fails when there is double empty path and there is nothing we can do about it -> `{"": {"": "value}}`.

The further I looked, I was more confident that the `JsonSurfer` is actually not good fit for document shredding. It does a very bad job of defining the path you are in.. Converting that path to bracketed and then again to array is also very unintuitive. Was there any other idea that was consider here? Why didn't we simply traverse the `JsonNode` here and simply apply handling based on the node type. It would save us a lot of time as well that we spend in checking for different states and conditions?

@tatu-at-datastax any idea if there is a utility for traversing the JSON tree? We need something better than that `JsonSurfer`

EricBorczuk: JsonSurfer was used from the very beginning of the docs API (even before I started on it) - so I don't have a good reason for why it is used. More than likely we could toss it out in favor of a simple traversal of Jackson nodes...and I guess that would allow data like `{"": {"": "value"}}`

But also, to reiterate - is this actually what we want? If we allow storing of empty string (rather than erroring out), will users ever be using this seriously, or will they always be using it by mistake? In other words, what is the use-case for storing empty-string as a field name in JSON (I know it's valid JSON, in practice though it seems like people would only be doing this by mistake)

tatu-at-datastax: @ivansenic when I looked at code earlier, I had the feeling that JsonSurfer might indeed be sort of overkill; and also add some overhead we could avoid. This because it offers more advanced filtering, traversal, but ones that we either don't need, or can implement more efficiently to fit our exact use case.

I would probably operate directly on JsonNode and not try to find higher level abstractions because I think that our traversal is quite distinct so general-purpose tools would not be good matches.

ivansenic: @tatu-at-datastax @EricBorczuk Would you two agree on the following:

1. Do allow storing of empty keys as this is a valid JSON and thus we correctly serve the users?
2. Completely refactor the shredding and implement as Tatu suggested plain simple recursive `JsonNode` traversal? We would even have more dedicated testing possibilities then per node type.

tatu-at-datastax: On internal replacement for `JsonNode`, I think @ivansenic 's point on keeping track of path would be relevant: our own "StargateJsonValue" (or whatever) could hold on to path indicator(s) as needed; avoid need for String concatenation and so forth.

tatu-at-datastax: Agree on (2) as that could have multiple benefits, including simpler implementation.

I am bit ambivalent on (1) -- use of empty String as key is valid JSON and if we are confident we can make it work 100%, I'd be for it. My main concern is whether there are cases where we simply cannot make it work reliably, or that are there are ambiguous cases. I think you @ivansenic and @EricBorczuk are in better position to decide on this one.

EricBorczuk: The main issue I see with (1) is that there's no difference between a missing pN column value and one that is set to empty string because a user supplied it (i.e. we don't store `null` for missing clustering columns because that isn't allowed) - so in the next step when we want wild-card `*` matching to work with empty string fields (which internally adds a clause `WHERE pN > ''`), we won't really be able to get empty-string fields to match.

ivansenic: What about storing a fixed value, like we do for the root document and empty arrays/maps?

tatu-at-datastax: Do we store actual String values as-is, or JSON-escaped ones, in leaf value fields? If former, it may be difficult to find any placeholder value that could not be stored by users.

ivansenic: @tatu-at-datastax Yea we have those already, have a look here https://github.com/stargate/stargate/blob/7bfee16e3767646541e3cc3f982fc5775e1d2194/restapi/src/main/java/io/stargate/web/docsapi/service/query/DocsApiConstants.java#L46-L52

ivansenic:
Setting to blocked until we decide how to handle empty keys and until we re-implement the shredding. Tasks pending:

* https://github.com/stargate/stargate/pull/1439

ivansenic:
This can be fixed once https://github.com/stargate/stargate/pull/1617 is merged.

ivansenic:
Closing for now, as we will not provide a fix at the moment. PR to handle proper error message: https://github.com/stargate/stargate/pull/1663.

PR #1418, opened 2021-11-19
Add more create operations

olim7t:


grpc-proto/proto/schema.proto
```diff
@@ -74,7 +74,19 @@ message CqlMaterializedView {
...
+      uint32 replication_factor = 1;
+  }
+  message NetworkReplication {
+      map<string, uint32> replication_factors = 2;
+  }
+  oneof replication {
+    SimpleReplication simple_replication = 3;
+    NetworkReplication network_replication = 4;
+  }
+  bool durable_writes = 5;
```
olim7t, line 89: Just experimenting here, I wanted to see what it looks like if we stick to the options that we've been exposing so far through the APIs.
This makes the implementation much simpler since it matches exactly what the query builder offers.

grpc/src/main/java/io/stargate/grpc/service/CreateTableHandler.java
```diff
@@ -35,15 +36,15 @@
       CqlTableCreate createTable,
       Persistence.Connection connection,
       Persistence persistence,
-      QueryBuilder queryBuilder,
+      TypedValue.Codec valueCodec,
```
olim7t, line 39: The query builder has a static view of the schema, so it has to be built at the last moment.

grpc-proto/proto/schema.proto
tatu-at-datastax: I like that. If we want to leave something like "options", it could perhaps be named "extra_options" or something, to indicate it contains things not exposed through specific fields? Although at the same time if we do not expose anything else right now it may make more sense to just remove it -- no pre-mature extensibility as we do not know future and can not really know how new options are best exposed before knowing what they are.

grpc/src/main/java/io/stargate/grpc/service/CreateTableHandler.java
tatu-at-datastax: So basically `valueCodec` needs to be passed to allow us to construct `QueryBuilder` when needed?

grpc-proto/proto/schema.proto
jeffreyscarpenter: I had something like this in previously and removed it. My preference is to just keep keyspace and table options as map<string, string> since we don't have any evidence of any particlular options is needed by any clients. If they need them, they can parse them.

grpc/src/main/java/io/stargate/grpc/service/CreateTableHandler.java
olim7t: Yes.
We could rebuild it every time too, but that seems inefficient since it can be cached.

grpc-proto/proto/schema.proto
mpenick: +1 the concrete API. This looks really nice to use. I can almost never remember the syntax required for the replication settings and this alleviates that.

What do you think about using a message type for the replication factor in `NetworkReplication`?

```proto
message CqlKeyspace {
  // ...
  message NetworkReplication {
    message ReplicationFactor {
      string datacenter = 1; // This is more descriptive of what the parameter is.
      uint32 num_replicas = 2;
    }
    repeated ReplicationFactor replication_factors = 2;
  }
}
```

One issue is it doesn't enforce distinct values for the datacenter name.

mpenick: This can be made optional using `BoolValue`:
```proto
google.protobuf.BoolValue durable_writes = 5;
```

Otherwise, it would default to `false` instead of not provided.

mpenick: @jeffreyscarpenter 

> we don't have any evidence of any particlular options is needed by any clients.

This is true for tables, but I think that when creating a keyspace providing the replication settings is required.

olim7t:
Closing this now that we have removed the existing operations in #1561.

PR #1424, opened 2021-11-22
doing document population in parallel batches

ivansenic:
**What this PR does**:

Improves performance for the document population when searching. We need to fetch X number of documents at the end of the search, however, the `concatMap` operator is doing sequential pull of documents in the same thread, meaning that docs are loaded sequentially in fact. With this update we are buffering up to 5 documents and firing the population in parallel in different RX threads.

There are two other places where the `concatMap` is used, we need to check if there we can also benefit from parallel execution. However, this will be done in separate PR.

Same quick load test against DSE locally shows this improvement:

```
master
---
TotalCount=10000
Period(ms)=238957
Throughput(ops/sec)=41.85
Min=14884
Mean=214634.14
StdDev=94173.16
50.000ptile=227540
90.000ptile=328466
99.000ptile=407633
99.900ptile=435421
99.990ptile=465829
99.999ptile=471597
Max=471597

#pr-branch
TotalCount=10000
Period(ms)=118896
Throughput(ops/sec)=84.11
Min=16572
Mean=93782.68
StdDev=55451.12
50.000ptile=81985
90.000ptile=170131
99.000ptile=236322
99.900ptile=276037
99.990ptile=286785
99.999ptile=288096
Max=288096
``` 

-

**Checklist**
- [x] Changes manually tested


EricBorczuk:
I ran just the `or` test locally, and it seems to match the data you've said above! ~1.7x faster with your branch :)

EricBorczuk (approved):

PR #1428, opened 2021-11-22
Add a query parameter onto write APIs to allow assigning TTL

### LINKED ISSUE #1414: Enhancement request to include TTL for Documents in Astra DB
A client currently using Couchbase has indicated Astra DB is missing the Document Expiry feature. We like to suggest an enhancement request to include TTL for Documents in Astra DB.

Please refer to Couchbase documentation for details: https://docs.couchbase.com/server/current/eventing/eventing-examples-docexpiry.html

EricBorczuk:
**What this PR does**:
Allows assigning a TTL to a document.

Fixes #1414

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
@@ -131,16 +131,18 @@ public class DocumentDB {
         this.dataStore
             .queryBuilder()
             .select()
-            .function(ImmutableList.of(QueryBuilderImpl.FunctionCall.ttl("leaf")))
+            .function(
+                ImmutableList.of(
+                    QueryBuilderImpl.FunctionCall.ttl(DocsApiConstants.LEAF_COLUMN_NAME)))
             .from(namespace, collection)
-            .where(BuiltCondition.of("key", Predicate.EQ, docId))
+            .where(BuiltCondition.of(DocsApiConstants.KEY_COLUMN_NAME, Predicate.EQ, docId))
             .build()
             .execute()
             .join()
             .rows();
     try {
       return result.stream()
-          .map(row -> row.getInt("ttl(leaf)"))
+          .map(row -> row.getInt(String.format("ttl(%s)", DocsApiConstants.LEAF_COLUMN_NAME)))
           .filter(Objects::nonNull)
           .mapToInt(Integer::valueOf)
           .max()
--- a/testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
+++ b/testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
@@ -403,6 +403,20 @@ public abstract class BaseDocumentApiV2Test extends BaseIntegrationTest {
   }
 
   @Test
+  public void testPatchWithAutoTtlNullParent() throws IOException, InterruptedException {
+    JsonNode obj1 = OBJECT_MAPPER.readTree("{ \"delete this\": \"in 5 seconds\" }");
+    JsonNode obj2 = OBJECT_MAPPER.readTree("{ \"match the parent\": \"this\", \"a\": \"b\" }");
+    // No ttl on the parent
+    RestUtils.put(authToken, collectionPath + "/1", obj1.toString(), 200);
+    RestUtils.put(authToken, collectionPath + "/1/b?ttl=auto", obj2.toString(), 200);
+
+    TimeUnit.SECONDS.sleep(5);
+
+    String res = RestUtils.get(authToken, collectionPath + "/1/b?raw=true", 200);
+    assertThat(OBJECT_MAPPER.readTree(res)).isEqualTo(obj2);
+  }
+
+  @Test
   public void testWeirdButAllowedKeys() throws IOException {
     JsonNode obj = OBJECT_MAPPER.readTree("{ \"$\": \"weird but allowed\" }");
     RestUtils.put(authToken, collectionPath + "/1/path", obj.toString(), 200);


restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -304,6 +304,26 @@ private Query getInsertQueryForTable(ImmutableKeyspaceAndTable info) {
         .join();
   }
 
+  private Query getInsertQueryForTableUsingTtl(ImmutableKeyspaceAndTable info, int ttl) {
```
ivansenic, line 307: Can we have `int` -> `Integer` and set TTL if the object is not null.. Then we don't have to duplicate the logic in the above method and have simply:

```
private Query getInsertQueryForTable(ImmutableKeyspaceAndTable info) {
    return getInsertQueryForTable(info, null);
}
```

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -379,8 +403,13 @@ public BoundQuery getInsertStatement(
 
     ImmutableKeyspaceAndTable info =
         ImmutableKeyspaceAndTable.builder().keyspace(keyspaceName).table(tableName).build();
-    // Hits a cache if the insert has been called for this table in the last 5 minutes
-    Query insertQuery = insertQueryForTable.get(info);
+    Query insertQuery;
+    if (ttl != null) {
```
ivansenic, line 407: and then you don't need this..

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -126,6 +126,11 @@ public Response postDoc(
       @ApiParam(value = "the name of the collection", required = true) @PathParam("collection-id")
           String collection,
       @ApiParam(value = "The JSON document", required = true) String payload,
+      @ApiParam(
+              value = "Include this to put a time-to-live on the document in the data store",
+              required = false)
+          @QueryParam("ttl")
+          Integer ttl,
```
ivansenic, line 133: Validation that is bigger than 0? This way I can pass negative values..

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -212,6 +218,11 @@ public Response writeManyDocs(
               required = false)
           @QueryParam("id-path")
           String idPath,
+      @ApiParam(
+              value = "Include this to put a time-to-live on the document in the data store",
+              required = false)
+          @QueryParam("ttl")
+          Integer ttl,
```
ivansenic, line 225: same here, validate positive

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -350,6 +368,11 @@ public Response putDocPath(
           @NotNull(message = "payload not provided")
           @NotBlank(message = "payload must not be empty")
           String payload,
+      @ApiParam(
+              value = "Include this to put a time-to-live on the document in the data store",
+              required = false)
+          @QueryParam("ttl")
+          Integer ttl,
```
ivansenic, line 375: and here

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -424,6 +448,11 @@ public Response patchDoc(
...
```
ivansenic, line 455: and here

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -502,6 +532,11 @@ public Response patchDocPath(
...
```
ivansenic, line 539: and here

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -660,6 +660,7 @@ private void writeNewArrayState(
         bindParams,
         processedPath,
         timeSource.currentTimeMicros(),
+        null,
```
ivansenic, line 663: is this intended? What actually about the documents we insert with TTL, but then don't update with TTL.. This smells like a problem imo?

restapi/src/test/java/io/stargate/web/docsapi/dao/DocumentDBTest.java
```diff
@@ -92,7 +92,7 @@ public void getInsertStatement() {
     values[idx++] = true; // bool_value
     assertThat(idx).isEqualTo(values.length);
 
-    BoundQuery query = documentDB.getInsertStatement("keyspace", "table", 1, values);
+    BoundQuery query = documentDB.getInsertStatement("keyspace", "table", 1, values, null);
```
ivansenic, line 95: can we have one test with TTL please? I also see that we don't have it in other unit tests? Can we add?

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -354,6 +354,28 @@ public void testEscapableKeyPut() throws IOException {
             "{\"description\":\"Request invalid: payload must not be empty.\",\"code\":422}");
   }
 
+  @Test
+  public void testTtlPut() throws IOException {
+    JsonNode obj1 = OBJECT_MAPPER.readTree("{ \"delete this\": \"in ten seconds\" }");
+    JsonNode obj2 = OBJECT_MAPPER.readTree("{ \"do not delete\": \"this\", \"a\": \"b\" }");
+    JsonNode subObj2 = OBJECT_MAPPER.readTree("{ \"delete this as well\": \"in ten seconds\" }");
+    RestUtils.put(authToken, collectionPath + "/1?ttl=10", obj1.toString(), 200);
```
ivansenic, line 362: hmm is TTL in seconds? can we decrease to `1` and sleep only one second as well, no need to wait this long?

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -778,6 +800,30 @@ public void testWriteManyDocs() throws IOException {
...
+    // Create documents using multiExample that creates random ID's
+    URL url = Resources.getResource("multiExample.json");
+    String body = Resources.toString(url, StandardCharsets.UTF_8);
+    String resp = RestUtils.post(authToken, collectionPath + "/batch?ttl=10", body, 202);
+    JsonNode respBody = OBJECT_MAPPER.readTree(resp);
+    ArrayNode documentIds = (ArrayNode) respBody.requiredAt("/documentIds");
+    assertThat(documentIds.size()).isEqualTo(27);
+
+    try {
+      Thread.sleep(11000);
```
ivansenic, line 814: same here, no need to make test wait for 10s

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -354,6 +354,28 @@ public void testEscapableKeyPut() throws IOException {
...
+  @Test
+  public void testTtlPut() throws IOException {
+    JsonNode obj1 = OBJECT_MAPPER.readTree("{ \"delete this\": \"in ten seconds\" }");
+    JsonNode obj2 = OBJECT_MAPPER.readTree("{ \"do not delete\": \"this\", \"a\": \"b\" }");
+    JsonNode subObj2 = OBJECT_MAPPER.readTree("{ \"delete this as well\": \"in ten seconds\" }");
+    RestUtils.put(authToken, collectionPath + "/1?ttl=10", obj1.toString(), 200);
+    RestUtils.put(authToken, collectionPath + "/2", obj2.toString(), 200);
+    RestUtils.put(authToken, collectionPath + "/2/a?ttl=10", subObj2.toString(), 200);
+
+    try {
```
ivansenic, line 366: can we use awailability here:

```
.await()
  .atLeast(Duration.ofSeconds(..))
  .untilAsserted(() -> {

  })
```

or using `between` function..

ivansenic (changes requested):
Left some comments and ideas.. One thing that worries me is patching..

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
EricBorczuk: yea I think I need to allow supplying a TTL to this as well

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -660,6 +668,7 @@ private void writeNewArrayState(
         bindParams,
         processedPath,
         timeSource.currentTimeMicros(),
+        ttl,
```
ivansenic, line 671: What I was thinking was the following situation:

1. I add a document with TTL
2. Then I patch same document at some point (not providing TTL, because how would I calculate now what should it be
3. When TTL is over part inserted in step one gets deleted, but patched part is still there..

ivansenic (commented):
Tests failing, not sure you understood my patching problem..

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
EricBorczuk: The TTL value is sticky - If you set it and then don't touch it, it will not be overwritten.

So for your case, in 1. you would set it, and since in 2. you are leaving it `null` (and not re-setting it), it will be left alone, so 3. would not occur.

This is vaguely mentioned here (you have to set it explicitly to 0 to "cancel" the TTL): https://docs.datastax.com/en/dse/5.1/cql/cql/cql_using/useExpireExample.html

EricBorczuk:
Fixing the tests now!

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -286,22 +286,39 @@ public void tableExists(String keyspaceName, String tableName) {
...
-    return dataStore
-        .prepare(
-            dataStore
-                .queryBuilder()
-                .insertInto(info.getKeyspace(), info.getTable())
-                .value(markers)
-                .timestamp()
-                .build())
-        .join();
+    if (ttl != null) {
```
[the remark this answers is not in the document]
EricBorczuk: I've been told that explicitly using duzzt intermediate classes is not a good idea because they are transient

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
ivansenic: I think this is not true.. Let's say at step (1) I am inserting a doc and that adds 10 rows, all will have TTL. In the step (2) I  patch the document and add 5 rows without the TTL. In step (3) I am left with only 5 rows from the patch, right?

It's easier when we have a single row insert and update, but here we work against docs and they result in multiple rows. That's what I had in mind.

EricBorczuk: oh never mind - there is an issue here because we are deleting a whole set of rows and then adding an entire different set of rows...yea I'm not sure how to reconcile this one, other than to select out the current TTL of the rest of the document and then explicitly set to that value.

That kind of ruins read-before-write for our PATCH/PUT operations though...I'll have to think on this a bit more

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
dougwettlaufer: What does the error look like when using an annotation? Is it consistent with our other 400s?

ivansenic: Yes there is peace of code that takes that message from annotation and forms error response in the common form -> [JerseyViolationExceptionMapper](https://github.com/stargate/stargate/blob/master/restapi/src/main/java/io/stargate/web/impl/JerseyViolationExceptionMapper.java)..

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
ivansenic: Yes, but I think we can not solve this so easily.. We can say that's a limitation, however, I guess this is something users would not be happy about..

EricBorczuk: Yeah this is no good as a limitation because if you set TTL on a document and then PATCH without a TTL, you have no way of getting it to behave in a "default" or reasonable manner - the chunk of document you patch with will persist forever and there's no way around it...also there's no way to get it to match the current TTL of the rest of the document without read-before-write. Maybe I need to add a parameter to make the patch do a slower operation (a read first) in order to attempt to match the TTL of the document that exists...

jeffreyscarpenter: What behavior is desirable / expected here? From a user perspective, I'd expect TTL to apply at the document level, not that portions of the document could expire at different times. I'm not sure about the idea of updating TTL on an existing document, whether that would be a good idea or not.

EricBorczuk: @jeffreyscarpenter the expected behavior would be that if you set a TTL on a newly created document, and then you patch a sub-path of the document, the patched sub-path should have the same TTL as the document.

Currently if you update a document that has a TTL, the patched part would have no TTL (and therefore would unintentionally last forever) - making it automatically try to match the surrounding document's TTL would require read-before-write, which would impact performance every time (even when the document has no TTL, because you would have to check the database to see if the document has any TTL)

So the only way to implement this I think would be slightly more un-ergonomic because documents are really more than one row...and maybe every time you patch a document that you know has a TTL, you have to supply `&ttl=match`, in order to do the more expensive read-before-write and match the TTL of the document. And if you don't supply it, the sub-document will last forever (no TTL).

I'm open really to any other suggestion, this is surprisingly complex for what I thought was going to be a straightforward impl

EricBorczuk: Maybe I will call it `ttl=auto` to make it a lil clearer

tatu-at-datastax: Was this issue considered when document shredding idea was first conceived? And if so, what were the expected semantics? It really sounds like a very complicated issue at low level, even if I agree that from user perspective it's most logical that the whole document has same TTL settings.

Or alternatively that one should not have auto-expiring documents at all (but possible allow auto-expiring values, i.e. per row)?

EricBorczuk:
So I added the option `ttl=auto` when you PATCH or PUT to a sub-path of a document - what this does is query for the document's TTL and uses that value as the TTL of the new patched sub-document.

This is expected to be a bit slower, which is why it will NOT be the default behavior for writing; if you exclude `ttl` you will always get permanent data that is written without read-before-write.

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
EricBorczuk: It was definitely not considered to begin with :/ but since Cassandra allows TTL on rows, a user that is coming from cassandra to docs API currently wants this feature...

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -123,6 +126,30 @@ public Keyspace getKeyspace(String keyspaceName) {
     return dataStore.schema().keyspaces();
   }
 
+  public Integer getTtlForDocument(String namespace, String collection, String docId) {
+    List<Row> result =
+        this.dataStore
+            .queryBuilder()
+            .select()
+            .function(ImmutableList.of(QueryBuilderImpl.FunctionCall.ttl("leaf")))
```
ivansenic, line 134: can we use constants here `LEAF`, `KEY`, we already have them

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
```diff
@@ -502,6 +538,12 @@ public Response patchDocPath(
           @NotNull(message = "payload not provided")
           @NotBlank(message = "payload must not be empty")
           String payload,
+      @ApiParam(
+              value =
+                  "Include this to put a time-to-live on the document in the data store; use 'auto' to make the TTL match that of the parent document.",
+              required = false)
+          @QueryParam("ttl")
+          String ttl,
```
ivansenic, line 546: Well this is confusing, it's `int` or `string`, I am not sure if this is really the best way.. how would this be complicated from the user perspective? Actually in HTTP query params are always strings right?

ivansenic (commented):
I personally think it's OK to do it this way.. I have kind of a strange feeling that TTL can be both integer or auto, but I guess we can live with it.. I would only ask to add a INT test when `ttl=auto`, but there is no existing TTL in the document..

restapi/src/main/java/io/stargate/web/docsapi/resources/DocumentResourceV2.java
EricBorczuk: yea to the user all params are strings

ivansenic (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
jeffreyscarpenter: In my opinion, there should only be one TTL for a document. This is kind of an advanced feature and if someone is opting in to use it, they should do so knowing that they are asking for a read-before-write. Would it help if the TTL was stored as document metadata?

tatu-at-datastax: Conceptually I agree but I am not sure how viable this is since I don't know if:

1. there actually is separate per-document metadata (I thought it's just a bunch of rows (each representing a leaf node))
2. underlying storage will operate on per-row (or per-field-value) basis and has no concept of "this set of rows work logically as a bigger entity"

Now obviously one can, with lots of work and overhead, work around this by synchronizing all rows. But it seems fragile and possibly a losing battle. It may lead to the worst kinds of flaws -- things are very rare, but with enough traffic, not rare enough. So we could get cases where arbitrary document nodes just disappear due to TTL value being off, having to investigate, and without much to go on.

So I hope we are very confident that our desire to make TTL work for Documents does not come to bite us in near future.

Does Mongo support TTL for documents or parts therein?

EricBorczuk:
Closing in favor of updated/rebased: https://github.com/stargate/stargate/pull/1670

PR #1430, opened 2021-11-22
Tiny improvements to REST/getTable IT checks to avoid passing with no columns

tatu-at-datastax:
**What this PR does**:

Adds checks for REST API integration test, "getTable()" method, to avoid false pass if no column definitions returned.
Problem found with SGv2 prototyping.

No issue filed

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
+++ b/testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
@@ -303,6 +303,7 @@ public class RestApiv2Test extends BaseIntegrationTest {
     assertThat(table.getKeyspace()).isEqualTo("system");
     assertThat(table.getName()).isEqualTo("local");
     assertThat(table.getColumnDefinitions()).isNotNull();
+    assertThat(table.getColumnDefinitions()).isNotEmpty();
   }
 
   @Test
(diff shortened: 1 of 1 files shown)


testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -343,6 +346,9 @@ public void getTableComplex() throws IOException {
     assertThat(columnDefinition)
         .usingRecursiveComparison()
         .isEqualTo(new ColumnDefinition("col1", "frozen<map<date, varchar>>", false));
+
+    // but also check we seem to have right number of columns as well
+    assertThat(table.getColumnDefinitions()).hasSize(4);
```
ivansenic, line 351: why not combining this all together at line 339:

```
assertThat(table.getColumnDefinitions())
   .isNotNull()
   .isNotEmpty()
   .hasSize(4)
```

although only `hasSize` would be enough and you can combine column assertion:


```
assertThat(table.getColumnDefinitions())
   .hasSize(4)
   .anySatisfy(columnDefinition -> assertThat(columnDefinition)
        .usingRecursiveComparison()
        .isEqualTo(new ColumnDefinition("col1", "frozen<map<date, varchar>>", false)
  );
```

ivansenic (approved):

mpenick (approved):

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
tatu-at-datastax: Thanks! Will change along suggestion. One possible reason to keep some checks separate is for more granular detection (which of checks failed), but hopefully line number matches failing part of assertion (or assertion fail message).

PR #1431, opened 2021-11-22
Prototype structured query representation in gRPC

olim7t:
A small experiment to see what it would look like to pass structured query representations instead of raw CQL strings in our gRPC API.

This models part of the SELECT statement (not including: SELECT JSON, GROUP BY, ORDER BY, LIMIT, PER PARTITION LIMIT, ALLOW FILTERING).

The server part is not implemented yet, but it would essentially boil down to: generating a CQL string from the representation, and then passing it to our existing execution code.

I believe the right approach is to work from the CQL grammar, to make sure we cover all possible cases (or at least leave the door open to implement them in the future). If something looks overly complicated, ask and I can show query examples.

### Pros:
- no need for a query builder on the client. The generated Java code is a bit verbose, but that can be improved with utility methods (see `CqlHelper.java`).
- easier query introspection on the server (is this statement a read or a write? what is the primary key? etc).

### Cons:
- verbosity. This is only a small portion of the language and there is already a lot in the proto file.
- possibly more complicated for users not familiar with CQL.
- more sensitive to compatibility issues: if we get something wrong, or Cassandra changes the CQL language in an incompatible way, we might have to break the API in the future.


grpc-proto/src/main/java/io/stargate/grpc/CqlHelper.java
```diff
@@ -0,0 +1,76 @@
...
+/** Helper methods to build common patterns in CQL queries. */
+public class CqlHelper {
+
+  public static Statement newStatement(Select.Builder selectBuilder) {
+    return Statement.newBuilder().setSelect(selectBuilder).build();
+  }
+
+  public static Selector.Builder selectColumn(String columnName) {
+    return Selector.newBuilder().setColumnName(columnName);
+  }
```
olim7t, line 36: The goal here is to simplify the 80% most common use cases.

Selecting a column is super common.
Calling a function on a map element where the map is a UDT field is technically possible, but not likely so if the client needs it they can write the full boilerplate.

ivansenic: How complicated you think would be that we can have a DSL like helper for each statement type? This would effectively be even better than a builder? Meaning you can program like:

```
select()
   .keyspace(..)
   .table(..)
   .column(..)
   .and()
   .column(..)
```

I think this is a huge developer benefit.. Instead of mixing the builder and helper, create our own DSL on top of the builder..

ivansenic: They are a great example of easy to use DSL -> https://github.com/awaitility/awaitility

grpc-proto/src/test/java/io/stargate/grpc/TmpCqlExamples.java
```diff
@@ -0,0 +1,67 @@
...
+
+  public static void main(String[] args) {
+    Statement statement;
+
+    // SELECT * FROM ks.tbl
+    statement = newStatement(Select.newBuilder().setKeyspaceName("ks").setTableName("tbl"));
+
+    // SELECT c1, c2 FROM ks.tbl
+    statement =
+        newStatement(
```
tomekl007, line 25: The builder looks elegant, however, I think that we should discuss more:
> possibly more complicated for users not familiar with CQL.

I think it contradicts the direction that we wanted to achieve, that is abstracting away CQL from end-users as much as possible and not tight couple our APIs to the Cassandra backend.

mpenick:
> easier query introspection on the server (is this statement a read or a write? what is the primary key? etc).

mpenick:
> easier query introspection on the server (is this statement a read or a write? what is the primary key? etc).

Are we trying to avoid parsing CQL to introspect? I'm pretty sure it's not too expensive and may not be too hard to reuse the existing grammar(s) that are widely available. Another advantage to parsing server-side is the CQL native protocol and the existing string query gRPC API can also introspect.

I understand going this route if the API is truly preferred by users and/or is easier to use, but not if it's to avoid parsing.

grpc-proto/src/main/java/io/stargate/grpc/CqlHelper.java
olim7t: I don't know, that sounds like a lot of work, essentially double the effort to add another layer. I'd rather keep things simple.

grpc-proto/src/test/java/io/stargate/grpc/TmpCqlExamples.java
tatu-at-datastax: I am not sure I follow (or maybe I am missing context) -- this would not add (or reduce) need for CQL knowledge but offer another method of building CQL, mostly to be used by developers who build new APIs.

tatu-at-datastax:
> > easier query introspection on the server (is this statement a read or a write? what is the primary key? etc).
> 
> Are we trying to avoid parsing CQL to introspect? I'm pretty sure it's not too expensive and may not be too hard to reuse the existing grammar(s) that are widely available. Another advantage to parsing server-side is the CQL native protocol and the existing string query gRPC API can also introspect.
> 
> I understand going this route if the API is truly preferred by users and/or is easier to use, but not if it's to avoid parsing.

It is about sequence of parsing, validating, using, generating/serializing queries; a series of no-value-add operations; keeping operations in a more usable form (CQL string can only really be used after it gets parsed, or somehow massaged).
So avoiding unnecessary back and forth processing which does add overhead, complexity and potentially flaws.
A secondary benefit is easier programmatic generation of queries via builders.

I agree that it is not necessarily a burning performance problem to solve: while CQL parsing is not free (I'd expect it to be slower, per character, than decoding json or xml, given syntax), query strings tend to be short, and much depends on kind of object model that is built on contents.

But no, I don't think the main reason is performance.

mpenick:
> It is about sequence of parsing, validating, using, generating/serializing queries; a series of no-value-add operations; keeping operations in a more usable form (CQL string can only really be used after it gets parsed, or somehow massaged).

grpc-proto/src/test/java/io/stargate/grpc/TmpCqlExamples.java
tomekl007: In this approach, the caller application needs to know more about CQL grammar than just passing a plain CQL. The builder code will be used in more places in the caller code. It may possibly add some more tight coupling to Cassandra (as we may propagate it in the builder). If we would need to switch the backed from C* to other (that was one of the ideas of Stargate) it may be harder to map this new structure to a query, than just allowing user to pass CQL.

tatu-at-datastax: Ok that part makes sense wrt knowledge. But one thing to note is that this is not an end-user API -- end users would use APIs, and APIs (and developers of APIs) would use structured approach. So abstraction would be at Document API or GraphQL or REST level.

I agree it is worth discussing at any rate.

olim7t:
Closing this now since this is likely not the direction we are taking.

PR #1436, opened 2021-11-23
improved error handling on describe operations

jeffreyscarpenter:
avoiding NPEs for missing keyspace/tables


grpc/src/main/java/io/stargate/grpc/service/SchemaHandler.java
```diff
@@ -42,25 +46,31 @@ public static void describeKeyspace(
...
-      //                .setDatacenterName(datacenter)
-      //                .setReplicationFactor(Integer.parseInt(replication.get(datacenter)))
-      //                .build());
-      //      }
-      //    }
+      HashMap<String, String> replication = new HashMap<String, String>(keyspace.replication());
+      if (replication.containsKey("class")) {
+        String strategyName = replication.remove("class");
+        if (strategyName.equals("SimpleStrategy")) {
+          cqlKeyspaceBuilder.putOptions("replication", Replication.simpleStrategy(1).toString());
```
tatu-at-datastax, line 62: I know it defaults to `1` but could it ever be anything else?

tatu-at-datastax (approved):

grpc/src/main/java/io/stargate/grpc/service/SchemaHandler.java
dougwettlaufer: It supports `N` but in practice it's usually `1`.
https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/architecture/archDataDistributeReplication.html#archDataDistributeReplication__simpleStrategy

PR #1438, opened 2021-11-24
maybe create table should not check for valid name if it exists

ivansenic:
**What this PR does**:

Small fix, although the JIT compiler should optimize that if branch and understands it's 100% missed, so we don't do the pattern compile and match on every call.. However, I like it explicit!

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
@@ -151,6 +151,13 @@ public class DocumentDB {
       throw new ErrorCodeRuntimeException(ErrorCode.DATASTORE_KEYSPACE_DOES_NOT_EXIST, message);
     }
 
+    // if table exists, do nothing
+    Table table = ks.table(tableName);
+    if (table != null) {
+      return false;
+    }
+
+    // otherwise, check that the name is  valid
     if (!tableName.matches("^[a-zA-Z0-9_]+$")) {
       String message =
           String.format(
@@ -159,8 +166,7 @@ public class DocumentDB {
       throw new ErrorCodeRuntimeException(ErrorCode.DATASTORE_TABLE_NAME_INVALID, message);
     }
 
-    if (ks.table(tableName) != null) return false;
-
+    // and create the table
     try {
       List<Column> columns = new ArrayList<>();
       columns.add(Column.create("key", Kind.PartitionKey, Type.Text));


restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -151,6 +151,13 @@ public boolean maybeCreateTable(String keyspaceName, String tableName) {
       throw new ErrorCodeRuntimeException(ErrorCode.DATASTORE_KEYSPACE_DOES_NOT_EXIST, message);
     }
 
+    // if table exists, do nothing
+    Table table = ks.table(tableName);
```
mpenick, line 155: This check is done for performance reasons? Instead of round-tripping to the datastore?

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -151,6 +151,13 @@ public boolean maybeCreateTable(String keyspaceName, String tableName) {
...
+    if (table != null) {
+      return false;
+    }
+
+    // otherwise, check that the name is  valid
     if (!tableName.matches("^[a-zA-Z0-9_]+$")) {
```
mpenick, line 161: Should this check be left to the datastore to validate? 

```
token@cqlsh> CREATE TABLE ks."non-standard-name?" (k int PRIMARY KEY, v int);
ConfigurationException: Table name must not be empty, more than 222 characters long, or contain non-alphanumeric-underscore characters (got "non-standard-name?")
```

We can catch that `ConfigurationException` error and change it to what you have here. This check is also missing the 222 character length limitation so maybe it makes sense to let the data store validate since it's more restrictive?

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
ivansenic: Yes, I guess so. This is called on every document write, so we don't want to go to the store every time.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
ivansenic: I am up for this, @EricBorczuk do you see any problem with that?

ivansenic: What are we doing about this one @EricBorczuk ? I am also fine to leave it as it is and spare the request to the data store if we know that the name is invalid?

ivansenic: OK I am merging as it is, saving the data store request if we already know that the table name is not valid.

PR #1439, opened 2021-11-25
refactored shredding of json payload

ivansenic:
**What this PR does**:
Provides simpler and more efficient document shredding based on the recursion. Operate only against the JsonNode and returns results in a minimal concrete class representation of shredding. The idea is that we add later on `JsonShreddedRow#bind(Query inserQuery)` and easily bind values based on the object state.

Open questions:

1. What should be the leaf & path values in case of empty objects and empty arrays?
2. Do we support primitives and what should be the leaf & path values if we do?
3. Should we escape also the prepend path? (seems this is not done at the moment)
4. Is the `ErrorCode.DOCS_API_GENERAL_INVALID_FIELD_NAME` still valid and when it should be thrown?

None.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
@@ -1,0 +1,174 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.stargate.web.docsapi.service;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.stargate.web.docsapi.exception.ErrorCode;
+import io.stargate.web.docsapi.exception.ErrorCodeRuntimeException;
+import io.stargate.web.docsapi.service.query.DocsApiConstants;
+import io.stargate.web.docsapi.service.util.DocsApiUtils;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Supplier;
+import javax.inject.Inject;
+
+public class JsonDocumentShredder {
+
+  @Inject private DocsApiConfiguration config;
+
+  @Inject private ObjectMapper objectMapper;
+
+  public JsonDocumentShredder(DocsApiConfiguration config, ObjectMapper objectMapper) {
+    this.config = config;
+    this.objectMapper = objectMapper;
+  }
+
+  public List<JsonShreddedRow> shred(
+      String payload, String documentId, List<String> subDocumentPath, boolean numericBooleans) {
+    try {
+      JsonNode node = objectMapper.readTree(payload);
+      return shred(node, documentId, subDocumentPath, numericBooleans);
+    } catch (JsonProcessingException e) {
+      throw new ErrorCodeRuntimeException(
+          ErrorCode.DOCS_API_INVALID_JSON_VALUE,
+          "Malformed JSON object found during shredding.",
+          e);
+    }
+  }
+
+  public List<JsonShreddedRow> shred(
+      JsonNode node, String documentId, List<String> subDocumentPath, boolean numericBooleans) {
+    Supplier<ImmutableJsonShreddedRow.Builder> rowBuilder =
+        () ->
+            ImmutableJsonShreddedRow.builder()
+                .key(documentId)
+                .maxDepth(config.getMaxDepth())
+                .addAllPath(subDocumentPath); // TODO should we also convert these parts with escape
+
+    List<JsonShreddedRow> result = new ArrayList<>();
+    processNode(node, rowBuilder, numericBooleans, result);
+    return result;
+  }
+
+  private void processNode(
+      JsonNode node,
+      Supplier<ImmutableJsonShreddedRow.Builder> rowBuilder,
+      boolean numericBooleans,
+      List<JsonShreddedRow> result) {
+    if (node.isArray()) {
+      processArrayNode(node, rowBuilder, numericBooleans, result);
+    } else if (node.isObject()) {
+      processObjectNode(node, rowBuilder, numericBooleans, result);
+    } else {
+      processValueNode(node, rowBuilder, numericBooleans, result);
+    }
+  }
+
+  private void processArrayNode(
+      JsonNode node,
+      Supplier<ImmutableJsonShreddedRow.Builder> rowBuilder,
+      boolean numericBooleans,
+      List<JsonShreddedRow> result) {
+    // empty array, simply create a reference to empty node and return
+    if (node.isEmpty()) {
+      ImmutableJsonShreddedRow row =
+          rowBuilder.get().stringValue(DocsApiConstants.EMPTY_ARRAY_MARKER).build();
+      result.add(row);
+      return;
+    }
+
+    // otherwise, iterate all nodes
+    int idx = 0;
+    for (JsonNode inner : node) {
+      // make sure we didn't exceed the maximum array length
+      if (idx >= config.getMaxArrayLength()) {
+        throw new ErrorCodeRuntimeException(ErrorCode.DOCS_API_GENERAL_ARRAY_LENGTH_EXCEEDED);
+      }
+
+      // convert the array index into path
+      // then create new next row builder
+      String arrayPath = "[" + DocsApiUtils.leftPadTo6(String.valueOf(idx)) + "]";
+      Supplier<ImmutableJsonShreddedRow.Builder> nextRowBuilder =
+          () -> rowBuilder.get().addPath(arrayPath);
+
+      // process inner node and increase the index
+      processNode(inner, nextRowBuilder, numericBooleans, result);
+      idx++;
+    }
+  }
+
+  private void processObjectNode(
+      JsonNode node,
+      Supplier<ImmutableJsonShreddedRow.Builder> rowBuilder,
+      boolean numericBooleans,
+      List<JsonShreddedRow> result) {
+    // empty object, simply create a reference to empty node and return
+    if (node.isEmpty()) {
+      ImmutableJsonShreddedRow row =
+          rowBuilder.get().stringValue(DocsApiConstants.EMPTY_OBJECT_MARKER).build();
+      result.add(row);
+      return;
+    }
+
+    node.fields()
+        .forEachRemaining(
+            field -> {
+              // TODO is this still valid
+              //  ErrorCode.DOCS_API_GENERAL_INVALID_FIELD_NAME
+
+              // escape the field path
+              // then create new next row builder
+              String fieldPath = DocsApiUtils.convertEscapedCharacters(field.getKey());
+              Supplier<ImmutableJsonShreddedRow.Builder> nextRowBuilder =
+                  () -> rowBuilder.get().addPath(fieldPath);
+
+              // process inner node and increase the index
+              processNode(field.getValue(), nextRowBuilder, numericBooleans, result);
+            });
+  }
+
+  private void processValueNode(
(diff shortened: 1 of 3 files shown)


jeffreyscarpenter (commented):
Does this change the way that data is actually stored in the database? Or would this just improve how we process data?

ivansenic:
> Does this change the way that data is actually stored in the database? Or would this just improve how we process data?

Processing only, currently we had the processing in [DocsShredder](https://github.com/stargate/stargate/blob/e18d5f9eb6f60e83e421570604b7a49f19e97f87/restapi/src/main/java/io/stargate/web/docsapi/service/DocsShredder.java).

restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  public List<JsonShreddedRow> shred(
+      JsonNode node, String documentId, List<String> subDocumentPath, boolean numericBooleans) {
+    Supplier<ImmutableJsonShreddedRow.Builder> rowBuilder =
+        () ->
+            ImmutableJsonShreddedRow.builder()
+                .key(documentId)
+                .maxDepth(config.getMaxDepth())
+                .addAllPath(subDocumentPath); // TODO should we also convert these parts with escape
```
EricBorczuk, line 63: I don't think we handle this case currently - but technically yes, if someone sends `PUT .../collections/x/id/path/with/a\.period/in/it` we should probably handle it...just a weird case overall

EricBorczuk (approved):
This looks great (and way more straightforward imo)!

restapi/src/main/java/io/stargate/web/docsapi/service/JsonShreddedRow.java
```diff
@@ -0,0 +1,64 @@
...
+
+  List<String> getPath();
+
+  // leaf is always actually the last path we have
+  default String getLeaf() {
+    List<String> path = getPath();
+    if (path != null && !path.isEmpty()) {
+      return path.get(path.size() - 1);
+    } else {
+      return null; // TODO is this fine? what do we return as leaf for empty objects and empty
```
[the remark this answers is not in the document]
ivansenic: What about the following, there are all valid JSON docs:

* `{}` - empty object
* `[]` - empty array
* `22` - primitives?

restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
ivansenic: adding the conversion just in case..

restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
```diff
@@ -0,0 +1,174 @@
...
+          rowBuilder.get().stringValue(DocsApiConstants.EMPTY_OBJECT_MARKER).build();
+      result.add(row);
+      return;
+    }
+
+    node.fields()
+        .forEachRemaining(
+            field -> {
+              // TODO is this still valid
+              //  ErrorCode.DOCS_API_GENERAL_INVALID_FIELD_NAME
```
ivansenic, line 134: @EricBorczuk what about this one? Is the field name restricted at the moment? I am seeing this in the current shredding:

```
                if (fieldName != null && DocsApiUtils.containsIllegalSequences(fieldName)) {
                  String msg =
                      String.format(
                          "Array paths contained in square brackets, periods, single quotes, and backslash are not allowed in field names, invalid field %s",
                          fieldName);
                  throw new ErrorCodeRuntimeException(
                      ErrorCode.DOCS_API_GENERAL_INVALID_FIELD_NAME, msg);
                }
```

EricBorczuk: Oh yes - sorry I didn't see that this was removed. Putting these characters in without explicitly escaping them with `\` should be disallowed

restapi/src/main/java/io/stargate/web/docsapi/service/JsonShreddedRow.java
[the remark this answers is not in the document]
EricBorczuk: empty object and empty array are treated as scalar values, and sending a payload that is just a scalar is illegal. By extension this also prevents all of those. cases from being written at the root of the document. See https://github.com/stargate/stargate/blob/master/restapi/src/main/java/io/stargate/web/docsapi/service/DocumentService.java#L216

dougwettlaufer (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/JsonShreddedRow.java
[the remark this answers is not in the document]
ivansenic: Is there any reason this is done so? Is the leaf value the problem? In the current implementation I allowed this with leaf being null in that case, I guess this would be a problem or? Maybe we can support it now?

restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
ivansenic: OK, hope I got this right, can you maybe quickly check my tests for these?

restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
ivansenic: Should these be also checked for the illegal chars then?

restapi/src/main/java/io/stargate/web/docsapi/service/JsonShreddedRow.java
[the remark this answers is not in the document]
ivansenic: And is this valid only for patching or also for the write a complete document? I guess this is also valid only when this is at the root of the object right?

ivansenic:
@EricBorczuk can you have another look and answer to the new questions.. I want to get this straight.. Thanks!

restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
```diff
@@ -0,0 +1,211 @@
...
+    // check if this is a valid root node
+    if (subDocumentPath.isEmpty()) {
+      checkRoot(node);
+    }
+
+    // sub-paths escaped
+    List<String> subPathsEscaped =
+        subDocumentPath.stream()
+            .map(DocsApiUtils::convertEscapedCharacters)
+            .collect(Collectors.toList());
```
ivansenic, line 68: @EricBorczuk maybe we should leave this to caller and not touch the subDocumentPath?

restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
```diff
@@ -0,0 +1,211 @@
...
+          rowBuilder.get().stringValue(DocsApiConstants.EMPTY_ARRAY_MARKER).build();
+      result.add(row);
+      return;
+    }
+
+    // otherwise, iterate all nodes
+    int idx = 0;
+    for (JsonNode inner : node) {
+      // make sure we didn't exceed the maximum array length
+      if (idx >= config.getMaxArrayLength()) {
```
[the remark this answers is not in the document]
ivansenic: Sure, great catch :+1:

restapi/src/main/java/io/stargate/web/docsapi/service/JsonShreddedRow.java
```diff
@@ -0,0 +1,63 @@
...
+    List<String> path = getPath();
+    if (path != null && !path.isEmpty()) {
+      return path.get(path.size() - 1);
+    } else {
+      throw new IllegalStateException("Shredded row does not contain a single path.");
+    }
+  }
+
+  @Nullable
+  String getStringValue();
```
mpenick, line 55: Since we're already boxing/unboxing these values. What do you think about using a "tagged union" representation?

```java
@Value.Immutable
public interface JsonShreddedRow {

  enum ValueType {
    Unknown,
    String,
    Double,
    Boolean,
  }

  ValueType type();

  Object value();
}
```

The representation is a bit more compact in the generated immutable code and the value extraction comparisons might be a bit faster (maybe).

```
    switch (row.type()) {
      case Unknown:
        // Other stuff
      case String:
        // String stuff
      case Double:
        // Double stuff
      case Boolean:
        // Boolean stuff
    }

    if (row.getStringValue() != null) {
      // String stuff
    } else if (row.getDoubleValue() != null) {
      // Double stuff
    } else if (row.getBooleanValue() != null) {
      // Boolean stuff
    } else {
      // Other stuff
    }
```

restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
```diff
@@ -0,0 +1,211 @@
...
+    } else if (!node.isNull()) {
+      builder.stringValue(node.asText());
+    }
+
+    // build and add to the results
+    ImmutableJsonShreddedRow row = builder.build();
+    result.add(row);
+  }
+
+  private Object convertToBackendBooleanValue(boolean value, boolean numericBooleans) {
```
mpenick, line 205: Does this conversion need to be done here? It's seems like this layer might not need to deal with this. Would it be possible to use `boolean` here and the conversion to the numeric type be done in the storage layer right before it's stored?

restapi/src/main/java/io/stargate/web/docsapi/service/JsonShreddedRow.java
ivansenic: Nope, I need all the values when storing the row as we do have three columns for the values.. So for me the switch does not make any sense.

restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
ivansenic: Yea seems that this makes sense.. I think I will apply this fix as well, shredder should not care about how are the values stored in the DB. Thanks :+1:

restapi/src/main/java/io/stargate/web/docsapi/service/JsonShreddedRow.java
mpenick: Ah, makes sense. So we store more than one of these values in a row?

restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
EricBorczuk: I think it would be reasonable to do it higher up, yea

restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
EricBorczuk: Yea, if we are allowing escaping like this we would likely have to make sure that "raw" unescaped chars aren't sent

EricBorczuk (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/JsonShreddedRow.java
ivansenic: No it's only one.. But I still need to bind the others to `null`..

restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
ivansenic: Done, resolving.

restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
ivansenic: I reverted the escaping and delegated to caller, so I would not be doing this as well. Responsibility of the caller.

restapi/src/main/java/io/stargate/web/docsapi/service/JsonShreddedRow.java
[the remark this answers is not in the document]
ivansenic: OK I'll merge but @EricBorczuk let's sync later about this.

restapi/src/main/java/io/stargate/web/docsapi/service/JsonShreddedRow.java
mpenick: Why do we need to bind the others to `null`? Can't they be left unset?

ivansenic: Just for the record, because the type of the JSON field can change on update, so we need to count for that upfront.

PR #1448, opened 2021-12-01
Add string-only QueryBuilder

olim7t:
Clone the persistence API's `QueryBuilder`, but the new version only produces plain query strings. This would be used by v2 services to generate the CQL queries they'll send to the bridge (if we go with a plain CQL approach). In the current REST prototype this is done with the Java driver's query builder, but we want to avoid pulling that dependency.

Notes:
* this duplicates everything. I didn't make any attempt to reuse code, because I think the persistence query builder won't exist anymore in the target v2 architecture.
* I duplicated the DSL as-is. There are things I might have done differently, but I don't want to restart everything from scratch.
* column types are passed as raw strings. This will work for our current needs with the REST API, we can introduce a more structured representation later if needed.

### DIFF
--- a/persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
+++ b/persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
@@ -15,12 +15,6 @@
  */
 package io.stargate.db.query.builder;
 
-import static com.datastax.oss.driver.shaded.guava.common.base.Preconditions.checkArgument;
-import static io.stargate.db.query.BindMarker.markerFor;
-import static java.lang.String.format;
-import static java.util.stream.Collectors.toList;
-import static java.util.stream.Collectors.toSet;
-
 import com.datastax.oss.driver.internal.core.util.Strings;
 import com.datastax.oss.driver.shaded.guava.common.collect.Sets;
 import com.github.misberner.apcommons.util.AFModifier;
@@ -58,6 +52,12 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import org.javatuples.Pair;
+
+import static com.datastax.oss.driver.shaded.guava.common.base.Preconditions.checkArgument;
+import static io.stargate.db.query.BindMarker.markerFor;
+import static java.lang.String.format;
+import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toSet;
 
 /** Convenience builder for creating queries. */
 @GenerateEmbeddedDSL(
--- a/persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
+++ b/persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
@@ -15,6 +15,12 @@
  */
 package io.stargate.db.query.builder;
 
+import static com.datastax.oss.driver.shaded.guava.common.base.Preconditions.checkArgument;
+import static io.stargate.db.query.BindMarker.markerFor;
+import static java.lang.String.format;
+import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toSet;
+
 import com.datastax.oss.driver.internal.core.util.Strings;
 import com.datastax.oss.driver.shaded.guava.common.collect.Sets;
 import com.github.misberner.apcommons.util.AFModifier;
@@ -52,12 +58,6 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import org.javatuples.Pair;
-
-import static com.datastax.oss.driver.shaded.guava.common.base.Preconditions.checkArgument;
-import static io.stargate.db.query.BindMarker.markerFor;
-import static java.lang.String.format;
-import static java.util.stream.Collectors.toList;
-import static java.util.stream.Collectors.toSet;
 
 /** Convenience builder for creating queries. */
 @GenerateEmbeddedDSL(
--- a/sgv2-service-common/pom.xml
+++ b/sgv2-service-common/pom.xml
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.stargate</groupId>
+        <artifactId>stargate</artifactId>
+        <version>1.0.44-SNAPSHOT</version>
+    </parent>
+    <artifactId>sgv2-service-common</artifactId>
+    <dependencies>
+        <dependency>
+            <groupId>com.github.misberner.duzzt</groupId>
+            <artifactId>duzzt-processor</artifactId>
+            <version>0.0.2</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.immutables</groupId>
+            <artifactId>value</artifactId>
+            <version>2.8.8</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>jsr305</artifactId>
+            <version>3.0.2</version>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
--- a/sgv2-service-common/pom.xml
+++ b/sgv2-service-common/pom.xml
@@ -26,6 +26,7 @@
             <groupId>com.google.code.findbugs</groupId>
             <artifactId>jsr305</artifactId>
             <version>3.0.2</version>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
--- a/sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/ReservedKeywords.java
+++ b/sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/ReservedKeywords.java
@@ -16,86 +16,86 @@
 package io.stargate.sgv2.common.cql;
 
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.Set;
-import java.util.concurrent.CopyOnWriteArraySet;
 
 public final class ReservedKeywords {
-  private static final String[] reservedKeywords =
-      new String[] {
-        "SELECT",
-        "FROM",
-        "WHERE",
-        "AND",
-        "ENTRIES",
-        "FULL",
-        "INSERT",
-        "UPDATE",
-        "WITH",
-        "LIMIT",
-        "USING",
-        "USE",
-        "SET",
-        "BEGIN",
-        "UNLOGGED",
-        "BATCH",
-        "APPLY",
-        "TRUNCATE",
-        "DELETE",
-        "IN",
-        "CREATE",
-        "KEYSPACE",
-        "SCHEMA",
-        "COLUMNFAMILY",
-        "TABLE",
-        "MATERIALIZED",
-        "VIEW",
-        "INDEX",
-        "ON",
-        "TO",
-        "DROP",
-        "PRIMARY",
-        "INTO",
-        "ALTER",
-        "RENAME",
-        "ADD",
-        "ORDER",
-        "BY",
-        "ASC",
-        "DESC",
-        "ALLOW",
-        "IF",
-        "IS",
-        "GRANT",
-        "OF",
-        "REVOKE",
-        "MODIFY",
-        "AUTHORIZE",
-        "DESCRIBE",
-        "EXECUTE",
-        "NORECURSIVE",
-        "TOKEN",
-        "NULL",
-        "NOT",
-        "NAN",
-        "INFINITY",
-        "OR",
-        "REPLACE",
-        "DEFAULT",
-        "UNSET",
-        "MBEAN",
-        "MBEANS",
-        "FOR",
-        "RESTRICT",
-        "UNRESTRICT"
-      };
 
-  private static final Set<String> reservedSet =
-      new CopyOnWriteArraySet<>(Arrays.asList(reservedKeywords));
+  private static final Set<String> KEYWORDS =
+      Collections.unmodifiableSet(
+          new HashSet<>(
+              Arrays.asList(
+                  "SELECT",
+                  "FROM",
+                  "WHERE",
+                  "AND",
+                  "ENTRIES",
+                  "FULL",
+                  "INSERT",
+                  "UPDATE",
+                  "WITH",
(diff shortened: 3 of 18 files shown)


sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/BuiltCondition.java
```diff
@@ -0,0 +1,157 @@
...
+
+import io.stargate.sgv2.common.cql.ColumnUtils;
+import java.util.Optional;
+import org.immutables.value.Value.Immutable;
+import org.immutables.value.Value.Style;
+import org.immutables.value.Value.Style.ImplementationVisibility;
+
+@Immutable
+@Style(visibility = ImplementationVisibility.PACKAGE)
+public interface BuiltCondition {
```
ivansenic, line 26: adding Javadocs would be awesome and so helpful for all developers, especially external ones

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/Column.java
```diff
@@ -0,0 +1,57 @@
...
+import io.stargate.sgv2.common.cql.ColumnUtils;
+import javax.annotation.Nullable;
+import org.immutables.value.Value;
+import org.immutables.value.Value.Immutable;
+
+@Immutable(prehash = true)
+public interface Column {
+  String name();
+
+  /** TODO do we ever need a more structured representation? */
```
ivansenic, line 27: what would `a more structured representation` be?

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/ConcreteValue.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.sgv2.common.cql.builder;
+
+class ConcreteValue<T> implements Value<T> {
```
ivansenic, line 18: should we also take advantage of `Immutable` here, so we get proper `hashCode` and `equals` that we are missing at the moment?

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/QueryBuilderImpl.java
```diff
@@ -0,0 +1,1369 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/** Convenience builder for creating queries. */
+@GenerateEmbeddedDSL(
+    modifier = AFModifier.DEFAULT,
+    autoVarArgs = false,
+    name = "QueryBuilder",
+    syntax = "(<keyspace>|<table>|<insert>|<update>|<delete>|<select>|<index>|<type>) build",
+    where = {
```
ivansenic, line 41: Is there any idea how we could rewrite this to something that provides better typed stuff in the middle of the query building? Do we at least have an idea what can this be?

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/Replication.java
```diff
@@ -0,0 +1,56 @@
...
+import static java.lang.String.format;
+
+import java.util.Map;
+
+/**
+ * Represents a keyspace replication string.
+ *
+ * <p>So something like "{ 'class': 'SimpleStrategy', 'replication_factor': 1}".
+ */
+public class Replication {
```
ivansenic, line 27: could also be immutable..

ivansenic (commented):
Looks good!

Since this is a new project I would strongly consider going for `Lombok` over `Immutable`, I guess they both have their pros and cons, but I like that with Lombok you can write short classes and you don't end up with the definition and implementation class, but only the single one..

persistence-api/src/main/java/io/stargate/db/query/builder/QueryBuilderImpl.java
```diff
@@ -59,6 +53,12 @@
 import javax.annotation.Nullable;
 import org.javatuples.Pair;
 
+import static com.datastax.oss.driver.shaded.guava.common.base.Preconditions.checkArgument;
```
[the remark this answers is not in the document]
olim7t: Might be an issue on my side, I'll check. The command-line formatter does not enforce a particular order and I think that's one of the last thing where we have discrepancies across the team.

sgv2-service-common/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+            <groupId>com.github.misberner.duzzt</groupId>
+            <artifactId>duzzt-processor</artifactId>
+            <version>0.0.2</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.immutables</groupId>
+            <artifactId>value</artifactId>
+            <version>2.8.8</version>
+            <scope>provided</scope>
```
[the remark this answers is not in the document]
olim7t: It's an annotation processor, it's just used to generate code at compile time, the dependency is not needed at runtime.

sgv2-service-common/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
```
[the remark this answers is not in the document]
olim7t: It brings the `@Nullable` annotation that's used to mark field optional in `@Immutable` classes.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/ReservedKeywords.java
```diff
@@ -0,0 +1,101 @@
...
+        "UNSET",
+        "MBEAN",
+        "MBEANS",
+        "FOR",
+        "RESTRICT",
+        "UNRESTRICT"
+      };
+
+  private static final Set<String> reservedSet =
+      new CopyOnWriteArraySet<>(Arrays.asList(reservedKeywords));
```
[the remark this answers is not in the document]
olim7t: What we're going for here is thread safety. But actually the set is immutable, so `Collections.unmodifiableSet` would be sufficient.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/Column.java
olim7t: Something like `ColumnType` in the persistence API.
Currently we are focused on the REST API, where types are just provided as strings in the JSON payload, so this is good enough.
But for GraphQL we have more elaborate use cases, like translating a GraphQL type into a CQL type.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/ConcreteValue.java
olim7t: We could, but I don't see a strong incentive given how simple the class is.
Note that we don't need value equality (in fact this type is never compared, used as map key, etc), so the default `equals` and `hashCode` are appropriate.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/ReservedKeywords.java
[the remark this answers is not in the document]
ivansenic: Yup.

sgv2-service-common/pom.xml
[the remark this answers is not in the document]
ivansenic: Hm, ok, maybe then this could also be on the `provided` scope?

Furthermore, I bumped to this: https://github.com/immutables/immutables/issues/647#issuecomment-313467651, there are multiple options for avoiding non-null annotation (like optional, or this value check style).. Not sure if anything of that is better than this, wdyt?

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/Column.java
ivansenic: Got it.. Yea maybe we change when we need it, or?

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/ConcreteValue.java
ivansenic: I am always afraid that one day somebody will decide to use such a class instance in a map.. That's why I am always looking to ensure that such POJO classes are properly implemented.. But I get your point..

jeffreyscarpenter (approved):

sgv2-service-common/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.stargate</groupId>
+        <artifactId>stargate</artifactId>
+        <version>1.0.44-SNAPSHOT</version>
+    </parent>
+    <artifactId>sgv2-service-common</artifactId>
+    <dependencies>
+        <dependency>
+            <groupId>com.github.misberner.duzzt</groupId>
```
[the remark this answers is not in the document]
tatu-at-datastax: I think there were some notes that maybe we could only copy-paste and use what has been generated once, but stop using the generator. Would this work here?

sgv2-service-common/pom.xml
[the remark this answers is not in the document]
tatu-at-datastax: The value of non-null/nullable usually seems to be for document as not many frameworks (and few if any runtimes) make any use of it. On plus side I think it is fine to then leave dependency as `provided` since during runtime annotation dependencies are not strictly needed (unless some library/framework itself needs it & if so they should bring in the dependency).

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/Column.java
tatu-at-datastax: I wonder if we need "type builder" (and perhaps -- but not necessarily -- "type decoder")?
What I mean is that when building type signatures like `frozen<map<uuid, tinyint>>` one could have simple builder, used optionally for structured cases.

I agree that this could be added at a later stage for sure; but I also suspect that SGv2 REST API would quite soon need it when converting between gRPC protoc and REST JSON.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/Predicate.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.sgv2.common.cql.builder;
+
+public enum Predicate {
```
tatu-at-datastax, line 18: While non-ambiguous within context, I wonder if it'd be good to name it `CqlPredicate` to distinguish from among other types. Similar questions about `Value`. I guess it depends on whether this would be exposed beyond immediate building of CQL query/statement or not -- if not, does not really matter. But if type is used and passed could sometimes be confusing.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/Value.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.sgv2.common.cql.builder;
+
+public interface Value<T> {
```
tatu-at-datastax, line 18: A small Javadoc might make sense here... also, name is very generic. I don't have a feel of if it'd be exposed -- probably not as it's just a base/tag interface? -- and if not, fine as-is I suppose. Maybe `CqlValue` would be better name regardless?

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/ValueModifier.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+package io.stargate.sgv2.common.cql.builder;
+
+import javax.annotation.Nullable;
+import org.immutables.value.Value.Immutable;
+import org.immutables.value.Value.Style;
+
+@Immutable
+@Style(visibility = Style.ImplementationVisibility.PACKAGE)
+public interface ValueModifier {
```
tatu-at-datastax, line 24: ... and if `Value` was renamed, this should follow (`CqlValueModifier`).

tatu-at-datastax:
I like it! Some questions on naming: I understand the desire to avoid too much prefixing, but wondering if, for this context and likely external use, `Cql`-prefix for a few externally visible types made sense? For chained operations not that important but otherwise generic nouns like `Value` and `Predicate` tend to be provided by multiple packages and make IDE auto-completion hassle.

sgv2-service-common/src/test/java/io/stargate/sgv2/common/cql/builder/QueryBuilderTest.java
```diff
@@ -0,0 +1,322 @@
...
+              .withReplication(Replication.simpleStrategy(1))
+              .build(),
+          "CREATE KEYSPACE \"Ks\" "
+              + "WITH replication = { 'class': 'SimpleStrategy', 'replication_factor': 1 }"),
+      arguments(
+          new QueryBuilder()
+              .create()
+              .keyspace("ks")
+              .ifNotExists()
+              .withReplication(Replication.simpleStrategy(1))
```
dougwettlaufer, line 58: What about `networkTopologyStrategy`?

sgv2-service-common/src/test/java/io/stargate/sgv2/common/cql/builder/QueryBuilderTest.java
```diff
@@ -0,0 +1,322 @@
...
+              + "WITH CLUSTERING ORDER BY (cc DESC)"),
+      arguments(
+          new QueryBuilder()
+              .create()
+              .table("ks", "tbl")
+              .column("k", "int", Column.Kind.PARTITION_KEY)
+              .withComment("'test' comment")
+              .withDefaultTTL(3600)
+              .build(),
+          "CREATE TABLE ks.tbl (k int, PRIMARY KEY ((k))) "
```
dougwettlaufer, line 134: Can we add a test for creating a table with comments? We want to make sure those work since docsapi uses them.

dougwettlaufer (approved):

sgv2-service-common/src/test/java/io/stargate/sgv2/common/cql/builder/QueryBuilderTest.java
tatu-at-datastax: Quick question on this: what does docsapi use it for? Annotating table as being used for document collection, configuration or such?

olim7t: This is a create table with a comment.
Do you mean executing the query rather than just generating it?

ivansenic: Comments are used for storing a JSON schema to validate each document against.

olim7t:
I'm going to merge this PR in the current state to get things moving.

Most remaining comments are about things that were already that way in the original code, so while not illegitimate I think they are off-topic in the context of this PR. We can address them in follow-up tickets.

Issue #1435, opened 2021-11-23
Implement Stargate V2 REST endpoints for "createTable" using new gRPC endpoint


tatu-at-datastax:
(similar to #1426)

Stargate V2 REST implementation has accessor for Table metadata but "createTable()" missing -- Integration tests for now use SGv1 version. Implementing creation natively is possibly the last missing piece needed for full end-to-end testing of the minimal functionality (create ks, table(s); insert data; get data).

tatu-at-datastax:
Implemented, 2 more ITs passing.

PR #1450, opened 2021-12-02
Fix StargateV2/REST ClusteringOrder enum return value (asc->ASC)

tatu-at-datastax:
**What this PR does**:

Fixes a small inconsistency in SGv2/REST table resource: clustering order is to be return as upper-case "ASC"/"DESC", not "asc"/"desc" -- this as per integration test; some documentation seems confusing. Either value is accepted on creation , CQL being (mostly?) case-insensitive.

Related to #1435.

**Checklist**
- [ ] Changes manually tested
- [x] Automated Tests added/updated -- one more IT passing (now 38 of 95) for Stargate V2

### DIFF
--- a/sgv2-restapi/README.md
+++ b/sgv2-restapi/README.md
@@ -1,0 +1,51 @@
+# Stargate V2 - REST service/API
+
+## General
+
+This sub-project is for prototype version of stand-alone REST API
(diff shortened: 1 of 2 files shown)


jeffreyscarpenter (commented):
Love the new README!

PR #1456, opened 2021-12-06
avoid going to props every time for supportsSecondaryIndex

ivansenic:
**What this PR does**:
Avoids going to the system props every time when `Persistence#supportsSecondaryIndex` is called. We call this quite often in the docs API due to the number boolean storage check. Thus, under heavy load we create here unnecessary contention point, as the `Hashtable` is synchronized..

I guess this property was not meant to be changed over time. Thus, a small fix.

image

**Checklist**
- [x] Changes manually tested
- ~Automated Tests added/updated~
- ~Documentation added/updated~

### DIFF
--- a/persistence-api/src/main/java/io/stargate/db/Persistence.java
+++ b/persistence-api/src/main/java/io/stargate/db/Persistence.java
@@ -34,6 +34,9 @@ import org.apache.cassandra.stargate.exceptions.AuthenticationException;
  */
 public interface Persistence extends SchemaAgreementChecker {
 
+  boolean SUPPORTS_SECONDARY_INDEX =
+      Boolean.parseBoolean(System.getProperty("stargate.persistence.2i.support.default", "true"));
+
   /** Name describing the persistence implementation. */
   String name();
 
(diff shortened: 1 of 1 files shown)


tatu-at-datastax:
Nice catch! This call can be surprisingly expensive I think, and regardless no point making dynamically as value cannot really be changed (afaik) on JVM.

tatu-at-datastax (approved):

tatu-at-datastax:
The only comment: there was some talk of centralizing names of various system properties in a constant file. I assume we don't have one and so it would not make sense to start working on that for just this one. But going forward seems like it'd be good to document properties we use.

ivansenic:
> The only comment: there was some talk of centralizing names of various system properties in a constant file. I assume we don't have one and so it would not make sense to start working on that for just this one. But going forward seems like it'd be good to document properties we use.

Long term, I think we should do what you implemented in V2, detailed config property classes..

mpenick (approved):

olim7t (approved):

PR #1458, opened 2021-12-06
SgV2: Separate "rows" endpoint into API interface, impl class

tatu-at-datastax:
**What this PR does**:

Separates formerly coupled REST endpoint API and implementation into interface (API, annotations) and implementation (no method annotations, actual implementation)

**Checklist**
- [x] Changes manually tested -- verify Swagger API works as well (no unit tests)
- [x] Automated Tests added/updated -- passes same tests as before changes (all endpoints covered)

### DIFF
--- a/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceApi.java
+++ b/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceApi.java
@@ -4,6 +4,7 @@ import com.codahale.metrics.annotation.Timed;
 import io.stargate.sgv2.restsvc.models.RestServiceError;
 import io.stargate.sgv2.restsvc.models.Sgv2RESTResponse;
 import io.stargate.sgv2.restsvc.models.Sgv2RowsResponse;
+import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
 import io.swagger.annotations.ApiResponse;
@@ -21,6 +22,7 @@ import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.Context;
+import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.PathSegment;
 import javax.ws.rs.core.Response;
 
@@ -28,9 +30,13 @@ import javax.ws.rs.core.Response;
  * Definition of REST API endpoint methods including both JAX-RS and Swagger annotations. No
  * implementations.
  *
- * <p>NOTE: class annotations cannot be included in the interface (at least for JAX-RS) and must be
- * included in the implementation class.
+ * <p>NOTE: JAX-RS class annotations cannot be included in the interface and must be included in the
+ * implementation class. Swagger annotations are ok tho.
  */
+@Api(
+    produces = MediaType.APPLICATION_JSON,
+    consumes = MediaType.APPLICATION_JSON,
+    tags = {"data"})
 public interface Sgv2RowsResourceApi {
   @Timed
   @GET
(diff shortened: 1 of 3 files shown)


sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceApi.java
```diff
@@ -0,0 +1,340 @@
...
+      value = {
+        @ApiResponse(code = 200, message = "OK", response = Sgv2RowsResponse.class),
+        @ApiResponse(code = 400, message = "Bad Request", response = RestServiceError.class),
+        @ApiResponse(code = 401, message = "Unauthorized", response = RestServiceError.class),
+        @ApiResponse(
+            code = 500,
+            message = "Internal server error",
+            response = RestServiceError.class)
+      })
+  public Response getRowWithWhere(
```
olim7t, line 52: Nit: `public` can be omitted in interface definitions, it's implied.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceApi.java
```diff
@@ -0,0 +1,340 @@
...
+          String token,
+      @ApiParam(value = "Name of the keyspace to use for the request.", required = true)
+          @PathParam("keyspaceName")
+          final String keyspaceName,
+      @ApiParam(value = "Name of the table to use for the request.", required = true)
+          @PathParam("tableName")
+          final String tableName,
+      @ApiParam(
+              value =
+                  "Value from the primary key column for the table. Define composite keys by separating values with slashes (`val1/val2...`) in the order they were defined. </br> For example, if the composite key was defined as `PRIMARY KEY(race_year, race_name)` then the primary key in the path would be `race_year/race_name` ",
```
olim7t, line 128: Can we split this with line breaks?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceApi.java
```diff
@@ -0,0 +1,340 @@
...
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.PathSegment;
+import javax.ws.rs.core.Response;
+
+/**
+ * Definition of REST API endpoint methods including both JAX-RS and Swagger annotations. No
+ * implementations.
+ *
+ * <p>NOTE: class annotations cannot be included in the interface (at least for JAX-RS) and must be
+ * included in the implementation class.
```
olim7t, line 32: `@Api` can be inherited, so maybe we can move it to the interface in order to have all the Swagger annotations in one place.

olim7t (approved):
👍 
I guess a point could be made that the Swagger annotations also document the code, but they're really too verbose and end up obscuring rather than helping.
So I prefer this approach to make the implementation easier to follow. If the parameters have good names, this should be enough to document their purpose.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceApi.java
tatu-at-datastax: I can try to see how Google indenter handles that.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceApi.java
tatu-at-datastax: Class annotation `@Api`? Ok, I can move that, makes sense. I only checked that JAX-RS ones cannot be moved.

tatu-at-datastax:
Agreed, it's not an optimal solution. But when just annotations produce 1/3 of all lines (partly due to auto-indenter using... lots of space) it does seem better to split things as that makes it easier to change implementation, where most work usually is.

I made suggested changes and will merge this; we can further improve aspects as things go.

Couple of things to note:

1. Only `interface` works: I tried abstract class and that did NOT work with Jersey at least (although some docs suggested it might)
2. While JAX-RS class annotations must be on impl class, Swagger seems smarter and allows annotations on either (but like JAX-RS, probably won't merge if both have them).

PR #1460, opened 2021-12-07
adding integration test using schema used in performance tests

jeffreyscarpenter:
Adding integration test using a schema commonly used for performance testing, to ensure readiness of the SGv2 REST API Service.

### DIFF
--- a/testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
+++ b/testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
@@ -46,6 +46,7 @@ import io.stargate.web.restapi.models.TableOptions;
 import io.stargate.web.restapi.models.TableResponse;
 import java.io.IOException;
 import java.lang.reflect.Method;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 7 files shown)


testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -2821,6 +2822,105 @@ public void testMixedCaseTable() throws IOException {
...
+    String timestamp = String.valueOf(Instant.now().getEpochSecond());
+    String serialNumber = "ABC123";
+    Map<String, String> row = new HashMap<>();
+    row.put("hour_created", timestamp);
+    row.put("serial_number", serialNumber);
+    row.put("machine_code", "DEF456");
+    row.put("part_name", "Engine");
+    row.put("part_number", "DEF456");
+    row.put("last_inspected_at", timestamp);
+    row.put("times_inspected", "2");
```
tatu-at-datastax, line 2892: While we are allowed to pass everything as String, we can also pass actual number. So would it make sense to try out both, either as two tests or two rows in same test?

tatu-at-datastax (approved):

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
jeffreyscarpenter: added an additional row insert to this test

PR #1461, opened 2021-12-07
Implement index operations in sgv2-restapi

olim7t:
With the new string-only query builder. This passes all index-related tests in `RestApiv2Test`.

### DIFF
--- a/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/models/Sgv2IndexAddRequest.java
+++ b/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/models/Sgv2IndexAddRequest.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.sgv2.restsvc.models;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.stargate.sgv2.common.cql.builder.CollectionIndexingType;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+import java.util.Map;
+import javax.validation.constraints.NotNull;
+
+@ApiModel("IndexAddRequest")
+public class Sgv2IndexAddRequest {
+
+  @NotNull private String column;
+  private String name;
+  private String type;
+  private CollectionIndexingType kind;
+  private Map<String, String> options;
+
+  private boolean ifNotExists = false;
+
+  public void setColumn(String column) {
+    this.column = column;
+  }
+
+  @ApiModelProperty(required = true, value = "Column name")
+  public String getColumn() {
+    return column;
+  }
+
+  public void setName(String name) {
+    this.name = name;
+  }
+
+  @ApiModelProperty(
+      value =
+          "Optional index name. If no name is specified, Cassandra names the index: table_name_column_name_idx.")
+  public String getName() {
+    return name;
+  }
+
+  public void setIfNotExists(boolean ifNotExists) {
+    this.ifNotExists = ifNotExists;
+  }
+
+  @ApiModelProperty(
+      value =
+          "Determines whether to create a new index if an index with the same name exists. Attempting to create an existing index returns an error unless this option is true.")
+  public boolean getIfNotExists() {
+    return ifNotExists;
+  }
+
+  public void setType(String type) {
+    this.type = type;
+  }
+
+  @ApiModelProperty(value = "A custom index class name or classpath.")
+  public String getType() {
+    return type;
+  }
+
+  public void setKind(CollectionIndexingType kind) {
+    this.kind = kind;
+  }
+
+  @JsonProperty("kind")
+  @ApiModelProperty(value = "The kind (ENTRIES, KEY, VALUES, FULL) of an index")
+  public CollectionIndexingType getKind() {
+    return kind;
+  }
+
+  @ApiModelProperty(value = "Options passed to a custom index")
+  public Map<String, String> getOptions() {
+    return options;
+  }
+
+  public void setOptions(Map<String, String> options) {
+    this.options = options;
+  }
+}
--- a/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/ResourceBase.java
+++ b/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/ResourceBase.java
@@ -5,13 +5,19 @@ import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.databind.json.JsonMapper;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.BytesValue;
+import com.google.protobuf.Int32Value;
 import io.stargate.proto.QueryOuterClass;
 import io.stargate.proto.Schema;
+import io.stargate.proto.StargateGrpc;
 import io.stargate.sgv2.restsvc.grpc.BridgeProtoValueConverters;
+import io.stargate.sgv2.restsvc.grpc.FromProtoConverter;
 import io.stargate.sgv2.restsvc.grpc.ToProtoConverter;
 import io.stargate.sgv2.restsvc.models.RestServiceError;
+import io.stargate.sgv2.restsvc.models.Sgv2RowsResponse;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Base64;
+import java.util.List;
 import java.util.Map;
 import javax.ws.rs.core.Response;
 
@@ -22,6 +28,7 @@ import javax.ws.rs.core.Response;
 public abstract class ResourceBase {
   protected static final ObjectMapper JSON_MAPPER = new JsonMapper();
   protected static final ObjectReader MAP_READER = JSON_MAPPER.readerFor(Map.class);
+  protected static final int DEFAULT_PAGE_SIZE = 100;
 
   protected static final BridgeProtoValueConverters PROTO_CONVERTERS =
       BridgeProtoValueConverters.instance();
@@ -101,5 +108,59 @@ public abstract class ResourceBase {
 
   protected static final boolean isStringEmpty(String str) {
     return (str == null) || str.isEmpty();
+  }
+
+  protected Response fetchRows(
+      StargateGrpc.StargateBlockingStub blockingStub,
+      int pageSizeParam,
+      String pageStateParam,
+      boolean raw,
+      String cql,
+      QueryOuterClass.Values.Builder values) {
+    QueryOuterClass.QueryParameters.Builder paramsB = parametersBuilderForLocalQuorum();
+    if (!isStringEmpty(pageStateParam)) {
+      paramsB =
+          paramsB.setPagingState(BytesValue.of(ByteString.copyFrom(decodeBase64(pageStateParam))));
+    }
+    int pageSize = DEFAULT_PAGE_SIZE;
+    if (pageSizeParam > 0) {
+      pageSize = pageSizeParam;
+    }
+    paramsB = paramsB.setPageSize(Int32Value.of(pageSize));
+
+    QueryOuterClass.Query.Builder b =
+        QueryOuterClass.Query.newBuilder().setParameters(paramsB).setCql(cql);
+    if (values != null) {
+      b = b.setValues(values);
+    }
+    final QueryOuterClass.Query query = b.build();
+    return Sgv2RequestHandler.handleMainOperation(
+        () -> {
+          QueryOuterClass.Response grpcResponse = blockingStub.executeQuery(query);
+
+          final QueryOuterClass.ResultSet rs = grpcResponse.getResultSet();
+          final int count = rs.getRowsCount();
+
+          String pageStateStr = extractPagingStateFromResultSet(rs);
+          List<Map<String, Object>> rows = convertRows(rs);
+          Object response = raw ? rows : new Sgv2RowsResponse(count, pageStateStr, rows);
+          return jaxrsResponse(Response.Status.OK).entity(response).build();
+        });
+  }
+
+  private static byte[] decodeBase64(String base64encoded) {
+    // TODO: error handling
+    return Base64.getDecoder().decode(base64encoded);
+  }
+
+  protected List<Map<String, Object>> convertRows(QueryOuterClass.ResultSet rs) {
+    FromProtoConverter converter =
+        BridgeProtoValueConverters.instance().fromProtoConverter(rs.getColumnsList());
+    List<Map<String, Object>> resultRows = new ArrayList<>();
+    List<QueryOuterClass.Row> rows = rs.getRowsList();
+    for (QueryOuterClass.Row row : rows) {
+      resultRows.add(converter.mapFromProtoValues(row.getValuesList()));
+    }
+    return resultRows;
   }
 }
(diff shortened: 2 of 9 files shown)


sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2IndexesResourceImpl.java
```diff
@@ -0,0 +1,177 @@
...
+          BridgeSchemaClient.create(blockingStub).findTable(keyspaceName, tableName);
+
+          String cql =
+              new QueryBuilder()
+                  .select()
+                  .from("system_schema", "indexes")
+                  .where("keyspace_name", Predicate.EQ, keyspaceName)
+                  .where("table_name", Predicate.EQ, tableName)
+                  .build();
+          return fetchRows(blockingStub, -1, null, true, cql, null);
```
olim7t, line 75: Per the v1 implementation.
This is inconsistent with other DDL resources, which return dedicated representations rather than the raw contents of the system table. But if we want full backward compatibility this is it.

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -472,8 +472,7 @@ public void createIndex(CqlSession session) throws IOException {
 
     ApiError response = objectMapper.readValue(body, ApiError.class);
     assertThat(response.getCode()).isEqualTo(HttpStatus.SC_BAD_REQUEST);
-    assertThat(response.getDescription())
-        .isEqualTo("Bad request: An index named test_idx already exists");
+    assertThat(response.getDescription()).contains("Index test_idx already exists");
```
olim7t, line 475: A few error messages are slightly different. I think that's acceptable as long as the status code is the same.

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -538,7 +537,7 @@ public void createCustomIndex(CqlSession session) throws IOException {
     Map<String, String> optionsReturned = row.get().getMap("options", String.class, String.class);
 
     assertThat(optionsReturned.get("class_name")).isEqualTo(indexType);
-    assertThat(optionsReturned.get("target")).isEqualTo("\"lastname\"");
+    assertThat(optionsReturned.get("target")).isEqualTo("lastname");
```
olim7t, line 540: That one puzzles me. I'm pretty sure unquoted is the right answer, that's what I get when creating the index via cqlsh. I don't know how it can end up quoted in the v1 test.

testing/src/main/java/io/stargate/it/http/RestApiExtension.java
```diff
@@ -228,6 +230,14 @@ private Instance(StargateEnvironmentInfo stargateEnvironmentInfo, RestApiParamet
         cmd.addArgument("-D" + e.getKey() + "=" + e.getValue());
       }
 
+      if (isDebug()) {
+        int debuggerPort = 5200;
+        cmd.addArgument(
+            "-agentlib:jdwp=transport=dt_socket,server=n,suspend=y,"
+                + "address=localhost:"
+                + debuggerPort);
+      }
```
olim7t, line 239: A bit off-topic, but this is nice to have as it allows debugging the REST server in integration tests (as explained [here](https://github.com/stargate/stargate/blob/master/DEV_GUIDE.md#debugging-integration-tests)).

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2IndexesResourceImpl.java
jeffreyscarpenter: Should we add a comment to note this is for backward compatibility?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/models/Sgv2IndexAddRequest.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+package io.stargate.sgv2.restsvc.models;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.stargate.sgv2.common.cql.builder.CollectionIndexingType;
+import io.swagger.annotations.ApiModelProperty;
+import java.util.Map;
+import javax.validation.constraints.NotNull;
+
+public class Sgv2IndexAddRequest {
```
tatu-at-datastax, line 24: I think we'll need to add `@ApiModel("IndexAddRequest") -- or whatever external name we want -- for Swagger to use. Otherwise will use unqualified class name as external name.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2IndexesResourceImpl.java
tatu-at-datastax: Could update this google doc:

https://docs.google.com/document/d/159QfQ1X3BbDErBNB5QSkg3f7j73RW2dbwD8zVZ90sGc/edit?pli=1

or create new one.

I agree we should collect open issues/concerns somewhere.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2IndexesResourceImpl.java
```diff
@@ -0,0 +1,177 @@
...
+              grpcFactory
+                  .constructBlockingStub()
+                  .withCallCredentials(new StargateBearerToken(token));
+
+          Schema.CqlTable table =
+              BridgeSchemaClient.create(blockingStub).findTable(keyspaceName, tableName);
+          if (table.getColumnsList().stream().noneMatch(c -> columnName.equals(c.getName()))) {
+            return jaxrsServiceError(
+                    Response.Status.NOT_FOUND,
+                    String.format("Column '%s' not found in table.", columnName))
```
tatu-at-datastax, line 112: Maybe add name of the table as well?

tatu-at-datastax (approved):

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
tatu-at-datastax: This is always bit of gray area; I agree that users should not rely on failure message details. We haven't really documented those anywhere either, outside of tests.

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
tatu-at-datastax: Yeah quoted seems wrong... unless it'd be for reason like retaining casing?

@jeffreyscarpenter Do you think this might have been `"lastName"` in SGv1, changed to all lower-case for SGv2?

olim7t: Right, it was mixed case on `master`.
I've reverted to the original code, the v2 implementation handles case the same.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/models/Sgv2IndexAddRequest.java
olim7t: Is that model name surfaced anywhere? I don't see it in the [generated docs](https://stargate.io/docs/stargate/1.0/attachments/restv2.html#operation/createIndex).

tatu-at-datastax: /swagger-ui endpoint exposes it under `Models`.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2IndexesResourceImpl.java
jeffreyscarpenter: Suggest we move the contents of that gdoc to a discussion post

testing/src/main/java/io/stargate/it/http/RestApiExtension.java
jeffreyscarpenter: cool!

jeffreyscarpenter (approved):

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2IndexesResourceImpl.java
tatu-at-datastax: @jeffreyscarpenter +1 -- does copy-paste work well between the two?

jeffreyscarpenter: I recommend the Google Docs plugin called "Docs to Markdown" by Ed Bacher which you can get in the Google workspace marketplace.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/ResourceBase.java
```diff
@@ -102,4 +109,56 @@ protected static final boolean isAuthTokenInvalid(String authToken) {
...
+                  ByteString.copyFrom(Sgv2RowsResourceImpl.decodeBase64(pageStateParam))));
+    }
+    int pageSize = DEFAULT_PAGE_SIZE;
+    if (pageSizeParam > 0) {
+      pageSize = pageSizeParam;
+    }
+    paramsB = paramsB.setPageSize(Int32Value.of(pageSize));
+
+    QueryOuterClass.Query.Builder b =
+        QueryOuterClass.Query.newBuilder().setParameters(paramsB.build()).setCql(cql);
```
mpenick, line 135: Tip: These builder methods support passing the builder w/o `.build()`.

```diff
-        QueryOuterClass.Query.newBuilder().setParameters(paramsB.build()).setCql(cql);
+        QueryOuterClass.Query.newBuilder().setParameters(paramsB).setCql(cql);
```

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/ResourceBase.java
```diff
@@ -102,4 +109,56 @@ protected static final boolean isAuthTokenInvalid(String authToken) {
...
+  protected Response fetchRows(
+      StargateGrpc.StargateBlockingStub blockingStub,
+      int pageSizeParam,
+      String pageStateParam,
+      boolean raw,
+      String cql,
+      QueryOuterClass.Values.Builder values) {
+    QueryOuterClass.QueryParameters.Builder paramsB = parametersBuilderForLocalQuorum();
+    if (!isStringEmpty(pageStateParam)) {
+      // surely there must better way to make Protobuf accept plain old byte[]? But if not:
```
mpenick, line 122: I really think this is the best you can do here esp. because of the `ByteValue` wrapper.

olim7t: I just moved this code but got it, I'll remove the comment.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2IndexesResourceImpl.java
olim7t: This is the message that was used in v1.

tatu-at-datastax: Ah ok. Yeah, then let's use that.

Would be good to go over error messages in future but not right now. Thanks!

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2IndexesResourceImpl.java
```diff
@@ -0,0 +1,177 @@
...
+          if (isStringEmpty(tableName)) {
+            return jaxrsBadRequestError("tableName must be provided").build();
+          }
+
+          StargateGrpc.StargateBlockingStub blockingStub =
+              grpcFactory
+                  .constructBlockingStub()
+                  .withCallCredentials(new StargateBearerToken(token));
+
+          BridgeSchemaClient.create(blockingStub).findTable(keyspaceName, tableName);
```
ivansenic, line 66: How expensive is to create stub and client on every request? I know it depends on the `token` but wander if there is better solution, or is this how the grpc would work by default?

Also, seems that this is actually gonna be a huge perf issue, let me try to explain, maybe I am wrong:
* In the `GrpcClientFactory` you construct the `ManagedChannelBuilder` and set it to use the `directExecutor`
* The Javadoc there is clear `Execute application code directly in the transport thread. Depending on the underlying transport, using a direct executor may lead to substantial performance improvements. However, it also requires the application to not block under any circumstances.`, but I am not sure how this relates to the gRPC clients
* Then you have the blocking stub in all calls, which
    * makes me think `directExecutor` should not be used (not sure)
    * will make a context switch every time you make a call.. I experienced this issue with the docs api perf tests, you have a bunch of "http" threads accepting the requests and then they block on calling the bridge, it's massive context switch all the time

I don't understand why are we not using the standard async stub here? You would just need to wire the `AsyncResponse` and have an implementation of the `StreamObserver` that actually reacts on the gRPC response, transforms to the HTTP response and pass it to the Async response.. We do something similar in the [AsyncObserver](https://github.com/stargate/stargate/blob/c44bcb22c4b6175de89f636f8fc09c8881554c30/restapi/src/main/java/io/stargate/web/docsapi/resources/async/AsyncObserver.java#L36) but for the Rx.. Note that some frameworks provide OOB support and wrappers for Stubs in order to bridge them to Rx or Reactor, meaning that it takes effort for users to actually implement and bridge HTTP async response to gRPC stream observer and vice versa..

I guess this has nothing to do with this PR but is a bigger issue with the V2.. If we are to measure the performance of the new architecture we should avoid mistakes done in V1.. And V1 is blocking all the way..

ivansenic (commented):
Left a small comment about the blocking stub usage..

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2IndexesResourceImpl.java
tatu-at-datastax: Our first goal is to prove that functionality works, and see what performance is. Then we can focus on improvements -- schema access, for example, is probably going to be too expensive (as implemented) to be feasible now.

I think we really should get blocking variant working with limited amount of work and then figure out how to convert.
If there are simple improvements to use of blocking style those could be added first of course.

Also need to consider that JAX-RS is fully blocking so we need to make sure that async operation on gRPC access actually measurable helps. I know async approach can help with massively parallel operation, and SGv2 is proxy-like enough to fit into that pattern. But just because something is non-blocking does not automatically mean it is more efficient so need to make sure we compare things.

olim7t: > I don't understand why are we not using the standard async stub here? You would just need to wire the `AsyncResponse` and have an implementation of the `StreamObserver` that actually reacts on the gRPC response, transforms to the HTTP response and pass it to the Async response..

Agreed, I plan to look at that next.

> Also need to consider that JAX-RS is fully blocking

How so? With `@Suspended` and `AsyncResponse` it's possible to implement non-blocking resources.

ivansenic: Some additional observations I made:

* the thread-per-request model can actually be fine here, as besides some basic processing of the request the threads does a call outside this JVM and blocks, which actually is fine (opposite to V1, which blocks so that other thread does the expensive CPU work in the same JVM)
* so Jetty used by Dropwizard is actually blocking, although they do support asynchronous processing.. To be fully reactive you need to have whatever `reactive-netty` or `underthrow` as the server.. mapping this to the async stub will provide benefits with to performance and resource used, I am certain of this..
* I am fine sticking with blocking until we actually move away from Dropwizard.. Implementing async processing did not showcase heavy improvements in the docs-api V1, however this can be due to the fact of how is the async support implemented in Jetty (you still get a request per thread)

tatu-at-datastax: I agree with points above by @ivansenic . I believe that fully async operations can help with highly concurrent systems. I am also not against exploring async processing within current system, to see how simple/complex changes would be.
Just want to make sure we get benefits we want and decide which parts to replace in which order.

@olim7t yes, true, servlet API (and them Jersey) do have async processing model. But by default handling is blocking so changes would be needed.

PR #1466, opened 2021-12-08
add support to v2 REST API for additional types

jeffreyscarpenter:
Adding support for additional types referenced in performance testing - decimal, double, timestamp. See for example NoSQLBench http tests here: https://github.com/nosqlbench/nosqlbench/tree/main/driver-http/src/main/resources/activities/baselines

### DIFF
--- a/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
+++ b/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
@@ -4,10 +4,9 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.uuid.impl.UUIDUtil;
-import com.google.protobuf.ByteString;
+import io.stargate.grpc.Values;
 import io.stargate.proto.QueryOuterClass;
 import java.math.BigDecimal;
-import java.nio.ByteBuffer;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.UUID;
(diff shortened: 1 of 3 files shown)


tatu-at-datastax (approved):

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
```diff
@@ -204,6 +205,18 @@ public JsonNode jsonNodeFrom(QueryOuterClass.Value value) {
     }
   }
 
+  protected static final class DecimalCodec extends FromProtoValueCodec {
+    @Override
+    public Object fromProtoValue(QueryOuterClass.Value value) {
+      return Values.decimal(value);
+    }
```
olim7t, line 212: Do we really need this method?
It looks like every implementation is going to be a direct call to the corresponding `Values.<typeName>` method.

tatu-at-datastax: I did not check that, but if it really is `value.getDecimal()`, that'd be better?

mpenick (approved):
Looks good. Should there be some sort of `testDecimalHandling` test? Or is part of the "steel thread" to just get the current tests working?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
jeffreyscarpenter: I collapsed the method

jeffreyscarpenter:
> Looks good. Should there be some sort of `testDecimalHandling` test? Or is part of the "steel thread" to just get the current tests working?

I extended the integration tests to leverage `decimal` types. But we probably do need to start creating unit tests for this service soon.

PR #1467, opened 2021-12-08
Extract gRPC stub creation to a filter

olim7t:
Avoids repeating the same 5 lines in every method.

### DIFF
--- a/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/impl/GrpcStubFactory.java
+++ b/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/impl/GrpcStubFactory.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.sgv2.restsvc.impl;
+
+import io.stargate.proto.StargateGrpc.StargateBlockingStub;
+import io.stargate.sgv2.restsvc.resources.AuthenticationFilter;
+import javax.inject.Inject;
+import javax.ws.rs.container.ContainerRequestContext;
+import org.glassfish.hk2.api.Factory;
+
+/**
+ * Provides the stub that we created in {@link AuthenticationFilter} to the Jersey context, so that
+ * it can be injected into resource methods with {@code @Context}.
+ */
+public class GrpcStubFactory implements Factory<StargateBlockingStub> {
+
+  private final ContainerRequestContext context;
+
+  @Inject
+  public GrpcStubFactory(ContainerRequestContext context) {
+    this.context = context;
+  }
+
+  @Override
+  public StargateBlockingStub provide() {
+    return (StargateBlockingStub) context.getProperty(AuthenticationFilter.GRPC_STUB_KEY);
+  }
+
+  @Override
+  public void dispose(StargateBlockingStub instance) {
+    // intentionally empty
+  }
+}
--- a/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/impl/GrpcStubFactory.java
+++ b/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/impl/GrpcStubFactory.java
@@ -16,13 +16,13 @@
 package io.stargate.sgv2.restsvc.impl;
 
 import io.stargate.proto.StargateGrpc.StargateBlockingStub;
-import io.stargate.sgv2.restsvc.resources.AuthenticationFilter;
+import io.stargate.sgv2.restsvc.resources.CreateGrpcStubFilter;
 import javax.inject.Inject;
 import javax.ws.rs.container.ContainerRequestContext;
 import org.glassfish.hk2.api.Factory;
 
 /**
- * Provides the stub that we created in {@link AuthenticationFilter} to the Jersey context, so that
+ * Provides the stub that we created in {@link CreateGrpcStubFilter} to the Jersey context, so that
  * it can be injected into resource methods with {@code @Context}.
  */
 public class GrpcStubFactory implements Factory<StargateBlockingStub> {
@@ -36,7 +36,7 @@ public class GrpcStubFactory implements Factory<StargateBlockingStub> {
 
   @Override
   public StargateBlockingStub provide() {
-    return (StargateBlockingStub) context.getProperty(AuthenticationFilter.GRPC_STUB_KEY);
+    return (StargateBlockingStub) context.getProperty(CreateGrpcStubFilter.GRPC_STUB_KEY);
   }
 
   @Override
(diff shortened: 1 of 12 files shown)


sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/impl/GrpcStubFactory.java
```diff
@@ -0,0 +1,46 @@
...
+import io.stargate.sgv2.restsvc.resources.AuthenticationFilter;
+import javax.inject.Inject;
+import javax.ws.rs.container.ContainerRequestContext;
+import org.glassfish.hk2.api.Factory;
+
+/**
+ * Provides the stub that we created in {@link AuthenticationFilter} to the Jersey context, so that
+ * it can be injected into resource methods with {@code @Context}.
+ */
+public class GrpcStubFactory implements Factory<StargateBlockingStub> {
```
olim7t, line 28: This is slightly more convoluted than I'd like.
I wish there was a way to register the stub in the request scope directly from the filter, but it seems like we need this additional factory.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/AuthenticationFilter.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.Provider;
+
+/**
+ * Validates that the auth token is present and pre-builds the gRPC stub before each request.
+ *
+ * <p>Note that the resource class or method must be annotated with {@link Authenticated} in order
+ * for this filter to run.
```
olim7t, line 31: This is necessary because the Swagger and health resources must not require the token.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceApi.java
```diff
@@ -37,6 +39,13 @@
...
     consumes = MediaType.APPLICATION_JSON,
     tags = {"data"})
+@ApiImplicitParams({
+  @ApiImplicitParam(
+      name = "X-Cassandra-Token",
+      paramType = "header",
+      value = "The token returned from the authorization endpoint. Use this token in each request.",
+      required = true)
+})
 public interface Sgv2RowsResourceApi {
```
olim7t, line 49: Thankfully this works at the class level, we don't need to repeat it for every method. I've checked the generated Swagger docs.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/impl/GrpcStubFactory.java
tatu-at-datastax: Quick question: since `Factory` comes from Glassfish, does this mean that the registration mechanism is Jersey-specific? Which would mean it is one more thing to port over to another framework if we choose to move to something like Quarkus.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceApi.java
```diff
@@ -1,10 +1,13 @@
 package io.stargate.sgv2.restsvc.resources;
 
 import com.codahale.metrics.annotation.Timed;
+import io.stargate.proto.StargateGrpc;
 import io.stargate.sgv2.restsvc.models.RestServiceError;
 import io.stargate.sgv2.restsvc.models.Sgv2RESTResponse;
 import io.stargate.sgv2.restsvc.models.Sgv2RowsResponse;
 import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiImplicitParam;
+import io.swagger.annotations.ApiImplicitParams;
```
tatu-at-datastax, line 10: Duplicate import?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceApi.java
tatu-at-datastax: Nice!

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Authenticated.java
```diff
@@ -0,0 +1,25 @@
...
+package io.stargate.sgv2.restsvc.resources;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import javax.ws.rs.NameBinding;
+
+/** Annotates resources that we want filtered with {@link AuthenticationFilter}. */
+@NameBinding
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Authenticated {}
```
tatu-at-datastax, line 25: Maybe `StargateAuthenticated`?

tatu-at-datastax:
Looks good, I like it!

One question: how does `@Authenticated` get wired to provider? I can follow everything else but that mapping...
Maybe it is somehow based on `@NameBinding`?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/AuthenticationFilter.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @see GrpcStubFactory
+ */
+@Provider
+@Authenticated
```
tatu-at-datastax, line 36: Ah. So this is where "name-binding" (see https://stackoverflow.com/questions/38520881/using-name-binding-annotations-in-jersey) is applied...

tatu-at-datastax (approved):

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceApi.java
olim7t: No, the second one has an extra "s" at the end 😃

olim7t:
> One question: how does `@Authenticated` get wired to provider?

The filter is also annotated with `@Authenticated`, that's how it gets associated with it.

https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest/filters-and-interceptors.html#d0e9931

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceApi.java
tatu-at-datastax: Ah. Yes. :)

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Authenticated.java
olim7t: Yeah I wasn't sure about the name, it doesn't actually perform authentication, just check that the token is present. I might go with `@CreateGrpcStub`, which is more descriptive.

tatu-at-datastax: Sounds good to me, more specific to actual functioning.

olim7t:
Force-pushed to rebase on latest v2.0.0 and adapt the new index resource.
I also renamed the annotation.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/impl/GrpcStubFactory.java
olim7t: Yes, I think it's implementation-specific.
With that being said, other Glassfish types are already referenced in `RestServiceServer`, so there will be a small porting effort no matter what.

tatu-at-datastax: Agreed.

PR #1469, opened 2021-12-09
Support a gRPC bi-directional streaming API

tomekl007:
**What this PR does**:
It provides bi-directional streaming support for gRPC.
1. It creates new methods in `proto` that returns and accept streams: https://github.com/stargate/stargate/pull/1469/files#diff-8b8c8408217f2daf34c555cd1d4b708e0eeb20b48b012a3945fad5f66a8f9c89R30-R37
2. Provides an implementation for both methods:
https://github.com/stargate/stargate/pull/1469/files#diff-7f7a21a1c4175ab8da70285ca06f2bffccec8b59fbb74e91bde9132f9426aa00R81-R111
3. Adds reactive support: https://github.com/stargate/stargate/pull/1469/files#diff-686679d482e4dea27fb8737d4028019563846a3a8f1d88871a0f4b9002ac70b5R18-R38 leveraging the `Reactor` framework.

**Open questions:**
- do we want to wait until all client submitted queries via the `onNext` are processed before completing the `Observer` (`onComplete` and aggregating in-flight)? For more info, see this: https://github.com/stargate/stargate/pull/1469#discussion_r772934106 - **decided and implemented**
- Do we want to create a separate module for gRPC reactive to not pollute client applications with reactive dependencies if they don't need them? see this: https://github.com/stargate/stargate/pull/1469#discussion_r773040936 - **will be done in a separate task**

Please ignore all `System.print.line` calls. I will remove them once the PR will be reviewed and ready. 

**Checklist**
- [x] handle errors better (i.e.: next, error, next) should work without closing the stream via the `onError`
- [x] validate ordering of responses in ITs
- [x] Changes manually tested


grpc/src/main/java/io/stargate/grpc/service/GrpcService.java
```diff
@@ -70,6 +71,49 @@ public void executeBatch(Batch batch, StreamObserver<Response> responseObserver)
...
+
+      @Override
+      public void onCompleted() {
+        System.out.println(
+            "onCompleted server side "
+                + " thread: "
+                + Thread.currentThread().getName()
+                + " inFlight: "
+                + inFlight.get());
+        while (inFlight.get() > 0) {
```
tomekl007, line 108: java-driver reactive support checks if all emitted events were processed, and only then sends the `onComplete`:
https://github.com/datastax/java-driver/blob/b7e384ea2500c1d0ef9cc882604128e6a486b2da/core/src/main/java/com/datastax/dse/driver/internal/core/cql/reactive/ReactiveResultSetSubscription.java#L239-L240

tomekl007: According to reactive specification:
```
An Observable may emit no items at all. An Observable may also never terminate with either an OnCompleted or an OnError notification. That is to say that it is proper for an Observable to issue no notifications, to issue only an OnCompleted or an OnError notification, or to issue only OnNext notifications.
```
[The Contract Governing Notifications
](https://reactivex.io/documentation/contract.html)
So we may consider not supporting `onComplete` from our side and have never terminating `StreamObserver`

grpc-proto/pom.xml
```diff
@@ -54,6 +68,15 @@
           <protocArtifact>com.google.protobuf:protoc:3.12.0:exe:${os.detected.classifier}</protocArtifact>
           <pluginId>grpc-java</pluginId>
           <pluginArtifact>io.grpc:protoc-gen-grpc-java:1.35.0:exe:${os.detected.classifier}</pluginArtifact>
+          <protocPlugins>
+            <protocPlugin>
+              <id>reactor-grpc</id>
```
tomekl007, line 73: we might consider creating a dedicated module `grpc-proto-reactive` for reactive gRPC support to not require our clients to add import reactive libraries (via transitive dependencies) if they do not need reactive support.

grpc/src/main/java/io/stargate/grpc/service/GrpcService.java
```diff
@@ -61,13 +62,86 @@ public void executeQuery(Query query, StreamObserver<Response> responseObserver)
...
+          if (exceptionHandler.getExceptionOccurred()) {
+            return;
+          }
+          // todo do we want to wait until in-flight is empty?
```
tomekl007, line 139: we may add some delay between checks

grpc/src/main/java/io/stargate/grpc/service/StreamingBatchHandler.java
```diff
@@ -0,0 +1,32 @@
...
+      Persistence persistence,
+      StreamObserver<QueryOuterClass.Response> responseObserver,
+      AtomicLong inFlight,
+      ExceptionHandler exceptionHandler) {
+    super(batch, connection, persistence, responseObserver, exceptionHandler);
+    this.inFlight = inFlight;
+  }
+
+  @Override
+  protected synchronized void setSuccess(QueryOuterClass.Response response) {
```
tomekl007, line 23: One important find, according to docs https://grpc.github.io/grpc-java/javadoc/io/grpc/stub/StreamObserver.html:  
```
Since individual StreamObservers are not thread-safe, 
if multiple threads will bewriting to a StreamObserver concurrently, 
the application must synchronize calls.
```

We are returning a `StreamObserver` to a caller. All coordination between client and server is done via this one `StreamObserver`. 

Unfortunately, our threading model allows executing `onNext()` on different netty threads. One example output for `Query`:
```
INFO  [Exec Stream Pumper] 2021-12-21 15:55:20,368 ProcessRunner.java:78 - stargate0-0> Executing: RESULT PREPARED cf6d6092b675cffd17f124d83fc57ed8 io.stargate.db.Result$PreparedMetadata@72c1727f (resultMetadata=io.stargate.db.Result$ResultMetadata@60f7c308) idIdempotent=true
INFO  [Exec Stream Pumper] 2021-12-21 15:55:20,381 ProcessRunner.java:78 - stargate0-0> executeWithRetry#setSuccess:  Thread:Native-Transport-Requests-2
INFO  [Exec Stream Pumper] 2021-12-21 15:55:20,381 ProcessRunner.java:78 - stargate0-0> buildResponse: EMPTY RESULT Thread:Native-Transport-Requests-3
INFO  [Exec Stream Pumper] 2021-12-21 15:55:20,381 ProcessRunner.java:78 - stargate0-0> executeWithRetry#setSuccess:  Thread:Native-Transport-Requests-3
```
Note that different threads `Thread:Native-Transport-Requests-2` and `Thread:Native-Transport-Requests-3` are calling `setSuccess` (and therefore `onNext`).

and for `Batch`:
```
INFO  [Exec Stream Pumper] 2021-12-21 16:05:57,741 ProcessRunner.java:78 - stargate0-0> executeWithRetry#setSuccess:  Thread:Native-Transport-Requests-1
INFO  [Exec Stream Pumper] 2021-12-21 16:05:57,742 ProcessRunner.java:78 - stargate0-0> executeWithRetry#setSuccess:  Thread:Native-Transport-Requests-2
INFO  [Exec Stream Pumper] 2021-12-21 16:05:57,742 ProcessRunner.java:78 - stargate0-0> SetSuccess for Batch: Thread: Native-Transport-Requests-1
INFO  [Exec Stream Pumper] 2021-12-21 16:05:57,742 ProcessRunner.java:78 - stargate0-0> SetSuccess for Batch: Thread: Native-Transport-Requests-2
```
I've started noticing this problem because of this error: https://github.com/grpc/grpc-java/issues/4581. 
It manifests only when running `Batch` queries (and `ExecuteBatchStreamingTest`)
I've tried solving this issue by synchronizing `responseObserver.onNext(response);` (in `StreamingBatchHandler` and `StreamingQueryHandler`) but it does not solve the issue. We might need to rethink and re-work our threading model for bi-directional streaming.

olim7t: `streamObserver` is shared across multiple `StreamingBatchHandler`s. Making the method synchronized makes each handler synchronize on itself. This is not enough, you need to use the same object.

This should fix the problem:
```java
synchronized (responseObserver) {
  responseObserver.onNext(response);
}
```

However I'm not a big fan of acquiring a lock in `setSuccess`, because it runs on one of the Persistence's native protocol threads. It's kind of unfortunate that gRPC imposes that on the client application, it makes it hard to write fully async server implementations. I'm not sure if there is a good alternative, at the very least we should evaluate what `onNext` does, to determine how long the thread might be blocked.

tomekl007: I've confirmed that this is a problem by changing the 
```
executor = SHARED.newExecutor(1, Integer.MAX_VALUE, "transport", "Native-Transport-Requests");
```
in the `Cassandra311Persistence` - it imposes that all `onNext` are called on the same thread. In this situation, there is no error.

The synchronization on `responseObserver` works, thank you! However, I am not sure about the performance impact of it because as you mentioned, we will block the C* persistence thread. We need to performance test it for sure.

Maybe we could consider changing the threading model for `gRPC`, to make prepare executing on the `grpc-netty-worker` thread, instead of Cassandra native-thread.
Current Context:
1. The incoming gRPC request is executed on the `netty-grpc-thread`
2. At some point, the thread is switched to Cassandra internal thread (native-transport) it's during the `prepare()` method call on persistence
3. after the `prepare()`, all subsequent executions are executed using `native-transport thread`, and not `netty-grpc-thread`.
It’s this method:
```
private CompletionStage<Prepared> maybePrepared(String cql, @Nullable String keyspace) {
    System.out.println("maybePrepared()" + " Thread:" + Thread.currentThread().getName());
    Parameters parameters =
        (keyspace == null)
            ? Parameters.defaults()
            : ImmutableParameters.builder().defaultKeyspace(keyspace).build();

    Prepared preparedInCache = connection.getPrepared(cql, parameters);
    //    CompletableFuture<Prepared> c = new CompletableFuture<>();

    return preparedInCache != null
        ? CompletableFuture.completedFuture(preparedInCache)
        : connection.prepare(cql, parameters); // todo here thread switches, maybe try to switch back?
  }
```
At the beginning of `maybePrepared`, the `netty-thread` is still executing this action. However, `connection.prepare` is returning CompletableFuture that is executed on C* custom thread. We could consider changing `C*prepare()` to not run on the separate executor:
```
public CompletableFuture<Result.Prepared> prepare(String query, Parameters parameters) {
      return executeRequestOnExecutor(
          parameters,
          // The queryStartNanoTime is not used by prepared message, so it doesn't really matter
          // that it's only computed now.
          System.nanoTime(),
          () -> new PrepareMessage(query));
    }

```
(note the `executeRequestOnExecutor`) but on the caller `grpc-netty` thread.
We could also try to get back the thread context (to netty-grpc thread) instead of switching to `native-cassandra`.
However, I am not sure how this can be achieved using `CompletableFuture` API.

ivansenic: @tomekl007 I would really like to understand this problem in more details and try to help if possible.

ivansenic: OK I think I get it.. 

`2.` won't work you must not block on the `netty-grpc-thread`, and if prepare goes to the data store, boom you are blocked.
`3.` is also not good, because this would decrease the max throughput to the data store and Connection is used by all APIs

So you can effectively do only `1.` Al alternative to synchronized, you can also create a new dedicated executor service with max thread size 1 and do something like

```
CompletableFuture.runAsync(() -> responseObserver.onNext(response), singleThreadedExecutor)
```

this way all interaction with the observer is done on a single thread only. However, not sure if this is more costly than that `synchronized` keyword..

tomekl007: Yes, I need firstly to performance test option 1. If it turns out that it is impacting the performance, I will experiment with a dedicated thread pool - but it may also be a bottleneck (even more significant than synchronizing `onNext()`)

mpenick: The all depends on how expensive `onNext()` , is it adding to back of queue or is waiting on a system call. If it's really cheap operation then locking is not a big deal.

tomekl007: Yes, there is also no contention between `StreamObeservers` for different clients. The contention is only for a given `StreamObserver` that is processed for a specific caller. So, in the end, the cost of contention might be negligible.

grpc-proto/pom.xml
ivansenic: this seems as a good idea..

grpc/src/main/java/io/stargate/grpc/service/BatchHandler.java
```diff
@@ -93,6 +94,8 @@ protected void validate() throws Exception {
 
   @Override
   protected CompletionStage<ResponseAndTraceId> buildResponse(Result result) {
+    System.out.println(
```
ivansenic, line 97: There are bunch of `System.out.println` calls in this PR, I guess that this will be removed?

grpc/src/main/java/io/stargate/grpc/service/ExceptionHandler.java
```diff
@@ -0,0 +1,270 @@
...
+
+  private final StreamObserver<QueryOuterClass.Response> responseObserver;
+  private final AtomicBoolean exceptionOccurred = new AtomicBoolean(false);
+
+  public ExceptionHandler(StreamObserver<QueryOuterClass.Response> responseObserver) {
+
+    this.responseObserver = responseObserver;
+  }
+
+  protected void handleException(Throwable throwable) {
```
ivansenic, line 53: can you add some inline comments explaining this if a bit?

grpc/src/main/java/io/stargate/grpc/service/ExceptionHandler.java
```diff
@@ -0,0 +1,270 @@
...
+      synchronized (responseObserver) {
+        responseObserver.onError(throwable);
+      }
+    } else if (throwable instanceof UnhandledClientException) {
+      onError(Status.UNAVAILABLE, throwable);
+    } else if (throwable instanceof PersistenceException) {
+      handlePersistenceException((PersistenceException) throwable);
+    } else {
+      LOG.error("Unhandled error returning UNKNOWN to the client", throwable);
+      synchronized (responseObserver) {
```
ivansenic, line 69: can this and synchronized on the like 60 be somehow also moved to the private method? would be nice to have synchronized in a single place only..

grpc/src/main/java/io/stargate/grpc/service/MessageStreamObserver.java
```diff
@@ -0,0 +1,88 @@
...
+            + " inFlight: "
+            + inFlight.get());
+    while (inFlight.get() > 0) {
+      if (exceptionHandler.getExceptionOccurred()) {
+        return;
+      }
+      // todo do we want to wait until in-flight is empty?
+      System.out.println("waiting while inFlight = 0");
+    }
+    synchronized (responseObserver) {
```
ivansenic, line 84: hmm this is now a second class calling it this way, how about having a `SynchronizedStreamObserver` which implements the delegate pattern and calls all the methods in the sync block?

grpc/src/main/java/io/stargate/grpc/service/MessageStreamObserver.java
```diff
@@ -0,0 +1,88 @@
...
+            + Thread.currentThread().getName()
+            + " inFlight: "
+            + inFlight.get());
+    while (inFlight.get() > 0) {
+      if (exceptionHandler.getExceptionOccurred()) {
+        return;
+      }
+      // todo do we want to wait until in-flight is empty?
+      System.out.println("waiting while inFlight = 0");
+    }
```
ivansenic, line 83: what s the deal here, you are spinning the cpu here or? should we not parnNano here or smth like this?

grpc/src/main/java/io/stargate/grpc/service/StreamingBatchHandler.java
```diff
@@ -0,0 +1,33 @@
...
+  protected void setSuccess(QueryOuterClass.Response response) {
+    System.out.println(
+        "SetSuccess for Batch: " + response + "Thread: " + Thread.currentThread().getName());
+    synchronized (responseObserver) {
+      responseObserver.onNext(response);
+    }
+    inFlight.decrementAndGet();
+    // do not invoke onComplete. The caller(client) may invoke it
+    // once it completes sending a stream of queries
+  }
```
ivansenic, line 32: should call to onNext be in a try/finally? if you have exception in this call, you will never decrement the inFlight?

grpc/src/main/java/io/stargate/grpc/service/StreamingQueryHandler.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  @Override
+  protected synchronized void setSuccess(QueryOuterClass.Response response) {
+    synchronized (responseObserver) {
+      responseObserver.onNext(response);
+    }
+    inFlight.decrementAndGet();
+    // do not invoke onComplete. The caller(client) may invoke it
+    // once it completes sending a stream of queries
```
ivansenic, line 39: same here

ivansenic (commented):
Did not yet look into tests, but there are some improvement suggestions..

grpc/src/main/java/io/stargate/grpc/service/ExceptionHandler.java
tomekl007: done in: https://github.com/stargate/stargate/pull/1469/commits/11f43ae296ac6db8dc1c989a1c2a7c9cdc947af9

grpc/src/main/java/io/stargate/grpc/service/MessageStreamObserver.java
tomekl007: great idea, done in: https://github.com/stargate/stargate/pull/1469/commits/f2d2ebacec45d0a4a2b2c26292ca000835f94e70

grpc/src/main/java/io/stargate/grpc/service/StreamingBatchHandler.java
tomekl007: good catch, done in: https://github.com/stargate/stargate/pull/1469/commits/5f01aefc8e2b69be30be295914fb437145e0f63a

grpc/src/main/java/io/stargate/grpc/service/MessageStreamObserver.java
tomekl007: I've implemented it using `ScheduledExecutor`, following the same pattern as we do for the `waitForSchemaAgreement`, done in:
https://github.com/stargate/stargate/pull/1469/commits/ba17b2db867d814790e045e29a63496b2ae68eda

grpc/src/main/java/io/stargate/grpc/service/SynchronizedStreamObserver.java
```diff
@@ -0,0 +1,49 @@
...
+
+  private final StreamObserver<V> streamObserver;
+
+  public SynchronizedStreamObserver(StreamObserver<V> streamObserver) {
+    this.streamObserver = streamObserver;
+  }
+
+  @Override
+  public void onNext(V value) {
+    synchronized (this) {
```
ivansenic, line 31: I would say that `synchronized(streamObserver)` is better option here.. I know this will work as well, but with my suggestion you are syncing on the `streamObserver` and you would be respecting anybody from outside this class doing the same.. I know this should not happen, but seems like a more valid design for me? wdyt?

if you accept this, of course change it everywhere and update that class level java doc

grpc/src/main/java/io/stargate/grpc/service/MessageStreamObserver.java
```diff
@@ -0,0 +1,133 @@
...
+    if (exceptionHandler.getExceptionOccurred()) {
+      return;
+    }
+
+    executor.schedule(
+        () -> waitForAllInFlightProcessed(onCompleteRetries - 1, shouldComplete),
+        DELAY_BETWEEN_ON_COMPLETE_CHECKS_MS,
+        TimeUnit.MILLISECONDS);
+    try {
+      shouldComplete.get();
```
ivansenic, line 98: What thread gets blocked here actually? And why do we actually block here..

Shouldn't this just be a signal from the client side there are no more requests? Meaning, hey there are no more requests coming, you can signal `onComplete` once you have processed all the in flight requests...

So if I may I would propose a design change here.. I would make `StreamingBatchHandler` and `StreamingQueryHandler` only pass the response back to the `MessageStreamObserver` and the stream observer can hold the whole state and act accordingly. This would also make `StreamingBatchHandler` and `StreamingQueryHandler` completely stateless, and would not have to do any `inFlight` stuff..

And I am certain this approach is actually a bug, why would you block the client side, I don't think this is going in line with the observable signaling..

testing/src/main/java/io/stargate/it/grpc/ExecuteBatchStreamingTest.java
```diff
@@ -0,0 +1,158 @@
...
+
+    // all inserted records may be not visible to the subsequent SELECT query
+    // because all reactive calls are non-blocking. Therefore, we need to wait for response of two
+    // batch insert
+    // queries
+    Awaitility.await()
+        .atMost(10, TimeUnit.SECONDS)
+        .until(
+            () -> {
+              return responses.size() == 2;
```
ivansenic, line 107: if this is async, `responses` should be synced list, like `CopyOnWriteArratList`

grpc/src/test/java/io/stargate/grpc/service/MessageStreamObserverTest.java
```diff
@@ -0,0 +1,119 @@
...
+    StreamObserver<QueryOuterClass.Response> responseStreamObserver =
+        new SynchronizedStreamObserver<>(callerStreamObserver);
+    ExceptionHandler exceptionHandler = new ExceptionHandler(responseStreamObserver);
+    BiFunction<QueryOuterClass.Response, AtomicLong, MessageHandler<QueryOuterClass.Response, ?>>
+        handlerProducer =
+            (response, inFlight) -> {
+              inFlight.decrementAndGet(); // inFlight is decremented
+              return mock(MessageHandler.class);
+            };
+    ScheduledExecutorService executor = Executors.newScheduledThreadPool(1);
```
ivansenic, line 35: I think executor should be correctly shutdown in the `@AfterEach`, you can extract as a field of the class and still init in every test with the required amount of threads

grpc/src/test/java/io/stargate/grpc/service/MessageStreamObserverTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+    MessageStreamObserver<QueryOuterClass.Response> observer =
+        new MessageStreamObserver<>(
+            responseStreamObserver, exceptionHandler, handlerProducer, executor, 10);
+
+    // when
+    observer.onNext(QueryOuterClass.Response.newBuilder().build());
+
+    // then should immediately complete and call the caller onComplete
+    assertDoesNotThrow(observer::onCompleted);
```
ivansenic, line 45: what's `assertDoesNotThrow` needed for? if it throws anything, the test will fail

grpc/src/test/java/io/stargate/grpc/service/MessageStreamObserverTest.java
```diff
@@ -0,0 +1,119 @@
...
+        new MessageStreamObserver<>(
+            responseStreamObserver, exceptionHandler, handlerProducer, executor, 10);
+
+    // when
+    observer.onNext(QueryOuterClass.Response.newBuilder().build());
+
+    // then wait for inFlight = 0 and complete
+    assertDoesNotThrow(
+        observer
+            ::onCompleted); // onCompleted will not return immediately because inFlight is not == 0
```
ivansenic, line 79: `onCompleted will not return immediately because inFlight is not == 0`

Exactly what I talked about above.. fyi if you apply my changes, you can still verify with:

```
verify(callerStreamObserver, timeout(1000).times(1)).onCompleted();
```

testing/src/main/java/io/stargate/it/grpc/ExecuteBatchStreamingTest.java
```diff
@@ -0,0 +1,158 @@
...
+        .isEqualTo(
+            new HashSet<>(
+                Arrays.asList(
+                    rowOf(Values.of("a"), Values.of(1)),
+                    rowOf(Values.of("b"), Values.of(2)),
+                    rowOf(Values.of("c"), Values.of(3)))));
+  }
+
+  @Test
+  public void streamingQueryWithError(@TestKeyspace CqlIdentifier keyspace) {
```
ivansenic, line 126: can we have a test case when first batch is OK, and second fails?

testing/src/main/java/io/stargate/it/grpc/ExecuteQueryStreamingTest.java
```diff
@@ -0,0 +1,137 @@
...
+    ResultSet rs = response.getResultSet();
+    assertThat(new HashSet<>(rs.getRowsList()))
+        .isEqualTo(
+            new HashSet<>(
+                Arrays.asList(
+                    rowOf(Values.of("a"), Values.of(1)), rowOf(Values.of("b"), Values.of(2)))));
+  }
+
+  @Test
+  public void streamingQueryWithError(@TestKeyspace CqlIdentifier keyspace) {
```
ivansenic, line 109: same here, can we have a test case when first query is OK, and second fails?

ivansenic (changes requested):
I think we have a flaw in the design here.. Would be up in discussing more if needed.. Only putting the `Request changes` so that we can re-think everything..

grpc/src/main/java/io/stargate/grpc/service/MessageStreamObserver.java
tomekl007: >Shouldn't this just be a signal from the client side there are no more requests? Meaning, hey there are no more requests coming, you can signal onComplete once you have processed all the in-flight requests

That is actually the case, the client is sending `onComplete` once all `onNext()` are emitted:
see for example: `simpleReactiveBiDirectionalQueries()`:
```
   emitter.next(
                  cqlQuery("INSERT INTO test (k, v) VALUES ('a', 1)", queryParameters(keyspace)));
              emitter.next(
                  cqlQuery(
                      "INSERT INTO test (k, v) VALUES (?, ?)",
                      queryParameters(keyspace),
                      Values.of("b"),
                      Values.of(2)));
              emitter.complete(); // client-side complete signal
```
regarding blocking `onComplete()`:

I first implemented the `onComplete` as not-blocking and without any in-flight.
However, all `next()` and once the `onComplete()` is issued by the caller, there still might be not-processed requests.
Without waiting for in-flight, that not-processed ' onNext()`s are not propagated to the caller because the stream will be closed by `complete()`. This is why decided to wait in `onComplete()`
(see this for more info: https://github.com/stargate/stargate/pull/1469#discussion_r772934106)
The other solution would be to impose on the client to track all `onNext()` and only if all of them succeeded to call the `onComplete`. I am not sure if this has a good UX. Wdyt?

ivansenic: But this is not what I am talking about.. You did good with the `inFlight` counting imo.. I just say let's switch the approach a bit:
* instead of blocking get here, set an `AtomicBoolean noMoreRequests` to true.. that's actually the only thing we need..
* change the `Streaming..Handler`s to delegate back the response to the `MessageStreamObserver` in the `setSuccess`
* based on the current state of the `noMoreRequests` and `inFlight` make a decision what to do

tomekl007: done in: 
https://github.com/stargate/stargate/pull/1469/commits/825c3dd263fa3b6798290ad059126fd32f059e16

grpc/src/test/java/io/stargate/grpc/service/MessageStreamObserverTest.java
tomekl007: done

testing/src/main/java/io/stargate/it/grpc/ExecuteBatchStreamingTest.java
tomekl007: good catch, I've implemented it and there is a problem with ordering as well, the `onError()` is called so all in-flight requests are canceled. I need to handle it by introducing a new type for streaming (either Result or Error) to not call `onError` for such a case.

tomekl007:
We need a new `StreamingResponse` type that can contain Error, proposition:
```
message StreamingResponse {
  oneof message{
    Response response = 1;
    google.rpc.Status status = 2;
  }
}
```
and in the service:
```
service Stargate {
rpc ExecuteQueryStream(stream Query) returns (stream StreamingResponse) {}
rpc ExecuteBatchStream(stream Batch) returns (stream StreamingResponse) {}
}
```

tomekl007:
Note on ordering, as per documentation:
`gRPC guarantees message ordering within an individual RPC call.`
So for bi-directional streaming, if the caller invokes two subsequent `onNext()`:
```
requestObserver.onNext(
        cqlQuery("INSERT INTO test (k, v) VALUES ('a', 1)", queryParameters(keyspace)));

    requestObserver.onNext(
        cqlQuery(
            "INSERT INTO test (k, v) VALUES (?, ?)",
            queryParameters(keyspace),
            Values.of("b"),
            Values.of(2)));
```
there is no guarantee that they will be in order as both may be executed via a separate RPC call.
If the caller needs ordering of operations, it should use our `executeBatchStream` API:
```
 QueryOuterClass.Batch.newBuilder()
                      .addQueries(cqlBatchQuery("INSERT INTO test (k, v) VALUES ('a', 1)"))
                      .addQueries(
                          cqlBatchQuery(
                              "INSERT INTO test (k, v) VALUES (?, ?)",
                              Values.of("b"),
                              Values.of(2)))
                      .setParameters(batchParameters(keyspace))
                      .build()
```
that allows executing N queries within a single RPC call.

grpc/src/main/java/io/stargate/grpc/service/SingleExceptionHandler.java
```diff
@@ -0,0 +1,29 @@
+package io.stargate.grpc.service;
+
+import io.grpc.Metadata;
+import io.grpc.Status;
+import io.grpc.stub.StreamObserver;
+import io.stargate.proto.QueryOuterClass;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class SingleExceptionHandler extends ExceptionHandler {
```
ivansenic, line 10: should we call this better `TerminatingExceptionHandler` or smth like this?

grpc/src/main/java/io/stargate/grpc/service/SuccessHandler.java
```diff
@@ -0,0 +1,7 @@
+package io.stargate.grpc.service;
+
+import io.stargate.proto.QueryOuterClass;
+
+public interface SuccessHandler {
```
ivansenic, line 5: `StreamingSuccessHandler`?

grpc/src/main/java/io/stargate/grpc/service/streaming/MessageStreamObserver.java
```diff
@@ -0,0 +1,97 @@
...
+   * item, and it may then follow those emission notifications by either an OnCompleted or an
+   * OnError notification, but not both.
+   *
+   * <p>If there was any error (server-side or client-side) we are ignoring the onComplete() signal.
+   */
+  @Override
+  public void onCompleted() {
+    clientSignalComplete.set(true);
+
+    if (inFlight.get() == 0) {
```
ivansenic, line 93: smart :+1:

grpc/src/main/java/io/stargate/grpc/service/streaming/StreamingExceptionHandler.java
```diff
@@ -0,0 +1,54 @@
...
+  protected void onError(
+      @Nullable Status status, @Nonnull Throwable throwable, @Nullable Metadata trailer) {
+    // propagate streaming error as a Status
+    successHandler.handleResponse(
+        QueryOuterClass.StreamingResponse.newBuilder()
+            .setStatus(convertStatus(status, throwable))
+            .build());
+  }
+
+  private com.google.rpc.Status convertStatus(
```
ivansenic, line 33: how this differ from the other exception handler and is this something we can contribute to the abstract class?

grpc/src/main/java/io/stargate/grpc/service/streaming/MessageStreamObserver.java
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * It is creating new handler for each request and invokes the {@link MessageHandler#handle()}
+   * method.
+   *
+   * @param value Batch or Query
+   */
+  @Override
+  public void onNext(MessageT value) {
+    inFlight.incrementAndGet();
+    streamingHandlerFactory.create(value, this, exceptionHandler).handle();
```
ivansenic, line 65: why are we creating the handler object on every request? Is this really designed correctly, that we have a handler per query? how does this influence the GC? shouldn't the query be parameter of the handle method? Or is there a specific reason for this?

ivansenic (commented):
Seems way better, there are still few small questions open.. I can give approval once we clarify those..

Some additional generic comments:
* new classes and interfaces should have a brief java docs, so that devs easier understand the class itself and how it interacts with the rest of the code..

grpc/src/main/java/io/stargate/grpc/service/streaming/MessageStreamObserver.java
tomekl007: this is the current design where the handler is per request. We would need @mpenick input on why it was designed this way. We may consider investigating re-design but this is definitely out of the scope of this ticket.

grpc/src/main/java/io/stargate/grpc/service/SuccessHandler.java
tomekl007: done

grpc/src/main/java/io/stargate/grpc/service/SingleExceptionHandler.java
tomekl007: I would prefer to have this distinction as `Single` and `Streaming`

grpc/src/main/java/io/stargate/grpc/service/streaming/MessageStreamObserver.java
ivansenic: Can we at least open an issue as a starting point?

grpc/src/main/java/io/stargate/grpc/service/SingleExceptionHandler.java
ivansenic: OK, as you wish..

grpc-proto/pom.xml
tomekl007: yeah, but it is not as simple as it sounds so I think we may consider doing it in a separate task.
I've tried to create a new module `grpc-proto-reactive`, but this module need to have both `proto` files (`stargate.proto` and `query.proto`). 
I've tried to make the new module inherit the `grpc-proto`:
```
    <dependencies>
        <dependency>
            <groupId>io.projectreactor</groupId>
            <artifactId>reactor-core</artifactId>
            <version>3.4.13</version>
        </dependency>
    </dependencies>

    <build>
        <extensions>
            <extension>
                <groupId>kr.motd.maven</groupId>
                <artifactId>os-maven-plugin</artifactId>
                <version>1.6.2</version>
            </extension>
        </extensions>
        <plugins>
            <plugin>
                <groupId>org.xolstice.maven.plugins</groupId>
                <artifactId>protobuf-maven-plugin</artifactId>
                <version>0.6.1</version>
                <configuration>
                    <protoSourceRoot>proto</protoSourceRoot>
                    <protocArtifact>com.google.protobuf:protoc:3.12.0:exe:${os.detected.classifier}</protocArtifact>
                    <pluginId>grpc-java</pluginId>
                    <pluginArtifact>io.grpc:protoc-gen-grpc-java:1.35.0:exe:${os.detected.classifier}</pluginArtifact>
                    <protocPlugins>
                        <protocPlugin>
                            <id>reactor-grpc</id>
                            <groupId>com.salesforce.servicelibs</groupId>
                            <artifactId>reactor-grpc</artifactId>
                            <version>${grpc.reactor.version}</version>
                            <mainClass>com.salesforce.reactorgrpc.ReactorGrpcGenerator</mainClass>
                        </protocPlugin>
                    </protocPlugins>
                </configuration>
                <executions>
                    <execution>
                        <goals>
                            <goal>compile</goal>
                            <goal>compile-custom</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
        </plugins>
    </build>
</project>
```
So it would have structure like `grpc-proto-reactive` -> `grpc-proto` -> `stargate parent` but it does not seem to work:
```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[WARNING] 'parent.relativePath' of POM io.stargate.grpc:grpc-proto-reactive:1.0.45-SNAPSHOT (/Users/tomasz.lelek/IntelliJ_workspace/stargate/grpc-proto-reactive/pom.xml) points at io.stargate:stargate instead of io.stargate.grpc:grpc-proto, please verify your project structure @ line 6, column 13
[FATAL] Non-resolvable parent POM for io.stargate.grpc:grpc-proto-reactive:1.0.45-SNAPSHOT: Could not find artifact io.stargate.grpc:grpc-proto:pom:1.0.45-SNAPSHOT in datastax-artifactory (https://repo.sjc.dsinternal.org/artifactory/dse) and 'parent.relativePath' points at wrong local POM @ line 6, column 13
```
Adding the:
```
  <dependency>
        </dependency>
```
as a dependency also does not work because `proto` files are not imported. We would need to figure this out.

grpc/src/main/java/io/stargate/grpc/service/streaming/MessageStreamObserver.java
tomekl007: done

ivansenic (approved):
I think we reached a state where this can be merged.. Well done @tomekl007

PR #1473, opened 2021-12-10
rework codec timestamps to use ISO string

jeffreyscarpenter:
determined that I had misunderstood how REST API was accepting timestamps - requires ISO strings


sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
```diff
@@ -301,6 +305,18 @@ private UUID uuidFrom(QueryOuterClass.Value value) {
...
 
+  protected static final class TimestampCodec extends FromProtoValueCodec {
+    @Override
+    public Object fromProtoValue(QueryOuterClass.Value value) {
+      return Instant.ofEpochMilli(value.getInt()).toString();
+    }
+
+    @Override
+    public JsonNode jsonNodeFrom(QueryOuterClass.Value value) {
+      return jsonNodeFactory.textNode(Instant.ofEpochMilli(value.getInt()).toString());
```
ivansenic, line 316: wait is this the same as the 

`jsonNodeFactory.textNode(fromProtoValue(value));`

EDIT: OK I see this would not compile due to the `Object`.. We should have made `FromProtoValueCodec` generic

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/ToProtoValueCodecs.java
```diff
@@ -384,5 +386,31 @@ public UUIDCodec(String typeDesc) {
...
+      return cannotCoerce(value);
+    }
+
+    @Override
+    public QueryOuterClass.Value protoValueFromStringified(String value) {
+      try {
+        // TODO: this implementation requires full date/time specification including timezone
+        // we could support more flexibility in format as requested in
+        // https://github.com/stargate/stargate/issues/839
+        return Values.of(Instant.parse(value).toEpochMilli());
```
ivansenic, line 408: Are we sure that this format is requested? `DateTimeFormatter.ISO_INSTANT`? Looked here https://docs.datastax.com/en/cql-oss/3.x/cql/cql_reference/timestamp_type_r.html and it seems it's fine, I was confused a bit with this section:

```
where Z is the RFC-822 4-digit time zone, expressing the time zone's difference from UTC. For example, for example the date and time of Feb 3, 2011, at 04:05:00 AM, GMT:

2011-02-03 04:05+0000
2011-02-03 04:05:00+0000
2011-02-03T04:05+0000
2011-02-03T04:05:00+0000
```

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -2862,7 +2862,7 @@ public void testTimestampHandling() throws IOException {
     assertThat(tableResponse.getName()).isEqualTo(tableName);
 
     // insert a row
-    long timestamp = Instant.now().toEpochMilli();
+    String timestamp = Instant.now().toString();
```
ivansenic, line 2865: hardcode the expected string here?

ivansenic (commented):
Can we add a unit test for the `ToProtoValueCodecs`? Seems strange that this is not tested?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/ToProtoValueCodecs.java
tatu-at-datastax: I think this code is probably copied from SGv1 REST `Converters` or some similar source. Which may or may not conform fully to expected CQL formats.

tatu-at-datastax (approved):

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
tatu-at-datastax: It is possible we might only need `JsonNode` variant but originally I was not sure; for serialization either works.
I don't think generics would necessarily have helped; my experience on things like this is that they tend to cause as much issues as they solve.

olim7t (changes requested):
For anyone reviewing along with the v1 code:
- "ToProto" codecs are equivalent to `Converters.toCqlValue(ColumnType, Object)`
- "FromProto" codecs are equivalent to `Converters.toJsonValue(Object)`

I know this is not part of this PR, but v1 converted large number to JSON strings, this is not the case here, e.g. `FromProtoValueCodecs.LongCodec`. Using JSON numbers can cause interoperability issues for large numbers. For example, javascript does not handle integers larger than 2^53 correctly.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/ToProtoValueCodecs.java
olim7t: Correct, this comes directly from the v1 code.

I think it would be a good idea to handle all the formats supported in CQL literals. We did it in the Java driver and also in the GraphQL module (see `TimestampCoercing`).

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/ToProtoValueCodecs.java
```diff
@@ -384,5 +499,126 @@ public UUIDCodec(String typeDesc) {
...
+        return Values.of((byte[]) value);
+      } else if (value instanceof ByteBuffer) {
+        return Values.of((ByteBuffer) value);
+      }
+      return cannotCoerce(value);
+    }
+
+    @Override
+    public QueryOuterClass.Value protoValueFromStringified(String value) {
+      return Values.of(value.getBytes());
```
tatu-at-datastax, line 619: Hmmmh. If String, would have to be base64-encoded. So I suspect this is wrong.

tatu-at-datastax: From SGv1 `Converters`:

```
Base64.getDecoder().decode(value)
```

is what is needed I think. May throw exception for invalid base64 encoding.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
```diff
@@ -195,6 +201,18 @@ public JsonNode jsonNodeFrom(QueryOuterClass.Value value) {
     }
   }
 
+  protected static final class FloatCodec extends FromProtoValueCodec {
+    @Override
+    public Object fromProtoValue(QueryOuterClass.Value value) {
+      return value.getFloat();
```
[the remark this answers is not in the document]
tatu-at-datastax: @olim7t Javascript is the bane of our existence, using `double` type for everything and like you say limiting integer magnitude to 53-bit (not 64). We probably should follow SGv1 pattern, even if it results in possible problems for non-JS clients that then must do String-to-long conversion unexpectedly.

I wonder if we should start a Discussion to keep track of all known possible problem cases, wrt bounds checks, conversions, leading to additional testing.
I think we may choose to address these in near future but perhaps after completing API implementation (that is may choose the order in which to tackle these currently "invisible" issues; ones that are not covered by tests)

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/ToProtoValueCodecs.java
jeffreyscarpenter: Perhaps we should move `TimestampCoercing` (and any similar classes) into the new `sgv2-service-common` module?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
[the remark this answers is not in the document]
jeffreyscarpenter: Discussion, or issue?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/ToProtoValueCodecs.java
jeffreyscarpenter: Suggest we should use `io.stargate.core.util.ByteBufferUtils' for inbound/outbound conversions, sound correct?

tatu-at-datastax: If they are general purpose yes (no proto/grpc type dependencies). Assuming this would be in sgv2-service-common?

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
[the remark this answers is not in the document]
tatu-at-datastax: Actually, maybe first bigger discussion on SGv2 REST API implementation state/completeness, by end point.
I will start one based on that earlier google doc.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
```diff
@@ -301,6 +332,66 @@ private UUID uuidFrom(QueryOuterClass.Value value) {
...
+    @Override
+    public JsonNode jsonNodeFrom(QueryOuterClass.Value value) {
+      return jsonNodeFactory.textNode(Values.inet(value).toString());
+    }
+  }
+
+  protected static final class BlobCodec extends FromProtoValueCodec {
+    @Override
+    public Object fromProtoValue(QueryOuterClass.Value value) {
+      return ByteBufferUtils.toBase64(Values.bytes(value));
```
tatu-at-datastax, line 386: Hmmh. Maybe add a note that we actually have 2 choices here:

1. `byte[]` actual decoded blob contents
2. `String`: base64-encoded textual version

I think both work for intended purpose (serialization from basic Java values) but maybe worth pointing out there is a choice being made.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/ToProtoValueCodecs.java
tatu-at-datastax: Eventually yes; maybe defer refactoring for later as I suspect we might learn more about all kinds of things we want to move around. Although that leads to the question of whether we should track such things (another GH discussion about refactoring of value conversion functionality?)

tatu-at-datastax (approved):

jeffreyscarpenter:
> Can we add a unit test for the `ToProtoValueCodecs`? Seems strange that this is not tested?

We've prioritized integration tests in the interest of speed for first milestone, we do plan to add unit tests in milestone 2.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
olim7t: I'm seeing this in the v1 code:
```
    // This covers [...] blobs (which Jackson already
    // converts to base64 natively), and time types (which are handled by registering JavaTimeModule
    // with the object mapper -- see Server.java).
    return cqlValue;
```
Why not do the same here? Just return a `ByteBuffer` and Jackson will handle it:
```
return Values.byteBuffer(value);
```

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/ToProtoValueCodecs.java
jeffreyscarpenter: This is currently in `core` but does not appear to be used in any coordinator modules on the backend. We might be able to move to `sgv2-service-common`, but I think we can defer that decision for now.

PR #1474, opened 2021-12-10
Simplify error handling

olim7t:
Rely on built-in Dropwizard mechanisms to eliminate boilerplate.
https://www.dropwizard.io/en/latest/manual/core.html#error-handling

Some of these changes might be controversial so I've isolated them into separate commits in case we decide to revert them.


sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/impl/DefaultExceptionMapper.java
```diff
@@ -0,0 +1,86 @@
...
+      if (family.equals(Response.Status.Family.REDIRECTION)) {
+        return response;
+      }
+      if (family.equals(Response.Status.Family.SERVER_ERROR)) {
+        logException(exception);
+      }
+
+      return Response.fromResponse(response)
+          .type(MediaType.APPLICATION_JSON_TYPE)
+          .entity(new RestServiceError(exception.getLocalizedMessage(), response.getStatus()))
```
olim7t, line 50: This class duplicates DropWizard's default `LoggingExceptionMapper`, the only change is that we use `RestServiceError` instead of `io.dropwizard.jersey.errors.ErrorMessage` here (and at line 61).

The reason being that our error payload has a `description` field, whereas Dropwizard's is named `message` 😞 

If we consider that that is an acceptable breaking change we can remove this class.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/impl/GrpcExceptionMapper.java
```diff
@@ -0,0 +1,93 @@
...
+import javax.ws.rs.ext.ExceptionMapper;
+
+public class GrpcExceptionMapper implements ExceptionMapper<StatusRuntimeException> {
+  @Override
+  public Response toResponse(StatusRuntimeException grpcE) {
+    Status.Code sc = grpcE.getStatus().getCode();
+
+    // Handling partly based on information from:
+    // https://developers.google.com/maps-booking/reference/grpc-api/status_codes
+    switch (sc) {
```
olim7t, line 31: Copied verbatim from `Sgv2RequestHandler` (now deleted).

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RequestHandler.java
```diff
@@ -1,167 +0,0 @@
...
-   * @param action Call to make to get {@link Response} to return in case no {@code Exception}
-   *     caught.
-   * @return {@link Response} to return from REST endpoint; may be success or failure.
-   */
-  public static Response handleMainOperation(Callable<Response> action) {
-    try {
-      return action.call();
-    } catch (StatusRuntimeException grpcE) { // from gRPC
-      return failResponseForGrpcException(grpcE);
-    } catch (NotFoundException nfe) { // does this actually occur?
```
olim7t, line 45: I doubt this is thrown by our code.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RequestHandler.java
```diff
@@ -1,167 +0,0 @@
...
-          .build();
-    }
-    // Do log underlying Exception with Stack trace since this is unknown, unexpected:
-    logger.error("Unrecognized error when executing request", e);
-    return Response.status(Response.Status.INTERNAL_SERVER_ERROR)
-        .entity(
-            new RestServiceError(
-                "Server error: " + e.getMessage(),
-                Response.Status.INTERNAL_SERVER_ERROR.getStatusCode()))
-        .build();
```
olim7t, line 156: Dropwizard already does this by default, see `DefaultExceptionMapper.logException`.

I don't think we should special case `IllegalArgumentException` et al, those should be caught by the application code and handled when they indicate a business error. Otherwise we might encounter unexpected ones and accidentally leak sensitive info in an error message.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceImpl.java
```diff
@@ -63,7 +65,7 @@ public Response getRowWithWhere(
       final String sortJson,
       final HttpServletRequest request) {
     if (isStringEmpty(where)) {
-      return jaxrsBadRequestError("where parameter is required").build();
+      throw new WebApplicationException("where parameter is required", Status.BAD_REQUEST);
```
olim7t, line 68: IMHO this equivalent in terms of readability and verbosity. In addition:
- it makes control flow easier if you want to fail from a deep code path
- I think it will be handled better when we switch to async (to be verified, but I think throwing from a method will complete the async request, so we won't need to change the code).

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceImpl.java
```diff
@@ -81,7 +83,7 @@ public Response getRowWithWhere(
 
     // !!! To Be Implemented:
 
-    return jaxrsResponse(Response.Status.NOT_IMPLEMENTED).build();
+    return Response.status(Status.NOT_IMPLEMENTED).build();
```
olim7t, line 86: Same here, not really worth a separate method IMHO.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceImpl.java
ivansenic: Yea, but usually you want to avoid the exception throwing for control flows, as it's come with the perf penalty for constructing and throwing the exception (like strack trace gathering).. But to me it seems that `where` should be always set, so we are not on the "hot" path here..

ivansenic (approved):
Very nice, approving with pleasure.. :+1:

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/impl/GrpcExceptionMapper.java
```diff
@@ -0,0 +1,93 @@
...
+      case OUT_OF_RANGE:
+      case INTERNAL:
+      case DATA_LOSS:
+        break;
+    }
+    return grpcResponse(
+        sc, Response.Status.INTERNAL_SERVER_ERROR, "Unhandled gRPC failure", grpcE.getMessage());
+  }
+
+  private static Response grpcResponse(
```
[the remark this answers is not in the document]
tatu-at-datastax: We possibly should log about unrecognized cases, or possibly some subset of known ones. But in case it's invalid request, for example, indicating problem to caller seems best; avoid log spam since the exception is not due to system failing but due to caller providing bad input. I think we should only log when there is likely flaw in our handling; not when we get bad input that is properly handled.

olim7t:
Force-pushed to rebase on latest v2.0.0.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceImpl.java
olim7t: Agreed, I wouldn't do it for regular logic, but for error handling it feels acceptable.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/impl/DefaultExceptionMapper.java
```diff
@@ -0,0 +1,86 @@
...
+    // If we're dealing with a web exception, we can service certain types of request (like
+    // redirection or server errors) better and also propagate properties of the inner response.
+    if (exception instanceof WebApplicationException) {
+      final Response response = ((WebApplicationException) exception).getResponse();
+      Response.Status.Family family = response.getStatusInfo().getFamily();
+      if (family.equals(Response.Status.Family.REDIRECTION)) {
+        return response;
+      }
+      if (family.equals(Response.Status.Family.SERVER_ERROR)) {
+        logException(exception);
```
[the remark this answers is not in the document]
olim7t: Good catch, I'll see how I can refactor the code.

PR #1476, opened 2021-12-10
remove reflection usage from the request hot path in C4

ivansenic:
**What this PR does**:

Decreases the usage of the reflection in the `persistence-cassandra-4.0`. I see no reason the `execute` should be called with the reflection.. In addition, reading tracing info should not be done always..

No isse.

**Checklist**
Should not require any test changes.

### DIFF
--- a/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
+++ b/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
@@ -438,8 +438,18 @@ public class Conversion {
 
   public static Result toResult(
       ResultMessage resultMessage, org.apache.cassandra.transport.ProtocolVersion version) {
-    return toResultInternal(resultMessage, version)
-        .setTracingId(ReflectionUtils.getTracingId(resultMessage));
+    return toResult(resultMessage, version, true);
+  }
+
+  public static Result toResult(
+      ResultMessage resultMessage,
+      org.apache.cassandra.transport.ProtocolVersion version,
+      boolean includeTracingInfo) {
+    Result result = toResultInternal(resultMessage, version);
+    if (includeTracingInfo) {
+      result.setTracingId(ReflectionUtils.getTracingId(resultMessage));
+    }
+    return result;
   }
 
   private static Result toResultInternal(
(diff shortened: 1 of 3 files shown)


persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
```diff
@@ -438,8 +438,18 @@ private static PersistenceException addSuppressed(
 
   public static Result toResult(
       ResultMessage resultMessage, org.apache.cassandra.transport.ProtocolVersion version) {
-    return toResultInternal(resultMessage, version)
-        .setTracingId(ReflectionUtils.getTracingId(resultMessage));
+    return toResult(resultMessage, version, true);
+  }
+
+  public static Result toResult(
```
dougwettlaufer, line 444: Should we create an issue to add this to the other persistence modules?

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/ReflectionUtils.java
```diff
@@ -84,10 +79,6 @@
     }
   }
 
-  static Response execute(Request request, QueryState queryState, long queryStartNanoTime) {
```
dougwettlaufer, line 87: Looks like they did change this to public
https://github.com/apache/cassandra/commit/a7c4ba9eeecb365e7c4753d8eaab747edd9a632a#diff-8c4c40abb1aac546629cbf00371d0f93bac1de96d20f064bec2b31968b2e9a71R219

dougwettlaufer (approved):

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/ReflectionUtils.java
ivansenic: Yep..

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Conversion.java
ivansenic: Not sure if it's needed.. Here it's really more about avoid that reflection call if you know trace information will not be needed..

PR #1482, opened 2021-12-13
improve type conversions in SGv2 REST service

jeffreyscarpenter:


sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
```diff
@@ -252,12 +279,12 @@ public JsonNode jsonNodeFrom(QueryOuterClass.Value value) {
   protected static final class CounterCodec extends FromProtoValueCodec {
     @Override
     public Object fromProtoValue(QueryOuterClass.Value value) {
-      return String.valueOf(value.getInt());
+      return String.valueOf(Values.int_(value));
```
tatu-at-datastax, line 282: I don't think this is right: remove `String.valueOf()`.

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -2883,19 +2900,26 @@ public void testTimestampHandling() throws IOException {
...
     row2.put("est_unit_cost", 38.95);
     row2.put("est_unit_cost_updated", timestamp2);
     row2.put("inspection_notes", "frayed cable");
     row2.put("mean_failure_time_hours", 5917321.12334);
+    row2.put("rating", (float) 92.6);
+    row2.put("stars", (byte) 4);
+    row2.put("likes", (short) 926);
+    row2.put("test_runs", BigInteger.ONE);
+    row2.put("source_ip", "FE80:CD00:0:CDE:1257:0:211E:729C");
+    row2.put("description", (new String("AQID//7gEiMB")).getBytes());
```
tatu-at-datastax, line 2922: No need to use String constructor. Also: should pass encoding (although being ASCII not a big deal), `StandardCharsets.UTF_8` or such.

olim7t:
Looks like the branch contains all the commits from your previous PR. You should be able to rebase to only keep the last commit:
```
git rebase --onto v2.0.0 v2/jeff/sgv2-time^ v2/jeff/sgv2-time
```

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
jeffreyscarpenter: `String.valueOf()` was in there originally though, I didn't add that.

olim7t: Comment above:
```
  // NOTE! Should be able to just use `LongCodec` but SGv1 seems to expect JSON String as
  // return value, and not more logical choice of integral number
```
This is the interoperability concern I mentioned in the previous PR. v1 does that for CQL bigint (Java long) too.

olim7t (approved):
In `BlobCodec`:
```java
      // 13-Dec-2021, jsc: Use base-64 encoded value here
      return ByteBufferUtils.toBase64(Values.bytes(value));
```

I think `return Values.byteBuffer(value)` would work too, Jackson will serialize the `ByteBuffer` to Base64 by default.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
tatu-at-datastax: Oh. My bad -- I somehow didn't see it is for `Counter` and (not sure how) thought it was for numbers.

So never mind. I should have checked the context since I bumped into this myself etc :-D

PR #1485, opened 2021-12-14
rework null handling in FromProtoValueCodecs

jeffreyscarpenter:
improve handling of null (unset) values returned from gRPC API in SGv2 REST API service.

### DIFF
--- a/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
+++ b/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
@@ -3,7 +3,6 @@ package io.stargate.sgv2.restsvc.grpc;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
-import io.stargate.core.util.ByteBufferUtils;
 import io.stargate.grpc.Values;
 import io.stargate.proto.QueryOuterClass;
 import java.time.Instant;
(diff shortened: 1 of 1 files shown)


sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/grpc/FromProtoValueCodecs.java
```diff
@@ -293,86 +326,112 @@ public JsonNode jsonNodeFrom(QueryOuterClass.Value value) {
...
     }
   }
 
   protected static final class TimestampCodec extends FromProtoValueCodec {
     @Override
     public Object fromProtoValue(QueryOuterClass.Value value) {
-      return Instant.ofEpochMilli(value.getInt()).toString();
+      return (value.getInnerCase() == QueryOuterClass.Value.InnerCase.NULL)
+          ? null
+          : Instant.ofEpochMilli(value.getInt()).toString();
```
jeffreyscarpenter, line 363: @mpenick we've been trying to use the operations on `io.stargate.grpc.Values` where possible but there is no `Values.timestamp() operation. Should there be?

mpenick: It's encoded in the `int` field: https://github.com/stargate/stargate/blob/v1.0.45/grpc-proto/proto/query.proto#L93-L94

jeffreyscarpenter: agree, but what I'm saying is there are operations on `Values`corresponding to other types that can be encoded in `int`:

- `Values.int_()`
- `Values.smallint()`
- `Values.tinyint()`

olim7t (approved):

tatu-at-datastax:
Looks good, makes sense.

PR #1488, opened 2021-12-15
write documents with the new shredding and writer

ivansenic:
**ATTENTION: This PR removes support for the `x-www-form-urlencoded` content type when writing docs.**

**What this PR does**:
A refactored write of documents based on top of the new shredding. Beside shredding, we have these improvements:
* query preparing is done in parallel
* write document path does not need a delete query
* cleaner code, better tests (added batch validation in `ValidatingDataStore`) and separation of concerns in components

Note that document patching and batch writes is not part of this PR and will be done after this one is merged. The final goal would be to remove majority of existing code in `DocumentResourceV2`, `DocumentService` and `DocumentDB`.

Internal issue.

**Checklist**
- [x] Changes manually tested
- [x] proper error codes in `ReactiveDocumentService`
- [x] test code for `ReactiveDocumentService` and `DocumentWriteService`

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/write/DocumentWriteService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/write/DocumentWriteService.java
@@ -1,0 +1,224 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.stargate.web.docsapi.service.write;
+
+import io.reactivex.rxjava3.core.Single;
+import io.reactivex.rxjava3.core.SingleSource;
+import io.reactivex.rxjava3.schedulers.Schedulers;
+import io.stargate.db.datastore.DataStore;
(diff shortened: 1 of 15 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/write/DocumentWriteService.java
```diff
@@ -0,0 +1,224 @@
...
+
+  // executes batch
+  private SingleSource<ResultSet> executeBatch(
+      DataStore dataStore, List<BoundQuery> boundQueries, ExecutionContext context) {
+
+    // trace queries in context
+    boundQueries.forEach(context::traceDeferredDml);
+
+    // then execute batch
+    boolean loggedBatch = useLoggedBatches.orElse(dataStore.supportsLoggedBatches());
```
dougwettlaufer, line 213: Why not do this in the constructor so we don't need the `orElse` everywhere?

restapi/src/main/java/io/stargate/web/docsapi/service/write/db/InsertQueryBuilder.java
```diff
@@ -0,0 +1,129 @@
...
+
+public class InsertQueryBuilder {
+
+  private final int maxDepth;
+  private final List<ValueModifier> insertValueModifiers;
+
+  /**
+   * Constructs the query builder for inserting document rows based on the defined max depth.
+   *
+   * @param maxDepth Mac depth of the document storage
```
dougwettlaufer, line 40:
```suggestion
   * @param maxDepth Max depth of the document storage
```

restapi/src/main/java/io/stargate/web/docsapi/service/write/DocumentWriteService.java
ivansenic: Because the data store is supplied in the method call. The writer needs a pointer to `DataStore` the write should be executed on.

ivansenic:
@EricBorczuk I would be up in merging this by the end of the day, do you wanna have a look?

EricBorczuk (approved):

PR #1491, opened 2021-12-15
Docker container for REST API

jeffreyscarpenter:


sgv2-restapi/Dockerfile
```diff
@@ -0,0 +1,8 @@
+# TODO: change to more modern JVM
```
ivansenic, line 1: what stops us to do this now?

sgv2-restapi/pom.xml
```diff
@@ -259,6 +259,20 @@
       <plugin>
         <artifactId>maven-surefire-plugin</artifactId>
       </plugin>
+      <!-- Plugin for building Docker images -->
+      <plugin>
+        <groupId>com.google.cloud.tools</groupId>
+        <artifactId>jib-maven-plugin</artifactId>
```
ivansenic, line 265: So what's the deal with the plugin and the `build.sh`/`Dockerfile` files, which one will be used?

sgv2-restapi/pom.xml
```diff
@@ -259,6 +259,20 @@
...
+        <version>3.1.4</version>
+        <configuration>
+          <from>
+            <image>adoptopenjdk:8</image>
```
ivansenic, line 269: this image is not the same as the one in the `Dockerfile`

sgv2-restapi/pom.xml
jeffreyscarpenter: Note this is a draft PR and not intended for review yet. However, the plugin is the one that will be used, the Dockerfile and associated script will be deleted in the next push.

sgv2-restapi/pom.xml
jeffreyscarpenter: ignore the dockerfile, it will be deleted

sgv2-restapi/Dockerfile
jeffreyscarpenter: nothing, @tatu-at-datastax will be doing this soon

sgv2-restapi/pom.xml
```diff
@@ -262,6 +262,33 @@
...
+          </to>
+          <!-- See: https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#container-object-->
+          <container>
+            <creationTime>USE_CURRENT_TIMESTAMP</creationTime>
+            <filesModificationTime>${maven.build.timestamp}</filesModificationTime>
+            <ports>
+              <port>8082</port>
+            </ports>
+            <jvmFlags>
+              <jvmFlag>-Ddw.stargate.grpc.host=localhost</jvmFlag>
```
ivansenic, line 285: Hmm not sure what the defaults mean here wrt to the image building.. Shouldn't these defaults be in the default configuration and not bounded to the images itself? Furthermore, will environment variables for the same be treated as higher priority and overwrite those?

jeffreyscarpenter: The properties can be overridden using the `JAVA_TOOL_OPTIONS` environment variable as described here https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#how-do-i-set-parameters-for-my-image-at-runtime. 

I agree there is some awkwardness with using Java properties. @tatu-at-datastax we could also discuss changing the REST API service so that it uses environment variables in addition to / instead of Java properties? Whichever approach we choose, I think it's good practice to have the default settings defined in the image. We should also document how to override them in the README. After we decide how the properties will be set I'd be happy to update this PR to build the image and update the documentation accordingly.

ivansenic: Well I guess that Dropwizard can accept properties coming from different sources right?

jeffreyscarpenter: Yes, but I believe the plan is to remove Dropwizard imminently

ivansenic: Yes but other frameworks do the same, env vars always have highest priority..

tatu-at-datastax: If I remember correctly, it is possible to pipe in ENV variables to DropWizard config, but it is not enabled by default (for security reasons; not everyone wants to allow such overrides, I think).
If env variables are the main mechanism (k8s seems to favor them; Java world considered them ancient and obsolete, go figure -- everything old is now new hotness :) ), we could also just consider passing them through system properties (`-Ddw.some.setting=${MY_ENV_VARIABLE)`) for now? And then consider rework when we re-platform.

tatu-at-datastax, line 285: On Grpc host: this would then build an image that expects gRPC to be running on same node, right?
Which works for testing, and some of smaller perf tests, but would not work when splitting. Unless we use something to do routing externally (service mesh / sidecar) and just bind via localhost?

dougwettlaufer: > I think it's good practice to have the default settings defined in the image

Typically defaults are defined in code that way you have the same behavior inside and outside of Docker. I am +1 on environment variables though.

jeffreyscarpenter: pushed a change that uses environment variables to set java properties, as suggested by @tatu-at-datastax. While this does work, it is slightly awkward and does require setting defaults for all environment variables.

jeffreyscarpenter: You'd need to override the value (which you'd now do via environment variable given the latest commit).

sgv2-restapi/pom.xml
```diff
@@ -262,6 +262,46 @@
...
+          </to>
+          <!-- See: https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#container-object-->
+          <container>
+            <creationTime>USE_CURRENT_TIMESTAMP</creationTime>
+            <filesModificationTime>${maven.build.timestamp}</filesModificationTime>
+            <ports>
+              <port>8082</port>
+            </ports>
+            <environment>
+              <STARGATE_GRPC_HOST>localhost</STARGATE_GRPC_HOST>
```
tatu-at-datastax, line 285: I assume these are default values that may be overridden? (which makes sense)

jeffreyscarpenter: Yes exactly. I can put documentation into the README.

PR #1493, opened 2021-12-15
SGv2: exclude old validation api, update to latest DropWizard

tatu-at-datastax:
**What this PR does**:

Fixes a version conflict that occurs if building Docker image of Stargate V2 REST service; conflict between older 1.x validation api jar brought to by swagger (and for some reason, directly?) vs DropWizard.
For some reason does not break things with "fat jar" (maybe order dependant) but does with cleaner Docker build.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,12 +15,12 @@
     <doclint>none</doclint>
     <driver.version>4.9.0</driver.version>
     <xml-format.skip>true</xml-format.skip>
-    <dropwizard.version>2.0.21</dropwizard.version>
+    <dropwizard.version>2.0.26</dropwizard.version>
     <!--
       Note that this is currently aligned with the metrics version that DropWizard transitively depends on.
       It's probably a good idea to keep the two versions in sync.
     -->
-    <dropwizard.metrics.version>4.1.19</dropwizard.metrics.version>
+    <dropwizard.metrics.version>4.1.27</dropwizard.metrics.version>
     <!--
       Note that this is currently aligned with the Jetty version that DropWizard transitively depends on.
       It's probably a good idea to keep the two versions in sync.
--- a/sgv2-restapi/pom.xml
+++ b/sgv2-restapi/pom.xml
@@ -94,14 +94,17 @@
       <artifactId>jetty-servlet</artifactId>
     </dependency>
     <dependency>
-      <groupId>javax.validation</groupId>
-      <artifactId>validation-api</artifactId>
-      <version>1.1.0.Final</version>
-    </dependency>
-    <dependency>
       <groupId>io.swagger</groupId>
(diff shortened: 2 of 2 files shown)


tatu-at-datastax:
Will merge to unblock work on Docker image creation.

pom.xml
```diff
@@ -15,12 +15,12 @@
     <doclint>none</doclint>
     <driver.version>4.9.0</driver.version>
     <xml-format.skip>true</xml-format.skip>
-    <dropwizard.version>2.0.21</dropwizard.version>
+    <dropwizard.version>2.0.26</dropwizard.version>
```
ivansenic, line 18: I think I remember when we were doing this before, other dependencies should also be checked and I think updated.. Have the feeling that this is not enough..

ivansenic: Furthermore I see you are merging V2 PR without approvals, what's the reason for it? 

I do understand you want to be faster, but that's eventually gonna be merged to master and I don't think we will do a big review then, but rather count that all in V2 branch was reviewed.. Or what's the idea?

tatu-at-datastax: The idea is probably not to merge to `master`/`mainline` any more (although that was considered at some point), but that it will replace master/mainline as for v2.0.0.

I try to wait for approval in general, but in this case I wanted to unblock Jeff -- although in hindsight I could just have asked him to approve it.

tatu-at-datastax: On DropWizard version: I updated dw-metrics as suggested. It is possible that jetty version could use upgrade; I'll check that now. If you remember others that likely need upgrade, please let me know.

dougwettlaufer: > The idea is probably not to merge to master/mainline any more

If the `v2.0.0` branch isn't going to be merged then how will all of this be reviewed?

tatu-at-datastax: Right, there needs to be a review process closer to when we consider work not to be prototype any more.
Regardless of whether there is actual merge (which may or may not be feasible) or simply switching mainline to be another branch.

The reason I think merge/non-merge is not a big issue is because so far very little code would change in merge-style way: almost everything is technically new code (under new modules) and there are few changes to SGv1 codebase itself. And of SGv1 changes made some have been backported.

Perhaps this isolation will change after REST module (except for Documents API that likely will be extracted to another module); if so we will need to decide if we do actually want a merge. My assumption was simply that merging has limited value & not merging leaves as fully functional V1 for however long anyone wants it (granted this can be handled by branching separate `v1` from `master` before merge so maybe that's a moot point).

ivansenic: * +1 for merging and creating a `v1` prior to that
* +1 for also having protection rules on the `v2` branch asap

PR #1496, opened 2021-12-17
[WIP] Grpc close fix using a netty listener

tomekl007:
Created for demonstration purposes, DO NOT REVIEW.


mpenick:
This was another approach to closing gRPC connections when a tenant is removed from a coordinator. The grpc java library doesn't provide a good way to do this. I think we need to go back to the drawing board.

PR #1498, opened 2021-12-17
(SGv2) Add GET method implementations (all, by-id) to UDT resource

### LINKED ISSUE #1479: SGv2/REST: implement support for "Stringified" versions of structured types (List, Map, Tuple)
Stargate V1 REST endpoints accept both JSON-typed values (numbers, Arrays, Objects) and so-called "Stringified" alternatives: latter are needed for Path parameters (primary keys) but are also accepted for regular column values.
Format for these values is not very well defined but decoder exist in SGv1/restapi codebase (class `Converters`, I think).
There are also couple of basic Integration tests covering handling, although mostly as side effects (tests that verify other endpoints but use List values and pass Stringified argument).

tatu-at-datastax:
**What this PR does**:

Completes Stargate V2 / REST UDT endpoint implementation by adding "get-all UDTs" and "get specific UDT".

Includes some fixes (like support for decoding `List` and `Set` gRPC/proto types) needed by functionality.

Fixes #1498

**Checklist**
- [ ] Changes manually tested
- [x] Automated Tests added/updated -- passes 3 more formerly failing tests

### DIFF
--- a/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/ResourceBase.java
+++ b/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/ResourceBase.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.databind.json.JsonMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.BytesValue;
 import com.google.protobuf.Int32Value;
@@ -154,6 +155,17 @@ public abstract class ResourceBase {
     List<QueryOuterClass.Row> rows = rs.getRowsList();
     for (QueryOuterClass.Row row : rows) {
       resultRows.add(converter.mapFromProtoValues(row.getValuesList()));
+    }
+    return resultRows;
+  }
+
+  protected static ArrayNode convertRowsToArrayNode(QueryOuterClass.ResultSet rs) {
+    FromProtoConverter converter =
+        BridgeProtoValueConverters.instance().fromProtoConverter(rs.getColumnsList());
+    ArrayNode resultRows = JSON_MAPPER.createArrayNode();
+    List<QueryOuterClass.Row> rows = rs.getRowsList();
+    for (QueryOuterClass.Row row : rows) {
+      resultRows.add(converter.objectNodeFromProtoValues(row.getValuesList()));
     }
     return resultRows;
   }
--- a/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResourceImpl.java
+++ b/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResourceImpl.java
@@ -24,8 +24,6 @@ import io.stargate.sgv2.common.cql.builder.BuiltCondition;
 import io.stargate.sgv2.common.cql.builder.Predicate;
 import io.stargate.sgv2.common.cql.builder.QueryBuilder;
 import io.stargate.sgv2.common.cql.builder.Replication;
-import io.stargate.sgv2.restsvc.grpc.BridgeProtoValueConverters;
-import io.stargate.sgv2.restsvc.grpc.FromProtoConverter;
 import io.stargate.sgv2.restsvc.models.Sgv2Keyspace;
 import io.stargate.sgv2.restsvc.models.Sgv2RESTResponse;
 import io.stargate.sgv2.restsvc.resources.CreateGrpcStub;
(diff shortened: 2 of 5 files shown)


sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/ResourceBase.java
```diff
@@ -158,6 +159,17 @@ protected Response fetchRows(
     return resultRows;
   }
 
+  protected static ArrayNode convertRowsToArrayNode(QueryOuterClass.ResultSet rs) {
```
tatu-at-datastax, line 162: Demoted from "Keyspaces" endpoint as it is now used from multiple places. Could conceivably move to shared codec class, but since our future wrt gRPC/Bridge is uncertain did not refactor further.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResourceImpl.java
```diff
@@ -73,15 +70,14 @@ public Response getAllKeyspaces(
             .from("system_schema", "keyspaces")
             .build();
 
-    QueryOuterClass.Query query =
-        QueryOuterClass.Query.newBuilder().setParameters(paramsB.build()).setCql(cql).build();
+    QueryOuterClass.Query query = QueryOuterClass.Query.newBuilder().setCql(cql).build();
```
tatu-at-datastax, line 73: Did not have any bound parameters (probably copy-pasted from somewhere else).

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -2755,7 +2755,7 @@ public void listAllTypes() throws IOException {
     List<Map<String, String>> fields = (List<Map<String, String>>) response.get(0).get("fields");
     assertThat(fields.size()).isEqualTo(1);
     assertThat(fields.get(0).get("name")).isEqualTo("firstname");
-    assertThat(fields.get(0).get("typeDefinition")).isEqualTo("varchar");
+    assertThat(fields.get(0).get("typeDefinition")).isEqualTo("text");
```
tatu-at-datastax, line 2758: This is a weird one: I don't like "fixing" tests like this but as far as I can see, field was created as "text", and CQLsh/console returns "text" in such cases -- so I don't know how or why SGv1 would translate this into "varchar" somehow. I also don't know how I would reliably translate it either, given that it seems there is no 1-to-1 mapping between couple of textual types that all become "text".

So, open to suggestions here. I also note that our test coverage for UDT field/column types is rather... limited.

ivansenic: Interesting. I can not help here, but I do notice that if this is wrong in the V1 we need a fix there asap as well?

tatu-at-datastax: This: https://github.com/stargate/stargate/issues/1499

tatu-at-datastax: That one is tricky so I will revert test change here to get this PR in (with 1 remaining "extra" fail, which was already failing for v2.0.0) and then tackle this test once we can resolve #1499.

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -2755,6 +2755,9 @@ public void listAllTypes() throws IOException {
     List<Map<String, String>> fields = (List<Map<String, String>>) response.get(0).get("fields");
     assertThat(fields.size()).isEqualTo(1);
     assertThat(fields.get(0).get("name")).isEqualTo("firstname");
+
+    // 20-Dec-2021, tatu: This is due to [stargate#1499], so fails with v2.0.0
```
tatu-at-datastax, line 2759: @ivansenic As per my comment, leaving test as-is, until SGv1 part can be addressed -- or we change behavior within v2.0.0.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResourceImpl.java
```diff
@@ -73,15 +70,14 @@ public Response getAllKeyspaces(
...
     QueryOuterClass.Response grpcResponse = blockingStub.executeQuery(query);
 
     final QueryOuterClass.ResultSet rs = grpcResponse.getResultSet();
 
     // two-part conversion: first from proto to JsonNode for easier traversability,
```
jeffreyscarpenter, line 78: does this comment still apply?

jeffreyscarpenter (approved):

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2KeyspacesResourceImpl.java
tatu-at-datastax: Technically yes, but I Could probably change it or just remove.

PR #1505, opened 2021-12-21
update and unify dependencies

ivansenic:
This PR does a maintenance on the dependencies. Here is the full list of the things done:

* centralize the `immutables` version and updates to latest
* centralize the `swagger-jersey2-jaxrs` and update to latest
* update all test frameworks to the latest (except `mockito` which is updated to the latest 3.x line)
* organization of the dependencies in all projects in groups - stargate, C, 3rd party, test
* removing the `javax.validation-api` and `javax.annotation-api` from the `persistence-api` as this makes no sense at all
* `javatuples` version centralized, and defined correct scope for the projects that depend on it
* setting the explicit maven compiler java version to `1.8`
* remove all `<scope>compile</scope>` from the poms, as this is default
* updates `dropwizard` to latest version `2.0.27`

Note that of course there is bunch of other dependencies to be updated. Here is my overview on the most important ones:
* `cassandra-all` from `3.11.9` to `3.11.11`
* DS OSS `java-driver-core` from `4.9.0` to `4.13.0`

Tasks:
* [ ] Rebase before merge

### DIFF
--- a/auth-api/pom.xml
+++ b/auth-api/pom.xml
@@ -36,12 +36,6 @@
       <artifactId>persistence-api</artifactId>
       <version>${project.version}</version>
       <scope>provided</scope>
-      <exclusions>
-        <exclusion>
-          <groupId>org.hibernate.validator</groupId>
-          <artifactId>hibernate-validator</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
 
     <!-- 3rd party dependencies -->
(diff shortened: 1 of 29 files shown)


dougwettlaufer (approved):

ivansenic:
@tatu-at-datastax @jeffreyscarpenter FYI: I am about to merge this and this could break your `v2` branch.. Not sure how much you depend on the main pom..

tatu-at-datastax:
@ivansenic thank you fro heads up. I hope it merges fine; our dependencies should use parent pom but there's just one new module at this point.

PR #1506, opened 2021-12-22
Rename Column.Type.Varchar to Text, remove protocol code 10 (fixes #1499)

### LINKED ISSUE #1499: Persistence API exposes Utf8Type as `varchar` instead of `text`
Stargate V1 REST API's "list all User Defined Types" seems to apply a translation to fields of type `text` and return them as `varchar`. Although these 2 types are generally interchangeable (see https://docs.datastax.com/en/cql-oss/3.x/cql/cql_reference/cql_data_types_c.html or https://stackoverflow.com/questions/17530230/cassandra-text-vs-varchar) it seems better to return `text` if type is declared as `text`: especially since "cqlsh" (and Astra CQL console) both report these fields as "text" for both `text` and `varchar` declarations.

olim7t:
**What this PR does**:
CQL types `varchar` and `text` are synonyms, they can be used interchangeably. User-facing Cassandra features (DESCRIBE, name of the type in system tables...) always use `text`, even if the user specified `varchar` at creation time.

Stargate does the opposite: text types are always returned as `varchar`. In addition, `Column.Type` has an extra `Text` constant that is never used (it corresponds to native protocol code 10, which was removed in v3, the oldest version that Stargate supports).

This PR does two things:
* remove the obsolete `Column.Type` constant for code 10 (that used to called `Text`).
* in addition, also rename the `Varchar` constant to `Text`. This is a bit more controversial, but I think it's nice for the Persistence API to be consistent with Cassandra. (alternatively, we could keep the name `Varchar` internally, and only format it as `text` when it gets returned to clients).

Fixes #1499

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/grpc/src/test/java/io/stargate/grpc/service/ExecuteQueryTest.java
+++ b/grpc/src/test/java/io/stargate/grpc/service/ExecuteQueryTest.java
@@ -254,7 +254,7 @@ public class ExecuteQueryTest extends BaseGrpcServiceTest {
             .addExpected(
                 ColumnSpec.newBuilder()
                     .setName("c2")
-                    .setType(TypeSpec.newBuilder().setBasic(TypeSpec.Basic.TEXT)))
+                    .setType(TypeSpec.newBuilder().setBasic(TypeSpec.Basic.VARCHAR)))
             .addExpected(
                 ColumnSpec.newBuilder()
                     .setName("c3")
@@ -269,7 +269,7 @@ public class ExecuteQueryTest extends BaseGrpcServiceTest {
                         TypeSpec.newBuilder()
                             .setList(
                                 TypeSpec.List.newBuilder()
-                                    .setElement(TypeSpec.newBuilder().setBasic(Basic.TEXT)))))
+                                    .setElement(TypeSpec.newBuilder().setBasic(Basic.VARCHAR)))))
             .build(false),
         ColumnMetadataBuilder.builder()
             .addActual(Column.create("l", Type.List.of(Type.Int)))
@@ -304,7 +304,7 @@ public class ExecuteQueryTest extends BaseGrpcServiceTest {
                         TypeSpec.newBuilder()
                             .setMap(
                                 TypeSpec.Map.newBuilder()
-                                    .setKey(TypeSpec.newBuilder().setBasic(TypeSpec.Basic.TEXT))
+                                    .setKey(TypeSpec.newBuilder().setBasic(TypeSpec.Basic.VARCHAR))
                                     .setValue(
                                         TypeSpec.newBuilder().setBasic(TypeSpec.Basic.BIGINT)))))
             .build(false),
--- a/grpc/src/test/java/io/stargate/grpc/service/ExecuteQueryTest.java
+++ b/grpc/src/test/java/io/stargate/grpc/service/ExecuteQueryTest.java
@@ -68,13 +68,13 @@ public class ExecuteQueryTest extends BaseGrpcServiceTest {
     final String releaseVersion = "4.0.0";
 
     ResultMetadata resultMetadata =
-        Utils.makeResultMetadata(Column.create("release_version", Type.Varchar));
+        Utils.makeResultMetadata(Column.create("release_version", Type.Text));
     Prepared prepared =
         new Prepared(
             Utils.STATEMENT_ID,
             Utils.RESULT_METADATA_ID,
             resultMetadata,
-            Utils.makePreparedMetadata(Column.create("key", Type.Varchar)),
+            Utils.makePreparedMetadata(Column.create("key", Type.Text)),
             false,
             false);
     when(connection.prepare(eq(query), any(Parameters.class)))
@@ -167,12 +167,12 @@ public class ExecuteQueryTest extends BaseGrpcServiceTest {
         // Invalid arity
         arguments(
             org.assertj.core.util.Arrays.array(
-                Column.create("k", Type.Varchar), Column.create("v", Type.Int)),
+                Column.create("k", Type.Text), Column.create("v", Type.Int)),
             org.assertj.core.util.Arrays.array(Values.of("a")),
             "Invalid number of bind values. Expected 2, but received 1"),
         // Invalid type
         arguments(
-            org.assertj.core.util.Arrays.array(Column.create("k", Type.Varchar)),
+            org.assertj.core.util.Arrays.array(Column.create("k", Type.Text)),
             org.assertj.core.util.Arrays.array(Values.of(1)),
             "Invalid argument at position 1"));
   }
@@ -245,7 +245,7 @@ public class ExecuteQueryTest extends BaseGrpcServiceTest {
     return Stream.of(
         ColumnMetadataBuilder.builder()
             .addActual(Column.create("c1", Column.Type.Int))
-            .addActual(Column.create("c2", Column.Type.Varchar))
+            .addActual(Column.create("c2", Column.Type.Text))
             .addActual(Column.create("c3", Column.Type.Uuid))
             .addExpected(
                 ColumnSpec.newBuilder()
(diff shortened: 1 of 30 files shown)


olim7t:
Checking that REST still accepts `varchar` from the client:
```
curl --request POST 127.0.0.2:8082/v2/schemas/keyspaces/test/tables --header "X-Cassandra-Token: $AUTH_TOKEN" --header "Content-Type: application/json" --header "Accept: application/json" --data '{"name":"foo", "columnDefinitions": [{"name":"k","typeDefinition":"varchar"}], "primaryKey":{"partitionKey":["k"]}}'
```

olim7t:
Same for GraphQL:
```
mutation {
  createTable(
    keyspaceName: "test"
    tableName: "foo"
    partitionKeys: [{ name: "k", type: { basic: VARCHAR } }]
  )
}
```

tatu-at-datastax:
Makes sense to me, but I think other team members probably should review this too before merging.

persistence-api/src/main/java/io/stargate/db/schema/Column.java
```diff
@@ -359,7 +358,7 @@ public boolean isSet() {
 
     Smallint(19, Short.class, false, "16-bit signed integer"),
 
-    Text(10, String.class, true, "UTF-8 encoded string"),
+    Text(13, String.class, true, "UTF-8 encoded string"),
```
ivansenic, line 361: Can this have some kind of backward compatibility issue? I guess not, but just to make sure?

persistence-api/src/main/java/io/stargate/db/schema/Column.java
```diff
@@ -712,6 +709,9 @@ private static UserDefinedType findUDTType(Keyspace keyspace, String udtName, bo
     }
 
     private static Type parseBaseType(String str) {
+      if (str.equalsIgnoreCase("varchar")) {
```
ivansenic, line 712: revert to avoid null issues `"varchar".equalsIgnoreCase(str)`

ivansenic (commented):
I guess this looks good, not sure why tests are not green.. Seems also that we internally never used `varchar` (like in docs api, internal tables, etc), or?

persistence-api/src/main/java/io/stargate/db/schema/Column.java
olim7t: I think it's fairly safe given that code 10 doesn't exist in the protocol. It's not possible for a client to send it -- nor accept it -- unless it was buggy.

grpc/src/test/java/io/stargate/grpc/service/ExecuteQueryTest.java
```diff
@@ -254,7 +254,7 @@ public void columnMetadata(
             .addExpected(
                 ColumnSpec.newBuilder()
                     .setName("c2")
-                    .setType(TypeSpec.newBuilder().setBasic(TypeSpec.Basic.TEXT)))
+                    .setType(TypeSpec.newBuilder().setBasic(TypeSpec.Basic.VARCHAR)))
```
olim7t, line 257: gRPC is a bit tricky because `query.proto` also defines both constants `TEXT = 0xA; VARCHAR = 0x0D;`. I don't feel comfortable removing or renaming here, since that would be a breaking change of the protocol.

Maybe we can deprecate `TEXT`, and justify keeping `VARCHAR` for 0x0D because that's the "internal" name used by the native protocol, and gRPC lives at that same low level.

olim7t:
Sorry about the repeated pushes, I'm having trouble running integration tests locally.

grpc/src/test/java/io/stargate/grpc/service/ExecuteQueryTest.java
ivansenic: This starts to be quite confusing.. So other APIs would work with `text` and gRPC would know only `varchar`?

olim7t: Yeah ideally I would have changed it too, but we have to consider backward compatibility...

I'm starting to wonder if this issue is worth "fixing" at all... If both names are synonymous, then returning `varchar` by default is not a bug, just a different behavior from what Cassandra usually does.

tatu-at-datastax: I guess if it is documented that is an option, but requires all new services to translate to `varchar`, and maybe confuse developers when CQL shell and other direct mechanisms expose `text`.
And to me it still seems like a bug to fix as gRPC was added relatively recently, although issue itself came with Stargate persistence. Too bad it was not caught with tests earlier, but it seems there has been some confusion between 2 options for longer time.

ivansenic: So what's the deal now with grpc, are we keeping it in the proto, but removing support overall? I am fine in having it in a proto as deprecated as long as we will not end up in error once used.. Maybe additional INT test here would be nice?

ivansenic (approved):
I think this is in a good state now and have a strong feeling we are not braking anything and yet improving..

ivansenic:
Blocks #1510 if merged first.

health-checker/src/main/java/io/stargate/health/StorageHealthChecker.java
```diff
@@ -60,7 +60,7 @@
           .name(TABLE_NAME)
           .addColumns(
               ImmutableColumn.create(PK_COLUMN_NAME, Column.Kind.PartitionKey, Column.Type.Uuid),
-              ImmutableColumn.create(VALUE_COLUMN_NAME, Column.Kind.Regular, Column.Type.Varchar))
+              ImmutableColumn.create(VALUE_COLUMN_NAME, Column.Kind.Regular, Column.Type.Text))
```
dougwettlaufer, line 63: What happens when SG is updated to this version in an already running system?

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -2217,7 +2217,7 @@ public void addColumn() throws IOException {
     createKeyspace(keyspaceName);
     createTable(keyspaceName, tableName);
 
-    ColumnDefinition columnDefinition = new ColumnDefinition("name", "varchar");
+    ColumnDefinition columnDefinition = new ColumnDefinition("name", "text");
```
dougwettlaufer, line 2220: Should we keep this or duplicate the test to verify backwards compatibility?

dougwettlaufer (approved):
Looks good but just to confirm, if something is already stored as `varchar` and then this version is rolled out are we still able to retrieve the row?

olim7t:
> Looks good but just to confirm, if something is already stored as `varchar` and then this version is rolled out are we still able to retrieve the row?

Yes.
It's stored as `UTF8Type` and transferred as code 10 over the wire, we're not changing any of that. Just the way it's surfaced through the persistence API.

tatu-at-datastax:
@olim7t This might be ready to merge? It would, I think, resolve the last failing `RestApiV2Test` for SGv2.

ivansenic:
@olim7t I merged #1510, which means you will need to rebase and:

* update the type conversion in all `Converersion` classes
* update all `ConverersionTest` classes

Hopefully the compile itself shows all the places explicitly.. Pls remove `TODO` from `Converersion` classes when done..

olim7t:
I ran more tests with gRPC: the only place where `TypeSpec.Basic` is exposed to the client is in response metadata, where we already defaulted to `VARCHAR`. So there is very little chance that `TEXT` was referenced anywhere.

On the v2.0.0 there is a `CreateTable` operation that takes `TypeSpec.Basic` as input, but it already converts both gRPC types to `Text` so it will rebase nicely (`DdlMessageHandler.convertType`):
```
      case TEXT:
      case VARCHAR:
        return Column.Type.Text;
```

olim7t:
Force-pushed to rebase on latest changes.

PR #1507, opened 2021-12-22
updated persistence-cassandra-3.11 to 3.11.11

ivansenic:
**What this PR does**:
Updates the `persistence-cassandra-3.11` to use the latest version `3.11.11`. The `cassandra-driver-core` stayed on `3.0.1` in this version, so no update there needed.

**Tasks:**
* [x] extend `README.md` with the C update instructions

### DIFF
--- a/persistence-cassandra-3.11/README.md
+++ b/persistence-cassandra-3.11/README.md
@@ -1,0 +1,18 @@
+# Persistence Cassandra 3.11
+
+This module represents the implementation of the [persistence-api](../persistence-api) for the Cassandra `3.11.x` version.
+
+## Cassandra version update
+
+The current Cassandra version this module depends on is `3.11.11`.
+In order to update to a newer patch version, please follow the guidelines below:
+
+* Update the `cassandra.version` property in the [pom.xml](pom.xml).
+* Check what is the version of the `com.datastax.cassandra:cassandra-driver-core` in the `org.apache.cassandra:cassandra-all` for the updated version. 
+This dependency is set as optional in the `cassandra-all`, but we need it to correctly handle UDFs.
+Set the version of the driver in the `cassandra.bundled-driver.version` property in the [pom.xml](pom.xml).
+* Change the version in the [Cassandra311MetricsRegistryTest.java](src/test/java/org/apache/cassandra/metrics/Cassandra311MetricsRegistryTest.java) to the new one.
+* Check if the new version has a transitive dependency to `org.apache.cassandra:cassandra-thrift`, and if it does remove that dependency from our [pom.xml](pom.xml).
+The `cassandra-thrift` was explicitly added when updating to `3.11.11` as it was not anymore in the `cassandra-all`.
+* Make sure everything compiles and tests are green.
+* Update this `README.md` file with the new or updated instructions.
--- a/persistence-cassandra-3.11/README.md
+++ b/persistence-cassandra-3.11/README.md
@@ -15,6 +15,7 @@ Set the version of the driver in the `cassandra.bundled-driver.version` property
 * Check if the new version has a transitive dependency to `org.apache.cassandra:cassandra-thrift`, and if it does remove that dependency from our [pom.xml](pom.xml).
 The `cassandra-thrift` was explicitly added when updating to `3.11.11` as it was not anymore in the `cassandra-all`.
 * Update the [CI Dockerfile](../ci/Dockerfile) and set the new version in the `ccm create` command related to 3.11.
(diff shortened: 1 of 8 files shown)


ivansenic:
So we depend on the class `org.apache.cassandra.thrift.cassandraConstants` available in the `org.apache.cassandra:cassandra-thrift` in order to read the current thrift version. However, starting from version `3.11.11` `cassandra-all` is not including the `cassandra-thrift` dependency anymore. 

Now the solution could be:
* Read thrift version from somewhere else, I have no idea where this could be..
* Depend ourself on the `org.apache.cassandra:cassandra-thrift:3.11.11` merely for having this class
* Hard-code the version ourselves and check on every update

@stargate/core-contributors Advice? 

I am doing a dependency solution in order to run CI until we decide what we want.

persistence-cassandra-3.11/pom.xml
```diff
@@ -52,6 +52,11 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
```
tatu-at-datastax, line 55: If this is temporary, no problem, but if it was to stay could definitely use a comment explaining the need for inclusion.

mpenick:
@ivansenic Maybe others disagree, but Thrift version doesn't matter for Stargate (at least right now). I'm of the opinion that we hard code the version in a constant w/ some explanation.

Yeah, `MigrationManager.MIGRATION_DELAY_IN_MS` been removed:
https://github.com/apache/cassandra/commit/a960d87e05f01000a758d8e7e5a58be57d13eb33#diff-8191bf68e083d923ddd54125087c550e0fa3136ea0c62d722e79a5d1a1fbe030

It was moved to `MigrationCoordinator` as private variable:
https://github.com/apache/cassandra/commit/a960d87e05f01000a758d8e7e5a58be57d13eb33#diff-ec0b9395200cd1a3aa3a8523a8bbd77850930b483decb3fa68e9508d1bcdb625R69

ivansenic:
> MigrationManager.MIGRATION_DELAY_IN_MS

* +1 for hard-coding the thrift version with an explanation in comments
* for the `MIGRATION_DELAY_IN_MS` I see no other way as to hard-code as well

In any case I would be adding an update C version section to the read me with all steps to pay attention to.

persistence-cassandra-3.11/pom.xml
ivansenic: Removed and hard-coded the version.

ivansenic:
Now it's getting interesting, the `org.apache.cassandra.service.StorageService`  in the 3.11.11 also depends on the `cassandra-thrift` , same class as we do.. The Stargate wont' start due to the `Caused by: java.lang.ClassNotFoundException: org.apache.cassandra.thrift.Cassandra$Iface`. So either something was heavily wrong in the pom of the 3.11.11 or I don't know what was the intended usage here.

Anyway two options from my view:
* Add ourselves the dep to the `cassandra-thrift` 
* Update to `3.11.10` until this is resolved

@mpenick @tatu-at-datastax wdyt?

persistence-cassandra-3.11/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+This module represents the implementation of the [persistence-api](../persistence-api) for the Cassandra `3.11.x` version.
+
+## Cassandra version update
+
+The current Cassandra version this module depends on is `3.11.11`.
+In order to update to a newer patch version, please follow the guidelines below:
+
+* Update the `cassandra.version` property in the [pom.xml](pom.xml).
+* Check what is the version of the `com.datastax.cassandra:cassandra-driver-core` in the `org.apache.cassandra:cassandra-all` for the updated version.
```
jeffreyscarpenter, line 11: I don't understand what this means. Is it saying to look at a setting in a Maven repo or the Apache Cassandra codebase?

jeffreyscarpenter (approved):

dougwettlaufer (approved):

persistence-cassandra-3.11/README.md
ivansenic: Transitive dependencies can be seen in the pom.xml, on the web links like https://mvnrepository.com/artifact/org.apache.cassandra/cassandra-all/3.11.11, or using `./mvnw dependency:tree -pl persistence-cassandra-3.11`.. Should I define the explicit way to do it? What would you like the best? Any proposal for rephrasing it?

ivansenic: Added more detailed explanation. Considering done.

PR #1510, opened 2021-12-23
closes #781: remove cassandra-all dependency from persistence-commons

### LINKED ISSUE #781: Remove cassandra-3.11 dependency from persistence-common
Breaking up https://github.com/stargate/stargate/issues/768 one of the tasks remove cassandra-3.11 dependency from persistence-common module.

So far we've been lucky that there haven't been breaking interface changes between C* 3.11 and DSE/C* 4.0 since we pull dependencies from cassandra-3.11 into all of our persistence modules. We need to clean this up by removing the dependency from persistence-common and if necessary splitting some of the classes into a new, shared module.

ivansenic:
**What this PR does**:
The beauty for the end of the year.. This PR removes the `cassandra-all:3.11.x` dependency from the `persistence-commons`. As `persistence-commons` is used by all persistence implementations, we were actually having a compile time against `3.11.x` for all C/DSE versions. This was wrong, and hopefully now things are more clean.

Every problem I encountered is carefully solved in each commit of this PR, so commit by commit review might be more meaningful:

* Not using anymore the C `TimeSource`, but our own. We had on in the `restapi` so I just moved it to ~`persistence-api`~ `core` and extended the interface.
* Excluded index options is added as abstract method to the `AbstractCassandraSchemaConverter`, each implementation can now implement on it's own
* existing `SeedProvider` is moved to abstract one, and each persistence implementation can extend now and implement correct interface
* similar stuff with the `StargateConfigSnitch`, where we already had a distinction between 3.11 & 4.0, so changes came out nicely here
* and for the end I removed `ColumnUtils` and added direct type to column mappings in the `Conversion` classes, as apparently `ColumnUtils` are not used anywhere else.

I did not add any tests for new classes, I might do this.

*Note that no behavior is broken here, it's 1-1 as it was. However, the modules depending on the `persistence-commons` need to be updated. Thus, we could think about releasing this under `1.1.0`.*

Fixes #781

**Checklist**
- [x] Check progress on #1506 
- [x] Changes manually tested


ivansenic:
**Warn:** The commit https://github.com/stargate/stargate/pull/1510/commits/51229fcbfbd60840fceda768c7f222b4553d099f adds the override for the `getLocalDatacenter` and `getLocalRack` for the `StargateConfigSnitch` in the DSE. 

DSE was using until now the version form the commons, however, there the `GossipingPropertyFileSnitch` in 3.11 had no local datacenter and rack methods (the snitch interface was not defining it). I am not sure how was this working until now and was there a fallback to the default method in the updated snitch interface. However, in C4 we already implemented these new methods and simply return for the local values what ever the Stargate snitch config has. I used the same approach here. Can anybody verify that this is OK?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateConfigSnitch.java
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * <p><i>Note that similar copy of this class exists in all persistence implementations, thus please
+ * apply changes to all of them if needed.</i>
+ */
+public class StargateConfigSnitch extends GossipingPropertyFileSnitch {
+
+  private final StargateSnitchProperties snitchProperties;
+
+  public StargateConfigSnitch() throws ConfigurationException {
+    this(new StargateSnitchProperties("DEFAULT_DC", "DEFAULT_RACK"));
```
dougwettlaufer, line 37: Should we keep the existing defaults of `dc-1` and `rack1`?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateConfigSnitch.java
```diff
@@ -0,0 +1,82 @@
...
+  public String getLocalRack() {
+    return snitchProperties.getRack();
+  }
+
+  @Override
+  public String toString() {
+    return "StargateConfigSnitch{myDC='"
+        + snitchProperties.getDc()
+        + '\''
+        + ", myRack='"
```
dougwettlaufer, line 78:
```suggestion
    return "StargateConfigSnitch{DC='"
        + snitchProperties.getDc()
        + '\''
        + ", rack='"
```

persistence-api/src/main/java/io/stargate/db/util/TimeSource.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.stargate.db.util;
```
jeffreyscarpenter, line 18: For SGv2, we're trying to eliminate dependence of the API Services on the `persistence-api`. Since the document API implementation uses this class, and you're now proposing to use it in the persistence implementations, it might be better to move this interface to the`core` module.

dougwettlaufer (approved):

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateConfigSnitch.java
ivansenic: These above are existing defaults, have a look at [StargateConfigSnitch.java](https://github.com/stargate/stargate/blob/master/persistence-common/src/main/java/io/stargate/db/datastore/common/StargateConfigSnitch.java) being used at the moment. 

I am fine with changing the defaults to anything better if you @dougwettlaufer have a proposal?

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateConfigSnitch.java
ivansenic: This was a stupid copy paste from the existing class, will update in all the new classes now..

persistence-api/src/main/java/io/stargate/db/util/TimeSource.java
ivansenic: Hmm, there was a reason it's not in the `core`.. I will double check, on the first look seems possible yea..

ivansenic: Moved to `core`, let's see if tests are green..

jeffreyscarpenter (approved):

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/StargateConfigSnitch.java
dougwettlaufer: Looks like they're set to the C* defaults [here](https://github.com/stargate/stargate/blob/master/stargate-starter/src/main/java/io/stargate/starter/Starter.java#L353-L354) so by the time we get to `StargateConfigSnitch` they should be set anyway.

ivansenic: OK, I am merging, but we can follow-up on this if needed..

PR #1518, opened 2022-01-04
Add a test for REST API, for existing "where ... $exists" behavior.

### LINKED ISSUE #1517: Missing integration tests for REST API operators $EXISTS, $CONTAINS(|KEY|ENTRY)
It looks like `RestApiV2Test` only covers use of "simple" WHERE operators ($EQ and such) and $IN, but not 4 other set operators.
It would make sense to add tests, especially so they can help with Stargate V2 refactoring.

tatu-at-datastax:
**What this PR does**:

Add one more IT for REST API (and extends another)

Fixes #1517

**Checklist**
- [ ] Changes manually tested


testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -75,6 +76,7 @@
   private String tableName;
   private static String authToken;
   private String host;
+  private String restUrlBase;
```
tatu-at-datastax, line 79: This is added to help with merging to Stargate V2 where we need(ed) to override things; may over time change usage here.

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -107,6 +109,7 @@ public MapGetResponseWrapper() {
   @BeforeEach
   public void setup(TestInfo testInfo, StargateConnectionInfo cluster) throws IOException {
     host = "http://" + cluster.seedAddress();
+    restUrlBase = host + ":8082";
```
tatu-at-datastax, line 112: Was not 100% sure if "cql port" of cluster would work here or not? 8082 appears to be hardcoded in code all over tho.

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -953,15 +956,62 @@ public void getRowsWithInQuery() throws IOException {
...
+
+    // First let's try to find something that does not exist
+    String whereClause = "{\"id\":{\"$eq\":\"1\"},\"extra\":{\"$exists\":true}}";
+    String body =
+        RestUtils.get(
+            authToken,
+            String.format(
+                "%s/v2/keyspaces/%s/%s?where=%s&raw=true",
+                restUrlBase, keyspaceName, tableName, whereClause),
+            HttpStatus.SC_OK);
```
tatu-at-datastax, line 1012: Here we get odd failure:

```
{"description":"Bad request: Invalid value provided for 'extra': Java value true of type 'java.lang.Boolean' is not a valid value for CQL type varchar","code":400}
```

which looks a lot like a bug. But I want to sanity check it first in case I just wrote bad query :)

tatu-at-datastax:
Changed things a bit so that 2 tests (one unit test, one IT) actually verify the way Stargate v1 REST API (v2) actually works wrt "$EXISTS". Bigger contextual discussion on right/wrong behavior is at #1519 -- I suspect deprecation makes sense, but we still need to verify that existing behavior remains as-is.

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
tatu-at-datastax: Apparently this just is the way things work for now; see #1519 for follow-up discussion.

restapi/src/test/java/io/stargate/web/resources/WhereParserTest.java
```diff
@@ -119,6 +119,26 @@ public void testParseValueEmpty() {
             "Value entry for field name, operation $gt was expecting a value, but found null.");
   }
 
+  // 04-Jan-2022, tatu: Verifies existing behavior as used by Stargate REST 1.0,
```
jeffreyscarpenter, line 122: Test is correct, but might look wrong if someone doesn't realize  that the behavior of $exists is expected to be equivalent to $EQ. It might be helpful to add a note here.

jeffreyscarpenter (approved):

restapi/src/test/java/io/stargate/web/resources/WhereParserTest.java
tatu-at-datastax: Good point; will add bit more info.

PR #1521, opened 2022-01-05
closes #1520: updated CI docker image, correct ccm versions, updated …

### LINKED ISSUE #1520: CI docker file uses old versions when creating and removing ccm
All versions are wrong..

```
# Create clusters to pre-download necessary artifacts
RUN ccm create -v 4.0.0 stargate_40 && ccm remove stargate_40
RUN ccm create -v 3.11.8 stargate_311 && ccm remove stargate_311
RUN ccm create -v 6.8.13 --dse stargate_dse68 && ccm remove stargate_dse68
```

ivansenic:
**What this PR does**:

Updates our CI docker image:
* updated Ubuntu
* set `ccm` to "prebake" currently used versions for the INT tests

Fixes #1520

**Checklist**
- [x] add `README.md` with update instructions
- [x] create `Makefile` for easier workflow

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -13,18 +13,19 @@ The docker image is directly referenced in the [cloudbuild.yaml](../cloudbuild.y
 
 ### Prerequisites
 
-In order to push the new docker image to the GCP image registry `gcr.io/stargateio`, besides having permissions to do so, you should authenticate and setup your docker configuration using:
+In order to push the new docker image to the GCP image registry `gcr.io/stargateio`, besides having permissions to do so, you should authenticate and set up your docker configuration using:
 
 ```bash
 gcloud auth login
 gcloud auth configure-docker
 ```
 
+That said, it's assumed that `gcloud` is available on the system, as it's going to be used for getting the existing image tags in the [Makefile](./Makefile).
+
 ### Process
 
 * Perform wanted updates related to the image.
 * Bump the `VERSION` variable in the [Makefile](Makefile).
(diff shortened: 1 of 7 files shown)


ci/README.md
```diff
@@ -0,0 +1,33 @@
...
+```bash
+gcloud auth login
+gcloud auth configure-docker
+```
+
+### Process
+
+* Perform wanted updates related to the image.
+* Bump the `VERSION` variable in the [Makefile](Makefile).
+**This step is mandatory, otherwise you will overwrite the existing image.**
```
dougwettlaufer, line 27: Why not make it a bash script that takes a version as input? I guess for the super paranoid you could always query GCR first and make sure the version doesn't exist before you tag and push.

ivansenic: I wanted to avoid the bash script, because with the Makefile it's cleaner imo and you can do `make build` or `make push`, and it's kind of easy :) But yea if you think script would be better I am also fine with this..

I can query GCR first, this is also very reasonable imo..

dougwettlaufer (approved):
Looks reasonable. Up to you if you want to do the extra credit of querying GCR to verify the version doesn't get overwritten.

ci/README.md
ivansenic: Done with the question and option to abort:

```
[10:08:35] ise:ci <ise-1520-ci-image-update>$ make push
The image with the version v1.0.6 exists, do you want to overwrite [Y/n]?
```

PR #1523, opened 2022-01-05
Follow up to #1477: implement $exists, $contains, $containsKey and $containsEntry operations

### LINKED ISSUE #1477: SGv2/REST: implement "getWithWhere()"
The last data-access REST operation not yet implemented is "getWithWhere()"; should be implemented.

tatu-at-datastax:
**What this PR does**:

Implement support for "$exists", "$contains", "$containsKey" and "$containsEntry"operations for StargateV2/REST, get-with-where

Fixes #1477

**Checklist**
- [ ] Changes manually tested
- [x] Automated Tests added/updated -- resolves one recently added previously failing test


testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -822,7 +822,7 @@ public void getRowsWithQuery() throws IOException {
     JsonNode data = getResponseWrapper.getData();
     assertThat(data.size()).isEqualTo(1);
     assertThat(data.at("/0/id").asText()).isEqualTo(rowIdentifier);
-    assertThat(data.at("/0/firstName").asText()).isEqualTo("John");
+    assertThat(data.at("/0/firstname").asText()).isEqualTo("John");
```
tatu-at-datastax, line 825: This is due to unfortunate discrepancy between SGv1 and SGv2 for this test class; not a difference in behavior per se (that is, SGv2 creates field with lower-case name so it is returned as such -- handling itself is case-insensitive).

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/WhereParser.java
```diff
@@ -65,7 +65,7 @@ public WhereParser(Schema.CqlTable tableDef, ToProtoConverter converter) {
 
       for (Map.Entry<String, JsonNode> operandAndValue : asIterable(condition.fields())) {
         String rawOp = operandAndValue.getKey();
-        FilterOp operator = FilterOp.decode(rawOp);
+        FilterOp operation = FilterOp.decode(rawOp);
```
tatu-at-datastax, line 68: Unifying to use term "operation" since our docs (Swagger) use that term.

jeffreyscarpenter (approved):

PR #1531, opened 2022-01-06
Fix 3 ITs that now expect "text" as type for all columns, tables etc.

tatu-at-datastax:
**What this PR does**:

Changes mapping from gRPC internal column type of `VARCHAR` (used internally by persistence) into `TEXT` which is the expected external type (see #1499).

No separate PR filed.

**Checklist**
- [ ] Changes manually tested
- [x] Automated Tests added/updated -- passes 3 regressed tests for StargateV2


tatu-at-datastax:
@olim7t does this make sense?

olim7t (approved):

PR #1534, opened 2022-01-07
Fix an IT in `SchemaOperationsTest` that broke due to Varchar/Text changes

tatu-at-datastax:
**What this PR does**:

Fixes a minor problem with existing StargateV2 integration test, regarding schema access.

No PR filed for this regression

**Checklist**
- [ ] Changes manually tested
- [x] Automated Tests added/updated -- fixes one failing integration test
- [ ] 


tatu-at-datastax:
@olim7t Does this make sense? My understanding is that internally gRPC will use "varchar" but externally we expose "text".
I noticed test failing in CI, as well as locally; passes after change.

jeffreyscarpenter (approved):

PR #1535, opened 2022-01-07
Docker images for Stargate v2

jeffreyscarpenter:
Build Docker images for Stargate v2 in Stargate repo to make it simpler to use images for testing during development


Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+# GraphQL
+EXPOSE 8080
+
+# Auth
+EXPOSE 8081
+
+# Health
+EXPOSE 8084
+
+# this will include all persistence jars, we'll remove the ones we don't want below
```
tatu-at-datastax, line 32: Would this still leave those jars as part of the image (or rather, use up space)? I recall Docker image construction is sometimes tricky as snapshot is taken between steps.

docker-compose/cassandra-3.11/docker-compose.yml
```diff
@@ -0,0 +1,75 @@
+version: '2'
+
+services:
+  backend-1:
```
tatu-at-datastax, line 4: I am not familir with Docker compose, so.... is this naming convention that must be followed? Or can we use more meaningful names? (I assume former)

tatu-at-datastax: Oh actually it looks like this can be freely selected (looking at later entries). So would it make sense to call it "cassandra-1" or "storage-1"?

tatu-at-datastax (approved):

docker-compose/cassandra-3.11/docker-compose.yml
tatu-at-datastax: I assume this mostly just follows existing naming convention within Datastax tho. If so, perhaps that makes sense.

jeffreyscarpenter: I preserved the previous conventions from the copies in [docker images](https://github.com/stargate/docker-images), but agree this could be more descriptive.

Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+# Auth
+EXPOSE 8081
+
+# Health
+EXPOSE 8084
+
+# we use a script to build a directory with contents of ./stargate-lib, omitting all persistence jars
+# we can then add the proper set of persistence jars to each of the images below
+ARG LIBDIR
+COPY ${LIBDIR} /stargate-lib/
```
tatu-at-datastax, line 28: Wouldn't this copy all jars, including persistence ones?

docker-compose/cassandra-3.11/start_cass_311.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
```
tatu-at-datastax, line 1: Newbie question: does this refer to `docker-compose.yml` via naming convention (defaulting)? Or how is the dependency indicated?

docker-compose/cassandra-4.0/README.md
```diff
@@ -0,0 +1,24 @@
...
+specified in `docker-compose.yml`:
+
+`./start_cass_40.sh`
+
+# Stargate with embedded Cassandra 4.0 (developer mode) 
+
+This alternate configuration that runs the Stargate coordinator node in developer mode, so that no
+separate Cassandra cluster is required. This can be run directly with the command:
+
+`docker-compose -f docker-compose-dev-mode.yaml up`
```
tatu-at-datastax, line 15: Ahh. I think this answers my earlier question; there's default name, but one that can be overridden.

ivansenic:
@jeffreyscarpenter I would like to invest time in reviewing this, but would need more context here.. So what are we trying to achive, are all the files newly written or was there some starting point, is all of this staying after in v2 or is something gonna become obsolete? Thanks in advance..

jeffreyscarpenter:
> @jeffreyscarpenter I would like to invest time in reviewing this, but would need more context here.. So what are we trying to achive, are all the files newly written or was there some starting point, is all of this staying after in v2 or is something gonna become obsolete? Thanks in advance..

Context:
One of the overall goals for v2 is to be more Docker/Kubernetes friendly. This PR represents one step in that direction, to to build Docker images in the main stargate repo as part of the standard build process so they can be more easily used during development. (This is different than the current process as implemented via the `docker-images` repo, which requires a Stargate release.) Other subsequent PRs will add actions to build, tag and publish these Docker images as part of our release process, and provide options to run some tests using these images using Docker and Kubernetes. It is likely that we'll want to deprecate the `docker-images` repo when v2 becomes the main version, but we don't have to make a final decision on that now.

Details:
The Dockerfile creates images that are similar to those created by `docker-images`, although they use a more recent JVM and are more streamlined. A single Dockerfile is used for simplicity and so we can have a common base image for the different coordinator images. The coordinator images have names with prefix `coordinator-` instead of `stargate-`. This provides a more accurate name of what these images contain and will help avoid confusion with v1 images. The fact that images will be tagged `v2.0.0-ALPHA1` (and following) when published will also help avoid unintentional use of the wrong images by current v1 users.

The Dockerfile also contains instructions for building the REST API image, although this could easily be broken out into a separate Dockerfile, for example if we decide to move service implementations to their own GitHub repos at some point in the future. You'll note that it uses Java 11.

The `docker-compose` files are largely based on the examples from `docker-images` with updates to have consistent naming. There are also new compose files that support running coordinators in "developer mode" that can be used for quick startup without having to start a Cassandra cluster. I've already found these pretty useful for local testing.

Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+
+FROM base as coordinator-dse-68
+COPY stargate-lib/persistence-api*.jar stargate-lib/persistence-dse*.jar /stargate-lib/
+
+#
+# Dockerfile for building REST API image
+#
+
+# Using Java 11
+FROM openjdk:11.0.13-jre-slim as restapi
```
ivansenic, line 48: is there anything that stops us from using java 17 here?

Dockerfile
ivansenic: Seems to me as well.. Ahh OK, seen the trick in the build file.. Why not simply going for:

```
RUN rm /stargate-lib/persistence*.jar
```

build_docker_images.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+set -euo pipefail
+
+# Script to generate Docker images for Stargate
+# Assumes that you have done a complete build so that all jars have been created, i.e.:
+#   mvn install -P dse
```
ivansenic, line 6: I think we should always point to a wrapper and skip tests:

`./mvnw install -P dse -DskipTests=true`

build_docker_images.sh
```diff
@@ -0,0 +1,28 @@
...
+# Create a temp directory under current directory to use as a staging area for image creation
+# This is a workaround since Docker COPY command provides no way to exclude files.
+# We could build a base image and then remove files, but this is wasteful
+# Note: Docker will not recognize files that are not under current directory
+LIBDIR=./tmp-${RANDOM}
+mkdir ${LIBDIR}
+cp ./stargate-lib/* $LIBDIR
+rm ${LIBDIR}/persistence*.jar
+
+#ls ${LIBDIR}
```
ivansenic, line 19: I guess this can be omitted if you apply the above proposed `RUN rm ..`?

docker-compose/cassandra-3.11/.env
```diff
@@ -0,0 +1,3 @@
+COMPOSE_PROJECT_NAME=cass311-stargate
+CASSTAG=3.11.11
```
ivansenic, line 2: Can you extend the `README.md` in the `persistence-cassandra-3.11` and add the update of this value to the list of tasks we need to do when upgrading the version?

docker-compose/cassandra-3.11/docker-compose.yml
```diff
@@ -0,0 +1,75 @@
...
+    image: cassandra:${CASSTAG}
+    networks:
+      - stargate
+    ports:
+      - 9044:9042
+    mem_limit: 2G
+    environment:
+      - HEAP_NEWSIZE=128M
+      - MAX_HEAP_SIZE=1024M
+      - CASSANDRA_SEEDS=cassandra-1
```
ivansenic, line 14: hmm is this correct, point to itself as the seed?

docker-compose/cassandra-3.11/docker-compose.yml
```diff
@@ -0,0 +1,75 @@
...
+    networks:
+      - stargate
+    mem_limit: 2G
+    depends_on:
+      - cassandra-2
+    environment:
+      - HEAP_NEWSIZE=128M
+      - MAX_HEAP_SIZE=1024M
+      - CASSANDRA_SEEDS=cassandra-1
+      - CASSANDRA_CLUSTER_NAME=c3-${CASSTAG}-cluster
```
ivansenic, line 39: Not sure why we have one coordinator and three C* nodes, but would it actually make more sense to have two each? Also not sure if we can use `replicas` here, at least for the coordinator?

docker-compose/cassandra-4.0/.env
```diff
@@ -0,0 +1,3 @@
+COMPOSE_PROJECT_NAME=cass40-stargate
+CASSTAG=4.0
```
ivansenic, line 2: should we also fix this to the version we currently support `4.0.1`?

docker-compose/dse-6.8/.env
```diff
@@ -0,0 +1,3 @@
+COMPOSE_PROJECT_NAME=dse68-stargate
+DSETAG=6.8.13
```
ivansenic, line 2: we use `6.8.16`

kubernetes/k8ssandra-values.yaml
```diff
@@ -0,0 +1,38 @@
+cassandra:
+  version: "4.0.0"
```
ivansenic, line 2: 4.0.1?

kubernetes/k8ssandra-values.yaml
```diff
@@ -0,0 +1,38 @@
...
+  cassandraLibDirVolume:
+    storageClass: local-path
+    size: 5Gi
+  allowMultipleNodesPerWorker: true
+  heap:
+    size: 1G
+    newGenSize: 1G
```
ivansenic, line 9: well this seems odd? new gen size matches the whole heap?

kubernetes/k8ssandra-values.yaml
```diff
@@ -0,0 +1,38 @@
...
+        - name: default
+stargate:
+  enabled: true
+  image:
+    registry: stargateio
+    repository: sgv2-coordinator-4_0
+    tag: v2.0.0-ALPHA1
+    pullPolicy: IfNotPresent
+  replicas: 1
+  heapMB: 256
```
ivansenic, line 30: should we push this a bit up?

ivansenic (changes requested):
Does not compile, probably needs rebasing, can not test the scripts.. I added some other comments as well, have a look..

I guess that the restapi image would become obsolete once we move to new framework, but for now it's good.. Really good work @jeffreyscarpenter :+1:

Dockerfile
jeffreyscarpenter: @ivansenic I originally had a step in the image creation process that copied in all of the jar files and then removed the undesired ones but @tatu-at-datastax pointed out that this would be unnecessarily bloat the image since the files are still technically there but just don't appear in the directory. Plus it s a bit more complex than removing all `persistence*.jar` files, there are actually ones you want to keep per image.

Dockerfile
jeffreyscarpenter: probably not, but I suggest we wait on this until we select the updated Java framework, which may make this section of the Dockerfile obsolete.

kubernetes/k8ssandra-values.yaml
jeffreyscarpenter: I'm going to pull the Kubernetes config from this PR, it is still a WIP that I included inadvertently.

docker-compose/cassandra-3.11/start_cass_311.sh
jeffreyscarpenter: yes, that's correct. `docker-compose.yml` is the default name.

build_docker_images.sh
jeffreyscarpenter: ok

docker-compose/cassandra-3.11/docker-compose.yml
jeffreyscarpenter: confirmed this is not necessary, removing

docker-compose/dse-6.8/.env
jeffreyscarpenter: fixing

docker-compose/cassandra-4.0/.env
jeffreyscarpenter: fixing

docker-compose/cassandra-3.11/.env
jeffreyscarpenter: fixing

docker-compose/cassandra-3.11/README.md
```diff
@@ -0,0 +1,26 @@
...
+specified in `docker-compose.yml`:
+
+`./start_cass_311.sh`
+
+# Stargate with embedded Cassandra 3.11 (developer mode)
+
+This alternate configuration that runs the Stargate coordinator node in developer mode, so that no
+separate Cassandra cluster is required. This can be run directly with the command:
+
+`docker-compose -f docker-compose-dev-mode.yaml up`
```
ivansenic, line 15: all `docker-compose` files have a `.yml` extension and not `.yaml`.. Either docs or files should be adapted..

ivansenic (changes requested):
The branch is still not compiling:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project testing: Compilation failure: Compilation failure: 
[ERROR] /home/ise/workspace/datastax/stargate/stargate/testing/src/main/java/io/stargate/it/grpc/SchemaTest.java:[45,13] cannot find symbol
[ERROR]   symbol:   method describeKeyspace(io.stargate.proto.Schema.DescribeKeyspaceQuery)
[ERROR]   location: variable stub of type io.stargate.proto.StargateGrpc.StargateBlockingStub
[ERROR] /home/ise/workspace/datastax/stargate/stargate/testing/src/main/java/io/stargate/it/grpc/SchemaTest.java:[64,70] cannot find symbol
[ERROR]   symbol:   method getName()
[ERROR]   location: class io.stargate.proto.QueryOuterClass.TypeSpec.Udt
[ERROR] /home/ise/workspace/datastax/stargate/stargate/testing/src/main/java/io/stargate/it/grpc/SchemaTest.java:[68,36] cannot find symbol
[ERROR]   symbol:   method getName()
[ERROR]   location: class io.stargate.proto.QueryOuterClass.TypeSpec.Udt
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <args> -rf :testing
```

I run the `./build_docker_images.sh` and nothing is being built.. I get:
```
$ ./build_docker_images.sh 
cp: -r not specified; omitting directory './stargate-lib/felix-cache'
cp: -r not specified; omitting directory './stargate-lib/log'
```

So I don't get it, is this working for you? If so then there is some issue with the mac/linux shell compatibility.. Let me know and maybe I can dig deeper?

docker-compose/cassandra-4.0/start_cass_40.sh
```diff
@@ -0,0 +1,32 @@
...
+docker-compose up -d coordinator restapi
+
+# Wait until stargate is up before bringing up zeppelin
+
+echo ""
+echo "Waiting for stargate to start up..."
+while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8082/health)" != "200" ]]; do
+    printf '.'
+    sleep 5
+done
```
ivansenic, line 32:
```suggestion
while [ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8082/health)" != "200" ]; do
    printf '.'
    sleep 5
done
```

ivansenic (approved):
I am approving, but there are 4 remaining comments and some suggestions to be solved please :pray:

docker-compose/cassandra-4.0/start_cass_40.sh
jeffreyscarpenter: fixed

docker-compose/cassandra-3.11/start_cass_311.sh
```diff
@@ -0,0 +1,32 @@
...
+docker-compose up -d coordinator restapi
+
+# Wait until stargate is up before bringing up zeppelin
+
+echo ""
+echo "Waiting for stargate to start up..."
+while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8082/health)" != "200" ]]; do
+    printf '.'
+    sleep 5
+done
```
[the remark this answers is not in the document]
jeffreyscarpenter: fixed

docker-compose/dse-6.8/start_dse_68.sh
```diff
@@ -0,0 +1,34 @@
...
+docker-compose up -d coordinator restapi
+
+# Wait until stargate is up before bringing up zeppelin
+
+echo ""
+echo "Waiting for stargate to start up..."
+while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8082/health)" != "200" ]]; do
+    printf '.'
+    sleep 5
+done
```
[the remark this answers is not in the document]
jeffreyscarpenter: fixed

Dockerfile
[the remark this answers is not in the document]
jeffreyscarpenter: agree, also changing persistence base image to `openjdk:8-jre-slim`, this will give us latest JVMs. We do not need strict reproducibility of Docker image builds.

docker-compose/cassandra-3.11/docker-compose.yml
```diff
@@ -0,0 +1,74 @@
...
+      - DATACENTER_NAME=datacenter1
+      - ENABLE_AUTH=true
+  restapi:
+    image: stargateio/restapi:${SGTAG}
+    depends_on:
+      - coordinator
+    networks:
+      - stargate
+    ports:
+      - 8082:8082
```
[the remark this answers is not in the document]
jeffreyscarpenter: fixed, this port expose command was left in by mistake.

build_docker_images.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+set -euo pipefail
+
+# Script to generate Docker images for Stargate
+# Assumes that you have done a complete build so that all jars have been created, i.e.:
+#   ./mvnw install -P dse -DskipTests=true
```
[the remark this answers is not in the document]
jeffreyscarpenter: added clean to the sample `mvnw` command

docker-compose/cassandra-3.11/README.md
jeffreyscarpenter: updated readme files.

tatu-at-datastax:
The only additional comment/question I have is about use of `latest` for JDK: I can see benefits of always using the latest, as 
JDK maintainers have very good tracker with compatibility (regressions rare). But at the same time it is usually a good practice to know exact version that is being packaged.

So: do we think in general that we should just use the latest patch for given JDK major version, trusting `latest` tagging?

ivansenic:
> The only additional comment/question I have is about use of `latest` for JDK: I can see benefits of always using the latest, as JDK maintainers have very good tracker with compatibility (regressions rare). But at the same time it is usually a good practice to know exact version that is being packaged.
> 
> So: do we think in general that we should just use the latest patch for given JDK major version, trusting `latest` tagging?

I was the one voting for the latest.. It's only to get the Java patches and security improvements as fast as possible.. Otherwise, you kinda forget to update the image base.. Same is already happening for us in the `docker-images` repo, we rarely update the base version.

tatu-at-datastax:
@ivansenic I agree that it's very easy to just never update anything. And I personally have never observed an actual regression by latest JDK patches. So I'm fine with this. I just wanted to ask if anyone felt differently since I know there are developers who consider reproducible build (and explicit versions) important.

So, for JDKs, let's just use the latest patch for major version as suggested.

PR #1537, opened 2022-01-10
Add support for UDT inserts/gets in Stargate V2 REST API

tatu-at-datastax:
**What this PR does**:

Adds support for inserting/getting User Defined Types (UDTs) in Stargate V2 REST API.

No PR filed.

**Checklist**
- [ ] Changes manually tested
- [x] Automated Tests added/updated -- passes test added in #1536

### DIFF
--- a/testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
+++ b/testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
@@ -1494,6 +1494,45 @@ public class RestApiv2Test extends BaseIntegrationTest {
   }
 
   @Test
+  public void getRowsWithUDT() throws IOException {
+    createKeyspace(keyspaceName);
+
+    // create UDT: note -- UDT names must be lower-case it seems (mixed case fails)
+    String udtString =
+        "{\"name\": \"test_udt\", \"fields\":"
+            + "[{\"name\":\"name\",\"typeDefinition\":\"text\"},"
+            + "{\"name\":\"age\",\"typeDefinition\":\"int\"}]}";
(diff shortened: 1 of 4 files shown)


testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -1493,6 +1493,45 @@ public void getRows() throws IOException {
     assertThat(data.get(0).get("firstname")).isEqualTo("John");
   }
 
+  @Test
+  public void getRowsWithUDT() throws IOException {
+    createKeyspace(keyspaceName);
+
+    // create UDT: note -- UDT names must be lower-case it seems (mixed case fails)
```
jeffreyscarpenter, line 1500: What is the failure? I'm not aware of a requirement for UDT names to be lower case

jeffreyscarpenter (approved):

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
tatu-at-datastax: I'll check this again, I don't remember off-hand. Observed with V1.

tatu-at-datastax: @jeffreyscarpenter Failure message:

    ERROR [main] 2022-01-10 09:23:37,246 RestUtils.java:279 - {"description":"Bad request: Malformed type name: testUdt","code":400}

which does not occur if `test_udt` is used. I wonder if this is due to added quoting/escaping.

PR #1551, opened 2022-01-13
Implement /metrics for SGv2/REST, change matching ITs to pass

tatu-at-datastax:
**What this PR does**:

Adds proper implementation of `/metrics` endpoint for Stargate V2 REST API; modify matching tests to use proper module names for tests to pass.

No PR filed, part of Stargate V2 work.

**Checklist**
- [ ] Changes manually tested


testing/src/main/java/io/stargate/it/http/RestApiMetricsTest.java
```diff
@@ -96,22 +104,33 @@ public void restApiHttpRequestMetrics() throws IOException {
                       metric ->
                           assertThat(metric)
                               .contains("method=\"GET\"")
-                              .contains("module=\"restapi\"")
+                              // NOTE! Hyphens remain and are NOT converted to underscores
```
ivansenic, line 107: this is fine, tag values are as is..

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/impl/RestServiceStarter.java
```diff
@@ -59,10 +58,10 @@ public void start() throws Exception {
     // pass whatever we want/need
 
     // Then Metrics-related things: hard-coded initially
-    final Metrics metricsImpl = new MetricsImpl();
+    final MetricsImpl metricsImpl = new MetricsImpl();
```
ivansenic, line 61: So to enable percentiles you should call the `MeterRegistryConfiguration.configure(meterRegistry);` using the `meterRegistry` from the `MetricsImpl`..

However, until now this was done in the constructor by calling the `MetricsImpl#initPrometheusMeterRegistry` to configure everything.. But I guess the `MetricRegistry` is out of the play now, so you maybe changed some code in the `MetricsImpl` for the v2?

Anyway this is solvable easily.

testing/src/main/java/io/stargate/it/http/RestApiMetricsTest.java
```diff
@@ -50,9 +49,13 @@
   // TODO
   @SuppressWarnings("unused") // referenced in @StargateSpec
   public static void buildParameters(StargateParameters.Builder builder) {
-    builder.putSystemProperties(
-        TestingServicesActivator.HTTP_TAG_PROVIDER_PROPERTY,
-        TestingServicesActivator.TAG_ME_HTTP_TAG_PROVIDER);
+    // 13-Jan-2022, tatu: In StargateV1 HTTP Tag Provider is registered using OSGi.
+    //    SGv2 does not have equivalent mechanism implemented yet, so tag functionality
+    //    NOT verified currently.
+    /*builder.putSystemProperties(
```
ivansenic, line 55: This is a problem, we need to find a solution for this.. Not sure if we should postpone this, but I don't feel comfortable to disable tests and let this go in.. We will forget about it later on.. 

However, I have no idea how to do this now.. Maybe we should sync here..

ivansenic (commented):
Percentiles are solvable, other thing not so easily..

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/impl/RestServiceStarter.java
tatu-at-datastax: I use `MetricsImpl` exactly as-is with no changes so far. But initialization/construction may have changed.

Odd thing is that percentiles seem to be added for some metrics by default, but not for specific ones being tested.
At least when running thing locally and having a look at `/metrics` endpoint.
So I guess I do not understand logic of them getting added or not.

testing/src/main/java/io/stargate/it/http/RestApiMetricsTest.java
tatu-at-datastax: My primary concern right now is to get the integration test to completely pass, since this helps catch many other regressions. It'd be great to keep tags working if possible, but I think that since we also need to tackle change of frameworks, perhaps metrics part would be part of this effort -- the way service is boostrapped/initialized will change a bit anyway so making things work with DropWizard here may be work that does not help making things work with the newer framework.

Still, yes, I think we should sync.

tatu-at-datastax: And just to make it clear: since tags are added in (and maybe used) Prod, this functionality would need to be fixed/added back before SGv2/REST would go in Prod. So disabling of testing, if any, should be temporary.
And not "just leave it out".

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/impl/RestServiceStarter.java
tatu-at-datastax: I think `MetricsRegistry` should also exist.

But one thing I was wondering about was the whole way in which static initialization was called from non-static context. And whether that ordering might be fragile so that seemingly unrelated change in set up of other components might trigger changes in timing of metrics registry initialization.

tatu-at-datastax: I am not sure about those changes; @jeffreyscarpenter is bit more familiar with it, but I think this was straight copy-paste to split REST metrics test from others.

jeffreyscarpenter (approved):

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/impl/RestServiceStarter.java
jeffreyscarpenter: I am missing the context. What code was copy-pasted?

tatu-at-datastax:
Will merge and file a follow-up issue for work to connect metrics tagger for v2.0.0, likely as part of the new framework -- registration of Tag provider depends to a degree on framework and we may want to improve on registration mechanism (f.ex maybe avoid system properties).

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/impl/RestServiceStarter.java
tatu-at-datastax: @jeffreyscarpenter  There was a test class like `MetricsTest`, containing tests for metrics of all APIs/services; a new one (like `RestApiMetricsTest`) was extracted from that. I was thinking you had done the initial part earlier and might have context.

ivansenic: Yes the copy paste is fine, what about support for the `@StargateSpec(parametersCustomizer = "buildParameters")`? So I was not lazy and debugged this:

* The second JVM that you fire in the INT tests and that I guess represents the `REST v2` is completely ignoring the those properties
* It's because `RestApiExtension` has different way of supplying properties and this was not included in the test

Anyway I fixed this in #1571

PR #1557, opened 2022-01-14
Handle bridge auth for background operations

olim7t:
With gRPC being the internal transport in v2, there are certain operations that are executed in the background, for example maintaining a schema metadata cache. In that context we don't have a user token to authenticate, so this change introduces a special "admin token", that will be managed as a shared secret between the gRPC server and the client Stargate services.

Obviously this is not for public gRPC. We plan to split the two APIs, this would only exist in the internal one.

### DIFF
--- a/core/src/main/java/io/stargate/core/grpc/BridgeConfig.java
+++ b/core/src/main/java/io/stargate/core/grpc/BridgeConfig.java
@@ -1,0 +1,18 @@
+package io.stargate.core.grpc;
+
+public class BridgeConfig {
+
+  /**
+   * The token that can be used by Stargate services to query the gRPC bridge when no user token is
+   * available (e.g. background queries).
+   */
+  public static final String ADMIN_TOKEN;
+
+  static {
+    String propertyName = "stargate.grpc.admin_token";
+    ADMIN_TOKEN = System.getProperty(propertyName);
+    if (ADMIN_TOKEN == null || ADMIN_TOKEN.isEmpty()) {
+      throw new IllegalStateException(String.format("'%s' must be provided", propertyName));
+    }
+  }
+}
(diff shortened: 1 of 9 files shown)


ivansenic (commented):
Hmm, I understand the need for this change, but isn't this change alone not sufficient. Now you would be able to actually bypass the authentication in the publicly available GRPC? Shouldn't we immediately add the "protection" against this? Or am I getting something wrong?

Sure we could add this in another PR, but this then becomes top-priority and something we must do before merging v2?

prabhatjha:
I think both @olim7t and @ivansenic are agreeing that we can not bypass some sort of authentication, difference should be around timing. Given that it has potential to expose data we should make it secure from the get go.

ivansenic (commented):
Yea this seems better.. For the final version in v2, I think following things do make sense:

* The `admin_token` can not be used in the public gRPC API
* For the internal point, if `admin_token` is set on the coordinator then it must be provided
* To allow easier dev, we should allow internal point communication without the `admin_token`, otherwise it becomes a bit complex when developing

olim7t:
> The `admin_token` can not be used in the public gRPC API

Yes, that's the intention. It's not very clear right now, but I plan on working on separating the internal and external APIs next.

> For the internal point, if `admin_token` is set on the coordinator then it must be provided

I think it should be mandatory everywhere. If services work without setting it explicitly that kind of defeats the purpose.

> To allow easier dev, we should allow internal point communication without the `admin_token`, otherwise it becomes a bit complex when developing

`StargateExtension` already sets it on the bridge. It would only need to be added when starting services like `RestServiceStarter`.

tatu-at-datastax (approved):

PR #1559, opened 2022-01-18
Fix #1558: allow insertion of JSON `null` values via REST API

### LINKED ISSUE #1558: Should allow JSON `null` values for inserts via REST API
Currently Stargate REST API does not accept `null` values from JSON passed for insert endpoint. Attempt to, for example, add explicit `null` for UUID-valued non-key column results in HTTP 400 failure with error message of:

```
{"description":"Bad request: Invalid Uuid value 'null': expected a string","code":400}
```

But while most of the time `null`s are not explicitly inserted insertion should be legal for non-key CQL fields.

tatu-at-datastax:
**What this PR does**:

Allows use of `null` values with JSON payload when inserting rows using REST API.

Also improves testing, including `null` inserting case (for UUIDs): no tests covered this case; and in general JSON-payload taking inserts were (and are) under tests compared to "stringified" payload.

Fixes #1558

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/restapi/src/main/java/io/stargate/web/resources/Converters.java
+++ b/restapi/src/main/java/io/stargate/web/resources/Converters.java
@@ -203,7 +203,9 @@ public class Converters {
   /** Converts an incoming JSON value into a Java type suitable for the given column type. */
   @SuppressWarnings("unchecked")
   public static Object toCqlValue(Column.ColumnType type, Object jsonValue) {
-
+    if (jsonValue == null) {
+      return null;
+    }
     if (jsonValue instanceof String) {
       return toCqlValue(type, (String) jsonValue);
     }
--- a/testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
+++ b/testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
@@ -24,7 +24,6 @@ import com.datastax.oss.driver.api.core.cql.ResultSet;
 import com.datastax.oss.driver.api.core.cql.Row;
 import com.datastax.oss.driver.api.core.cql.SimpleStatement;
 import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableMap;
-import com.fasterxml.jackson.annotation.JsonValue;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.JsonNode;
@@ -74,17 +73,6 @@ import org.junit.jupiter.api.extension.ExtendWith;
 @ExtendWith(CqlSessionExtension.class)
 @CqlSessionSpec()
 public class RestApiv2Test extends BaseIntegrationTest {
-  // Due to Guava's null hatred, ImmutableMap cannot contain nulls.
-  // But we have a need for null values in JSON so here's something that
-  // will get serialized as JSON null.
-  final Object GUAVA_NULL_VALUE =
-      new Object() {
-        @JsonValue
-        public Integer nullValue() {
-          return null;
-        }
-      };
-
   private String keyspaceName;
   private String tableName;
   private static String authToken;
@@ -1520,7 +1508,12 @@ public class RestApiv2Test extends BaseIntegrationTest {
             ImmutableMap.of(
                 "id", "1", "data", Arrays.asList(28, false, "foobar"), "alt_id", altUid1),
             ImmutableMap.of(
-                "id", "2", "data", Arrays.asList(39, true, "bingo"), "alt_id", GUAVA_NULL_VALUE)));
+                "id",
+                "2",
+                "data",
+                Arrays.asList(39, true, "bingo"),
+                "alt_id",
+                objectMapper.nullNode())));
     JsonNode json = readRawRowsBySingleKey(keyspaceName, tableName, "2");
     assertThat(json.size()).isEqualTo(1);
     assertThat(json.at("/0/id").asText()).isEqualTo("2");
--- a/testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
+++ b/testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
@@ -22,6 +22,8 @@ import com.datastax.oss.driver.api.core.CqlSession;
 import com.datastax.oss.driver.api.core.cql.ResultSet;
 import com.datastax.oss.driver.api.core.cql.Row;
 import com.datastax.oss.driver.api.core.cql.SimpleStatement;
+import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableMap;
+import com.fasterxml.jackson.annotation.JsonValue;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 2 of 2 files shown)


restapi/src/main/java/io/stargate/web/resources/Converters.java
```diff
@@ -203,7 +203,9 @@ public static ValueModifier colToValue(String name, Object value, AbstractTable
   /** Converts an incoming JSON value into a Java type suitable for the given column type. */
   @SuppressWarnings("unchecked")
   public static Object toCqlValue(Column.ColumnType type, Object jsonValue) {
-
+    if (jsonValue == null) {
```
tatu-at-datastax, line 206: Looks like this is sufficient as the actual fix: driver(s) seem capable of then translating value as necessary for backend access (or failing if "null"s are actually not allowed)

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -71,6 +73,16 @@
 @ExtendWith(CqlSessionExtension.class)
 @CqlSessionSpec()
 public class RestApiv2Test extends BaseIntegrationTest {
+  // Due to Guava's null hatred, ImmutableMap cannot contain nulls.
+  // But we have a need for null values in JSON so here's something that
+  // will get serialized as JSON null.
+  private final Object GUAVA_NULL_VALUE =
```
tatu-at-datastax, line 79: As per comment, Guava barfs of `null`s as not just Map keys but values. But we can relatively easily use a "token" that serializes (via Jackson) as `null`; this is one possible way to achieve that.

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -71,6 +73,16 @@
...
+      new Object() {
+        @JsonValue
```
tatu-at-datastax, line 81: Means literally "use return value of this method as value to actually serialize": Jackson calls method, gets `null`, serializes as JSON null.

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -1467,17 +1479,22 @@ public void getRows() throws IOException {
...
+        Arrays.asList("id text", "data tuple<int,boolean,text>", "alt_id uuid"),
         Collections.singletonList("id"),
         Collections.emptyList());
+    String altUid1 = UUID.randomUUID().toString();
     insertTestTableRows(
         Arrays.asList(
-            Arrays.asList("id 1", "data (28,false)"), Arrays.asList("id 2", "data (39,true)")));
+            // Put UUID for the first row; leave second one empty/missing
+            Arrays.asList("id 1", "data (28,false,'foobar')", "alt_id " + altUid1),
+            Arrays.asList("id 2", "data (39,true,'bingo')")));
```
tatu-at-datastax, line 1497: Will NOT yet attempt explicit null, since Stringified handler will not accept it. But whether it should or not is open to further discussion -- there is no easy way to tell apart String value "null" from something that we'd like to be `null`.

So for now fix is strictly for JSON payload case.

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -3311,14 +3361,22 @@ private void createTestIndex(
...
   private List<Map<String, String>> insertTestTableRows(List<List<String>> rows)
       throws IOException {
     final List<Map<String, String>> insertedRows = new ArrayList<>();
     for (List<String> row : rows) {
       Map<String, String> rowMap = new HashMap<>();
       for (String kv : row) {
-        String[] parts = kv.split(" ");
+        // Split on first space, leave others in (with no limit we'd silently
+        // drop later space-separated parts)
+        String[] parts = kv.split(" ", 2);
```
tatu-at-datastax, line 3379: Although no tests exhibited the issue, I had this problem earlier when trying to write a test: "trailing" values were silently just ignored. Instead it will now split String into exactly 2 parts, with the first space.
(previously would have split to however many segments and drop all but 2: something for which there is no use case)

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -933,13 +946,12 @@ public void getRowsWithQueryAndInvalidSort() throws IOException {
...
 
     String whereClause = String.format("{\"id\":{\"$eq\":\"%s\"}}", rowIdentifier);
-    String body =
-        RestUtils.get(
-            authToken,
-            String.format(
-                "%s:8082/v2/keyspaces/%s/%s?where=%s&sort={\"expense_id\"\":\"desc\"}",
-                host, keyspaceName, tableName, whereClause),
-            HttpStatus.SC_BAD_REQUEST);
+    RestUtils.get(
```
tatu-at-datastax, line 949: Removed reference to return value as ErrorProne complains about it.

restapi/src/main/java/io/stargate/web/resources/Converters.java
ivansenic: Yep this seems enough..

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
ivansenic: Why not using a simple `NullNode.getInstance()`? You anyway have `Map<String, Object>` in the rows creation?

ivansenic (approved):
I don't really understand the test changes.. The code is so confusing that it's impossible to review.. I give approve, but count that you have the last say if changes are tested enough or not..

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
tatu-at-datastax: Actually, good point, will change.

Yes, plain `Map` used in many places over ImmutableMap. So perhaps better avoid its use here in general.

tatu-at-datastax:
@ivansenic Aside from comments for changes which make sense, are there specific places where changes do not make sense? This test class is way too big and existing code confusing in place.

Some changes to tests are not strictly minimal: I think existing tests leave out validation of state so I sometimes add more explicit verification.

testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
```diff
@@ -1900,6 +1938,26 @@ public void updateRowRaw() throws IOException {
     @SuppressWarnings("unchecked")
     Map<String, String> data = objectMapper.readValue(body, Map.class);
     assertThat(data).containsAllEntriesOf(rowUpdate);
+
+    // Also verify that we can "delete" lastName
```
tatu-at-datastax, line 1942: (that is, set field to `null` with update -- exercises null usage via different path)

jeffreyscarpenter (approved):

ivansenic:
> @ivansenic Aside from comments for changes which make sense, are there specific places where changes do not make sense? This test class is way too big and existing code confusing in place.
> 
> Some changes to tests are not strictly minimal: I think existing tests leave out validation of state so I sometimes add more explicit verification.

@tatu-at-datastax Yea I meant existing code, not your updates..

tatu-at-datastax:
Ah ok :)

Just asked since I don't feel my changes were super clear either. But they need to work with existing code so trying to limit changes.

PR #1561, opened 2022-01-19
Remove gRPC create operations

olim7t:
**What this PR does**:
Remove dedicated `CreateXxx` operations from the gRPC API. They are somewhat redundant, since CREATE statements can be constructed with the query builder, and passed to the generic `ExecuteQuery`.

Keeping our API lean is beneficial:
- less code to maintain
- less divergence from the v1 codebase
  - easier merges (the last one was particularly bad, specifically because of these two operations)
  - extracting a separate "internal" API will be easier if the code is less entangled

Side note: once `SchemaCache` becomes the primary way to access schema metadata, we could also remove describe operations.

N/A

**Checklist**
- [x] Changes manually tested


tatu-at-datastax (approved):

tatu-at-datastax:
Assuming tests in `RestApiv2Test` still pass should be good. I think removal of these methods added earlier in v2 has been discussed and agreed upon.

PR #1562, opened 2022-01-19
refactor update document sub-path to use new write service

ivansenic:
**What this PR does**:
Continues refactoring of the Document API write path.. This PR handles updates of sub-documents and execution of built-in functions.

:information_source: *Improvement:* Put with the JSON primitive that is targeting a sub-document path is now allowed. Meaning you can do a put in a form (cc @polandll):

```
curl -X POST 
    -H "Content-Type: application/json" \
    -d 'true' \
    https://.../v2/namespaces/keyspace/collection/document-id/key
```

:information_source: *Bug fix:* `$push` and `$pop` to the nested arrays were not working, meaning you could not push to a document like `[{"array":[]}]` targeting the nested array with `/doc-id/[0]/array/function`

:information_source: Note that I removed `HttpHeaders` and `UriInfo` parameters from resource methods as they are not used.

Internal ticket.

**Open questions for @EricBorczuk regarding built in functions**
* What was the intention with the `ExecuteBuiltInFunction` DTO and the `value` being an object? How to convert this to the correct JsonNode? I saw you used object all way, but that's not possible anymore.
* Can I move the `value` to be `JsonNode`? If yes, how do we want to handle cases where there is no value in the `$push`, until now this was resolved as the bad request, but new implementation would construct the `NullNode` and insert null to the array.
* Another options is to use `valueToTree` conversion (current code changes have that).
* Do any unit tests exist for the execute built-in functions? Could not locate any.

**Checklist**
- [x] Clear-up the Reactive document service now
- [x] ~Delete unused code~ (still not possible)
- [x] Resolve `//TODO`s added
- [x] Unit tests for the execute built-in functions

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
@@ -36,6 +36,7 @@ import io.stargate.auth.Scope;
 import io.stargate.auth.SourceAPI;
 import io.stargate.auth.UnauthorizedException;
 import io.stargate.core.util.TimeSource;
+import io.stargate.db.datastore.ResultSet;
 import io.stargate.db.datastore.Row;
 import io.stargate.web.docsapi.dao.DocumentDB;
 import io.stargate.web.docsapi.dao.Paginator;
@@ -63,9 +64,7 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
-import javax.ws.rs.core.PathSegment;
 import org.javatuples.Pair;
-import org.jsfr.json.JsonSurferJackson;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -78,7 +77,6 @@ public class ReactiveDocumentService {
   @Inject DocumentWriteService writeService;
   @Inject JsonConverter jsonConverter;
   @Inject JsonSchemaHandler jsonSchemaHandler;
-  @Inject DocsShredder docsShredder;
   @Inject JsonDocumentShredder jsonDocumentShredder;
   @Inject ObjectMapper objectMapper;
   @Inject TimeSource timeSource;
@@ -92,7 +90,6 @@ public class ReactiveDocumentService {
       DocumentWriteService writeService,
       JsonConverter jsonConverter,
       JsonSchemaHandler jsonSchemaHandler,
-      DocsShredder docsShredder,
       JsonDocumentShredder jsonDocumentShredder,
       ObjectMapper objectMapper,
       TimeSource timeSource,
@@ -102,7 +99,6 @@ public class ReactiveDocumentService {
     this.writeService = writeService;
     this.jsonConverter = jsonConverter;
     this.jsonSchemaHandler = jsonSchemaHandler;
-    this.docsShredder = docsShredder;
     this.jsonDocumentShredder = jsonDocumentShredder;
     this.objectMapper = objectMapper;
     this.timeSource = timeSource;
@@ -579,7 +575,7 @@ public class ReactiveDocumentService {
    * from the end of an array, respectively.
    *
    * @param db @link DocumentDB} to execute the function in
-   * @param keyspace the keyspace to execute the function against
+   * @param namespace the keyspace to execute the function against
    * @param collection the collection to execute the function against
    * @param id the document ID to execute the function against
    * @param funcPayload information about the function to execute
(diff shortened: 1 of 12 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -177,28 +173,55 @@ public ReactiveDocumentService(
...
+   * document with the same ID will be overwritten.
+   *
+   * @param db {@link DocumentDB} to write in
+   * @param namespace Namespace
+   * @param collection Collection name
+   * @param documentId The ID of the document to update
+   * @param payload Document represented as JSON string
+   * @param context Execution content
+   * @return Document response wrapper containing the generated ID.
+   */
```
ivansenic, line 191: wrong javadoc

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -562,24 +585,36 @@ public ReactiveDocumentService(
...
+          // based on type switch the result
+          Maybe<JsonNode> result = null;
           if (function == BuiltInApiFunction.ARRAY_PUSH) {
-            result =
-                handlePush(
-                    db, keyspace, collection, id, funcPayload.getValue(), pathString, context);
+            // explicit null check to ensure we have a value to push
+            if (null == funcPayload.getValue()) {
+              throw new ErrorCodeRuntimeException(ErrorCode.DOCS_API_INVALID_JSON_VALUE);
+            }
```
ivansenic, line 610: Seems to be handled in the resource, remove

ivansenic (commented):
Self-review..

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -562,24 +585,36 @@ public ReactiveDocumentService(
...
+
+            // TODO Eric what's the desired object here and how would this me mapped anyway?
```
EricBorczuk, line 612: Yea I think it is reasonable to make it a JsonNode higher up, since virtually all the time this piece of data has to be JSON of some sort

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -649,6 +649,51 @@ public void testPutDocPathProfile() throws IOException {
     assertThat(OBJECT_MAPPER.readTree(resp).get("profile").isEmpty()).isFalse();
   }
 
+  @Test
+  public void testPutOverwriteWithPrimitive() throws IOException {
```
EricBorczuk, line 653: :+1:

testing/src/main/java/io/stargate/it/http/docsapi/BaseDocumentApiV2Test.java
```diff
@@ -3513,6 +3540,34 @@ public void testBuiltInPushPopFunction() throws IOException {
                 "[\"New York Bulls\",\"Los Angeles Kings\",\"Golden State Warriros\",\"Huston Rocket\"]"));
   }
 
+  @Test
+  public void testBuiltInPushPopFunctionWithNestedArray() throws IOException {
```
EricBorczuk, line 3544: Nice bugfix here!

EricBorczuk (approved):
This looks great! I like the refactor and think it will lend itself to upcoming changes. I pointed out above where there is a test for builtin functions (there might not be a unit test, however). The added tests look good as well!

tatu-at-datastax:
One tiny thing that may or may not be useful to know: in addition to `NullNode`, there is also `MissingNode` which represents absence of `JsonNode`. Jackson exposes that with some methods (`node.path()`, `node.at()`), and it allows avoiding null checks for traversal. But it could also add as a token node if there is need to indicate difference between explicit `null` value, and actual absence of value.

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -562,24 +585,36 @@ public ReactiveDocumentService(
...
+            //  should we fix the value in the Function payload to JsonNode?
```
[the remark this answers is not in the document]
ivansenic: @tatu-at-datastax @EricBorczuk Setting the `JsonNode value` in the [ExecuteBuiltInFunction](https://github.com/stargate/stargate/blob/576cc162a0a3592b0a59720666e76108888b7ae0/restapi/src/main/java/io/stargate/web/docsapi/models/dto/ExecuteBuiltInFunction.java) cannot distinguish between null and missing node.. There is an issue here that explains it [FasterXML/jackson-databind#3214](https://github.com/FasterXML/jackson-databind/issues/3214)

So I would like to have it, but seems that it's not working at least at the moment.. If you have any other idea I would be up for a fix.. Until then, we can keep it like this I guess.. We want to allow pushing of `null`s, but we also want that to be explicit and report bad request in case value is not set..

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
ivansenic: See comment https://github.com/stargate/stargate/pull/1562#discussion_r788530425

restapi/src/main/java/io/stargate/web/docsapi/service/write/DocumentWriteService.java
```diff
@@ -139,9 +140,48 @@ public DocumentWriteService(TimeSource timeSource, DocsApiConfiguration config)
...
+      String collection,
+      String documentId,
+      List<String> subDocumentPath,
+      List<JsonShreddedRow> rows,
+      boolean numericBooleans,
+      ExecutionContext context) {
+    // TODO for review
+    //  should we double check that each row begins with the given sub-path
+    //  ensuring correctness of the update
+
```
ivansenic, line 181: @EricBorczuk what about this todo? passed `subDocumentPath` is only used here for the delete query.. It's expected that rows are already created with `subDocumentPath` in mind.. Basically it means you can call this method to update a document at the sub-path, but provide rows that point to some completely other place in the document..

WDYT?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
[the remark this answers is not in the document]
EricBorczuk: Yeah I guess for now it has to remain this way, then

restapi/src/main/java/io/stargate/web/docsapi/service/write/DocumentWriteService.java
EricBorczuk: Yeah I think it's reasonable as a precondition to make sure that the sub path matches the rows' path

ivansenic: Added. Will merge on green.

PR #1564, opened 2022-01-19
Fix error handling in SchemaNotificationsHandler

olim7t:
**What this PR does**:
- ensure that `responseObserver.onError` in `SchemaNotificationsHandler` is called at most once (there were CI failures indicating that it was called multiple times).
- synchronize calls to the observer; not sure if this was a direct factor, but we've seen in other PRs that this is necessary.
- be more defensive about unexpected errors in `onSchemaChange`; otherwise exceptions bubble up to the Cassandra code that invoked the listener, and end up failing a completely unrelated request.

N/A

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/grpc/src/main/java/io/stargate/grpc/service/SchemaNotificationsHandler.java
+++ b/grpc/src/main/java/io/stargate/grpc/service/SchemaNotificationsHandler.java
@@ -35,6 +35,7 @@ class SchemaNotificationsHandler implements EventListener {
 
   private final Persistence persistence;
   private final ServerCallStreamObserver<SchemaNotification> responseObserver;
+  private boolean isFailed;
 
   SchemaNotificationsHandler(
       Persistence persistence, StreamObserver<SchemaNotification> responseObserver) {
@@ -52,30 +53,32 @@ class SchemaNotificationsHandler implements EventListener {
 
   private void onSchemaChange(
       Type type, Target target, String keyspace, String name, List<String> argumentTypes) {
-    try {
-      SchemaChange.Builder change =
-          SchemaChange.newBuilder().setChangeType(type).setTarget(target).setKeyspace(keyspace);
-      if (name != null) {
-        change.setName(StringValue.of(name));
+    synchronized (responseObserver) {
+      if (isFailed) {
+        return;
       }
-      change.addAllArgumentTypes(argumentTypes);
-
-      SchemaNotification.Builder notification = SchemaNotification.newBuilder().setChange(change);
-      if (type != Type.DROPPED || target != Target.KEYSPACE) {
-        notification.setKeyspace(SchemaHandler.buildKeyspaceDescription(keyspace, persistence));
-      }
-      synchronized (responseObserver) {
-        responseObserver.onNext(notification.build());
-      }
-    } catch (Throwable t) {
       try {
-        synchronized (responseObserver) {
-          responseObserver.onError(t);
+        SchemaChange.Builder change =
+            SchemaChange.newBuilder().setChangeType(type).setTarget(target).setKeyspace(keyspace);
+        if (name != null) {
+          change.setName(StringValue.of(name));
         }
-      } catch (Throwable t2) {
-        // Be defensive here because the Cassandra internals don't guard against listener errors.
-        t2.addSuppressed(t);
-        LOG.warn("Unexpected error while notifying error", t2);
+        change.addAllArgumentTypes(argumentTypes);
+
+        SchemaNotification.Builder notification = SchemaNotification.newBuilder().setChange(change);
+        if (type != Type.DROPPED || target != Target.KEYSPACE) {
+          notification.setKeyspace(SchemaHandler.buildKeyspaceDescription(keyspace, persistence));
+        }
+        responseObserver.onNext(notification.build());
+      } catch (Throwable t) {
+        try {
+          isFailed = true;
+          responseObserver.onError(t);
+        } catch (Throwable t2) {
+          // Be defensive here because the Cassandra internals don't guard against listener errors.
+          t2.addSuppressed(t);
+          LOG.warn("Unexpected error while notifying error", t2);
+        }
       }
     }
   }
--- a/grpc/src/main/java/io/stargate/grpc/service/SchemaNotificationsHandler.java
+++ b/grpc/src/main/java/io/stargate/grpc/service/SchemaNotificationsHandler.java
@@ -45,18 +45,27 @@ class SchemaNotificationsHandler implements EventListener {
 
   public void handle() {
     synchronized (responseObserver) {
-      responseObserver.setOnCancelHandler(() -> persistence.unregisterEventListener(this));
-      responseObserver.setOnCloseHandler(() -> persistence.unregisterEventListener(this));
+      responseObserver.setOnCancelHandler(
+          () -> {
+            LOG.info("{} Unregister after cancellation", SchemaNotificationsHandler.this);
+            persistence.unregisterEventListener(this);
+          });
+      responseObserver.setOnCloseHandler(
+          () -> {
+            LOG.info("{} Unregister after error", SchemaNotificationsHandler.this);
+            persistence.unregisterEventListener(this);
+          });
     }
     persistence.registerEventListener(this);
   }
 
   private void onSchemaChange(
       Type type, Target target, String keyspace, String name, List<String> argumentTypes) {
+    LOG.info("{} onSchemaChange({}, {}, {}, {})", this, type, target, keyspace, name);
     synchronized (responseObserver) {
-      if (isFailed) {
-        return;
-      }
+      //      if (isFailed) {
+      //        return;
+      //      }
       try {
         SchemaChange.Builder change =
             SchemaChange.newBuilder().setChangeType(type).setTarget(target).setKeyspace(keyspace);
@@ -69,9 +78,11 @@ class SchemaNotificationsHandler implements EventListener {
         if (type != Type.DROPPED || target != Target.KEYSPACE) {
           notification.setKeyspace(SchemaHandler.buildKeyspaceDescription(keyspace, persistence));
         }
+        LOG.info("{} invoke responseObserver.onNext", this);
         responseObserver.onNext(notification.build());
       } catch (Throwable t) {
         try {
+          LOG.info("{} invoke responseObserver.onError", this, t);
           isFailed = true;
           responseObserver.onError(t);
         } catch (Throwable t2) {
--- a/grpc/src/main/java/io/stargate/grpc/service/SchemaNotificationsHandler.java
+++ b/grpc/src/main/java/io/stargate/grpc/service/SchemaNotificationsHandler.java
@@ -43,11 +43,11 @@ class SchemaNotificationsHandler implements EventListener {
   }
 
   public void handle() {
+    synchronized (responseObserver) {
+      responseObserver.setOnCancelHandler(() -> persistence.unregisterEventListener(this));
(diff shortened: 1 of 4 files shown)


grpc/src/main/java/io/stargate/grpc/service/SchemaNotificationsHandler.java
```diff
@@ -66,7 +66,14 @@ private void onSchemaChange(
       }
       responseObserver.onNext(notification.build());
     } catch (Throwable t) {
-      responseObserver.onError(t);
+      try {
```
tatu-at-datastax, line 69: Should these 2 statements be on separate try-catch blocks, as they could both potentially throw exception?

tatu-at-datastax (approved):

tatu-at-datastax:
Would be nice to have a description of what the underlying issue was and how the fix applies. Not a long description but just a summary.

grpc/src/main/java/io/stargate/grpc/service/SchemaNotificationsHandler.java
```diff
@@ -35,6 +35,7 @@
 
   private final Persistence persistence;
   private final ServerCallStreamObserver<SchemaNotification> responseObserver;
+  private boolean isFailed;
```
olim7t, line 38: Just unregistering in `setOnCloseHandler` did not completely get rid of the duplicate calls, maybe there is a race condition that still allows another event to be processed.

So I had to add this to explicitly check from our code. It doesn't add much complexity since we can piggyback on the existing synchronization.

ivansenic: This looks heavily to the issues that we were seeing with the bi-directional streaming.. Is this a streaming handler? If so you can not call `onError` once you hit the exception.. You rather need to send the exception back as the result of the call.. 

You must call `onError` only once.. This however, kills the client connection and thus you can not continue streaming the responses..

Maybe I am completely wrong as I am jumping here without a context, but if you have duplicate calls for the `onError` it's probably something done wrong.. And I don't think the solution in the PR is actually solving the root issue..

ivansenic (changes requested):
Requesting changes just to block the merge until we figure out the real cause here.. Glad to revert if I am wrong :)

grpc/src/main/java/io/stargate/grpc/service/SchemaNotificationsHandler.java
tomekl007: +1 to @ivansenic comment. We cannot call `onError` twice, according to the contract it can be called only once.
Please take a look how I solved it in:
https://github.com/stargate/stargate/pull/1469 and the wrapper type:
https://github.com/stargate/stargate/pull/1469/files#diff-edefd35d46c51b4410ea5494061770729922d6fa5548f158078c5e4fd37a35f4R356

olim7t: > We cannot call `onError` twice

Agreed, this is what the PR fixes.

> You rather need to send the exception back as the result of the call..

I've considered it but I don't think it's the right approach here. Errors are unexpected here. It's not like bidirectional streaming where an exception can be a legitimate response to one of the queries. If the code throws it's a sign that something is seriously wrong, so I think failing the observer is the right decision, it's up to the client to decide if it wants to resubscribe to notifications.

> And I don't think the solution in the PR is actually solving the root issue..

Fair enough, the fix ensures that extra calls to `onError` are ignored, but I didn't get to the bottom of why they happen in the first place.
What should happen is that after the first error, the callback from `setOnCloseHandler` is executed, we unregister from the persistence and therefore `onSchemaChange` should not happen again. As alluded above I suspect there is a race condition that allows multiple events to go through before we are unregistered.

I'll push a commit with debug logs to try to get more details.

olim7t: The error happens when a keyspace that contained a table gets dropped.
1. We get a notification for the table:
    ```
    INFO  [MigrationStage:1] 2022-01-20 17:22:44,659 onSchemaChange(DROPPED, TABLE, ks1, tbl1)
    ```
2. we try to build the keyspace description to include in the notification. But the keyspace is already gone so this throws, and we notify the observer of the error:
    ```
    INFO  [MigrationStage:1] 2022-01-20 17:22:44,662 invoke responseObserver.onError
    io.grpc.StatusException: NOT_FOUND: Keyspace not found
      at io.grpc.Status.asException(Status.java:543)
      at io.stargate.grpc.service.SchemaHandler.buildKeyspaceDescription(SchemaHandler.java:66)
      at io.stargate.grpc.service.SchemaNotificationsHandler.onSchemaChange(SchemaNotificationsHandler.java:79)
    ```
3. then we get a notification for the keyspace:
    ```
    INFO  [MigrationStage:1] 2022-01-20 17:22:44,663 onSchemaChange(DROPPED, KEYSPACE, ks1, null)
    ```
4. we try to notify the observer, but this fails because it's already complete:
    ```
    INFO  [MigrationStage:1] 2022-01-20 17:22:44,663 invoke responseObserver.onNext
    ...
    INFO  [MigrationStage:1] 2022-01-20 17:22:44,663 invoke responseObserver.onError
    java.lang.IllegalStateException: Stream was terminated by error, no further calls are allowed
    ```
5. meanwhile, the onClose handler gets invoked because of the first error. As I suspected, there is a race condition here, we were able to process the second event before unregistering.
    ```
    INFO  [grpc-default-worker-ELG-3-10] 2022-01-20 17:22:44,663 Unregister after error
    ```

So the root issue is indeed at step 2, I'll fix that.
But because of the race condition I need to keep `isFailed`.

tatu-at-datastax: Very good detective work @olim7t!

Will be interesting to know how the issue can be resolved: perhaps metadata about Keyspace, in context of DROP-Table, would only need Keyspace name? And not actual properties.
(or maybe nothing else would be included but processing will try to construct `Keyspace` object with metadata).

ivansenic: Hmm from client perspective if something is wrong internally on the server, I am not sure if I would like to get that error back.. Yes you say it's client responsibility to reconnect, but from the client perspective I just want to listen on schema changes.. So actually I don't except any errors reported back, except for the "protocol" ones.. Can we re-think this and actually not call `onError` here at all.. Makes no sense from the client point of view..

olim7t: > Hmm from client perspective if something is wrong internally on the server, I am not sure if I would like to get that error back..

That was my initial approach in #1403. But the problem is that the client is now operating on a potentially stale version of the schema, without any indication that something went wrong. Propagating the error gives it a chance to start over from a fresh copy of the schema (see `SchemaCache.Observer.onError`).

tatu-at-datastax:
It sounds like we might need a meeting to discuss approach here: Ivan has expressed valid concerns and this is pretty critical functionality for REST/SGv2.

Now: to me it seems that there must be a way for server to indicate that some problem prevented it from serving a schema change event. There is literally no way to ensure that things could not break at some point (running out of memory; implementation bugs). Such problems can either cause "poison pill" events (trying to re-send fails again and again), or drop updates. If caller relies on 100% accuracy, it is not possible for server to just swallow failures and proceed.
Instead there should in my opinion be SOME mechanism of indicating that server had a hard failure and that the update stream cannot be considered complete. This could be done by closing connection; or for sending specific error notification/event. I don't know what the right way is, but I do not see that just ignoring server side failure in general works.

But if such an error notification is sent (whether explicit or just by closing connection), client would need to bootstrap all information it seems, to be completely safe. If it simply resubscribes, it may lose update notifications that occurred during time connection was closed and reconnection.

I guess there is still an option of trying to ignore the server error and accepting possible missing update(s). If we really decided to go that route, we would at least have to log some clear signal on these occurrences.

Ideally we would make it extremely unlikely that such errors occur, obviously, since these would have major performance impacts (potentially all Coordinator/API connections needing to bootstrap again).

WDYT? Do I completely misunderstand the problem?

ivansenic:
I think we are now here all on the same page.. Thanks @olim7t for pointing out that if information about and error is not propagated down to the client, the client can has a wrong state of the cache.

That's said we need to agree on:
* How is missed event propagated to the client
   * Option 1: `onError` called and client need to explicitly re-connect
   * Option 2: `message` without breaking the connection
* No matter how is client informed, how will that dirty schema state gonna be handled? And if way of handling it on the client favors one of the two above options.

I am not sure what's the best, I leave this to the @olim7t. If the current behavior is already supported in the `SchemaCache.Observer.onError`, then maybe we can go as is. I am removing my request for changes for now.

ivansenic (commented):
Removing the requested changes for now. OK this is not possible.. I am fine with approving once we all agree.

olim7t:
> But if such an error notification is sent (whether explicit or just by closing connection), client would need to bootstrap all information it seems, to be completely safe. If it simply resubscribes, it may lose update notifications that occurred during time connection was closed and reconnection.

Agreed, and that's what `SchemaCache` does: the reconnection process fetches a fresh copy of the schema, also taking care to record and replay any notification that might arrive during those queries (just like the initial call to `newInstance`).

> WDYT? Do I completely misunderstand the problem?

No, that's exactly it :)

olim7t:
> * How is missed event propagated to the client
>   
>   * Option 1: `onError` called and client need to explicitly re-connect
>   * Option 2: `message` without breaking the connection
>
> * No matter how is client informed, how will that dirty schema state gonna be handled? And if way of handling it on the client favors one of the two above options.

I would say option 1 is a bit easier to implement.
The part about "recording notifications during the initial schema fetch" is handled by `InitReplayingStreamObserver`. Currently this is a one-time switch: the observer starts in "recording" mode, then you switch to "ready" (replay recorded events and then pass-through) and it stays that way forever.
If we switch to option 2, we are keeping the same observer for the lifetime of the cache, so we need the ability to turn back to recording mode. That means more moving parts in an already fairly complicated class.

tatu-at-datastax:
Ok, good, agreed it's good to make sure we are on same page & I think that is the case now.
Connection failure with reconnect and re-sync make sense.

I think that on short term this all makes sense, but this also seems like an area where any remaining concerns can have big negative effects: schema out-of-sync can cause rare but nasty, hard-to-diagnose issues.
The other possible concern that I have mentioned earlier is the size of transfers for bootstrapping/resync.

I also think that these concerns can be alleviated by producing Metrics on important events (and relevant metadata): how often is full schema requested, what is payload size; how frequent are regular updates; how many subscribers do coordinators have; how often do connections fail due to server-side/client-side issues.

tatu-at-datastax (approved):

grpc/src/main/java/io/stargate/grpc/service/SchemaNotificationsHandler.java
```diff
@@ -43,33 +47,63 @@
   }
 
   public void handle() {
+    synchronized (responseObserver) {
+      responseObserver.setOnCancelHandler(() -> persistence.unregisterEventListener(this));
+      responseObserver.setOnCloseHandler(() -> persistence.unregisterEventListener(this));
+    }
```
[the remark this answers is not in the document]
olim7t: For non-streaming handlers I think we're fine because observer methods will only be invoked from a single thread.

Whether we synchronize on `this` or `responseObserver` does not matter, as long as it's done consistently. But generally I prefer using a private field, this avoids issues if a caller also happens to synchronize on this object and deadlocks (granted the risk is low here).

PR #1569, opened 2022-01-20
Double-quote UDTs in type strings passed to QueryBuilder

olim7t:
**What this PR does**:
In the REST API, CQL type definitions are passed as raw strings in the JSON payload. This ensures that we properly double-quote UDT names before appending them to the CQL query, so that case-sensitive names are handled correctly.

N/A

**Checklist**
- [x] Changes manually tested


sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/Strings.java
```diff
@@ -15,7 +15,43 @@
...
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
 public class Strings {
```
tatu-at-datastax, line 26: Maybe `CqlStrings` to avoid naming collisions?

tatu-at-datastax (approved):

tatu-at-datastax:
LGTM. If performance ever became an issue there are some things that could be done (like first check if type name only has lower-case characters, if so, return as-is). But code looks efficient enough as-is to avoid processing for non-parameterized type definitions so I'm not worried.

PR #1572, opened 2022-01-21
updated delete document paths to use new write service

ivansenic:
**What this PR does**:
Refactor of the delete document.. Simple and clean, although it was already pretty straight forward.. *New feature*: `profile` query param is available on the delete endpoints as well now. 

Internal.

**Checklist**
- [x] Unit test updates for the `ReactiveDocumentService`
- [ ] Resolve two `TODO`s, @EricBorczuk please advise..
- [x] Delete existing code

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
@@ -231,6 +231,66 @@ public class ReactiveDocumentService {
   }
 
   /**
+   * Deletes a document with given ID in the given namespace and collection.
+   *
+   * @param db {@link DocumentDB} to write in
+   * @param namespace Namespace
+   * @param collection Collection name
+   * @param documentId The ID of the document to delete
+   * @param context Execution content @@return Flag representing if the operation was success.
+   */
+  public Single<Boolean> deleteDocument(
+      DocumentDB db,
+      String namespace,
+      String collection,
+      String documentId,
+      ExecutionContext context) {
+    List<String> subPath = Collections.emptyList();
+    return deleteDocument(db, namespace, collection, documentId, subPath, context);
+  }
+
+  /**
+   * Deletes a document with given ID in the given namespace and collection at the specified
+   * sub-path. Any previously existing sub-document at the given path will be removed.
+   *
+   * @param db {@link DocumentDB} to write in
+   * @param namespace Namespace
+   * @param collection Collection name
+   * @param documentId The ID of the document to delete
+   * @param subPath Sub-path of the document to delete. If empty will delete the whole doc.
+   * @param context Execution content
+   * @return Flag representing if the operation was success.
+   */
+  public Single<Boolean> deleteDocument(
+      DocumentDB db,
+      String namespace,
+      String collection,
+      String documentId,
+      List<String> subPath,
+      ExecutionContext context) {
+
+    return Single.defer(
+            () -> {
+              // authentication for writing before anything
+              authorizeWrite(db, namespace, collection, Scope.DELETE);
+
+              // pre-process to support array elements
+              List<String> subPathProcessed = processSubDocumentPath(subPath);
+
+              // call write document
+              return writeService.deleteDocument(
+                  db.getQueryExecutor().getDataStore(),
+                  namespace,
+                  collection,
+                  documentId,
+                  subPathProcessed,
+                  context);
+            })
+        // TODO should we extract applied here and return that
+        .map(any -> true);
+  }
+
+  /**
    * Searches for documents in the whole collection.
    *
    * @param db {@link DocumentDB} to search in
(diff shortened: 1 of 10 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/write/DocumentWriteService.java
```diff
@@ -237,6 +237,71 @@ public DocumentWriteService(TimeSource timeSource, DocsApiConfiguration config)
...
+      String documentId,
+      List<String> subDocumentPath,
+      ExecutionContext context) {
+    // create and cache the remove query prepare
+    Single<? extends Query<? extends BoundQuery>> deleteQueryPrepare =
+        prepareDeleteDocumentQuery(dataStore, keyspace, collection, subDocumentPath);
+
+    // execute when prepared
+    return deleteQueryPrepare
+        .observeOn(Schedulers.computation())
```
mpenick, line 287: I'm curious why we want this operation to run on a "compuation" thread pool? Why not run it all on the "io" thread pool? Wouldn't this result in migration from one type to the other type of thread?

ivansenic: It's to do binding op off the TPC thread (DSE) or Native Transport Thread (C*).. Yes it will offload to another thread.. We confirmed in the perf tests and JMH tests that there is not much of a difference, if any.. However, if we don't go away from those "storage" threads then they will be doing any task that is in the followup of this reactive sequence.. Meaning even logging the output of the request or writing to the request buffer..

ivansenic:
@EricBorczuk any chance you can review so I can merge and go forward?

restapi/src/main/java/io/stargate/web/docsapi/service/write/DocumentWriteService.java
```diff
@@ -237,6 +237,71 @@ public DocumentWriteService(TimeSource timeSource, DocsApiConfiguration config)
...
+        .map(
+            prepared -> {
+              // take current timestamp for binding
+              long timestamp = timeSource.currentTimeMicros();
+
+              // return bind delete
+              // TODO Eric I feel keeping deletes always with the timestamp - 1 feels right
```
EricBorczuk, line 294: Yes this is alright - I know that when writing chunks of documents there is a delete statement in the batch that has `timestamp - 1` as its value, so to keep it consistent this should be fine

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -230,6 +230,66 @@ public ReactiveDocumentService(
...
+              // call write document
+              return writeService.deleteDocument(
+                  db.getQueryExecutor().getDataStore(),
+                  namespace,
+                  collection,
+                  documentId,
+                  subPathProcessed,
+                  context);
+            })
+        // TODO should we extract applied here and return that
```
EricBorczuk, line 289: Does applied actually return whether data was modified? I had assumed that running a delete that matches no rows would still return `applied` as `true`

If that's not the case, then we could actually differentiate between "the deletion occurred on existing doc" (204 no content) and "the deletion didn't occur because there was nothing there" (404 not found)

restapi/src/main/java/io/stargate/web/docsapi/service/write/DocumentWriteService.java
EricBorczuk: So this is a way to avoid blocking?

EricBorczuk (approved):
Looks good! I addressed your two TODO statements - hope that helps :)

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
ivansenic: I am actually not sure, however seems that this can be controlled with the `IF EXISTS`: https://docs.datastax.com/en/cql-oss/3.3/cql/cql_reference/cqlDelete.html? What do you say, add this to the deletion and correctly go for the different HTTP code? Seems cool? Not sure if this will be considered as API breaking?

restapi/src/main/java/io/stargate/web/docsapi/service/write/DocumentWriteService.java
ivansenic: It;s to avoid any extra work being done on the TCP thread and a safety net of not invoking new data store calls on this thread, which has caused issues in the past..

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
tatu-at-datastax: Not sure if it matters but for idempotency perspective returning success in both cases is usually considered better I think: success means that Table no longer exists and so calling DELETE (for timeouts etc) multiple times is safe.
404 can be overloaded as well since it can signal bad URL amongst other things.

But this is a good question as it applies to other APIs as well: ideally we should follow the same guideline.

ivansenic: OK, I will leave the TODO for now and keep this things as is.. We can change in the V2 if we want..

PR #1574, opened 2022-01-21
add separate release workflow for v2

jeffreyscarpenter:
Setting up a separate workflow to build v2 releases since the current release.yml workflow is set up to work off of `master`. The v2 release builds from the `v2.0.0` branch and is triggered by tags matching `v2.*.*`. The original release workflow is modified to trigger off of `v1.*.*` tags instead of `v*.*.*`.

This workflow will be extended in a followup PR to build Docker images, which will be an a difference from v1 which builds Docker images via the separate `docker-images` repo.

### DIFF
--- a/.github/workflows/release-v2.yml
+++ b/.github/workflows/release-v2.yml
@@ -3,7 +3,7 @@ name: Release
 on:
   push:
     tags:
-      - 'v1.*.*'
+      - 'v2.*.*'
 
 jobs:
   create-release:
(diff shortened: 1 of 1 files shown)


tatu-at-datastax:
Ok so this is initially just a copy of the default one, which will now become "v1-only" release workflow? And in future this one will be improved to do docker image build/publish too? Makes sense.

tatu-at-datastax (approved):

ivansenic (approved):
Seems fine..

jeffreyscarpenter:
> I guess `CHANGELOG.md` will not be updated when we release V2, or?

We will want to list the latest v1 and v2 changes that are incorporated. I am investigating the Changelog generation tool we've been using to see what output it produces and whether any manual editing will be required. I envision incorporating any needed changes in a separate PR.

PR #1579, opened 2022-01-24
Extract gRPC bridge as a separate bundle

olim7t:
**What this PR does**:
Extract an "internal" gRPC service that will serve as the persistence backend for service implementations. Revert the original gRPC API mostly to its v1 state.

Where to start: new `bridge` module. This reuses a a lot of code from `grpc`, I went with a direct dependency for simplicity we could also extract a common utility module. The new service definition is directly in `grpc-proto`.

I didn't keep any of the new v2 definitions on the original public API, but it's something that can be done if we see some value there.

N/A

**Checklist**
- [x] Changes manually tested


bridge/src/test/java/io/stargate/bridge/service/BaseBridgeTest.java
```diff
@@ -0,0 +1,97 @@
...
+        server.shutdown().awaitTermination();
+      }
+      if (executor != null) {
+        executor.shutdownNow();
+      }
+      if (clientChannel != null) {
+        clientChannel.shutdown().awaitTermination(60, TimeUnit.SECONDS);
+      }
+    } catch (InterruptedException e) {
+      e.printStackTrace();
```
tatu-at-datastax, line 67: Not sure it makes sense to print this? (maybe just left by IDE auto-complete)

olim7t: That was a mindless copy-paste from `BaseGrpcServiceTest`. I don't think it matters a lot, nobody is interrrupting the test threads. But we can rethrow instead.

tatu-at-datastax (approved):

PR #1584, opened 2022-01-25
Update release-v2.yml

jeffreyscarpenter:
Rename the workflow so it won't be confused with v1 release workflow

### DIFF
--- a/.github/workflows/release-v2.yml
+++ b/.github/workflows/release-v2.yml
@@ -144,7 +144,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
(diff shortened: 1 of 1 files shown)


tatu-at-datastax (approved):
Good idea. Might want to add "v1" for the default one unless it has it already?

jeffreyscarpenter:
> Good idea. Might want to add "v1" for the default one unless it has it already?

I think it's better to leave the v1 release as-is.

tatu-at-datastax:
> > Good idea. Might want to add "v1" for the default one unless it has it already?
> 
> I think it's better to leave the v1 release as-is.

I was only thinking of adding it in message, not label. But not a big deal either way.

PR #1585, opened 2022-01-26
use doFinally in the `DseConnection#executeRequest` to reset warnings

ivansenic:
**What this PR does**:
The ` ClientWarn.instance.resetWarnings();` is not called if the reactive sequence ends up in error. The error handler does call the `convertExceptionWithWarnings` which calls the `ClientWarn.instance.getAndClearWarnings()`, however, although name suggest different the `getAndClearWarnings()` is not clearing the thread local.

Proposed fix ensures that after either `onSuccess` or `onError` handlers are called, we reset the state.

**I would be very thankful if anybody has an idea about properly testing this?**

No ticket.

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
+++ b/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
@@ -535,7 +535,6 @@ public class DsePersistence
                 .execute(queryState, queryStartNanoTime)
                 .map(
                     response -> {
-                      try {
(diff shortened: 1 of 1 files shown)


tatu-at-datastax:
Sounds good. Might actually fix a problem with mismatched warnings (another request that runs on a thread where clearing failed). Will not do anything for the Map retention problem but that is something DSE side needs to address.

tatu-at-datastax (approved):

mpenick (approved):

PR #1590, opened 2022-01-27
Add tag/publish options to docker build script for v2

jeffreyscarpenter:
This adds flags to the `build_docker_images.sh` script:

- `-p` - causes the images to be built for all supported platform architectures and pushed to Docker Hub (assumes you are logged in). This is intended to be used as part of automated builds (see #1584).
- `-t <version>` - overrides the default tag that will be applied to the image with the one you provide. By default the tag consists of the version is obtained from the parent `pom.xml` file, prepended with `v`, i.e. `v2.0.0`.

### DIFF
--- a/build_docker_images.sh
+++ b/build_docker_images.sh
@@ -5,8 +5,40 @@ set -euo pipefail
 # Assumes that you have done a complete build so that all jars have been created, i.e.:
 #   ./mvnw clean install -P dse -DskipTests=true
 
+#
+# Defaults
+#
+
+# generate for local platform, don't push
+DOCKER_FLAGS=
+
 # extract Stargate version from project pom file
-export SGTAG=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+SGTAG="v$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)"
+
+#
+# overrides via command line
+#
+
+while getopts ":pt:" opt; do
+  case $opt in
+    p)
+      DOCKER_FLAGS="--platform linux/amd64,linux/arm64 --push"
+      ;;
+    t)
+      SGTAG=$OPTARG
+      ;;
+    \?)
+      echo "Invalid option: -$OPTARG" >&2
+      exit 1
+      ;;
+  esac
+done
+
+echo "Building version $SGTAG"
+
+#
+# Persistence images
+#
 
 # Create a temp directory under current directory to use as a staging area for image creation
 # This is a workaround since Docker COPY command provides no way to exclude files.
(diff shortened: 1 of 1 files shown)


build_docker_images.sh
```diff
@@ -5,7 +5,48 @@ set -euo pipefail
...
+
+#
+# Defaults
+#
+
+# generate for local platform, don't push
+DOCKER_FLAGS=
+
+# extract Stargate version from project pom file
+SGTAG="v$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)"
```
tatu-at-datastax, line 24: This looks useful, something I could borrow :)

(Maven's version extractor is pretty good in general, although in this case it just prints the whole version so  bo parsing/extraction needed).

build_docker_images.sh
```diff
@@ -16,13 +57,15 @@ mkdir ${LIBDIR}
 cp ./stargate-lib/* $LIBDIR
 rm ${LIBDIR}/persistence*.jar
 
-#ls ${LIBDIR}
-
-docker build --target coordinator-4_0 --build-arg LIBDIR="$LIBDIR" -t stargateio/coordinator-4_0:$SGTAG .
-docker build --target coordinator-3_11 --build-arg LIBDIR="$LIBDIR" -t stargateio/coordinator-3_11:$SGTAG .
-docker build --target coordinator-dse-68 --build-arg LIBDIR="$LIBDIR" -t stargateio/coordinator-dse-68:$SGTAG .
+docker buildx build --target coordinator-4_0 --build-arg LIBDIR="$LIBDIR" -t stargateio/coordinator-4_0:$SGTAG $DOCKER_FLAGS .
```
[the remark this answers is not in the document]
jeffreyscarpenter: Specifying `buildx` causes Docker to use a different build engine from the default. You can read more about the differences in general in articles like [this](https://www.cloudsavvyit.com/12441/what-is-dockers-buildkit-and-why-does-it-matter/). 

But for our purposes here the most important thing is the ability to build images for multiple platforms. Note that the scripts in the `docker-images` repo (which this will eventually replace) uses `buildx`. See an example [here](https://github.com/stargate/docker-images/blob/master/cassandra-4.0/build.sh#L16).

build_docker_images.sh
jeffreyscarpenter: Just to clarify, are you suggesting there is a better way to do this?

tatu-at-datastax: Sorry, no, not at all. I was just remembering reading about logic Maven tooling extracts components of all kinds of exotic version Strings, and does it well & has documentation outlining how it works.
Which is really not relevant here, just a detour :)

Issue #1577, opened 2022-01-24
Invalid UUID (0 bytes) served by gRPC for SGv2/REST API


tatu-at-datastax:
Our perf tests for StargateV2/REST seem to have odd failures reported when accessing rows inserted: stack trace:

```
ERROR [dw-176 - GET /v2/keyspaces/baselines/iot?where=%7B%22machine_id%22%3A%7B%22%24eq%22%3A%2212d1012c-47fa-441c-a41e-87
3c048d2535%22%7D%2C%22sensor_name%22%3A%7B%22%24eq%22%3A%22wire_gauge%22%7D%7D&page-size=10] 2022-01-24 11:53:00,025 Defau
ltExceptionMapper.java:59 - Error handling a request: f9a67463d3518a01
java.lang.IllegalArgumentException: Expected 16 bytes for a uuid values, got 0
        at io.stargate.grpc.Values.uuid(Values.java:258)
        at io.stargate.sgv2.restsvc.grpc.FromProtoValueCodecs$UUIDCodec.fromProtoValue(FromProtoValueCodecs.java:368)
        at io.stargate.sgv2.restsvc.grpc.FromProtoConverter.mapFromProtoValues(FromProtoConverter.java:41)
        at io.stargate.sgv2.restsvc.resources.ResourceBase.convertRows(ResourceBase.java:157)
        at io.stargate.sgv2.restsvc.resources.ResourceBase.fetchRows(ResourceBase.java:146)
        at io.stargate.sgv2.restsvc.resources.Sgv2RowsResourceImpl.getRowWithWhere(Sgv2RowsResourceImpl.java:114)
```

which suggests that somehow either entries are stored incorrectly (corrupt data), or, perhaps more likely, that gRPC API is changing missing/null value into incorrect representation.

I have not been able to reproduce the issue with regular POSTs via REST API; data in this case is generated by Fallout tests, but not sure if it's via straight CQL or REST (sgv2) API.

Tests are here:

https://fallout.sjc.dsinternal.org/tests/ui/jenkins-dse@datastax.com/sgv2-http-rest-timeseries-cass4

EricBorczuk:
[...]

PR #1591, opened 2022-01-27
Add a test case to reproduce #1577 fail on v2.0.0

tatu-at-datastax:
**What this PR does**:

Reproduces issue #1577

The key requirements are:

* Partition key has `UUID` value
* At least 2 entries are read; of these partition key for first is fine, second fails due to "empty" (by either gRPC server noticing, or sending malformed key)

Root cause is within gRPC service handlers but this reproduces it via REST/SGv2 since that is where it was first observed (as part of Fallout/Haxx perf test).

Won't fix anything yet.

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
+++ b/testing/src/main/java/io/stargate/it/http/RestApiv2Test.java
@@ -1654,26 +1654,47 @@ public class RestApiv2Test extends BaseIntegrationTest {
             Arrays.asList("id " + mainKey, "id2 a", "name Bob"),
             Arrays.asList("id " + mainKey, "id2 b", "name Joe"),
             Arrays.asList("id " + mainKey, "id2 x", "name Patrick"),
-            Arrays.asList("id " + secondKey, "id2 c", "name Rick"),
             Arrays.asList("id " + secondKey, "id2 e", "name Alice")));
 
-    // get first page
+    // get first page; cannot use "raw" mode as we need pagingState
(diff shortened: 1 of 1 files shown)


tatu-at-datastax:
Ready for review, although we may want to wait for fix to #1577 which will go through StargateV1 (master branch).

tatu-at-datastax:
One new test failure; we could check this in later as a regression test, or, could mark as `@Disabled`, change to enabled later on. Not sure what makes most sense.

jeffreyscarpenter:
I don't have a major objection to this, although technically the underlying issue is a bug in the gRPC service. So why not add a test case that focuses on gRPC service rather than on REST service? Also, REST service is moving to the bridge. I guess adding it is not going to hurt anything, except perhaps continuing to add complexity to the IT test suite. If you do add now, agree adding `@Disabled` until #1577 fix is pushed down is a good choice.

tatu-at-datastax:
@jeffreyscarpenter Fair question, and this is due to "backwards" reproductions: test was added before the root cause was known. There are now 2 other tests for gRPC directly (unit test, integration test) that do specifically check for the root cause.

I think I'd like to add this in addition only because of existing shallow coverage of some of the functionality -- we do not cover combinations of features. But I think I should probably extend this test slightly, to do actual pagination.

jeffreyscarpenter:
> I think I'd like to add this in addition only because of existing shallow coverage of some of the functionality -- we do not cover combinations of features. But I think I should probably extend this test slightly, to do actual pagination.

Good point, agree.

PR #1593, opened 2022-01-27
Add validation for #1592; also add a test for UUID case (failing)

### LINKED ISSUE #1592: More gRPC `ValueCodec` implementations should verify validity of data decoded (ByteBuffer->proto Value)
(note: inspired by #1577 )

Looks like following `ValueCodec`s only enforce size limitations on encoding, but not decoding:

* `DecimalCodec` (actually has some checking, might be ok)
* `InetCodec`
* `VarintCodec` (also has some checking)

(validation was added to `UuidCodec` and is helping pinpoint apparent conversion issue/corruption with UUID values)

It seems useful to add such checks as they are very low cost and can prevent corrupt data from being served downstream: this becomes particularly important with StargateV2 when gRPC bridge will becomes the universal interface.

*EDIT*: Actually codecs for `Decimal` and `VarInt` have some checking, and seem to coerce 0-length data into `null`.
There is no comment explaining this behavior: looks like bit of defensive programming. But what is not clear is whether this is something that SHOULD or SHOULD NOT be done -- it could help or it might mask an actual problem?

tatu-at-datastax:

Fixes #1592 (add more validation) but also related to #1577

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/grpc/src/main/java/io/stargate/grpc/codec/DecimalCodec.java
+++ b/grpc/src/main/java/io/stargate/grpc/codec/DecimalCodec.java
@@ -43,9 +43,7 @@ public class DecimalCodec implements ValueCodec {
 
   @Override
   public Value decode(@NonNull ByteBuffer bytes, @NonNull ColumnType type) {
-    if (bytes.remaining() == 0) {
-      return null;
-    } else if (bytes.remaining() < 4) {
+    if (bytes.remaining() < 4) {
       throw new IllegalArgumentException(
           "Invalid decimal value, expecting at least 4 bytes but got " + bytes.remaining());
     }
--- a/grpc/src/main/java/io/stargate/grpc/codec/VarintCodec.java
+++ b/grpc/src/main/java/io/stargate/grpc/codec/VarintCodec.java
@@ -37,7 +37,7 @@ public class VarintCodec implements ValueCodec {
   public Value decode(@NonNull ByteBuffer bytes, @NonNull ColumnType type) {
     if (bytes.remaining() == 0) {
       throw new IllegalArgumentException(
-          "Invalid decimal value, expecting non-empty Bytes but got 0");
+          "Invalid varint value, expecting non-empty Bytes but got 0");
     }
 
     return Value.newBuilder()
--- a/grpc/src/main/java/io/stargate/grpc/codec/VarintCodec.java
+++ b/grpc/src/main/java/io/stargate/grpc/codec/VarintCodec.java
@@ -36,7 +36,8 @@ public class VarintCodec implements ValueCodec {
   @Override
   public Value decode(@NonNull ByteBuffer bytes, @NonNull ColumnType type) {
     if (bytes.remaining() == 0) {
-      return null;
+      throw new IllegalArgumentException(
+              "Invalid decimal value, expecting non-empty Bytes but got 0");
     }
 
     return Value.newBuilder()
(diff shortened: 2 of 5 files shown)


grpc/src/main/java/io/stargate/grpc/codec/DecimalCodec.java
```diff
@@ -43,9 +43,7 @@ public ByteBuffer encode(@NonNull Value value, @NonNull ColumnType type) {
 
   @Override
   public Value decode(@NonNull ByteBuffer bytes, @NonNull ColumnType type) {
-    if (bytes.remaining() == 0) {
-      return null;
```
tatu-at-datastax, line 47: This could not have worked: protobuf/grpc code will throw NPE upon getting `null`. Null handling is done by caller, before this method gets called.

grpc/src/main/java/io/stargate/grpc/codec/VarintCodec.java
```diff
@@ -36,7 +36,8 @@ public ByteBuffer encode(@NonNull Value value, @NonNull ColumnType type) {
   @Override
   public Value decode(@NonNull ByteBuffer bytes, @NonNull ColumnType type) {
     if (bytes.remaining() == 0) {
-      return null;
```
tatu-at-datastax, line 39: This could not have worked: protobuf/grpc code will throw NPE upon getting `null`. Null handling is done by caller, before this method gets called.

tatu-at-datastax:
@EricBorczuk @mpenick Ready for review, I think -- adds checks in codecs, a new failing test (that is Disabled to pass CI).

May want to add tests for `InetCodec`, `DecimalCodec` and `VarintCodec` as the next step (along with possible fixes as needed).

Would also be good to have this tested at unit test level but I don't yet know things well enough to try that.

testing/src/main/java/io/stargate/it/grpc/ExecuteQueryTest.java
```diff
@@ -184,6 +188,32 @@ public void simpleQueryWithPaging(@TestKeyspace CqlIdentifier keyspace)
     assertThat(rs.getPagingState()).isNotNull();
   }
 
+  @Test
+  @Disabled("Until issue #1577 fixed")
```
tatu-at-datastax, line 192: Could separate into different PR (one with the fix), as an alternative. But thought this is easier way to share.

mpenick (approved):

EricBorczuk (approved):
This all looks good! Asked one question about the test for UUID

testing/src/main/java/io/stargate/it/grpc/ExecuteQueryTest.java
```diff
@@ -184,6 +188,32 @@ public void simpleQueryWithPaging(@TestKeyspace CqlIdentifier keyspace)
...
+  public void simpleQueryWithPagingUuidKey(@TestKeyspace CqlIdentifier keyspace) {
+    StargateBlockingStub stub = stubWithCallCredentials();
+    final String uuid = UUID.randomUUID().toString();
+
+    for (int i = 0; i < 5; i++) {
+      stub.executeQuery(
+          cqlQuery(
+              String.format("INSERT INTO test_uuid (id, value) VALUES (%s, %d)", uuid, i),
```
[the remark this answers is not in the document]
tatu-at-datastax: Might be a good idea if it's easy to generalize? I do have a PR to fix the actual issue, along with unit test so will think about this one after getting that in.

PR #1601, opened 2022-01-31
Start refactoring RESTv2 Schema CRUD tests into separate test class

### LINKED ISSUE #1600: Split REST/Schema tests out of (too) big `RestApiv2Test` class
Currently we have a single huge test class, `RestApiv2Test` covering the whole REST v2 API: it is above 3600 lines of code.
That is bit unwieldy to work on.

So it seems reasonable to carve out some tests into at least one more test class: we can start by creating another class that contains Schema CRUD methods. We may consider more granular break-up in future if that makes sense.
One consideration on short term is that of StargateV1/StargateV2 refactoring: splitting should be done in a way to keep it easy to merge SGv1->SGv2 changes. This should be easy enough to achieve, just needs to be considered when doing initial split.

tatu-at-datastax:
**What this PR does**:

As per issue title, start of refactoring to reduce unwieldy size of existing `RestApiv2Test` class.
Also includes minor cleanup within tests, when that makes sense.

Fixes #1600

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/testing/src/main/java/io/stargate/it/http/RestApiv2SchemaTest.java
+++ b/testing/src/main/java/io/stargate/it/http/RestApiv2SchemaTest.java
@@ -15,7 +15,6 @@ import io.stargate.web.restapi.models.RESTResponseWrapper;
 import java.io.IOException;
 import java.lang.reflect.Method;
 import java.util.Optional;
-
 import net.jcip.annotations.NotThreadSafe;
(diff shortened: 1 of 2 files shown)


ivansenic (commented):
Awesome.. We need this, tests classes are huge and are not readble atm..

jeffreyscarpenter (approved):

PR #1602, opened 2022-01-31
Improve Docker-compose scripts

jeffreyscarpenter:
This PR improves the docker-compose scripts for Stargate v2 by defaulting the Stargate version used to the one specified in the parent `pom.xml` file. This can be overridden by the `-t` (for "tag") flag on the `start_*.sh` scripts. Also improved `README` files. 

**Checklist**
- [x] Changes manually tested


docker-compose/cassandra-3.11/start_cass_311.sh
```diff
@@ -1,5 +1,25 @@
 #!/bin/sh
 
+# Set variable SGTAG
+
+SGTAG="v$(../../mvnw -f ../.. help:evaluate -Dexpression=project.version -q -DforceStdout)"
```
ivansenic, line 5: :partying_face:

docker-compose/cassandra-3.11/docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
     networks:
       - stargate
     ports:
-      - 9044:9042
+      - 9042:9042
```
ivansenic, line 9: there was a discussion around this in [[stargate/docker]()](https://github.com/stargate/docker-images/pull/95#discussion_r795954868).. now when I think about this: 
* `9042` (default) should be mapped to Stargate.. 
* and I am not sure if we need to expose the CQL port for C, if yes, then let's keep it `9044`? I mean we should want all CQL traffic to go through Stargate or?

docker-compose/cassandra-4.0/docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
     networks:
       - stargate
     ports:
-      - 9044:9042
+      - 9042:9042
```
ivansenic, line 9: same as above?

docker-compose/dse-6.8/docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
     networks:
       - stargate
     ports:
-      - 9044:9042
+      - 9042:9042
```
ivansenic, line 9: same

ivansenic (approved):
Looks good, however I had a problem creating images once again:

```
$ ./build_docker_images.sh 
Building version v2.0.0-ALPHA-2-SNAPSHOT
cp: -r not specified; omitting directory './stargate-lib/felix-cache'
cp: -r not specified; omitting directory './stargate-lib/log'
```
So these two directories are created on the run time.. So they should not be included in the docker images, that's fine.. However, creating images fails.. Should we add:

```
cp ./stargate-lib/*.jar $LIBDIR
cp ./stargate-lib/logback.xml $LIBDIR
````

So we only copy jars and logback config.. Of course this *is not* happening when you run `clean install` previously.
```

docker-compose/cassandra-3.11/docker-compose.yml
jeffreyscarpenter: yeah, I kind of kept this convention from the original examples in `docker-images`. I assumed that exposing CQL port on one of the Cassandra nodes was for debugging purposes, but I agree with your comment. I'm not sure that this is a practice that we want to recommend.  Will remove this.

jeffreyscarpenter:
> So we only copy jars and logback config.. Of course this _is not_ happening when you run `clean install` previously.

I like your proposed fix and will incorporate.

PR #1605, opened 2022-02-01
refactor patching of the documents

### LINKED ISSUE #1604: Update documents checks for schema also on document update and patch
Update of the documents is checking for the JSON  schema correctness also when only a sub-document is updated. This is wrong as schema relates to a complete document and not only the portion of it. 

Check is also executed when patching, which is completely wrong, the schema validation would only validated on the patching data. Should be disabled for patching for now.

Two options for update:
* Make sure schema is checked only if the whole document is updated
* Don't do any schema checks on update

We actually need to understand if this is a missing feature and if all operations have to be checked against the schema.

ivansenic:
**What this PR does**:
Refactoring of the document patching to use the new shredding and service. Updates from the existing version:
* Patching of a sub-document with a primitive is now possible
* Patching is not allowed if JSON schema exists on a collection (see #1604 for more info, fixed root patching)

Unfortunately I needed to refactor a bit around delete document query builders, thus a bit more code changes than expected.

Internal issue, but fixes #1604.

**Checklist**
- [x] Changes manually tested
- [x] Unit test for write service
- [x] Unit test for the document service
- [x] Delete old unused code 
- [x] Solve patch JSON schema validation
- [x] Resolve `TODO`, @EricBorczuk please help


ivansenic:
@polandll Here also I changed the notes in the Swagger annotations..

ivansenic:
@EricBorczuk From my point of view this is finished.. Please review and let's resolve that one `TODO`.. Would like to merge asap so it goes into next Stargate version.

restapi/src/main/java/io/stargate/web/docsapi/resources/ReactiveDocumentResourceV2.java
```diff
@@ -283,6 +284,132 @@ public void putDocPath(
...
+          @NotBlank(message = "payload must not be empty")
+          String payload,
+      @ApiParam(
+              value = "Whether to include profiling information in the response (advanced)",
+              defaultValue = "false")
+          @QueryParam("profile")
+          Boolean profile,
+      @Context HttpServletRequest request,
+      @Suspended AsyncResponse asyncResponse) {
+    // TODO Eric should we actually fail the PATCH if the table is not there?
```
EricBorczuk, line 382: Hmm...maybe - I think the reason why we eagerly create the table is because PUT does so as well, but we could also say that PATCH requires that there already is data to patch onto...that probably makes more sense

restapi/src/main/java/io/stargate/web/docsapi/service/write/DocumentWriteService.java
```diff
@@ -237,15 +233,158 @@ public DocumentWriteService(TimeSource timeSource, DocsApiConfiguration config)
...
+        prepareDeleteDocumentQuery(deleteArrayBuilder, dataStore, keyspace, collection)
+            .observeOn(Schedulers.computation());
+
+    // create and cache the insert query prepare
+    Single<? extends Query<? extends BoundQuery>> insertPrepared =
+        prepareInsertDocumentRowQuery(dataStore, keyspace, collection)
+            .observeOn(Schedulers.computation());
+
+    // execute when both done
+    return Single.zip(
```
EricBorczuk, line 335: :+1:

EricBorczuk (approved):
This looks good! You could make patching illegal on nonexistent tables; that might help differentiate it from PUT in the user's mind too

restapi/src/main/java/io/stargate/web/docsapi/resources/ReactiveDocumentResourceV2.java
ivansenic: So I decided that I don't want to change current behavior, so that I don't break anybody.. Feel more comfortable keeping the existing behavior..

PR #1606, opened 2022-02-01
Fix release-v2 workflow

jeffreyscarpenter:
The change pushed in #1584 inadvertently deleted a line from the release-v2 workflow which causes a parsing error every time Google cloud build tries to read the file, which is on every push. This PR is to add the line back in.


ivansenic (approved):
Nice catch, was not able to find it myself..

PR #1609, opened 2022-02-02
relates to #1604: fix update of sub-document with JSON schema existing

ivansenic:
**What this PR does**:
Fixes a bug introduced in the #1572 which resulted in checking the JSON schema being valid on sub-document updates. This action should be forbidden. Note that the similar bug exists for the `PATCH` operation on a root document and that will be fixed in #1605.

@polandll I updated the Swagger annotations here to clarify this for the user.

Relates to #1604 and fixes the update (`PUT`) operations.

**Checklist**
- [x] Changes manually tested


tomekl007 (approved):
LGTM

PR #1615, opened 2022-02-03
fixed merge error that changed expected result string

jeffreyscarpenter:
fix test broken by merge error


tatu-at-datastax (approved):

tatu-at-datastax:
@jeffreyscarpenter I think this must be changed to use regexp. Cassandra 4.0 gives:

```
"Bad request: A user type with name 'udt1' already exists"
```

vs 3.11 (against which local tests run) 

```
Bad request: A user type of name KEYSPACE.TABLENAME already exists
```

so I'd probably only look for "udt1" and "already exists" as either chained ".contains()" or regexp.

PR #1616, opened 2022-02-03
Refactor RestApiExtension to support integration testing of other API services

jeffreyscarpenter:
The `RestApiExtension` was added to SGv2 to support integration testing in which the REST API Service is deployed in its own JVM, separate from the coordinator node. This PR refactors the extension, renaming it to `ApiServiceExtension` and abstracting out details specific to the REST API so that it can be reusable for testing other services. 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/DEV_GUIDE.md
+++ b/DEV_GUIDE.md
@@ -221,13 +221,13 @@ have done a recent build, for example:
 Then you can run the individual test using the `-Dit.test` option, for example:
 
 ```sh
-mvn -pl testing -Pit-cassandra-3.11 verify -Dit.test=RestApiv2RowsTest
+mvn -pl testing -Pit-cassandra-3.11 verify -Dit.test=RestApiv2Test
 ```
 
 You can even run a single case (method):
 
 ```sh
-mvn -pl testing -Pit-cassandra-3.11 verify -Dit.test=RestApiv2RowsTest#getRowsWithQuery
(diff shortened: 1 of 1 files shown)


testing/src/main/java/io/stargate/it/http/ApiServiceParameters.java
```diff
@@ -0,0 +1,137 @@
...
+  @Value.Default
+  // Example: 8088
+  default int servicePort() {
+    return 0;
+  }
+
+  @Value.Default
+  // Example: dw.server.connector.port
+  default String servicePortPropertyName() {
+    return "Override me";
```
tatu-at-datastax, line 62: If this is truly required, could/should it be left with no implementation at all? (or maybe Immutables requires this?)

tatu-at-datastax: I guess this was just refactored from the existing class with no changes.

tatu-at-datastax (approved):

testing/src/main/java/io/stargate/it/http/ApiServiceParameters.java
jeffreyscarpenter: Good point. I found a better way to do this with `@Value.Parameter` which doesn't require a default value.

PR #1617, opened 2022-02-04
refactor batch write of documents

### LINKED ISSUE #1610: Documents not checked for valid JSON schema on batch write
JSON schema validation is bypassed completely on batch write.

ivansenic:
*Work in progress, but ready to review.*

**What this PR does**:
Refactors the batch write of the documents. Final PR of the doc write path refactoring. This PR solves the following bugs in addition:
* adds JSON schema validation for each document inserted if one is present
* recognizes the duplicate IDs in a batch of documents and rejects with `400`, until now only one of the documents sharing the same ID was written
* tries to align with the API docs saying that the `MultiDocsResponse#documentIds` should return `The ids of the documents created, in order of creation` (not finalized yet)

Fixes #1610 and internal ticket.

**Implementation discussions**
This implementation changed the original one, where all queries needed for all the documents to be written were part of a single batch statement. This statement was fired and could be success or not. This would result in the multiple partition batch, which if any query fails, makes sure that none of the DML statement are executed. So it's all or nothing.

This implementation actually fires a batch per document in-parallel, making each batch statement a single partition batch, thus more effective. However, this would mean that if one document write fails, other will succeed. My idea would be that the result of the HTTP call sends back the IDs of documents that were successfully written. Write errors are effectively ignored.

I would like to hear opinions about this change, is this something that improves stuff and would be acceptable as the API change (although we would fulfill the contract with returning the IDs of the written documents).

@mpenick @EricBorczuk Let's discuss please.

**Checklist**
- [x] Changes manually tested
- [x] INT test for duplicated IDs
- [x] INT test for batch schema check
- [x] Unit test for the `ReactiveDocumentService`
- [x] Finalize implementation discussions
- [x] Error handling of a single failed write
- [x] Remove unused code


jeffreyscarpenter:
Looks like tests are failing with code format error?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -154,6 +159,101 @@ public ReactiveDocumentService(
...
+                }
+              }
+
+              // generate the Single for each document write
+              List<Single<ResultSet>> writeAll =
+                  documentRowsMap.entrySet().stream()
+                      .map(
+                          entry -> {
+                            if (useUpdate) {
+                              return writeService.updateDocument(
```
EricBorczuk, line 230: So the difference between `updateDocument` and `writeDocument` is that updating does a deletion of certain paths before write?

restapi/src/main/java/io/stargate/web/docsapi/service/util/DocsApiUtils.java
```diff
@@ -40,12 +41,27 @@
...
   private DocsApiUtils() {}
 
+  public static Optional<JsonPointer> pathToJsonPointer(String path) {
+    if (null == path) {
+      return Optional.empty();
+    }
+
+    String separator = String.valueOf(JsonPointer.SEPARATOR);
+    String adaptedPath =
+        path.replaceAll(DocsApiUtils.PERIOD_PATTERN.pattern(), separator)
```
EricBorczuk, line 58: :+1:

EricBorczuk (commented):
The changes look good direction-wise, I'm happy to see that the resource/endpoint files are getting consolidated. The actual implementation also looks good wrt batching. I guess the final issue would be how we handle partial failures in the batch write process - does it currently return only the ID's that successfully got written? Could it also return maybe the indexes of documents whose write failed?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
ivansenic: yep, and write does not.. We can use write when we have auto generated IDs..

ivansenic:
@EricBorczuk OK then, I kept the a batch per document, making this more efficient.. I am returning IDs of the documents that have succeeded.. I guess this is fine when you supply the ID in the document, but when it's auto generated you don't know what's actually inserted. However, this is exactly a place where we can add `ordered` attribute and insert doc by doc, meaning you know by the amount of returned IDs where the failure happened.

Is this fine then?

ivansenic:
@EricBorczuk This is finalized from my point of view.
@mpenick What's your opinion on the changed impl details.

EricBorczuk (approved):

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -154,6 +159,128 @@ public ReactiveDocumentService(
...
+                                            db.treatBooleansAsNumeric(),
+                                            context);
+                                      }
+                                    })
+
+                                // if successful return pair with true
+                                .map(result -> Pair.with(documentId, true))
+
+                                // on any error mark this document id as failed
+                                .onErrorReturn(t -> Pair.with(documentId, false));
```
mpenick, line 265: Is this the exception information thrown away here? Or is this handled exception logged somewhere else?

restapi/src/main/java/io/stargate/web/docsapi/resources/ReactiveDocumentResourceV2.java
```diff
@@ -164,6 +166,75 @@ public void postDoc(
...
+      code = 202)
+  @ApiResponses(
+      value = {
+        @ApiResponse(code = 202, message = "Accepted", response = MultiDocsResponse.class),
+        @ApiResponse(code = 400, message = "Bad request", response = ApiError.class),
+        @ApiResponse(code = 401, message = "Unauthorized", response = ApiError.class),
+        @ApiResponse(code = 403, message = "Forbidden", response = ApiError.class),
+        @ApiResponse(code = 500, message = "Internal Server Error", response = ApiError.class)
+      })
+  @Path("collections/{collection-id}/batch")
```
mpenick, line 184: What happens when a batch only partially succeeds? Should we mention that the updates to multiple docs is not atomic  or isolated?

mpenick: > This implementation actually fires a batch per document in-parallel, making each batch statement a single partition batch, thus more effective. However, this would mean that if one document write fails, other will succeed. My idea would be that the result of the HTTP call sends back the IDs of documents that were successfully written. Write errors are effectively ignored.

I see. You've answered this. We should mention these limitations in the docs.

mpenick (approved):
We should prominently mention that updates to multiple documents using a batch is not atomic or isolated.

Also, minorly concerned that we're throwing away error information, that might be logged or added to metrics.

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
ivansenic: Ok we can log the exception for sure..

restapi/src/main/java/io/stargate/web/docsapi/resources/ReactiveDocumentResourceV2.java
ivansenic: @mpenick @EricBorczuk I am not kinda worried about this not atomic batch.. Actually this means the `../batch` endpoint will always return `200`.. This seems wrong in a way and we would never get it in the metrics (OK logs would be there).. Maybe I should have keep the behavior until now and fire one big batch for all documents, which succeeds or not.. Not sure what to do, can we talk 5 mins about this please?

mpenick: I think it's fine too as long as its semantics are documented.

ivansenic:
@polandll Lorina, we changed the semantic of the batch document write in this ticket. I adapted the Swagger API docs, not sure if this has to be adapted by you as well somewhere else.. 

I have checked the public docs on the stargate.io and batch is not even mentioned there.. So maybe it would be a good idea to add that as a section and explain the new approach. I can create a ticket or help you our with anything you need?

PR #1623, opened 2022-02-07
Extract new StargateBridgeClient abstraction

olim7t:
**What this PR does**:
- introduce a new `StargateBridgeClient`, that services will use instead of interacting with stubs directly. This will allow us to handle common concerns in a centralized place. In particular, the initial motivation for this change was to offer a read-after-write guarantee for schema operations.
- revamp the schema cache (now `StargateBridgeSchema`). It is now read-through, the first client to access a keyspace will trigger the persistence request. Schema notifications merely invalidate the entry, to force a refresh on next access.


sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultGrpcSchema.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class DefaultGrpcSchema implements GrpcSchema {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultGrpcSchema.class);
+
+  private static final int KEYSPACE_CACHE_SIZE = 1000;
+  private static final Duration KEYSPACE_CACHE_TTL = Duration.ofMinutes(5);
```
olim7t, line 44: These are tentative, we can adjust them as needed, maybe make them configurable.

tatu-at-datastax: Sounds good. Would be nice to somehow collect caching settings, document, but I don't know what'd be the best way. Perhaps we will just collect at a later point (easy enough to grep Caffeine Cache references)

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultGrpcSchema.java
```diff
@@ -0,0 +1,191 @@
...
+            .build();
+    registerChangeObserver();
+  }
+
+  @Override
+  public CompletionStage<CqlKeyspaceDescribe> getKeyspaceAsync(String keyspaceName) {
+    LOG.debug("getKeyspaceAsync({})", keyspaceName);
+    CompletionStage<CqlKeyspaceDescribe> existing = keyspaceCache.getIfPresent(keyspaceName);
+    if (existing != null) {
+      LOG.debug("Found entry in the cache, returning");
```
tatu-at-datastax, line 73: Maybe add `keyspaceName` in message (otherwise not very useful log statement)

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/GrpcClient.java
```diff
@@ -13,21 +13,18 @@
...
 import io.stargate.proto.Schema.CqlKeyspaceDescribe;
 
-/**
- * Allows clients to register with {@link SchemaCache} to get notified when keyspaces change.
- *
- * <p>Note no guarantee is made as to which threads will invoke those methods; implementations
- * should use proper synchronization if needed, and not block.
- */
-public interface SchemaListener {
+public interface GrpcClient {
```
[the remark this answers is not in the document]
tatu-at-datastax: lol git. :)

grpc-proto/proto/schema.proto
```diff
@@ -96,12 +96,14 @@ message GetSchemaNotificationsParams {
...
 
 message SchemaNotification {
-  SchemaChange change = 1;
-
-  // The full metadata of the keyspace that was impacted by the change.
-  // This is present for every kind of change, except if the keyspace doesn't exist anymore at the
-  // time of the notification (note that this is not only for a DROP KEYSPACE event, for example if
-  // a keyspace that contained a table gets dropped, the DROP TABLE event will also have no
-  // keyspace).
-  CqlKeyspaceDescribe keyspace = 2;
```
olim7t, line 106: I removed this to save bandwidth, the client is not using it anymore.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultGrpcSchema.java
olim7t: It should be fairly easy to correlate with the log from line 70, if the logs contain thread names.
But I can also merge them together...

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultGrpcSchema.java
olim7t: Also metrics

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultGrpcClient.java
```diff
@@ -0,0 +1,83 @@
...
+
+  @Override
+  public Response executeQuery(Query request) {
+    Response response =
+        ClientCalls.blockingUnaryCall(
+            channel, StargateBridgeGrpc.getExecuteQueryMethod(), callOptions, request);
+    if (response.hasSchemaChange()) {
+      // Ensure the impacted keyspace is fresh before we return control to the caller
+      String keyspaceName = decorateKeyspaceName(response.getSchemaChange().getKeyspace());
+      schema.forceKeyspaceRefresh(keyspaceName);
```
mpenick, line 62: The reason this is not just an invalidation is to not put the latency in the next query that requires this information?

tatu-at-datastax: I think it is to enforce/improve causality for common case: someone creates/changes a table, then does insert... without wait, DML might arrive at Service faster than schema sync update did, and possibly see wrong version of table (or that it was missing)?

olim7t: Actually that was an oversight, it can be a simple invalidation. We don't need to wait, if the caller tries to get the keyspace immediately after it will trigger the background query, like any other client. I'll make the change.

tatu-at-datastax:
At risk of Bike Shedding, I think the name should reflect "Bridge" part of grpc client -- so maybe `GrpcBridgeClient` if not even `StargateBridgeClient` without necessarily mentioning grpc (I'm fine it being included just not mandatory).

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/CreateGrpcClientFilter.java
```diff
@@ -0,0 +1,90 @@
...
+ * Validates that the auth token is present and pre-builds the gRPC client before each request.
+ *
+ * <p>Note that the resource class or method must be annotated with {@link CreateGrpcStub} in order
+ * for this filter to run.
+ *
+ * @see GrpcClientJerseyFactory
+ */
+@Provider
+@CreateGrpcStub
+public class CreateGrpcClientFilter implements ContainerRequestFilter {
```
mpenick, line 40: Maybe rename to match new names `CreateStargateBridgeClientFilter`? Not a big deal though.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
```diff
@@ -0,0 +1,156 @@
...
+    CqlKeyspace keyspace = describe.getCqlKeyspace();
+    return CqlKeyspaceDescribe.newBuilder(describe)
+        .setCqlKeyspace(
+            CqlKeyspace.newBuilder(keyspace).setName(stripTenantPrefix(keyspace.getName())).build())
+        .build();
+  }
+
+  private String stripTenantPrefix(String fullKeyspaceName) {
+    if (fullKeyspaceName.startsWith(tenantPrefix)) {
+      return tenantPrefix;
```
mpenick, line 152: Should this return `tenantPrefix` here? Maybe I'm misunderstanding, but  I thought if `fullKeyspace = "abc_ksname"` and `tenantPrefix = "abc"` this would return `"ksname"`.

olim7t: Good catch, will fix

PR #1629, opened 2022-02-09
Update startup scripts and docker images to incorporate bridge token

jeffreyscarpenter:
Updating startup scripts and docker compose configurations to pass the same bridge token to the SGv2 REST API service and coordinator. Included as part of this are the introduction of a new `starctl-service-rest` script to start the REST API service, which can be used locally, and is also incorporated into the REST docker image as the entrypoint. There is also a convenience `start_all_local.sh` script to start both REST and the coordinator via `starctl`/`starctl-service-rest` with the same bridge token. Finally, the SGv2 REST service pom was updated so that its fat jar is placed under the `stargate-lib/rest` directory. This will help improve how we build the `stargate-jars.zip` for the next release.

Note: will likely need #1628 and possibly #1627 incorporated before merging this.


starctl-service-rest
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
```
tatu-at-datastax, line 1: Maybe add brief code comment on purpose of the script?

starctl-service-rest
```diff
@@ -0,0 +1,76 @@
...
+    exit 1
+fi
+
+if [[ "$_java" ]]; then
+    jvmver=$("$_java" -version 2>&1 | awk -F '"' '/version/ {print $2}')
+    JVM_VERSION=${jvmver%_*}
+    JVM_PATCH_VERSION=${jvmver#*_}
+
+    if [ "$JVM_VERSION" \< "1.8" ] ; then
+        echo "Stargate requires Java 8u40 or later."
```
tatu-at-datastax, line 19: I assume this comes from existing codebase, but I wonder if there's much point... 1.8.40 is ancient as-is. Could reduce less useful checking.
(specifically thinking that looking for 1.8+ is fine, but patch level is bit arbitrary)

tatu-at-datastax:
Looks like Starter test(s) failed for some reason.

PR #1634, opened 2022-02-12
Port remaining resources to use StargateBridgeClient

olim7t:
**What this PR does**:
Finish migrating all the REST v2 resources to use `StargateBridgeClient`.


sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/StargateBridgeClientFactory.java
```diff
@@ -16,14 +16,18 @@
...
 import java.util.concurrent.ScheduledExecutorService;
 
 public interface StargateBridgeClientFactory {
 
   static StargateBridgeClientFactory newInstance(
-      Channel channel, String adminToken, ScheduledExecutorService executor) {
-    return new DefaultStargateBridgeClientFactory(channel, adminToken, executor);
+      Channel channel,
+      String adminToken,
+      SchemaRead.SourceApi sourceApi,
```
olim7t, line 28: Something I had missed in the previous PR. `REST` was hard-coded in the common code.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
```diff
@@ -90,15 +96,16 @@ public Response executeBatch(Batch request) {
   }
 
   @Override
-  public CqlKeyspaceDescribe getKeyspace(String keyspaceName) throws UnauthorizedKeyspaceException {
-    if (!authorizeSchemaRead(SchemaReads.keyspace(keyspaceName, SchemaRead.SourceApi.REST))) {
+  public Optional<CqlKeyspaceDescribe> getKeyspace(String keyspaceName)
+      throws UnauthorizedKeyspaceException {
```
olim7t, line 100: Switched to optional. As much as I'm comfortable using null in the cache, this is a higher level API exposed to other modules.

Also, I made `UnauthorizedKeyspaceException` (and its new table counterpart) unchecked. There is a catch-all mapper `BridgeAuthorizationExceptionMapper` in the web layer. This make it easier to deal with those errors, in particular in code that composes optionals.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/StargateBridgeClientFactory.java
tatu-at-datastax: Good catch then.

One relatively immediate challenge to consider, however: how do make this extensible for external APIs/Services when we no longer control set of APIs? Should there be some sort of "other" value for someone developing new ones, for example.

tatu-at-datastax (approved):

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2IndexesResourceImpl.java
```diff
@@ -54,7 +52,8 @@ public Response getAllIndexesForTable(
       throw new WebApplicationException("tableName must be provided", Status.BAD_REQUEST);
     }
 
-    BridgeSchemaClient.create(blockingStub).findTable(keyspaceName, tableName);
+    // check that we're authorized for the table
+    bridge.getTable(keyspaceName, tableName);
```
ivansenic, line 56: Should the `ResourceBase` class provide a `callWithTable` which actually accepts the Table object, so we don't call the `getTable` twice? Maybe I am reading wrong and this is not possible :relaxed:

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2IndexesResourceImpl.java
```diff
@@ -85,12 +84,18 @@ public Response addIndex(
...
+        .flatMap(
+            table ->
+                table.getColumnsList().stream()
+                    .filter(c -> columnName.equals(c.getName()))
+                    .findAny())
+        .orElseThrow(
+            () ->
+                new WebApplicationException(
+                    String.format("Column '%s' not found in table.", columnName),
+                    Status.NOT_FOUND));
```
ivansenic, line 98: This will be the error also if the table is not there.. do a `.map` instead of the `.flatMap` and move exception after the `.findAny()`

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2IndexesResourceImpl.java
```diff
@@ -126,17 +131,21 @@ public Response dropIndex(
...
+        .getTable(keyspaceName, tableName)
+        .flatMap(
+            table ->
+                table.getIndexesList().stream()
+                    .filter(i -> indexName.equals(i.getName()))
+                    .findAny())
+        .orElseThrow(
+            () ->
+                new WebApplicationException(
+                    String.format("Index '%s' not found.", indexName), Status.NOT_FOUND));
```
ivansenic, line 144: same here..

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
```diff
@@ -90,15 +96,16 @@ public Response executeBatch(Batch request) {
...
+    if (!authorizeSchemaRead(SchemaReads.keyspace(keyspaceName, sourceApi))) {
       throw new UnauthorizedKeyspaceException(keyspaceName);
     }
     return getAuthorizedKeyspace(keyspaceName);
   }
 
-  private CqlKeyspaceDescribe getAuthorizedKeyspace(String keyspaceName) {
-    return stripTenantPrefix(schema.getKeyspace(addTenantPrefix(keyspaceName)));
+  private Optional<CqlKeyspaceDescribe> getAuthorizedKeyspace(String keyspaceName) {
+    return Optional.of(stripTenantPrefix(schema.getKeyspace(addTenantPrefix(keyspaceName))));
```
ivansenic, line 108: I am thinking that the whole Tenant prefix stuff, should be extracted to a "util" pojo class that can accept the tenant id and then provide methods like `stripTenantPrefix` and `addTenantPrefix`.. It can be tested good in the unit test and reused where needed..

These lines (https://github.com/stargate/stargate/pull/1634/files#diff-da5eacd2c92cb5e4fa71fa2de0240eed08a2dcd1ad8287cdebc4abe34edc04c9R199-R223) seems as the candidate for extraction here..

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
```diff
@@ -107,20 +114,55 @@ private CqlKeyspaceDescribe getAuthorizedKeyspace(String keyspaceName) {
...
+                .map(n -> SchemaReads.keyspace(n, sourceApi))
                 .collect(Collectors.toList()));
     List<CqlKeyspaceDescribe> keyspaces = new ArrayList<>(names.size());
     int i = 0;
     for (String name : names) {
-      CqlKeyspaceDescribe ks;
-      // Null check because it could have been deleted since we listed the names
-      if (authorizations.get(i) && (ks = getAuthorizedKeyspace(name)) != null) {
-        keyspaces.add(ks);
+      if (authorizations.get(i)) {
```
ivansenic, line 122: is `int i` always zero here?

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
```diff
@@ -107,20 +114,55 @@ private CqlKeyspaceDescribe getAuthorizedKeyspace(String keyspaceName) {
...
+        .map(
+            keyspace -> {
+              List<Schema.CqlTable> tables = keyspace.getTablesList();
+              List<Boolean> authorizations =
+                  authorizeSchemaReads(
+                      tables.stream()
+                          .map(t -> SchemaReads.table(keyspaceName, t.getName(), sourceApi))
+                          .collect(Collectors.toList()));
+              List<Schema.CqlTable> authorizedTables = new ArrayList<>(tables.size());
+              int i = 0;
```
ivansenic, line 155: here also, `int i = 0` is not changed or?

ivansenic: Do we have a unit test for the `DefaultStargateBridgeClient`

ivansenic (commented):
Had a look, left few comments..

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
olim7t: I'll add the test in a follow-up PR, it's going to take a bit longer than expected and I don't want to hold other work that depends on this PR.

PR #1640, opened 2022-02-15
Fix #1639: refactor Materialized View REST API ITs into separate test class

### LINKED ISSUE #1639: Refactor Rest API Integration tests that test Materialized Views into separate class
Since some persistence backends do not have support for Materialized Views, tests that exercise them will fail on such backends.
Currently we have at least one test in `RestApiv2RowsTest` (method `getAllRowsFromMaterializedView()`).
It would make sense to create a separate IT class to make it easier to exclude running of this test specifically for relevant backends.

tatu-at-datastax:
**What this PR does**:

Further refactors REST API v2 tests to move:

* Materialized View related tests into new class

The reason being that further testing by downstream projects has issues with a backend that does not (yet?) support Materialized Views, and test exclusion is easier with more granular division of tests methods into classes.

Fixes #1639

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/testing/src/main/java/io/stargate/it/http/restapi/RestApiTestBase.java
+++ b/testing/src/main/java/io/stargate/it/http/restapi/RestApiTestBase.java
@@ -8,18 +8,25 @@ import io.stargate.it.BaseIntegrationTest;
 import io.stargate.it.http.RestUtils;
 import io.stargate.it.http.models.Credentials;
 import io.stargate.it.storage.StargateConnectionInfo;
+import io.stargate.web.restapi.models.ColumnDefinition;
 import io.stargate.web.restapi.models.GetResponseWrapper;
+import io.stargate.web.restapi.models.PrimaryKey;
+import io.stargate.web.restapi.models.TableAdd;
+import io.stargate.web.restapi.models.TableResponse;
 import java.io.IOException;
 import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import org.apache.http.HttpStatus;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.TestInfo;
 
 /** Intermediate IT base class for use by REST API integration tests. */
-public class RestApiTestBase extends BaseIntegrationTest {
+abstract class RestApiTestBase extends BaseIntegrationTest {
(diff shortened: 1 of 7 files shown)


testing/src/main/java/io/stargate/it/http/restapi/RestApiTestBase.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.apache.http.HttpStatus;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.TestInfo;
+
+/** Intermediate IT base class for use by REST API integration tests. */
+abstract class RestApiTestBase extends BaseIntegrationTest {
```
jeffreyscarpenter, line 29: Just FYI, this name may create some confusion upon merging into the v2.0.0 package, where a `io.stargate.it.http.BaseRestApiTest` has been added already.

tatu-at-datastax: Ohhh. Thanks, I'll change that then!

tatu-at-datastax: Ok I think I'll get this merged for `master`, merge & refactor in v2.0.0 to combine classes.

jeffreyscarpenter (approved):

testing/src/main/java/io/stargate/it/http/restapi/RestApiTestBase.java
ivansenic: I would say this functionality already exists, check this peace of code :arrow_down:  https://github.com/stargate/stargate/blob/d0d0ddf15f05c9185d3e0b6ac9a0d3df4cb32ae7/testing/src/main/java/io/stargate/it/http/docsapi/CollectionsResourceIntTest.java#L83-L94

So `@CqlSessionSpec` and `@TestKeyspace CqlIdentifier keyspace` do the trick.. I like this annotation approach way more..

ivansenic: And I am not convinced create table and insert rows should be done is such a way.. The `CqlSessionSpec` has `initQueries` properties that can create table easily or add data transparently for the test setup..

ivansenic (approved):
Left a comment..

PR #1642, opened 2022-02-16
ability to add extra tags for each gRPC call

ivansenic:
**What this PR does**:
Adds possibility to add extra tags for each gRPC server-side call. This extends the original Micrometer implementation, as that is not supported yet there.

There's nothing too interesting here, only new interceptor, OSGi gymnastics to make everything work and integration test to confirm everything. I am not sure if unit tests for the `TaggingMetricCollectingServerInterceptor` is needed here?

Internal issue.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/grpc/src/main/java/io/micrometer/core/instrument/binder/grpc/TaggingMetricCollectingServerInterceptor.java
+++ b/grpc/src/main/java/io/micrometer/core/instrument/binder/grpc/TaggingMetricCollectingServerInterceptor.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.micrometer.core.instrument.binder.grpc;
+
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+import io.grpc.ServerCall;
+import io.grpc.ServerCallHandler;
+import io.grpc.Status;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.Timer;
+import io.stargate.grpc.metrics.api.GrpcMetricsTagProvider;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.UnaryOperator;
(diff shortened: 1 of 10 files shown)


grpc/src/main/java/io/micrometer/core/instrument/binder/grpc/TaggingMetricCollectingServerInterceptor.java
```diff
@@ -0,0 +1,147 @@
...
+  private MetricSet metricsFor(ServerCall<?, ?> call) {
+    MethodDescriptor<?, ?> method = call.getMethodDescriptor();
+    Tags callTags = tagProvider.getCallTags(call);
+
+    // check if there is any extension, if not return super
+    // otherwise use our implementation
+    if (Tags.empty().equals(callTags)) {
+      return super.metricsFor(method);
+    } else {
+      return new MetricSet(
```
mpenick, line 106: It looks like micrometer tries to cache the `MetricSet` per method if it can: https://github.com/micrometer-metrics/micrometer/blob/v1.7.3/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/AbstractMetricCollectingInterceptor.java#L164

Is that something we should try to do (using tags/method)? Or is this not a big deal performance-wise?

grpc/src/main/java/io/micrometer/core/instrument/binder/grpc/TaggingMetricCollectingServerInterceptor.java
```diff
@@ -0,0 +1,147 @@
...
+ * Extension of the micrometer's {@link MetricCollectingServerInterceptor} that enables us to add
+ * extra tags on each {@link ServerCall}.
+ */
+public class TaggingMetricCollectingServerInterceptor extends MetricCollectingServerInterceptor {
+
+  /**
+   * The total number of requests received, same as
+   * MetricCollectingServerInterceptor#METRIC_NAME_SERVER_REQUESTS_RECEIVED.
+   */
+  private static final String METRIC_NAME_SERVER_REQUESTS_RECEIVED =
```
mpenick, line 44: Unfortunate that `MetricCollectingServerInterceptor` made these constants private... or it would have been nice if `newRequestCounterFor()` etc. returned builders.

mpenick (approved):
Looks great.

grpc/src/main/java/io/micrometer/core/instrument/binder/grpc/TaggingMetricCollectingServerInterceptor.java
ivansenic: Yea they are caching as there are no extra tags per call.. And they are able to cache per method.. In our case you can not cache, as tags can change on each call..

However, we do use the caching if the tag provider returns empty tags.. So in case of Noop tag provider we will use original impl.. Note that this works because the amount of tags returned by the tag provider have always to be same size..

grpc/src/main/java/io/micrometer/core/instrument/binder/grpc/TaggingMetricCollectingServerInterceptor.java
ivansenic: Yes this is really unfortunate.. Maybe I can try to provide a fix for the micrometer quickly..

PR #1648, opened 2022-02-17
Add a listener mechanism to StargateBridgeSchema

olim7t:
**What this PR does**:
Add a listener mechanism to `StargateBridgeSchema`. This will be useful for the GraphQL API.
We don't need all the different `onUpdated/onRemoved...` notifications, callers will follow the same pattern as the cache itself: whenever anything changes, invalidate their copy and refresh (that's what the GraphQL cache does in v1 anyway).

N/A

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeSchema.java
+++ b/sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeSchema.java
@@ -17,6 +17,8 @@ package io.stargate.sgv2.common.grpc;
 
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.RemovalCause;
+import com.github.benmanes.caffeine.cache.RemovalListener;
 import io.grpc.Channel;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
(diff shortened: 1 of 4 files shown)


sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeSchema.java
```diff
@@ -199,7 +216,15 @@ public void onCompleted() {
...
   private void scheduleReconnect() {
-    executor.schedule(this::registerChangeObserver, getAndIncreaseDelayMs(), TimeUnit.MILLISECONDS);
+    try {
+      if (!executor.isShutdown()) {
+        executor.schedule(
+            this::registerChangeObserver, getAndIncreaseDelayMs(), TimeUnit.MILLISECONDS);
+      }
+    } catch (RejectedExecutionException e) {
+      // The code above can occasionally race, avoid printing a full stack trace
+      LOG.warn("Reconnection task was rejected, this is normal if the server is shutting down");
```
EricBorczuk, line 226: Maybe include just the `getMessage`? Maybe that's not that useful though overall

EricBorczuk (approved):

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeSchema.java
tatu-at-datastax: I would suggest adding `getMessage()` at first; if it proves to be useless can remove but maybe better not to assume it is.

PR #1650, opened 2022-02-17
closes #1649: fixed BundleAvailableCondition to support nested classes

### LINKED ISSUE #1649: `@IfBundleAvailable` does not support nested tests classes
The JUnit condition `BundleAvailableCondition` is not working with the nested test classes. Seems this is only used for gRPC test at the moment. I guess it was added once `grpc` modules were not in the default profile.

We can delete it or fix as this can be a valuable test util in the future.

ivansenic:
**What this PR does**:
This annoyed me for second time now, and I decided to fix it.. Trick is to use `SearchOption.INCLUDE_ENCLOSING_CLASSES`.. Confirmed that it works with nested classes as well now.. Confirmed that not existing bundles would result in ignoring tests with `@IfBundleAvailable(bundleName = "grpc1")` :

```
Unable to find bundle 'grpc1'

Unable to find bundle 'grpc1'

Unable to find bundle 'grpc1'

Unable to find bundle 'grpc1'

Process finished with exit code 0
```

Fixes #1649.

**Checklist**

### DIFF
--- a/testing/src/main/java/io/stargate/it/storage/BundleAvailableCondition.java
+++ b/testing/src/main/java/io/stargate/it/storage/BundleAvailableCondition.java
@@ -9,6 +9,7 @@ import org.junit.jupiter.api.extension.ConditionEvaluationResult;
 import org.junit.jupiter.api.extension.ExecutionCondition;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.platform.commons.support.AnnotationSupport;
+import org.junit.platform.commons.support.SearchOption;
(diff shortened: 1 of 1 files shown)


tatu-at-datastax (approved):
As per my note on the issue, fix itself looks solid, although we may want to consider just removing the annotation altogether as it isn't and won't be needed (if I understood things correctly).

ivansenic:
I have no idea if we would need this in future, but until we don't get rid of the OSGi i t can be handy..

PR #1654, opened 2022-02-18
grcp metric interceptor to be before the connection one

ivansenic:
**What this PR does**:
Makes the gRPC metrics interceptor before the connection interceptor.. This should enable us to correctly catch the unauthorized calls in the metrics..

Internal.

@mpenick can you ack this is OK?


ivansenic:
Closing as order was already correct. Seems that the last registered interceptor will be the first one in gRPC. Strange, but luckily there was a test already to verify the correct behavior.

PR #1647, opened 2022-02-17
Second part of #1644 fix: C*4 from 4.0.1 to 4.0.3 upgrade

tatu-at-datastax:
**What this PR does**:

Second upgrade for #1644: Cassandra 4.x dependency from 4.0.1 to 4.0.3

**Which issue(s) this PR fixes**:

Part of #1644.

**Checklist**
- [ ] Changes manually tested
- [x] Automated Tests added/updated
- [ ] Documentation added/updated
- [x] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

PR #1656, opened 2022-02-24
(WIP) Try to reproduce and fix race condition (#1647)

tatu-at-datastax:
**What this PR does**:

Fixes #1647

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/testing/src/main/java/io/stargate/it/grpc/streaming/ExecuteBatchStreamingTest.java
+++ b/testing/src/main/java/io/stargate/it/grpc/streaming/ExecuteBatchStreamingTest.java
@@ -51,6 +51,33 @@ import org.junit.jupiter.api.extension.ExtendWith;
       "CREATE TABLE IF NOT EXISTS test (k text, v int, PRIMARY KEY(k, v))",
     })
 public class ExecuteBatchStreamingTest extends GrpcIntegrationTest {
+  private static class StreamObserverWithList implements StreamObserver<StreamingResponse> {
+    private final List<StreamingResponse> responses;
+    private final AtomicReference<Throwable> errorHolder;
+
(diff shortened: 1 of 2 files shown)


testing/src/main/java/io/stargate/it/grpc/streaming/ExecuteBatchStreamingTest.java
```diff
@@ -116,31 +130,45 @@ public void onCompleted() {}
...
+              .addQueries(
+                  cqlBatchQuery(
+                      "INSERT INTO test (k, v) VALUES (?, ?)", Values.of("b"), Values.of(2)))
+              .setParameters(batchParameters(keyspace))
+              .build());
+    }
+
+    Awaitility.await()
+        .atMost(30, TimeUnit.SECONDS)
+        .untilAsserted(() -> assertThat(responses.size()).isEqualTo(2 * ROUNDS));
```
tatu-at-datastax, line 158: Fails here only getting 1000 -- but I don't think that actually reproduces the NPE we want to catch; both since no matching error found in logs and since suggested fix for `QueryProcessor` access does not solve the fail.
So I suspect assertion here is incorrect.

ivansenic:
@tatu-at-datastax I don't know what was the idea, but here you don't have upgrade to C* `4.0.2`?

tatu-at-datastax:
@ivansenic Correct, the original PR (#1647) has that.

Ohh... shoot. I am an idiot.

ivansenic:
Let s close this one, you can port anything you want to the original PR.

tatu-at-datastax:
> Let s close this one, you can port anything you want to the original PR.

Yup.

PR #1661, opened 2022-02-25
closes #1659: proper handling of retries for unprepared errors in grpc

### LINKED ISSUE #1659: gRPC to correctly hande retries in case of the `PreparedQueryNotFoundException`
In the https://github.com/stargate/stargate/pull/1647, we had to add retries in the gRPC in case of the `PreparedQueryNotFoundException`. These retries have no retry limits at the moment. This is in line with the [PersistenceBackedDataStore](https://github.com/stargate/stargate/blob/f892620938119752ca9a04690140d02a12e8f3d3/persistence-api/src/main/java/io/stargate/db/datastore/PersistenceBackedDataStore.java#L115-L137), that always does a retry on such exception.

gRPC implements it's own `RetryPolicy` which defines in what situations can retries be done. The `RetryPolicy` should be extended in order to support `onUnprepared(int retryCount)`.

We need to agree on:
- Should default retry policy allow limitless retries in case of the unprepared exception?
- If not, what is the max retry limit?

The `PersistenceExceptionTest` should be extended to cover this case.

ivansenic:
**What this PR does**:
Extends the grpc `RetryPolicy` in order to allow specific retries for the `UNPREPARED` error code. Currently allows limites retries, but we should agree in this PR if this is fine or not. See #1659 for more details.

Fixes #1659

**Checklist**

### DIFF
--- a/grpc/src/main/java/io/stargate/grpc/retries/DefaultRetryPolicy.java
+++ b/grpc/src/main/java/io/stargate/grpc/retries/DefaultRetryPolicy.java
@@ -31,7 +31,7 @@ import org.slf4j.LoggerFactory;
  * <p>This is a very conservative implementation: it triggers a maximum of one retry per request,
  * and only in cases that have a high chance of success (see the method javadocs for detailed
  * explanations of each case). The exception is the {@link
- * RetryPolicy#onUnprepared(PreparedQueryNotFoundException, int)}, which allows limitless retries.
+ * RetryPolicy#onUnprepared(PreparedQueryNotFoundException, int)}, which allows 2 retries.
  */
 @ThreadSafe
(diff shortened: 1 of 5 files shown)


ivansenic:
@mpenick Let's agree if limitless retries are fine? Can adapt to whatever we want..

mpenick:
I think I'd prefer we cap the number of retries. Default to 1 or 2, and potentially have a property to set to something higher. I'm worried at the potential for something like this to spin in a loop because of a defect.

ivansenic:
Fixed to 2 retries. As other methods don't expose a property for their limits, I kept the same behavior here as well.. Let's see if this passes our test now..

jeffreyscarpenter (approved):

PR #1662, opened 2022-02-25
closes #1660: non-api endpoint metrics to be reported with different …

### LINKED ISSUE #1660: Metrics for non-API endpoints should be reported as different module
Currently we have a set of the health and utility endpoints in each module. The metrics for those endpoints should not be recorded with module tag that is same as the module where they are existing. With this we want to have a clear aggregated metric views on the real API endpoints, without interference of those utility endpoints.

The endpoints are:

* *restapi*
   * `^root$`
   * `^/health$`
   * `^/swagger.*`
* *graphqlapi*
   * `^/playground$`
   * `^/graphql-schema$`

ivansenic:
**What this PR does**:
Enables possibility to tag specific endpoints with a `module=basemodule-other` tag when reporting metrics. See #1660 for more information.

Fixes #1660.

**Checklist**

### DIFF
--- a/graphqlapi/src/main/java/io/stargate/graphql/web/DropwizardServer.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/web/DropwizardServer.java
@@ -38,6 +38,7 @@ import io.stargate.graphql.web.resources.FilesResource;
 import io.stargate.graphql.web.resources.GraphqlCache;
 import io.stargate.graphql.web.resources.PlaygroundResource;
 import io.stargate.metrics.jersey.MetricsBinder;
+import java.util.Arrays;
 import java.util.EnumSet;
 import javax.servlet.DispatcherType;
 import javax.servlet.FilterRegistration;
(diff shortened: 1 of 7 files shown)


ivansenic:
@jeffreyscarpenter Let's quickly confirm that this is something we need..

graphqlapi/src/main/java/io/stargate/graphql/web/DropwizardServer.java
```diff
@@ -50,6 +51,9 @@
 
 public class DropwizardServer extends Application<Configuration> {
 
+  public static final String[] NON_API_URI_REGEX =
+      new String[] {"^/playground$", "^/graphql-schema$"};
```
tatu-at-datastax, line 55: These could be combined into a single regexp, unless there is a specific reason for keeping them separate?
(combining would result in a single RE that is faster to evaluate than going over 2 separately)

jeffreyscarpenter (approved):
Looks good to me

graphqlapi/src/main/java/io/stargate/graphql/web/DropwizardServer.java
ivansenic: Yes of course, stupid me..

PR #1663, opened 2022-02-28
closes #1220: json with empty field names not supported

### LINKED ISSUE #1220: DocsApi allows for creating json with value and empty key

ivansenic:
**What this PR does**:
As agreed with @EricBorczuk, allowing empty keys in JSON, although they are valid, would represent quite a challenge for us in the documents API. Thus, we decided that for now we explicitly forbid this and return `400` in case such document is written. We assumed this will not affect users, as empty JSON keys are an anti-pattern. The handling might change in future.

Fixes #1220.

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
@@ -165,6 +165,13 @@ public class JsonDocumentShredder {
             field -> {
               String fieldName = field.getKey();
 
+              if (fieldName.isEmpty()) {
+                String msg =
+                    "JSON objects containing empty field names are not supported at the moment.";
+                throw new ErrorCodeRuntimeException(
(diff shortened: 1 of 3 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/JsonDocumentShredder.java
```diff
@@ -165,6 +165,13 @@ private void processObjectNode(
             field -> {
               String fieldName = field.getKey();
 
+              if (fieldName.isEmpty()) {
+                String msg =
+                    "JSON objects containing empty field names are not supported at the moment.";
```
tatu-at-datastax, line 170: nit: maybe just "not supported" since otherwise we are implying there's a chance/plan for changing.
If users want this to be supported they can of course request it.

tatu-at-datastax (approved):
I like it -- I think it is better to start with more strict, re-consider with real use case, user request, if need be.

PR #1669, opened 2022-03-01
Improve logging by ConfigStoreYaml

tatu-at-datastax:
**What this PR does**:

Adds more information on logging by `ConfigStoreYaml`, both for failures and successful loads.

No issue filed, needed for troubleshooting

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreYaml.java
+++ b/config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreYaml.java
@@ -25,14 +25,19 @@ import io.stargate.config.store.api.ConfigStore;
 import io.stargate.config.store.api.ConfigWithOverrides;
 import io.stargate.config.store.api.MissingModuleSettingsException;
 import io.stargate.config.store.yaml.metrics.CacheMetricsRegistry;
+import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.time.Duration;
 import java.util.Collections;
 import java.util.Map;
 import java.util.concurrent.CompletionException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


jeffreyscarpenter (approved):

config-store-yaml/src/main/java/io/stargate/config/store/yaml/ConfigStoreYaml.java
```diff
@@ -56,28 +61,41 @@ public ConfigStoreYaml(Path configFilePath, Ticker ticker, MetricRegistry metric
...
         throw new MissingModuleSettingsException(
             String.format(
-                "The loaded configuration map: %s, does not contain settings from a given module: %s",
-                result, moduleName));
+                "The loaded configuration map (from '%s'): %s, does not contain settings from a given module: %s",
+                configFilePath, result, moduleName));
       }
-      return new ConfigWithOverrides(
-          Collections.unmodifiableMap(result.get(moduleName)), moduleName);
+      logger.info(
```
jeffreyscarpenter, line 90: I'm ambivalent about making this info vs debug but I think its ok. this should only happen once in the life of a stargate node right?

tatu-at-datastax: I definitely hope so. I can easily change it to `debug` if it appears this gets logged too often; that's a valid concern for Integration Tests I think.

tatu-at-datastax: Alternatively I could move the message to actual loading: How about I'll do that actually...

tatu-at-datastax:
@jeffreyscarpenter Changed based on your suggestion: access (via cache) is now `debug`; actual loading (less common since there's just single config file shared by components) at `info`. I can change that to `debug` too if need be but I suspect it really is once per Coordinator run.

PR #1670, opened 2022-03-01
Add in TTL work, with tests

### LINKED ISSUE #1414: Enhancement request to include TTL for Documents in Astra DB

EricBorczuk:
**What this PR does**:
Allows assigning a TTL to a document. Updated version of https://github.com/stargate/stargate/pull/1428

Fixes #1414 

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
@@ -2,6 +2,7 @@ package io.stargate.web.docsapi.dao;
 
 import com.datastax.oss.driver.api.core.servererrors.AlreadyExistsException;
 import com.datastax.oss.driver.shaded.guava.common.base.Splitter;
+import com.google.common.collect.ImmutableList;
 import io.stargate.auth.AuthenticationSubject;
 import io.stargate.auth.AuthorizationService;
 import io.stargate.auth.Scope;
@@ -9,12 +10,13 @@ import io.stargate.auth.SourceAPI;
 import io.stargate.auth.UnauthorizedException;
 import io.stargate.db.datastore.DataStore;
 import io.stargate.db.datastore.ResultSet;
+import io.stargate.db.datastore.Row;
 import io.stargate.db.query.BoundQuery;
 import io.stargate.db.query.Predicate;
 import io.stargate.db.query.Query;
 import io.stargate.db.query.builder.BuiltCondition;
 import io.stargate.db.query.builder.QueryBuilder;
-import io.stargate.db.query.builder.ValueModifier;
+import io.stargate.db.query.builder.QueryBuilderImpl;
 import io.stargate.db.schema.Column;
 import io.stargate.db.schema.Column.Kind;
 import io.stargate.db.schema.Column.Type;
@@ -28,12 +30,13 @@ import io.stargate.web.docsapi.service.ExecutionContext;
 import io.stargate.web.docsapi.service.QueryExecutor;
 import io.stargate.web.docsapi.service.json.DeadLeaf;
 import io.stargate.web.docsapi.service.query.DocsApiConstants;
-import io.stargate.web.docsapi.service.util.ImmutableKeyspaceAndTable;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
@@ -118,6 +121,32 @@ public class DocumentDB {
 
   public Set<Keyspace> getKeyspaces() {
     return dataStore.schema().keyspaces();
+  }
+
+  public Integer getTtlForDocument(String namespace, String collection, String docId) {
+    List<Row> result =
+        this.dataStore
+            .queryBuilder()
+            .select()
+            .function(
+                ImmutableList.of(
+                    QueryBuilderImpl.FunctionCall.ttl(DocsApiConstants.LEAF_COLUMN_NAME)))
+            .from(namespace, collection)
+            .where(BuiltCondition.of(DocsApiConstants.KEY_COLUMN_NAME, Predicate.EQ, docId))
+            .build()
+            .execute()
+            .join()
+            .rows();
+    try {
+      return result.stream()
+          .map(row -> row.getInt(String.format("ttl(%s)", DocsApiConstants.LEAF_COLUMN_NAME)))
+          .filter(Objects::nonNull)
+          .mapToInt(Integer::valueOf)
+          .max()
+          .getAsInt();
+    } catch (NoSuchElementException e) {
+      return null;
+    }
   }
 
   public void writeJsonSchemaToCollection(String namespace, String collection, String schemaData) {
@@ -286,25 +315,6 @@ public class DocumentDB {
           ErrorCode.DATASTORE_TABLE_DOES_NOT_EXIST,
           String.format("Collection %s does not exist.", tableName));
     }
-  }
-
-  private Query getInsertQueryForTable(ImmutableKeyspaceAndTable info) {
-    // Add a bind position for each column
-    List<ValueModifier> markers = new ArrayList<>(allColumnNames.size());
-    for (String columnName : allColumnNames) {
-      markers.add(ValueModifier.marker(columnName));
-    }
-
-    // Add a bind position for a timestamp, build, and return
-    return dataStore
-        .prepare(
-            dataStore
-                .queryBuilder()
-                .insertInto(info.getKeyspace(), info.getTable())
-                .value(markers)
-                .timestamp()
-                .build())
-        .join();
   }
 
   private void createDefaultIndexes(String keyspaceName, String tableName)
--- a/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
@@ -2,7 +2,6 @@ package io.stargate.web.docsapi.dao;
 
 import com.datastax.oss.driver.api.core.servererrors.AlreadyExistsException;
 import com.datastax.oss.driver.shaded.guava.common.base.Splitter;
-import com.google.common.collect.ImmutableList;
 import io.stargate.auth.AuthenticationSubject;
 import io.stargate.auth.AuthorizationService;
 import io.stargate.auth.Scope;
@@ -10,13 +9,11 @@ import io.stargate.auth.SourceAPI;
 import io.stargate.auth.UnauthorizedException;
 import io.stargate.db.datastore.DataStore;
 import io.stargate.db.datastore.ResultSet;
-import io.stargate.db.datastore.Row;
 import io.stargate.db.query.BoundQuery;
 import io.stargate.db.query.Predicate;
 import io.stargate.db.query.Query;
 import io.stargate.db.query.builder.BuiltCondition;
 import io.stargate.db.query.builder.QueryBuilder;
-import io.stargate.db.query.builder.QueryBuilderImpl;
 import io.stargate.db.schema.Column;
 import io.stargate.db.schema.Column.Kind;
 import io.stargate.db.schema.Column.Type;
@@ -35,8 +32,6 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
-import java.util.NoSuchElementException;
-import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
@@ -121,32 +116,6 @@ public class DocumentDB {
 
   public Set<Keyspace> getKeyspaces() {
     return dataStore.schema().keyspaces();
-  }
-
-  public Integer getTtlForDocument(String namespace, String collection, String docId) {
-    List<Row> result =
-        this.dataStore
-            .queryBuilder()
-            .select()
-            .function(
-                ImmutableList.of(
-                    QueryBuilderImpl.FunctionCall.ttl(DocsApiConstants.LEAF_COLUMN_NAME)))
-            .from(namespace, collection)
-            .where(BuiltCondition.of(DocsApiConstants.KEY_COLUMN_NAME, Predicate.EQ, docId))
-            .build()
-            .execute()
-            .join()
-            .rows();
-    try {
-      return result.stream()
-          .map(row -> row.getInt(String.format("ttl(%s)", DocsApiConstants.LEAF_COLUMN_NAME)))
-          .filter(Objects::nonNull)
-          .mapToInt(Integer::valueOf)
-          .max()
-          .getAsInt();
-    } catch (NoSuchElementException e) {
-      return null;
-    }
   }
 
   public void writeJsonSchemaToCollection(String namespace, String collection, String schemaData) {
--- a/restapi/src/main/java/io/stargate/web/docsapi/resources/ReactiveDocumentResourceV2.java
+++ b/restapi/src/main/java/io/stargate/web/docsapi/resources/ReactiveDocumentResourceV2.java
@@ -359,10 +359,11 @@ public class ReactiveDocumentResourceV2 {
           @NotBlank(message = "payload must not be empty")
           String payload,
       @ApiParam(
-              value = "Include this to make the TTL match that of the parent document.",
+              value =
+                  "Include this to make the TTL match that of the parent document. Requires read-before-write if set to true",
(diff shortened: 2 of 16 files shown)


restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -110,6 +110,39 @@ public ReactiveDocumentService(
...
+    try {
+      Integer ttlValue = Integer.parseInt(ttlRaw);
+      if (subPath.isEmpty()) {
+        return ttlValue;
+      } else {
+        throw new ErrorCodeRuntimeException(ErrorCode.DOCS_API_INVALID_TTL_NUMBER);
+      }
+    } catch (NumberFormatException e) {
+      if (!subPath.isEmpty()) {
+        if (!ttlRaw.equalsIgnoreCase("auto")) {
```
tatu-at-datastax, line 134: I would suggest explicitly looking for "auto" first since exceptions are rather expensive things to throw -- so creating and catching `NumberFormatException`s can be wasteful. More efficient to check for special case(s) first.
Same goes for empty String actually, probably better to check those 2 before attempting number parsing.

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -120,6 +123,32 @@ public Keyspace getKeyspace(String keyspaceName) {
...
+          .map(row -> row.getInt(String.format("ttl(%s)", DocsApiConstants.LEAF_COLUMN_NAME)))
+          .filter(Objects::nonNull)
+          .mapToInt(Integer::valueOf)
+          .max()
+          .getAsInt();
+    } catch (NoSuchElementException e) {
+      return null;
+    }
+  }
+
```
ivansenic, line 151: We should avoid adding any new things in the `DocumentDB` and rather as such functionality to the `DocumentSearchService`.. I am also fine to have a separate `DocumentTTLService`, but point is that this is implemented in the same way as we do it for the document read:

* dedicated query builder
* tests for query builder
* reactive service implementation that does separate prepare and execute, and returns a Single<Integer> (can be empty in case there is no ttl)
* test for that service implementation

restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java
```diff
@@ -288,25 +317,6 @@ public void tableExists(String keyspaceName, String tableName) {
...
-        .prepare(
-            dataStore
-                .queryBuilder()
-                .insertInto(info.getKeyspace(), info.getTable())
-                .value(markers)
-                .timestamp()
-                .build())
-        .join();
-  }
-
```
ivansenic, line 309: good catch! This is exactly what I am talking about, removing peace by peace from the `DocumentDB`.. For V2 we need to do this..

restapi/src/main/java/io/stargate/web/docsapi/resources/ReactiveDocumentResourceV2.java
```diff
@@ -391,6 +424,12 @@ public void patchDoc(
           @NotNull(message = "payload not provided")
           @NotBlank(message = "payload must not be empty")
           String payload,
+      @ApiParam(
+              value = "Include this to put a time-to-live on the document in the data store",
+              required = false)
+          @QueryParam("ttl")
+          @Min(value = 1, message = "TTL value must be a positive integer.")
+          Integer ttl,
```
ivansenic, line 432: wait this is patch, you can not submit the TTL here? we need a different parameter like `ttl-auto=true`, right?

restapi/src/main/java/io/stargate/web/docsapi/resources/ReactiveDocumentResourceV2.java
```diff
@@ -319,6 +346,12 @@ public void putDocPath(
           @NotNull(message = "payload not provided")
           @NotBlank(message = "payload must not be empty")
           String payload,
+      @ApiParam(
+              value =
+                  "Include this to put a time-to-live on the document in the data store; use 'auto' on a sub-document to make the TTL match that of the parent document.",
+              required = false)
+          @QueryParam("ttl")
+          String ttl,
```
ivansenic, line 354: let's have a different name for the param and make it a boolean, see other comment

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -125,6 +158,7 @@ public ReactiveDocumentService(
       String namespace,
       String collection,
       String payload,
+      Integer ttl,
```
ivansenic, line 161: add java doc

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -305,9 +344,11 @@ public ReactiveDocumentService(
       String collection,
       String documentId,
       String payload,
-      ExecutionContext context) {
+      String ttl,
```
ivansenic, line 347: please no string! Here integer is fine, this is where we allow the ttl to be specified..

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -330,7 +372,10 @@ public ReactiveDocumentService(
       String documentId,
       List<String> subPath,
       String payload,
-      ExecutionContext context) {
+      String ttl,
```
ivansenic, line 375: same here.. as we need to handle both update of the whole document with ttl int and the ttl auto I propse:

1. `updateDocument` accepts ttl int and calls `updateDocumentInternal`
2. add `updateSubDocument` with ttl-auto boolean and if it's `true` resolves the ttl and calls the `updateDocumentInternal` with the value.
3. this method is renamed to `updateDocumentInternal` and accepts only ttl as int

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -381,9 +428,11 @@ public ReactiveDocumentService(
       String collection,
       String documentId,
       String payload,
-      ExecutionContext context) {
+      String ttl,
```
ivansenic, line 431: patch only accepts ttl-auto boolean, right?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -407,7 +457,10 @@ public ReactiveDocumentService(
       String documentId,
       List<String> subPath,
       String payload,
-      ExecutionContext context) {
+      String ttl,
```
ivansenic, line 460: same here

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -1036,9 +1090,11 @@ public ReactiveDocumentService(
       String keyspace,
       String collection,
       String id,
+      String ttl,
```
ivansenic, line 1093: Integer

ivansenic (changes requested):
We need more clear service API and execution paths.. Main thing is to go away from the ttl as string and split into two params.. There is a bunch of needed changes here, so take your time.. if you wish we can split this into multiple PRs, if this is easier for you, for example:

1. PR-1 extension of the query builder and the DocumentWriteService to support TTL
2. PR-2 add service to provide current TTL for a document (current DocumentDB impl, but see suggestion)
3. PR-3 rest of the needed changes, HTTP API and reactive document service

This would also make it easier to review, WDYT?

EricBorczuk:
Instead of the separation you suggested, I will be putting the bulk of this step:
```PR-2 add service to provide current TTL for a document (current DocumentDB impl, but see suggestion)```

In a separate PR.

File: restapi/src/main/java/io/stargate/web/docsapi/dao/DocumentDB.java [hunk 12]
```diff
@@ -120,6 +123,32 @@ public Keyspace getKeyspace(String keyspaceName) {
     return dataStore.schema().keyspaces();
   }
 
+  public Integer getTtlForDocument(String namespace, String collection, String docId) {
```
EricBorczuk, line 126: This gets removed in https://github.com/stargate/stargate/pull/1684

EricBorczuk:
Of these changes:

PR-1 extension of the query builder and the DocumentWriteService to support TTL
PR-2 add service to provide current TTL for a document (current DocumentDB impl, but see suggestion)
PR-3 rest of the needed changes, HTTP API and reactive document service

#1 and #3 are fixed in this PR! Please review and I'll hold off merging https://github.com/stargate/stargate/pull/1684 to not clutter things

restapi/src/main/java/io/stargate/web/docsapi/service/query/search/db/impl/DocumentTtlQueryBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Supplier;
+
+/** Simple query builder to get the TTL of a document. */
+public class DocumentTtlQueryBuilder extends AbstractSearchQueryBuilder {
+
+  /** Doc id to target. */
+  private final String documentId;
```
ivansenic, line 40: this is now not needed, also why Codacy complains..

restapi/src/main/java/io/stargate/web/docsapi/resources/ReactiveDocumentResourceV2.java
```diff
@@ -319,6 +358,11 @@ public void putDocPath(
           @NotNull(message = "payload not provided")
           @NotBlank(message = "payload must not be empty")
           String payload,
+      @ApiParam(
+              value = "Include this to make the TTL match that of the parent document.",
+              required = false)
+          @QueryParam("ttl-auto")
+          Boolean ttlAuto,
```
ivansenic, line 365: should we also make it a `boolean` here and not the object bool?

restapi/src/main/java/io/stargate/web/docsapi/resources/ReactiveDocumentResourceV2.java
```diff
@@ -319,6 +358,11 @@ public void putDocPath(
           @NotNull(message = "payload not provided")
           @NotBlank(message = "payload must not be empty")
           String payload,
+      @ApiParam(
+              value = "Include this to make the TTL match that of the parent document.",
```
ivansenic, line 362: can we add that this requires read before write if `true`.. Also to the other methods with `ttlAuto`

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -305,9 +346,44 @@ public ReactiveDocumentService(
...
+   * @return Document response wrapper containing the generated ID.
+   */
+  public Single<DocumentResponseWrapper<Void>> updateSubDocument(
+      DocumentDB db,
+      String namespace,
+      String collection,
+      String documentId,
+      List<String> subPath,
+      String payload,
+      Boolean ttlAuto,
```
ivansenic, line 375: same here should we just have a `boolean`?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -381,9 +463,10 @@ public ReactiveDocumentService(
       String collection,
       String documentId,
       String payload,
+      Integer ttl,
```
ivansenic, line 466: OK this is a mistake, how can a patch have a ttl as int? should be a `ttlAuto`? This also means you don't have to introduce that new `patchSubDocument` it can still be one method.. Right?

restapi/src/test/java/io/stargate/web/docsapi/service/ReactiveDocumentServiceTest.java
```diff
@@ -197,13 +197,14 @@ public void happyPath() throws Exception {
               eq(collection),
```
ivansenic, line 197: This class is missing 5 tests updates for the TTL:

* write with ttl
* write batch with ttl
* update with ttl
* update sub-path with ttl auto
* patch with ttl auto

ivansenic (changes requested):
Looks good, way better than on the first review.. I left some minor comments for improvements, should be easy to fix.. However, we miss bunch of tests for the reactive document service..

ivansenic:
@EricBorczuk small ping here, not sure if you saw my review from last week.. See no progress here..

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
EricBorczuk: I believe this method is entirely unused, that's why I missed it. Deleting it in favor of the sub-document impl and renaming that method

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -442,31 +536,13 @@ public ReactiveDocumentService(
...
-  public Single<Boolean> deleteDocument(
-      DocumentDB db,
-      String namespace,
-      String collection,
-      String documentId,
-      ExecutionContext context) {
-    List<String> subPath = Collections.emptyList();
-    return deleteDocument(db, namespace, collection, documentId, subPath, context);
-  }
-
```
ivansenic, line 469: wait why is this deleted now?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -1040,15 +1116,19 @@ public ReactiveDocumentService(
...
+                writeService.updateDocument(
+                    db.getQueryExecutor().getDataStore(),
+                    keyspace,
+                    collection,
+                    id,
+                    processedPath,
+                    rows,
+                    ttl,
+                    db.treatBooleansAsNumeric(),
+                    context));
```
ivansenic, line 1131: is this additional fix for the push function? :+1:

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -110,13 +112,46 @@ public ReactiveDocumentService(
...
+  private Single<Integer> determineTtl(
+      DocumentDB db,
+      String namespace,
+      String collection,
+      String documentId,
+      ExecutionContext context) {
+
+    return Single.defer(
+        () -> {
+          authorizeRead(db, namespace, collection);
```
ivansenic, line 124: I am starting to think if this is needed? Means if you have only write permissions, you can not actually do update or patch with the TTL auto? Is this correct?

restapi/src/main/java/io/stargate/web/docsapi/service/query/DocumentSearchService.java
```diff
@@ -180,6 +181,16 @@
         .take(1);
   }
 
+  public Flowable<RawDocument> getDocumentTtlInfo(
```
ivansenic, line 184: Java docs missing here.. Not sure if this was added or I missed this..

ivansenic (approved):
Great stuff.. Have a look on a few remaining comments, especially that authz read on determine TTL..

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
EricBorczuk: The method is never used...all calls currently are using line 467's `deleteDocument` directly

restapi/src/main/java/io/stargate/web/docsapi/service/query/DocumentSearchService.java
EricBorczuk: Yea this existed in previous reviews, must have just missed javadoc. Will add

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
EricBorczuk: Oh that's a good point...Yeah maybe this shouldn't be here then (even though it's technically a read, people should be able to update/patch regardless)

PR #1671, opened 2022-03-01
Populate client info on all queries

jeffreyscarpenter:
This fixes an issue discovered in internal testing that was causing queries that should fail with UNAUTHORIZED error to succeed with some authentication providers.
 
**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/grpc/src/main/java/io/stargate/grpc/service/interceptors/NewConnectionInterceptor.java
+++ b/grpc/src/main/java/io/stargate/grpc/service/interceptors/NewConnectionInterceptor.java
@@ -129,23 +129,20 @@ public class NewConnectionInterceptor implements ServerInterceptor {
         authenticationService.validateToken(info.token(), info.headers());
 
     AuthenticatedUser user = authenticationSubject.asUser();
-    Connection connection;
-    if (!user.isFromExternalAuth()) {
-      SocketAddress remoteAddress = info.remoteAddress();
-      // This is best effort attempt to set the remote address, if the remote address is not the
-      // correct type then use a dummy value. Note: `remoteAddress` is almost always a
-      // `InetSocketAddress`.
-      InetSocketAddress inetSocketAddress =
-          remoteAddress instanceof InetSocketAddress
-              ? (InetSocketAddress) remoteAddress
-              : DUMMY_ADDRESS;
-      connection = persistence.newConnection(new ClientInfo(inetSocketAddress, null));
-    } else {
-      connection = persistence.newConnection();
-    }
+
+    SocketAddress remoteAddress = info.remoteAddress();
+    // This is best effort attempt to set the remote address, if the remote address is not the
+    // correct type then use a dummy value. Note: `remoteAddress` is almost always a
+    // `InetSocketAddress`.
+    InetSocketAddress inetSocketAddress =
+        remoteAddress instanceof InetSocketAddress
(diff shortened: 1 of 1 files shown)


grpc/src/main/java/io/stargate/grpc/service/interceptors/NewConnectionInterceptor.java
```diff
@@ -129,23 +129,20 @@ private Connection newConnection(RequestInfo info) throws UnauthorizedException
...
+    InetSocketAddress inetSocketAddress =
+        remoteAddress instanceof InetSocketAddress
+            ? (InetSocketAddress) remoteAddress
+            : DUMMY_ADDRESS;
+    ClientInfo clientInfo = new ClientInfo(inetSocketAddress, null);
+    Connection connection = persistence.newConnection(clientInfo);
     connection.login(user);
     if (user.token() != null) {
-      connection.clientInfo().ifPresent(c -> c.setAuthenticatedUser(user));
+      clientInfo.setAuthenticatedUser(user);
```
mpenick, line 145: This is a really good find.  If I remember right, `ClientInfo` was optional at one point, but that's no longer that case. It looks like `AuthenticateUser` information passed to `ClientInfo.setAuthenticationUser()` is serialized and pass a a custom payload through the CQL layer of the persistence backend. This is then used by `StargateQueryHandler#authorizeByToken()` to validate.

mpenick: Allowing all operations when the authenticated user (token) is not set is not good. The persistence layer shouldn't allow a the authenticated user to not be set when there's an authz service configured. That doesn't belong in this PR so I created a follow up issue: https://github.com/stargate/stargate/issues/1673

mpenick (approved):

grpc/src/main/java/io/stargate/grpc/service/interceptors/NewConnectionInterceptor.java
tatu-at-datastax: I assume it is not problematic necessarily, but just to make sure: @olim7t how about the "admin token" usage with schema sync -- APIs/services needing to access schema metadata during background population of caches, without specific authenticated user?

olim7t: Right, the bridge admin token is taking advantage of that: https://github.com/stargate/stargate/blob/v2.0.0/bridge/src/main/java/io/stargate/bridge/service/interceptors/BridgeNewConnectionInterceptor.java#L20-L23

So if we close the loophole we'd have to special-case the admin token somewhere else.

olim7t: PS- I think this is also the case in v1 whenever we call `DataStoreFactory.createInternal()`.

olim7t (approved):

grpc/src/main/java/io/stargate/grpc/service/interceptors/NewConnectionInterceptor.java
tatu-at-datastax: It may be the case then that we need to document things better to ensure all usage is aware of security consequences of such "admin" connections.

PR #1681, opened 2022-03-07
Fix #1680: update Jackson dependency to 2.12.6

### LINKED ISSUE #1680: Update Jackson dependency to 2.12.6 (from 2.10 and others)
Based on latest reports from Snyk, it looks like we should unify and upgrade dependencies to Jackson versions.
The main version Stargate components like `restapi` seems to use is 2.10.5 which is ok (as long as it's beyond 2.9 we won't get tons of CVEs), but based on my work on StargateV2 we should have no problem upgrading to the latest 2.12 patch. And while we could go all the way 2.13.2 (released yesterday) it seems safer to go with something that has been around for a while and has no known vulnerabilities. Hence 2.12.6.

tatu-at-datastax:
**What this PR does**:

Adds managed dependency to `jackson-bom` version 2.12.6, which will default Jackson dependencies to that version throughout project.

No issue filed; Snyk reported issues wrt Jackson 2.10.5 on another project.

**Checklist**
- [x] Changes manually tested
- [ ] Automated Tests added/updated -- will rely on existing CI

### DIFF
--- a/config-store-yaml/pom.xml
+++ b/config-store-yaml/pom.xml
@@ -28,7 +28,6 @@
       <dependency>
         <groupId>com.fasterxml.jackson.dataformat</groupId>
         <artifactId>jackson-dataformat-yaml</artifactId>
-        <version>2.11.4</version>
       </dependency>
       <dependency>
         <groupId>com.github.ben-manes.caffeine</groupId>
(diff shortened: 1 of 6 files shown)


tatu-at-datastax:
Ugh. Failure, right, but looks like GCB log access UI has changed. Not clear how to actually download FULL logs any more.

tatu-at-datastax:
Was able to figure out the failing test; need to get #1686 merged first as a pre-req -- there was an actual bug exposed by (but not caused by) version upgrade.

ivansenic (approved):
Great :+1:

PR #1683, opened 2022-03-08
(WIP) Alternative Jackson 2.12.6 upgrade PR

### LINKED ISSUE #1680: Update Jackson dependency to 2.12.6 (from 2.10 and others)

tatu-at-datastax:
DO NOT MERGE!

**What this PR does**:

Upgrades most of Jackson deps to 2.12.6 -- but unlike earlier PR #1681 leaves overrides for one module to see if that avoids test fails.

Fixes #1680

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/config-store-yaml/pom.xml
+++ b/config-store-yaml/pom.xml
@@ -28,7 +28,6 @@
       <dependency>
         <groupId>com.fasterxml.jackson.dataformat</groupId>
         <artifactId>jackson-dataformat-yaml</artifactId>
-        <version>2.11.4</version>
       </dependency>
       <dependency>
         <groupId>com.github.ben-manes.caffeine</groupId>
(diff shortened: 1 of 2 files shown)


tatu-at-datastax:
Ok, based on work on the main PR, I think this will also fail with one test -- problem actual, not just with tests.

PR #1686, opened 2022-03-08
Fix #1685 handling of `timestamp` values on REST/json output

### LINKED ISSUE #1685: REST API serializes `timestamp` values as POJOs: should be ISO-8601
(note: found when working on #1680 

So: looks like `timestamp` fields are serialized as POJOs because by default Jackson does not recognize Java 8 `Instant` values as anything special. So you'd see something like:

```
{"epochSecond":1619203342, "nano":139000000}
```

Unfortunately there is no test coverage, it seems.
But it is possible to modify `RestApiv2RowsTest.getRowsWithTimestampQuery()` to cover this case.

I think I will try to resolve this issue first before proceeding with #1680.

tatu-at-datastax:
**What this PR does**:

Fixes json serialization of `timestamp` values: instead of meaningless JSON objects will be serialized as ISO-8601 Strings. Adds minor test extension to verify the fix; formerly no tests seem to verify actual value that comes back.

Fixes #1685

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/restapi/pom.xml
+++ b/restapi/pom.xml
@@ -52,6 +52,14 @@
       <groupId>io.dropwizard</groupId>
       <artifactId>dropwizard-core</artifactId>
     </dependency>
+    <!-- Since restapi serializes Java 8 date/time values, we must use
+         JSR-310 module. Version provided by parent.
+      -->
+    <dependency>
+      <groupId>com.fasterxml.jackson.datatype</groupId>
+      <artifactId>jackson-datatype-jsr310</artifactId>
+      <version>2.10.5</version>
+    </dependency>
     <dependency>
       <groupId>com.github.java-json-tools</groupId>
       <artifactId>json-schema-validator</artifactId>
--- a/restapi/src/main/java/io/stargate/web/resources/Converters.java
+++ b/restapi/src/main/java/io/stargate/web/resources/Converters.java
@@ -25,6 +25,8 @@ import com.datastax.oss.driver.shaded.guava.common.collect.Lists;
 import com.datastax.oss.driver.shaded.guava.common.collect.Maps;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import io.stargate.db.datastore.Row;
 import io.stargate.db.query.Modification.Operation;
 import io.stargate.db.query.Predicate;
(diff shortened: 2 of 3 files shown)


restapi/pom.xml
```diff
@@ -52,6 +52,14 @@
       <groupId>io.dropwizard</groupId>
       <artifactId>dropwizard-core</artifactId>
     </dependency>
+    <!-- Since restapi serializes Java 8 date/time values, we must use
+         JSR-310 module. Version provided by parent.
+      -->
+    <dependency>
+      <groupId>com.fasterxml.jackson.datatype</groupId>
+      <artifactId>jackson-datatype-jsr310</artifactId>
+      <version>2.10.5</version>
```
tatu-at-datastax, line 61: Only temporarily specify here (will follow up with another PR); for now 2.10.5 is used via DropWizard.

restapi/src/main/java/io/stargate/web/resources/Converters.java
```diff
@@ -72,6 +74,14 @@
 public class Converters {
 
   private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+  static {
+    // For Java 8 date/time types we need:
+    OBJECT_MAPPER.registerModule(new JavaTimeModule());
```
tatu-at-datastax, line 80: Without this `java.time.Instant` becomes that odd JSON Object which can not be deserialized (nor easily used from Java). Module adds explicit serializer, deserializer.

restapi/src/main/java/io/stargate/web/resources/Converters.java
```diff
@@ -72,6 +74,14 @@
...
+    // and by default date/time values written as numbers but we prefer ISO-8601:
+    OBJECT_MAPPER.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
```
tatu-at-datastax, line 82: But even with module, by default timestamp values are serialized as numbers; in this case an odd floating-point number. This is rarely what is wanted so change to use textual formatting, ISO-8601 value that is format being sent in the first place.

testing/src/main/java/io/stargate/it/http/RestApiv2RowsTest.java
```diff
@@ -597,6 +597,7 @@ public void getRowsWithTimestampQuery() throws IOException {
     assertThat(data.size()).isEqualTo(1);
     assertThat(data.get(0).get("id")).isEqualTo("1");
     assertThat(data.get(0).get("firstName")).isEqualTo("John");
+    assertThat(data.get(0).get("created")).isEqualTo(timestamp);
```
tatu-at-datastax, line 600: And here's the minimal change to verify we get value we expect back. Ideally would have more tests but currently there are no existing tests to augment so this should suffice on short term.

tatu-at-datastax:
Transient failure for C*4.0 test `ExecuteBatchStreamingTest.manyStreamingBatch` (fragile), will re-run.
Should be passing otherwise.

tatu-at-datastax:
Passed with retry.

jeffreyscarpenter (approved):
This is a good change and consistent with how Cassandra ecosystem tools (e.g. `cqlsh`) typically handle timestamps.

PR #1688, opened 2022-03-09
Fix eviction race for prepared statements on Cassandra 4.0

### LINKED ISSUE #1679: `ExecuteBatchStreamingTest` fails sporadically for the C* 4.0
Unfortunately, the gRPC test `ExecuteBatchStreamingTest.manyStreamingBatch` fails sporadically even after merging the https://github.com/stargate/stargate/pull/1661. Seems that two retries are not enough and that we can hit the situation that both retries also fail. 
I don't see any other solution, than to increase the number of retries. We might leave it being limited, but maybe 2 is too less.

mpenick:
CASSANDRA-15252 (and CASSANDRA-17248) changed the behavior of prepared
statements which can cause multiple prepares of the same query to evict
each other from the prepared cache. This manifest itself as
`PreparedQueryNotFoundException` exceptions when attempting to execute a
prepared statement that was just prepared then evicted by another
concurrent prepare.

This fix serializes prepares for similar queries by locking them;
allowing one prepare to finish completely to avoid the case where
simultaneous prepares evict each other. It uses lock striping to
hopefully avoid making prepares too slow for concurrent preparers.

Fixes #1679

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Cassandra40Persistence.java
+++ b/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Cassandra40Persistence.java
@@ -7,6 +7,7 @@ import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
 import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableMap;
 import com.datastax.oss.driver.shaded.guava.common.collect.Iterables;
 import com.datastax.oss.driver.shaded.guava.common.util.concurrent.Uninterruptibles;
+import com.google.common.util.concurrent.Striped;
 import io.stargate.auth.AuthorizationService;
 import io.stargate.core.util.TimeSource;
 import io.stargate.db.Authenticator;
@@ -38,6 +39,7 @@ import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Lock;
 import java.util.function.Supplier;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Cassandra40Persistence.java
```diff
@@ -500,12 +509,22 @@ protected void loginInternally(io.stargate.db.AuthenticatedUser user) {
...
-          parameters,
-          // The queryStartNanoTime is not used by prepared message, so it doesn't really matter
-          // that it's only computed now.
-          System.nanoTime(),
-          () -> new PrepareMessage(query, parameters.defaultKeyspace().orElse(null)));
+      // The behavior of prepared statements was changed as part of CASSANDRA-15252
+      // (and CASSANDRA-17248) which can cause multiple prepares of the same query to evict each
+      // other from the prepared cache. A few Stargate APIs eagerly prepare queries which
+      // increases this chance and to avoid this we serialize prepares for similar queries.
+      Lock lock = PREPARE_LOCKS.get(parameters.defaultKeyspace().map(k -> query + k).orElse(query));
```
tatu-at-datastax, line 516: So this calculates hash for striped locks from query string, qualified by default namespace, if any?

Although purely for bucketing of potentially conflicting queries into same bucket, wouldn't it be enough to only consider `query` and ignore default keyspace? Should still use locking where it needs to be, unlikely to cause less congestion, and avoids String concatenation? So:

```
Lock lock = PREPARE_LOCKS.get(query)
```

Not a big deal but just wanted to see if my understanding is correct.

tatu-at-datastax (approved):

persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Cassandra40Persistence.java
mpenick: Yeah, i think that make sense.  It would only become contentious when the same query strings are prepared in several different keyspaces. Which is probably not likely? I'm not sure.  I'm fine with just choosing the bucket based on `query`.  

It might be worth avoiding the concatenate for the common case to avoid some contention on the less likely case,

tatu-at-datastax: I have no data to back this up so I am just speculating.

I am ok with using concatenation too, not a big deal, so it's up to you.

PR #1689, opened 2022-03-09
Update starctl scripts for v2

jeffreyscarpenter:
Rethinking the approach I had originally executed for start scripts for v2. 

Original v2 approach: new script `start_all_local.sh` starts `starctl` (coordinator) and `starctl-service-rest` (new script to start rest service)

Updated v2 approach: `starctl` is the primary script which starts `starctl-coordinator` and `starctl-service-rest`

In the updated approach, `starctl` basically replaces what start_all_local.sh was doing, and with more reliable management of the forked processes).

Docker images updated as required.

Change to the v2 release script will need to be made on master as well.

**Checklist**
- [x] Changes manually tested


starctl
```diff
@@ -1,177 +1,27 @@
...
-        STARGATE_ARGS+=("--disable-mbean-registration")
-    fi
+if [ -z $BRIDGE_TOKEN ]; then
+  export BRIDGE_TOKEN=mockAdminToken
+fi
 
-    if [[ ! -z "$DISABLE_BUNDLES_WATCH" ]]; then
-        STARGATE_ARGS+=("--disable-bundles-watch")
-    fi
+# start REST API service
```
tatu-at-datastax, line 20: Would it make sense to start by starting coordinator, before APIs/Services? Since latter require former.
Probably wouldn't make much sense as both take time to spin up but could reduce time for APIs need to wait for coordinator.

tatu-at-datastax (approved):

tatu-at-datastax:
Makes sense: I like the idea of parent process, modular "child" processes. Should be easier to compose for setups with a subset of APIs/services too.

starctl
jeffreyscarpenter: On the one hand: The API services should be robust enough to handle things correctly if the coordinator is not present (i.e return 503 if no coordinator available?), which is why I didn't bother to try to put in logic for this. 

On the other hand: might be good to fork `starctl-coordinator` first with a short delay to make sure itstarts up correctly. Traditionally the `starctl` (now `starctl-coordinator`) has required parameters. If you just type `./starctl` and don't provide any parameters, it will exit with an error message but the REST service is started anyway. In this case you wouldn't be left with an orphan process since `starctl` does a `wait` on the forked processes and kills them if it is killed, but the error message from `starctl-coordinator` could be obscured.

tatu-at-datastax: Yes, I 100% agree that Services need to be robust enough to handle outages, including start up delay.

Also seems like configurable optional delay(s) would be a nice (future?) improvement.

Good point about challenges wrt piping of failures -- would be also great, if possible, to catch some sanity-check style startup failure cases wrt coordinator.

tatu-at-datastax:
@jeffreyscarpenter should this be merged now?

PR #1690, opened 2022-03-11
gRPC: add support for CQL duration type

olim7t:
**What this PR does**:
Add `CqlDuration` type to represent durations on the client. 
Add missing bits to bind duration values and read duration columns.

Note: this is based on v2.0.0 because I started it in the context of GraphQL v2, but it can be backported to v1.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/grpc-proto/src/main/java/io/stargate/grpc/VIntCoding.java
+++ b/grpc-proto/src/main/java/io/stargate/grpc/VIntCoding.java
@@ -1,0 +1,181 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+// Protocol Buffers - Google's data interchange format
+// Copyright 2008 Google Inc.  All rights reserved.
+// https://developers.google.com/protocol-buffers/
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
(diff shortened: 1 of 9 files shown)


grpc-proto/src/main/java/io/stargate/grpc/Values.java
```diff
@@ -100,6 +100,10 @@ public static Value of(LocalTime value) {
     return Value.newBuilder().setTime(value.toNanoOfDay()).build();
   }
 
+  public static Value of(CqlDuration duration) {
+    return Value.newBuilder().setDuration(ByteString.copyFrom(duration.encode())).build();
```
olim7t, line 104: The duration is encoded directly on the client.
Alternatively, we could use a protobuf message with `months`, `days` and `nanos` fields, but since we'll need this encoding when we eventually send the value to Cassandra, I figured we might as well pass it all the way through.

tatu-at-datastax:
Cass-4.0 build was failing, probably/possibly due to that flaky test; triggered retry.

grpc-proto/src/main/java/io/stargate/grpc/Values.java
mpenick: That's less usable from other languages/platforms, but if this is an internal API maybe it's not a big deal.

tatu-at-datastax: I had the same concern: if it is not exposed to clients (i.e. gRPC public access is not affected) it's fine to have as implementation detail. But if decoding/encoding is needed everywhere, a simple scheme might be preferable.

But these types are shared between public gRPC and Bridge aren't they?

grpc-proto/src/main/java/io/stargate/grpc/VIntCoding.java
```diff
@@ -0,0 +1,181 @@
...
+
+  private static int numberOfExtraBytesToRead(int firstByte) {
+    // we count number of set upper bits; so if we simply invert all of the bits, we're golden
+    // this is aided by the fact that we only work with negative numbers, so when upcast to an int
+    // all
+    // of the new upper bits are also set, so by inverting we set all of them to zero
+    return Integer.numberOfLeadingZeros(~firstByte) - 24;
+  }
+
+  private static final ThreadLocal<byte[]> encodingBuffer =
```
tatu-at-datastax, line 113: Is it worth having this for a tiny 9 byte buffer? Yes, it reduces churn a bit but not sure it's a net win due to need for `ThreadLocal` access.

mpenick: If you do `final byte[] enocodingBuffer = new byte[9]` locally and pass it into `encodeVInt()` maybe it'll inline that method then escape analysis will make that a stack alloc? I don't know how to go about analyzing that. :)

tatu-at-datastax: Even without that I think that allocation (and GC) of 9-byte array is trivially small thing, not worth worrying about unless in a tight loop or something.
In fact this is kind of thing that I think really should not be optimized unless it shows up in profiling, heap dump or something.

olim7t: Yeah that seems a bit overkill, I've removed it.

tatu-at-datastax (approved):

PR #1698, opened 2022-03-16
gRPC: Revisit index metadata

olim7t:
**What this PR does**:
Revisit index metadata to include all information available from the persistence.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/bridge/src/main/java/io/stargate/bridge/service/SchemaHandler.java
+++ b/bridge/src/main/java/io/stargate/bridge/service/SchemaHandler.java
@@ -21,6 +21,7 @@ import io.grpc.StatusException;
 import io.grpc.stub.StreamObserver;
 import io.stargate.db.Persistence;
 import io.stargate.db.query.builder.Replication;
+import io.stargate.db.schema.CollectionIndexingType;
 import io.stargate.db.schema.Column;
 import io.stargate.db.schema.Index;
 import io.stargate.db.schema.Keyspace;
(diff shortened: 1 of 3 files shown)


grpc-proto/proto/schema.proto
```diff
@@ -32,11 +32,23 @@ enum ColumnOrderBy {
 
 // INDEXES
 
+enum IndexingType {
+  DEFAULT = 0;
```
olim7t, line 36: The indexing type is optional. I opted for a default "empty" value rather than adding a wrapper message (not even sure if protobuf provides an equivalent of `StringValue` for enums out of the box).

tatu-at-datastax (approved):

bridge/src/main/java/io/stargate/bridge/service/SchemaHandler.java
```diff
@@ -174,11 +176,25 @@ private static CqlTable buildCqlTable(Table table) throws StatusException {
   }
 
   private static CqlIndex buildSecondaryIndex(SecondaryIndex index) {
-    return CqlIndex.newBuilder()
-        .setName(index.name())
-        .setColumnName(index.column().name())
-        .setCustomType(StringValue.newBuilder().setValue(index.indexTypeName()).build())
```
olim7t, line 180: `index.indexTypeName()` is hard-coded to "Secondary index". That's not very useful in gRPC since the client already knows that it's dealing with a `CqlIndex` (as opposed to a materialized view or table).
I removed the `customType` field completely.

grpc-proto/proto/schema.proto
tatu-at-datastax: +1 for explicit value, given that we don't have to worry about backwards compatibility yet.

jeffreyscarpenter (approved):

PR #1703, opened 2022-03-18
Prepare common components for GraphQL v2

olim7t:
**What this PR does**:
Extract the generic parts of #1651 for easier reviews.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/grpc-proto/src/main/java/io/stargate/grpc/BytesValues.java
+++ b/grpc-proto/src/main/java/io/stargate/grpc/BytesValues.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.grpc;
+
+import com.google.protobuf.ByteString;
+import com.google.protobuf.BytesValue;
+import java.nio.ByteBuffer;
+import java.util.Base64;
+
+public class BytesValues {
+
+  public static BytesValue of(byte[] bytes) {
+    return BytesValue.of(ByteString.copyFrom(bytes));
+  }
+
+  public static BytesValue of(ByteBuffer byteBuffer) {
+    return BytesValue.of(ByteString.copyFrom(byteBuffer));
+  }
+
+  public static BytesValue ofBase64(String base64) {
+    return of(Base64.getDecoder().decode(base64));
+  }
+
+  public static byte[] toBytes(BytesValue value) {
+    return value.getValue().toByteArray();
+  }
+
+  public static ByteBuffer toByteBuffer(BytesValue value) {
+    return value.getValue().asReadOnlyByteBuffer();
+  }
+
+  public static String toBase64(BytesValue value) {
+    return Base64.getEncoder().encodeToString(toBytes(value));
+  }
+
+  private BytesValues() {
+    // intentionally empty
+  }
+}
--- a/sgv2-service-common/src/main/java/io/stargate/sgv2/common/futures/Futures.java
+++ b/sgv2-service-common/src/main/java/io/stargate/sgv2/common/futures/Futures.java
@@ -15,9 +15,11 @@
  */
 package io.stargate.sgv2.common.futures;
 
+import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
+import java.util.stream.Collectors;
 
 public class Futures {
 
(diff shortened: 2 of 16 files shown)


olim7t:
It makes sense to add the async bridge methods to this (they're not used in CQL-first but will be used in schema-first). I've cherry-picked a commit that I had on a local branch.

grpc-proto/src/main/java/io/stargate/grpc/BytesValues.java
```diff
@@ -0,0 +1,52 @@
...
+ * limitations under the License.
+ */
+package io.stargate.grpc;
+
+import com.google.protobuf.ByteString;
+import com.google.protobuf.BytesValue;
+import java.nio.ByteBuffer;
+import java.util.Base64;
+
+public class BytesValues {
```
ivansenic, line 23: As I wrote in the original PR, maybe we should add a quick unit test for this..

sgv2-service-common/src/main/java/io/stargate/sgv2/common/futures/Futures.java
```diff
@@ -53,4 +56,20 @@
...
+
+  public static <T> CompletableFuture<T> failedFuture(Exception e) {
+    CompletableFuture<T> f = new CompletableFuture<>();
+    f.completeExceptionally(e);
+    return f;
+  }
+
+  /** Returns a future of all the results, or a failed future if any of the inputs fails. */
+  public static <T> CompletionStage<List<T>> sequence(List<CompletionStage<T>> futures) {
+    return CompletableFuture.allOf(futures.toArray(new CompletableFuture<?>[0]))
```
ivansenic, line 68: `new CompletableFuture<?>[futures.size()]`

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
```diff
@@ -76,120 +82,172 @@
...
+    ClientCalls.asyncUnaryCall(call, query, observer);
+    return observer
+        .asFuture()
+        .thenApply(
+            response -> {
+              if (response.hasSchemaChange()) {
+                // Ensure we're not holding a stale copy of the impacted keyspace by the time we
+                // return control to the caller
+                String keyspaceName = addTenantPrefix(response.getSchemaChange().getKeyspace());
+                schema.removeKeyspace(keyspaceName);
```
ivansenic, line 98: what kind of schema object this is? is it thread safe?

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
```diff
@@ -76,120 +82,172 @@
...
+  public CompletionStage<Optional<CqlKeyspaceDescribe>> getKeyspaceAsync(String keyspaceName) {
+    return authorizeSchemaReadAsync(SchemaReads.keyspace(keyspaceName, sourceApi))
+        .thenApply(
+            authorized -> {
+              if (!authorized) {
+                throw new UnauthorizedKeyspaceException(keyspaceName);
+              }
+              return authorized;
+            })
+        .thenCompose(__ -> getAuthorizedKeyspace(keyspaceName));
```
ivansenic, line 123: this can be shorthanded?

```
        .thenCompose(
            authorized -> {
              if (!authorized) {
                throw new UnauthorizedKeyspaceException(keyspaceName);
              }
              return getAuthorizedKeyspace(keyspaceName);
            })

File: sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java [hunk 5]
```diff
@@ -76,120 +82,172 @@
...
+                      .map(n -> SchemaReads.keyspace(n, sourceApi))
+                      .collect(Collectors.toList());
+              return authorizeSchemaReadsAsync(reads)
+                  .thenApply(authorizations -> removeUnauthorized(names, authorizations));
+            });
+    return authorizedNamesFuture.thenCompose(this::getAuthorizedkeyspaces);
+  }
+
+  private <T> List<T> removeUnauthorized(List<T> elements, List<Boolean> authorizations) {
+    assert elements.size() == authorizations.size();
```
ivansenic, line 150: throw proper exception and msg instead of `assert`?

ivansenic (approved):
Look good, however, seems there are no tests added? Was this forgotten or simply they are not written.. Seems we have a bunch of places that should be testes, especially the `DefaultStargateBridgeClient`?

I approved, but there are also tests failing.. Ping me if I need to have another look..

sgv2-service-common/src/main/java/io/stargate/sgv2/common/futures/Futures.java
olim7t: Pre-sizing triggers an IntelliJ inspection:
> There are two styles to convert a collection to an array:
> - A pre-sized array, for example, c.toArray(new String[c.size()])
> - An empty array, for example, c.toArray(new String[0])
>
> In older Java versions, using a pre-sized array was recommended, as the reflection call necessary to create an array of proper size was quite slow.
> However, since late updates of OpenJDK 6, this call was intrinsified, making the performance of the empty array version the same, and sometimes even better, compared to the pre-sized version. Also, passing a pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray calls. This may result in extra nulls at the end of the array if the collection was concurrently shrunk during the operation.

(the last part about concurrent collections does not apply here)

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
olim7t: I guess it depends how we decide to use (or not use) assertions... Here it's intended more as documenting an invariant of the method: `authorizations` is always the result of a call to `authorizeSchemaReadsAsync`, which returns a list of the same size as its argument. So different sizes cannot really happen, unless there is a programming error.
If I didn't use an assertion here, I would probably just remove the statement altogether.

PR #1704, opened 2022-03-21
updating version of nimbus-jose-jwt library

jeffreyscarpenter:
Update the version of nimbus-jose-jwt library used in the JWT-based authentication module from 9.0.1 to 9.21

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/auth-jwt-service/pom.xml
+++ b/auth-jwt-service/pom.xml
@@ -33,7 +33,7 @@
     <dependency>
       <groupId>com.nimbusds</groupId>
       <artifactId>nimbus-jose-jwt</artifactId>
-      <version>9.0.1</version>
+      <version>9.21</version>
     </dependency>
 
     <!-- Test dependencies -->


tatu-at-datastax (approved):

tatu-at-datastax:
Oh. I have a suggestion here; if added as a managed dependency, it would also affect C2 I think,
because `c2/pom.xml` has `io.stargate:stargate` as its parent pom.

Basically, look at:

```
  <dependencyManagement>
    <dependencies>
```

and you can see there are a few definitions which will add the expected version to use WITHOUT adding actual dependency. Actual dependency, when added, should then exclude version number.

So I would suggest doing this to get a more consistent set of versions.

jeffreyscarpenter:
> Oh. I have a suggestion here; if added as a managed dependency, it would also affect C2 I think, because `c2/pom.xml` has `io.stargate:stargate` as its parent pom.
> 
> Basically, look at:
> 
> ```
>   <dependencyManagement>
>     <dependencies>
> ```
> 
> and you can see there are a few definitions which will add the expected version to use WITHOUT adding actual dependency. Actual dependency, when added, should then exclude version number.
> 
> So I would suggest doing this to get a more consistent set of versions.

Can we do that in a follow up issue?

tatu-at-datastax:
> > Oh. I have a suggestion here; if added as a managed dependency, it would also affect C2 I think, because `c2/pom.xml` has `io.stargate:stargate` as its parent pom.
> > Basically, look at:
> > ```
> >   <dependencyManagement>
> >     <dependencies>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and you can see there are a few definitions which will add the expected version to use WITHOUT adding actual dependency. Actual dependency, when added, should then exclude version number.
> > So I would suggest doing this to get a more consistent set of versions.
> 
> Can we do that in a follow up issue?

Yes, let's do that.

PR #1706, opened 2022-03-21
Add `makePagingState` to grpc bridge

EricBorczuk:
**What this PR does**:
Adds makePagingState as a function to grpc bridge, allowing someone who has the keyspace, table, and column byte state of a row to make a paging state from it. I've been trying to find a way to manually test, but any advice on how to test would be welcome...

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
+++ b/bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
@@ -112,4 +112,11 @@ public class BridgeService extends StargateBridgeGrpc.StargateBridgeImplBase {
             request, GrpcService.CONNECTION_KEY.get(), authorizationService, responseObserver)
         .handle();
   }
+
+  @Override
+  public void makePagingState(
+      Schema.MakePagingStateParams request, StreamObserver<Schema.PagingState> responseObserver) {
+    PagingStateHandler.makePagingState(
+        request, GrpcService.CONNECTION_KEY.get(), persistence, responseObserver);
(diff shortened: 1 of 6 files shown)


ivansenic (changes requested):
I am confident we will not need this.. There is no reason why would we call per row the make paging state from the client side.. This is enormous overhead.. 

Instead, we would add a secondary `ExecuteQuery(Query)` that returns a different `Response`. That `Response` will not contain a set of Rows, but rather an extended version version, where beside a Row we have paging state and the comparable bytes.. This way all make page state and get comparable bytes is done on the bridge and there is still one network call per query.

I am sorry for the work invested, but the agreement was to investigate if this is possible, not create a full blown implementation and a PR. A bit of misunderstanding between @EricBorczuk and me..

EricBorczuk:
Yeah this one's my bad, I had misunderstood some of the discussions from the last week or so; I forgot that the paging state is eagerly generated per-row currently, and thought that instead it would be generated once on the last row that was touched from each query - agree that this would be a huge performance penalty as-is. At the very least I know now that generating the paging state on the bridge side is quite easy to do, for the future when we want to add the gRPC function that returns Rows decorated with paging state and comparable bytes. To be honest, I personally didn't find it to be entirely a waste because I got to learn about how everything is set up :) 

Closing this!

ivansenic:
It's good that we have the reference to the branch where this was done :+1:

PR #1711, opened 2022-03-22
Remove StargateBridgeSchema

olim7t:
**What this PR does**:
- remove `StargateBridgeSchema`. Schema requests are now executed by `StargateBridgeClient`, with the user's token.
- the client now queries the bridge for every `getKeyspace()` call, but we also added a checksum mechanism to optimize the response size when the client already has the latest version.
- make authorization optional for schema metadata operations (avoids an extra bridge call if we're going to do another authorized query right after)
- remove bridge "admin token".

TODO:
- [x] tune the size and TTL of the keyspace cache (in `DefaultStargateBridgeClientFactory`)
- [x] I think we should also remove the gRPC bridge operations that are no longer used: `DescribeTable` and `GetSchemaNotifications`.
- [x] add `CqlKeyspace.decoratedName` in the proto definition, and fill it


tatu-at-datastax:
Quick question/confirmation: so, a lot of "remove tenant prefix" calls are removed as unnecessary because gRPC calls are now per-tenant/tenant-aware? So handled by Bridge side, not client-side?

olim7t:
> Quick question/confirmation: so, a lot of "remove tenant prefix" calls are removed as unnecessary because gRPC calls are now per-tenant/tenant-aware? So handled by Bridge side, not client-side?

All bridge calls are now made in the context of a HTTP request, so they are always associated with a particular user, which has a tenant.

ivansenic:
> Note that all bridge requests are now made with a "tenant-aware" gRPC connection. This means we would also get rid of the bridge "admin token".

I don't think this is true.. I though you would use admin token and the tenant-id as the param for getting the schema.. How would you otherwise authorize on the bridge that tenant id belongs to the given auth token?

tatu-at-datastax:
> > Note that all bridge requests are now made with a "tenant-aware" gRPC connection. This means we would also get rid of the bridge "admin token".
> 
> I don't think this is true.. I though you would use admin token and the tenant-id as the param for getting the schema.. How would you otherwise authorize on the bridge that tenant id belongs to the given auth token?

I think this means there is the actual per-tenant auth token to use instead of global admin token, and this auth token is used to verify the match. So all calls are authenticated directly.

ivansenic:
> > > Note that all bridge requests are now made with a "tenant-aware" gRPC connection. This means we would also get rid of the bridge "admin token".
> > 
> > 
> > I don't think this is true.. I though you would use admin token and the tenant-id as the param for getting the schema.. How would you otherwise authorize on the bridge that tenant id belongs to the given auth token?
> 
> I think this means there is the actual per-tenant auth token to use instead of global admin token, and this auth token is used to verify the match. So all calls are authenticated directly.

But wait, we are talking about the `rpc DescribeKeyspace(DescribeKeyspaceQuery) returns (CqlKeyspaceDescribe) {}` and passing the token of the user?  There is a direct access to the persistenace and it's schema object, there is no additional authz on the keyspace you want to desribe done there, nor anything done with the connection object that was created using the tenant auth token.. Meaning any valid token can fetch any keyspace or? Or will the `decorateKeyspaceName` solve this and return only the tenant based keyspaces eventually?

Let's please clarify this, I am getting this wrong most likely..

olim7t:
> But wait, we are talking about the `rpc DescribeKeyspace(DescribeKeyspaceQuery) returns (CqlKeyspaceDescribe) {}` and passing the token of the user? There is a direct access to the persistenace and it's schema object, there is no additional authz on the keyspace you want to desribe done there, nor anything done with the connection object that was created using the tenant auth token.. Meaning any valid token can fetch any keyspace or? Or will the `decorateKeyspaceName` solve this and return only the tenant based keyspaces eventually?

When you pass a tenant id in the gRPC headers, the bridge decorates keyspace names automatically. See `SchemaHandler.describeKeyspace`.

Regarding authorization, it is indeed not done on the bridge, but on the client: for example `DefaultStargateBridgeClient.getkeyspaceAsync()`. This made sense when the bridge allowed "anonymous" access (via admin token), but if we remove that, then I'm starting to think we should move authorization to the bridge for consistency.

tatu-at-datastax:
Ok yes, Schema access does need to be somehow authenticated. Maybe this is something where "bridge access token" could still work -- less secure than true authentication but should reduce likelihood of a rogue service from scanning schema metadata?
Or would it be simple enough to just call regular authn like other Bridge endpoints?

ivansenic:
If we do implement auth for the get keyspace on the bridge, we should only do it when hash is not supplied, as we are not leaking anything when hash is there and data did not change.. 

Not sure what's the best though, this would mean that any op you do requires the `authorizeSchemaRead` for that token, which was not the case until now I think.. Until now we only validated the token.. This could be a breaking change imo! That's said, we might even expose this breaking change by calling the `rpc AuthorizeSchemaReads(AuthorizeSchemaReadsRequest) returns (AuthorizeSchemaReadsResponse)` from the client side..

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/StargateBridgeClient.java
```diff
@@ -60,17 +60,25 @@ default Response executeBatch(Batch batch) {
...
+   *     you build a client response directly from the metadata, then authorization should be done
+   *     by this method.
+   * @throws UnauthorizedKeyspaceException (wrapped in the returned future) if {@code
+   *     checkIfAuthorized} is set, and the client is not allowed to access this keyspace. However,
+   *     this check is shallow: if the caller surfaces keyspace elements (tables, UDTs...) to the
+   *     end user, it must authorize them individually with {@link #authorizeSchemaReads(List)}.
    */
-  CompletionStage<Optional<CqlKeyspaceDescribe>> getKeyspaceAsync(String keyspaceName);
+  CompletionStage<Optional<CqlKeyspaceDescribe>> getKeyspaceAsync(
+      String keyspaceName, boolean checkIfAuthorized);
```
olim7t, line 76: I ended up adding a parameter for authorization. Having a default was not intuitive either way, I like that this forces the caller to think about it.

bridge/src/main/java/io/stargate/bridge/service/SchemaHandler.java
```diff
@@ -49,27 +49,39 @@
...
-          persistence.decorateKeyspaceName(query.getKeyspaceName(), GrpcService.HEADERS_KEY.get());
-      Keyspace keyspace = persistence.schema().keyspace(decoratedKeyspace);
-      if (keyspace == null) {
-        throw Status.NOT_FOUND.withDescription("Keyspace not found").asException();
-      }
-      responseObserver.onNext(buildKeyspaceDescription(keyspace));
+    String decoratedName =
+        persistence.decorateKeyspaceName(query.getKeyspaceName(), GrpcService.HEADERS_KEY.get());
+
+    Keyspace keyspace = persistence.schema().keyspace(decoratedName);
```
olim7t, line 62: `CqlKeyspaceDescribe.CqlKeyspace.getName()` will be the decorated name (we don't strip the tenant prefix).
In a way this is a feature because the client will need the decorated name if it needs to maintain a global cache, like GraphQL does.

But actually it's probably better to have a separate `getDecoratedName()` field for that, I'll make the change.

olim7t (commented):
Updated with the findings of #1711 and other discussions.
This is good for review now.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
```diff
@@ -111,22 +100,66 @@
...
+    fetchKeyspaceFromBridge(keyspaceName, cachedHash)
+        .whenComplete(
+            (fetched, error) -> {
+              if (error != null) {
+                result.completeExceptionally(error);
+              } else if (fetched == null) {
+                // Keyspace does not exist. Update the cache if we previously thought it did.
+                if (cached != null) {
+                  keyspaceCache.invalidate(keyspaceName);
+                }
```
ivansenic, line 136: hmm are you sure you are not getting back the error on keyspace not existing, in my tests I saw that these calls do not end in `OK` status? Maybe you should move this to the `error` case?

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
```diff
@@ -253,29 +287,15 @@ private Channel addMetadata(Channel channel, String tenantId) {
...
-      return fullKeyspaceName.substring(tenantPrefix.length());
+  static class DescribeKeyspaceObserver extends UnaryStreamObserver<CqlKeyspaceDescribe> {
+    @Override
+    public void onError(Throwable t) {
+      if (t instanceof StatusRuntimeException
+          && ((StatusRuntimeException) t).getStatus().getCode() == Status.Code.NOT_FOUND) {
+        onNext(null);
+      } else {
+        super.onError(t);
+      }
```
ivansenic, line 298: ok so this is a hack.. not sure if this should not be directly handled in the handler, this way is a bit hidden.. Leaving to you..

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClientFactory.java
```diff
@@ -15,34 +15,30 @@
...
 
 class DefaultStargateBridgeClientFactory implements StargateBridgeClientFactory {
 
   private final Channel channel;
   private final SchemaRead.SourceApi sourceApi;
-  private final DefaultStargateBridgeSchema schema;
+  private final Cache<String, CqlKeyspaceDescribe> keyspaceCache =
+      Caffeine.newBuilder()
+          // TODO tune max size, TTL...
+          .maximumSize(10000)
```
ivansenic, line 32: So this is global cache, shared between all instances? 10K max keyspaces.. Yea we should tune this, this can be a lot of memory imo -> 1GB if keyspace in average takes 100KB.. Not sure what the average is..

ivansenic (commented):
Reviewed, since there is a lot of rest api stuff, someone from rest api team should have a look as well and give final approval imo..

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2ColumnsResourceImpl.java
```diff
@@ -62,6 +63,7 @@ public Response createColumn(
         bridge,
         keyspaceName,
         tableName,
+        false,
```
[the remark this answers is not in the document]
olim7t: Column creation is done with a CQL query: `ALTER TABLE... ADD COLUMN` CQL query (line 76 below).

That CQL query will be authorized automatically by the bridge, so we don't need to pre-authorize.

bridge/src/main/java/io/stargate/bridge/service/SchemaHandler.java
olim7t: I'll add the separate field I alluded to above, hopefully it will clarify things.

bridge/src/main/java/io/stargate/bridge/service/SchemaHandler.java
```diff
@@ -49,27 +49,39 @@
...
-    } catch (StatusException e) {
-      responseObserver.onError(e);
+    } else {
+      try {
+        CqlKeyspaceDescribe description = SchemaHandler.buildKeyspaceDescription(keyspace);
+        responseObserver.onNext(description);
+        responseObserver.onCompleted();
+      } catch (StatusException e) {
+        responseObserver.onError(e);
+      }
```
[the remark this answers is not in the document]
olim7t: I think we should remove the gRPC `DescribeTable` call.

Currently it's not used at all, keyspaces already contain the tables, and managing the cache at keyspace granularity is good enough IMHO.

grpc-proto/proto/schema.proto
```diff
@@ -85,6 +85,11 @@ message CqlKeyspace {
 
 message DescribeKeyspaceQuery {
   string keyspace_name = 1;
+  // The value of `CqlKeyspaceDescribe.hash` the last time the client described that keyspace.
+  // - if it matches the current hash on the server, the server will return an empty response, to
+  //   indicate that the client can continue using the last known value.
+  // - if it doesn't match, or is not provided, the server will return a regular response.
+  google.protobuf.Int32Value hash = 2;
```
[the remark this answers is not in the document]
tatu-at-datastax: We could return hash value for otherwise empty case too? But I guess that would prevent returning a pre-constructed empty response so that's perhaps why we don't do that.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2ColumnsResourceImpl.java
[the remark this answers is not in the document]
tatu-at-datastax: Ok so just to make sure I understand: no need to check read-operation because proper authnz will be needed for actual write operation anyway.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClientFactory.java
tatu-at-datastax: If and when caffeine caches allow specifying weight of items, could use some simple heuristics for calculating weight for cached entries.

Could also consider ttl (in hours probably) to evict least-recently used entries for cases where cache is not full by entry count.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2IndexesResourceImpl.java
```diff
@@ -133,7 +136,7 @@ public Response dropIndex(
 
     CqlTable table =
         bridge
-            .getTable(keyspaceName, tableName)
+            .getTable(keyspaceName, tableName, false)
             .orElseThrow(() -> new WebApplicationException("Table not found", Status.NOT_FOUND));
```
tatu-at-datastax, line 140: Existing code so no absolute need to change but really should include table name in exception message.
(can be fixed in a later commit -- maybe I should go over REST API and systematically add missing info)

tatu-at-datastax (approved):

tatu-at-datastax:
REST changes make sense to me, beyond @ivansenic 's comments and review. Added some notes but none that needs to be addressed before merge I think.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
olim7t: Yeah definitely better to handle it here, I changed it.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClientFactory.java
olim7t: I did a few tests regarding `CqlKeyspaceDescribe` size:
- 20 tables with 20 columns/table retains 63KB.
- 200 tables with 75 columns/table (the Astra limits) retains 2.2MB.

200 tables is a lot, and even then not all of them are going to have that many columns. I think an order of magnitude of a few hundred KBs is realistic. So a cache size of 1000 sounds appropriate.

olim7t:
Any opinion on removing unused bridge operations?
```
rpc DescribeTable(DescribeTableQuery) returns (CqlTable) {}
rpc GetSchemaNotifications(GetSchemaNotificationsParams) returns (stream SchemaNotification) {}
```
The bridge definition is a semi-public API: adding stuff later is easy, but changing or removing stuff without breaking the API will be nearly impossible once we go GA. So I think we should be very conservative as to what we add to it now, even if that means deleting code.

tatu-at-datastax:
> Any opinion on removing unused bridge operations?
> 
> ```
> rpc DescribeTable(DescribeTableQuery) returns (CqlTable) {}
> rpc GetSchemaNotifications(GetSchemaNotificationsParams) returns (stream SchemaNotification) {}
> ```
> 
> The bridge definition is a semi-public API: adding stuff later is easy, but changing or removing stuff without breaking the API will be nearly impossible once we go GA. So I think we should be very conservative as to what we add to it now, even if that means deleting code.

Agreed: let's remove these. Especially the latter since I don't see any use for it at this point. And like you said earlier, former can be implemented in terms of Keyspace access anyway. If we find specific need for it we can re-implement it.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClientFactory.java
ivansenic: Yup, data never lies.. Maybe we can add some extra protection with soft values/keys.. Just to avoid really hard fail here with OOM..

bridge/src/main/java/io/stargate/bridge/service/SchemaHandler.java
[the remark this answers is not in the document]
ivansenic: Fine with me.

ivansenic:
Great work @olim7t

olim7t:
> Revert "Remove obsolete KeyspaceInvalidationListener"

Sorry for the back and forth, I realized that this will still be needed for services that build their own cache on top of keyspace metadata, like GraphQL.

tatu-at-datastax:
> > Revert "Remove obsolete KeyspaceInvalidationListener"
> 
> Sorry for the back and forth, I realized that this will still be needed for services that build their own cache on top of keyspace metadata, like GraphQL.

But would this require implementing of stateful cache update, something we are not designing to do due to complexities in CNDB?

olim7t:
> But would this require implementing of stateful cache update, something we are not designing to do due to complexities in CNDB?

No, that's just invalidation from the local cache in the bridge client (either when the entry timed out, or the client received NOT_FOUND while checking the hash). There are no push notifications coming from the bridge.

tatu-at-datastax:
> > But would this require implementing of stateful cache update, something we are not designing to do due to complexities in CNDB?
> 
> No, that's just invalidation from the local cache in the bridge client (either when the entry timed out, or the client received NOT_FOUND while checking the hash). There are no push notifications coming from the bridge.

Ok, this makes sense, thank you for confirming.

olim7t:
Removed invalidation listener again, I'm really struggling to figure out how caching is going to work in GraphQL 😞 
I'm going to focus on merging this, I'll amend the code in #1651 if needed.

PR #1712, opened 2022-03-22
Dependency check workflow

jeffreyscarpenter:
This PR adds a dependency check workflow using the Maven Versions plugin. This can be triggered manually and is also set to run once daily (we can adjust).

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/.github/workflows/dependency-check.yml
+++ b/.github/workflows/dependency-check.yml
@@ -1,0 +1,45 @@
+name: Check Dependencies
+
+on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron:  '0 12 * * *'
+  workflow_dispatch:
+
+jobs:
+  check-dependencies:
+    name: Check Dependencies
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: "master"
+      - name: Setup Java JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: Setup Maven
+        env:
+          ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_USERNAME }}
+          ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_PASSWORD }}
+        run: |
+          mkdir -p ~/.m2
+          cat <<EOF > ~/.m2/settings.xml
+          <settings>
+            <servers>
+              <server>
+                <id>artifactory-snapshots</id>
+                <username>${ARTIFACTORY_USERNAME}</username>
+                <password>${ARTIFACTORY_PASSWORD}</password>
+              </server>
+              <server>
+                <id>artifactory-releases</id>
+                <username>${ARTIFACTORY_USERNAME}</username>
+                <password>${ARTIFACTORY_PASSWORD}</password>
+             </server>
+           </servers>
+          </settings>
+          EOF
+      - name: Check Maven Dependencies
+        run: |
+          mvn versions:display-dependency-updates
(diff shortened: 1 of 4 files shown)


tatu-at-datastax:
I forgot to ask earlier: is this simply to update to the latest official version (either within current minor version or globally), or based on some vulnerability information? Seems like it could be either:

1. "There is a later released version, want to update" OR
2. "Version you depend on has a report issue(s), but there is a newer safer version"

I guess both can be useful for different use cases.

+1 for getting some sort of report, and I guess we'll see what gets reported.

tatu-at-datastax (approved):

jeffreyscarpenter:
Basically, this workflow is just a bunch of setup so we can run this command:

```
mvn versions:display-dependency-updates
```

The command does have [options](https://www.mojohaus.org/versions-maven-plugin/display-dependency-updates-mojo.html) to control whether to report major or minor version updates. 

This is just going to give us a listing of what upgrades are possible, it does not include vulnerability info. There are other options available in the GH workflow library for checking vulnerabilities, I can look into those as well. Both types of tools could prove useful.

graphqlapi/pom.xml
```diff
@@ -104,10 +104,6 @@
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-servlets</artifactId>
     </dependency>
-    <dependency>
-      <groupId>org.eclipse.jetty</groupId>
-      <artifactId>jetty-server</artifactId>
-    </dependency>
```
ivansenic, line 110: so this is not needed or what?

restapi/pom.xml
```diff
@@ -158,11 +158,6 @@
       <artifactId>assertj-core</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-        <groupId>org.mockito</groupId>
-        <artifactId>mockito-junit-jupiter</artifactId>
-        <scope>test</scope>
-    </dependency>
```
ivansenic, line 165: same here? I am pretty sure we use classes from this package in the tests?

graphqlapi/pom.xml
jeffreyscarpenter: It's a duplicate of a dependency already listed above. The dependency checker chokes on it.

restapi/pom.xml
jeffreyscarpenter: Removal of duplicate entry

.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,87 @@
...
+                <id>artifactory-releases</id>
+                <username>${ARTIFACTORY_USERNAME}</username>
+                <password>${ARTIFACTORY_PASSWORD}</password>
+             </server>
+           </servers>
+          </settings>
+          EOF
+      - name: Check Maven Dependencies
+        run: |
+          mvn dependency-check:aggregate
```
jeffreyscarpenter, line 82: Note that the output of the tool will mention that because the project jars have not been built yet, it is building a "virtual" representation. However when I modified this line and included the maven `package` target, it then complained that the jars were snapshot version and built the virtual representation anyway. I don't see a configuration option that would work around this. So I've elected to not waste time on the compilation step.

maven-version-rules.xml
```diff
@@ -0,0 +1,8 @@
+<ruleset comparisonMethod="maven"
+         xmlns="http://mojo.codehaus.org/versions-maven-plugin/rule/2.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://mojo.codehaus.org/versions-maven-plugin/rule/2.0.0 http://mojo.codehaus.org/versions-maven-plugin/xsd/rule-2.0.0.xsd">
+    <ignoreVersions>
+        <ignoreVersion type="regex">(?i).*[-_\.](alpha|b|beta|rc|m|ea)[-_\.]?[0-9]*</ignoreVersion>
```
jeffreyscarpenter, line 6: this helps the tool to ignore non-final release versions of things such as alpha, beta, rc releases.

PR #1713, opened 2022-03-24
gRPC: Optimize DescribeKeyspace when metadata hasn't changed

olim7t:
**What this PR does**:
Allow client to send a hash of their known version when describing a keyspace. If the metadata hasn't changed, the bridge will return an empty response.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
+++ b/bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
@@ -15,6 +15,8 @@
  */
 package io.stargate.bridge.service;
 
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
 import io.grpc.Status;
 import io.grpc.StatusException;
 import io.grpc.stub.StreamObserver;
@@ -31,9 +33,14 @@ import io.stargate.proto.QueryOuterClass.Query;
 import io.stargate.proto.QueryOuterClass.Response;
 import io.stargate.proto.Schema;
 import io.stargate.proto.StargateBridgeGrpc;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.ScheduledExecutorService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class BridgeService extends StargateBridgeGrpc.StargateBridgeImplBase {
+
+  private static final Logger LOG = LoggerFactory.getLogger(BridgeService.class);
 
   private static final Schema.CqlKeyspaceDescribe EMPTY_KEYSPACE_DESCRIPTION =
       Schema.CqlKeyspaceDescribe.newBuilder().build();
@@ -43,6 +50,12 @@ public class BridgeService extends StargateBridgeGrpc.StargateBridgeImplBase {
 
   private final ScheduledExecutorService executor;
   private final int schemaAgreementRetries;
+
+  private final LoadingCache<Keyspace, Schema.CqlKeyspaceDescribe> descriptionCache =
+      Caffeine.newBuilder()
+          // TODO adjust size and TTL
+          .maximumSize(1000)
+          .build(SchemaHandler::buildKeyspaceDescription);
 
   public BridgeService(
       Persistence persistence,
@@ -108,11 +121,21 @@ public class BridgeService extends StargateBridgeGrpc.StargateBridgeImplBase {
       responseObserver.onCompleted();
     } else {
       try {
-        Schema.CqlKeyspaceDescribe description = SchemaHandler.buildKeyspaceDescription(keyspace);
+        Schema.CqlKeyspaceDescribe description = descriptionCache.get(keyspace);
         responseObserver.onNext(description);
         responseObserver.onCompleted();
-      } catch (StatusException e) {
-        responseObserver.onError(e);
+      } catch (CompletionException e) {
+        Throwable cause = e.getCause();
+        if (cause instanceof StatusException) {
+          responseObserver.onError(e);
+        } else {
+          LOG.error("Internal error while building keyspace description", cause);
+          responseObserver.onError(
+              Status.INTERNAL
+                  .withDescription("Internal error while building keyspace description")
+                  .asException());
+        }
+        descriptionCache.invalidate(keyspace);
       }
     }
   }
--- a/bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
+++ b/bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
@@ -15,9 +15,12 @@
  */
 package io.stargate.bridge.service;
 
+import io.grpc.Status;
+import io.grpc.StatusException;
 import io.grpc.stub.StreamObserver;
 import io.stargate.auth.AuthorizationService;
 import io.stargate.db.Persistence;
+import io.stargate.db.schema.Keyspace;
 import io.stargate.grpc.service.GrpcService;
 import io.stargate.grpc.service.SingleBatchHandler;
 import io.stargate.grpc.service.SingleExceptionHandler;
@@ -31,6 +34,9 @@ import io.stargate.proto.StargateBridgeGrpc;
 import java.util.concurrent.ScheduledExecutorService;
 
 public class BridgeService extends StargateBridgeGrpc.StargateBridgeImplBase {
+
+  private static final Schema.CqlKeyspaceDescribe EMPTY_KEYSPACE_DESCRIPTION =
+      Schema.CqlKeyspaceDescribe.newBuilder().build();
 
   private final Persistence persistence;
   private final AuthorizationService authorizationService;
@@ -88,7 +94,27 @@ public class BridgeService extends StargateBridgeGrpc.StargateBridgeImplBase {
   public void describeKeyspace(
       Schema.DescribeKeyspaceQuery request,
       StreamObserver<Schema.CqlKeyspaceDescribe> responseObserver) {
-    SchemaHandler.describeKeyspace(request, persistence, responseObserver);
+
+    String decoratedName =
+        persistence.decorateKeyspaceName(request.getKeyspaceName(), GrpcService.HEADERS_KEY.get());
+
+    Keyspace keyspace = persistence.schema().keyspace(decoratedName);
+    if (keyspace == null) {
+      responseObserver.onError(
+          Status.NOT_FOUND.withDescription("Keyspace not found").asException());
+    } else if (request.hasHash() && request.getHash().getValue() == keyspace.hashCode()) {
+      // Client already has the latest version, don't resend
+      responseObserver.onNext(EMPTY_KEYSPACE_DESCRIPTION);
+      responseObserver.onCompleted();
+    } else {
+      try {
+        Schema.CqlKeyspaceDescribe description = SchemaHandler.buildKeyspaceDescription(keyspace);
+        responseObserver.onNext(description);
+        responseObserver.onCompleted();
+      } catch (StatusException e) {
+        responseObserver.onError(e);
+      }
+    }
   }
 
   @Override
--- a/bridge/src/main/java/io/stargate/bridge/service/SchemaHandler.java
+++ b/bridge/src/main/java/io/stargate/bridge/service/SchemaHandler.java
@@ -69,7 +69,8 @@ class SchemaHandler {
 
   static CqlKeyspaceDescribe buildKeyspaceDescription(Keyspace keyspace) throws StatusException {
 
-    CqlKeyspaceDescribe.Builder describeResultBuilder = CqlKeyspaceDescribe.newBuilder();
+    CqlKeyspaceDescribe.Builder describeResultBuilder =
+        CqlKeyspaceDescribe.newBuilder().setHash(keyspace.hashCode());
     CqlKeyspace.Builder cqlKeyspaceBuilder = CqlKeyspace.newBuilder();
     cqlKeyspaceBuilder.setName(keyspace.name());
 
--- a/grpc-proto/proto/schema.proto
+++ b/grpc-proto/proto/schema.proto
@@ -85,6 +85,11 @@ message CqlKeyspace {
 
 message DescribeKeyspaceQuery {
   string keyspace_name = 1;
+  // The value of `CqlKeyspaceDescribe.hash` the last time the client described that keyspace.
+  // - if it matches the current hash on the server, the server will return an empty response, to
+  //   indicate that the client can continue using the last known value.
+  // - if it doesn't match, or is not provided, the server will return a regular response.
+  google.protobuf.Int32Value hash = 2;
 }
 
 message CqlKeyspaceDescribe {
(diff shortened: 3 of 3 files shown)


bridge/src/main/java/io/stargate/bridge/service/SchemaHandler.java
```diff
@@ -69,7 +69,8 @@ public static void describeKeyspace(
 
   static CqlKeyspaceDescribe buildKeyspaceDescription(Keyspace keyspace) throws StatusException {
 
-    CqlKeyspaceDescribe.Builder describeResultBuilder = CqlKeyspaceDescribe.newBuilder();
+    CqlKeyspaceDescribe.Builder describeResultBuilder =
+        CqlKeyspaceDescribe.newBuilder().setHash(keyspace.hashCode());
```
olim7t, line 73: Using the Java hashcode for a quick first prototype.
If we estimate that the risk of collision is too high, it should be pretty easy to implement another algorithm.

bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
```diff
@@ -88,7 +94,27 @@ public void executeBatch(Batch batch, StreamObserver<Response> responseObserver)
...
+    if (keyspace == null) {
+      responseObserver.onError(
+          Status.NOT_FOUND.withDescription("Keyspace not found").asException());
+    } else if (request.hasHash() && request.getHash().getValue() == keyspace.hashCode()) {
+      // Client already has the latest version, don't resend
+      responseObserver.onNext(EMPTY_KEYSPACE_DESCRIPTION);
+      responseObserver.onCompleted();
+    } else {
+      try {
+        Schema.CqlKeyspaceDescribe description = SchemaHandler.buildKeyspaceDescription(keyspace);
```
olim7t, line 111: We could cache this to avoid recomputing it for every client. It's a bit more complicated because we'll need to invalidate the cache when the hash changes, so overall I'm not sure if it's worth it.

bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
```diff
@@ -88,7 +94,27 @@ public void executeBatch(Batch batch, StreamObserver<Response> responseObserver)
...
-    SchemaHandler.describeKeyspace(request, persistence, responseObserver);
+
+    String decoratedName =
+        persistence.decorateKeyspaceName(request.getKeyspaceName(), GrpcService.HEADERS_KEY.get());
+
+    Keyspace keyspace = persistence.schema().keyspace(decoratedName);
+    if (keyspace == null) {
+      responseObserver.onError(
+          Status.NOT_FOUND.withDescription("Keyspace not found").asException());
+    } else if (request.hasHash() && request.getHash().getValue() == keyspace.hashCode()) {
```
ivansenic, line 105: How expensive is the `keyspace.hashCode()`? OK I see this is pre-hashed, awesome :+1:

bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
ivansenic: I think it's definitely worth it.. Here are some reasons why:

* first there will be multiple instances of same REST services, meaning that getting schema only once per version is not actually true, if the same tenant hits another instance, it has to go and get it
* second, with the optimization on the bridge you actually allow the REST side service to have no caching at all.. This can give a significant relaxation wrt to memory.. I know this is not the case for the GraphQL because caching allows you to build the GrapQL schema, but maybe for other services it is.. Maybe I am wrong here..

bridge/src/main/java/io/stargate/bridge/service/SchemaHandler.java
ivansenic: This has to be good enough.. Has to be!

grpc-proto/proto/schema.proto
```diff
@@ -85,6 +85,11 @@ message CqlKeyspace {
 
 message DescribeKeyspaceQuery {
   string keyspace_name = 1;
+  // The value of `CqlKeyspaceDescribe.hash` the last time the client described that keyspace.
+  // - if it matches the current hash on the server, the server will return an empty response, to
+  //   indicate that the client can continue using the last known value.
+  // - if it doesn't match, or is not provided, the server will return a regular response.
+  google.protobuf.Int32Value hash = 2;
```
ivansenic, line 92: this is optional right?

ivansenic (approved):
Looks great, how about the client side updates? Next PR? We need to ensure in tests that schema changes that do not touch a specific keyspace does return same hash code.. Meaning you have target keyspace that is not changing, but updating the schema for other, the hash for first one stays consistent..

Regarding the preprocessing of the schema, I am not sure, instinct says it is beneficial.. But I do understand your point as well, especially that hash is related to single keyspace and not the global schema..

ivansenic:
Run the perf test with this brach quickly, having a hash makes this exactly where we want it to be:

**Max 1,000 RPS (actual at 999.9)**
```
Summary:
  Count:	100000
  Total:	100.00 s
  Slowest:	8.43 ms
  Fastest:	0.06 ms
  Average:	0.13 ms
  Requests/sec:	999.98

Response time histogram:
  0.063 [1]     |
  0.900 [99668] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  1.736 [161]   |
  2.573 [51]    |
  3.410 [35]    |
  4.246 [35]    |
  5.083 [22]    |
  5.919 [15]    |
  6.756 [7]     |
  7.593 [4]     |
  8.429 [1]     |

Latency distribution:
  10 % in 0.10 ms 
  25 % in 0.10 ms 
  50 % in 0.11 ms 
  75 % in 0.12 ms 
  90 % in 0.14 ms 
  95 % in 0.16 ms 
  99 % in 0.36 ms 

Status code distribution:
  [OK]   100000 responses 
```

**Unlimited RPS (actual at 25.5K)**
```
Summary:
  Count:	100000
  Total:	3.91 s
  Slowest:	12.60 ms
  Fastest:	0.07 ms
  Average:	1.13 ms
  Requests/sec:	25599.54

Response time histogram:
  0.068  [1]     |
  1.321  [71372] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  2.574  [24791] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  3.827  [2949]  |∎∎
  5.080  [486]   |
  6.333  [171]   |
  7.586  [68]    |
  8.839  [19]    |
  10.092 [45]    |
  11.345 [58]    |
  12.598 [40]    |

Latency distribution:
  10 % in 0.46 ms 
  25 % in 0.66 ms 
  50 % in 0.97 ms 
  75 % in 1.40 ms 
  90 % in 1.92 ms 
  95 % in 2.38 ms 
  99 % in 3.71 ms 

Status code distribution:
  [OK]   100000 responses   

```

bridge/src/main/java/io/stargate/bridge/service/SchemaHandler.java
tatu-at-datastax: For initial prototype yes, but I would suggest that instead of Java String.hash() which is not really very high quality, we'd use same thing as Cassandra: Murmur3 hash:

https://docs.datastax.com/en/cassandra-oss/3.x/cassandra/architecture/archPartitionerM3P.html

Murmur algorithm is not cryptographic but it is very fast, has good distribution. I think you can also make it produce 64-bit hash which I think would be better than 32-bit for our case.

I know this is bit of an internal implementation detail but since it will be communicated across Bridge we need to decide on it relatively early.

grpc-proto/proto/schema.proto
tatu-at-datastax: As per my earlier comment, I'd prefer 64-bit hash.

tatu-at-datastax:
@ivansenic Awesome test results!

grpc-proto/proto/schema.proto
tatu-at-datastax: Is `Int32Value` used to allow explicit "missing" vs defaulting to 0? That is, this is similar to `Integer` vs `int`.
Makes sense I guess and avoids having to have special handling for value `0`.

olim7t: Yes. It generate a `hasHash()` method that allows you to check if the value was explicitly set (vs a simple `sint32` that would default to 0 when not set).

olim7t:
> Quick question: does code need to treat `0` hash code same as missing? I think proto3 uses Java value defaults for "missing" (undefined) values for optional values.

No, by virtue of using `Int32Value` in the request, "unset" is different from 0.

> Javascript does not have 64-bit int type

The current hash is a 32-bit `int`.

olim7t:
> keep a cache of `keyspaceName=>CqlKeyspaceDescribe`

I forgot that `StargateBridgeClient` is short-lived 😡 
Maybe we can have a global cache shared by all clients, but then we'll need to handle tenants, and probably keep authorization separate.

olim7t:
@ivansenic I've added a cache on the bridge side. If the client queries with no hash, the bridge will return the full response every time, but only build the `CqlKeyspaceDescribe` once. That can give us an idea of how the "no cache on service side" approach compares in term of performance.

grpc-proto/proto/schema.proto
tatu-at-datastax: One more thing: use of `String.hashCode()` (or any Java specified `hashCode()`) is probably not a good idea if we want to allow non-Java clients. `String.hashCode()` is in theory just implementation detail (in practice it will never be changed as that'd break too much code), i.e. not specified and (theoretically) not reliably implementable on other platforms.
Use of externally defined standard hash is preferable as implementations exist for many languages.

tatu-at-datastax:
> > keep a cache of `keyspaceName=>CqlKeyspaceDescribe`
> 
> I forgot that `StargateBridgeClient` is short-lived 😡 Maybe we can have a global cache shared by all clients, but then we'll need to handle tenants, and probably keep authorization separate.

Ugh.

Yes, for content-hash based access to be beneficial, client-side does need caching. That in turn would require tenant-awareness in some form.
So I guess there is a little bit of design needed.

I think it might be time to figure out how to make Tenant-awareness more of a first-class concern with Stargate V2. It may make sense to define "default Tenant" concept, so that some id is always carried along and there is no non-tenant functionality.

bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
```diff
@@ -88,7 +107,37 @@ public void executeBatch(Batch batch, StreamObserver<Response> responseObserver)
...
+      // Client already has the latest version, don't resend
+      responseObserver.onNext(EMPTY_KEYSPACE_DESCRIPTION);
+      responseObserver.onCompleted();
+    } else {
+      try {
+        Schema.CqlKeyspaceDescribe description = descriptionCache.get(keyspace);
```
ivansenic, line 124: OK let me test this wrt perf.. I guess you are aware that this is not being updated on schema changes..

ivansenic:
> @ivansenic I've added a cache on the bridge side. If the client queries with no hash, the bridge will return the full response every time, but only build the `CqlKeyspaceDescribe` once. That can give us an idea of how the "no cache on service side" approach compares in term of performance.

Just tested the branch and this gives very bad results.. It's even worse than building the representation every time (concurrency added on that cache now).. So I guess, client side cache is needed, and there is no need to cache anything on the server side, as most likely the serialization and transfer takes most of the CPU cycles and not the persistance-api to grpc-protos Java transformation..

**Max 1,000 RPS (actual 256.78)**:

```
Summary:
  Count:	10000
  Total:	38.94 s
  Slowest:	1.42 s
  Fastest:	17.07 ms
  Average:	189.49 ms
  Requests/sec:	256.78

Response time histogram:
  17.067   [1]    |
  157.162  [4348] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  297.258  [4270] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  437.354  [1091] |∎∎∎∎∎∎∎∎∎∎
  577.450  [196]  |∎∎
  717.546  [59]   |∎
  857.642  [19]   |
  997.738  [11]   |
  1137.834 [1]    |
  1277.929 [3]    |
  1418.025 [1]    |

Latency distribution:
  10 % in 63.31 ms 
  25 % in 113.44 ms 
  50 % in 172.14 ms 
  75 % in 242.31 ms 
  90 % in 325.16 ms 
  95 % in 386.65 ms 
  99 % in 554.17 ms 

Status code distribution:
  [OK]   10000 responses   
```

But it's all fine, having data to actually come to the right design and approach seems fine.. Thumbs up for @olim7t for being right from the beginning..

tatu-at-datastax:
Ok: one last (?) comment on calculating hash value: I think we really need to do it in a way that is well documented and can be reimplemented by non-Java clients. This need not be done immediately but should (I think) be done before official external release.

To me the most straight-forward way would be to

1. serialize and cache gRPC/proto representation as byte[]/ByteBuffer
2. calculate hash using Murmur3 algorithm (or similar).
3. cache byte[]/ByteBuffer representation to avoid serialization penalty

We can and need to figure out why performance @ivansenic found was so bad -- I am pretty sure there's a logical reason and we can resolve it -- for this to work of course.

On very short term we can just piggy-back on `hashCode()` gRPC returns, I think.

Oh and we do need test(s) to ensure that caching works, i.e. repeated calls will use cached copy (can just verify that cache size does increase when requesting same thing).

ivansenic:
I though we agreed we won't have any caching, and Java `hashCode()` on the keyspace object seems fine?

olim7t:
The client doesn't need to recompute the hash, only store what the bridge returned and reuse it for the next request.

tatu-at-datastax:
@ivansenic @olim7t Actually you are right: client does not need to compute the hash. And given that really is a server-side implementation detail.

Given this, I guess it's ok to go with Java object `hashCode()` values, even `String.hashCode()` (which is not great). And if there is need to change the calculation that can be done easily.

tatu-at-datastax:
@ivansenic On server-side caching: that is needed in some form to retain hash code value, isn't it? So it is just a question of what representation of cached entry is needed (Java object, grpc/proto serialized response).
I am fine with caching Java objects for now; improvements if we care can be done later, no need to add in this PR.

bridge/src/main/java/io/stargate/bridge/service/SchemaHandler.java
tatu-at-datastax: As per our discussion, my point was off: client does NOT need to (re)compute hash at all. So it's a server-side implementation detail, except for type used (which needs to be agreed on).

Given this, using Java classes `hashCode()` is reasonable initial implementation. Can be changed if need be, without clients having to care.

olim7t:
Closing this, I'm going to consolidate all schema-related changes in #1711.

PR #1714, opened 2022-03-24
Fix prepared query not found in Cassandra 4.0 by adding lock to StargateQueryHandler

jeffreyscarpenter:
We are regularly getting test failures with condition timeout on `ExecuteBatchStreamingTest.manyStreamingBatch`

```
Step #4 - "cassandra-4.0": [INFO] Results:
Step #4 - "cassandra-4.0": [INFO] 
Step #4 - "cassandra-4.0": [ERROR] Errors: 
Step #4 - "cassandra-4.0": [ERROR]   ExecuteBatchStreamingTest.manyStreamingBatch:157 ? ConditionTimeout Assertion ...
```

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Cassandra40Persistence.java
+++ b/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/Cassandra40Persistence.java
@@ -7,7 +7,6 @@ import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
 import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableMap;
 import com.datastax.oss.driver.shaded.guava.common.collect.Iterables;
 import com.datastax.oss.driver.shaded.guava.common.util.concurrent.Uninterruptibles;
-import com.google.common.util.concurrent.Striped;
 import io.stargate.auth.AuthorizationService;
 import io.stargate.core.util.TimeSource;
 import io.stargate.db.Authenticator;
@@ -39,7 +38,6 @@ import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.concurrent.locks.Lock;
 import java.util.function.Supplier;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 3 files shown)


tatu-at-datastax:
@ivansenic @mpenick Do you think it's just timeout, or actually related to that race condition? To me logs seemed to suggest latter, but I thought that had gotten fixed.

jeffreyscarpenter:
Even if this is not the root cause, it is certainly not going to hurt anything. Can I get an approve?

ivansenic (changes requested):
We should not increase the timeout, as it will just make the CI longer.. The root cause is the prepared issue, as you can see in the logs:

```
Step #4 - "cassandra-4.0": error: 
Step #4 - "cassandra-4.0": expected: 0
Step #4 - "cassandra-4.0":  but was: 13
Step #4 - "cassandra-4.0": 
Step #4 - "cassandra-4.0": code: 13
Step #4 - "cassandra-4.0": message: "INTERNAL: Prepared query with ID 5629fd7bbd270c04199914945c9bca54 not found (either the query was not prepared on this host (maybe the host has been restarted?) or you have prepared too many queries and it has been evicted from the internal cache)"
Step #4 - "cassandra-4.0": details {
Step #4 - "cassandra-4.0":   type_url: "type.googleapis.com/google.rpc.ErrorInfo"
Step #4 - "cassandra-4.0":   value: "\n\355\001Prepared query with ID 5629fd7bbd270c04199914945c9bca54 not found (either the query was not prepared on this host (maybe the host has been restarted?) or you have prepared too many queries and it has been evicted from the internal cache)"
Step #4 - "cassandra-4.0": }
```

ivansenic:
The fix @mpenick implemented is not good, the `executeRequestOnExecutor` returns `CompletableFuture` and thus represents an async operation.. The lock is locked only until the task is submitted to the executor.. I will fix this and push changes to this PR..

```
    @Override
    public CompletableFuture<Result.Prepared> prepare(String query, Parameters parameters) {
      // The behavior of prepared statements was changed as part of CASSANDRA-15252
      // (and CASSANDRA-17248) which can cause multiple prepares of the same query to evict each
      // other from the prepared cache. A few Stargate APIs eagerly prepare queries which
      // increases this chance and to avoid this we serialize prepares for similar queries.
      Lock lock = PREPARE_LOCKS.get(query);
      try {
        lock.lock();
        return executeRequestOnExecutor(
            parameters,
            // The queryStartNanoTime is not used by prepared message, so it doesn't really matter
            // that it's only computed now.
            System.nanoTime(),
            () -> new PrepareMessage(query, parameters.defaultKeyspace().orElse(null)));
      } finally {
        lock.unlock();
      }
    }
```

ivansenic:
The solution in https://github.com/stargate/stargate/pull/1714/commits/f79718450603893e146410266b2b53d53eb8dc2e moves the lock to the query handler.. I added `@RepeatedTest(1000)` for the CI to run it multiple times, will move back to one if all green..

mpenick (approved):
That's rather embarrassing (for me). Locking an asynchronous operation is a new failure mode for me.

Thanks, good catch. LGTM.

tatu-at-datastax:
Looks good although I hope @mpenick can review.

Just one question: to get good release notes should we change title here, or create separate issue?
This because what we have here is not a test issue but fixing an actual underlying bug -- so should be noted as such.

ivansenic:
OK returned to single test execution, Mike has reviewed it, I am merging on green. Adapted the title as well.

tatu-at-datastax:
All checks now pass, merge it @ivansenic ! :-)

ivansenic (approved):

PR #1745, opened 2022-03-29
Force C*4 persistence to use latest Netty (4.1.75-final)

tatu-at-datastax:
**What this PR does**:

Adds managed Netty dependency (using netty-bom) to force use of the latest Netty version by Cassandra-4.0 backend.
Could do the same for 3.11, likely; but NOT for dse backend (uses forked Netty so that would need to be upgraded).

The reason for upgrade is security scan that shows some problems with older Netty jars.

No issue filed (yet).

**Checklist**
- [ ] Changes manually tested
- [x] Automated Tests added/updated -- relies on CI/IT to verify that things still work

### DIFF
--- a/persistence-cassandra-4.0/pom.xml
+++ b/persistence-cassandra-4.0/pom.xml
@@ -18,6 +18,20 @@
     -->
     <cassandra.bundled-driver.version>3.11.0</cassandra.bundled-driver.version>
   </properties>
+  <dependencyManagement>
+    <dependencies>
+      <!-- 29-Mar-2022, tatu: Need to force Netty overrides here
+        -->
+      <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-bom</artifactId>
+        <version>${netty.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
   <dependencies>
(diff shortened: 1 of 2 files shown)


tatu-at-datastax:
Looks like main level managed dependency does not work as DSE persistence backend fails unit tests with:

    java.lang.NoClassDefFoundError: io/netty/channel/epoll/Aio

which is apparently dropped from Netty 4.1.x.

But it should be possible to force at least Cassandra 4.0 backend to use later version (and possibly, if we want to, Cassandra 4.11)

jeffreyscarpenter (approved):

persistence-cassandra-4.0/pom.xml
```diff
@@ -18,6 +18,20 @@
...
+      <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-bom</artifactId>
+        <version>${netty.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
```
ivansenic, line 34: Will this work if there are no direct netty dependencies in the pom, meaning will it also affect the transitive deps?

tatu-at-datastax: Yes.

PR #1746, opened 2022-03-29
Upgrade Java driver 4.10->4.13

tatu-at-datastax:
**What this PR does**:

After initial increase in Cassandra Java driver to 4.10, let's try to go all the way to 4.13 (was: 4.14 initially).
(4.14 was a bridge too far initially, with an odd `esri-geometry-api` dependency miss problem).

Found couple of minor issues:

* org.json transitive dependency assumed, but library not really needed (code is operating on Maps)
* Accidental use of Jackson 1.x annotations instead of 2.x
* Missing Jackson dependency for `restapi` (not breaking since a few things bring it transitively, but good practice to be explicit)

NOTE: as a follow-up 4.13->4.14 it looks like DSE persistence module will need dependency to this:

```
<dependency>
  <groupId>com.esri.geometry</groupId>
  <artifactId>esri-geometry-api</artifactId>
  <version>1.2.1</version>
</dependency>
 ```

which 4.13 has strict dependency to; 4.14 defines as `optional`. But adding direct dependency from `persistence-dse-6.8/pom.xml` did NOT resolve the IT failure so there seems to be need to do something else.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/persistence-api/src/test/java/io/stargate/db/datastore/PersistenceBackedResultSetTest.java
+++ b/persistence-api/src/test/java/io/stargate/db/datastore/PersistenceBackedResultSetTest.java
@@ -40,8 +40,6 @@ import java.util.concurrent.ExecutionException;
 import java.util.function.Predicate;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 class PersistenceBackedResultSetTest {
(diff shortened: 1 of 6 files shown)


jeffreyscarpenter (approved):

tatu-at-datastax:
Looks like getting to 4.14 is tricky enough so that I'll first want to see if 4.13 works already with changes.
And then can ponder on what to do wrt "esri-geometry-api" dependency.

PR #1747, opened 2022-03-29
Dependency check suppression

jeffreyscarpenter:
Suppressing false positives in dependency-check report.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/dependency-check-suppression.xml
+++ b/dependency-check-suppression.xml
@@ -11,4 +11,15 @@
         <cve>CVE-2017-9431</cve>
         <cve>CVE-2020-7768</cve>
     </suppress>
+    <suppress>
+        <notes><![CDATA[
+      False positives related to Stargate grpc artifact with versions 1.0.x.
+      ]]></notes>
+        <packageUrl regex="true">^pkg:maven/io\.stargate\.grpc/grpc@.*$</packageUrl>
+        <cve>CVE-2017-7860</cve>
+        <cve>CVE-2017-7861</cve>
+        <cve>CVE-2017-8359</cve>
+        <cve>CVE-2017-9431</cve>
+        <cve>CVE-2020-7768</cve>
+    </suppress>
 </suppressions>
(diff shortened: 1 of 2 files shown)


dependency-check-suppression.xml
```diff
@@ -0,0 +1,40 @@
...
+        <cve>CVE-2017-7861</cve>
+        <cve>CVE-2017-8359</cve>
+        <cve>CVE-2017-9431</cve>
+        <cve>CVE-2020-7768</cve>
+    </suppress>
+    <suppress>
+        <notes><![CDATA[
+      False positives related to Stargate grpc artifact with versions 1.0.x.
+      ]]></notes>
+        <packageUrl regex="true">^pkg:maven/io\.stargate\.grpc/grpc@.*$</packageUrl>
```
jeffreyscarpenter, line 18: can probably combine this with previous block with a wildcard, it's all the same CVEs

dependency-check-suppression.xml
```diff
@@ -0,0 +1,40 @@
...
+    <suppress>
+        <notes><![CDATA[
+      False positive, jackson-databind-2.12.6.1.jar specifically pulled in under https://github.com/stargate/stargate/pull/1741
+      ]]></notes>
+        <packageUrl regex="true">^pkg:maven/com\.fasterxml\.jackson\.core/jackson\-databind@.*$</packageUrl>
+        <cve>CVE-2020-36518</cve>
+    </suppress>
+    <suppress>
+        <notes><![CDATA[
+      False positives related to testing module - Kotlin
```
tatu-at-datastax, line 34: Does the tool report vulns in test-scoped dependencies too? Those are rarely useful (since it's rather hard to exploit test deps), maybe plug-in can be configured to fully ignore all of them?

tatu-at-datastax (approved):

dependency-check-suppression.xml
jeffreyscarpenter: There is a `skipTestScope` option available. I tried enabling that option yesterday and it made no difference in the output, so I did not commit the change. We can always add that later if we determine we are getting a lot of noise.

PR #1750, opened 2022-03-30
Add comparable bytes to row decorator

### LINKED ISSUE #1716: Extend `RowDecorator` interface to return comparable bytes for a row
Introduce new method that uses the bytes comparable API on the `DecoratedKey` to return comparable bytes for each `Row`. Since C* 3.11 and 4.0 depend on https://github.com/apache/cassandra/pull/1294 to be merged, for now return the empty bytes array in the relevant `RowDecoratorImpl`.

Extend the `PersistenceTest` or `DsePersistenceIT` directly test in order to test the new feature in the similar fashion as currently in the `testRowDecorator`.

EricBorczuk:
**What this PR does**:

Fixes #1716

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/persistence-api/src/main/java/io/stargate/db/RowDecorator.java
+++ b/persistence-api/src/main/java/io/stargate/db/RowDecorator.java
@@ -31,5 +31,9 @@ public interface RowDecorator {
    */
   <T extends Comparable<T>> ComparableKey<T> decoratePartitionKey(Row row);
 
+  /**
+   * Generates the comparable byte value of the {@link Row}, using the "comparable bytes" API of the
+   * underlying datastore. If no such API exists, an empty ByteBuffer is returned.
+   */
   ByteBuffer getComparableBytes(Row row);
 }
--- a/persistence-api/src/main/java/io/stargate/db/RowDecorator.java
+++ b/persistence-api/src/main/java/io/stargate/db/RowDecorator.java
@@ -17,6 +17,7 @@ package io.stargate.db;
 
 import io.stargate.db.datastore.ResultSet;
 import io.stargate.db.datastore.Row;
+import java.util.stream.Stream;
 
 /**
  * A table-specific interface for extracting key column values from a {@link ResultSet} {@link Row}
@@ -29,4 +30,6 @@ public interface RowDecorator {
    * same order that queries iterate / paginate over the Cassandra data ring.
    */
   <T extends Comparable<T>> ComparableKey<T> decoratePartitionKey(Row row);
+
+  Stream<Byte> getComparableBytes(Object... rawKeyValues);
 }
--- a/persistence-api/src/main/java/io/stargate/db/RowDecorator.java
+++ b/persistence-api/src/main/java/io/stargate/db/RowDecorator.java
@@ -17,7 +17,7 @@ package io.stargate.db;
 
 import io.stargate.db.datastore.ResultSet;
 import io.stargate.db.datastore.Row;
-import java.util.stream.Stream;
+import java.nio.ByteBuffer;
 
 /**
  * A table-specific interface for extracting key column values from a {@link ResultSet} {@link Row}
@@ -31,5 +31,5 @@ public interface RowDecorator {
    */
   <T extends Comparable<T>> ComparableKey<T> decoratePartitionKey(Row row);
 
-  Stream<Byte> getComparableBytes(Row row);
+  ByteBuffer getComparableBytes(Row row);
 }
--- a/persistence-api/src/main/java/io/stargate/db/RowDecorator.java
+++ b/persistence-api/src/main/java/io/stargate/db/RowDecorator.java
@@ -31,5 +31,5 @@ public interface RowDecorator {
    */
   <T extends Comparable<T>> ComparableKey<T> decoratePartitionKey(Row row);
 
-  Stream<Byte> getComparableBytes(Object... rawKeyValues);
+  Stream<Byte> getComparableBytes(Row row);
 }
--- a/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/RowDecoratorImpl.java
+++ b/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/RowDecoratorImpl.java
@@ -19,12 +19,35 @@ import io.stargate.db.AbstractRowDecorator;
 import io.stargate.db.ComparableKey;
 import io.stargate.db.schema.TableName;
 import java.util.stream.Collectors;
+import java.util.Spliterator;
+import java.util.stream.StreamSupport;
 import org.apache.cassandra.db.Clustering;
 import org.apache.cassandra.db.DecoratedKey;
 import org.apache.cassandra.db.PartitionPosition;
 import org.apache.cassandra.schema.TableMetadata;
+import org.apache.cassandra.utils.ByteSource.Peekable;
+import static org.apache.cassandra.utils.ByteSource.END_OF_STREAM;
 
 public class RowDecoratorImpl extends AbstractRowDecorator {
+
+  /**
+   * Utility to treat a ByteSource as an Iterable, to easily convert to Stream.
+   * This can likely be reused in the other RowDecoratorImpl's once ByteSource is available in Cassandra 3.11/4.0.
+   */
+  private static class ByteSourceIterable implements Iterable<Byte> {
+    Peekable src;
+    public ByteSourceIterable(Peekable src) {
+      this.src = src;
+    }
+
+    public Byte next() {
+      return src.next().byteValue();
+    }
+
+    public boolean hasNext() {
+      return src.peek() != END_OF_STREAM;
+    }
+  }
 
   private final TableMetadata metadata;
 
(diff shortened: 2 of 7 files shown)


persistence-api/src/main/java/io/stargate/db/RowDecorator.java
```diff
@@ -29,4 +30,6 @@
    * same order that queries iterate / paginate over the Cassandra data ring.
    */
   <T extends Comparable<T>> ComparableKey<T> decoratePartitionKey(Row row);
+
+  Stream<Byte> getComparableBytes(Object... rawKeyValues);
```
EricBorczuk, line 34: I decided to use Stream<Byte> here because `ByteSource` (the class that comparable bytes returns) is best represented as an iterable of bytes, rather than a list with known length - so the consumer of this method would be responsible for either consuming the whole stream and creating a list, or passing the stream along (maybe gRPC supports this?)

EricBorczuk:
Not sure why dse-build is consistently failing here...

persistence-api/src/main/java/io/stargate/db/RowDecorator.java
tatu-at-datastax: But isn't `Stream<Byte>` very inefficient? `Byte` is wrapper type and that seems ... like a very inefficient abstraction, and at least wrt performance worse than `ByteSource`.

EricBorczuk: ah I should use ByteBuffer 🤦 idk what I was thinking

tatu-at-datastax: Yeah that's a solid choice. Np! :)

ivansenic: Yep good points @tatu-at-datastax

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/RowDecoratorImpl.java
```diff
@@ -43,4 +50,16 @@ public RowDecoratorImpl(TableName tableName, TableMetadata tableMetadata) {
...
+
+  @Override
+  public ByteBuffer getComparableBytes(Row row) {
+    Clustering key = metadata.partitionKeyAsClusteringComparator().make(primaryKeyValues(row));
+    DecoratedKey decoratedKey = metadata.partitioner.decorateKey(key.serializeAsPartitionKey());
+    ByteSource src = decoratedKey.asComparableBytes(ByteComparable.Version.DSE68);
+    ByteArrayOutputStream baos = new ByteArrayOutputStream();
+    for (int i = src.next(); i != END_OF_STREAM; i = src.next()) {
+      baos.write(i);
+    }
```
ivansenic, line 62: should we simply use: `com.datastax.bdp.db.tries.util.ByteSourceUtil#readBytes(org.apache.cassandra.utils.ByteSource)`

restapi/src/test/java/io/stargate/db/datastore/AlphabeticalOrderPartitionKeyDecorator.java
```diff
@@ -43,4 +44,11 @@ public AlphabeticalOrderPartitionKeyDecorator(List<Column> allColumns) {
     //noinspection unchecked
     return (ComparableKey<T>) new ComparableKey<>(String.class, decorated);
   }
+
+  @Override
+  public ByteBuffer getComparableBytes(Row row) {
+    // TODO replace this with the relevant row's byte-comparable value when
+    // https://github.com/apache/cassandra/pull/1294 is ready
+    return ByteBuffer.wrap(new byte[] {});
```
ivansenic, line 52: let's implement as same as the in the above method, just move that string bytes into the byte buffer?

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
```diff
@@ -1964,6 +1964,31 @@ public void testRowDecorator() throws ExecutionException, InterruptedException {
...
+
+    ResultSet rs1 = dataStore.execute(selectAll).get();
+    RowDecorator dec1 = rs1.makeRowDecorator();
+    ByteBuffer src = dec1.getComparableBytes(rs1.one());
+    if (backend.isDse()) {
+      assertThat(src.array().length).isGreaterThan(0);
+    } else {
+      assertThat(src.array().length).isEqualTo(0);
+    }
+  }
```
ivansenic, line 1990: this is not enough, we need the same test as in the ` testRowDecorator`, just confirming that bytes are created is not enough, we need to confirm that comapring is done correctly.. should be one-to-one with the `testRowDecorator` and then test comparing using a unsigned bytes comparator .. And sure only active for DSE now..

ivansenic (changes requested):
Looks good, but test needs to be better..

persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/RowDecoratorImpl.java
EricBorczuk: ah I didn't see that this existed - I will use it :+1:

tatu-at-datastax: Sounds good, although which package is that from? (bdp means... DSE?)

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
EricBorczuk: I added a better test for this - but in order to do so I had to have `ByteComparable` in scope and so had to add a DSE dependency to the persistence-test module...hope that's ok.

Maybe it can be reconfigured to _only_ import that package

persistence-test/pom.xml
```diff
@@ -8,6 +8,23 @@
   </parent>
   <groupId>io.stargate.db</groupId>
   <artifactId>persistence-test</artifactId>
+  <properties>
```
ivansenic, line 11: This is definitely not needed.. Will comment on the test more..

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
```diff
@@ -1964,6 +1966,42 @@ public void testRowDecorator() throws ExecutionException, InterruptedException {
...
+    ResultSet rs1 = dataStore.execute(selectAll).get();
+    RowDecorator dec1 = rs1.makeRowDecorator();
+    List<Row> rows = rs1.rows();
+    if (backend.isDse()) {
+      List<String> allComparableByteStrings =
+          rows.stream()
+              .map(dec1::getComparableBytes)
+              .map(ByteComparable::fixedLength)
+              .map(x -> x.byteComparableAsString(ByteComparable.Version.DSE68))
+              .collect(Collectors.toList());
```
ivansenic, line 1993: You shouldn't depend on these DSE related interfaces? What's the issue with comparing bytes in unsigned order directly?

I know there is `UnsignedBytes` in Guava, this should have something we could use.. Or use something like this (take from `PureJavaOperations` class):

```
    @Override
    public int compare(ByteBuffer buffer1, ByteBuffer buffer2) {
      int end1 = buffer1.limit();
      int end2 = buffer2.limit();
      for (int i = buffer1.position(), j = buffer2.position(); i < end1 && j < end2; i++, j++) {
        int a = (buffer1.get(i) & 0xff);
        int b = (buffer2.get(j) & 0xff);
        if (a != b) {
          return a - b;
        }
      }
      return buffer1.remaining() - buffer2.remaining();
    }
```

persistence-test/src/main/java/io/stargate/it/PersistenceTest.java
```diff
@@ -1964,6 +1966,42 @@ public void testRowDecorator() throws ExecutionException, InterruptedException {
...
+      Collections.sort(allComparableByteStrings);
+      for (int idx = 0; idx < allComparableByteStrings.size(); idx++) {
+        String cb1 = allComparableByteStrings.get(idx);
+        String cb2 =
+            ByteComparable.fixedLength(dec1.getComparableBytes(rows.get(idx)))
+                .byteComparableAsString(ByteComparable.Version.DSE68);
+        assertThat(cb1.compareTo(cb2)).isEqualTo(0);
+      }
+    }
```
ivansenic, line 2002: Not sure what's going on here, why simply not using and adapting the code from the above method: 

```
while (it1.hasNext()) {
      assertThat(it2.hasNext()).isTrue();
      Row r1 = it1.next();
      Row r2 = it2.next();
      first = first == null ? r1 : first;
      last = r1;
      assertEq(r1, r2, dec1, dec2);
      if (p1 == null) {
        p1 = r1;
      }
      assertGtEq(r1, p1, dec1, dec2);
    }
    assertThat(it2.hasNext()).isFalse();
    assertGt(last, first, dec1, dec2);
}    
```

ivansenic (changes requested):
Tests still not looking good, new dependency is definitely not a good choice.

ivansenic (approved):
Yup that's it..

PR #1757, opened 2022-04-04
Add bridge method to get supported persistence features

olim7t:
**What this PR does**:
Trivial change to expose `Persistence.supports*` methods through the bridge. This will be needed for both GraphQL and Docs API.

/

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
+++ b/bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
@@ -37,6 +37,7 @@ public class BridgeService extends StargateBridgeGrpc.StargateBridgeImplBase {
 
   private final ScheduledExecutorService executor;
   private final int schemaAgreementRetries;
+  private final Schema.SupportedFeaturesResponse supportedFeaturesResponse;
 
   public BridgeService(
       Persistence persistence,
@@ -54,6 +55,12 @@ public class BridgeService extends StargateBridgeGrpc.StargateBridgeImplBase {
     this.authorizationService = authorizationService;
     this.executor = executor;
     this.schemaAgreementRetries = schemaAgreementRetries;
+    this.supportedFeaturesResponse =
+        Schema.SupportedFeaturesResponse.newBuilder()
+            .setSecondaryIndexes(persistence.supportsSecondaryIndex())
+            .setSai(persistence.supportsSAI())
+            .setLoggedBatches(persistence.supportsLoggedBatches())
+            .build();
   }
 
   @Override
@@ -98,5 +105,13 @@ public class BridgeService extends StargateBridgeGrpc.StargateBridgeImplBase {
     new AuthorizationHandler(
             request, GrpcService.CONNECTION_KEY.get(), authorizationService, responseObserver)
         .handle();
+  }
+
+  @Override
+  public void getSupportedFeatures(
+      Schema.SupportedFeaturesRequest request,
+      StreamObserver<Schema.SupportedFeaturesResponse> responseObserver) {
+    responseObserver.onNext(supportedFeaturesResponse);
+    responseObserver.onCompleted();
   }
 }
--- a/grpc-proto/proto/schema.proto
+++ b/grpc-proto/proto/schema.proto
@@ -141,3 +141,20 @@ message SchemaRead {
 message AuthorizeSchemaReadsResponse {
   repeated bool authorized = 1;
 }
+
+// The arguments to a GetSupportedFeatures call.
+// Intentionally empty, there are no arguments at this time but this is intended for future
+// extensibility.
+message SupportedFeaturesRequest {}
+
+// The response to a GetSupportedFeatures call.
+message SupportedFeaturesResponse {
+  // Whether the persistence backend supports regular Cassandra secondary indexes.
+  bool secondary_indexes = 1;
+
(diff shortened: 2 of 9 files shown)


tatu-at-datastax:
Sounds good. Would it be feasible to add integration tests for (at least some of) the backends to verify implementations as well? That is, that DSE-6.8 for example reports features we expect it to return, end-to-end.

olim7t:
> Sounds good. Would it be feasible to add integration tests for (at least some of) the backends to verify implementations as well? That is, that DSE-6.8 for example reports features we expect it to return, end-to-end.

It's feasible but pretty low-value, the actual implementations are either hard-coded or rely on a system property.

tatu-at-datastax:
> > Sounds good. Would it be feasible to add integration tests for (at least some of) the backends to verify implementations as well? That is, that DSE-6.8 for example reports features we expect it to return, end-to-end.
> 
> It's feasible but pretty low-value, the actual implementations are either hard-coded or rely on a system property.

Fair enough.

tatu-at-datastax (approved):

grpc-proto/proto/schema.proto
```diff
@@ -141,3 +141,20 @@ message SchemaRead {
...
+message SupportedFeaturesResponse {
+  // Whether the persistence backend supports regular Cassandra secondary indexes.
+  bool secondary_indexes = 1;
+
+  // Whether the persistence backend supports Storage Attached Indexes.
+  bool sai = 2;
+
+  // Whether the persistence backend supports logged CQL batches.
+  bool logged_batches = 3;
+}
```
ivansenic, line 160: My idea was to solve this with configuration on the Docs API.. I want to avoid having to make this call.. Meaning I will have something like:

```
stargate.data-store-support.secondaryIndexes=
stargate.data-store-support.sai=
stargate.data-store-support.logged-batch=
```

Then I would create profiles for each supported data store, for example, `application-DSE.yaml` which would contain props for the DSE. Then when starting the microservice you can select the profile based on the underlying data store.

Not saying that this is not good, just explaining the option I had in mind. This makes less configuration, but with that extra call.. Not sure if this makes testing harder, but yea I am fine with both options..

bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
```diff
@@ -99,4 +106,12 @@ public void authorizeSchemaReads(
             request, GrpcService.CONNECTION_KEY.get(), authorizationService, responseObserver)
         .handle();
   }
+
+  @Override
```
ivansenic, line 110: can we add a simple java docs

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/StargateBridgeClient.java
```diff
@@ -166,4 +167,10 @@ default Response executeBatch(Batch batch) {
   default boolean authorizeSchemaRead(SchemaRead schemaRead) {
     return authorizeSchemaReads(Collections.singletonList(schemaRead)).get(0);
   }
+
+  CompletionStage<SupportedFeaturesResponse> getSupportedFeaturesAsync();
+
+  default SupportedFeaturesResponse getSupportedFeatures() {
+    return Futures.getUninterruptibly(getSupportedFeaturesAsync());
+  }
```
ivansenic, line 175: same here, java docs

ivansenic (approved):
Looks great.. Have a look on my first comment..

grpc-proto/proto/schema.proto
olim7t: > My idea was to solve this with configuration on the Docs API.

The downside is that this relies entirely on the user getting it right... If there is a configuration error, the service will try to use the wrong features and get errors.

> I want to avoid having to make this call.

Yeah that's a valid concern.
In the GraphQL API it doesn't really matter because the call is not on a hot path.
But if needed I could memoize the result at the `StargateBridgeClientFactory` level, so that the gRPC call is only done once (the response is always the same for a given persistence implementation).

tatu-at-datastax: This is all-around problematic because conceptually Service is asking for Persistence-side feature and ideally it would ask and not be statically configured. But I do understand that for practical reasons it may be necessary for Service to configure such settings so as not to have to connect to Persistence.

Perhaps we cannot unify all such introspection behind same API; so maybe changes here should be configured wrt how well they work for GraphQL use case. And @ivansenic's settings may need to be handled differently.
I am not sure this is the way to go but wanted to mentioned that different use cases sometimes require different solutions.

olim7t: > we can call it once on the app startup and consider that it's fixed.. But this is not possible with the current routing approach, as you need to do it in the request context.

We can do it for the first request that asks for it, and then cache the result.

ivansenic: True.

bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
olim7t: The method inherits the javadoc of the generated parent class, which is in turn copied from the proto file:
```
    /**
     * <pre>
     * Checks which features are supported by the persistence backend.
     * </pre>
     */
```

Issue #1719, opened 2022-03-28
Setup observability for the Quarkus based microservices


ivansenic:
Use the available Quarkus metrics and tracing support and extend it to add the tenant awareness. 

* Add the `http.server.requests.counter` with tenant-id as the tag and if possible exception in case of error responses.
* See if enriching the trace with the tenant id is possible easily.

In addition we should also:

* enable client-side gRPC metrics & tracing
* use Prometheus as the registry
* see available options for the tracing registries, especially for the development use-case

ivansenic:
Fixed in #1760.

PR #1760, opened 2022-04-05
closes #1717: boostratping the Docs API V2 with Quarkus

### LINKED ISSUE #1717: Bootstrap Document API V2 project with Quarkus
The new Document API V2 micro-service will use the Quarkus as the base Java framework. The bootstrapping should prepare the project for future development and take car of the shared concepts and concerns:

* Maven setup and integration with the current project structure
* OpenAPI/Swagger setup
* Hibernate validator setup
* Unified exception handling implementation
* Header-based authentication support
* Base implementation for:
   * Token awareness
   * Tenant awareness
* Docker image setup 
* Lombok support (optional)

ivansenic:
**What this PR does**:

This PR provides the Quarkus bootstrapping for the Docs API V2.

The best way to review this is to first read the `README.MD` file, I know it's long, but once you read it, it will be easier to understand all the code and configuration. Some major concerns are solved here in addition to the stuff provided by the Quarkus itself. Those are not related to the Document API, but to new micro services in general:

* header-based security
* multi-tenancy
* Cassandra token resolving
* gRPC metadata passing

There is an `ExampleResource` that explains how the flow should look like.

The PR does not contain any document related code. Note that only `.java` files, `pom.xml`, `application.yaml`, banner config and `README.MD` files were added by me. Everything rest is generated and I guess should not be reviewed.

It's important to mention that this project is independent for now, it's not part of the parent project. We need to find a way to have projects with different Java target in the mono-repo. Quarkus supports only Java 11+ and I bootstrapped with Java 17 actually. And used some nice new features as well. I guess the right way to do this would be [toolchains](https://maven.apache.org/guides/mini/guide-using-toolchains.html). Or optionally breaking out from the main repository..
 
Code is not formatted, and CI is not checking this project for now. All of this will be done, once we connect to the parent project. The observability parts (metrics, tracing) are also not included as we have a separate issue for this -> #1719.

The majority of this code should go to some kind of `svg2-common-quarkus` project, but there is no need to do this, until we have another project moving to Quarkus. 

The project for now depends only on the `grpc-proto`, we will connect it do the existing `sgv2-service-common` project once we need features from it.

I ll present more in-depth details on my demo on Thursday. Unit tests for all new components will be added once we agree on the approach being valid.

Relates to #1717.

**Checklist**
- [x] Changes manually tested
- [x] Finalize OpenAPI dev props 
- [x] Unit tests
- [x] Solve formatting, the `fmt-maven-plugin` does not support Java 17
- [x] Integrate into the CI so tests are verified
- [x] Quarkus update to `2.8.0.Final`
    - check support for `@Nested` fixed
    - check support for `MockitoExtension` fixed


sgv2-docsapi/README.md
```diff
@@ -0,0 +1,236 @@
...
+| `stargate.tenant-resolver.fixed.tenant-id` | `String` | unset       | Tenant identifier value if the `fixed` type is used.                                                                                |
+
+#### [Cassandra token configuration](src/main/java/io/stargate/sgv2/docsapi/config/TokenResolverConfig.java)
+*Configuration mapping for the Cassandra token resolving.*
+
+| Property                                     | Type     | Default             | Description                                                                                                                |
+|----------------------------------------------|----------|---------------------|----------------------------------------------------------------------------------------------------------------------------|
+| `stargate.token-resolver.type`               | `String` | `principal`         | Token resolver type. If unset, empty token is used. Possible options are `header`, `subdomain` `fixed`, `custom` or unset. |
```
ivansenic, line 82: `subdomain` -> `principal`

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/example/ExampleResource.java
```diff
@@ -0,0 +1,117 @@
...
+@SecurityRequirement(name = Constants.OPEN_API_DEFAULT_SECURITY_SCHEME)
+public class ExampleResource {
+
+    @Inject
+    GrpcClients grpcClients;
+
+    @Inject
+    RequestInfo requestInfo;
+
+    @Operation(
```
ivansenic, line 60: I did not extract OpenAPI definitions into the interface as the restapi v2, but this is definitely an option imo..

sgv2-docsapi/README.md
```diff
@@ -0,0 +1,236 @@
...
+./mvnw package -Pnative -Dquarkus.native.container-build=true
+```
+
+You can then execute your native executable with: `./target/sgv2-docsapi-${project.version}-runner`
+
+If you want to learn more about building native executables, please consult [Maven tooling](https://quarkus.io/guides/maven-tooling).
+
+### Creating a docker image
+
+You can create a docker image named `io.stargate/docsapi` using:
```
tatu-at-datastax, line 150: @jeffreyscarpenter What would our SGv2 naming convention be, wrt existing REST API docker image?

sgv2-docsapi/README.md
```diff
@@ -0,0 +1,236 @@
...
+```shell script
+./mvnw package -Pnative -Dquarkus.native.container-build=true -Dquarkus.container-image.build=true
+```
+
+If you want to learn more about building container images, please consult [Container images](https://quarkus.io/guides/container-image).
+
+## Quarkus Extensions
+
+This project uses various Quarkus extensions, modules that run on top of a Quarkus application.
+You can list, add and remove the extensions using the `quarkus ext` command.
```
tatu-at-datastax, line 165: What does this do? (modify a config file, or pom.xml maybe?). That is, how does it add extension(s).

sgv2-docsapi/pom.xml
```diff
@@ -0,0 +1,182 @@
+<?xml version="1.0"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
```
tatu-at-datastax, line 2: Should this have "io.stargate:stargate" as its parent? (to get various properties, some managed dependencies)

sgv2-docsapi/pom.xml
```diff
@@ -0,0 +1,182 @@
...
+    <quarkus.generate-code.grpc.scan-for-proto>io.stargate.grpc:grpc-proto</quarkus.generate-code.grpc.scan-for-proto>
+    <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
+    <quarkus.platform.group-id>io.quarkus.platform</quarkus.platform.group-id>
+    <quarkus.platform.version>2.7.5.Final</quarkus.platform.version>
+    <surefire-plugin.version>3.0.0-M5</surefire-plugin.version>
+  </properties>
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>${quarkus.platform.group-id}</groupId>
```
tatu-at-datastax, line 25: Is this redirection needed? (maybe it is recommended by Quarkus project or something). That is, why not just include `io.quarkus.platform` and `quarkus-bom` directly?

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/common/RequestInfo.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.ws.rs.core.SecurityContext;
+import java.util.Optional;
+
+/**
+ * The request information containing the tenant ID and the Cassandra key. This bean is @{@link RequestScoped}.
+ * <p>
+ * Uses the registered {@link TenantResolver} and {@link CassandraTokenResolver} to optionally resolve the tenant ID and the Cassandra token.
+ */
+@RequestScoped
+public class RequestInfo {
```
tatu-at-datastax, line 36: Maybe `StargateRequestInfo`?

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/common/exception/model/dto/ApiError.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+public record ApiError(
+
+        @Schema(description = "A human readable description of the error state.", example = "Could not create collection `custom-users`, it has invalid characters. Valid characters are alphanumeric and underscores.")
+        String description,
+
+        @Schema(description = "The internal number referencing the error state.", example = "22000")
+        int code,
+
+        // TODO Is this still needed and used? This was not part of the public API.
```
tatu-at-datastax, line 37: Seems generally useful? But agreed, should be documented, public, if so.

tatu-at-datastax:
Looks good. One bigger question: since many of the abstractions are probably to be shared, would it make sense to add another repo already, instead of adding shared components under `sgv2-docsapi`? I guess they are easy enough to move as the second step once we need them so adding them under Docs API initially is fine, keeps things simple.

sgv2-docsapi/README.md
ivansenic: I checked, you have `io.stargate/restapi`, so I followed the same :)

sgv2-docsapi/README.md
ivansenic: so basically changes the `pom.xml`, not sure if there are any other operations for certain extenstions..

sgv2-docsapi/pom.xml
ivansenic: Mentioned in the PR description that is not yet connected to the parent..

sgv2-docsapi/pom.xml
ivansenic: This is auto-generated setup by Quarkus CLI, don't want to touch this at all.. Not sure how this works all together..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/common/RequestInfo.java
ivansenic: Funny, I had it as `StargateRequestInfo` at the beginning, then switched, sure will revoke..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/common/exception/model/dto/ApiError.java
ivansenic: In the current `restapi` project this is not documented as the API parameter.. I also think it's not part of any examples, this is why I ask..

ivansenic:
> 

Exactly, keeping it simple for now, I need 2h max to extract to new project and set up everything when needed..

ivansenic:
@tatu-at-datastax What do you think about this kind of property setup:

```
stargate:
  auth: 
    header-based:
       ...
    token-resolver:
       ...
  multi-tenancy:
    tenant-resolver:  
       ...
```

Not sure..

jeffreyscarpenter:
> The majority of this code should go to some kind of `svg2-common-quarkus` project, but there is no need to do this, until we have another project moving to Quarkus.

Agree with deferring the refactoring, but suggest code should move to `sgv2-service-common`instead of creating another new module.

jeffreyscarpenter:
not sure of the best place to capture this, but want to make sure we consider other factors as we work toward integrating this more broadly in the project:

- adapting our maven multi-module structure to support multiple Java versions
- moving common code to sgv2-service-common 
- update CI workflows to sure the additional JDK is present
- incorporate docker image publishing from Quarkus-based services into the release workflow

sgv2-docsapi/README.md
tatu-at-datastax: Good, good, just wanted to mention!

sgv2-docsapi/pom.xml
tatu-at-datastax: Ah. Yes, makes sense due to tooling challenges etc.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/common/exception/model/dto/ApiError.java
tatu-at-datastax: Yeah I think that API endpoints were just copied as-is; but auth token handling changed a few times to/from explicitly passing vs. handled by filter. Or something similar.
But bottomline being that the way restapi does may well be lacking things.

tatu-at-datastax:
> @tatu-at-datastax What do you think about this kind of property setup:
> 
> ```
> stargate:
>   auth: 
>     header-based:
>        ...
>     token-resolver:
>        ...
>   multi-tenancy:
>     tenant-resolver:  
>        ...
> ```
> 
> Not sure..

Looks clear enough? Do you have concerns, alternatives to consider?

tatu-at-datastax:
> > The majority of this code should go to some kind of `svg2-common-quarkus` project, but there is no need to do this, until we have another project moving to Quarkus.
> 
> Agree with deferring the refactoring, but suggest code should move to `sgv2-service-common`instead of creating another new module.

Although eventually it might not make a big difference, my feeling is that what @ivansenic suggests makes sense: framework-dependant/specific things may make sense in a package separate from more general things.

Actually I think the biggest reason for doing this is to make it practical to use frameworks other than Quarkus; if so, such impls would not want to get Q dependencies. I guess it is debatable as to how big feature this (being framework agnostic) is.

ivansenic:
> 

This is a new proposal, the old props are in the current `application.yaml` file..

ivansenic:
> not sure of the best place to capture this, but want to make sure we consider other factors as we work toward integrating this more broadly in the project:
> 
>     * adapting our maven multi-module structure to support multiple Java versions
> 
>     * moving common code to sgv2-service-common
> 
>     * update CI workflows to sure the additional JDK is present
> 
>     * incorporate docker image publishing from Quarkus-based services into the release workflow

All valid points :+1: 

However not everything can be done at once.. 

* regarding the parent project we need to agree how to handle this.. 
* common project can be added once we need another quarkus project, don't want to do premature optimization atm
* yup CI has to be solved rather sooner than later
* docker image publishing can be done at the end imo

ivansenic:
Updated to latest SG version, fixed docker image creation, added tests, addressed some review comments, etc.

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/common/token/impl/HeaderTokenResolverTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+import javax.enterprise.inject.Instance;
+import javax.inject.Inject;
+import java.util.Map;
+import java.util.Optional;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.when;
+
+@QuarkusTest
```
EricBorczuk, line 38: These tests seem really straightforward and easy to set up :+1:

EricBorczuk (approved):
This looks great! Seems like it'll be pretty straightforward to use quarkus, I like it a lot overall.

I think I had one minor question about mvnw, but that was just for my own edification really

sgv2-docsapi/mvnw
```diff
@@ -0,0 +1,316 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
ivansenic: That's only until we merge it with the parent project.. This will be sorted out..

ivansenic:
@stargate/core @stargate/core-contributors Can anybody explain the deal with the `cloudbuild.yaml` which should resolve deps for the builds and update the "m2" bucket?

* Why is it running, when all the other build files would also push to the bucket on completion?
* If this was intended to be a dependency to other builds checks, why is this not the case?

PR #1769, opened 2022-04-12
[WIP] Multiple java versions support in mono repo

ivansenic:
*This PR is a draft at the moment. Created in order to check the CI and discuss the approach.*

**What this PR does**:

Connects the Quarkus project to the parent stargate. Uses [toolchains](https://maven.apache.org/guides/mini/guide-using-toolchains.html) to compile every project in target version and run test in the same version as well.

Accomplished things:

* updates maven wrapper to latest
* CI stuff updated
* build and tests green

Current issues:
* formatter plugin does not support Java 17
* `IdempotencyAnalyzerTest` fails on CI, no idea why
* Setting up the IDE for this kind of the project setup has to figured out

Tasks pending:
* updating development quide
* revert not needed changes

Relates to #1717 

**Checklist**
- [ ] Changes manually tested


tatu-at-datastax:
Would it be possible to first do a separate PR for update of Maven wrapper from Takari (0.5.6) to Apache one (3.1.0)?
As I recall these are pretty similar and latter should work just fine for our use. I think I made similar change to some of Jackson components when someone contributed a PR for the upgrade.

Doing this would reduce diff to just actual meaningful changes.

tatu-at-datastax:
Ok I think those changes were just merged in, so all that is needed is to merge/rebase this PR from v2.0.0 to get Maven wrapper changes that way (and remove from diff).

ivansenic:
@tatu-at-datastax sure fix the wrapper first.. if you want to talk about the work done I am happy to help..

tatu-at-datastax:
@ivansenic sounds good -- yes, let's sync up. Created PR #1784 just for maven wrapper changes in the meantime.

tatu-at-datastax:
@ivansenic merged #1784 so diff is now bit smaller. Toolchains changes make sense wrt what is needed, only need to discuss how to manage `toolschains.xml` files developers have, assuming we go this route.

tatu-at-datastax:
@ivansenic I wrote up https://github.com/stargate/stargate/discussions/1790 about my failed attempts for an alternative (use of either `maven.compiler.release`, or `maven.compiler.source`/`maven.compiler.target`). I was able to get things to compile with JDK 14 but not JDK 17, and even then it seemed quite fragile. I wasn't able to get CI to work.

The only (?) other alternative I can think of would be refactoring of our repos into 2:

1. Coordinator+persistence (backend side) monorepo, with JDK 8 only (probably would just reuse existing `stargate`)
2. New JDK 17 based services (all sgv2-xxx ones)

which would allow both repos to use just one JDK. We may want to do that split anyway, but that would need bigger discussion.

For now the main open question would be if some tooling could help in creating `toolchains.xml` for developer, but otherwise this approach seems simple.

tatu-at-datastax:
@ivansenic I wonder why the CI fails here for unit tests: failed test(s) from `IdempotencyAnalyzerTest`, with:

```
Step #4 - "cassandra-4.0": [ERROR] io.stargate.db.cassandra.impl.idempotency.IdempotencyAnalyzerTest.validateIdempotencyOfQueries(String, boolean)[1]  Time elapsed: 0.033 s  <<< ERROR!
Step #4 - "cassandra-4.0": org.apache.cassandra.exceptions.InvalidRequestException: unconfigured table my_table
```

Matching line is

```
    ParsedStatement.Prepared prepare = raw.prepare(ClientState.forInternalCalls());
```

and I think (but could be wrong) that it's always for `persistence-cassandra-3.11`. But maybe just because it's the first backend run.

tatu-at-datastax:
@ivansenic Looks like errorprone setup is removed currently as well? Having tried my best to upgrade version, as well as try to make it work with other JDKs (without toolchains) I noticed it was very difficult to try to make it work (in fact, I couldn't make it work with ANY other version or changed configuration -- and configuration we use looks different from all examples I've seen).

ivansenic:
> 

@tatu-at-datastax would love to sync with you about this issue. I think pure compile target can not work imo, because of the tests.. So I still think it's toolchains or separate or separate projects in the same repo..

ivansenic:
Closing this as we will not pursue this option.

PR #1773, opened 2022-04-13
[WIP] closes #1772: reference to the google api common protos

### LINKED ISSUE #1772: Stargate gRPC protos can not be used due to the missing dependency
As I already mentioned in the #1758, the Stargate `.proto` files from the `grpc-proto` project can not be used for code generation. I am hitting this already second time. The first time with the grpc perf tests and now with the Quarkus code-generation.

```
Step #5 - "build-java17": [INFO] --- quarkus-maven-plugin:2.8.0.Final:generate-code (default) @ sgv2-docsapi ---
Step #5 - "build-java17": google/rpc/status.proto: File not found.
Step #5 - "build-java17": query.proto:19:1: Import "google/rpc/status.proto" was not found or had errors.
Step #5 - "build-java17": query.proto:391:5: "google.rpc.Status" is not defined.
Step #5 - "build-java17": schema.proto:19:1: Import "query.proto" was not found or had errors.
Step #5 - "build-java17": schema.proto:59:12: "ColumnSpec" is not defined.
Step #5 - "build-java17": schema.proto:60:12: "ColumnSpec" is not defined.
Step #5 - "build-java17": schema.proto:61:12: "ColumnSpec" is not defined.
Step #5 - "build-java17": schema.proto:70:12: "ColumnSpec" is not defined.
Step #5 - "build-java17": schema.proto:71:12: "ColumnSpec" is not defined.
Step #5 - "build-java17": schema.proto:72:12: "ColumnSpec" is not defined.
Step #5 - "build-java17": schema.proto:73:12: "ColumnSpec" is not defined.
Step #5 - "build-java17": schema.proto:101:12: "TypeSpec.Udt" is not defined.
Step #5 - "build-java17": [INFO] ------------------------------------------------------------------------
Step #5 - "build-java17": [INFO] BUILD FAILURE
Step #5 - "build-java17": [INFO] ------------------------------------------------------------------------
Step #5 - "build-java17": [INFO] Total time:  15.538 s
Step #5 - "build-java17": [INFO] Finished at: 2022-04-13T09:39:11Z
Step #5 - "build-java17": [INFO] ------------------------------------------------------------------------
Step #5 - "build-java17": [ERROR] Failed to execute goal io.quarkus.platform:quarkus-maven-plugin:2.8.0.Final:generate-code (default) on project sgv2-docsapi: Quarkus code generation phase has failed: InvocationTargetException: Failed to generate Java classes from proto files: [/workspace/sgv2-docsapi/target/protoc-protos-from-dependencies
```

The reason is the missing `google/rpc/status.proto: File not found` file. Unfortunately this import is not implicit and those files are not available (only `google/protobuf` is). I am not sure how this compiles and how this even works, but I had to clone the [api-common-protos](https://github.com/googleapis/api-common-protos) repository and create a symbolic link locally to solve this like:

```
[10:49:34] ise:stargate <ise-1717-quarkus-bootstrap>$ ls -all grpc-proto/proto/
total 40
drwxrwxr-x 2 ise ise  4096 Apr  4 15:06 .
drwxrwxr-x 5 ise ise  4096 Apr  4 15:13 ..
-rw-rw-r-- 1 ise ise  1776 Apr  4 15:06 bridge.proto
lrwxrwxrwx 1 ise ise    57 Mar 23 10:50 google -> /home/ise/workspace/open-source/api-common-protos/google/
-rw-rw-r-- 1 ise ise 19319 Mar 28 11:16 query.proto
-rw-rw-r-- 1 ise ise  3984 Apr  4 15:06 schema.proto
-rw-rw-r-- 1 ise ise  1274 Mar 17 14:28 stargate.proto
```

However, this actually makes our protos not usable. The CI fails for me now on the Quarkus project. The only reason this "works" in other Stargate V2 projects is because the dependecy is not on the protos, but on the generated classes in the `grpc-proto` project. But this is not OK.

Furthremore, I check the the other grpc clients, as I was wandering how would this work there.. Well seems that all the clients are having their own copy of the proto files:
* https://github.com/stargate/stargate-grpc-go-client/tree/main/proto
* https://github.com/stargate/stargate-grpc-node-client/tree/main/proto

And those do not reference the `google/rpc/status.proto` at all. Seems that the version of the proto files in the clients is completely different from what's currently in Stargate. Which again feels wrong.

IMO this is a major issue and a major blocker for both us and users of the grpc protos.

ivansenic:
**What this PR does**:
Potential fix for the #1772, adds `proto-google-common-protos` due to the missing `google/rpc/status.proto` reference in the `query.proto`. 

Fixes #1172

### DIFF
--- a/grpc-proto/pom.xml
+++ b/grpc-proto/pom.xml
@@ -30,6 +30,11 @@
       <groupId>io.grpc</groupId>
       <artifactId>grpc-stub</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>proto-google-common-protos</artifactId>
+      <version>2.8.2</version>
+    </dependency>
 
     <!-- Test dependencies -->
     <dependency>


ivansenic:
I am not sure if this is the best solution. Maybe we could simply copy the related google proto file and include it in our source code. We anyway depend on the specific version of it.

The alternative is to keep a dependency, but this does not really help when generating as I am forced to generate the whole model from the google common protos, instead of resolving just that one type. See https://github.com/stargate/stargate/pull/1760/commits/7bacb9e433c5a70a84d9779f7e88f447ba29e6d5 for reference usage.

I would like @mpenick or somebody with more grpc experience to say what's the best here.

ivansenic:
> Seems good to me. I couldn't think of a better way to fix this.

OK @olim7t pointed out that the added dependency is already a transitive dep of the `io.grpc:grpc-protobuf:jar`.. So this explicit definition is not needed.. Only alternative would be copying the proto file directly to our code base. Not sure if this would make it any better..

tatu-at-datastax:
Assuming this dependency works and does not add tons of other stuff, no problem. But it does like `google/rpc/status.proto` really is a small leaf value, and unlikely to ever really change in meaningful way, so
I would be fine with simply adding that file as-is to our repo.

ivansenic:
> Assuming this dependency works and does not add tons of other stuff, no problem. But it does like `google/rpc/status.proto` really is a small leaf value, and unlikely to ever really change in meaningful way, so I would be fine with simply adding that file as-is to our repo.

Yes so keeping this open until we agree if adding that file to the repo directly would be fine.

olim7t:
This can be closed now I think.

ivansenic:
> This can be closed now I think.

I am fine with closing and not doing anything about this, since the proto is available in the google jar. But the client libs would need to add the google rpc proto file directly. Which open a question why the grpc clients are not up-to-date with the public protos in Stargate.

PR #1778, opened 2022-04-14
Decouple bridge from public gRPC service (fixes #1770)

### LINKED ISSUE #1770: Decouple Bridge from gRPC service
The Stargate v2 Bridge module is currently constructed as an extension of the gRPC module

- The bridge proto files are part of the `grpc-protos` module
- The bridge implementation reuses classes from the gRPC implementation

This coupling has had some benefits in terms of enabling us to bootstrap the bridge quickly, but also some undesirable side effects:

- on multiple occasions we have had to consider carefully changes we want to make to the bridge that actually require changes to classes in the gRPC module, potentially changing behavior for public API users
- under #1763 we're potentially making changes to the gRPC protos for behavior we want to add to the bridge. Even though these would be non-breaking changes to our public API, they are still changes, and we'd need to consider downstream impacts on documentation, etc.
- we've had at least one defect that is arguably caused by bridge changes impacting gRPC behavior in downstream projects (see https://github.com/riptano/c2/issues/453 where the bridge appears to be impacting gRPC metrics reporting)

We should consider making a clean break between these two modules with separate protos and separate implementations.

olim7t:
**What this PR does**:
Lots of changes but the majority of them are trivial package name changes. In summary:
- create new `bridge-proto` module and move `bridge.proto` and `schema.proto` to it. Also copy `query.proto` because they reference it. Change package names in the proto files (this impacts the generated code, and in turn all the downstream code that references it: `sgv2-*` modules + a few integration tests).
- remove `bridge.proto` and `schema.proto` from `grpc-proto`. Revert the changes of #1763 from `query.proto`.
- remove the `bridge => grpc` dependency. Instead, duplicate all the relevant Java code in `bridge`.
- I've also removed some parent classes from `bridge`, they're not needed because the bridge doesn't have reactive implementations. Arguably this could make it a bit harder to port future fixes from `grpc`, so it's in a separate commit in case we decide to revert it.
- revert the changes of #1763 from `grpc`.

Fixes #1770

**Checklist**
- [x] Changes manually tested


sgv2-docsapi/pom.xml
```diff
@@ -33,7 +34,7 @@
   <dependencies>
     <dependency>
       <groupId>io.stargate.grpc</groupId>
```
jeffreyscarpenter, line 36: Should we make the group ID `io.stargate.bridge`?

jeffreyscarpenter (approved):
Would it be true to say that as a result of this PR the `grpc-protos` and `grpc` modules should be essentially the same as on the v1 branch?

Overall I think we are doing the right thing with this separation, but when I see duplicate copies of the `Values` class and the various `Codec` classes it does give me pause. Should we think about keeping a single copy of things that are concerned with formatting type data?

olim7t:
> Would it be true to say that as a result of this PR the `grpc-protos` and `grpc` modules should be essentially the same as on the v1 branch?

Yes.
I haven't tried to compare them but will do.

> Overall I think we are doing the right thing with this separation, but when I see duplicate copies of the `Values` class and the various `Codec` classes it does give me pause. Should we think about keeping a single copy of things that are concerned with formatting type data? (**although, once we've separated the type protos, I guess that prevents reuse of type conversion code**)

Exactly. We could try to keep a smaller subset of messages common (`TypeSpec`, `Value`...), but if we find that they have to diverge in the future we'll be back to square one. I feel like it's better to go all or nothing.

bridge/src/main/java/io/stargate/bridge/codec/DurationCodec.java
```diff
@@ -0,0 +1,90 @@
...
+    try {
+      int months = (int) VIntCoding.readVInt(in);
+      int days = (int) VIntCoding.readVInt(in);
+      long nanos = VIntCoding.readVInt(in);
+      return Value.newBuilder()
+          .setDuration(
+              QueryOuterClass.Duration.newBuilder().setMonths(months).setDays(days).setNanos(nanos))
+          .build();
+    } catch (IOException e) {
+      // cannot happen
```
tatu-at-datastax, line 69: Really? Not even for corrupt input? (yes, I know this is from existing code, not new)

tatu-at-datastax:
Yeah, it is unfortunate that so much code is duplicated. But at the same time, input and/or output types being separate ones with no inheritance (and with protobuf, it's probably difficult to try to use interfaces or anything anyway) I am not sure there is an easy way to reuse codecs or even refactor chunks.
But everyone working in codebase needs to be aware of the duplication.

So I am ok with duplication as the price for decoupling.

ivansenic:
I just glanced over this, having all that duplicated code makes a huge red alarm in my head.. Not sure if there is any alternative, but we might wanna discuss this.. Is there no chance to introduce some kind of dependency between `bridge-protos` and `grpc-protos`? This would feel natural to me. Not sure dough, did not spent too much time on this.

olim7t:
> Is there no chance to introduce some kind of dependency

The stated goal of the ticket was to not have a dependency 🙂 
I don't have a strong preference either way, but we need to accept the tradeoffs and pick one.

jeffreyscarpenter:
I confess I thought we had already merged this. We should go ahead and commit to this.

tatu-at-datastax:
Agreed with @jeffreyscarpenter : no one likes extra duplication, but right now trying to figure out some form of reuse is probably not a good use of our time. Protobuf does not have good mechanisms (afaik) for modular reuse and I think decoupling public API from internal API has some value too despite drawbacks.

PR #1779, opened 2022-04-14
Add exception mappers for grpc status runtime exception and ErrorCodes

EricBorczuk:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested


EricBorczuk:
@ivansenic I saw in the description of the ticket that an implementation was needed for `io.stargate.web.impl.JerseyViolationExceptionMapper` - but I'm not sure why that's the case, since that exception is only in dropwizard. Did you mean for there to be some other general exception that needed special handling?

ivansenic:
> @ivansenic I saw in the description of the ticket that an implementation was needed for `io.stargate.web.impl.JerseyViolationExceptionMapper` - but I'm not sure why that's the case, since that exception is only in dropwizard. Did you mean for there to be some other general exception that needed special handling?

@EricBorczuk yes I thought about the validation exception that would be raised in the Quarkus.. Not sure what that would be, maybe check the docs or the source code in Quarkus and implemented exception handlers.. we definitely do not need anything from dropwizard..

EricBorczuk:
The Exception thrown by the validations in quarkus is just the plain `javax.validation.ValidationException` - I added a mapper for it!

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/common/exception/StatusRuntimeExceptionMapper.java
```diff
@@ -22,22 +22,80 @@
...
+    Status.Code sc = grpcE.getStatus().getCode();
 
-    return switch (status.getCode()) {
-      case UNAVAILABLE -> RestResponse.status(Response.Status.BAD_GATEWAY, body);
-      case UNAUTHENTICATED -> RestResponse.status(Response.Status.UNAUTHORIZED, body);
-      default -> RestResponse.status(Response.Status.INTERNAL_SERVER_ERROR, body);
-    };
+    // Handling partly based on information from:
+    // https://developers.google.com/maps-booking/reference/grpc-api/status_codes
+    switch (sc) {
```
ivansenic, line 39: This `switch` does nothing as you are not returning anything.. This effectively always ends up going to the line `91`. 

See above should be:

```
return swicth() {
}
```

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/common/exception/StatusRuntimeExceptionMapper.java
```diff
@@ -22,22 +22,80 @@
...
+        // Start with codes we know how to handle
+      case FAILED_PRECONDITION:
```
ivansenic, line 41: Each of the cases should have a dedicated unit test.. Otherwise, mistakes like the one above are not easy to catch..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/common/exception/StatusRuntimeExceptionMapper.java
```diff
@@ -22,22 +22,80 @@
...
+      case ALREADY_EXISTS:
+      case RESOURCE_EXHAUSTED:
+      case ABORTED:
+      case OUT_OF_RANGE:
+      case INTERNAL:
+      case DATA_LOSS:
+        break;
+    }
+
+    return grpcResponse(
```
ivansenic, line 91: move to line 88 as default return for all, maybe a `default` case would make sense?

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/exception/ErrorCode.java
```diff
@@ -0,0 +1,202 @@
...
+
+  DATASTORE_TABLE_NAME_INVALID(
+      Response.Status.BAD_REQUEST,
+      "The table name contains invalid characters. Valid characters are alphanumeric and underscores."),
+
+  /** Document API. */
+  DOCS_API_GENERAL_ARRAY_LENGTH_EXCEEDED(
+      Response.Status.BAD_REQUEST,
+      String.format(
+          "Max array length of %s exceeded.", DocsApiConfiguration.DEFAULT.getMaxArrayLength())),
```
ivansenic, line 42: I am not sure that we will have `DocsApiConfiguration` anymore.. Can you please avoid using it for now and copy the values directly in the default error messages.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/config/DocsApiConfiguration.java
```diff
@@ -0,0 +1,53 @@
+package io.stargate.sgv2.docsapi.config;
+
+/** The configuration for the document API. */
+public interface DocsApiConfiguration {
```
ivansenic, line 4: remove for now

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/exception/ValidationExceptionMapper.java
```diff
@@ -0,0 +1,19 @@
...
+import javax.ws.rs.core.Response;
+import org.jboss.resteasy.reactive.RestResponse;
+import org.jboss.resteasy.reactive.RestResponse.ResponseBuilder;
+import org.jboss.resteasy.reactive.server.ServerExceptionMapper;
+
+/** Simple exception mapper for the {@link ValidationExceptionMapper}. */
+public class ValidationExceptionMapper {
+
+  @ServerExceptionMapper
+  public RestResponse<ApiError> validationException(ValidationException exception) {
```
ivansenic, line 14: are you sure this is thrown on the invalid input? I am afraid we would cover way bigger scope with this one.. Thinking that the `ResteasyReactiveViolationException` could be the one we should target. Not sure?

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/common/exception/StatusRuntimeExceptionMapperTest.java
```diff
@@ -0,0 +1,48 @@
...
+        return Response.Status.SERVICE_UNAVAILABLE;
+      case UNIMPLEMENTED:
+        return Response.Status.NOT_IMPLEMENTED;
+      default:
+        return Response.Status.INTERNAL_SERVER_ERROR;
+    }
+  }
+
+  @Test
+  public void testErrorCodeMappings() {
```
ivansenic, line 34: as said, one test for each case.. Mock the `StatusRuntimeException` if needed.

ivansenic (changes requested):
There seems to be an issue with the gRPC exception mapping.. In addition to all the comments, we did not cover the general case that's being handled here: https://github.com/stargate/stargate/blob/cecaae9b0fa029ad79986f3946d3d6cf11cd768e/restapi/src/main/java/io/stargate/web/docsapi/resources/error/ErrorHandler.java#L65-L72. Any idea how can we do this as well?

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/common/exception/StatusRuntimeExceptionMapper.java
EricBorczuk: Instead of `return switch...`, each case is doing the `return`, so I think this is working

ivansenic: ok could be actually yea.. just extend the tests then, sorry for confusion..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/exception/ValidationExceptionMapper.java
EricBorczuk: I changed to ResteasyReactiveViolationException - for some reason I wasn't able to find it anywhere in my search through quarkus code before, thanks for finding it!

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/exception/ResteasyReactiveViolationExceptionMapperTest.java
```diff
@@ -0,0 +1,22 @@
+package io.stargate.sgv2.docsapi.api.exception;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.quarkus.hibernate.validator.runtime.jaxrs.ResteasyReactiveViolationException;
```
tatu-at-datastax, line 5: Not sure we can do anything about this, but this like an internal package shaded (repackage in different Java package) by Quarkus for its own use, originally from RESTeasy's public API.
So just wondering why that would be exposed by Quarkus. If it does I guess we don't have much choice than refer to it, just seems bit off.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/exception/ValidationExceptionMapper.java
ivansenic: Maybe @Tatu is right and this was peeking into internal api.. As I wrote in the comment, was not sure myself as well.. Let's stick to the `javax.validation.ConstraintViolationException` which is a parent class of this one.. Not sure, maybe also the first approach was OK..

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/exception/ResteasyReactiveViolationExceptionMapperTest.java
ivansenic: Yes wrote it in the comment above, you might be right and this might be wrong.. Eric can map it back to the original exception or the other one which is a parent class of this one.

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/common/exception/StatusRuntimeExceptionMapperTest.java
```diff
@@ -0,0 +1,322 @@
...
+    StatusRuntimeException error = new StatusRuntimeException(Status.ABORTED);
+    RestResponse<ApiError> response = mapper.statusRuntimeException(error);
+    assertThat(response.getEntity().description().endsWith(error.getMessage())).isTrue();
+    assertThat(
+            response
+                .getEntity()
+                .description()
+                .contains(
+                    String.format(
+                        "(%s->%s)", Status.Code.ABORTED, getResponseStatus(Status.Code.ABORTED))))
```
ivansenic, line 45: nit: can we have a status directly added here, without going to that method.. It's way easier to read, this way method defines what should be the result.. Then the `getResponseStatus` can also be removed.

ivansenic (approved):
Giving approval but:

* tests are not green
* the mapping to the `ResteasyReactiveViolationException` was obviously wrong, let's map it back to the `ConstraintViolationException`

EricBorczuk:
With the ExceptionMapper, the nearest superclass of the Exception that's thrown is the mapper that's used - so I think a RuntimeException mapper should work for the generic case. Will add that!

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/exception/ResteasyReactiveViolationExceptionMapperTest.java
tatu-at-datastax: Sounds good.

PR #1781, opened 2022-04-15
Fix Values.byteBuffer to return non-empty buffer (fixes #1676)

### LINKED ISSUE #1676: Values.byteBuffer returns an empty buffer
In [Values](https://github.com/stargate/stargate/blob/master/grpc-proto/src/main/java/io/stargate/grpc/Values.java#L232-L238):
```java
  public static ByteBuffer byteBuffer(Value value) {
    checkInnerCase(value, InnerCase.BYTES);

    ByteBuffer bytes = ByteBuffer.allocate(value.getBytes().size());
    value.getBytes().copyTo(bytes);
    return bytes;
  }
```

This does not reset the position after copying, resulting in an empty buffer.

Fix:

```java
  public static ByteBuffer byteBuffer(Value value) {
    checkInnerCase(value, InnerCase.BYTES);

    return value.getBytes().asReadOnlyByteBuffer();
  }
```

olim7t:
Fixes #1676

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/grpc-proto/src/main/java/io/stargate/grpc/Values.java
+++ b/grpc-proto/src/main/java/io/stargate/grpc/Values.java
@@ -78,6 +78,10 @@ public class Values {
     return Value.newBuilder().setBytes(ByteString.copyFrom(value)).build();
   }
 
+  /**
+   * Note that this method "consumes" its argument: after the call, {@code value} will have its
+   * position set to its limit, and thus appear empty.
+   */
   public static Value of(ByteBuffer value) {
     return Value.newBuilder().setBytes(ByteString.copyFrom(value)).build();
   }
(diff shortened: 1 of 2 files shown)


grpc-proto/src/main/java/io/stargate/grpc/Values.java
```diff
@@ -232,9 +236,7 @@ public static double double_(Value value) {
   public static ByteBuffer byteBuffer(Value value) {
     checkInnerCase(value, InnerCase.BYTES);
 
-    ByteBuffer bytes = ByteBuffer.allocate(value.getBytes().size());
-    value.getBytes().copyTo(bytes);
-    return bytes;
+    return value.getBytes().asReadOnlyByteBuffer();
```
tatu-at-datastax, line 239: Makes sense, as per issue description. Would benefit from Javadoc stating it too since call sequence does not immediately suggest (at least to me) the behavior.

tatu-at-datastax (approved):

PR #1785, opened 2022-04-19
Simple approach to multi-JDK builds

tatu-at-datastax:
**What this PR does**:

Attempts to support use of different JDK targets for monorepo build without tool chains.

Related to discussion #1790.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -7,7 +7,7 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 RUN apt-get update \
     && apt-get install -y \
     openjdk-8-jdk \
-    openjdk-17-jdk \
+    openjdk-14-jdk \
     git \
     gcc \
     python-dev \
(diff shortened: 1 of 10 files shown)


tatu-at-datastax:
First minimal check in, compilation using `./mvnw -DskipTests clean package`

* Succeeds with JDK 8 and JDK 14
* Fails with JDK 17 due to Duzzt generator's issues

Would need to change CI JDK settings to verify that compilation with JDK 14 (say) works with running most tests with JDK 8; and then try JDK 14 features on one of V2 services (REST most likely).

tatu-at-datastax:
Ci passes without changes, as expected (it's using JDK 8 so why shouldn't it).
Need to see how to test use of Java 11 or 14 features, run tests with either.

ivansenic:
> Hmmh. Trying to run unit tests locally, realized one big fundamental problem here: if we use JDK 14 (etc), we simply cannot run tests of things that require JDK 8. While compilation may (and does) succeed, runtime is different: for runtime we must have JDK 8. With integration tests we could perhaps punt this to CI but not with unit tests.

Yes that's why I said that toolchains are only option to have different Java level projects in the same repo..

PR #1787, opened 2022-04-19
Bumping version for next release

olim7t:


CHANGELOG.md
```diff
@@ -1,5 +1,47 @@
 # Changelog
 
+## [v1.0.54](https://github.com/stargate/stargate/tree/v1.0.54) (2022-04-18)
+
+[Full Changelog](https://github.com/stargate/stargate/compare/v2.0.0-ALPHA-7...v1.0.54)
```
jeffreyscarpenter, line 5: This isn't right. v2 shouldn't be in the history

CHANGELOG.md
```diff
@@ -1,5 +1,47 @@
...
+
+- Bootstrap Document API V2 project with Quarkus [\#1717](https://github.com/stargate/stargate/issues/1717)
+- Extend Bridge to support per-row page state and comparable bytes [\#1715](https://github.com/stargate/stargate/issues/1715)
+
+**Merged pull requests:**
+
+- Update Maven wrapper plugin settings to simplify PR \#1769 [\#1784](https://github.com/stargate/stargate/pull/1784) ([tatu-at-datastax](https://github.com/tatu-at-datastax))
+- update the CI image to include needed v2 changes [\#1774](https://github.com/stargate/stargate/pull/1774) ([ivansenic](https://github.com/ivansenic))
+
+## [v2.0.0-ALPHA-7](https://github.com/stargate/stargate/tree/v2.0.0-ALPHA-7) (2022-04-07)
```
jeffreyscarpenter, line 17: there shouldn't be any v2 references in this changelog on master

CHANGELOG.md
olim7t: Right, something went wrong, let me check...

jeffreyscarpenter: I just don't think the tool we are using to generate changelogs is equipped to address our situation of two active release streams, so manual cleanup is required for each v1 and v2 release unfortunately. But we won't be in this situation forever.

jeffreyscarpenter (approved):

PR #1793, opened 2022-04-21
Update startup and Docker scripts for GraphQL

olim7t:
**Checklist**
- [x] Changes manually tested

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,3 +60,20 @@ ENV STARGATE_REST_PORT=8082
 COPY stargate-lib/rest/sgv2*.jar stargate-lib/rest/
 COPY ./starctl-service-rest /starctl-service-rest
 ENTRYPOINT ["./starctl-service-rest"]
+
+
+#
+# Dockerfile for building GraphQL API image
+#
+
+FROM openjdk:11-jre-slim as graphqlapi
+
+EXPOSE 8080
+
+ENV STARGATE_BRIDGE_HOST=localhost
+ENV STARGATE_BRIDGE_PORT=8091
+ENV STARGATE_GRAPHQL_PORT=8080
+
+COPY stargate-lib/graphql/sgv2*.jar stargate-lib/graphql/
+COPY ./starctl-service-graphql /starctl-service-graphql
+ENTRYPOINT ["./starctl-service-graphql"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -72,7 +72,7 @@ EXPOSE 9080
 
 ENV STARGATE_BRIDGE_HOST=localhost
 ENV STARGATE_BRIDGE_PORT=8091
-ENV STARGATE_GRAPHQL_PORT=8080
+ENV STARGATE_GRAPHQL_PORT=9080
 
 COPY stargate-lib/graphql/sgv2*.jar stargate-lib/graphql/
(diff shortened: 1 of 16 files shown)


starctl-service-graphql
```diff
@@ -0,0 +1,69 @@
...
+      exit 2
+    fi
+else
+    if [[ ! -d $STARGATE_HOME ]]; then
+      echo "STARGATE_HOME: '$STARGATE_HOME' is not a directory"
+      exit 2
+    fi
+fi
+
+eval declare -a JAVA_OPTS=("$GRAPHQL_JAVA_OPTS")
```
jeffreyscarpenter, line 43: This is a good improvement. Should we do something similar for the REST script?

docker-compose/cassandra-3.11/start_cass_311.sh
```diff
@@ -41,4 +41,4 @@ docker-compose up -d cassandra-3
 # Bring up the stargate
 
 docker-compose up -d coordinator restapi
-
+docker-compose up -d coordinator graphqlapi
```
jeffreyscarpenter, line 44: This could all be a single line i.e. 

```
docker-compose up -d coordinator restapi graphqlapi
```

docker-compose/cassandra-3.11/start_cass_311.sh
```diff
@@ -40,5 +40,4 @@ docker-compose up -d cassandra-3
 
 # Bring up the stargate
 
-docker-compose up -d coordinator restapi
-docker-compose up -d coordinator graphqlapi
+docker-compose up -d coordinator restapi coordinator graphqlapi
```
jeffreyscarpenter, line 43: Will this work with coordinator listed twice?

jeffreyscarpenter (approved):

starctl-service-rest
```diff
@@ -40,7 +40,7 @@ else
     fi
 fi
 
-eval declare -a JAVA_OPTS=("$JAVA_OPTS")
+eval declare -a JAVA_OPTS=("$REST_JAVA_OPTS")
```
tatu-at-datastax, line 43: Should similar change be made to `starctl-service-rest`, to allow different JAVA_OPTS for different services?
(and/or should there be both `JAVA_OPTS` and service-specific ones?)

olim7t: This is `starctl-service-rest`.
I made the change because that's the way it was done in the c2 repo.

docker-compose/cassandra-3.11/start_cass_311.sh
olim7t: 😳

starctl-service-rest
tatu-at-datastax: Whops. Sorry about noise, yes, you are right. That change makes sense and I somehow read this to be for graphql one.

Dockerfile
```diff
@@ -68,7 +68,7 @@ ENTRYPOINT ["./starctl-service-rest"]
 
 FROM openjdk:11-jre-slim as graphqlapi
 
-EXPOSE 8080
+EXPOSE 9080
```
tatu-at-datastax, line 71: I guess this makes sense. Is there an issue to point to wrt port change?
Should also be mentioned in readme of sgv2-graphqlapi at least.

PR #1794, opened 2022-04-22
Add utility to deal with TypeSpec instances

olim7t:
I originally did this as part of #1771. I'm moving it to an independent PR because I see other areas where it could be reused (e.g. #1726), so better merge it ASAP.

Issue #1725, opened 2022-03-28
Port Document API utility services to the V2


ivansenic:
This issue should handle all utilities that are mainly gonna be a copy-only. A small number of changes could be related to moving to the new configuration model and a new dependency injection. Test should be moved as well.

All services should be inject-able.

Target classes:

* `ConditionParser`
* `ExpressionParser`
*  `DocsShredder`
* `DocsApiUtils` (can remain as static utility class)

ivansenic:
Done in #1831.

Issue #1720, opened 2022-03-28
Implement `TableManager` based on Stargate V2 common schema handling


ivansenic:
The Document API V2 needs a `TableManager` for managing document related Cassandra tables using the Stargate V2 common schema handling. The responsibilities of this service should be:

* Create Document API table
* Drop Document API table
* Check is valid Document API table - does exists and the table schema valid
* ~Add JSON schema to an existing and valid Document API table~ (in #1824)
* ~Delete JSON schema from the existing Document API table~ (in #1824)

ivansenic:
Done in #1806.

PR #1796, opened 2022-04-23
Adapt Document API condition model to gRPC based types (fixes #1726)

### LINKED ISSUE #1726: Adapt Document API condition model to gRPC based types
The Document API V1 condition model in the `io.stargate.web.docsapi.service.query.condition` should be migrated off the `persistance-api` and implemented against the gRPC based models for result set and query conditions. All unit tests should be adapted as well.

An analysis should be done if the `ColumnSpec` is needed to perform the conversion, or a general static document API table structure would be enough.

olim7t:
Fixes #1726

**Checklist**
- [ ] Changes manually tested


sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/DocsApiConfiguration.java
```diff
@@ -0,0 +1,53 @@
+package io.stargate.sgv2.docsapi.service;
+
+/** The configuration for the document API. */
+public interface DocsApiConfiguration {
```
olim7t, line 4: Ported as-is, but this will probably be converted to Quarkus config.

grpc-proto/src/main/java/io/stargate/grpc/TypeSpecs.java
```diff
@@ -0,0 +1,458 @@
...
+import io.stargate.proto.QueryOuterClass.TypeSpec.Map;
+import io.stargate.proto.QueryOuterClass.TypeSpec.Set;
+import io.stargate.proto.QueryOuterClass.TypeSpec.Tuple;
+import io.stargate.proto.QueryOuterClass.TypeSpec.Udt;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+public class TypeSpecs {
```
olim7t, line 29: Temporarily cherry-picked from #1794 to avoid PRs on top of PRs. I'll remove the commit once #1794 gets merged.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/ExtendedRow.java
```diff
@@ -0,0 +1,73 @@
...
+package io.stargate.sgv2.docsapi.service.util;
+
+import io.stargate.grpc.Values;
+import io.stargate.proto.QueryOuterClass.ColumnSpec;
+import io.stargate.proto.QueryOuterClass.Row;
+import io.stargate.proto.QueryOuterClass.Value;
+import java.util.List;
+
+/** A wrapper around the protobuf-generated {@link Row}, to add the column metadata. */
+public class ExtendedRow {
```
olim7t, line 25: Maybe we should move this to `sgv2-service-common` and generalize its use. I've been passing the row and columns separately in GraphQL, but here it really makes sense to have a wrapper with all the predicates.

olim7t (commented):
First batch of classes since I got to a stable state.

olim7t:
Force-pushed to rebase on latest v2.0.0 (decoupled bridge + `TypeSpecs` utility).

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/RawDocument.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+package io.stargate.sgv2.docsapi.service;
+
+import io.stargate.sgv2.docsapi.service.util.ExtendedRow;
+import java.util.List;
+
+public class RawDocument {
+
+  private final List<ExtendedRow> rows;
```
olim7t, line 23: I only ported the bare minimum to make the code run.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/DocsApiUtils.java
```diff
@@ -15,14 +15,123 @@
...
+import io.stargate.sgv2.docsapi.api.exception.ErrorCodeRuntimeException;
 import io.stargate.sgv2.docsapi.service.query.DocsApiConstants;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 public final class DocsApiUtils {
```
olim7t, line 29: Same here, the original code has more methods.

olim7t (commented):
Second round of classes, I think that covers all we need for this API.

There are unit test failures due to weird Quarkus classloader errors, I think it doesn't like the fact that we both get the proto code from `bridge-proto` and regenerate it.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/DocsApiConfiguration.java
ivansenic: Yes this is available via the `DocumentProperties` class now..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/DocsApiConstants.java
```diff
@@ -0,0 +1,118 @@
...
+import io.stargate.bridge.grpc.TypeSpecs;
+import io.stargate.bridge.proto.QueryOuterClass.ColumnSpec;
+import io.stargate.bridge.proto.QueryOuterClass.TypeSpec;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+/** Constants needed in the Documents API. */
+public interface DocsApiConstants {
```
ivansenic, line 28: Also `DocumentProperties#tableProperties` can be used now.. We should not bring this back in the game.. The `DocumentProperties` can be injected anywhere..

sgv2-docsapi/pom.xml
```diff
@@ -45,6 +45,11 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>io.stargate</groupId>
+      <artifactId>sgv2-service-common</artifactId>
```
ivansenic, line 50: What's this exactly needed for here?

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/ExtendedRow.java
ivansenic: Yea wrapper makes sense, few things:

* can this also be immutable
* can this be part of some kind of the `.model` package? Also the naming makes it confusing with the extended row we implemented on the bridge, but yea.. Maybe `RowWrapper` or something like this..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/DocsApiUtils.java
ivansenic: The rest can be handled as part of the #1725 

Just not sure now how to handle the `DocsApiConstants` and the static usage here. Maybe in #1725 we are making this not static anymore..

ivansenic (commented):
Did a quick look onto this in the train.. Does look very very good.. Basically the only issue I can see is that we opted in not having the docs constants anymore as static, but configurable.. So we need to find a way to make this happen.. Maybe as a second step..

I still don't understand:
* why we need a connection to the commons project in this PR
* and what's the deal with testing and using different generated proto classes, we should only have on the classpath the one that Quarkus generated..

olim7t:
> why we need a connection to the commons project in this PR

It's needed for `i.s.s.common.cql.builder.QueryBuilder`, that we'll use to build the string queries passed to the bridge. Currently it's not used directly yet, but the code references other classes of that package, such as `BuiltCondition`.

> and what's the deal with testing and using different generated proto classes, we should only have on the classpath the one that Quarkus generated..

We have a dependency to `bridge-proto` to get the proto file, but it also contains the generated classes.
Also `sgv2-service-common` needs the generated classes (including for things that we actually use in the docs project, such as the query builder).

Regenerating the proto files in the service is a new thing, we didn't need to do that for REST and GraphQL.

sgv2-docsapi/pom.xml
```diff
@@ -152,6 +152,28 @@
...
+                  <fileset dir="${project.build.directory}/generated-sources/grpc/io/stargate/bridge/proto">
+                    <include name="QueryOuterClass.java"/>
+                    <include name="Schema.java"/>
+                  </fileset>
+                </delete>
+              </target>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
```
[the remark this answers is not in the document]
ivansenic: Can we just exclude the generated classes from the `common`, already this this for #1720 [work](https://github.com/stargate/stargate/tree/ise-1720-table-manager-schema) and all works fine, tests are green.. I am building queries, sending them over the bridge and it's all fine.. My setup is the following:

```
    <dependency>
      <exclusions>
        <exclusion>
          <groupId>*</groupId>
          <artifactId>*</artifactId>
        </exclusion>
      </exclusions>
    </dependency>
    <dependency>
      <artifactId>sgv2-service-common</artifactId>
      <exclusions>
        <exclusion>
          <groupId>*</groupId>
          <artifactId>*</artifactId>
        </exclusion>
      </exclusions>
    </dependency>
 ```
 
I get all the common classes, for sure if I use anything that depends on the Dropwizard or something else I will get a runtime error, but for now it's good.. And if we create a dedicated common for the Quarkus or framework-agnostic it would be better..

File: sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/model/RowWrapper.java [hunk 9]
```diff
@@ -0,0 +1,78 @@
...
+
+import io.stargate.bridge.grpc.Values;
+import io.stargate.bridge.proto.QueryOuterClass.ColumnSpec;
+import io.stargate.bridge.proto.QueryOuterClass.Row;
+import io.stargate.bridge.proto.QueryOuterClass.Value;
+import java.util.List;
+import java.util.Objects;
+
+/** A wrapper around the protobuf-generated {@link Row}, to add the column metadata. */
+public class RowWrapper {
```
ivansenic, line 26: * this could be immutable
* we could port this here as lazy evaluated prop: https://github.com/stargate/stargate/pull/1797/files#diff-073be9eaa3353270c2e732f3821a3d15a0f721ad06ba57b7d1797408c37dea92R59-R66

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/model/RowWrapper.java
```diff
@@ -0,0 +1,78 @@
...
+  private Value getValue(String columnName) {
+    int i = firstIndexOf(columnName);
+    if (i < 0) {
+      throw new IllegalArgumentException(String.format("Column '%s' does not exist", columnName));
+    }
+    return row.getValues(i);
+  }
+
+  /** @return the index, or <0 if the column does not exist. */
+  private int firstIndexOf(String columnName) {
```
ivansenic, line 69: could be replaced by the map lookup if we implement the suggestion above

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/RawDocument.java
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.sgv2.docsapi.service;
+
+import io.stargate.sgv2.docsapi.model.RowWrapper;
+import java.util.List;
+
+public class RawDocument {
```
ivansenic, line 21: should be in `model` package as well. quick java docs

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/ExpressionParser.java
```diff
@@ -0,0 +1,326 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+import javax.inject.Inject;
+
+public class ExpressionParser {
```
ivansenic, line 49: Should be annotated with `@ApplicationScoped` or `@Singleton` so it can be injected in other components and/or tests.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/condition/ConditionParser.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Simple service that wraps all available raw filter values and connects them to a {@link
+ * ConditionProvider}, enabling parsing of the filter nodes.
+ */
+public class ConditionParser {
```
ivansenic, line 36: also annotation for the `@ApplicationScoped` or `@Singleton`

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/query/ExpressionParserIntTest.java
```diff
@@ -0,0 +1,1085 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+// INT test in terms that it works against components and has no mocks
+// TODO ISE: move to the testing when DI allows
+@ExtendWith(MockitoExtension.class)
+class ExpressionParserIntTest {
```
ivansenic, line 64: this should be a `@QuarkusTest`

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/DocsApiUtils.java
ivansenic: Maybe we can make it non-static immediately.. Better sooner than later..

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/query/condition/ConditionParserTest.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import org.apache.commons.lang3.RandomStringUtils;
+import org.apache.commons.lang3.RandomUtils;
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+class ConditionParserTest {
```
ivansenic, line 48: also quarkus test,

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/Literal.java
```diff
@@ -15,7 +15,9 @@
  */
 package io.stargate.sgv2.common.cql.builder;
 
-class Literal implements Term {
+import java.util.Objects;
+
+public class Literal implements Term {
```
ivansenic, line 20: immutable?

ivansenic (commented):
Had a wider look, still looks very good.. We need to solve few issues:

1. That dependency to the common and the introduced classes
2. Kicking out `DocsApiConstants` (a bit of work needed)
3. Figuring out what to do with the `DocsApiUtils`, should it be not-static any more or not..
4. Make some of the tests `@QuarkusTest`

I can take over?

sgv2-docsapi/pom.xml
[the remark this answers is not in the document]
olim7t: That doesn't solve the problem, unit tests are still failing.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/RawDocument.java
olim7t: This is copied directly from the v1 code, I followed the existing package structure.

ivansenic:
Force pushed to rebase on the latest v2..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/DocsApiConstants.java
ivansenic: Done.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/DocsApiUtils.java
ivansenic: Kept is static, removed references to the constants..

ivansenic:
Did a reset author on the copied classes to keep the blame correct.. We should ensure this while porting the classes, otherwise will all get mixed up..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/DocsApiUtils.java
```diff
@@ -0,0 +1,266 @@
+/*
```
EricBorczuk, line 1: Ah good to see that this is ported here :+1:

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/common/model/RowWrapper.java
```diff
@@ -0,0 +1,109 @@
...
+   * @return Value of the column as boolean.
+   * @throws IllegalArgumentException If column does not exist.
+   */
+  default Boolean getBoolean(String columnName) {
+    return Values.bool(getValue(columnName));
+  }
+
+  /** @return Map of column names to index in the {@link #row()}. */
+  @org.immutables.value.Value.Derived
+  default Map<String, Integer> columnIndexMap() {
```
EricBorczuk, line 87: Nice :+1: this class will be straightforward to use

EricBorczuk (approved):
This looks good! Glad to see that the Utils and RowWrapper are part of this too, it lays good foundation for future work

olim7t (commented):
Can't approve my own PR, but the latest changes look good to me 👍

PR #1805, opened 2022-04-29
Revisit value handling in the query builder

olim7t:
- only allow QueryOuterClass.Value as literals
- allow explicit bind markers

### DIFF
--- a/sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/Literal.java
+++ b/sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/Literal.java
@@ -15,15 +15,20 @@
  */
 package io.stargate.sgv2.common.cql.builder;
 
-class Literal implements Term {
+import io.stargate.bridge.proto.QueryOuterClass.Value;
+import java.util.Objects;
 
-  private final Object value;
+public class Literal implements Term {
 
-  public Literal(Object value) {
-    this.value = value;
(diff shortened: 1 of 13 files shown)


sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/Literal.java
```diff
@@ -15,15 +15,20 @@
  */
 package io.stargate.sgv2.common.cql.builder;
 
-class Literal implements Term {
+import io.stargate.bridge.proto.QueryOuterClass.Value;
+import java.util.Objects;
 
-  private final Object value;
+public class Literal implements Term {
```
ivansenic, line 21: mentioned already in #1796 , this is a nice candidate for immutable..

ivansenic (approved):
Seems good to me.. Hope that tests would catch anything that I might have missed..

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/Literal.java
olim7t: Maybe we can address this in a follow-up ticket, I don't want to delay this...

But I think we are over-using `@Value.Immutable`: in this case the class is already immutable and it doesn't need value equality. So it wouldn't add any value.

olim7t: And just after writing this, I notice that I've implemented `equals` and `hashCode` on another branch... But still, it's not that hard to do manually.

PR #1806, opened 2022-05-05
relates to #1720: simple cached schema management 

ivansenic:
**What this PR does**:

Simple schema management for the Docs API V2 and Quarkus based applications in general.. Provides only keyspace fetching for now.. It's fully reactive implementation, with the support for the tenant aware keyspace cache.

Uses the built-in cache support from Quarkus that enables out-of-the-box:

* specifying of all cache properties via configuration
* metrics about cache hit rates

Small exception mapping was done in addition, in order to move to correct package and remove test warnings.

Relates to #1720

**Checklist**


ivansenic:
@olim7t This is the schema management I was talking about today.. Have a look, it's pretty simple and already tested well, reactive, has cache metric and tracing support oob.. What do you think?

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/common/SchemaManager.java
```diff
@@ -0,0 +1,213 @@
...
+  // fetches the keyspace from the bridge and caches it
+  @CacheResult(cacheName = "keyspace-cache")
+  protected Uni<Schema.CqlKeyspaceDescribe> fetchKeyspace(
+      @CacheKey String keyspaceName, @CacheKey Optional<String> tenantId, StargateBridge bridge) {
+    // create request
+    Schema.DescribeKeyspaceQuery request =
+        Schema.DescribeKeyspaceQuery.newBuilder().setKeyspaceName(keyspaceName).build();
+
+    // memoize
+    return bridge.describeKeyspace(request).memoize().indefinitely();
```
EricBorczuk, line 197: Should this be indefinite? Is it memoizing the result? What happens if someone drops a keyspace and then re-creates it with different replication or other options?

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/common/exception/BridgeAuthorizationExceptionMapperTest.java
```diff
@@ -0,0 +1,42 @@
...
+  public void happyPath() {
+    BridgeAuthorizationExceptionMapper mapper = new BridgeAuthorizationExceptionMapper();
+    BridgeAuthorizationException error = new UnauthorizedKeyspaceException("my_keyspace");
+
+    try (RestResponse<ApiError> response = mapper.bridgeAuthorizationException(error)) {
+      assertThat(response.getStatus()).isEqualTo(401);
+      assertThat(response.getEntity().description())
+          .isEqualTo("Unauthorized keyspace: my_keyspace");
+    }
+    ;
```
EricBorczuk, line 40: Is this `;` necessary?

EricBorczuk (approved):
This looks good, I did have one question re: the keyspace caching that might be important in some cases

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/common/SchemaManager.java
ivansenic: This is a memoize of the uni that's being cached, the only purpose is not to re-execute the request on the bridge once the item is pulled in the cache.. With the reactive frameworks in case of a double subscription to the same producer, the whole chain is re-executed..

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/common/exception/BridgeAuthorizationExceptionMapperTest.java
ivansenic: Of course not :smile:

PR #1807, opened 2022-05-05
Use decorated name to cache keyspaces in bridge client

olim7t:
**What this PR does**:
Fix a bug in a bridge client: keyspaces were cached under their simple name, regardless of the tenant.

/

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
+++ b/sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
@@ -69,6 +69,11 @@ class DefaultStargateBridgeClient implements StargateBridgeClient {
 
   private final Channel channel;
   private final CallOptions callOptions;
+  // A note on tenant handling: the tenant id is sent in the gRPC metadata, so the bridge will
+  // handle it transparently; client-facing methods like `executeQueryAsync()` and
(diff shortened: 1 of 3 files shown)


sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
```diff
@@ -130,7 +130,9 @@ public CompletionStage<Optional<CqlKeyspaceDescribe>> getKeyspaceAsync(
   private CompletionStage<CqlKeyspaceDescribe> getAuthorizedKeyspace(String keyspaceName) {
     CompletableFuture<CqlKeyspaceDescribe> result = new CompletableFuture<>();
 
-    CqlKeyspaceDescribe cached = keyspaceCache.getIfPresent(keyspaceName);
+    String decoratedKeyspaceName = decorateKeyspaceName(keyspaceName);
```
tatu-at-datastax, line 133: Not sure where it'd make most sense, but I think a note on why decorated keyspace name is needed locally vs not sent via bridge (since bridge side decorates it?) would be useful.
And maybe `fetchKeyspaceFromBridge()` could take `undecoratedKeyspaceName` or something to make this clear(er)?

tatu-at-datastax (approved):

PR #1810, opened 2022-05-06
Migrate write query builders to Document API v2

### LINKED ISSUE #1728: Implment `WriteBridgeService` in the Document API V2
The new `WiriteBridgeService` should be based off the existing `DocumentWriteService`. The query builder implementations should be building a gRPC-based queries. The `DataStore` should be substituted with the bridge gRPC client, but the client can become the inject-able dependency. The reactive framework should be adapted to the Quarkus default.

olim7t:
**What this PR does**:
See title.

Fixes #1728 (partially)

**Checklist**
- [x] Changes manually tested


sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/db/AbstractDeleteQueryBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+import io.stargate.sgv2.common.cql.builder.BuiltCondition;
+import io.stargate.sgv2.common.cql.builder.Predicate;
+import io.stargate.sgv2.common.cql.builder.QueryBuilder;
+import io.stargate.sgv2.common.cql.builder.Term;
+import io.stargate.sgv2.docsapi.api.common.properties.document.DocumentProperties;
+import java.util.ArrayList;
+import java.util.List;
+
+/** Shared base class for all document query builder. */
+public abstract class AbstractDeleteQueryBuilder {
```
olim7t, line 33: Do we want to convert these to `@Value.Immutable` as well?

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/write/db/DeleteDocumentQueryBuilderTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+@QuarkusTest
+@TestProfile(DeleteDocumentQueryBuilderTest.Config.class)
+class DeleteDocumentQueryBuilderTest {
+
+  private static final int MAX_DEPTH = 4;
+  private static final DocsApiTestSchemaProvider SCHEMA_PROVIDER =
+      new DocsApiTestSchemaProvider(MAX_DEPTH);
+  private static final String KEYSPACE_NAME = SCHEMA_PROVIDER.getKeyspace().getName();
+  private static final String COLLECTION_NAME = SCHEMA_PROVIDER.getTable().getName();
```
olim7t, line 44: I think eventually the cleaner solution would be to inject `DocumentProperties` into `DocsApiTestSchemaProvider`. I can make the change everywhere if we agree.

olim7t (commented):
First round of changes: migrate all the "query builder" classes referenced by the service.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/db/DeleteSubDocumentPathQueryBuilder.java
```diff
@@ -0,0 +1,80 @@
...
+  @Override
+  protected void addWhereConditions(List<BuiltCondition> whereConditions) {
+    if (subDocumentPath.size() > maxDepth) {
+      throw new ErrorCodeRuntimeException(ErrorCode.DOCS_API_GENERAL_DEPTH_EXCEEDED);
+    }
+
+    int pathSize = exactPath ? maxDepth : subDocumentPath.size();
+    // if we have the sub-document path add that as well
+    for (int i = 0; i < pathSize; i++) {
+      whereConditions.add(BuiltCondition.of("p" + i, Predicate.EQ, Term.marker()));
```
ivansenic, line 61: and here

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/db/AbstractDeleteQueryBuilder.java
ivansenic: Naah we can keep as is..

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/write/db/DeleteDocumentQueryBuilderTest.java
ivansenic: If you have an idea it would be great to optimize.. In addition, maybe this should be connected with a fixed profile that defines the `maxDepth`. Like having a single profile for smaller `maxDepth`, be 4 or whatever.. That would give us:

* less code duplication for test setup
* faster tests, as for each different test profile Quarkus have to restart the context?

ivansenic (approved):
I know it s a draft, but this is awesome and lets merge it right away.. We can add other need stuff as in the second PR..

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/write/db/DeleteDocumentQueryBuilderTest.java
olim7t: Implemented in "Refactor unit tests" commit.

olim7t:
This is in a good state if we want to merge now.
There are a couple of things that could still be improved:
- insert queries are always used in CQL batches, and delete queries all but in one place, so the builders could produce `BatchQuery` instances, in order to avoid converting later.
- `buildQuery` should also produce a proto builder (e.g. `BatchQuery.Builder`), it's not necessary to build the proto object since we're going to rebuild anyway in `bind()`. But that will require changes in `QueryBuilderImpl`.

We can address both when migrating the service.

sgv2-docsapi/pom.xml
```diff
@@ -119,6 +119,11 @@
       <artifactId>jbool_expressions</artifactId>
       <version>1.23</version>
     </dependency>
+    <dependency>
```
ivansenic, line 122: Just saw this, we are migrating to the open api v3 seagger annotations.. So it s not anymore `@ApiModelProperty`, but `@Schema` annotations. You can see example in the ExampleResource and its DTO model.. This dependency must go out.

ivansenic (commented):
Swagger V2 annotations are out of the game. Quick update?

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/models/QueryInfo.java
```diff
@@ -0,0 +1,70 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import io.swagger.annotations.ApiModelProperty;
+import org.immutables.value.Value;
+
+@JsonSerialize(as = ImmutableQueryInfo.class)
+@JsonDeserialize(as = ImmutableQueryInfo.class)
+@Value.Immutable(lazyhash = true)
+public interface QueryInfo {
+
+  @ApiModelProperty("CQL query text")
```
ivansenic, line 29: same here

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/ExecutionContext.java
```diff
@@ -0,0 +1,126 @@
...
+      executionInfoMap.merge(cql, QueryInfo.of(cql, numRows), QueryInfo::combine);
+    }
+
+    @Override
+    public void traceDeferredDml(Query query) {
+      int numRows = 0;
+
+      // TODO the query is already built, so we have no way of estimating the rows impacted
+
+      traceCqlResult(query, numRows);
```
ivansenic, line 106: Guess not from the client side :( or?

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/TimeSource.java
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.stargate.sgv2.docsapi.service.util;
+
+/** A simple interface capable of providing current time timestamps in millis and micros. */
+public interface TimeSource {
```
ivansenic, line 21: What do we need this for? Did not found a usage in this PR? Would we need to depend on it, I hope not..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/db/DeleteSubDocumentArrayQueryBuilder.java
```diff
@@ -0,0 +1,64 @@
...
+    whereConditions.add(BuiltCondition.of(targetColumns, Predicate.GTE, Term.marker()));
+    whereConditions.add(BuiltCondition.of(targetColumns, Predicate.LTE, Term.marker()));
+  }
+
+  @Override
+  protected void addBindValues(List<QueryOuterClass.Value> values) {
+    super.addBindValues(values);
+
+    values.add(Values.of("[000000]"));
+    values.add(Values.of("[999999]"));
```
ivansenic, line 62: actually this is also wrong now, this should be related to the `documentProperties.maxArrayLength() - 1`.. So if max array length is `1000`, we should have `"[999]"` here..

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/DocsApiTestSchemaProvider.java
```diff
@@ -24,48 +24,35 @@
...
 import io.stargate.sgv2.docsapi.service.common.model.ImmutableRowWrapper;
 import io.stargate.sgv2.docsapi.service.common.model.RowWrapper;
 import java.util.List;
 import java.util.stream.Stream;
+import javax.inject.Inject;
+import javax.inject.Singleton;
 import org.apache.commons.lang3.RandomStringUtils;
 
 // utility class that can construct the schema for the docs api tests
+@Singleton
```
ivansenic, line 40: Sure totally forgot the injection here is also easy in the tests.. Awesome stuff :+1:

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/testprofiles/MaxDepth4.java
```diff
@@ -0,0 +1,27 @@
...
+ *  See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package io.stargate.sgv2.docsapi.testprofiles;
+
+import io.quarkus.test.junit.QuarkusTestProfile;
+import java.util.Map;
+
+public class MaxDepth4 implements QuarkusTestProfile {
```
ivansenic, line 22: Can we call this `MaxDepth4TestProfile` and `MaxDepth8TestProfile` since all the others test profiles also have the `TestProfile` suffix?

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/util/DocsApiUtilsQuarkusTest.java
```diff
@@ -0,0 +1,274 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import javax.inject.Inject;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+@QuarkusTest
+@TestProfile(MaxDepth4.class)
+class DocsApiUtilsQuarkusTest {
```
ivansenic, line 38: do you think it's not possible to pull the other DocsApi test here?

ivansenic (approved):
Awesome stuff, thanks so much for the test improvements, way better.. Can you just make sure to rebase before merging, Eric did merge two things yesterday, and I am not sure if the tests in those two commits would break this now..

But I would say we need this asap as it brings a really nice set of updates..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/TimeSource.java
ivansenic: Naah leave it, is fine, forgot about the writetime stuff. Is the query builder for grpc supporting this?

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/util/DocsApiUtilsQuarkusTest.java
ivansenic: Yea MockitoExtension won t work with QuarkusTest. Opened the ticket there, said this is not supported and nor will be due to the classloading they do for tests.

olim7t:
Rebased on latest v2.0.0, addressed comments and removed the "Prepare other classes..." commit (I've kept it on a local branch).

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/util/DocsApiUtilsQuarkusTest.java
olim7t: Ok, that shouldn't be a big problem in practice, once you start using QuarkusTest you have the whole context and you're less likely to mock things.
And in any case it's just the extension that doesn't work, mocks can still be created explicitly with `Mockito.mock`, or MockitoAnnotations.openMocks` if we want to keep the annotations.

ivansenic: you have the `@InjectMock` support directly from Quarkus, you can mock any bean.. So the problem would only be in auto-mock-init non-bean classes or captors..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/db/DeleteSubDocumentArrayQueryBuilder.java
ivansenic: Seems fixed now..

ivansenic:
Guess this can be merged :rocket:

PR #1820, opened 2022-05-16
Mpenick/change grpc bridge tenant header

mpenick:
Changes the tenant ID header for the gRPC bridge from `host` to `tenant-id`. Using `host` causes nginx to error with `400 Bad Request` because it considers having both `:authority` and `host` as a duplicate header for "host".

No issue for this PR.

**Checklist**
- [ ] Changes manually tested (not yet)
~- [ ] Documentation added/updated~


olim7t (approved):

ivansenic (approved):
We have it as `X-Tenant-Id` in the Docs api to align it to the http headers. Also the token is passed as `X-Cassandra-Token`, so maybe nice to align. Anyway what ever it is at the end we can align.

Strange that no test updates were needed here 🤔

tatu-at-datastax:
> We have it as `X-Tenant-Id` in the Docs api to align it to the http headers. Also the token is passed as `X-Cassandra-Token`, so maybe nice to align. Anyway what ever it is at the end we can align.
> 
> Strange that no test updates were needed here 🤔

I don't think we test tenant-routing at all until maybe in CNDB?

I agree wrt naming, would be nice to align, assuming gRPC is fine with it.

mpenick:
@ivansenic Updated to `x-tenant-id` (casing doesn't matter in http2, it's a lower cased anyway).

PR #1822, opened 2022-05-16
Skip bridge authentication for GetSupportedFeatures (fixes #1821)

### LINKED ISSUE #1821: Bridge get supported features should be available without C* token
Executing the get suported features on the bridge should not require the Cassandra token. Not only that the C* auth does not make sense here, it also disables any one-time initialization process on the startup. The auth interceptor should not require the token in this case.

olim7t:
**What this PR does**:
Skip bridge authentication for GetSupportedFeatures

Fixes #1821

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/bridge/src/main/java/io/stargate/bridge/service/interceptors/NewConnectionInterceptor.java
+++ b/bridge/src/main/java/io/stargate/bridge/service/interceptors/NewConnectionInterceptor.java
@@ -2,6 +2,7 @@ package io.stargate.bridge.service.interceptors;
 
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
+import com.google.protobuf.Descriptors;
 import io.grpc.Context;
 import io.grpc.Contexts;
 import io.grpc.Grpc;
@@ -15,6 +16,7 @@ import io.grpc.Status;
 import io.stargate.auth.AuthenticationService;
 import io.stargate.auth.AuthenticationSubject;
 import io.stargate.auth.UnauthorizedException;
+import io.stargate.bridge.proto.Bridge;
 import io.stargate.bridge.service.BridgeService;
 import io.stargate.db.AuthenticatedUser;
 import io.stargate.db.ClientInfo;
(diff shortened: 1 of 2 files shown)


bridge/src/main/java/io/stargate/bridge/service/interceptors/NewConnectionInterceptor.java
```diff
@@ -77,37 +81,38 @@ public <ReqT, RespT> Listener<ReqT> interceptCall(
     try {
       Context context = Context.current();
 
-      String token = headers.get(TOKEN_KEY);
-      if (token == null) {
-        call.close(Status.UNAUTHENTICATED.withDescription("No token provided"), new Metadata());
-        return new NopListener<>();
+      if (shouldCreateConnection(call)) {
```
tatu-at-datastax, line 84: Looks like the logic for "shouldCreateConnection()" is bit heavy-weight, and in 99% of cases would return `true`, to require non-null token.
Given this, would this make sense to swap the logic to only check if connection-create is needed if `token` is `null`? To distinguish between error (rare) and GetSupportedFeatures (rare as well).

ivansenic: Branch prediction ;)

ivansenic: And that `equals` should not be heavy at all.. Maybe I am missing something..

ivansenic (approved):
Looks good..

bridge/src/main/java/io/stargate/bridge/service/interceptors/NewConnectionInterceptor.java
olim7t: I guess we could rewrite the code like this:
```java
if (token == null) {
  if (shouldCreateConnection(call)) {
    // error out and return
  }
  // fallthrough to return statement
} else {
  // create and attach connection
}
return Contexts.interceptCall(context, call, headers, next);
```
Then we are not completely skipping auth for GetSupportedFeatures, just making it optional: if a token is present, we still create a connection. It won't be used for this call but that does not really matter, since connections are cached.

But I'm not sure it's worth it: as Ivan said the check is not expensive, and I find the original version easier to follow.

PR #1833, opened 2022-05-20
Implement WriteDocumentService in Document API v2 (fixes #1728)

### LINKED ISSUE #1728: Implment `WriteBridgeService` in the Document API V2

olim7t:
**What this PR does**:
Implement WriteBridgeService in Document API v2

Fixes #1728

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/models/QueryInfo.java
+++ b/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/models/QueryInfo.java
@@ -34,7 +34,12 @@ public interface QueryInfo {
   @JsonProperty("executionCount")
   int execCount();
 
-  @Schema(description = "The total number of rows fetched by this query")
+  @Schema(
+      description =
+          "The total number of rows fetched or modified by this query. "
+              + "Note that this is underestimated when the query involves deletions: there is no "
+              + "efficient way (without a read-before-write) to count how many rows a CQL DELETE will "
+              + "remove, so it's always counted as 1.")
   @JsonProperty("rowCount")
   int rowCount();
 
--- a/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/models/QueryInfo.java
+++ b/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/models/QueryInfo.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.sgv2.docsapi.models;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+import org.immutables.value.Value;
+
+@JsonSerialize(as = ImmutableQueryInfo.class)
+@JsonDeserialize(as = ImmutableQueryInfo.class)
+@Value.Immutable(lazyhash = true)
+public interface QueryInfo {
+
+  @Schema(description = "CQL query text")
+  @JsonProperty("cql")
+  String preparedCQL();
+
+  @Schema(description = "The number of times this query was executed")
+  @JsonProperty("executionCount")
+  int execCount();
+
+  @Schema(description = "The total number of rows fetched by this query")
+  @JsonProperty("rowCount")
+  int rowCount();
+
+  static QueryInfo of(String cql, int rowCount) {
+    return of(cql, 1, rowCount);
+  }
+
+  static QueryInfo of(String cql, int execCount, int rowCount) {
+    return ImmutableQueryInfo.builder()
+        .execCount(execCount)
+        .rowCount(rowCount)
+        .preparedCQL(cql)
+        .build();
+  }
+
+  static QueryInfo combine(QueryInfo i1, QueryInfo i2) {
+    String cql = i1.preparedCQL();
+
+    if (!cql.equals(i2.preparedCQL())) {
+      throw new IllegalStateException(
+          "Unable to combine stats from different CQL queries: "
+              + cql
+              + " and "
+              + i2.preparedCQL());
+    }
+
+    return ImmutableQueryInfo.builder()
+        .preparedCQL(cql)
+        .execCount(i1.execCount() + i2.execCount())
+        .rowCount(i1.rowCount() + i2.rowCount())
+        .build();
+  }
+}
--- a/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/ExecutionContext.java
+++ b/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/ExecutionContext.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.sgv2.docsapi.service;
+
+import io.stargate.bridge.proto.QueryOuterClass.Query;
+import io.stargate.sgv2.docsapi.models.ExecutionProfile;
+import io.stargate.sgv2.docsapi.models.ImmutableExecutionProfile;
+import io.stargate.sgv2.docsapi.models.QueryInfo;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.stream.Collectors;
+
+public abstract class ExecutionContext {
+
+  public static final ExecutionContext NOOP_CONTEXT = new NoOpContext();
+
+  public static ExecutionContext create(Boolean enabled) {
+    if (Boolean.TRUE.equals(enabled)) {
+      return new ProfilingContext("root");
+    } else {
+      return NOOP_CONTEXT;
+    }
+  }
+
+  public abstract ExecutionContext nested(String description);
+
+  /** Records information about a query when its results are retrieved. */
+  public abstract void traceCqlResult(Query query, int numRows);
+
+  /**
+   * Immediately records information about a DML query that is scheduled to be executed at a later
+   * time.
+   */
+  public abstract void traceDeferredDml(Query query);
+
+  public abstract ExecutionProfile toProfile();
+
+  private static class NoOpContext extends ExecutionContext {
+    @Override
+    public void traceCqlResult(Query query, int numRows) {
+      // nop
+    }
+
+    @Override
+    public void traceDeferredDml(Query query) {
+      // nop
+    }
+
+    @Override
+    public ExecutionContext nested(String description) {
+      return this;
+    }
+
+    @Override
+    public ExecutionProfile toProfile() {
+      return null;
+    }
+  }
+
+  private static class ProfilingContext extends ExecutionContext {
+    private final Queue<ProfilingContext> steps = new ConcurrentLinkedQueue<>();
+    private final Map<String, QueryInfo> executionInfoMap = new ConcurrentHashMap<>();
+    private final String description;
+
+    private ProfilingContext(String description) {
+      this.description = description;
+    }
+
+    @Override
+    public final ExecutionContext nested(String description) {
+      ProfilingContext step = new ProfilingContext(description);
+      steps.add(step);
+      return step;
+    }
+
+    @Override
+    public void traceCqlResult(Query query, int numRows) {
+      String cql = query.getCql();
+      executionInfoMap.merge(cql, QueryInfo.of(cql, numRows), QueryInfo::combine);
+    }
+
+    @Override
+    public void traceDeferredDml(Query query) {
+      int numRows = 0;
+
+      // TODO the query is already built, so we have no way of estimating the rows impacted
+
+      traceCqlResult(query, numRows);
+    }
+
+    @Override
+    public ExecutionProfile toProfile() {
+      List<QueryInfo> queries = new ArrayList<>(executionInfoMap.values());
+      // Sort by most fetched pages (i.e. execution count), then by fetched rows
+      queries.sort(
+          Comparator.comparing(QueryInfo::execCount)
+              .thenComparing(QueryInfo::rowCount)
+              .thenComparing(QueryInfo::preparedCQL)
+              .reversed());
+
+      return ImmutableExecutionProfile.builder()
+          .description(description)
+          .queries(queries)
+          .nested(steps.stream().map(ProfilingContext::toProfile).collect(Collectors.toList()))
+          .build();
+    }
+  }
+}
--- a/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/ExecutionContext.java
+++ b/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/ExecutionContext.java
@@ -99,11 +99,11 @@ public abstract class ExecutionContext {
 
     @Override
     public void traceDeferredDml(Query query) {
-      int numRows = 0;
-
-      // TODO the query is already built, so we have no way of estimating the rows impacted
-
-      traceCqlResult(query, numRows);
+      // This method is invoked for batched DML queries. For INSERTS, we know the number of
+      // impacted rows is always 1. For DELETES, there's no way to estimate the exact number of rows
+      // (e.g. it could be a sub-document with an arbitrary number of fields), so count them as 1 as
+      // well.
+      traceCqlResult(query, 1);
     }
 
     @Override
--- a/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/TimeSource.java
+++ b/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/TimeSource.java
@@ -23,8 +23,6 @@ public interface TimeSource {
   /** Microseconds since {@link System#currentTimeMillis() epoch}. */
   long currentTimeMicros();
 
-  /**
-   * @return the current time in milliseconds
-   */
+  /** @return the current time in milliseconds */
   long currentTimeMillis();
 }
--- a/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/TimeSource.java
+++ b/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/TimeSource.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.stargate.sgv2.docsapi.service.util;
+
+/** A simple interface capable of providing current time timestamps in millis and micros. */
+public interface TimeSource {
+
+  /** Microseconds since {@link System#currentTimeMillis() epoch}. */
+  long currentTimeMicros();
+
+  /**
+   * @return the current time in milliseconds
+   */
+  long currentTimeMillis();
+}
--- a/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/TimeSourceConfiguration.java
+++ b/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/TimeSourceConfiguration.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package io.stargate.sgv2.docsapi.service.util;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.inject.Produces;
+
+public class TimeSourceConfiguration {
+
+  @Produces
+  @ApplicationScoped
+  TimeSource system() {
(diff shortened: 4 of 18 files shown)


sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/ExecutionContext.java
```diff
@@ -0,0 +1,126 @@
...
+    public void traceCqlResult(Query query, int numRows) {
+      String cql = query.getCql();
+      executionInfoMap.merge(cql, QueryInfo.of(cql, numRows), QueryInfo::combine);
+    }
+
+    @Override
+    public void traceDeferredDml(Query query) {
+      int numRows = 0;
+
+      // TODO the query is already built, so we have no way of estimating the rows impacted
```
olim7t, line 104: I'm not sure how we can implement this in the v2 architecture.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/DocumentWriteService.java
```diff
@@ -0,0 +1,115 @@
...
+              long timestamp = timeSource.currentTimeMicros();
+              return rows.stream()
+                  .map(
+                      row ->
+                          insertQueryBuilder.bind(
+                              query, documentId, row, timestamp, numericBooleans))
+                  .toList();
+            })
+        .flatMap(boundQueries -> executeBatch(bridge, boundQueries, context.nested("ASYNC INSERT")))
+    // TODO reschedule outside of gRPC executor?
```
olim7t, line 93: What is the threading model in v2? Do we want to reschedule this on a different executor?

I'm not sure if that's really necessary, unless the caller chains really compute-intensive work on it. Context switching has a cost too.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/TimeSourceConfiguration.java
```diff
@@ -0,0 +1,40 @@
...
+ *  See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package io.stargate.sgv2.docsapi.service.util;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.inject.Produces;
+
+public class TimeSourceConfiguration {
```
olim7t, line 22: I followed the model of the Properties classes here. Should I move this to a "configuration" subpackage?

olim7t (commented):
First method + set up test infrastructure

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/bridge/ValidatingStargateBridge.java
```diff
@@ -0,0 +1,171 @@
...
+    public AbstractIntegerAssert<?> assertExecuteCount() {
+      return assertThat(executeCount.get());
+    }
+
+    void executed() {
+      executeCount.incrementAndGet();
+    }
+  }
+
+  public static class QueryExpectation extends QueryAssert {
```
ivansenic, line 116: I know it's needed atm, but as you are on it, can we add a support for validation if the query was `enriched`. Needed for the reading side..

ivansenic (approved):
Know is draft, but could already approve this.. Great stuff imo..  @olim7t is getting into Quarking mode..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/DocumentWriteService.java
```diff
@@ -0,0 +1,115 @@
...
+   * @param keyspace Keyspace to store document in.
+   * @param collection Collection the document belongs to.
+   * @param documentId Document ID.
+   * @param rows Rows of this document.
+   * @param ttl the time-to-live of the rows (seconds)
+   * @param numericBooleans If numeric boolean should be stored.
+   * @param context Execution content for profiling.
+   * @return Single containing the {@link ResultSet} of the batch execution.
+   */
+  public Uni<ResultSet> writeDocument(
```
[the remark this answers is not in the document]
ivansenic: OK just saw one method for now, but guess you will be adding others later on..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/ExecutionContext.java
olim7t: So I guess I'll just remove `QueryInfo.rowCount()`.

olim7t: The v1 implementation does not return an exact count either:
```java
RowsImpacted rowsImpacted = ((BoundDMLQuery) query).rowsUpdated();
if (rowsImpacted.isKeys()) {
  numRows = rowsImpacted.asKeys().primaryKeys().size();
}
```

For inserts, this works. But also, the count is always 1.
For deletes:
- `DeleteSubDocumentKeysQueryBuilder` uses an IN clause, which only returns an exact count _if that clause covers the last clustering column_. In our model this is unlikely to happen: a subdocument based at the last _pX_could only be empty. So in practice the IN condition will be on a higher-level column, the delete will be considered a range query, and therefore we'll count 0 rows (`isKeys()) == false`).
- other deletes are always range queries, so they return 0 too.

Bottom line: we can hard-code `numRows` to 1 and be at least as accurate as v1.


<details>
<summary>v1 test code</summary>

```java
package io.stargate.db.query.builder;

import com.google.common.collect.ImmutableList;
import io.stargate.db.query.BoundDMLQuery;
import io.stargate.db.query.Predicate;
import io.stargate.db.query.RowsImpacted;
import io.stargate.db.query.TypedValue;
import io.stargate.db.schema.Column;
import io.stargate.db.schema.Schema;

public class Tmp {

  private static final Schema SCHEMA =
      Schema.build()
          .keyspace("ks")
          .table("tbl")
          .column("id", Column.Type.Text, Column.Kind.PartitionKey)
          .column("p0", Column.Type.Text, Column.Kind.Clustering)
          .column("p1", Column.Type.Text, Column.Kind.Clustering)
          .column("p2", Column.Type.Text, Column.Kind.Clustering)
          .column("v", Column.Type.Text, Column.Kind.Regular)
          .build();

  public static void main(String[] args) {
    QueryBuilder queryBuilder = new QueryBuilder(SCHEMA, TypedValue.Codec.testCodec(), null);
    BoundDMLQuery bound =
        ((BoundDMLQuery)
            queryBuilder
                .delete()
                .from("ks", "tbl")
                .where("id", Predicate.EQ)
                .where("p0", Predicate.IN)
                .build()
                .bind("id", ImmutableList.of("a", "b", "c")));
    RowsImpacted rowsImpacted = bound.rowsUpdated();

    System.out.printf(
        "IsKeys: %s, estimated count: %d%n",
        rowsImpacted.isKeys(),
        rowsImpacted.isKeys() ? rowsImpacted.asKeys().primaryKeys().size() : 0);
  }
}
```

</details>

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/DocumentWriteService.java
```diff
@@ -0,0 +1,115 @@
...
+      List<JsonShreddedRow> rows,
+      Integer ttl,
+      boolean numericBooleans,
+      ExecutionContext context) {
+
+    StargateBridge bridge = requestInfo.getStargateBridge();
+
+    return Uni.createFrom()
+        .item(() -> insertQueryBuilder.buildQuery(keyspace, collection, ttl))
+        .map(
```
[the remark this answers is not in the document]
olim7t: I'd rather introduce it as needed, otherwise it can be confusing.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/ExecutionContext.java
ivansenic: Good  investigation.. Let's do it then like that.. Can we just update the API documentation in the `QueryInfo`:

```
 @Schema(description = "The total number of rows fetched by this query")
  @JsonProperty("rowCount")
  int rowCount();
```

So mention this is approximate number of rows impacted or smth like that

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/DocumentWriteService.java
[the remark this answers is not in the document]
ivansenic: Fine with me..

ivansenic:
I see no reason not to merge this on green CI..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/DocumentWriteService.java
```diff
@@ -0,0 +1,506 @@
...
+
+              // delete existing subpath
+              AbstractDeleteQueryBuilder deleteQueryBuilder =
+                  subDocumentPath.isEmpty()
+                      ? new DeleteDocumentQueryBuilder(documentProperties)
+                      : new DeleteSubDocumentPathQueryBuilder(
+                          subDocumentPath, false, documentProperties);
+              QueryOuterClass.Query deleteQuery =
+                  deleteQueryBuilder.buildQuery(keyspace, collection);
+              queries.add(deleteQueryBuilder.bind(deleteQuery, documentId, timestamp - 1));
```
olim7t, line 195: Unlike v1, the gRPC API doesn't have an explicit concept of a prepared statement, queries are prepared automatically.
So the separation between `buildQuery()` and `bind()` feels a bit artificial, everything could be done in a single call.
The only reason would be to avoid re-building the CQL string every time, but that means we would need to cache by keyspace+collection (+extra fields for the other builders) at the class level. I'm not convinced this would make a huge difference, feels more like premature optimization.

Any thoughts?

ivansenic: Yea get it.. I would avoid caching.. We can leave it as it, or create `buildAndBind` shortcut if you like.. Although for the inser query it makes sense in creating one time and binding different values.. I guess this is a nice option..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/DocumentWriteService.java
```diff
@@ -0,0 +1,506 @@
...
+  public Uni<ResultSet> deleteDocument(
+      String keyspace,
+      String collection,
+      String documentId,
+      List<String> subDocumentPath,
+      ExecutionContext context) {
+    return deleteDocumentInternal(keyspace, collection, documentId, subDocumentPath, context);
+  }
+
+  private Uni<ResultSet> deleteDocumentInternal(
```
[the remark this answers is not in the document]
ivansenic: Perfect!

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/DocumentWriteService.java
```diff
@@ -0,0 +1,506 @@
...
+            () -> {
+              long timestamp = timeSource.currentTimeMicros();
+              AbstractDeleteQueryBuilder deleteQueryBuilder =
+                  subDocumentPath.isEmpty()
+                      ? new DeleteDocumentQueryBuilder(documentProperties)
+                      : new DeleteSubDocumentPathQueryBuilder(
+                          subDocumentPath, false, documentProperties);
+              QueryOuterClass.Query deleteQuery =
+                  deleteQueryBuilder.buildQuery(keyspace, collection);
+              return deleteQueryBuilder.bind(deleteQuery, documentId, timestamp);
```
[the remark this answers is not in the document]
ivansenic: I think all delete queries should be `timestamp - 1`, but not sure.. @EricBorczuk can you help please?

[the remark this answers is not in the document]
ivansenic: But guess does not make a difference here.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/db/DeleteDocumentQueryBuilder.java
```diff
@@ -22,7 +22,30 @@
...
+ * xyz |  a |    |    |         1
+ * xyz |  b |    |    |         2
+ * </pre>
+ *
+ * The following DELETE conditions will be generated:
+ *
+ * <pre>
+ * WHERE key = 'xyz'
+ * </pre>
+ */
```
ivansenic, line 48: LOVE THOSE :heart:

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/db/InsertQueryBuilder.java
```diff
@@ -106,6 +107,9 @@ public Query bind(
     values.addValues(row.getDoubleValue() == null ? Values.NULL : Values.of(row.getDoubleValue()));
     values.addValues(convertToBackendBooleanValue(row.getBooleanValue(), numericBooleans));
 
+    if (ttl != null) {
+      values.addValues(Values.of(ttl));
+    }
```
ivansenic, line 112: did not know that `ttl` can also be the param.. was this possible in V1 as well? Thanks for the fix..

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/bridge/AbstractBridgeTest.java
```diff
@@ -0,0 +1,53 @@
...
+import io.stargate.sgv2.docsapi.api.common.StargateRequestInfo;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+
+/**
+ * Provides an infrastructure to mock {@link StargateBridge}.
+ *
+ * <p>Subclasses must be annotated with {@link QuarkusTest}.
+ */
+public abstract class AbstractBridgeTest {
```
ivansenic, line 33: There is already a `BridgeTest` that mocks the bridge by actually firing up the grpc server.. Can we just name this different, maybe `AbstractValidatingBridgeTest`? Sorry I missed this in last review..

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/write/DocumentWriteServiceTest.java
```diff
@@ -0,0 +1,1192 @@
...
+import javax.inject.Inject;
+import org.apache.commons.lang3.RandomStringUtils;
+import org.apache.commons.lang3.RandomUtils;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+@QuarkusTest
+@TestProfile(MaxDepth4TestProfile.class)
+class DocumentWriteServiceTest extends AbstractBridgeTest {
```
ivansenic, line 53: So how many changes were needed in this test?

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/write/DocumentWriteServiceTest.java
```diff
@@ -0,0 +1,1192 @@
...
+                  Values.of(2.2d),
+                  Values.NULL,
+                  Values.of(timestamp))
+              .inBatch(expectedBatchType)
+              .returningNothing();
+
+      service
+          .writeDocument(keyspaceName, tableName, documentId, rows, null, context)
+          .await()
+          .atMost(ASYNC_WAIT_DURATION);
```
ivansenic, line 142: Would propose to use the multiny assertion:

```
 service
          .writeDocument(keyspaceName, tableName, documentId, rows, null, context)
          .subscribe()
          .withSubscriber(UniAssertSubscriber.create())
          .awaitItem()
          .assertItem(...) // optional step, if needed
          .assertCompleted();
```

The we also don't need the `ASYNC_WAIT_DURATION`

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/DocumentWriteService.java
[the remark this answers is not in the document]
olim7t: I think the idea is to order deletes before inserts when both are present in the batch (see [this](https://stackoverflow.com/questions/30317877/cassandra-batch-statement-execution-order)).

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/db/InsertQueryBuilder.java
olim7t: The v1 implementation actually turns it into a bind variable under the hood. It also supports mixing such implicit markers with explicit ones well, but in v2 that would be a bit more complicated. So it's simpler to be explicit everywhere.

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/write/DocumentWriteServiceTest.java
olim7t: Lots of small changes, for example wrapping the expected values in `Values.of()`. But they're all very simple so it's a pretty straightforward copy-paste and adapt.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/DocumentWriteService.java
```diff
@@ -0,0 +1,506 @@
...
+        Batch.newBuilder()
+            .setType(type)
+            .setParameters(
+                QueryOuterClass.BatchParameters.newBuilder()
+                    .setConsistency(
+                        QueryOuterClass.ConsistencyValue.newBuilder()
+                            .setValue(queriesConfig.consistency().writes())));
+    boundQueries.forEach(
+        query ->
+            batch.addQueries(
```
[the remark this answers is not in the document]
olim7t: On second thoughts, let's stay focused and do it in a follow-up PR.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/models/QueryInfo.java
```diff
@@ -0,0 +1,75 @@
...
+  @JsonProperty("executionCount")
+  int execCount();
+
+  @Schema(
+      description =
+          "The total number of rows fetched or modified by this query. "
+              + "Note that this is underestimated when the query involves deletions: there is no "
+              + "efficient way (without a read-before-write) to count how many rows a CQL DELETE will "
+              + "remove, so it's always counted as 1.")
+  @JsonProperty("rowCount")
```
tatu-at-datastax, line 43: Note: can just use `@JsonProperty` with no arguments -- should then use method name as-is because there is no `get` prefix. May of course be left as is for documentation purposes if that makes sense.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/TimeSource.java
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.stargate.sgv2.docsapi.service.util;
+
+/** A simple interface capable of providing current time timestamps in millis and micros. */
```
tatu-at-datastax, line 20: Many frameworks prefer use of nanos as the basic building block: is there a reason for preferring microseconds?
Nanosecond timers are usually not based off the epoch, unlike milliseconds, so conversion between nanos and milliseconds is not necessarily possible.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/configuration/TimeSourceConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+public class TimeSourceConfiguration {
+
+  @Produces
+  @ApplicationScoped
+  TimeSource system() {
+    return new TimeSource() {
+
+      @Override
+      public long currentTimeMicros() {
+        return currentTimeMillis() * 1000;
```
tatu-at-datastax, line 32: If we are to use this implementation and microseconds, we might consider instead using

    System.nanoTime() / 1000L;

since that would give us accuracy. The one caveat is that the number gotten is not compatible with `System.currentTimeMillis()` as it is not based on epoch (of `1970-01-01` UTC) so either all of these should be based on `nanoTime()` or values should not be mixed.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/db/DeleteSubDocumentArrayQueryBuilder.java
```diff
@@ -28,7 +28,34 @@
...
+ * { "a": 1, "b": [2, 3] }
+ * </pre>
+ *
+ * That gets stored as the following rows:
+ *
+ * <pre>
+ * key | p0 | p1    | p2 | dbl_value
+ * --- +----+-------+----+-----------
+ * xyz |  a |       |    |         1
+ * xyz |  b | [000] |    |         2
```
tatu-at-datastax, line 46: I am probably wrong here, but I thought index was expressed with 6 digits earlier?

tatu-at-datastax:
Looks good (from my limited understanding). Only one real comment/question: for timing, should this use nanosecond timers (from `System.timeNanos()`)? Given how fast systems have become, in many cases millisecond resolution is bit too coarse for collecting timing information. Most metrics system do this (micrometer included).

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/db/DeleteSubDocumentArrayQueryBuilder.java
olim7t: This is now configurable via `documentProperties.maxArrayLength()` (see `addBindValues` below). But yes the default is indeed 6.
I went with 3 to keep the example compact, but actually it doesn't matter, I'll change it.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/configuration/TimeSourceConfiguration.java
olim7t: `nanoTime()` can only be used to measure elapsed time, so we can't use it to represent a fixed timestamp.

With that being said, I agree that this deserves some thought; millisecond granularity sounds very coarse, and under high load many concurrent requests could end up using the same timestamp. We could borrow a few ideas from the Java driver's [monotonic generators](https://github.com/datastax/java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/internal/core/time/MonotonicTimestampGenerator.java).

I suggest we address this in a follow-up ticket.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/db/DeleteSubDocumentArrayQueryBuilder.java
tatu-at-datastax: Ok.

Configurability both makes sense and seems dangerous: mostly because changing the setting could make existing data invalid, if I understand it correctly, if changing the setting when restarting. Unless there is something in code to convert data back and forth.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/util/configuration/TimeSourceConfiguration.java
olim7t: #1850

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/db/DeleteSubDocumentArrayQueryBuilder.java
ivansenic: Configuration is needed as anyway we have different props atm for the serverless and classic for example..

ivansenic:
> System.timeNanos()

Am I missing something, [Java 17 API](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/System.html) does not have this method.. And `nanoTime()` relates to execution time of the single JVM which can not be used to mark current time..

tatu-at-datastax:
Typo, yes, it is `System.nanoTime()` not "times()".

I am not sure I understand the other comment: yes, as I also said, there is no fixed offset (value `0` has no fixed point): I guess you are saying that for some usage we would like to know "real" time of starting?
If so yes it would be problematic. It would be possible to denote start time with `currentTimeMillis()` for that use but I can see how it could lead to complicated usage and confusion.

So if we are ok with resolution of milliseconds only, that seems ok. If so, I would not add accessors for microseconds however, unless there is a way to provide true sub-millisecond resolution.

PR #1834, opened 2022-05-20
Fix #1828: Add timing diagnostics

### LINKED ISSUE #1828: Add instrumentation to SGv2 REST API for measuring timing and reporting timing breakdown
As part of on-going performance analysis of Stargate v2 REST API (and eventually other APIs), it makes sense to add a mechanism that collects timing over specific tasks of request processing. Parts like:

* Input processing
* Schema access
* DB Read access
* DB Write access
* Output processing

(note that not all requests do all of above; also note that order of some operations also vary)

Timing of these should always be collected and used to update simple (dropwizard) Metrics objects (Timers likely) but also optionally logged to output (disabled by default but enabled by a System property) for more manual analysis.
Overhead of collection and DW metrics should not add measurable overhead; logging could and is therefore only enabled when requested.

tatu-at-datastax:
**What this PR does**:

Will add metrics collection and (optional, disabled by default) outputting of timing information

Fixes #1828

**Checklist**
- [x] Changes manually tested


sgv2-service-common/src/main/java/io/stargate/sgv2/common/metrics/SchemaCacheStatsRecorder.java
```diff
@@ -0,0 +1,85 @@
...
+import com.github.benmanes.caffeine.cache.stats.CacheStats;
+import com.github.benmanes.caffeine.cache.stats.StatsCounter;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Implementation inspired by <a
+ * href="https://metrics.dropwizard.io/4.2.0/manual/caffeine.html">Dropwizart-metrics for
+ * Caffeine</a> but modified for out needs.
+ */
+public class SchemaCacheStatsRecorder implements StatsCounter {
```
ivansenic, line 16: Caffeine cache stats are supported out of the box with the Micrometer: https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java

sgv2-graphqlapi/src/main/java/io/stargate/sgv2/graphql/impl/GraphqlServiceServer.java
```diff
@@ -15,6 +15,7 @@
  */
 package io.stargate.sgv2.graphql.impl;
 
+import com.codahale.metrics.MetricRegistry;
```
ivansenic, line 18: We should definitely use Micrometer and not drop-wizard stuff.. The `MeterRegistry` should be available via the `Metrics metrics` instance. Not only it's better, but it would enable easier move to the Quarkus, as there Micrometer is used.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/metrics/ApiTimingDiagnostics.java
```diff
@@ -0,0 +1,71 @@
+package io.stargate.sgv2.common.metrics;
+
+import com.codahale.metrics.Timer;
+import java.util.function.Supplier;
+
+public class ApiTimingDiagnostics {
```
ivansenic, line 6: This is extremely annoying, we are adding another level of complexity here for collecting the metrics.. Not sure if this should be merged in this way.. 

Simple timing management could be done easily later on with [@Counted and @Timed annotations](https://quarkus.io/guides/micrometer#does-micrometer-support-annotations). For the more complicated situations, we should use what's available in Micrometer, and function based time measurement is available: https://micrometer.io/docs/concepts#_timers

ivansenic (commented):
Know it's a draft, but I already see this is going in the wrong direction. Happy to discuss more.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/metrics/ApiTimingDiagnostics.java
tatu-at-datastax: I am not super happy about intrusiveness, but at the same time I don't think annotation based approaches can easily work in providing per-call tracing style metrics. They are fine for aggregates over time, but I am also interested in specific calls.
One benefit for explicit calls is that it is pretty clear what is happening, instead of something magical that happens somewhere else.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/metrics/SchemaCacheStatsRecorder.java
tatu-at-datastax: Supporting Caffeine stats is either available via Dropwizard Metrics Caffeine plugin -- single-class -- as well.

But what I can do first, I think, is just to drop Cache instrumentation from this PR: I can see that cache hit rate is high so no need to verify. And we can tackle metrics side separately.

It does still leave the question of other metrics but at least removes one distraction.

sgv2-graphqlapi/src/main/java/io/stargate/sgv2/graphql/impl/GraphqlServiceServer.java
tatu-at-datastax: I am not against that and was wondering about Micrometer. The only concern I have is wrt how metrics are reported and accessible: right now `/metrics` endpoint is quite convenient. As long MM can be hooked similarly it does sound like a good idea to future-proof code.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/metrics/SchemaCacheStatsRecorder.java
tatu-at-datastax: Removed this class & cache instrumentation; separate issue.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/metrics/ApiTimingDiagnostics.java
tatu-at-datastax: Ok, so on Timers; this is about identical to DWM `Timer` so no objections to that part of handling (DWM->MM), as long as we can integrate MM metrics registry. At the same time use of these Timers is encapsulated within `ApiTimingDiagnostics` so it is also quite simple to change at a later point if need be.

But. This only addresses the aspect of exposing these via aggregate statistics over all calls. Like I said that is useful and all but does not help with tracing and collecting metrics for a single operation.
Passing callbacks (which I have started with) is certainly intrusive; an alternative by (ab)using `ThreadLocal`s would be ugly for other reasons.

Next thing I will be working on, and intended last part, is to make sampling of diagnostics to output configurable; disabled by default, and enabled only during testing.

tatu-at-datastax: Added sampling/sampler, and also did convert to use Micrometer instead of dropwizard-metrics. Metrics reporting is bit different and looks like we actually lose functionality by doing this (we can only calculate average and do not get percentiles? (tp95, tp99 etc)) which is unfortunate.
But since we are moving to MM anyway I guess it makes sense, and for some usage it's fine.
Would be good to know if there was a way to get percentiles

sgv2-graphqlapi/src/main/java/io/stargate/sgv2/graphql/impl/GraphqlServiceServer.java
tatu-at-datastax: Never mind above. Converted `Timer`s used to Micrometer as suggested. Thanks @ivansenic!

tatu-at-datastax:
Ok @ivansenic -- I think my initial take is ready. So, the idea is to both:

1. produce regular metrics reported via Micrometer, Prometheus, about db-read/db-write/schema-access phases of data (not schema) access points
2. optionally output sampling of per-request timing (configurable by percentage or by frequency; like 5% of traffic, or "one request sampled every 5 seconds"

I agree that part of the changes is intrusive, creation and handling of `ApiTimingDiagnotics` event around resource endpoint methods.
I tried to think of a way to avoid creating scope with callback but could not think of anything that works well. I could of course make it injected as parameter (similar to `@Context StargateBridgeClient bridge`) but not sure how to how to call matching method to close the timer scope (and smaller problem of passing name of timing sample which would probably be annotation parameter).
I think the `timed[Operation]` calls are less problematic (and probably harder to replace).
I am interested in hearing about your ideas.

The reason I think this would be useful beyond my local testing is that instrumentation should be low overhead to be enabled in production; probably with some low rate like sample every N seconds.

ivansenic (changes requested):
Let's discuss this on a call..

tatu-at-datastax:
> Let's discuss this on a call..

Will do.

ivansenic:
@tatu-at-datastax Let's close this for now.. In V2 we have tracing enabled, thus `@WithSpan` annotation can be added to measure additional times..

PR #1842, opened 2022-05-20
Enforce requested clustering column order on REST API

### LINKED ISSUE #1841: Table creation via REST API does not enforce requested order of clustering columns
The create table operation on the REST API appears to create clustering keys in the order the columns are declared, not in the order specified in the primary key declaration.

For example:

```
POST /v2/schemas/keyspaces/<your keyspace here>/tables

{
    "name": "some_table",
    "columnDefinitions": [
        {
            "name": "pk1",
            "typeDefinition": "text"
        },
        {
            "name": "ck1",
            "typeDefinition": "text"
        },
        {
            "name": "ck2",
            "typeDefinition": "text"
        }
    ],
    "primaryKey": {
        "partitionKey": [
            "pk1"
        ],
        "clusteringKey": [
            "ck2", "ck1"
        ]
    },
    "tableOptions": {
        "defaultTimeToLive": 0,
        "clusteringExpression": [
            {
                "column": "ck1",
                "order": "ASC"
            },
            {
                "column": "ck2",
                "order": "ASC"
            }
        ]
    }
}
```

When reading the table definition back via CQL, the primary key should reflect the order specified in the clustering key: `pk1, ck2, ck1`, but in fact it will read `pk1, ck1, ck2` (ordering of the clustering columns reversed).

jeffreyscarpenter:
Fixes #1841

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/restapi/src/main/java/io/stargate/web/restapi/resources/v2/schemas/TablesResource.java
+++ b/restapi/src/main/java/io/stargate/web/restapi/resources/v2/schemas/TablesResource.java
@@ -47,10 +47,12 @@ import io.swagger.annotations.ApiResponse;
 import io.swagger.annotations.ApiResponses;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


tatu-at-datastax:
From quick look I think partition key columns would also have same issue. Could add second IT method, or make this one use 2 partition and 2 clustering keys in "wrong" order (in column definitions).

jeffreyscarpenter:
> From quick look I think partition key columns would also have same issue. Could add second IT method, or make this one use 2 partition and 2 clustering keys in "wrong" order (in column definitions).

This is not an issue as ordering of partition key columns has no siginficance semantically - they are hashed together and order doesn't matter.

tatu-at-datastax:
Ah right. So may just look unexpected (ordering for creation statement), but functionally does not matter. Makes sense.

PR #1844, opened 2022-05-24
closes #1721: keyspace manager for Docs API v2

ivansenic:
**What this PR does**:

Introduced `KeyspaceManager` implementation for the Docs API v2:

* extended `SchemaManager` to support fetching of all keyspaces
* simple implementation based on the `SchemaManager`
* `@Authroized` version as well
* fully tested

**Which issue(s) this PR fixes**:
Fixes #1721

**Checklist**
- [x] Changes manually tested
- [x] Automated Tests added/updated
- [x] Documentation added/updated
- [x] Sync with #1840, depends what gets in first

PR #1840, opened 2022-05-20
closes #1722: collection resource impl

ivansenic:
*Depends on #1825.*

**What this PR does**:

Implements V2 `CollectionsResource`:
 * implements all endpoints except the collection upgrade (separate issue #1736)
 * almost all tests expect the happy path due to inability to create a keyspace at the moment
 * proper and extensive API documentation with different examples
 * extension of the `TableManager` and `SchemaManager` to handle needed features

**Which issue(s) this PR fixes**:

Relates to #1722, some tests missing in order to close the ticket.

**Checklist**
- [x] Changes manually tested
- [x] Automated Tests added/updated
- [x] Documentation added/updated
- [ ] Docs check with @polandll 
- [x] Sync with #1844 depending on what gets in first

PR #1847, opened 2022-05-24
Add JsonSchemaManager

### LINKED ISSUE #1824: Implement `JsonSchemaManager` based on Stargate V2 `TableManager`
Extracted tasks from #1720 related to the schema management:

* Add JSON schema to an existing and valid Document API table
* Get JSON schema from an existing and valid Document API table
* Delete JSON schema from the existing Document API table (maybe not needed, seems V1 does not have it)
* Check if JSON matches the schema

We could make this non-bound to the any specific `TableManager` impl and have methods like:

```
Uni<JsonNode> getJsonSchema(Uni<Schema.CqlTable> table);

Uni<Void> attachJsonSchema(Uni<Schema.CqlTable> table, JsonNode schema);

Uni<Boolean> validateJsonSchema(Uni<Schema.CqlTable> table, JsonNode document);
```

Then the components using the class could go for authorized or not authorized schema access.

**Note that caching of JSON schema is not needed anymore, as the table schemas are cached.**

EricBorczuk:
**What this PR does**:
Adds JsonSchemaManager, with some tests

Fixes #1824

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManager.java
+++ b/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManager.java
@@ -101,41 +101,6 @@ public class JsonSchemaManager {
    * Validates a JSON document against a given table's schema
    *
    * @param table the table that has a schema
-   * @param document the document, as stringified JSON
-   * @return a Uni with Boolean detailing whether or not the document complies with the schema.
-   */
-  public Uni<Boolean> validateJsonDocument(Uni<Schema.CqlTable> table, String document) {
-    return table
-        .onItem()
-        .ifNotNull()
-        .transform(
-            t -> {
-              String comment = t.getOptionsMap().getOrDefault("comment", null);
-              if (comment == null) {
-                // If there is no valid schema, then the document is valid
-                return true;
-              }
-              JsonNode jsonSchema;
-              try {
-                jsonSchema = objectMapper.readTree(comment);
-              } catch (JsonProcessingException e) {
-                // If there is no valid schema, then the document is valid
-                return true;
-              }
-
-              try {
-                validate(jsonSchema, document);
-              } catch (ProcessingException e) {
-                return true;
-              }
-              return true;
-            });
-  }
-
-  /**
-   * Validates a JSON document against a given table's schema
-   *
-   * @param table the table that has a schema
    * @param document the document, as JsonNode
    * @return a Uni with Boolean detailing whether or not the document complies with the schema.
    */
@@ -145,7 +110,6 @@ public class JsonSchemaManager {
         .ifNotNull()
         .transform(
             t -> {
-              System.out.println("the value2: " + t.getOptionsMap().get("comment"));
               String comment = t.getOptionsMap().getOrDefault("comment", null);
               if (comment == null) {
                 // If there is no valid schema, then the document is valid
@@ -166,17 +130,6 @@ public class JsonSchemaManager {
               }
               return true;
             });
-  }
-
-  private void validate(JsonNode schema, String value) throws ProcessingException {
-    final JsonNode tree;
-    try {
-      tree = objectMapper.readTree(value);
-    } catch (JsonProcessingException e) {
-      throw new ErrorCodeRuntimeException(
-          ErrorCode.DOCS_API_INVALID_JSON_VALUE, "Malformed JSON object found during read: " + e);
-    }
-    validate(schema, tree);
   }
 
   private void validate(JsonNode schema, JsonNode jsonValue) throws ProcessingException {
--- a/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManager.java
+++ b/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManager.java
@@ -21,6 +21,8 @@ import java.util.Iterator;
 import java.util.List;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Json Schema manager provides basic operations to store, retrieve, and use the JSON schema for a
@@ -28,6 +30,8 @@ import javax.inject.Inject;
  */
 @ApplicationScoped
 public class JsonSchemaManager {
+  private static final Logger logger = LoggerFactory.getLogger(JsonSchemaManager.class);
+
   @Inject ObjectMapper objectMapper;
 
   @Inject JsonSchemaQueryProvider jsonSchemaQueryProvider;
@@ -57,6 +61,7 @@ public class JsonSchemaManager {
               try {
                 return objectMapper.readTree(comment);
               } catch (JsonProcessingException e) {
+                logger.warn("Document table has comment, but it's not a valid JSON.");
                 return null;
               }
             });
@@ -105,28 +110,19 @@ public class JsonSchemaManager {
    * @return a Uni with Boolean detailing whether or not the document complies with the schema.
    */
   public Uni<Boolean> validateJsonDocument(Uni<Schema.CqlTable> table, JsonNode document) {
-    return table
-        .onItem()
-        .ifNotNull()
-        .transform(
-            t -> {
-              String comment = t.getOptionsMap().getOrDefault("comment", null);
-              if (comment == null) {
-                // If there is no valid schema, then the document is valid
-                return true;
-              }
-              JsonNode jsonSchema;
-              try {
-                jsonSchema = objectMapper.readTree(comment);
-              } catch (JsonProcessingException e) {
-                // If there is no valid schema, then the document is valid
+    return getJsonSchema(table)
+        .map(
+            jsonSchema -> {
+              if (jsonSchema == null) {
+                // If there is no valid JSON schema, then the document is valid
                 return true;
               }
 
               try {
                 validate(jsonSchema.get("schema"), document);
               } catch (ProcessingException e) {
-                return true;
+                throw new ErrorCodeRuntimeException(
+                    ErrorCode.DOCS_API_JSON_SCHEMA_PROCESSING_FAILED);
               }
               return true;
             });
--- a/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManager.java
+++ b/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManager.java
@@ -1,0 +1,96 @@
+package io.stargate.sgv2.docsapi.service.schema;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.github.fge.jsonschema.core.exceptions.ProcessingException;
+import com.github.fge.jsonschema.core.report.ProcessingMessage;
+import com.github.fge.jsonschema.core.report.ProcessingReport;
+import com.github.fge.jsonschema.main.JsonSchemaFactory;
+import io.smallrye.mutiny.Uni;
+import io.stargate.bridge.proto.QueryOuterClass;
+import io.stargate.bridge.proto.StargateBridge;
+import io.stargate.sgv2.docsapi.api.common.StargateRequestInfo;
+import io.stargate.sgv2.docsapi.api.exception.ErrorCode;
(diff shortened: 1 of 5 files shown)


sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManagerTest.java
```diff
@@ -0,0 +1,188 @@
+package io.stargate.sgv2.docsapi.service.schema;
```
EricBorczuk, line 1: Is there a way to test the exact query that is returning from the observed result? Also what other cases should I test

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManager.java
```diff
@@ -0,0 +1,96 @@
...
+public class JsonSchemaManager {
+  @Inject ObjectMapper objectMapper;
+
+  @Inject JsonSchemaQueryProvider jsonSchemaQueryProvider;
+
+  @Inject StargateRequestInfo requestInfo;
+
+  private final JsonSchemaFactory jsonSchemaFactory = JsonSchemaFactory.byDefault();
+
+  public Uni<JsonNode> getJsonSchema(TableManager tableManager, String keyspace, String table) {
```
EricBorczuk, line 38: I chose to use the TableManager and keyspace/table strings here, rather than a CqlTable, mostly to get the benefit of using authorization through the TableManager that is provided; also it might be more direct to just fetch the json schema out of the table directly rather than fetching the table and then fetching its comment...

But maybe just accessing the `comment` on a provided CqlTable works, too? Kind of makes a bunch of these methods very close to doing nothing though...

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/query/JsonSchemaQueryProvider.java
```diff
@@ -0,0 +1,42 @@
+package io.stargate.sgv2.docsapi.service.schema.query;
+
+import io.stargate.bridge.proto.QueryOuterClass;
+import io.stargate.sgv2.common.cql.builder.QueryBuilder;
+import io.stargate.sgv2.docsapi.config.QueriesConfig;
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
+@ApplicationScoped
+public class JsonSchemaQueryProvider {
```
ivansenic, line 10: add a test for this class (see examples in #1840 and #1844).. this is also how you can confirm query is correct

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManagerTest.java
ivansenic: There are examples in the #1840 and #1844.. I test query strings in the dedicated query provider test and then check there here that the query executed on the bridge as same as the one that a query provider would return..

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManagerTest.java
```diff
@@ -0,0 +1,188 @@
...
+          .assertItem(objectMapper.readTree(keyspace.getTables(0).getOptionsMap().get("comment")))
+          .assertCompleted();
+
+      verify(bridgeService).describeKeyspace(any(), any());
+      verifyNoMoreInteractions(bridgeService);
+    }
+  }
+
+  @Nested
+  class ValidateJsonSchema {
```
ivansenic, line 155: missing cases:

* no schema in table
* schema validation fails (object not complied with schema)

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManagerTest.java
```diff
@@ -0,0 +1,188 @@
...
+          .withSubscriber(UniAssertSubscriber.create())
+          .awaitItem();
+
+      verify(bridgeService).executeQuery(any(), any());
+      verifyNoMoreInteractions(bridgeService);
+    }
+  }
+
+  @Nested
+  class GetJsonSchema {
```
ivansenic, line 118: missing cases:

* no schema in table
* comment not a valid schema (imo should behave as no schema on table, or?)

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManagerTest.java
```diff
@@ -0,0 +1,188 @@
...
+            .addAllClusteringKeyColumns(clusteringColumns)
+            .addAllColumns(valueColumns);
+    return Schema.CqlKeyspaceDescribe.newBuilder()
+        .setCqlKeyspace(Schema.CqlKeyspace.newBuilder().setName(namespace))
+        .addTables(table)
+        .build();
+  }
+
+  @Nested
+  class AttachJsonSchema {
```
ivansenic, line 86: missing cases:

* schema validation fails

ivansenic (changes requested):
Looks very good.. Added requests improvements and changes..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManager.java
```diff
@@ -0,0 +1,96 @@
...
+
+  public Uni<QueryOuterClass.Response> attachJsonSchema(
+      String keyspace, String table, JsonNode schema) {
+    ProcessingReport report = jsonSchemaFactory.getSyntaxValidator().validateSchema(schema);
+    if (report.isSuccess()) {
+      ObjectNode wrappedSchema = objectMapper.createObjectNode();
+      wrappedSchema.set("schema", schema);
+      StargateBridge bridge = requestInfo.getStargateBridge();
+      return bridge.executeQuery(
+          jsonSchemaQueryProvider.attachSchemaQuery(keyspace, table, wrappedSchema.toString()));
```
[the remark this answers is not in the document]
EricBorczuk: toString() should be the equivalent of `write()` with pretty=false (so long as wrappedSchema is guaranteed non-null)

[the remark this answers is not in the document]
ivansenic: Hmm yes OK you are correct. But have a look on the JavaDocs:

```
Method that will produce (as of Jackson 2.10) valid JSON using default settings of databind, as String. If you want other kinds of JSON output (or output formatted using one of other Jackson-supported data formats) make sure to use ObjectMapper or ObjectWriter to serialize an instance, for example:
    String json = objectMapper.writeValueAsString(rootNode);
```

I hope that "default settings of databind" would work once we read it back..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManager.java
```diff
@@ -0,0 +1,190 @@
...
+            t -> {
+              String comment = t.getOptionsMap().getOrDefault("comment", null);
+              if (comment == null) {
+                return null;
+              }
+
+              try {
+                return objectMapper.readTree(comment);
+              } catch (JsonProcessingException e) {
+                return null;
```
ivansenic, line 60: maybe also a warn here for the logs, smth like `Document table has comment, but it's not a valid JSON schema`? With the exception of course..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManager.java
```diff
@@ -0,0 +1,190 @@
...
+              }
+
+              try {
+                validate(jsonSchema.get("schema"), document);
+              } catch (ProcessingException e) {
+                return true;
+              }
+              return true;
+            });
+  }
```
ivansenic, line 168: wait wait wait, why not going through the `getJsonSchema`, we don't want to double the same code:

```java
return getJsonSchema(table)
    .map(schema -> {
        if (null == schema) {
           return true:
        }     
        
       try {
          validate(jsonSchema.get("schema"), document);
        } catch (ProcessingException e) {
          return true;
        }
        return true;
    }
```

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManager.java
```diff
@@ -0,0 +1,190 @@
...
+
+  /**
+   * Validates a JSON document against a given table's schema
+   *
+   * @param table the table that has a schema
+   * @param document the document, as stringified JSON
+   * @return a Uni with Boolean detailing whether or not the document complies with the schema.
+   */
+  public Uni<Boolean> validateJsonDocument(Uni<Schema.CqlTable> table, String document) {
```
ivansenic, line 107: would remove this options, caller is responsible to convert to `JsonNode`, there is no reason to have this at all

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManager.java
```diff
@@ -0,0 +1,190 @@
...
+  private void validate(JsonNode schema, String value) throws ProcessingException {
+    final JsonNode tree;
+    try {
+      tree = objectMapper.readTree(value);
+    } catch (JsonProcessingException e) {
+      throw new ErrorCodeRuntimeException(
+          ErrorCode.DOCS_API_INVALID_JSON_VALUE, "Malformed JSON object found during read: " + e);
+    }
+    validate(schema, tree);
+  }
```
ivansenic, line 179: remove, only validate `JsonNode`

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManager.java
```diff
@@ -0,0 +1,190 @@
...
+
+  private void validate(JsonNode schema, JsonNode jsonValue) throws ProcessingException {
```
ivansenic, line 181: so what's with the processing exception, should we catch that in `validateJsonDocument` and report with some kind of the `ErrorCode`? Or is the intention to end up in the internal server error? That would be fine as well..

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManagerTest.java
```diff
@@ -0,0 +1,220 @@
...
+                    + "      \"description\": \"Product's price\"\n"
+                    + "    }\n"
+                    + "  },\n"
+                    + "  \"required\": [\"id\", \"name\", \"price\"]\n"
+                    + "}}")
+            .build();
+
+    JsonNode document = objectMapper.readTree("{\"id\":1, \"price\":1}");
+
+    assertThatThrownBy(
```
ivansenic, line 212: why is this expecting to throw something? `.awaitFailure().assertFailure()`?

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManagerTest.java
```diff
@@ -0,0 +1,220 @@
...
+
+      result.awaitItem().assertItem(null).assertCompleted();
+    }
+  }
+
+  @Nested
+  class ValidateJsonSchema {
+    @Test
+    public void happyPath() throws JsonProcessingException {
+      table = Schema.CqlTable.newBuilder().putOptions("comment", "{\"schema\": {}}").build();
```
ivansenic, line 150: is empty schema the best for validation, why not taking the same one from the line `185`?

ivansenic (approved):
Looks good, some advice on the code de-duplication and removal of unneeded stuff :rocket:

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/schema/JsonSchemaManager.java
EricBorczuk: Hmm this probably should bubble up into some ErrorCode...I'll change it

ivansenic: I guess we should check on V1 here!

ivansenic:
Green..

image

EricBorczuk:
Once I stopped being a doofus and figured out the problem by debugging the right way, it was so easy! xD

PR #1852, opened 2022-05-25
closes #1839: api error response in case of a missing token

### LINKED ISSUE #1839: Missing token should provide a response with body
In Docs API V2 the request with missing token will end in `401` but will have no body. We should:

* add a valid body similar to V1
* add a generic test for the protected paths (similar as the `TenantRequestMetricsFilterTest`)

ivansenic:
**What this PR does**:

Responds with `ApiError` in case of the missing token. In addition fixes the wrong `/api/v2` to `/v2`..

Fixes #1839.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/common/security/HeaderBasedAuthenticationMechanism.java
+++ b/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/common/security/HeaderBasedAuthenticationMechanism.java
@@ -17,6 +17,8 @@
 
 package io.stargate.sgv2.docsapi.api.common.security;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.quarkus.security.identity.IdentityProviderManager;
 import io.quarkus.security.identity.SecurityIdentity;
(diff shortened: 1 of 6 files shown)


sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/common/security/HeaderBasedAuthenticationMechanism.java
```diff
@@ -70,6 +79,51 @@ public Uni<ChallengeData> getChallenge(RoutingContext context) {
...
+              context.response().setStatusCode(status);
+
+              // create the response
+              String message =
+                  "Role unauthorized for operation: Missing token, expecting one in the %s header."
+                      .formatted(headerName);
+              ApiError apiError = new ApiError(message, status);
+              try {
+                // try to serialize
+                String response = objectMapper.writeValueAsString(apiError);
```
tatu-at-datastax, line 103: Not sure if it is easy/possible, but serializing as `byte[]` (writeValueAsBytes()) is bit more efficient than writing into intermediate `String`. Fastest would be writing directly into `OutputStream` but that's perhaps not available.

tatu-at-datastax (approved):

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/common/security/HeaderBasedAuthenticationMechanism.java
ivansenic: Good point.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/common/security/HeaderBasedAuthenticationMechanism.java
```diff
@@ -70,6 +79,51 @@ public Uni<ChallengeData> getChallenge(RoutingContext context) {
...
+                context
+                    .response()
+                    .headers()
+                    .set(HttpHeaders.CONTENT_LENGTH, String.valueOf(response.getBytes().length));
+
+                // write and map to true
+                return Uni.createFrom()
+                    .completionStage(
+                        context.response().write(response).map(true).toCompletionStage());
+              } catch (JsonProcessingException e) {
```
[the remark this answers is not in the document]
ivansenic: Sure. Done.

PR #1853, opened 2022-05-25
closes #2068: Enable gRPC reflection

### LINKED ISSUE #2068: Enable gRPC Server Reflection to help with API Consumer on-boarding
To help API Consumers on-boarding to Stargate's gRPC service, it would be helpful to enable gRPC Server Reflection.  This allows client tools to pull the proto files at runtime.  More info here: https://github.com/grpc/grpc-java/blob/master/documentation/server-reflection-tutorial.md#enable-server-reflection.

This came up while exploring Stargate's gRPC service using Postman.  Server Reflection offers a much faster time to first API call, especially among developers new to gRPC.

mpenick:
**What this PR does**:

Enables gRPC reflection. This is useful for gRPC tools that can utilize reflection information
e.g. Postman.

**Checklist**
- [X] Changes manually tested

### DIFF
--- a/grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
+++ b/grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
@@ -40,8 +40,7 @@ import org.slf4j.LoggerFactory;
 public class GrpcImpl {
   private static final Logger logger = LoggerFactory.getLogger(GrpcImpl.class);
   private static final Integer EXECUTOR_SIZE = Integer.getInteger("stargate.grpc.executor_size", 8);
-  private static final Boolean REFLECTION_ENABLED =
-      "true".equalsIgnoreCase(System.getProperty("stargate.grpc.reflection_enabled", "true"));
+  private static final Boolean REFLECTION_ENABLED = Boolean.parseBoolean(System.getProperty("stargate.grpc.reflection_enabled", "true"));
   private static final Integer SHUTDOWN_TIMEOUT_SECONDS =
       Integer.getInteger("stargate.grpc.shutdown_timeout_seconds", 60);
 
--- a/grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
+++ b/grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
@@ -40,7 +40,8 @@ import org.slf4j.LoggerFactory;
 public class GrpcImpl {
   private static final Logger logger = LoggerFactory.getLogger(GrpcImpl.class);
   private static final Integer EXECUTOR_SIZE = Integer.getInteger("stargate.grpc.executor_size", 8);
-  private static final Boolean REFLECTION_ENABLED = Boolean.parseBoolean(System.getProperty("stargate.grpc.reflection_enabled", "true"));
+  private static final Boolean REFLECTION_ENABLED =
+      Boolean.parseBoolean(System.getProperty("stargate.grpc.reflection_enabled", "true"));
   private static final Integer SHUTDOWN_TIMEOUT_SECONDS =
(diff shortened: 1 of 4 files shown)


grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
```diff
@@ -77,6 +78,7 @@ public GrpcImpl(
             .intercept(
                 new TaggingMetricCollectingServerInterceptor(
                     metrics.getMeterRegistry(), grpcMetricsTagProvider))
+            .addService(ProtoReflectionService.newInstance())
```
mpenick, line 81: Maybe make this optional?

mpenick:
This can be merged after a review.

grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
ivansenic: Done. Enabled by default.

grpc/src/main/java/io/stargate/grpc/service/interceptors/NewConnectionInterceptor.java
```diff
@@ -73,6 +74,11 @@ public NewConnectionInterceptor(
   @Override
   public <ReqT, RespT> Listener<ReqT> interceptCall(
       ServerCall<ReqT, RespT> call, Metadata headers, ServerCallHandler<ReqT, RespT> next) {
+    // Allow reflection calls through without authentication
+    if (call.getMethodDescriptor().getServiceName() != null
+        && call.getMethodDescriptor().getServiceName().equals(ServerReflectionGrpc.SERVICE_NAME)) {
+      return next.startCall(call, headers);
+    }
```
ivansenic, line 81: anybody having an idea if this would be in any case a security issue?

grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
```diff
@@ -39,6 +40,8 @@
 public class GrpcImpl {
   private static final Logger logger = LoggerFactory.getLogger(GrpcImpl.class);
   private static final Integer EXECUTOR_SIZE = Integer.getInteger("stargate.grpc.executor_size", 8);
+  private static final Boolean REFLECTION_ENABLED =
+      "true".equalsIgnoreCase(System.getProperty("stargate.grpc.reflection_enabled", "true"));
```
tatu-at-datastax, line 44: In other places I think we simply use " Boolean.parseBoolean("...", "true");". It's bit inconsistent to allow case-insensitiveness here. Up to you if you want to leave it as-is.

grpc/src/main/java/io/stargate/grpc/service/interceptors/NewConnectionInterceptor.java
tatu-at-datastax: I cannot think of obvious significant problems.

What I can think of is that allowing reflection does give out more information. It could perhaps also allow DoS approach if reflection calls are much more expensive to handle than failing authentication.
But I don't see either of these as a practical problem. And at least we have easy switch for turning it off if necessary.

tatu-at-datastax:
Nice simplification of tests @ivansenic! Looks good.

grpc/src/main/java/io/stargate/grpc/service/interceptors/NewConnectionInterceptor.java
ivansenic: Good points..

grpc/src/main/java/io/stargate/grpc/impl/GrpcImpl.java
ivansenic: Ok, updated..

PR #1854, opened 2022-05-25
Change REST API module and artifact names to "sgv2-restapi" for Stargate V2

tatu-at-datastax:
**What this PR does**:

Changes module name and artifact id (jar base name) used for REST API in Stargate V2: convention here is to add prefix of `sgv2-` to indicate Stargate major version; but since API itself does not change (and no version indicated with Stargate V1), leave that part as-is.
So basically change from

    sgv2-rest-service

to

    sgv2-restapi 

Graphql already uses this approach; Documents API was bundled with REST API in SGv1 do so that is slightly different but will likely follow this convention as well.

This change needs to be done now before we start installing things: earlier ALPHA releases will have old name but that should be ok.

No issue filed.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/sgv2-restapi/README.md
+++ b/sgv2-restapi/README.md
@@ -14,11 +14,11 @@ and uses standard DropWizard configuration approach.
 
 Project produces a single runnable uber/fat jar like:
 
-    sgv2-restapi/target/sgv2-rest-service-2.0.0-ALPHA-3-SNAPSHOT.jar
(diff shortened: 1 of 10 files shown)


olim7t:
Maybe also align the `artifactId` in the POM?

tatu-at-datastax:
> Maybe also align the `artifactId` in the POM?

Yes, thank you!

jeffreyscarpenter (approved):

Issue #1737, opened 2022-03-28
Independent end-to-end tests in Document API V2


ivansenic:
Currently all integration tests are implemented in the `testing` project, which makes thigh coupling between a micro-service and that project. Moving to the new framework enables us to have dedicated end-to-end tests directly in the new Document API microservice. This can be achieved with `@QuarkusIntegrationTest` and test containers, where all dependencies to execute the end-to-end test would be started using the test containers.

Further analysis is needed to understand if this option is valid choice and if it meets all the requirements we have.

ivansenic:
Two issues remain here open:

1. Currently enabling a profile `-P dse` deactivates the integration tests 
2. Running the integration tests on each PR.. We would most likely need to implement a dedicated github action
3. Figure out how to setup a default keyspace for each int tests that can support both native and non-native tests. Most likelt has to be a part of the `StargateTestResource`. See https://github.com/stargate/stargate/blob/v2.0.0/sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/CollectionsResourceIntegrationTest.java#L55-L59 and https://github.com/quarkusio/quarkus/issues/7690#issuecomment-596543310 for more information.

ivansenic:
In order to close this issue, let's update to the Quarkus `2.10.2` and solve the clean-up problem we had in the int tests.. Remove the `TODO` as we have a dedicated ticket now for end-to-end native image tests -> #1955

ivansenic:
Finalized in #1957.

PR #1859, opened 2022-05-30
closes #1724: namespace resource

### LINKED ISSUE #1724: Namespace REST Controller for Document API V2
Implement a V2 version of the `NamespacesResource`. As with all REST controllers in the new versions, we should:

* Have fully reactive implementation
* Have retires for `GET` operations
* Have a strong-typed DTO model
* Validates DTOs
* Have a well document API using the Open API v3

*Note that getBuiltInFunctions operation should be implemented in the separate built-in functions issue.*

ivansenic:
**What this PR does**:
Implementation of the namespace resource for the Docs API V2. Note that integration tests are created from scratch as they were not existing in the V1. Also, API documentation is corrected.

Fixes #1724
 
**Checklist**
- [x] Changes manually tested
- [x] Add mockito extension for namespace creation -> moved to https://github.com/stargate/stargate/issues/1737#issuecomment-1141110000
- [x] Resolve `TODO`s @EricBorczuk


sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/schemas/namespaces/NamespacesResource.java
```diff
@@ -0,0 +1,404 @@
...
+                    },
+                    schema =
+                        @org.eclipse.microprofile.openapi.annotations.media.Schema(
+                            implementation = ApiError.class))),
+        @APIResponse(ref = OpenApiConstants.Responses.GENERAL_401),
+        @APIResponse(ref = OpenApiConstants.Responses.GENERAL_500),
+        @APIResponse(ref = OpenApiConstants.Responses.GENERAL_503),
+      })
+  @GET
+  @Path("/{namespace:\\w+}")
```
EricBorczuk, line 183: I'm assuming this `\w` refers to the Regex class [A-Za-z0-9]+, which would not match certain characters that are technically allowed to be in keyspaces: [see here](https://docs.datastax.com/en/cql-oss/3.x/cql/cql_reference/ref-lexical-valid-chars.html#:~:text=characters%20in%20names-,Keyspace%20and%20table%20names%20can%20only%20contain%20alpha%2Dnumber%20characters,alpha%2Dnumeric%20characters%20and%20underscores.)

For instance, a keyspace like `eric-keyspace` is allowed as long as the value is enclosed in double-quotes (even though hyphen is a character that would not be matched by `\w`)

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/schemas/namespaces/model/dto/NamespaceDto.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import java.util.List;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+/** DTO for namespace datacenter. */
+public record NamespaceDto(
+
+    // TODO should we return replicas count in case of the simple strategy
```
EricBorczuk, line 28: I think we could return number of replicas - since it's just additional data and would be non-breaking, nothing wrong with that

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/schemas/namespaces/NamespacesResource.java
```diff
@@ -0,0 +1,404 @@
...
+              }
+              return false;
+            })
+        .recoverWithUni(
+            () ->
+                namespaceManager
+                    .createNamespace(namespaceName, getReplication(body))
+                    .map(
+                        created -> {
+                          // TODO should we also ship back the datacenters?
```
EricBorczuk, line 301: Same as with replica count, I think it's perfectly fine to add this non-breaking change in

EricBorczuk (approved):
This looks good - see my comment about using `\w`

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/schemas/namespaces/NamespacesResource.java
```diff
@@ -0,0 +1,404 @@
...
+  @Parameters(
+      value = {
+        @Parameter(name = "raw", ref = OpenApiConstants.Parameters.RAW),
+      })
+  @APIResponses(
+      value = {
+        @APIResponse(
+            responseCode = "200",
+            description =
+                "Call successful. Note that in case of unwrapping (`raw=true`), the response contains only the contents fof the `data` property.",
```
mpenick, line 102: fof --> of

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/schemas/namespaces/NamespacesResource.java
```diff
@@ -0,0 +1,404 @@
...
+            ref = OpenApiConstants.Parameters.NAMESPACE,
+            description = "The namespace name."),
+        @Parameter(name = "raw", ref = OpenApiConstants.Parameters.RAW),
+      })
+  @APIResponses(
+      value = {
+        @APIResponse(
+            responseCode = "200",
+            description =
+                "Call successful. Note that in case of unwrapping (`raw=true`), the response contains only the contents fof the `data` property.",
```
mpenick, line 158: fof -> of

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/schemas/namespaces/NamespacesResource.java
EricBorczuk: Oh if that's the case, then never mind! Yeah you're right, that restriction has existed since the beginning.

But then `\w` should suffice and you shouldn't need `\w+` ...right?

mpenick: underscore char is missing

mpenick (approved):

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/schemas/namespaces/NamespacesResource.java
ivansenic: :pray:

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/schemas/namespaces/model/dto/NamespaceDto.java
ivansenic: Hmm, I am not sure about this anymore.. Seems that replicas count in the `SimpleStrategy` is not really taking any effect, I set `3` when creating, but then get it back as `1`.. And the query is fine and everything.. We should open a separate ticket for this, it might be that is actually depending on the nodes in the cluster? Or?

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/schemas/namespaces/NamespacesResource.java
ivansenic: Done.

PR #1863, opened 2022-05-31
Allow resume mode to be settable on QueryParameters

### LINKED ISSUE #1862: Support different resume modes for the paging state for enriched queries
In the Bridge `QueryHandler` in case of the enriched queries we are always resuming with the `ResumeMode.NEXT_ROW`:

https://github.com/stargate/stargate/blob/18c16b3fc501f874f534db388d2651596cb16f40/bridge/src/main/java/io/stargate/bridge/service/QueryHandler.java#L156-L165

However, we need to be able to specify in the query parameters what resume mode we want. If the resume mode not specified, we should not get the page state at all.

I don't know if there will be additional `ResumeMode`s added in the future. I am fine in both designing the query parameter as the `boolean` or the `enum`.

EricBorczuk:
**What this PR does**:
Amends query.proto to allow adding a ResumeMode onto the query, which will be used when the Query is enriched.

Fixes #1862 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/bridge-proto/proto/query.proto
+++ b/bridge-proto/proto/query.proto
@@ -38,6 +38,11 @@ enum Consistency {
   LOCAL_ONE = 0x0A;
 }
 
+enum ResumeMode {
+  NEXT_ROW = 0x00;
+  NEXT_PARTITION = 0x01;
+}
+
 // A wrapper message for Consistency, for cases where the consistency level can be unset.
 message ConsistencyValue {
   Consistency value = 1;
@@ -226,6 +231,9 @@ message QueryParameters {
 
   // Whether to decorate the result set with comparable byte data and per-row paging states.
   bool enriched = 11;
+
+  // Configurable resume mode for page state creation; only used when enriched = true.
+  ResumeMode resumeMode = 12;
 }
 
 // A CQL column type.
--- a/bridge-proto/proto/query.proto
+++ b/bridge-proto/proto/query.proto
@@ -38,6 +38,8 @@ enum Consistency {
   LOCAL_ONE = 0x0A;
 }
 
+// The type of resume mode (either next row, or next partition) that should
+// be used when generating per-row paging states for enriched queries.
 enum ResumeMode {
   NEXT_ROW = 0x00;
   NEXT_PARTITION = 0x01;
@@ -233,6 +235,8 @@ message QueryParameters {
   bool enriched = 11;
 
   // Configurable resume mode for page state creation; only used when enriched = true.
+  // Changes the way that paging state is generated on each row of the result set, during enrichment.
+  // If unset, the paging state will not be populated for each row.
   ResumeMode resumeMode = 12;
 }
 
--- a/bridge/src/main/java/io/stargate/bridge/service/QueryHandler.java
+++ b/bridge/src/main/java/io/stargate/bridge/service/QueryHandler.java
@@ -154,14 +154,25 @@ public class QueryHandler extends MessageHandler<Query, Prepared> {
   }
 
   private ByteBuffer getPagingStateFromRow(
-      ByteBuffer resultSetPagingState, Row row, boolean lastInPage) {
+      ByteBuffer resultSetPagingState,
+      Row row,
+      QueryOuterClass.ResumeMode resumeMode,
+      boolean lastInPage) {
     if (lastInPage && resultSetPagingState == null) {
       return EXHAUSTED_PAGE_STATE;
     }
 
+    if (resumeMode == null) {
+      return null;
+    }
+
+    PagingPosition.ResumeMode internalResumeMode = PagingPosition.ResumeMode.NEXT_ROW;
+    if (resumeMode == QueryOuterClass.ResumeMode.NEXT_PARTITION) {
+      internalResumeMode = PagingPosition.ResumeMode.NEXT_PARTITION;
+    }
+
     return connection.makePagingState(
-        PagingPosition.ofCurrentRow(row).resumeFrom(PagingPosition.ResumeMode.NEXT_ROW).build(),
-        this.parameters);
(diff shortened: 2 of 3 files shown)


bridge/src/main/java/io/stargate/bridge/service/QueryHandler.java
```diff
@@ -154,14 +154,21 @@ private ByteBuffer getComparableBytesFromRow(
...
+      ByteBuffer resultSetPagingState,
+      Row row,
+      QueryOuterClass.ResumeMode resumeMode,
+      boolean lastInPage) {
     if (lastInPage && resultSetPagingState == null) {
       return EXHAUSTED_PAGE_STATE;
     }
 
+    PagingPosition.ResumeMode internalResumeMode = PagingPosition.ResumeMode.NEXT_ROW;
+    if (resumeMode != null && resumeMode.getValueDescriptor().getNumber() == 1) {
```
EricBorczuk, line 166: Is there a better way to match the enum's value to a PagingPosition.ResumeMode? Should I use the name of the enum instead?

mpenick: Use `QueryOuterClass.ResumeMode.NEXT_PARTITION`?

```diff
diff --git a/bridge/src/main/java/io/stargate/bridge/service/QueryHandler.java b/bridge/src/main/java/io/stargate/bridge/service/QueryHandler.java
index 21db6e2b6..47dadf815 100644
--- a/bridge/src/main/java/io/stargate/bridge/service/QueryHandler.java
+++ b/bridge/src/main/java/io/stargate/bridge/service/QueryHandler.java
@@ -26,6 +26,7 @@ import io.stargate.bridge.proto.QueryOuterClass.SchemaChange;
 import io.stargate.db.ClientInfo;
 import io.stargate.db.ImmutableParameters;
 import io.stargate.db.PagingPosition;
+import io.stargate.db.PagingPosition.ResumeMode;
 import io.stargate.db.Parameters;
 import io.stargate.db.Persistence;
 import io.stargate.db.Persistence.Connection;
@@ -40,6 +41,7 @@ import java.nio.ByteBuffer;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
@@ -163,7 +165,7 @@ public class QueryHandler extends MessageHandler<Query, Prepared> {
     }
 
     PagingPosition.ResumeMode internalResumeMode = PagingPosition.ResumeMode.NEXT_ROW;
-    if (resumeMode != null && resumeMode.getValueDescriptor().getNumber() == 1) {
+    if (Objects.equals(resumeMode, QueryOuterClass.ResumeMode.NEXT_PARTITION)) {
       internalResumeMode = PagingPosition.ResumeMode.NEXT_PARTITION;
     }
```

bridge-proto/proto/query.proto
```diff
@@ -38,6 +38,11 @@ enum Consistency {
   LOCAL_ONE = 0x0A;
 }
 
+enum ResumeMode {
+  NEXT_ROW = 0x00;
```
mpenick, line 42: Is the paging state data different for `NEXT_ROW` vs. `NEXT_PARTITION` or is it the same state data?

bridge/src/main/java/io/stargate/bridge/service/ValuesHelper.java
```diff
@@ -122,12 +123,12 @@ public static BoundStatement bindValues(Prepared prepared, Values values, ByteBu
 
   public static ResultSet processResult(Rows rows, QueryParameters parameters)
       throws StatusException {
-    return processResult(rows, parameters.getSkipMetadata(), null, null, null, null);
+    return processResult(rows, parameters.getSkipMetadata(), null, null, null, null, null);
```
mpenick, line 126: Maybe use `QueryOuterClass.ResumeMode.UNRECOGNIZED` instead of `null`? `QueryParameters#getResumeMode()` will no return `null` but will instead return `UNRECOGNIZED`.

bridge-proto/proto/query.proto
```diff
@@ -226,6 +231,9 @@ message QueryParameters {
 
   // Whether to decorate the result set with comparable byte data and per-row paging states.
   bool enriched = 11;
+
+  // Configurable resume mode for page state creation; only used when enriched = true.
```
ivansenic, line 235: Does not really explains everything, add:

* should reference that it's for the paging state creation for each row
* if not specified, that the paging state is not set then
* difference between resume modes, maybe at the enum declaration

ivansenic (changes requested):
Looks good, but can not approve this without a single test.. Tests should be there to confirm all cases:

* next row -> points to next row
* next partition -> points to the first row of the next partition
* not specified -> no paging state

bridge/src/main/java/io/stargate/bridge/service/QueryHandler.java
```diff
@@ -154,14 +154,25 @@ private ByteBuffer getComparableBytesFromRow(
...
+      ByteBuffer resultSetPagingState,
+      Row row,
+      QueryOuterClass.ResumeMode resumeMode,
+      boolean lastInPage) {
     if (lastInPage && resultSetPagingState == null) {
       return EXHAUSTED_PAGE_STATE;
     }
 
+    if (resumeMode == null) {
+      return null;
```
ivansenic, line 166: also questioning if this `null`, will throw `NullPointerException` once you try to set it to the row builder? Is this taken care of?

EricBorczuk: Yea this is handled in ValuesHelper (line 213)

mpenick (approved):

EricBorczuk:
Discussed with Ivan re: tests, we don't yet have testing for QueryHandler, so punting it for now.

PR #1868, opened 2022-06-02
Add System property "stargate.bridge.client.disableSchemaFreshnessCheck" for perf testing

tatu-at-datastax:
**What this PR does**:

Adds a new System property that can be set to `"true"` to disable freshness check by Bridge client; this allows performance testing of the scenario where after initial lookup cached Keyspace (and thereby Table) instances can be used locally (up until expiration) without gRPC call by service to Coordinator. This will help determine exact overhead of the freshness check.

Part of #1828

**Checklist**
- [x] Changes manually tested


tatu-at-datastax:
Not 100% sure if this makes sense to merge to `v2.0.0` or remain in a branch. Not universally needed but also not a lot of code to maintain.

ivansenic:
> Not 100% sure if this makes sense to merge to `v2.0.0` or remain in a branch. Not universally needed but also not a lot of code to maintain.

Not sure either..

olim7t:
I don't see any valid reason to ever enable this in production, so I think it should stay on a branch.

tatu-at-datastax:
Right, I would not propose to run it for production (obviously), but whether it'd be useful for testing purposes, using production builds/images, instead of having to maintain separate branch and require specific builds.
But for now it probably makes sense to stay in a branch unless specific need arises.

tatu-at-datastax:
No longer needed.

PR #1871, opened 2022-06-06
closes #1731: query executor for the Docs API V2

### LINKED ISSUE #1731: Re-implement Documents `QueryExecutor` to work against gRPC
Re-implement the `QueryExecutor` class and substitute dependency to the `persistence-api` model to the gRPC bridge model. The new QueryExecutor will depend on the extended response implemented in the https://github.com/stargate/stargate/issues/1715.

ivansenic:
**What this PR does**:

The implementation for the `QueryExecutor` for the V2. This is currently based on #1870, but can be reviewed anyway.

Changes from V1:
* resume mode is being explicitly requested at query time, instead of resolved at the time of making the page state; however, decision between next row or next partition stays the same
* callers can define the new parameter `fetchRowPaging` which will define if correct resume mode will be passed to the bridge or not, meaning that queries that will not be used for determining the user-facing page state can avoid any overhead, which would be V1 behavior actually
* figuring out the selected clustering columns can not be done in V2, thus we include all of them.. they are used in the comparison, which will actually look at all possible path columns (thus one test less than V1)
* Mutiny does not support ordered merge, so for now, we need to still interact with the RxJava3

Other improvements:
* simplified as much as possible the internal classes of the executor in V1 (removed `Page` and `QueryData`)
* row wrapping is optimized, to avoid constantly re-creating the index map
* configuration is adapted for the max search page

Fixes #1731

**Checklist**
- [x] #1870 merged


sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/executor/QueryExecutor.java
```diff
@@ -0,0 +1,619 @@
...
+    return result;
+  }
+
+  // all the doc path columns
+  // note that this differs from V1,
+  // as we can not know what clustering columns are selected
+  private List<String> docPathColumns() {
+    // TODO we push all columns as we can not know what's selected
+    //  make cached or smth
+    return documentProperties.tableColumns().pathColumnNames().stream().sorted().toList();
```
ivansenic, line 458: Fix!

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/bridge/ValidatingPaginator.java
```diff
@@ -0,0 +1,97 @@
...
+package io.stargate.sgv2.docsapi.bridge;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.nio.ByteBuffer;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+public class ValidatingPaginator {
```
ivansenic, line 25: add correct author tag everywhere

ivansenic (commented):
Self-review

EricBorczuk (approved):
This looks good to me! I asked one question about the conversion to Rx from Uni

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/executor/DocumentProperty.java
```diff
@@ -0,0 +1,125 @@
...
+   * <p>This index corresponds to the position of the query's paging state in the combined paging
+   * state.
+   */
+  abstract int queryIndex();
+
+  /** @return Comparable key that can be used for comparing the partition belonging for this row. */
+  @Value.Lazy
+  ByteString comparableKey() {
+    QueryOuterClass.Row row = rowWrapper().row();
+    return row.getComparableBytes().getValue();
```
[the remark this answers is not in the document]
ivansenic: It's not yet in and yes you are correct it's not working for the C* 3 and 4..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/executor/QueryExecutor.java
```diff
@@ -0,0 +1,619 @@
...
+  /**
+   * Runs the provided queries in parallel and merges their result sets according to the selected
+   * primary key columns, then groups the merged rows into {@link RawDocument} objects according to
+   * {@code keyDepth} primary key values.
+   *
+   * <p>Note: all queries must select the same set of columns from the same table.
+   *
+   * <p>Note: queries may use different {@code WHERE} conditions.
+   *
+   * <p>Note: the queries must select all the partition key columns and zero or more clustering key
```
olim7t, line 139: Isn't there only one partition key column?

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/executor/QueryExecutor.java
```diff
@@ -0,0 +1,619 @@
...
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
+/** Executes pre-built document queries, groups document rows and manages document pagination. */
+@ApplicationScoped
+public class QueryExecutor {
```
olim7t, line 54: It would really help to have a high-level explanation of this class: why searches can result into multiple CQL queries, the problem of combining paging states, key depth, exponential page sizes, etc.
Maybe with a few concrete examples (like I did in the write query builder classes), if we can find ones that are sufficiently compact yet meaningful.

From an outsider's point of view, the code is very hard to follow.

sgv2-docsapi/pom.xml
```diff
@@ -108,6 +108,16 @@
...
+    <dependency>
+      <groupId>io.smallrye.reactive</groupId>
+      <artifactId>mutiny-rxjava3</artifactId>
+      <version>1.5.0</version>
+    </dependency>
+    <dependency>
+      <groupId>com.github.akarnokd</groupId>
+      <artifactId>rxjava3-extensions</artifactId>
+      <version>3.1.1</version>
+    </dependency>
```
olim7t, line 120: Maybe allude to the fact that this is needed because of ordered merge in `QueryExecutor`?

Although one can also comment the dependencies and try to recompile to see where it's used...

olim7t (commented):
A few minor remarks, I'm afraid I don't understand the code enough to be more helpful.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/executor/DocumentProperty.java
```diff
@@ -0,0 +1,125 @@
...
+          ByteString.unsignedLexicographicalComparator()
+              .compare(docProperty1.comparableKey(), docProperty2.comparableKey());
+
+  /**
+   * A result set that points to a page where the row that describes this document property
+   * originates from.
+   */
+  abstract QueryOuterClass.ResultSet page();
+
+  /** A single row wrapper in {@link RowWrapper}. */
```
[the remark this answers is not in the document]
ivansenic: Yea, typo.. Fixing..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/executor/QueryExecutor.java
```diff
@@ -0,0 +1,619 @@
...
+
+    return new Accumulator(id, comparator, docKey.build(), property);
+  }
+
+  /**
+   * Builds a comparator that follows the natural order of rows in document tables, including all
+   * clustering columns (i.e., "path" columns).
+   */
+  private Comparator<DocumentProperty> rowComparator() {
+    List<Comparator<DocumentProperty>> comparators = new ArrayList<>();
```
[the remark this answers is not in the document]
ivansenic: This is outdated now and refactored into a specific component.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/executor/QueryExecutor.java
ivansenic: This was a leftover comment from V1, where query executor was created with a more generic usage in mind, will adapt.

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/executor/QueryExecutor.java
ivansenic: True. Will try to add some explanations..

sgv2-docsapi/pom.xml
ivansenic: Will add a comment.

PR #1872, opened 2022-06-06
Initial implementation of Host-to-tenant metrics tagger

### LINKED ISSUE #1565: Add support for registering custom `HttpMetricsTagProvider` in Stargate V2/REST
(note: follow-up to #1551)

Due to StargateV2 refactoring the mechanism used for adding HTTP metrics tag providers will not work (it is based on OSGi).
We need to re-connect this, but it seems that this should be part of work when moving to the new HTTP/Service framework (see discussion #1526).

This issue needs to be tackled before actual production deployments as tags are important for analytics, dashboards.

tatu-at-datastax:
**What this PR does**:

Adds equivalent Host/tenant Metrics tag for SGv2/REST API as what SGv1/REST does (although in slightly simpler way)

Fixes #1565

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/metrics-jersey/src/main/java/io/stargate/metrics/jersey/sgv2/TenantIdFromHostHeaderTagsProvider.java
+++ b/metrics-jersey/src/main/java/io/stargate/metrics/jersey/sgv2/TenantIdFromHostHeaderTagsProvider.java
@@ -1,0 +1,77 @@
+package io.stargate.metrics.jersey.sgv2;
+
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.jersey2.server.JerseyTagsProvider;
+import java.util.Collections;
+import javax.ws.rs.core.MultivaluedMap;
+import org.glassfish.jersey.server.ContainerRequest;
+import org.glassfish.jersey.server.monitoring.RequestEvent;
+
+/** Specialized {@link JerseyTagsProvider} used to extract Tenant Id from "Host" HTTP header. */
+public class TenantIdFromHostHeaderTagsProvider implements JerseyTagsProvider {
+  private static final String DEFAULT_TENANT_TAG_KEY = "tenant";
+
+  private final String tenantTagName;
+
+  /** Tags when tenant is unknown. */
+  private final Tags tagsForUnknown;
+
+  public TenantIdFromHostHeaderTagsProvider() {
+    this(DEFAULT_TENANT_TAG_KEY);
+  }
+
+  /**
+   * @param tenantTagName Tag name to use for Tenant Id, if any; {@code null} or empty String to
+   *     disable extraction
+   */
+  public TenantIdFromHostHeaderTagsProvider(String tenantTagName) {
+    if (tenantTagName == null || tenantTagName.isEmpty()) {
+      this.tenantTagName = null;
+      tagsForUnknown = null;
+    } else {
+      this.tenantTagName = tenantTagName;
+      tagsForUnknown = Tags.of(tenantTagName, "unknown");
+    }
+  }
+
+  @Override
+  public Iterable<Tag> httpRequestTags(RequestEvent event) {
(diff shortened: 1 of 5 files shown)


metrics-jersey/src/main/java/io/stargate/metrics/jersey/sgv2/TenantIdFromHostHeaderTagsProvider.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  protected Iterable<Tag> findHostTag(RequestEvent event) {
+    if (tenantTagName == null) {
+      return Collections.emptyList();
+    }
+    ContainerRequest request = event.getContainerRequest();
+    MultivaluedMap<String, String> headers = request.getHeaders();
+
+    // HTTP header names are case-insensitive but MultivaluedMap is case-sensitive
```
ivansenic, line 55: Feels better for me to have extracting logic in one place and set it in the request context.. But leaving to you to decide..

metrics-jersey/src/main/java/io/stargate/metrics/jersey/sgv2/TenantIdFromHostHeaderTagsProvider.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  /**
+   * @param tenantTagName Tag name to use for Tenant Id, if any; {@code null} or empty String to
+   *     disable extraction
+   */
+  public TenantIdFromHostHeaderTagsProvider(String tenantTagName) {
+    if (tenantTagName == null || tenantTagName.isEmpty()) {
+      this.tenantTagName = null;
+      tagsForUnknown = null;
```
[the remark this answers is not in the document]
tatu-at-datastax: Ok. I was wondering whether to allow it or not but since there is no current use case I'll just check that null is not passed instead.

metrics-jersey/src/main/java/io/stargate/metrics/jersey/sgv2/TenantIdFromHostHeaderTagsProvider.java
tatu-at-datastax: Ok, I tried to see how to reuse logic of/from `CreateStargateBridgeClientFilter` which has it but unfortunately it seems bit tricky because:

1. Extracting bit of code to get tenant id from `MultivaluedMap` would be nice but `metrics-jersey` does not depend on `sgv-service-common` nor vice versa. Worse, I don't think there's anything both depend on: `core` should not be depended on by SGv2 as it pulls in jersey-specific stuff.
2. Relying on Bridge filter would be possible by bit of refactoring (right now it only adds client in context but I could just stuff tenant id there) but would only be applied to endpoints that use annotation to get bridge client, leaving out health-check etc end points (which may or may not be a problem)
 
Now: I think that when we convert to Quarkus this is worth revisiting. But on short term I think this bit of duplication might make most sense. Not ideal but I'd prefer avoiding adding cross-dependencies as `metrics-jersey` package should go away from SGv2 once we upgrade framework.

tatu-at-datastax:
Ok, ready for review. Wasn't able to figure out how to reuse logic from `CreateStargateBridgeClientFilter` in a clean way so there is a bit of duplication, but I think that is ok as this will (need to) be resolved with move to Quarkus when filters are changed anyway. Still open to alternative ideas of course; ideally no duplication of that biz logic.

tatu-at-datastax:
@ivansenic WDYT?

mpenick (approved):
This seem reasonable to me.

tatu-at-datastax:
This is needed for 2.0.0-alpha-12 so merging.

ivansenic:
Looks good, sorry for not being available.

tatu-at-datastax:
@ivansenic no problem at all, thank you for your earlier input.

Issue #1738, opened 2022-03-28
Implement `ReadBridgeService` in the Document API V2


ivansenic:
The new `ReadBridgeService` should be based off the existing `DocumentSearchService`. The query builder implementations should be building a gRPC-based queries. The `DataStore` should be substituted with the bridge gRPC client, but the client can become the inject-able dependency. The reactive framework should be adapted to the Quarkus default.

PR #1874, opened 2022-06-07
 relates to #1738: queries for the document reads adapted for V2

ivansenic:
**What this PR does**:
First part of the #1738, where I only converted all the search query builders to the V2. Note that there was a slight change in the design from the V1, as V2 is not supporting the mixed mode markers & literals in the conditions.. Thus, everything is refactored to use a marker. 

`TTL` function was not in the `QueryBuilderImpl` in V2, I added the code, but was confused that there was no DSL chnage in V1 to reference this in `"select star? column* function* ((count|min|max|avg|sum|writeTimeColumn) as?)* "`. We should resolve this.

Relates to #1738<issue number>

**Checklist**
- [x] TTL function

### DIFF
--- a/sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/QueryBuilderImpl.java
+++ b/sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/QueryBuilderImpl.java
@@ -1523,6 +1523,14 @@ public class QueryBuilderImpl {
       return function(columnName, alias, "SUM");
     }
 
+    public static FunctionCall ttl(String columnName) {
+      return ttl(columnName, null);
+    }
+
+    public static FunctionCall ttl(String columnName, String alias) {
+      return function(columnName, alias, "TTL");
+    }
+
     public static FunctionCall writeTime(String columnName) {
       return writeTime(columnName, null);
     }
--- a/sgv2-docsapi/pom.xml
+++ b/sgv2-docsapi/pom.xml
@@ -109,6 +109,10 @@
       <artifactId>quarkus-opentelemetry-exporter-otlp</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+    </dependency>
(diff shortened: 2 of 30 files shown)


sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/QueryBuilderImpl.java
```diff
@@ -1523,6 +1523,14 @@ public static FunctionCall sum(String columnName, String alias) {
...
     }
 
+    public static FunctionCall ttl(String columnName) {
+      return ttl(columnName, null);
+    }
+
+    public static FunctionCall ttl(String columnName, String alias) {
+      return function(columnName, alias, "TTL");
+    }
+
```
ivansenic, line 1533: @olim7t Can you help here and let me know if I need to change the DSL as well.. Seems that `ttl` was not included in the V1, most likely was missed.  See here:

```
      @SubExpr(
          name = "select",
          definedAs =
              "select star? column* function* ((count|min|max|avg|sum|writeTimeColumn) as?)* "
                  + "from (where* perPartitionLimit? limit? groupBy* orderBy*) allowFiltering?"),
```

ivansenic: Other functions are there `(count|min|max|avg|sum|writeTimeColumn)`?

EricBorczuk (approved):

sgv2-service-common/src/main/java/io/stargate/sgv2/common/cql/builder/QueryBuilderImpl.java
ivansenic: Should I explore more in order to figure it out?

ivansenic: Tried to play around with it, but basically needs more that adding a simple action there.. It's not important, but could be a nice addition in the future..

PR #1876, opened 2022-06-08
Add optimistic handling of schema metadata (fixes #1873)

### LINKED ISSUE #1873: Consider optimistic handling of schema metadata
Stargate services often use the following pattern:
- get the schema metadata, for example with `StargateBridgeClient.getKeyspace()`
- build a DML query based on this schema
- execute that query

The current implementation of `getKeyspace()` (v2.0.0-ALPHA-11) uses a cache, but it sends a background gRPC query to check that the metadata is still up to date. We use `DescribeKeyspaceQuery.hash` to optimize the happy path when nothing has changed (by sending an empty response), but that's still an extra network hop.

We could save that call with an optimistic approach:
- the client builds its query based on its current version of the metadata
- it sends the query, keyspace name and keyspace hash to the bridge
- if the hash is up to date, the bridge returns the response
- otherwise:
  - the bridge sends a special response that contains the new keyspace metadata
  - the client rebuilds the query based on the new metadata, and retries
  - the client updates its cache

This retry process could be abstracted behind a new `StargateBridgeClient` method:
```java
Response executeQuery(String keyspaceName, Function<CqlKeyspaceDescribe, Query> query)
```

As a first step, we can implement a dummy version of that method (based on the current ones, no actual retries) and evaluate the impacts on the service code:
- do all cases fit the trivial scenario above?
- how does it affect readability?
- do we also need this for batches? (probably for docs)

Then we can add the actual implementation, which will require a new gRPC call.

olim7t:
**What this PR does**:
Add a way to execute queries optimistically, to avoid pre-fetching the schema (even if it's just a hash check) before each query.

Fixes #1873

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/bridge-proto/proto/schema.proto
+++ b/bridge-proto/proto/schema.proto
@@ -162,11 +162,21 @@ message SupportedFeaturesResponse {
 
 message QueryWithSchema {
   Query query = 1;
+  // The keyspace that the query operates on.
   string keyspace_name = 2;
+  // The version hash of the keyspace that the client used to build the query.
   sint32 keyspace_hash = 3;
 }
 
 message QueryWithSchemaResponse {
-  Response response = 1;
-  CqlKeyspaceDescribe new_keyspace = 2;
+  message NoKeyspace {}
+
+  oneof inner {
+    // The keyspace hadn't changed on the bridge side, the query succeeded
+    Response response = 1;
+    // The keyspace changed on the bridge side, this is the new metadata.
+    CqlKeyspaceDescribe new_keyspace = 2;
+    // The keyspace was deleted on the bridge side.
+    NoKeyspace no_keyspace = 3;
+  }
 }
--- a/bridge-proto/proto/schema.proto
+++ b/bridge-proto/proto/schema.proto
@@ -158,3 +158,15 @@ message SupportedFeaturesResponse {
   // Whether the persistence backend supports logged CQL batches.
   bool logged_batches = 3;
 }
+
+
+message QueryWithSchema {
+  Query query = 1;
+  string keyspace_name = 2;
+  sint32 keyspace_hash = 3;
+}
+
+message QueryWithSchemaResponse {
+  Response response = 1;
+  CqlKeyspaceDescribe new_keyspace = 2;
+}
--- a/sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/bridge/ValidatingStargateBridge.java
+++ b/sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/bridge/ValidatingStargateBridge.java
@@ -50,6 +50,12 @@ public class ValidatingStargateBridge implements StargateBridge {
   }
 
   @Override
+  public Uni<Schema.QueryWithSchemaResponse> executeQueryWithSchema(
+      Schema.QueryWithSchema request) {
+    throw new UnsupportedOperationException("Not implemented yet");
+  }
+
+  @Override
   public Uni<QueryOuterClass.Response> executeBatch(QueryOuterClass.Batch batch) {
     return batch.getQueriesList().stream()
         .map(
--- a/sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/bridge/ValidatingStargateBridge.java
+++ b/sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/bridge/ValidatingStargateBridge.java
@@ -52,7 +52,9 @@ public class ValidatingStargateBridge implements StargateBridge {
   @Override
   public Uni<Schema.QueryWithSchemaResponse> executeQueryWithSchema(
       Schema.QueryWithSchema request) {
-    throw new UnsupportedOperationException("Not implemented yet");
+    // Simulate a successful execution (keyspace was up to date)
+    return executeQuery(request.getQuery())
+        .map(response -> Schema.QueryWithSchemaResponse.newBuilder().setResponse(response).build());
   }
 
   @Override
--- a/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceImpl.java
+++ b/sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceImpl.java
@@ -72,25 +72,23 @@ public class Sgv2RowsResourceImpl extends ResourceBase implements Sgv2RowsResour
       throw new WebApplicationException(e.getMessage(), Status.BAD_REQUEST);
     }
 
-    return callWithTable(
-        bridge,
-        keyspaceName,
-        tableName,
-        false,
-        (tableDef) -> {
-          final ToProtoConverter toProtoConverter = findProtoConverter(tableDef);
+    QueryOuterClass.Response response =
+        queryWithTable(
+            bridge,
+            keyspaceName,
+            tableName,
+            tableDef -> {
+              final ToProtoConverter toProtoConverter = findProtoConverter(tableDef);
 
-          final List<BuiltCondition> whereConditions;
-          try {
-            whereConditions = new WhereParser(tableDef, toProtoConverter).parseWhere(where);
-          } catch (IllegalArgumentException e) {
-            throw new WebApplicationException(e.getMessage(), Status.BAD_REQUEST);
-          }
+              final List<BuiltCondition> whereConditions;
+              try {
+                whereConditions = new WhereParser(tableDef, toProtoConverter).parseWhere(where);
+              } catch (IllegalArgumentException e) {
+                throw new WebApplicationException(e.getMessage(), Status.BAD_REQUEST);
+              }
 
-          final QueryOuterClass.Query query;
-          if (columns.isEmpty()) {
-            query =
-                new QueryBuilder()
+              if (columns.isEmpty()) {
+                return new QueryBuilder()
                     .select()
                     .star()
                     .from(keyspaceName, tableName)
@@ -98,9 +96,8 @@ public class Sgv2RowsResourceImpl extends ResourceBase implements Sgv2RowsResour
                     .orderBy(sortOrder)
                     .parameters(parametersForPageSizeAndState(pageSizeParam, pageStateParam))
                     .build();
-          } else {
-            query =
-                new QueryBuilder()
+              } else {
+                return new QueryBuilder()
                     .select()
                     .column(columns)
                     .from(keyspaceName, tableName)
@@ -108,9 +105,9 @@ public class Sgv2RowsResourceImpl extends ResourceBase implements Sgv2RowsResour
                     .orderBy(sortOrder)
                     .parameters(parametersForPageSizeAndState(pageSizeParam, pageStateParam))
                     .build();
-          }
-          return fetchRows(bridge, query, raw);
-        });
+              }
+            });
+    return toHttpResponse(response, raw);
   }
 
   @Override
@@ -135,15 +132,14 @@ public class Sgv2RowsResourceImpl extends ResourceBase implements Sgv2RowsResour
       throw new WebApplicationException(e.getMessage(), Status.BAD_REQUEST);
     }
 
-    return callWithTable(
-        bridge,
-        keyspaceName,
-        tableName,
-        false,
-        (tableDef) -> {
-          final ToProtoConverter toProtoConverter = findProtoConverter(tableDef);
-          final QueryOuterClass.Query query =
-              buildGetRowsByPKQuery(
+    QueryOuterClass.Response response =
+        queryWithTable(
+            bridge,
+            keyspaceName,
+            tableName,
+            tableDef -> {
+              final ToProtoConverter toProtoConverter = findProtoConverter(tableDef);
+              return buildGetRowsByPKQuery(
                   keyspaceName,
                   tableName,
                   path,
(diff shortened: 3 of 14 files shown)


sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/StargateBridgeClient.java
```diff
@@ -47,6 +48,25 @@ default Response executeQuery(Query query) {
     return Futures.getUninterruptibly(executeQueryAsync(query));
   }
 
+  // TODO create async variant?
+  Response executeQuery(
+      String keyspaceName, Function<Optional<CqlKeyspaceDescribe>, Query> queryProducer);
```
olim7t, line 53: I started with a synchronous implementation, because it is more straightforward, and what the REST service uses.

If we move forward with this, I'll convert it to async like the other methods, and probably create batch variants too.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/schemas/Sgv2IndexesResourceImpl.java
```diff
@@ -135,20 +136,17 @@ public Response dropIndex(
...
-            .orElseThrow(() -> new WebApplicationException("Table not found", Status.NOT_FOUND));
+    // We need an explicit schema fetch here: queryWithTable() wouldn't work, because the initial
+    // "optimistic" query construction might operate on a table metadata that doesn't reflect a
+    // recent index addition yet, and therefore throw a 404.
+    Schema.CqlTable table = getTable(bridge, keyspaceName, tableName);
     if (!ifExists
         && table.getIndexesList().stream().noneMatch(i -> indexName.equals(i.getName()))) {
       throw new WebApplicationException(
           String.format("Index '%s' not found.", indexName), Status.NOT_FOUND);
     }
```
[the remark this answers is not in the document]
olim7t: > fetch and retry systematically, whenever the first query construction throws

Implemented in "Improve error handling" commit.

bridge-proto/proto/schema.proto
```diff
@@ -157,4 +157,16 @@ message SupportedFeaturesResponse {
...
+
+message QueryWithSchema {
+  Query query = 1;
+  string keyspace_name = 2;
+  sint32 keyspace_hash = 3;
+}
+
+message QueryWithSchemaResponse {
+  Response response = 1;
+  CqlKeyspaceDescribe new_keyspace = 2;
```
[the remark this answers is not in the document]
mpenick: Any advantage to using `oneof` here to codify those semantics? If they are mutually exclusive.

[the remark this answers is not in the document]
olim7t: Maybe I could use `oneof` with an empty `DeletedKeyspace` message as the third alternative.

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/bridge/ValidatingStargateBridge.java
```diff
@@ -49,6 +49,12 @@ public Uni<QueryOuterClass.Response> executeQuery(QueryOuterClass.Query query) {
     return expectation.execute(null, query.hasParameters() && query.getParameters().getEnriched());
   }
 
+  @Override
+  public Uni<Schema.QueryWithSchemaResponse> executeQueryWithSchema(
+      Schema.QueryWithSchema request) {
+    throw new UnsupportedOperationException("Not implemented yet");
```
ivansenic, line 55: why just not return same as `executeQuery`.. I don't think we will be using schema failure in validating tests..

ivansenic (commented):
Seems like a very complex peace of code, requires a lot of testing imo.. But also like a very nice extension to the bridge.. 

I am pretty sure docs API would not need this, at least not in the current form, as we anyway don't depend on a keyspace to create queries, our queries are always fixed.. The only use-case I could think of would be keyspace or table deleted, but they are still in our cache.. Schema changes to document tables should not occur... There are some other things about this impl, that most likely would not work in docs.. For example, the JSON schema validation that is set as a comment in a table.

Implementation in Quarks reactive mode would be also pretty simple for this..

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
```diff
@@ -107,6 +110,64 @@ public CompletionStage<Response> executeQueryAsync(Query query) {
     return observer.asFuture();
   }
 
+  @Override
+  public Response executeQuery(
+      String keyspaceName, Function<Optional<CqlKeyspaceDescribe>, Query> queryProducer) {
```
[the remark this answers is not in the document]
olim7t: When the keyspace does not exist.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
```diff
@@ -107,6 +110,64 @@ public CompletionStage<Response> executeQueryAsync(Query query) {
...
+    } else {
+      // Build the query with our local version
+      Query query;
+      try {
+        query = queryProducer.apply(Optional.of(keyspace));
+      } catch (Throwable t) {
+        // Always retry with the latest version, in case the error is caused by stale metadata
+        // (e.g. using a table that has not yet replicated to our local cache)
+        return forceFetchAndExecute(keyspaceName, decoratedKeyspaceName, queryProducer);
+      }
```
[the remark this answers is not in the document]
olim7t: I tried to explain it in the comment. See a concrete example [here](https://github.com/stargate/stargate/pull/1876#discussion_r892660032:~:text=olim7t-,2%20days%20ago,-This%20illustrates%20what).

testing/src/main/java/io/stargate/it/http/RestApiv2SchemaTest.java
```diff
@@ -1114,9 +1114,9 @@ public void indexCreateInvalid() throws IOException {
                 "%s/v2/schemas/keyspaces/%s/tables/invalid_table/indexes",
                 restUrlBase, keyspaceName),
             objectMapper.writeValueAsString(indexAdd),
-            HttpStatus.SC_NOT_FOUND);
+            HttpStatus.SC_BAD_REQUEST);
     ApiError response = objectMapper.readValue(body, ApiError.class);
-    assertThat(response.getCode()).isEqualTo(HttpStatus.SC_NOT_FOUND);
+    assertThat(response.getCode()).isEqualTo(HttpStatus.SC_BAD_REQUEST);
```
olim7t, line 1119: v1 has inconsistent rules here, missing table on an index creation is a 404, but for column operations it's a 400. Either we make the behavior consistent, or we'll need a way to tell `queryWithTable` which error to throw.

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceImpl.java
```diff
@@ -72,45 +72,42 @@ public Response getRowWithWhere(
...
+        queryWithTable(
+            bridge,
+            keyspaceName,
+            tableName,
+            tableDef -> {
+              final ToProtoConverter toProtoConverter = findProtoConverter(tableDef);
+
+              final List<BuiltCondition> whereConditions;
+              try {
+                whereConditions = new WhereParser(tableDef, toProtoConverter).parseWhere(where);
```
tatu-at-datastax, line 85: This is probably where we would need to make `WhereParser` indicate things like missing table/column
using specific exception, to allow schema fetch. Although maybe it's not a big issue for this particular query:
the case of adding a column and then immediately referencing column might not be common?
Probably more common with row insert/create/update case.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
```diff
@@ -107,6 +110,91 @@ public CompletionStage<Response> executeQueryAsync(Query query) {
...
+    CqlKeyspaceDescribe keyspace = keyspaceCache.getIfPresent(decoratedKeyspaceName);
+    if (keyspace == null) {
+      // Nothing in our local cache, check if the bridge has a more recent version
+      return forceFetchAndExecute(keyspaceName, decoratedKeyspaceName, tableName, queryProducer);
+    } else {
+      // Build the query with our local version
+      Query query;
+      try {
+        query = produceQuery(queryProducer, Optional.of(keyspace), tableName);
+      } catch (Throwable t) {
```
tatu-at-datastax, line 128: Is there a reason this is not `Exception` but `Throwable`? (I feel I may have asked same question before,
and it had to do with custom subtypes or something?)

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/bridge/ValidatingStargateBridge.java
olim7t: Fixed in latest version.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
```diff
@@ -107,6 +110,91 @@ public CompletionStage<Response> executeQueryAsync(Query query) {
...
+                QueryWithSchemaResponse.InnerCase innerCase = withSchemaResponse.getInnerCase();
+                switch (innerCase) {
+                  case RESPONSE:
+                    // Success, the keyspace hadn't changed on the bridge
+                    return CompletableFuture.completedFuture(withSchemaResponse.getResponse());
+                  case NEW_KEYSPACE:
+                    // The keyspace was updated, retry with that version
+                    CqlKeyspaceDescribe ks = withSchemaResponse.getNewKeyspace();
+                    keyspaceCache.put(decoratedKeyspaceName, ks);
+                    return executeQueryAsync(
```
tatu-at-datastax, line 158: Is there any danger this could loop indefinitely? That is, what mechanism makes the retry called only once?

testing/src/main/java/io/stargate/it/http/RestApiv2SchemaTest.java
tatu-at-datastax: Hmmmh. It seems like `404` would be better, but I guess 400 is ok too.

Maybe file a separate issue for resolving this discrepancy but use whatever logic is easiest here, for now, to get the optimized schema handling working even if at expense of error code inconsistence (whether between SGv1/SGv2 REST, or across REST operations).

sgv2-restapi/src/main/java/io/stargate/sgv2/restsvc/resources/Sgv2RowsResourceImpl.java
tatu-at-datastax: It sounds like retry on exception from query building already results in retry so that should be good. I'll go over PR again to see it, assuming it's all fine.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
olim7t: Yeah, it's unlikely that retrying on something like `Error` would be useful. Changing it.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
olim7t: This is not a recursive call, it invokes the original `executeQueryAsync()` method. If `produceQuery()` throws this time, the exception will propagate.

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
```diff
@@ -107,6 +110,91 @@ public CompletionStage<Response> executeQueryAsync(Query query) {
...
+      return forceFetchAndExecute(keyspaceName, decoratedKeyspaceName, tableName, queryProducer);
+    } else {
+      // Build the query with our local version
+      Query query;
+      try {
+        query = produceQuery(queryProducer, Optional.of(keyspace), tableName);
+      } catch (Throwable t) {
+        // Always retry with the latest version, in case the error is caused by stale metadata
+        // (e.g. using a table that has not yet replicated to our local cache)
+        return forceFetchAndExecute(keyspaceName, decoratedKeyspaceName, tableName, queryProducer);
```
tatu-at-datastax, line 131: Ah. This is where retry is for case of query building failing, possibly due to stale schema. And calls different method

sgv2-service-common/src/main/java/io/stargate/sgv2/common/grpc/DefaultStargateBridgeClient.java
tatu-at-datastax: Ahhhhh. I misread the invocation so it looked it was recursive call. But it isn't, there's just 1 arg.
Yes, that works.

tatu-at-datastax (approved):

PR #1877, opened 2022-06-08
Fix #1463 by adding keyspace filter as suggested

### LINKED ISSUE #1463: GET /indexes does not filter by keyspace
In `IndexesResource`:
```java
           BoundQuery query =
                restDB
                    .queryBuilder()
                    .select()
                    .column(columns)
                    .from(SYSTEM_SCHEMA, INDEXES_TABLE)
                    .where("table_name", Predicate.EQ, tableName)
                    .allowFiltering(true)
```
is wrong. If two tables with the same name exist in different keyspaces, all of their indexes will be returned.
It should be instead:
```java
           BoundQuery query =
                restDB
                    .queryBuilder()
                    .select()
                    .column(columns)
                    .from(SYSTEM_SCHEMA, INDEXES_TABLE)
                    .where("keyspace_name", Predicate.EQ, keyspaceName)
                    .where("table_name", Predicate.EQ, tableName)
```

tatu-at-datastax:
**What this PR does**:

Adds missing filter for keyspace in "get all indexes"; extends existing test case to cover it (fails without fix, passes with it)

Fixes #1463

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/testing/src/main/java/io/stargate/it/http/RestApiv2SchemaTest.java
+++ b/testing/src/main/java/io/stargate/it/http/RestApiv2SchemaTest.java
@@ -1146,19 +1146,38 @@ public class RestApiv2SchemaTest extends BaseIntegrationTest {
   @Test
   public void indexListAll() throws IOException {
     createTestKeyspace(keyspaceName);
+    final String altKeyspaceName = "alt_" + System.currentTimeMillis();
+    createTestKeyspace(altKeyspaceName);
     tableName = "tbl_createtable_" + System.currentTimeMillis();
-    createTestTable(
+    createTestTableIn(
+        keyspaceName,
+        tableName,
+        Arrays.asList("id text", "firstName text", "email list<text>"),
+        Collections.singletonList("id"),
+        null);
+    // We'll need another table in another keyspace to test [stargate#1463]
+    createTestTableIn(
+        altKeyspaceName,
         tableName,
         Arrays.asList("id text", "firstName text", "email list<text>"),
         Collections.singletonList("id"),
         null);
 
+    // Verify that neither keyspace (primary test; alt) have no indexes defined:
     String body =
         RestUtils.get(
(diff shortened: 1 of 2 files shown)


tatu-at-datastax:
@olim7t simple fix as per your suggestion.

mpenick (approved):

testing/src/main/java/io/stargate/it/http/RestApiv2SchemaTest.java
```diff
@@ -1186,6 +1206,17 @@ public void indexListAll() throws IOException {
...
+    // but also that the other table with same name (but in diff keyspace) does not
+    body =
+        RestUtils.get(
+            authToken,
+            String.format(
+                "%s/v2/schemas/keyspaces/%s/tables/%s/indexes",
+                restUrlBase, altKeyspaceName, tableName),
+            HttpStatus.SC_OK);
+    data = objectMapper.readValue(body, new TypeReference<List<Map<String, Object>>>() {});
+    assertThat(data).isEqualTo(Arrays.asList());
```
mpenick, line 1219: Curious, why `.isEqualTo(Array.asList())` vs `.isEmpty()`? Is there an advantage?

tatu-at-datastax:
Quick note: Stargate V2 listAllFilters works without changes due to the way it is implemented -- was not having this issue.

testing/src/main/java/io/stargate/it/http/RestApiv2SchemaTest.java
mpenick: Awesome idea! Thanks for sharing.

tatu-at-datastax: Glad this was useful! I have learned many things via code reviews here, esp. constructs others use that are new to me.

PR #1878, opened 2022-06-08
Add test for #1817: passes for SGv1/REST, will fail against SGv2/REST

### LINKED ISSUE #1817: Stargate V2 keyspace creation via REST API does not work with `datacenters` argument
Although simple keyspace creation with payload like:

```json
{ "name" : "keyspace1", "replicas":1 }
```

works (`replicas` being optional), it looks like more complicated case of

```
{ "name" : "keyspaceN",
   "datacenters" : [
       { "name":"dc1" },
       { "name":"dc2", "replicas":5}
   ]
}
```

fails. Code seems to assume that `datacenters` argument is already in format

```
{
  "dc1":1,
  "dc2":5
}
```

Test for this comes from a Postman collection used with Stargate V1 REST; there is no unit test for this yet.

tatu-at-datastax:
**What this PR does**:

Adds a unit test to verify #1817: passes against Stargate V1 REST API; will fail for SGv2/REST.

Does not yet fix an issue; will help fix #1817 for SGv2.

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/testing/src/main/java/io/stargate/it/http/RestApiv2SchemaTest.java
+++ b/testing/src/main/java/io/stargate/it/http/RestApiv2SchemaTest.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import net.jcip.annotations.NotThreadSafe;
 import org.apache.http.HttpStatus;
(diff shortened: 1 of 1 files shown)


tatu-at-datastax:
Noticed failures from CI due to C* setup on CI only configuring single dc, not two.
Since test is primarily verifying handling of requests and mapping to query builder, changed set up to do a single-DC keyspace creation, but using explicit DC definitions.
This is enough to cover specific reported bug.

PR #1890, opened 2022-06-14
Update java-driver-code dependency from 4.13.0 to 4.14.1

### LINKED ISSUE #1889: Update `java-driver-core` dependency to 4.14.1
Looks like there is at least one CVE related to version 4.13.0 of `java-driver-core`, picked up by our Snyk scanner.
Upgrade to 4.14.1 should fix this.

tatu-at-datastax:
**What this PR does**:

Updates `java-driver-core` dependency from 4.13.0 to 4.14.1 (latest), to address CVEs due to its dependencies.

Fixes #1889

**Checklist**
- [x] Changes manually tested -- actually tested via CI

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -721,6 +721,15 @@
         <artifactId>java-driver-core</artifactId>
         <version>${driver.version}</version>
       </dependency>
+      <!-- 14-Jun-2022, tatu: with java-driver-core 4.14.1 this dependency is optional
+            and not brought as compile-time dependency, so needs to be explicitly added
+           (see [stargate#1889] for details). But we better use consistent version
+        -->
+      <dependency>
+        <groupId>com.esri.geometry</groupId>
+        <artifactId>esri-geometry-api</artifactId>
+        <version>1.2.1</version>
+      </dependency>
       <dependency>
         <groupId>com.datastax.oss</groupId>
         <artifactId>java-driver-query-builder</artifactId>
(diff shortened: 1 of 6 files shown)


tatu-at-datastax:
Alas, DSE backend `GeoTypeTest` fails, due to driver 4.14.1 making dependency `com.esri.geometry:esri-geometry-api` from `compile-time` to `optional`. Will try to add the dependency next and try IT in question locally to see if this fixes the issue.

tatu-at-datastax:
Ok the issues was that nothing else actually depends on `persistence-dse-6.8` via Maven dependency, and so necessary types were not added transitively -- and an integration test would, upon failure (whether fatal or not) of geotypes handling get exception and fail to load the class (because OSGi).
The fix is to add the dependency either everywhere where `java-driver-core` is needed, or, seemingly safer, to `persistence-api` which is what typically brings in `java-driver-core`.

With this tests pass as expected.

In future we may also consider upgrading version of `esri-geometry-api` itself; but for now left it at version that driver itself defines (which is old). Upgrade possibly needed to upgrade further transitive dependencies.

pom.xml
```diff
@@ -716,6 +716,25 @@
         <artifactId>java-driver-shaded-guava</artifactId>
         <version>25.1-jre</version>
       </dependency>
+      <dependency>
```
tatu-at-datastax, line 719: Note: this is in `<dependencyManagement>` section so just defines global version to use, does not add actual dependency.

tatu-at-datastax:
@olim7t I think you might be the best person to review this -- just in case I forgot something. Change mostly needed to get Snyk happy, by upgrading `netty-codec` transitive dependency 4.1.60 -> 4.1.75 for Java driver (although has that baseline for persistence backends).

olim7t (approved):

PR #1903, opened 2022-06-22
Reduce GraphQLJava induced log noise

### LINKED ISSUE #1279: Configure GraphQL-java's `DataFetcherExceptionHandler` to reduce log noise
(note: see https://github.com/graphql-java/graphql-java/issues/2564 for the issue on grapql-java side)

Currently failures to fetch data for GraphQL query will get both indicated to caller and automatically logged (possibly multiple times). In case of automated logging at WARN level we will also always get a (lengthy) stack trace. Unfortunately this logging is done by graphql-java default handler, `SimpleDataFetcherExceptionHandler` and cannot be easily changed -- while it's not a big class and we could copy-paste, it seems better to try to use and configure default handler here, if possible.

An example of output (somewhat truncated):

```
Step #4 - "cassandra-4.0": WARN  [main] 2021-09-27 15:37:41,533 SimpleDataFetcherExceptionHandler.java:25 - Exception while fetching data (/m1) : graphql.GraphQLException: options can only de defined once in an @atomic mutation selection
Step #4 - "cassandra-4.0": java.util.concurrent.CompletionException: graphql.GraphQLException: options can only de defined once in an @atomic mutation selection
Step #4 - "cassandra-4.0": 	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:292)
Step #4 - "cassandra-4.0": 	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:308)
Step #4 - "cassandra-4.0": 	at java.util.concurrent.CompletableFuture.uniApply(CompletableFuture.java:607)
Step #4 - "cassandra-4.0": 	at java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:591)
Step #4 - "cassandra-4.0": 	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:488)
Step #4 - "cassandra-4.0": 	at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:1990)
Step #4 - "cassandra-4.0": 	at io.stargate.graphql.web.StargateGraphqlContext$BatchContext.setExecutionResult(StargateGraphqlContext.java:135)
Step #4 - "cassandra-4.0": 	at io.stargate.graphql.schema.cqlfirst.dml.fetchers.MutationFetcher.executeAsPartOfBatch(MutationFetcher.java:92)
Step #4 - "cassandra-4.0": 	at io.stargate.graphql.schema.cqlfirst.dml.fetchers.MutationFetcher.get(MutationFetcher.java:56)
Step #4 - "cassandra-4.0": 	at io.stargate.graphql.schema.cqlfirst.dml.fetchers.MutationFetcher.get(MutationFetcher.java:31)
Step #4 - "cassandra-4.0": 	at io.stargate.graphql.schema.CassandraFetcher.get(CassandraFetcher.java:33)
Step #4 - "cassandra-4.0": 	at graphql.execution.instrumentation.dataloader.DataLoaderDispatcherInstrumentation.lambda$instrumentDataFetcher$0(DataLoaderDispatcherInstrume
...
```

This happens about ~70 times during our CI tests but also affects products logging (depending on how many failures of these types are triggered by clients).

tatu-at-datastax:
**What this PR does**:

Upgrades version of `graphql-java` used and then uses new 18.x mechanism to override logging to reduce noise that currently clogs logs on cases where code handles exceptions already.
There is some positive impact for our CI/IT logs (gets rid of 73 non-error cases that look like errors), but more importantly should reduce production log noise.

Graphql-java fix itself here:

https://github.com/graphql-java/graphql-java/pull/2569

Fixes #1279

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/CassandraFetcherExceptionHandler.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/CassandraFetcherExceptionHandler.java
@@ -41,6 +41,5 @@ public class CassandraFetcherExceptionHandler extends SimpleDataFetcherException
   protected void logException(ExceptionWhileDataFetching error, Throwable exception) {
     // 24-Jun-2022, tatu: [stargate#1279] Since we handle exception at an outer level,
     //    should not log it here.
-    ;
   }
 }
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/CassandraFetcherExceptionHandler.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/CassandraFetcherExceptionHandler.java
@@ -15,6 +15,7 @@
  */
 package io.stargate.graphql.schema;
 
+import graphql.ExceptionWhileDataFetching;
 import graphql.execution.DataFetcherExceptionHandlerParameters;
 import graphql.execution.DataFetcherExceptionHandlerResult;
 import graphql.execution.SimpleDataFetcherExceptionHandler;
@@ -34,5 +35,12 @@ public class CassandraFetcherExceptionHandler extends SimpleDataFetcherException
     }
 
     return super.onException(handlerParameters);
+  }
+
+  @Override
+  protected void logException(ExceptionWhileDataFetching error, Throwable exception) {
+    // 24-Jun-2022, tatu: [stargate#1279] Since we handle exception at an outer level,
+    //    should not log it here.
+    ;
   }
 }
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/QueryFetcher.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/QueryFetcher.java
@@ -40,7 +40,7 @@ public class QueryFetcher extends DeployedFetcher<Object> {
 
   @SuppressWarnings("unchecked")
   private static final Coercing<ByteBuffer, String> BLOB_COERCING =
-      CqlScalar.BLOB.getGraphqlType().getCoercing();
+      (Coercing<ByteBuffer, String>) CqlScalar.BLOB.getGraphqlType().getCoercing();
 
   private final QueryModel model;
 
--- a/graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/CqlDirectives.java
+++ b/graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/CqlDirectives.java
@@ -635,7 +635,7 @@ public class CqlDirectives {
     ALL_AS_STRING =
         new SchemaPrinter(
                 SchemaPrinter.Options.defaultOptions()
-                    .includeDirectives(d -> !defaultDirectives.contains(d.getName()))
+                    .includeDirectives(d -> !defaultDirectives.contains(d))
                     .includeSchemaElement(e -> e != dummyQueryType))
             .print(schema);
 
--- a/graphqlapi/src/test/java/io/stargate/graphql/schema/GraphQlTestBase.java
+++ b/graphqlapi/src/test/java/io/stargate/graphql/schema/GraphQlTestBase.java
@@ -97,6 +97,9 @@ public abstract class GraphQlTestBase {
         GraphQL.newGraphQL(graphQlSchema)
             // Use parallel execution strategy for mutations (serial is default)
             .mutationExecutionStrategy(new AsyncExecutionStrategy())
+            // 24-Jun-2022, tatu: As per [stargate#1279] prevent graphql-java from adding
+            //    spurious logs at low level. Will not affect propagation of exceptions
+            .defaultDataFetcherExceptionHandler(CassandraFetcherExceptionHandler.INSTANCE)
             .build();
   }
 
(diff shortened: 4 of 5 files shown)


tatu-at-datastax:
Note: only version upgrade first, to verify new version of `graphql-java` works as-is (... or not).
And once it does, then add logging method override.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/CqlDirectives.java
```diff
@@ -635,7 +635,7 @@ public class CqlDirectives {
     ALL_AS_STRING =
         new SchemaPrinter(
                 SchemaPrinter.Options.defaultOptions()
-                    .includeDirectives(d -> !defaultDirectives.contains(d.getName()))
+                    .includeDirectives(d -> !defaultDirectives.contains(d))
```
ivansenic, line 638: This also looks strange, was this an error before?

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/fetchers/deployed/QueryFetcher.java
```diff
@@ -40,7 +40,7 @@ public class QueryFetcher extends DeployedFetcher<Object> {
 
   @SuppressWarnings("unchecked")
   private static final Coercing<ByteBuffer, String> BLOB_COERCING =
-      CqlScalar.BLOB.getGraphqlType().getCoercing();
+      (Coercing<ByteBuffer, String>) CqlScalar.BLOB.getGraphqlType().getCoercing();
```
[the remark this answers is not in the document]
tatu-at-datastax: This was necessary because of backwards-incompatible change where `getGraphqlType()` now returns "raw" type -- `Coercing<?,?>` -- instead actual types. I don't know why, and did not see mention in release notes (but there were many things so could have missed).
So I only changed it because this (or something else) is necessary due to API change. :-(

Maybe @olim7t has more context.

graphqlapi/src/main/java/io/stargate/graphql/schema/CassandraFetcherExceptionHandler.java
```diff
@@ -35,4 +36,11 @@ public DataFetcherExceptionHandlerResult onException(
 
     return super.onException(handlerParameters);
   }
+
+  @Override
+  protected void logException(ExceptionWhileDataFetching error, Throwable exception) {
```
tatu-at-datastax, line 41: So this is the actual fix: default implementation logs at warning level:

```
        logNotSafe.warn(error.getMessage(), exception);
```

graphqlapi/src/test/java/io/stargate/graphql/schema/GraphQlTestBase.java
```diff
@@ -97,6 +97,9 @@ public void setupEnvironment() {
         GraphQL.newGraphQL(graphQlSchema)
             // Use parallel execution strategy for mutations (serial is default)
             .mutationExecutionStrategy(new AsyncExecutionStrategy())
+            // 24-Jun-2022, tatu: As per [stargate#1279] prevent graphql-java from adding
```
tatu-at-datastax, line 100: @olim7t I hope there are no other side effects to be concerned about here? If there are, can just create an alternative test-exception-handler for the same purpose. But for now assuming production one works well.

graphqlapi/src/main/java/io/stargate/graphql/schema/CassandraFetcherExceptionHandler.java
```diff
@@ -35,4 +36,11 @@ public DataFetcherExceptionHandlerResult onException(
...
+    // 24-Jun-2022, tatu: [stargate#1279] Since we handle exception at an outer level,
+    //    should not log it here.
+    ;
```
tatu-at-datastax, line 44: Ugh. WTH? Codacy fails on this... what a clown show. Will change.

tatu-at-datastax:
Interesting. With changes applied in 2 places, number of logged warnings fell from 73 to 21 (i.e. most gone) but not to 0 as I had expected. Looks like there is yet another path through which `CassandraFetcherExceptionHandler` is not configured with `GraphQL.builder().defaultDataFetcherExceptionHandler(...)`.

tatu-at-datastax:
Ok now the noise is gone, ready for merge.

tatu-at-datastax:
@olim7t If you have a chance to have a look that'd be great: redundant logging is gone as per CI/IT output, and changes required by `grapql-java` seem straight-forward but since you know this codebase it'd be good to have another look.

graphqlapi/src/main/java/io/stargate/graphql/schema/graphqlfirst/processor/CqlDirectives.java
olim7t: Yeah it's a consequence of https://github.com/graphql-java/graphql-java/pull/2562. To be fair that PR mostly preserves backward compatibility, you can still use the incorrect `GraphQLDirective` in most places. But this is one exception where it's not possible (because the erasure of the two lambdas would be the same I think).

+1 to your change, however I think we should also migrate to `GraphQLAppliedDirective` where applicable (the old API still works, but better use the right one). I'll take a look.

olim7t: Actually it turns out that we mostly parse directive applications via the SDL API: `graphql.language.Directive`.

We have a few references to `GraphQLDirective`, but they relate to directive definitions, so they don't need to change.

olim7t (approved):

PR #1904, opened 2022-06-22
Test GH actions for CI

jeffreyscarpenter:
See what happens if we run integration tests using GH actions for a single profile (C* 4.0)

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -3,6 +3,7 @@ name: CI Experiment
 on:
   workflow_dispatch:
 
+jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -4,17 +4,19 @@ on:
   workflow_dispatch:
 
 jobs:
-  build:
-    name: Build
+  integration-test:
+    name: Integration Test
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           ref: "master"
-      - name: Setup Java JDK
-        uses: actions/setup-java@v1
+      - uses: actions/setup-java@v3
+        name: Setup Java JDK
         with:
-          java-version: 1.8
+          distribution: 'temurin'
+          java-version: 8
+          cache: 'maven'
       - name: Setup Maven
         env:
           ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_USERNAME }}
@@ -39,5 +41,5 @@ jobs:
           EOF
       - name: Build with Maven
         run: |
-          mvn verify -P it-cassandra-4.0
+          ./mvnw verify -P it-cassandra-4.0
 
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -1,0 +1,42 @@
+name: CI Experiment
+
+on:
+  workflow_dispatch:
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: "master"
+      - name: Setup Java JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: Setup Maven
+        env:
+          ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_USERNAME }}
+          ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_PASSWORD }}
+        run: |
+          mkdir -p ~/.m2
+          cat <<EOF > ~/.m2/settings.xml
+          <settings>
+            <servers>
+              <server>
+                <id>artifactory-snapshots</id>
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-test.yml
```diff
@@ -0,0 +1,43 @@
...
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
tatu-at-datastax, line 11: Not a big deal but there's v3 of `checkout` and `setup-java` too (couple of lines lower).
For `setup-java@v3` java-version needs to be `8` fwtw.

.github/workflows/ci-test.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          ref: "master"
+      - name: Setup Java JDK
+        uses: actions/setup-java@v1
+        with:
```
tatu-at-datastax, line 16: With `actions/setup-java@v3` there's also nice caching option; I have this for jackson:

```
      uses: actions/setup-java@v3
      with:
        distribution: 'temurin'
        java-version: ${{ matrix.java_version }}
        cache: 'maven'
```

(matrix just to test with Java 8, 11 and 17, here we can just put `8` in there)

tatu-at-datastax (approved):

.github/workflows/ci-test.yml
```diff
@@ -0,0 +1,43 @@
...
+                <id>artifactory-releases</id>
+                <username>${ARTIFACTORY_USERNAME}</username>
+                <password>${ARTIFACTORY_PASSWORD}</password>
+             </server>
+           </servers>
+          </settings>
+          EOF
+      - name: Build with Maven
+        run: |
+          mvn verify -P it-cassandra-4.0
```
ivansenic, line 42: should we use the wrapper please? and verify is doing the tests as well, was this the intention?

best would be imo:

* job 1 `Test` - ./mvnw test
* job 2 `Integration test C4` -  `/mvnw integration-test -P it-cassandra-4.0`
* job 3 `Integration test C3` -  `/mvnw integration-test`
* job 4 `Integration test DSE` -  `/mvnw integration-test -P dse`

First job can be prerequisite for the other three, which can run in parallel.. This will be the best, as you run unit tests only once and then only int tests for each profile..

.github/workflows/ci-test.yml
ivansenic: What's caching?

CONTRIBUTING.md
```diff
@@ -36,14 +36,12 @@ chances of your issue being dealt with quickly:
 ### <a name="submit-pr"></a> Submitting a Pull Request (PR)
 Before you submit your Pull Request (PR) consider the following guidelines:
 
-* Acceptance of the DataStax [contributor license agreement](https://cla.datastax.com/) (CLA) is required in order for us to be able to accept your code contribution. You will see this on the standard pull request checklist for the Stargate repositories.
-
```
ivansenic, line 40: was this intentional?

ivansenic (commented):
I like this, can we discuss further how can we go this path :) I also need smth similar for the docs api..

tatu-at-datastax:
@ivansenic `caching` would be, I think, for external Maven dependencies (keep non-empty Maven `~/.m2/repository` across builds I think) , to speed up builds.

ivansenic:
> @ivansenic `caching` would be, I think, for external Maven dependencies (keep non-empty Maven `~/.m2/repository` across builds I think) , to speed up builds.
> 
> Very quick overview here:
> 
> https://github.blog/changelog/2021-08-30-github-actions-setup-java-now-supports-dependency-caching/
> 
> and bit more on
> 
> https://github.com/actions/setup-java#caching-packages-dependencies

Ohh man this is awesome..

CONTRIBUTING.md
jeffreyscarpenter: I'll rebase

ivansenic:
This is merged but the experiment action is failing on the `ccm` creation.. Was the merge indented? Stupid ccm should be replaced with the test containers :)

tatu-at-datastax:
Yes, the idea was to get the initial branch closed and I would continue from it -- workflow is only manually triggered so while it is not functional, won't block anything.

.github/workflows/ci-test.yml
jeffreyscarpenter: @tatu-at-datastax: I like the approach @ivansenic recommends above, especially since it avoids an issue with the current GCB workflows in that each run executes the full suite of unit tests (i.e. all 3 persistence modules are tested in each of the 3 workflows). 

If our overall goal is to reduce runtime, we could even consider running all 4 jobs listed above in parallel? We're generally pretty good at checking in code that works :).

ivansenic: @jeffreyscarpenter you did not watch my _I code, I pollute_ presentation :smile: We should always confirm unit tests are green before firing the int tests.. This way we save unnecessary work if the those fail.. And those should be fast, under 2 mins for example..

ivansenic: But there is more to this, we cal always test docker image building as well for example..

PR #1905, opened 2022-06-23
closes #1733: added dead leaves deletion to the write bridge service

### LINKED ISSUE #1733: Implement dead leaf collection service in the Documents API V2
Since the existing implementation of the dead-leaf collection will not be ported to the V2, we need a new service `DeadLeafBridgeService` that executes a deletion of the dead leaves in the document table.

The collection of the dead leafs should be reviewed again, refactored if needed and documented properly as currently documentation is missing completely.

ivansenic:
**What this PR does**:

Dead leaves deletion for the documents V2. 

Fixes #1733

**Checklist**
- [ ] Resolve confusion about the delete all (TODO)..

### DIFF
--- a/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/DocumentWriteService.java
+++ b/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/DocumentWriteService.java
@@ -17,6 +17,7 @@
 
 package io.stargate.sgv2.docsapi.service.write;
 
+import com.google.common.base.Splitter;
 import io.opentelemetry.extension.annotations.WithSpan;
 import io.smallrye.mutiny.Uni;
 import io.stargate.bridge.proto.QueryOuterClass;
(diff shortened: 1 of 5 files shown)


ivansenic:
@EricBorczuk can you assist with the TODO please..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/DocumentWriteService.java
```diff
@@ -409,6 +416,94 @@ private Uni<ResultSet> deleteDocumentInternal(
...
+      } else if (DeadLeaf.ARRAY_LEAF.equals(deadLeaf)) {
+        deleteArray = true;
+      } else {
+        keysToDelete.add(deadLeaf.getName());
+      }
+    }
+
+    List<AbstractDeleteQueryBuilder> builders = new ArrayList<>();
+
+    // TODO Eric if we have delete all, why would we add this one below
```
EricBorczuk, line 487: This is a great point...

You could probably just rewrite this to
if (deleteAll) {
    // handle delete all
} else {
    if (deleteArray) {
    }
    if (keysToDelete) {
    }
}

Since if deleteAll is true you don't have to do the other two (but those other two could both fire off in theory)

EricBorczuk (approved):
This looks good!

I think you can rewrite the section you marked with `TODO` without any repercussion

PR #1911, opened 2022-06-24
Push images to ECR

mpenick:
Add images to our private AWS ECR registry as part of the release process.

**Checklist**
- [X] Changes manually tested

### DIFF
--- a/.github/workflows/release-v2.yml
+++ b/.github/workflows/release-v2.yml
@@ -166,7 +166,7 @@ jobs:
         uses: aws-actions/amazon-ecr-login@v1
       - name: Build and push (ECR)
         run: |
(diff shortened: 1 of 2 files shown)


mpenick:
Still testing this.

tatu-at-datastax:
Quick question: is this in addition to publishing to Docker Hub? I assume so, just verifying, since our users would be accessing images from some public registry.

jeffreyscarpenter:
I'm -1 on pushing images to private registry from our public repo.

tatu-at-datastax:
I think the problem here is that access to Dockerhub registry is unreliable due to rate limiting they do. So deployments can and do fail. This is not theoretical but actual problem being encountered and I remember this being a problem elsewhere too and (from what I understood) there is no simple solution.

So I think what we need is a some form of caching be that a completely separate registry, or some of pass-through cache.
It sounds like we should discuss the way we want to handle that.

PR #1922, opened 2022-06-28
relates to #1737: base github action for Docs API V2

ivansenic:
**What this PR does**:
Adds CI for the Docs V2 based on the GitHub actions.

Relates to #1737

### DIFF
--- a/.github/workflows/v2-docs-api.yaml
+++ b/.github/workflows/v2-docs-api.yaml
@@ -3,8 +3,11 @@ name: CI for Docs API V2
 on:
   push:
     branches: [ "v2.0.0" ]
+
   pull_request:
     branches: [ "v2.0.0" ]
+
+  workflow_dispatch:
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


ivansenic:
@tatu-at-datastax How does this actually works out, does this needs to be merged in order to check it out? Don't think so or?

tatu-at-datastax:
@ivansenic no need to merge, there is an option to run action from specific branch in UI.

ivansenic:
> @ivansenic no need to merge, there is an option to run action from specific branch in UI.
> 
> Although I guess in some cases merging is needed -- I think new workflows are only included from `master`. So if creating a new one it may be necessary to merge something, even just a placholder. And then UI will let you run it from any branch that has definition, I think.

@tatu-at-datastax but now I wrote the whole CI workflow.. Can I just merge this, would there be any side effects?

mpenick:
Is this just temporary? I'm fine with it as long as this is eventually unified with the other apis CI and release process.

ivansenic:
Thanks @mpenick, yes more improvements coming for sure..

PR #1926, opened 2022-06-29
Upgrade C*3 (3.11.13) and C*4 (4.0.4) dependencies

tatu-at-datastax:
**What this PR does**:

Upgrade C*3.11 and C*4.0 dependencies to 3.11.13 and 4.0.4, respectively

**Which issue(s) this PR fixes**:
Fixes #1920

**Checklist**
- [ ] Changes manually tested
- [x] Automated Tests added/updated
- [ ] Documentation added/updated
- [x] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

Issue #1920, opened 2022-06-27
Update persistence to C*3.11.13, 4.0.4


tatu-at-datastax:
There are newer versions of Cassandra 3.11 and 4.0 available so let's upgrade to:

* 3.11.13 (from 3.11.12)
* 4.0.4 (from 4.0.3)

One reason is to get a fix for:

https://issues.apache.org/jira/browse/CASSANDRA-17581

which affects our integration tests with latest Java 8 JDKs (one fail)

PR #1944, opened 2022-07-04
relates to #1920: update C* and DSE versions in the docker compose

ivansenic:
**What this PR does**:

Updates docker compose in the v2.0.0 after #1926.

Relates to #1920


tatu-at-datastax:
Thank you @ivansenic -- I forgot this part wrt v2, C*/DSE updates.

PR #1950, opened 2022-07-06
relates to #1730: document patch resource tests

### LINKED ISSUE #1730: REST Controllers for the write paths in the Document API V2
Implement 4 REST controllers in the Document V2 API, that handle following operations:

1. Write Controller (single and batch document writes)
2. Update Controller
3. Patch Controller
4. Delete Controller

Controllers should be based of the existing `ReactiveDocumentResourceV2` class, but based on the new framework. As with all REST controllers in the new versions, we should:

* Have fully reactive implementation
* Have a strong-typed DTO model
* Validates DTOs
* Have a well document API using the Open API v3

EricBorczuk:
**What this PR does**:

Fixes #1730 (adds tests)

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/WriteDocumentsService.java
+++ b/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/WriteDocumentsService.java
@@ -459,24 +459,20 @@ public class WriteDocumentsService {
     return readBridgeService
         .getDocumentTtlInfo(namespace, collection, documentId, ctx)
         .onItem()
+        .ifNotNull()
         .transform(
             rawDocument -> {
-              if (rawDocument == null) {
-                return 0;
-              }
               List<RowWrapper> rows = rawDocument.rows();
               if (rows.isEmpty()) {
                 return 0;
               }
-              try {
-                Long value =
-                    rows.get(0)
-                        .getLong(String.format("ttl(%s)", configuration.table().leafColumnName()));
+              RowWrapper row = rows.get(0);
+              String ttlColumn = "ttl(%s)".formatted(configuration.table().leafColumnName());
+              if (!row.isNull(ttlColumn)) {
+                Long value = row.getLong(ttlColumn);
                 return value.intValue();
-              } catch (IllegalArgumentException e) {
-                // getLong had a NULL value
-                return 0;
               }
+              return 0;
             })
         .onItem()
         .ifNull()
--- a/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/WriteDocumentsService.java
+++ b/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/WriteDocumentsService.java
@@ -13,6 +13,7 @@ import io.stargate.sgv2.docsapi.config.DocumentConfig;
 import io.stargate.sgv2.docsapi.service.ExecutionContext;
 import io.stargate.sgv2.docsapi.service.JsonDocumentShredder;
 import io.stargate.sgv2.docsapi.service.JsonShreddedRow;
+import io.stargate.sgv2.docsapi.service.common.model.RowWrapper;
 import io.stargate.sgv2.docsapi.service.query.ReadBridgeService;
 import io.stargate.sgv2.docsapi.service.schema.JsonSchemaManager;
 import io.stargate.sgv2.docsapi.service.util.DocsApiUtils;
@@ -459,12 +460,24 @@ public class WriteDocumentsService {
         .getDocumentTtlInfo(namespace, collection, documentId, ctx)
         .onItem()
         .transform(
-            rawDocument ->
-                rawDocument
-                    .rows()
-                    .get(0)
-                    .getLong(String.format("ttl(%s)", configuration.table().leafColumnName()))
-                    .intValue())
+            rawDocument -> {
+              if (rawDocument == null) {
+                return 0;
+              }
+              List<RowWrapper> rows = rawDocument.rows();
+              if (rows.isEmpty()) {
+                return 0;
+              }
+              try {
+                Long value =
+                    rows.get(0)
+                        .getLong(String.format("ttl(%s)", configuration.table().leafColumnName()));
+                return value.intValue();
+              } catch (IllegalArgumentException e) {
+                // getLong had a NULL value
+                return 0;
+              }
+            })
         .onItem()
         .ifNull()
         .switchTo(() -> Uni.createFrom().item(0));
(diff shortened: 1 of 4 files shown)


sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/WriteDocumentsService.java
```diff
@@ -459,12 +460,24 @@ private Uni<Integer> determineTtl(
...
-            rawDocument ->
-                rawDocument
-                    .rows()
-                    .get(0)
-                    .getLong(String.format("ttl(%s)", configuration.table().leafColumnName()))
-                    .intValue())
+            rawDocument -> {
+              if (rawDocument == null) {
+                return 0;
+              }
```
ivansenic, line 466: can we replace this with

`.onItem().IfNotNull()`

you are handling the null case at the end

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/WriteDocumentsService.java
```diff
@@ -459,12 +460,24 @@ private Uni<Integer> determineTtl(
...
+              try {
+                Long value =
+                    rows.get(0)
+                        .getLong(String.format("ttl(%s)", configuration.table().leafColumnName()));
+                return value.intValue();
+              } catch (IllegalArgumentException e) {
+                // getLong had a NULL value
+                return 0;
+              }
```
ivansenic, line 479: check that the `columnExists` to avoid throwing this exception all the time, as this is on the valid write path (exceptions are not free):

```
RowWrapper wrapper = rows.get(0);
String column ="ttl(%s)".formatted(configuration.table().leafColumnName());
if (wrapper.columnExists(column) {
    return wrapper.getLong(column);
} else {
   return 0;
}
```

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/DocumentPatchResourceIntegrationTest.java
```diff
@@ -0,0 +1,626 @@
...
+    }
+
+    @Test
+    public void happyPathNoCollection() {
+      given()
+          .header(Constants.AUTHENTICATION_TOKEN_HEADER_NAME, "")
+          .header("Content-Type", "application/json")
+          .body(DEFAULT_PAYLOAD)
+          .when()
+          .patch(BASE_PATH + "/test", DEFAULT_NAMESPACE, "newtable")
```
ivansenic, line 129: "newtable" - make random

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/DocumentPatchResourceIntegrationTest.java
```diff
@@ -0,0 +1,626 @@
...
+  @Nested
+  class PatchDocument {
+    @Test
+    public void happyPath() {
+      given()
+          .header(Constants.AUTHENTICATION_TOKEN_HEADER_NAME, "")
+          .header("Content-Type", "application/json")
+          .body(DEFAULT_PAYLOAD)
+          .when()
+          .patch(BASE_PATH + "/newdoc", DEFAULT_NAMESPACE, DEFAULT_COLLECTION)
```
ivansenic, line 75: can you please use random doc ids everywhere, so that there is no chance of mixup in the tests

and also the call in the way

```
.patch(BASE_PATH + "/{document-id}", DEFAULT_NAMESPACE, DEFAULT_COLLECTION, id)
```

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/DocumentPatchResourceIntegrationTest.java
```diff
@@ -0,0 +1,626 @@
...
+        .atMost(Duration.ofSeconds(10));
+
+    tableManager
+        .createCollectionTable(DEFAULT_NAMESPACE, DEFAULT_COLLECTION)
+        .await()
+        .atMost(Duration.ofSeconds(10));
+  }
+
+  @Nested
+  class PatchDocument {
```
ivansenic, line 67: missed V1 tests:

* testRootDocumentPatchNulls
* arbitrary table patch

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/DocumentPatchResourceIntegrationTest.java
```diff
@@ -0,0 +1,626 @@
...
+          .then()
+          .statusCode(404)
+          .body("code", equalTo(404))
+          .body(
+              "description", equalTo("Unknown namespace notakeyspace, you must create it first."));
+    }
+  }
+
+  @Nested
+  class PatchDocumentPath {
```
ivansenic, line 476: missed v1 tests:
*   testSubDocumentPatch

sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/DocumentPatchResourceIntegrationTest.java
ivansenic: Arbitrary table might be also missing from the write tests as well..

ivansenic (changes requested):
* Seems a bunch of V1 tests were missed here
* This needs to be re-based on top of the `origin/v2.0.0` without the write tests

Come on, let's wrap this up..

sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/WriteDocumentsService.java
EricBorczuk: The problem here is that the column exists, but getting its value is returning a null value when a non-null number value is expected (that is the IllegalArgumentException I am intercepting). I tried to change it to what you wrote above but am still encountering the same issue; an unexpected error occurs reading `Expected INT value, received NULL`

In order to fix this, I would probably have to change the way values are accessed in RowWrapper, or maybe create a "getOptionaLong" method instead...

ivansenic: Wait is there any way we can add the `wrapper.isNull()` call? @olim7t quick help?

ivansenic: @EricBorczuk could be this:

  boolean isNull(String columnName) {
    int i = firstIndexOf(columnName);
    if (i < 0) {
      throw new IllegalArgumentException(String.format("Column '%s' does not exist", columnName));
    }
     Value value = row().getValues(i);
     return value.hasNull();
  }

EricBorczuk: Yep this would work, I'll add it

ivansenic: Forget it, there is already `isNull` method!

PR #1954, opened 2022-07-06
Fix for #1953: add ability to override DropWizard app config file

### LINKED ISSUE #1953: Allow yaml configuration file overrides per-service
Currently all Stargate V1 services are based on DropWizard, and include `src/main/resources/config.yaml` for their default settings. While these settings can theoretically be overridden with System Properties (`-Ddw.http.port=8088`), there is no prefixing to avoid changing settings for all.
Since yaml configuration files are baked in jars, changing them is difficult as well.

To support things like configuring https endpoints for services it makes sense to allow specifying alternate configuration files, on per-service basis. This can be done by, for example:

1. Adding new System properties that, if specified, will point to file to read instead (`./my-service-config/rest-config.yaml`) of the default resource
2. Using naming convention (for example `[SERVICE-NAME]-config.yaml`, like `restapi-config.yaml`) and check for existence of such file, and if found using as replacement

It seems simple enough to do both, in this order of precedence.

tatu-at-datastax:
**What this PR does**:

Adds 2 ways to override per-service DropWizard config file (default: resource `config.yaml`):

1. System property `stargate.configurationFile.[MODULENAME]` to specify file reference
2. Naming convention: if file `[MODULENAME]-config.yaml` found, use that

to customize each service (MODULENAME, like `restapi`) as necessary.

As part of changes will also rename generic `config.yaml`defaults to `[MODULENAME]-config.yaml` for easier troubleshooting, consistency.

Fixes #1953

**Checklist**
- [ ] Changes manually tested


tatu-at-datastax:
Note: addition to "metrics-jersey" not optimal but it is the only shared module that already has DropWizard dependencies.

Also note that this won't work with Stargate V2.

tatu-at-datastax:
Manually tested that overrides work for REST API, added support for GraphQL and Auth API end points too.

metrics-jersey/src/main/java/io/stargate/metrics/jersey/dwconfig/StargateV1ConfigurationSourceProvider.java
```diff
@@ -0,0 +1,66 @@
...
+ *   <li>Explicit System Property of form "stargate.configurationFile.${MODULENAME}" specifying
+ *       exact filename of the configuration file to use
+ *   <li>Existence of file "${MODULENAME}-${CONFIGBASE} which is read as the configuration
+ *   <li>Classpath Resource with name "${CONFIGBASE}"
+ * </ol>
+ *
+ * (where "${MODULENAME}" and "${CONFIGBASE} are passed as arguments; former typically being
+ * something like {@code "restapi"} and latter {@code "config.yaml"})
+ */
+public class StargateV1ConfigurationSourceProvider implements ConfigurationSourceProvider {
```
ivansenic, line 25: and `core` does not work, this is really not the place for this.. but I get why you did it..

ivansenic, line 25: needs a proper unit test

ivansenic (commented):
Just had a quick glance on this.. What is the reason to have this? The real use case?

tatu-at-datastax:
> Just had a quick glance on this.. What is the reason to have this? The real use case?

A user wants to expose Stargate endpoints as http; configuration like https://stackoverflow.com/questions/35002780/setting-up-ssl-in-dropwizard works fine as long as there's a way to provide custom YAML config.
System property overrides (`-Ddw.http.port=9090`) won't work as they are global and would affect all endpoints (and would be bit inconvenient).

metrics-jersey/src/main/java/io/stargate/metrics/jersey/dwconfig/StargateV1ConfigurationSourceProvider.java
tatu-at-datastax: Agreed, not a great place but best of poor choices. And does seem right to add a new module for 1 class, for Stargate v1 only.

ivansenic:
> > Just had a quick glance on this.. What is the reason to have this? The real use case?
> 
> A user wants to expose Stargate endpoints as http; configuration like https://stackoverflow.com/questions/35002780/setting-up-ssl-in-dropwizard works fine as long as there's a way to provide custom YAML config. System property overrides (`-Ddw.http.port=9090`) won't work as they are global and would affect all endpoints (and would be bit inconvenient).

Understood. Yea, stupid OSGi hitting us again..

metrics-jersey/src/main/java/io/stargate/metrics/jersey/dwconfig/StargateV1ConfigurationSourceProvider.java
tatu-at-datastax: Tests added.

auth-api/src/main/java/io/stargate/auth/api/impl/AuthApiRunner.java
```diff
@@ -36,6 +36,6 @@ public AuthApiRunner(
   public void start() {
     AuthApiServer authApiServer =
         new AuthApiServer(this.authenticationService, this.metrics, this.httpMetricsTagProvider);
-    authApiServer.run("server", "config.yaml");
+    authApiServer.run("server", "authapi-config.yaml");
```
tatu-at-datastax, line 39: Mentioned in description: change `config.yaml` to service-specific to simplify overrides (can use same name and there is no confusion on what is used where) and as minor documentation. There is no benefit from having same name in the first place.

metrics-jersey/pom.xml
```diff
@@ -23,6 +23,10 @@
       <artifactId>micrometer-core</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>io.dropwizard</groupId>
+      <artifactId>dropwizard-configuration</artifactId>
```
tatu-at-datastax, line 28: Needed just for type `ConfigurationSourceProvider` that we implement and register to allow configuration resource/file overrides.

tatu-at-datastax:
Notes on configuration: to test https/ssl endpoints, you can create file `restapi-config.yaml` at the current working directory (repo root) when starting Stargate (with `starctl` f.ex).
Contents should look something like: (note: this is edited from original `restapi/src/main/resources/restapi-config.yaml`):

```yaml
server:
  type: rest-api
  applicationContextPath: /
  connector:
    type: https
    port: 8443
    keyStorePath: ./keystore/myKeyStore.jks
    keyStorePassword: "astraRulezOk"
    validateCerts: true
  requestLog:
    type: external
logging:
    type: external
```

To generate `keystore/myKeyStore.jks`, follow instructions on https://docs.oracle.com/cd/E19798-01/821-1841/gjrgy/  (or similar)

https://docs.oracle.com/cd/E19798-01/821-1841/gjrgy/

jeffreyscarpenter (approved):

PR #1958, opened 2022-07-07
closes #1736: upgrade actions for collections

### LINKED ISSUE #1736: Support document table upgrade functions in Document API V2
This issue groups all work needed for supporting the document table upgrade functions. Currently there is only one function:

1. `SAI_INDEX_UPGRADE`

The issue should refactor the way how the upgrade functions are modeled, allowing easier addition of new functions in the future. The implementation should cover existence of the service for the function execution as well as the resource for exposing functionality over the HTTP. 

Note that functions should be checked against the configuration in order to understand if it's possible to execute or not.

ivansenic:
**What this PR does**:

Adds the upgrade possibility for the collection.

I have to say that this feels a bit stragte, not sure if we should abandon this and simply always return there is no available upgrade and always throw an error on the `/upgrade` HTTP call. It's also pretty stupid that current API defines only a single possible upgrade type for each collection. It's also stupid that I can not actually test this in integration tests. However, now when we do have everything implemented in a  way to easily support additional upgrade types, maybe we can just merge it.

If we decide to merge, I have some questions:
* Is it OK if I do a simple manual test only (disable index creation on collection creation and thus be able to perform the upgrade)?
* Should upgrade stuff be outside `TableManager` or not?

I will add additional tests if we are to merge it..

**UPDATE: As agreed with @EricBorczuk we decided that upgrade `SAI_INDEX_UPGRADE`, as the only upgrade existing, does not make much sense. Thus, we are removing the support for it, keeping the API fully backward compatible and correct. In the future, we might fix the upgrade feature and add more possibilities that do make sense.**

Fixes #1736

**Checklist**
- [x] Changes manually tested
- [x] Re-check API docs


ivansenic:
@tatu-at-datastax can you give me an approval here as @EricBorczuk is out and you were in the loop about the issue.. Thx

PR #1959, opened 2022-07-07
closes #1761: copy and implement needed comprable bytes API for D…

### LINKED ISSUE #1761: Use comparable bytes API in persistence-cassandra-3.11 and persistence-cassandra-4.0
The #1716  extended the `RowDecorator` interface, however the implementations in the 3.11 and 4.0 depend on the https://github.com/apache/cassandra/pull/1294. We are leaving this for the end of the docs-api V2 work.

ivansenic:
_Draft, in-progress._

**What this PR does**:
As we depend on the comparable bytes API that are being release in the C* 4.2, we need to temporary copy the peace of it, so we do support C* 3 & 4 in Docs API V2. We will do so by supporting:

* `Murmur3Partitioner`
* `ByteOrderedPartitioner`
* `OrderPreservingPartitioner`

Other partitioners will be supported once 4.2 is released. This is also when we will remove this code and use the available code in the C*.

Fixes #1761 

**Checklist**
- [x] Enable Docs API V2 int tests
- [x] Support for C*3
- [x] Mention partitioner limitations in README
- [x] Stricter `io.stargate.it.PersistenceTest#testRowDecoratorComparableBytes`

### DIFF
--- a/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/RowDecoratorImpl.java
+++ b/persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/RowDecoratorImpl.java
@@ -126,7 +126,10 @@ public class RowDecoratorImpl extends AbstractRowDecorator {
     byte[] buf = new byte[initialBufferCapacity];
     int data;
     while ((data = byteSource.next()) != ByteSource.END_OF_STREAM) {
-      buf = ensureCapacity(buf, readBytes);
+      // inlined com.datastax.bdp.db.tries.util.ByteSourceUtil#ensureCapacity
+      if (readBytes == buf.length) {
+        buf = Arrays.copyOf(buf, readBytes * 2);
+      }
       buf[readBytes++] = (byte) (data & 0xFF);
     }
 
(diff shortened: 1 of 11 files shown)


persistence-cassandra-4.0/src/main/java/io/stargate/db/cassandra/impl/RowDecoratorImpl.java
```diff
@@ -43,17 +54,98 @@ private RowDecoratorImpl(TableName tableName, TableMetadata tableMetadata) {
...
+
+    // https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/dht/OrderPreservingPartitioner.java#L206-L210
+    if (token instanceof OrderPreservingPartitioner.StringToken) {
+      String tokenValue = (String) token.getTokenValue();
+      return ByteSource.of(tokenValue, version);
+    }
+
+    // we only support the above 3
+    // for other partitioner, log warn and wrap empty bytes
+    LOG.warn(
```
tatu-at-datastax, line 115: Is this only warning and not error wrt logging? Do we expect that this would actually happen?

ivansenic: And with all that, comparable bytes are only needed for `$or` searches.. Thus, I just actually created https://github.com/stargate/stargate/issues/1960

ivansenic:
Good, confirmed that C* 4.0 build passes now and C* 3.11 fails.. Will continue by fixing that and re-enabling all tests in the Docs API V2.

tatu-at-datastax (approved):

EricBorczuk (approved):

ivansenic:
Screw you codacy..

PR #1966, opened 2022-07-11
closes #1903: decrease fetch of the comparable bytes

### LINKED ISSUE #1960: Enhanced querying is not needed always in the `QueryExecutor`
In case when we have only one query, we don't have to fetch comparable bytes. Meaning that when we have a single query with no resume mode, we can save a lot by not requesting the enhanced bridge query execution.

ivansenic:
**What this PR does**:
Decreases the usage of generating comparable bytes, by setting enriched only when we have two queries or resume mode.. This will have considerable impact, especially in cases when we are populating documents..

Note that the resume mode could be isolated, and we can safely only request comparable bytes once we have multiple queries.. But currently bridge inspects the enriched flag for deciding if resume mode will be handled or not, thus we have a separate ticket for this (#1967).

Fixes #1960

**Checklist**

### DIFF
--- a/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/executor/DocumentPropertyComparator.java
+++ b/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/executor/DocumentPropertyComparator.java
@@ -1,11 +1,19 @@
 package io.stargate.sgv2.docsapi.service.query.executor;
 
+import com.google.protobuf.ByteString;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
 import org.apache.commons.lang3.StringUtils;
 
 public class DocumentPropertyComparator implements Comparator<DocumentProperty> {
+
+  /**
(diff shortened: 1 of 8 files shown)


sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/executor/DocumentPropertyComparator.java
```diff
@@ -1,12 +1,20 @@
 package io.stargate.sgv2.docsapi.service.query.executor;
 
+import com.google.protobuf.ByteString;
```
tatu-at-datastax, line 3: Not necessarily ideal to have this dependency (since this is not protobuf related?), but I guess it's fine as we already have this dependency for Bridge access.

ivansenic: Hmm but the bridge model data does expose this for bytes transfer.. Am I getting something wrong?

PR #1979, opened 2022-07-15
Briefest way to implement TLS configuration for CQL

### LINKED ISSUE #1001: Cql client encryption is not working

mpenick:
**What this PR does**:

This is one possible way of enable client encryption (TLS) for CQL. This uses the existing system property "stargate.unsafe.cassandra_config_path" to load the client encryption options from a `cassandra.yaml` (format) type file. 

It could be argued that a another approach would be to implement this in a way that's specific to Stargate using either CLI flags and/or a Stargate-specific configuration file. That approach might be nice, but it will not be a quick or simple to implement.

Fixes #1001 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/cql/src/main/java/io/stargate/cql/CqlActivator.java
+++ b/cql/src/main/java/io/stargate/cql/CqlActivator.java
@@ -22,12 +22,15 @@ import io.stargate.cql.impl.CqlImpl;
 import io.stargate.db.DbActivator;
 import io.stargate.db.Persistence;
 import io.stargate.db.metrics.api.ClientInfoMetricsTagProvider;
+import java.io.File;
+import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.net.InetAddress;
-import java.net.UnknownHostException;
 import java.util.Arrays;
 import java.util.List;
 import javax.annotation.Nullable;
 import org.apache.cassandra.config.Config;
+import org.apache.cassandra.config.YamlConfigurationLoader;
 
 public class CqlActivator extends BaseActivator {
   private CqlImpl cql;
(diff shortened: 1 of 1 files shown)


cql/src/main/java/io/stargate/cql/CqlActivator.java
```diff
@@ -86,6 +89,19 @@ protected List<ServicePointer<?>> dependencies() {
 
   private static Config makeConfig() {
     try {
+      Config c;
+
+      // This is the same system property as used by persistence backends for custom configuration
+      // using `cassandra.yaml`.
+      String cassandraConfigPath = System.getProperty("stargate.unsafe.cassandra_config_path", "");
```
mpenick, line 96: Thinking about this more. It's a bad idea to use this existing system property name because the configuration yaml format and available configuration could be different between CQL and the persistence backend's `cassandra.yaml` format. One example is audit logging, the available options is different between DSE and Cassandra 4.0.

PR #1980, opened 2022-07-15
Convert GraphQL v2 to Quarkus

olim7t:
Main changes:
- `GraphqlResourceBase` and subclasses: use injection to pass the bridge and GraphqlCache, adapt multipart handling code for Quarkus.
- `GraphqlCache`: use Quarkus config for enableDefaultKeyspace

Everything else is pretty much a copy-paste of the existing `sgv2-graphqlapi`, with a few changes to the imports. Note that fetcher implementations still rely on the synchronous methods of `StargateBridgeClient`; with the async/reactive bridge it would now be possible to make all fetchers async, but I consider this beyond the scope of this PR, see #421.

TODO:
- [x] find a way to add `stargate.graphql.enable-playground` to the config (how do we condition the inclusion of a REST resource at runtime?)
- [x] adapt (move?) the integration tests in `testing`
- [x] remove the original `sgv2-graphqlapi` at the root of the repo
- [x] handle OVERLOADED error (see `GraphqlResourceBase.executeAsync`)


tatu-at-datastax (approved):

apis/sgv2-graphqlapi/src/main/java/io/stargate/sgv2/graphql/web/resources/GraphqlCache.java
```diff
@@ -145,11 +146,12 @@ CompletionStage<Optional<GraphQL>> getGraphql(StargateBridgeClient bridge) {
...
           bridge.getKeyspaceAsync(keyspaceName, true);
 
-      return keyspaceFuture.thenCompose(
+      return keyspaceFuture.thenComposeAsync(
           maybeKeyspace ->
               maybeKeyspace
                   .map(keyspace -> handleExisting(keyspace, bridge))
-                  .orElse(handleMissing()));
+                  .orElse(handleMissing()),
+          Infrastructure.getDefaultWorkerPool());
```
olim7t, line 154: 1. Not super happy about that. I'm aware that the correct solution would be to rewrite everything from scratch to use async/reactive bridge methods only, but from a practical standpoint I'm not sure if we want to invest this effort right now.
2. I'm not sure if this is the best executor to use, I haven't found good docs about Quarkus' threading model.

apis/sgv2-graphqlapi/src/test/java/io/stargate/sgv2/graphql/integration/util/GraphqlIntegrationTest.java
```diff
@@ -0,0 +1,67 @@
...
+import io.stargate.bridge.proto.QueryOuterClass.QueryParameters;
+import io.stargate.bridge.proto.StargateBridge;
+import io.stargate.sgv2.api.common.StargateRequestInfo;
+import java.time.Duration;
+import javax.inject.Inject;
+import org.apache.commons.lang3.RandomStringUtils;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+
+public abstract class GraphqlIntegrationTest {
```
olim7t, line 32: For now, this relies on inheritance, which is a bit underwhelming. Requests are executed through the bridge.

I would prefer to port `CqlSessionExtension` from the v1 codebase, but it's not yet clear to me if/how a Junit extension can interact with a Quarkus test extension.

ivansenic: Yes, we should be able to have the `CqlSessionExtension` here as well, that communicates over the bridge or maybe even directly over the CQL port.. For this to happen, we need to expose `host` and `port` of the target CQL port in the integration tests.. Then you can use Cassandra driver to push needed cql statements.. 

I might be working on this while you are on holiday.. Let's sync once you are back..

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/common/CqlEnabledIntegrationTestBase.java
```diff
@@ -0,0 +1,74 @@
...
+import org.junit.jupiter.api.TestInstance;
+
+/**
+ * Base test class that exposes a Java driver {@link CqlSession} connected to the Stargate
+ * coordinator, allowing subclasses to create and populate their CQL schema.
+ *
+ * <p>Subclasses must be annotated with {@link io.quarkus.test.junit.QuarkusIntegrationTest}.
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public abstract class CqlEnabledIntegrationTestBase {
```
ivansenic, line 38: cool stuff, can be used in docs api as well.. Do you think in future having `CqlSession` like stuff as annotation would make sense? Or this is enough?

docker-compose/cassandra-3.11/docker-compose-dev-mode.yml
```diff
@@ -33,18 +33,5 @@ services:
...
-      - coordinator
-    networks:
-      - stargate
-    ports:
-      - 9080:9080
-    mem_limit: 2G
-    environment:
-      - STARGATE_BRIDGE_HOST=coordinator
-      - STARGATE_BRIDGE_PORT=8091
-      - STARGATE_GRAPHQL_PORT=9080
```
ivansenic, line 48: why removing the graphql from docker compose? we should still have it imo.. I know there is a ticket to add docs api now..

This applies to all docker related files below..

apis/sgv2-graphqlapi/src/test/java/io/stargate/sgv2/graphql/integration/cqlfirst/ApolloIntegrationTest.java
```diff
@@ -24,50 +24,59 @@
...
+@QuarkusIntegrationTest
+@TestProfile(IntegrationTestProfile.class)
+@ActivateRequestContext
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class ApolloIntegrationTest extends ApolloIntegrationTestBase {
 
   @Test
   public void getOrdersByValue() throws ExecutionException, InterruptedException {
-    ApolloClient client = getApolloClient("/graphql/" + keyspace);
+    ApolloClient client = getApolloClient("/graphql/" + keyspaceId.asInternal());
```
ivansenic, line 79: maybe `keyspaceId.asInternal()` should be a method in the `CqlFirstIntegrationTest`

apis/sgv2-graphqlapi/src/test/java/io/stargate/sgv2/graphql/integration/util/GraphqlIntegrationTest.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ */
+package io.stargate.sgv2.graphql.integration.util;
+
+import io.quarkus.test.common.http.TestHTTPResource;
+import io.stargate.sgv2.common.CqlEnabledIntegrationTestBase;
+
+public abstract class GraphqlIntegrationTest extends CqlEnabledIntegrationTestBase {
+
+  @TestHTTPResource protected String baseUrl;
```
ivansenic, line 24: will `@TestHTTPResource` work with `@QuarkusIntegrationTest`?

apis/sgv2-graphqlapi/src/test/java/io/stargate/sgv2/graphql/integration/util/GraphqlClient.java
```diff
@@ -115,4 +128,29 @@ protected String urlEncode(String parameter) {
...
+              "Timeout (%s) for GET request from URL '%s': service not running?",
+              request.url(), e.getClass().getName()));
+    }
+    assertThat(response.code()).isEqualTo(expectedStatusCode);
+
+    ResponseBody body = response.body();
+    assertThat(body).isNotNull();
+
+    return body.string();
+  }
```
ivansenic, line 155: why not using rest assured here that's given as default test framework for HTTP? we could get rid of the OkHttp?

ivansenic (commented):
I tried my best to give as much feedback as possible.. I think we should bring to the merging state as soon as possible and then have tickets for future improvements..

General notes:

* the api CI workflow was not extended to include `sgv2-graphql` for the integration tests.. If you disable fixed token in the int-tests, then you can to add your service to be docker-based tested
* can we get the set of tickets for improvements, for example removing `StargateBridgeClient`, async fetchers if possible and going into reactive programming model, etc (you know better)
* nit: banner setup i smissing

apis/sgv2-graphqlapi/src/main/java/io/stargate/sgv2/graphql/schema/FileSupport.java
```diff
@@ -0,0 +1,62 @@
...
+            String.format(
+                "Expected a FileInputStream, got %s",
+                input == null ? "null" : input.getClass().getSimpleName()));
+      }
+    }
+
+    @Override
+    public InputStream parseLiteral(Object input) throws CoercingParseLiteralException {
+      throw new CoercingParseLiteralException("Upload values must be specified as JSON variables");
+    }
```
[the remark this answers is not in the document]
olim7t: It's the enclosing class that's static, this method has to be implemented because it's defined by the interface.

apis/sgv2-graphqlapi/src/test/java/io/stargate/sgv2/graphql/integration/util/GraphqlIntegrationTest.java
olim7t: Yes, I've confirmed that it works.

apis/sgv2-graphqlapi/src/test/java/io/stargate/sgv2/graphql/integration/util/GraphqlClient.java
olim7t: Honestly out of laziness, adapting the existing code was quicker. But if RestAssured has a low-level querying API I can take a look.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/StargateBridgeClientImpl.java
```diff
@@ -34,7 +35,7 @@
 public class StargateBridgeClientImpl implements StargateBridgeClient {
 
   private static final Function<String, Uni<? extends Schema.CqlKeyspaceDescribe>>
-      MISSING_KEYSPACE = ks -> Uni.createFrom().failure(new UnauthorizedKeyspaceException(ks));
+      MISSING_KEYSPACE = ks -> Uni.createFrom().failure(MissingKeyspaceException.INSTANCE);
```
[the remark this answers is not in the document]
olim7t: Ah, good to know, I'll make the change.

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/common/CqlEnabledIntegrationTestBase.java
olim7t: This can probably be turned into a test resource, now that the contact point info is passed as system properties. The only requirement is that `StargateTestResource` needs to execute first, do you know if there is a way to enforce a reliable order?

apis/sgv2-graphqlapi/src/main/java/io/stargate/sgv2/graphql/web/resources/GraphqlResourceBase.java
```diff
@@ -0,0 +1,283 @@
...
+import javax.ws.rs.container.AsyncResponse;
+import javax.ws.rs.container.Suspended;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import org.jboss.resteasy.reactive.multipart.FileUpload;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Produces(MediaType.APPLICATION_JSON)
+public abstract class GraphqlResourceBase {
```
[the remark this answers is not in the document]
olim7t: Correct, the only change was to make the context object customizable, to simplify unit testing.

apis/sgv2-graphqlapi/src/main/resources/application.yaml
```diff
@@ -0,0 +1,49 @@
...
+      keyspace-cache:
+        maximum-size: 1000
+        expire-after-access: PT5M
+        metrics-enabled: true
+
+  grpc:
+    clients:
+      bridge:
+        host: ${STARGATE_BRIDGE_HOST:localhost}
+        port: ${STARGATE_BRIDGE_PORT:8091}
```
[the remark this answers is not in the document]
olim7t: Just because that's how it was done in pre-Quarkus v2 images. It makes it easier to adapt docker-compose files and operator code.

apis/sgv2-graphqlapi/src/main/resources/application.yaml
```diff
@@ -0,0 +1,49 @@
...
+      module: sgv2-graphqlapi
+    tenant-request-counter:
+      enabled: ${stargate.multi-tenancy.enabled}
+
+  multi-tenancy:
+    enabled: false
+
+quarkus:
+
+  http:
```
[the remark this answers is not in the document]
olim7t: It uses the defaults defined by `AuthConfig`.

apis/sgv2-graphqlapi/src/main/java/io/stargate/sgv2/graphql/web/resources/StargateGraphqlContext.java
```diff
@@ -21,27 +21,23 @@
...
+import io.stargate.sgv2.api.common.grpc.StargateBridgeClient;
 import io.stargate.sgv2.graphql.schema.CassandraFetcher;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.atomic.AtomicReference;
-import javax.servlet.http.HttpServletRequest;
 
 public class StargateGraphqlContext {
```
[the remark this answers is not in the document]
olim7t: This context object is passed to the GraphQL engine (see `input.context(...)` calls in `GraphqlResourceBase`), then it can be retrieved from fetchers with `DataFetchingEnvironment.getContext()`. So in a way it's GraphQL doing its own simplified version of injection.

Turning it into a bean would be feasible I guess, but then we'd need to make sure that every fetcher is a managed bean as well... Overall I don't think it's worth it.

apis/sgv2-graphqlapi/src/main/java/io/stargate/sgv2/graphql/web/resources/GraphqlCache.java
olim7t: Ticket for async: #421

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/common/CqlEnabledIntegrationTestBase.java
olim7t: Ok apparently there is `QuarkusTestResourceLifecycleManager.order()`, I'll give it a try.

olim7t: Hrmm actually a test resource might not be the best fit:

> These resources are started before the first test is run, and are closed at the end of the test suite.

There's no class-level lifecycle that would allow us to create the session and a dedicated keyspace before each test.

apis/sgv2-graphqlapi/README.md
```diff
@@ -253,21 +250,10 @@ The easiest way is to use the built-in GraphQL playground at http://localhost:80
...
-The integration tests are located in the [testing] module. There use two different approaches:
-
-* the tests that extend [ApolloTestBase] use the [Apollo client] library. They provide a good
-  example of using a "real-world" client, but require a lot of boilerplate: external definition
-  files, code generation...
-* the other tests use a more lightweight approach that takes GraphQL queries as plain strings, and
-  returns raw JSON responses. See [GraphqlClient] and its subclasses.
-
-If you add new tests, please favor the lightweight approach.
+<!-- TODO integration tests -->
```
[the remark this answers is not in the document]
olim7t: Instead of copy-pasting, maybe it should move to the top-level readme?

apis/sgv2-graphqlapi/src/main/java/io/stargate/sgv2/graphql/web/resources/PlaygroundResource.java
```diff
@@ -16,47 +16,68 @@
...
     // the percent signs that exist in the string.
     String formattedPlaygroundFile =
-        playgroundFile.replaceFirst("AUTHENTICATION_TOKEN", token == null ? "" : token);
+        html.replaceFirst("AUTHENTICATION_TOKEN", token == null ? "" : token);
 
     return Response.ok(formattedPlaygroundFile).build();
   }
+
+  private String getToken(HttpHeaders headers) {
+    List<String> tokenHeaders = headers.getRequestHeader("x-cassandra-token");
```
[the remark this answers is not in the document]
olim7t: Right, we can inject `RequestInfo` instead.

apis/sgv2-graphqlapi/src/main/resources/application.yaml
```diff
@@ -0,0 +1,49 @@
...
+    binder:
+      http-server:
+        # ignore all non-application uris, as well as the custom set
+        suppress-non-application-uris: true
+        ignore-patterns: /,/metrics,/playground,/swagger-ui.*,.*\.html
+
+    # exports at prometheus default path
+    export:
+      prometheus:
+        path: /metrics
```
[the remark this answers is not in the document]
olim7t: We didn't have it in v1 either.

The rationale is that for GraphQL, the useful information is in the GraphQL schema metadata itself (for example see the "schema" and "docs" tabs in the playground).

If you look at the REST resources, it's just [generic stuff](https://graphql.org/learn/serving-over-http/) that looks like any other web-based GraphQL service (REST is just the transport). So there's not a lot of value in documenting it, I doubt anyone would ever look at it.

apis/sgv2-graphqlapi/src/test/java/io/stargate/sgv2/graphql/integration/cqlfirst/ApolloIntegrationTest.java
olim7t: I'm not sure how best to explain it, but I think exposing the `CqlIdentifier` is better. That's the thing we want to make available; whether subclasses use the internal form or the CQL form is up to their use case.

apis/sgv2-graphqlapi/src/test/java/io/stargate/sgv2/graphql/integration/util/GraphqlClient.java
olim7t: Done.

apis/sgv2-graphqlapi/README.md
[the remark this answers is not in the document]
ivansenic: Fine with me, we can even move this to a separate task to speed merging this in. It's low priority anyway.

apis/sgv2-graphqlapi/src/main/java/io/stargate/sgv2/graphql/schema/FileSupport.java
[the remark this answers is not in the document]
ivansenic: My bad.

apis/sgv2-graphqlapi/src/main/java/io/stargate/sgv2/graphql/web/resources/PlaygroundResource.java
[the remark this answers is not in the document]
ivansenic: Yea that's cool, so we pass the token no matter how it's resolved (fixed, header or whatever)

apis/sgv2-graphqlapi/src/main/resources/application.yaml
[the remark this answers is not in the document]
ivansenic: OK fine with me. But let's have ticket for updating operator and docker compose stuff, then we can remove this later on..

apis/sgv2-graphqlapi/src/main/resources/application.yaml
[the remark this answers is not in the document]
ivansenic: Yea but you are not making Quarkus have any paths authenticated, everything can be accessed, is this the way to go? This is how it's done in the docs:

```
  # HTTP settings
  http:
    # TODO update both port and non-app paths after discussions
    port: 8180
    non-application-root-path: stargate

    # every /v2 path is authenticated by default
    # adapt if changing the authentication mechanism
    auth:
      proactive: false
      permission:
        default:
          paths: /v2/*
          policy: authenticated
```

This means, at least for the default `AuthConfig`, if the token header is not there, the request to that path would fail immediately, not even reaching the graphql, nor the bridge.. You are effectively saving at least one bridge call and all the processing in the graphql app itself.. 

For example, `/graphql/{keyspaceName}` seems like something that must have a token supplied.. Thus, I would expect all the endpoints that depend on having the token to be in that `paths` list.

apis/sgv2-graphqlapi/src/main/resources/application.yaml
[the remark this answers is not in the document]
ivansenic: Then let's disable it. At least the Swagger UI with the property.

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/common/CqlEnabledIntegrationTestBase.java
ivansenic: Yea I was more thinking about Junit5 extension..

apis/sgv2-graphqlapi/src/test/java/io/stargate/sgv2/graphql/integration/cqlfirst/ApolloIntegrationTest.java
ivansenic: Fine with me

apis/sgv2-graphqlapi/src/main/resources/application.yaml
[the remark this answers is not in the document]
ivansenic: And one more thing here, if you go with the proposal, then there is additional work needed to be done with the metrics.. There is a strange bug in Quarkus that does not correctly report paths in metrics once auth fails: https://github.com/quarkusio/quarkus/issues/24938

In order to overcome this, you need to define `match-patterns` in order to not have variable paths in the metrics like:

```
  # built-in micrometer properties
  micrometer:
    binder:
      http-server:
        # ignore all non-application uris, as well as the custom set
        suppress-non-application-uris: true
        ignore-patterns: /,/metrics,/swagger-ui.*,.*\.html

        # due to the https://github.com/quarkusio/quarkus/issues/24938
        # we need to define uri templating on our own for now
        # note that order is important
        match-patterns: |
          /v2/namespaces/[\\w]+/collections=/v2/namespaces/{namespace:\\w+}/collections,
          /v2/namespaces/[\\w]+/collections/[\\w]+=/v2/namespaces/{namespace:\\w+}/collections/{collection:\\w+},
          /v2/namespaces/[\\w]+/collections/[\\w]+/batch=/v2/namespaces/{namespace:\\w+}/collections/{collection:\\w+}/batch,
          /v2/namespaces/[\\w]+/collections/[\\w]+/json-schema=/v2/namespaces/{namespace:\\w+}/collections/{collection:\\w+}/json-schema,
          /v2/namespaces/[\\w]+/collections/[\\w]+/upgrade=/v2/namespaces/{namespace:\\w+}/collections/{collection:\\w+}/upgrade,
          /v2/namespaces/[\\w]+/collections/[\\w]+/(.*?)/.*/function=/v2/namespaces/{namespace:\\w+}/collections/{collection:\\w+}/{document-id}/{document-path:.*}/function,
          /v2/namespaces/[\\w]+/collections/[\\w]+/(.*?)/.*=/v2/namespaces/{namespace:\\w+}/collections/{collection:\\w+}/{document-id}/{document-path:.*},
          /v2/namespaces/[\\w]+/collections/[\\w]+/.*=/v2/namespaces/{namespace:\\w+}/collections/{collection:\\w+}/{document-id},
          /v2/schemas/namespaces/[\\w]+=/v2/schemas/namespaces/{namespace:\\w+}
```

For you this should be simple, as you only have two endpoints with path parameters:

* `/graphql/{keyspaceName}`
* `/graphql-files/keyspace/{keyspaceName}.graphql`

ivansenic:
Tried to answer to all the comments.. I'll do a final review on Thursday and we can hopefully merge then.

Since all the unit tests are green now, we should enable the int tests:

1. Add required maven setup https://github.com/stargate/stargate/blob/ae0f9fd63d3948065c289c0124eb2f8052c183ea/apis/sgv2-docsapi/pom.xml#L202-L221
2. Enable project in CI: https://github.com/stargate/stargate/blob/ae0f9fd63d3948065c289c0124eb2f8052c183ea/.github/workflows/apis-v2.yaml#L214-L215
3. Enable docker based test: https://github.com/stargate/stargate/blob/ae0f9fd63d3948065c289c0124eb2f8052c183ea/.github/workflows/apis-v2.yaml#L321-L323

apis/sgv2-graphqlapi/src/main/java/io/stargate/sgv2/graphql/web/resources/GraphqlResourceBase.java
```diff
@@ -0,0 +1,283 @@
...
+
+  public static final String APPLICATION_GRAPHQL = "application/graphql";
+
+  private static final Logger LOG = LoggerFactory.getLogger(GraphqlResourceBase.class);
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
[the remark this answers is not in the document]
olim7t: Done.

apis/sgv2-graphqlapi/src/main/java/io/stargate/sgv2/graphql/web/resources/PlaygroundResource.java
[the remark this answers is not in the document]
olim7t: Done.

apis/sgv2-graphqlapi/README.md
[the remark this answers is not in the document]
jeffreyscarpenter: I agree that we should prioritize merging to the `v.2.0.0` branch. That will make it easier to work on the documentation across all 3 of the APIs.

apis/sgv2-graphqlapi/pom.xml
```diff
@@ -10,6 +10,7 @@
   <properties>
     <quarkus.container-image.group>stargateio</quarkus.container-image.group>
     <quarkus.container-image.name>graphqlapi</quarkus.container-image.name>
+    <quarkus.container-image.tag>v${project.version}</quarkus.container-image.tag>
```
olim7t, line 13: Our existing code prepends a "v" before the version, whereas Quarkus doesn't. I don't mind either way but if we drop it we'll have to adapt the coordinator and docker-compose scripts.

docker-compose/cassandra-3.11/docker-compose.yml
```diff
@@ -75,11 +75,10 @@ services:
     networks:
       - stargate
     ports:
-      - 9080:9080
+      - 8085:8080
     mem_limit: 2G
     environment:
       - STARGATE_BRIDGE_HOST=coordinator
       - STARGATE_BRIDGE_PORT=8091
-      - STARGATE_GRAPHQL_PORT=9080
```
olim7t, line 83: We actually don't need to customize the port since the API is in its own container.

docker-compose/cassandra-3.11/docker-compose.yml
```diff
@@ -75,11 +75,10 @@ services:
     networks:
       - stargate
     ports:
-      - 9080:9080
+      - 8085:8080
```
olim7t, line 78: I changed the host port because I happen to have something already running on 9080 on my laptop.
Also it aligns better with what is used in the rest of the file.

apis/sgv2-graphqlapi/src/main/resources/application.yaml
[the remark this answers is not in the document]
ivansenic: @olim7t I figured out that operator will anyway have to change to enable multi-tenancy and setup for the tenant resolver, as well as the data store configuration.. Thus, I think you'll anyway need to migrate the operator.. These can stay for now, but just saying the in operator we can go directly to `-Dquarkus.grpc.clients.bridgde.host` or env equivalent ..

apis/sgv2-graphqlapi/pom.xml
ivansenic: Can you add this to docs as well.. the release process correctly sets the tag with `v`, so this would be only for local builds and keeping that docker-compose stuff working

docker-compose/cassandra-3.11/docker-compose.yml
ivansenic: you can still set the exposed port with `-Dquarkus.http.port` if needed..

docker-compose/cassandra-3.11/docker-compose-dev-mode.yml
```diff
@@ -40,11 +40,10 @@ services:
     networks:
       - stargate
     ports:
-      - 9080:9080
+      - 8085:8080
     mem_limit: 2G
     environment:
       - STARGATE_BRIDGE_HOST=coordinator
       - STARGATE_BRIDGE_PORT=8091
```
ivansenic, line 47: nit: you can also just set `QUARKUS_GRPC_CLIENTS_BRIDGE_HOST` here and remove those env vars from docker compose completely..

ivansenic (approved):
Had one more look, I would say I have no more comments.. Open discussions should be resolved, tests should be green and we can merge this when @olim7t decides to do it.

Great work +1

apis/sgv2-graphqlapi/README.md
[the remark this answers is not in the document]
olim7t: I created #2052.

apis/sgv2-graphqlapi/src/main/java/io/stargate/sgv2/graphql/schema/graphqlfirst/fetchers/deployed/MutationFetcher.java
```diff
@@ -237,8 +237,7 @@ protected QueryParameters buildParameters() {
         && serialConsistencyLevel == DEFAULT_SERIAL_CONSISTENCY) {
       return DEFAULT_PARAMETERS;
     } else {
-      return DEFAULT_PARAMETERS
-          .toBuilder()
+      return DEFAULT_PARAMETERS.toBuilder()
           .setConsistency(consistencyLevel)
           .setSerialConsistency(serialConsistencyLevel)
           .build();
```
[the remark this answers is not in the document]
olim7t: Follow-up ticket: #2053

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/common/CqlEnabledIntegrationTestBase.java
olim7t: Follow-up ticket: #2054

PR #1987, opened 2022-07-19
updated release workflow for V2

ivansenic:
**What this PR does**:

Main goal was to include `apis/` projects in the release by:
 * publishing docker images
 * bumping versions in the PR

However, I ended-up with a complete refactor of a release workflow.. Here is the overview on the jobs and updates made:

1. Create release
    * still creates release in the same way
    * however, `uploadUrl` is shared to other jobs as the output, no need for upload/download stuff
2. Build coordinator
    * still main task to upload `stargate-libs.zip` to the created release
    * now correctly depends on the `create-release` job and it's output
    * updated actions versions to latest (`actions/checkout@v3`, `actions/setup-java@v3`)
    * uses maven wrapper instead of maven
    * extra memory for maven for faster builds `MAVEN_OPTS: '-Xmx4g'`
3. Publish to OSSRH
    * part of the previous job `publish`, but just handling publishing the coordinator jars to the OSSRH
    * updated actions versions, use wrapper, etc
4. Publish docker images
    * part of the previous job `publish`, but just handling publishing the docker images
    * added support for publishing the `/api` as well, done with installing two java versions
    * same updates as everywhere else
5. Create PR for version update
    * part of the previous job `publish`, but just handling new PR for the new version
    * added support for updating the versions of `/api` as well
    * same updates as everywhere else

In addition I defined a manual trigger with the possibility to skip the publishing of the artifacts and docker images.

Internal issue.

**Checklist**
- [x] Manual trigger inputs with no push

### DIFF
--- a/.github/workflows/release-v2.yml
+++ b/.github/workflows/release-v2.yml
@@ -17,12 +17,16 @@ on:
         required: true
         type: string
 
+# global env vars, available in all jobs and steps
 env:
+  ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_USERNAME }}
+  ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_PASSWORD }}
   MAVEN_OPTS: '-Xmx4g'
 
 jobs:
 
   # creates a new release if it's not existing
+  # outputs the resolved release tag value in the release-tag output var
   # outputs the upload URL in the release-upload-url output var
   create-release:
     name: Create release
@@ -65,9 +69,6 @@ jobs:
           cache: maven
 
       - name: Setup Maven
-        env:
-          ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_USERNAME }}
-          ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_PASSWORD }}
         run: |
           mkdir -p ~/.m2
           cat <<EOF > ~/.m2/settings.xml
@@ -123,8 +124,6 @@ jobs:
 
       - name: Setup Maven
         env:
-          ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_USERNAME }}
-          ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_PASSWORD }}
           OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}
           OSSRH_TOKEN: ${{ secrets.OSSRH_TOKEN }}
         run: |
@@ -188,9 +187,6 @@ jobs:
       - run: echo "JAVA_17=$JAVA_HOME" >> $GITHUB_ENV
 
       - name: Setup Maven
-        env:
-          ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_USERNAME }}
-          ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_PASSWORD }}
         run: |
           mkdir -p ~/.m2
           cat <<EOF > ~/.m2/settings.xml
@@ -301,9 +297,6 @@ jobs:
           cache: maven
 
       - name: Setup Maven
-        env:
-          ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_USERNAME }}
-          ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_PASSWORD }}
         run: |
           mkdir -p ~/.m2
           cat <<EOF > ~/.m2/settings.xml
--- a/.github/workflows/release-v2.yml
+++ b/.github/workflows/release-v2.yml
@@ -8,10 +8,17 @@ on:
   workflow_dispatch:
     inputs:
       publish:
-        description: 'Should the publishing of docker images and JARs to OSSRH happen?'
+        description: 'Publish docker images and JARs to OSSRH?'
         required: true
         default: false
         type: boolean
+      tag:
+        description: 'Custom release tag value.'
+        required: true
+        type: string
+
+env:
+  MAVEN_OPTS: '-Xmx4g'
 
 jobs:
 
@@ -21,16 +28,24 @@ jobs:
     name: Create release
     runs-on: ubuntu-latest
     outputs:
+      release-tag: ${{ steps.resolve_tag.outputs.tag }}
       release-upload-url: ${{ steps.create_release.outputs.upload_url }}
     steps:
-      - name: Create Release
+      - name: Resolve tag
+        id: resolve_tag
+        run: |
+          TAG=${{ inputs.tag != null && inputs.tag || github.ref }}
+          echo "Resolved tag for the release $TAG"
+          echo "::set-output name=tag::${TAG}"
+
+      - name: Create release
         id: create_release
         uses: actions/create-release@v1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          tag_name: ${{ github.ref }}
-          release_name: Release ${{ github.ref }}
+          tag_name: ${{ steps.resolve_tag.outputs.tag }}
+          release_name: Release ${{ steps.resolve_tag.outputs.tag }}
           draft: false
           prerelease: true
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release-v2.yml
```diff
@@ -6,9 +6,14 @@ on:
       - 'v2.*.*'
 
 jobs:
+
+  # creates a new release if it's not existing
+  # outputs the upload URL in the releaseUploadUrl output var
```
ivansenic, line 11: wrong name of the var

ivansenic:
Should this be on the `master`, so that I can actually test stuff?

.github/workflows/release-v2.yml
```diff
@@ -128,45 +149,190 @@ jobs:
...
-      - name: Build and push (ECR)
+
+      - name: Build and push (coordinator, Amazon ECR)
+        if: ${{ inputs.publish != false }}
         run: |
           ./build_docker_images.sh -p -t ${{ steps.vars.outputs.tag }} -r ${{ secrets.ECR_REPOSITORY }} -a
-      - name: Update version numbers for next release
+
+      - name: Build and push (apis, Amazon ECR)
+        if: ${{ inputs.publish != false }}
```
mpenick, line 276: `${{ inputs.publish == true }}` seems easier to understand compared to `${{ inputs.publish != false }}`. ` `${{ inputs.publish }}` not work, or does it need `!=` (or `==`) for coercion?

mpenick: ah, nm, I understand, this work well for testing. :) Because absence is true.

ivansenic: But how would this work when it's not a manual trigger and there are no inputs?

.github/workflows/release-v2.yml
```diff
@@ -63,48 +75,57 @@ jobs:
...
+          java-version: '8'
+          cache: maven
+
       - name: Setup Maven
         env:
           ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_USERNAME }}
           ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_PASSWORD }}
           OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}
           OSSRH_TOKEN: ${{ secrets.OSSRH_TOKEN }}
+          MAVEN_OPTS: '-Xmx4g'
```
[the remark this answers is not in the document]
ivansenic: GH runner has 7GB of memory available, in my tests I figured-out that coordinator builds are like 2-3 times faster if I give more memory to the maven.. Seems that OSGi manifest creation is some kind of bottleneck there..

.github/workflows/release-v2.yml
```diff
@@ -128,45 +149,190 @@ jobs:
...
+          JAVA_HOME=$JAVA_8 ./mvnw versions:set -DremoveSnapshot versions:commit
+          JAVA_HOME=$JAVA_8 ./mvnw -B clean install -P dse -DskipTests 
+
+      # only set version here
+      - name: Install APIs
         run: |
-          mvn versions:set -DremoveSnapshot versions:commit && \
-          mvn -B -Dgpg.passphrase=${{ secrets.OSSRH_GPG_SECRET_KEY_PASSWORD }} clean deploy -DskipTests -P deploy,dse
+          cd apis/
+          JAVA_HOME=$JAVA_17 ./mvnw versions:set -DremoveSnapshot versions:commit
```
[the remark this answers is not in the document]
ivansenic: Yea basically I don't need it as I have afterwards:

```
JAVA_HOME=$JAVA_17 ./mvnw clean package -DskipTests -Dquarkus.container-image.build=true -Dquarkus.container-image.push=true -Dquarkus.container-image.tag=${{ steps.vars.outputs.tag }}
```

Which does clean package, build of docker and push in one command..

.github/workflows/release-v2.yml
mpenick: I didn't see your response. I understand, the absence is true to allow for testing.

.github/workflows/release-v2.yml
```diff
@@ -20,31 +33,30 @@ jobs:
...
           ref: "v2.0.0"
-      - name: Setup Java JDK
-        uses: actions/setup-java@v1
+
+      - name: Set up JDK 8
+        uses: actions/setup-java@v3
         with:
-          java-version: 1.8
+          distribution: 'temurin'
+          java-version: '8'
```
tatu-at-datastax, line 52: We can use either `8` for latest, or specify specific version like `8.0.332`. No strong opinion on how much to pin, although I did have issues with `ccm` failing on specific new version a while ago. But I guess we can pin to specific version if (and only if) JDK-related problems found.

.github/workflows/release-v2.yml
```diff
@@ -63,48 +75,57 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+
+      # TODO same here, why do we need the ref?
+      - uses: actions/checkout@v3
         with:
           ref: "v2.0.0"
+
+      # TODO this below is some magic I don't get, why not using ${{ github.ref }} as in create-release
```
tatu-at-datastax, line 110: Would it be easy to test to see if that'd work? The whole "echo" construct is needed for some reason by Github actions to set output variable for other steps to see I think. Sort of idiom although quite incomprehensible to me.

.github/workflows/release-v2.yml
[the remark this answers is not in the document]
tatu-at-datastax: Interesting that manifest creation is so slow. But I guess it depends on CLASSPATH size, and we have so many dependencies...

.github/workflows/release-v2.yml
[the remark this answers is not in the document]
tatu-at-datastax: So without specifying new version, does this just increase the patch number by one?

.github/workflows/release-v2.yml
```diff
@@ -128,45 +149,190 @@ jobs:
...
+      # install both Java 8 & Java 17, keep paths in vars
+      - name: Set up JDK 8
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: '8'
+          cache: maven
+      - run: echo "JAVA_8=$JAVA_HOME" >> $GITHUB_ENV
+
+      - name: Set up JDK 8
```
tatu-at-datastax, line 189: Minor typo, should be 17.

.github/workflows/release-v2.yml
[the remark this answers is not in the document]
ivansenic: Not even sure if it's needed to remove the `-SNAPSHOT` here, I am anyway tagging with the custom tag value?

.github/workflows/release-v2.yml
```diff
@@ -128,45 +149,190 @@ jobs:
...
+          ./mvnw versions:set -DremoveSnapshot versions:commit
+          ./mvnw -B -Dgpg.passphrase=${{ secrets.OSSRH_GPG_SECRET_KEY_PASSWORD }} clean deploy -DskipTests -P deploy,dse
+
+  # publishes the docker images for the coordinator and the APIs
+  publish-docker:
+    name: Publish docker images
+    runs-on: ubuntu-latest
+    steps:
+
+      # TODO same here, why do we need the ref?
```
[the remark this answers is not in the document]
jeffreyscarpenter: > Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. 
source: https://github.com/actions/checkout

Tatu is right, this is brittle. Let's omit these `with` overrides.

jeffreyscarpenter (approved):
Lots of great improvements here, many of them would apply to the v1 release workflow as well.

.github/workflows/release-v2.yml
ivansenic: Done.

.github/workflows/release-v2.yml
ivansenic: Done.

.github/workflows/release-v2.yml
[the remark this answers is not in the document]
ivansenic: Actually I realized this is major bug at the moment. Consider a following situation:

* I tag locally a commit that should be released
* Anybody merges a PR that pushes another commit on a master
* I now push that local tag to the remote
* The release is not checking out the commit that is tagged by the release, but the one from the PR that's being merged..

:exploding_head:

ivansenic:
Added few further improvements:

* when `input.publish` is `false` I am still building the images, but not pushing them
* checkout the reference, as suggested by @tatu-at-datastax and @jeffreyscarpenter 
* added debug output for the tag resolving as it seems not needed

ivansenic:
Pushed additional updates:
* fix `MAVEN_OPTS` usage..
* introduce `inputs.tag` for specification of the release tag value when running from a manual trigger
* removed strange tag resolving in the publish jobs

ivansenic:
`default-push-trigger-1 (stargateio) ` failing due to the #2001. Merging anyway.

PR #1988, opened 2022-07-19
Use dedicated test-common module instead of generating a test-jar (fixes #1976)

### LINKED ISSUE #1976: Consider alternative setup for sgv2-quarkus-common test JAR
In #1965 we generate a Maven "test-jar" from sgv2-quarkus-common. Test JARs generally don't work well in reactor builds (e.g. you need to `mvn install` first before running the tests of `sgv2-docsapi`).

Maybe we should have a separate `sgv2-quarkus-test-common` that is a regular module:
```
sgv2-quarkus-common <--- sgv2-quarkus-test-common <--- sgv2-docsapi
```
This also means moving some tests out of `sgv2-quarkus-common` to avoid a circular dependency, but we can put them in `sgv2-quarkus-test-common`.

olim7t:
**What this PR does**:
see above

Maven test-jars don't work well in reactor builds (need a `mvn install` before running tests in a dependent project).

This change allows a full `mvn test` from the `apis/` root, even if nothing has been installed in the local repository beforehand.

The only downside is that two unit tests (`BridgeDataStorePropertiesConfigurationTest` and `SchemaManagerTest`) need to move to the new module, even though they logically belong to the original `sgv2-quarkus-common`. This is to avoid circular dependencies.

Fixes #1976


apis/sgv2-quarkus-test-common/src/test/resources/application.yaml
```diff
@@ -0,0 +1,12 @@
+# by default ignore bridge for data store props in tests
+stargate:
+  data-store:
+    ignore-bridge: true
```
tatu-at-datastax, line 4: What does "ignore-bridge" do? Instead of calling bridge during tests it'll ... ?

olim7t: It controls whether we use the bridge's `GetSupportedFeatures` operation to find out if the persistence backend supports SAI, logged batches, etc. If false, this information must be provided via the configuration instead.

I couldn't find a reference in the Java code at first because it uses a different case: `DataStoreConfig.ignoreBridge()` 😡 Something to keep in mind when you try to find where config properties are used.

Then we use it in `DataStorePropertiesConfiguration`.

tatu-at-datastax: Thanks, that makes sense. Name is bit vague but comment can in config help, and I am not sure what would be a better name ("access-metadata-via-bridge" is sort of long f.ex)

ivansenic: I am fully up for name change of the config prop if you have a better idea?

ivansenic (changes requested):
Wait, I need to pull a break here.. I hopped that this would give us some improvements, but it does not:

1. executing the `./mvnw test` in the `/apis` already works fine on the master if nothing is previously installed, just tested that
2. starting the project with `cd sgv2-docsapi/ && ../mvnw quarkus:dev` still does not work, even with the changes.. I hoped this would fix that..

I actually see no improvements from the previous state, except everything is even more complex with one additional project and even that test classes not being anymore in the right project.. I am not seeing something?

olim7t:
> I actually see no improvements from the previous state

You know what, you're right. The issue I had in mind is an old Maven 2 bug that's been fixed forever. I'll close this PR.

> starting the project with `cd sgv2-docsapi/ && ../mvnw quarkus:dev` still does not work

That will never work. But you can do this from the project root:
```
./mvnw -pl sgv2-docsapi -am test
```
It builds the module and its dependencies.

PR #1990, opened 2022-07-19
V1/V2 Docs API compat test

EricBorczuk:
**What this PR does**:
Adds a small script that uses Docs API V1 to write documents and ensures that they can be retrieved and modified by V2.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/testing/src/main/java/io/stargate/it/http/docsapi/compatibility/DocumentsV1ToV2Compat.java
+++ b/testing/src/main/java/io/stargate/it/http/docsapi/compatibility/DocumentsV1ToV2Compat.java
@@ -1,0 +1,110 @@
+package io.stargate.it.http.docsapi.compatibility;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.stargate.it.http.RestUtils;
+import java.io.IOException;
+
+public class DocumentsV1ToV2Compat {
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  private static String authToken;
+  private static String host = "127.0.0.1";
+  private static String namespace = "test_namespace";
+  private static String collection = "test_collection";
+  private static String documentId = "test_document";
+
+  /*
+   * This will:
+   * 1) Create a document using the Documents API V1
+   * 2) Get that document using Documents API V2
+   * 3) Update the document using Documents API V2 and check that it was updated
+   * 4) Patch a sub-document using Documents API V2 and check that it was updated
+   * 5) Delete the document using Documents API V2 and check that it no longer exists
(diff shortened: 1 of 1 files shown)


jeffreyscarpenter (commented):
This is an interesting check but won't we end up deleting it soon, since we'll soon be disabling Docs API in coordinator?

EricBorczuk:
It's really just meant as a one-off to make sure that right now, what we have in V2 is correct/compatible; since it's not integrated into CI or anything, it won't interfere with any code changes and also it is meant to be deleted once we remove the V1 docs API

ivansenic (changes requested):
I like the approach very much, why not have this as code :)

The only thing I need is configurable hosts and ports for both V1 and V2.. On my setup the V1 would be `127.0.0.4:8083` and V2 `127.0.0.1:8180`.. You can make that configurable by system properties or main method arguments, both are fine with me.. Defaults can stay as is now..

Once you update I ll run the test..

EricBorczuk:
OK - should be able to configure using system properties now.

When you test locally, you'll have to create a keyspace manually called `test_namespace` before you run

testing/src/main/java/io/stargate/it/http/docsapi/compatibility/DocumentsV1ToV2Compat.java
```diff
@@ -0,0 +1,117 @@
...
+   * 4) Patch a sub-document using Documents API V2 and check that it was updated
+   * 5) Delete the document using Documents API V2 and check that it no longer exists
+   * 6) Get the document using Documents API V1 and check that it no longer exists
+   */
+  public static void main(String[] args) throws IOException {
+    v1Host = System.getProperty("doccapi.v1.host", "127.0.0.1");
+    v1Port = System.getProperty("doccapi.v1.port", "8083");
+    v2Host = System.getProperty("doccapi.v2.host", "127.0.0.1");
+    v2Port = System.getProperty("doccapi.v2.port", "8180");
+    authToken = RestUtils.getAuthToken(v2Host);
```
ivansenic, line 34: does not work, should be `v1Host`..

testing/src/main/java/io/stargate/it/http/docsapi/compatibility/DocumentsV1ToV2Compat.java
```diff
@@ -0,0 +1,117 @@
...
+import java.io.IOException;
+
+public class DocumentsV1ToV2Compat {
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  private static String authToken;
+  private static String v2Host;
+  private static String v2Port;
+  private static String v1Host;
+  private static String v1Port;
+  private static String namespace = "test_namespace";
```
ivansenic, line 16: could be parametrized as well..

ivansenic (approved):

PR #1992, opened 2022-07-20
Fix TLS (`client_encryption_options`) configuration for CQL

### LINKED ISSUE #1001: Cql client encryption is not working

mpenick:
**What this PR does**:

CQL TLS configuration is now loaded from a yaml  file specified using the system property `stargate.cql.config_path`. This very similar to `cassandra.yaml`, but it only contains the options that are specific to the CQL transport layer.

This also include a couple bug fixes:
* Removes usage of `ClientWarn`, this is handled by the persistence layer.
* Remove usage of `DatabaseDescriptor` which is not properly configured.

Substantial portions of the following files have been copied from C* 4.0.3:
* Config.java
* EncryptionOptions.java
* YamlConfigurationLoader.java
* SSLFactory.java

Fixes #1001

**Checklist**
- [x] Changes manually tested
- [x] Add an example `cql.yaml`  file to resources
- [x] Add tests for client certificates
- [x] Fix licences on `.java` files (either Stargate or Apache Cassandra)


tatu-at-datastax:
One thing that may or may not be relevant: instead of using SnakeYAML directly (btw do we not need a dependency from `pom.xml` to it? Or is it assumed we get it transitively from Cassandra), there's also Jackson YAML module:

https://github.com/FasterXML/jackson-dataformats-text/tree/2.14/yaml

which let's you use Jackson API to bind to POJOs, or `Map`s or whatever using YAML-backed `ObjectMapper`.
That could simplify things a bit. It does use SnakeYAML for actual decoding fwtw.

cql/src/main/java/io/stargate/cql/impl/CqlImpl.java
```diff
@@ -85,7 +85,6 @@ public void start() {
             .withEventLoopGroup(workerGroup)
             .withHost(nativeAddr);
 
-    TransportDescriptor.getNativeProtocolEncryptionOptions().applyConfig();
     if (!TransportDescriptor.getNativeProtocolEncryptionOptions().isEnabled()) {
```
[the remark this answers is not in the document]
mpenick: Thanks that's a good catch. Thanks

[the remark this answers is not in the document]
mpenick: Fixed.

mpenick:
> One thing that may or may not be relevant: instead of using SnakeYAML directly (btw do we not need a dependency from `pom.xml` to it? Or is it assumed we get it transitively from Cassandra), there's also Jackson YAML module:
> 
> https://github.com/FasterXML/jackson-dataformats-text/tree/2.14/yaml
> 
> which let's you use Jackson API to bind to POJOs, or `Map`s or whatever using YAML-backed `ObjectMapper`. That could simplify things a bit. It does use SnakeYAML for actual decoding fwtw.

Updated in this commit: https://github.com/stargate/stargate/pull/1992/commits/de500843beea07c377eaa381a46a6af01279d004

cql/src/main/java/io/stargate/cql/CqlActivator.java
```diff
@@ -15,21 +15,30 @@
...
-import java.net.UnknownHostException;
 import java.util.Arrays;
 import java.util.List;
 import javax.annotation.Nullable;
-import org.apache.cassandra.config.Config;
+import org.apache.cassandra.stargate.config.Config;
 
 public class CqlActivator extends BaseActivator {
+  public static ObjectMapper mapper =
+      new ObjectMapper(new YAMLFactory())
```
tatu-at-datastax, line 39: not a big deal but there's also `YAMLMapper` that constructs `YAMLFactory` (no need to change if no other changes made), which is a sub-class of `ObjectMapper`

cql/src/main/resources/cql.yaml
```diff
@@ -0,0 +1,126 @@
+
+# CQL transport config YAML
```
olim7t, line 2: It looks like this file is not referenced anywhere. Is it intended as an example?

olim7t (approved):

cql/src/main/resources/cql.yaml
mpenick: Yes, it's an an example file.

PR #1993, opened 2022-07-21
improving release v1 github workflow

ivansenic:
**What this PR does**:

Ported a subset of Release V2 improvements to the V1 as well.. V1 workflow is considerably simpler:

* introduced the manual trigger with the input that can disable pushing of the jars
* sharing upload url via the outputs
* fixed potential bug with referencing master branch in the release
* updated action version, improved maven setup, etc
* proper naming and deletion of unnecessary stuff

**Unfortunately testing a manual trigger is only possible when this is merged..**

Internal issue

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -100,6 +100,12 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      # TODO remove after testing
+      - name: Warn publish
+        if: ${{ inputs.publish != false }}
+        run: |
+          echo "ATTENTION: This job will publish JARs to the OOSRH!"
+
       # TODO resolve if this is needed at all
       - name: Set version
         id: vars
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,10 +5,23 @@ on:
     tags:
       - 'v1.*.*'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -63,48 +72,53 @@ jobs:
...
-        run: echo ::set-output name=URL::$(cat release-url/url.txt)
+          
+      # uploads the jars by referencing the release-upload-url from create-release job
       - name: Upload jars
         uses: actions/upload-release-asset@v1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          upload_url: ${{ steps.get_release_url.outputs.URL }}
+          upload_url: ${{needs.create-release.outputs.release-upload-url}}
```
mpenick, line 91: Should this be `steps.create-release...` instead of `needs.create-release...`?

mpenick (approved):

.github/workflows/release.yml
ivansenic: This uses the referenced job in the `needs` section, and then uses it's outputs: https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs

PR #1995, opened 2022-07-21
Move integration test utilities to sgv2-quarkus-common

olim7t:


apis/sgv2-quarkus-common/pom.xml
```diff
@@ -91,6 +91,11 @@
       <artifactId>rest-assured</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <scope>test</scope>
+    </dependency>
```
olim7t, line 98: We need to add this dependency here, but it also has to stay in `sgv2-docsapi`. That's because the test-jar shares the same POM as the main artifact, and test dependencies are not transitive.

tatu-at-datastax (approved):

mpenick (approved):

PR #2019, opened 2022-08-09
Separate executor in BridgeImpl

EricBorczuk:
This is a prospective change that Ivan and I observed improves performance of both reads and writes for docs API v2...I wanted to make this draft to ask what the prior reasoning on using the direct executor was. I think maybe @mpenick or @olim7t would know about that context!

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/bridge/src/main/java/io/stargate/bridge/impl/BridgeImpl.java
+++ b/bridge/src/main/java/io/stargate/bridge/impl/BridgeImpl.java
@@ -71,9 +71,7 @@ public class BridgeImpl {
             EXECUTOR_SIZE, GrpcUtil.getThreadFactory("bridge-stargate-executor", true));
     server =
         NettyServerBuilder.forAddress(new InetSocketAddress(listenAddress, port))
-            // `Persistence` operations are done asynchronously so there isn't a need for a separate
-            // thread pool for handling gRPC callbacks in `GrpcService`.
-            .directExecutor()
+            .executor(Executors.newScheduledThreadPool(16))
             .intercept(new NewConnectionInterceptor(persistence, authenticationService))
             .intercept(new MetricCollectingServerInterceptor(metrics.getMeterRegistry()))
             .addService(new BridgeService(persistence, authorizationService, executor))
--- a/bridge/src/main/java/io/stargate/bridge/impl/BridgeImpl.java
+++ b/bridge/src/main/java/io/stargate/bridge/impl/BridgeImpl.java
@@ -71,7 +71,9 @@ public class BridgeImpl {
             EXECUTOR_SIZE, GrpcUtil.getThreadFactory("bridge-stargate-executor", true));
     server =
         NettyServerBuilder.forAddress(new InetSocketAddress(listenAddress, port))
-            .executor(Executors.newScheduledThreadPool(16))
+            // `Persistence` operations are done asynchronously so there isn't a need for a separate
+            // thread pool for handling gRPC callbacks in `GrpcService`.
+            .directExecutor()
             .intercept(new NewConnectionInterceptor(persistence, authenticationService))
             .intercept(new MetricCollectingServerInterceptor(metrics.getMeterRegistry()))
             .addService(new BridgeService(persistence, authorizationService, executor))
--- a/bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
+++ b/bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
@@ -98,22 +98,29 @@ public class BridgeService extends StargateBridgeGrpc.StargateBridgeImplBase {
     Keyspace keyspace = persistence.schema().keyspace(decoratedName);
 
     if (keyspace == null) {
-      responseObserver.onNext(
-          Schema.QueryWithSchemaResponse.newBuilder()
-              .setNoKeyspace(Schema.QueryWithSchemaResponse.NoKeyspace.getDefaultInstance())
-              .build());
-      responseObserver.onCompleted();
+      executor.execute(
+          () -> {
+            responseObserver.onNext(
+                Schema.QueryWithSchemaResponse.newBuilder()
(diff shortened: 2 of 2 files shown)


mpenick:
We saw performance improvements in both throughput and latency in fallout (maybe Haxx?) tests from using `.directExecutor()` on the original gRPC API because operations in the gRPC service are already transitioned to another thread; either a DSE TPC or a C* thread is used and that makes the operations asynchronous from the Netty I/O thread. By not using `.directExecutor()`, the request is transitioned to gRPC threadpool to then be transitioned to another DSE/C* threads. The idea was that it removed one extra unnecessary context switch.  Also, very little CPU intensive work is done in the original gRPC service to cause performance issues on the Netty I/O threads. Maybe for the docapi workload that's not true? Or perhaps the bridge code might be doing something different or there's some performance regression?

mpenick:
It'd be interesting to narrow which specific gRPC bridge request are slower using `.directExecutor()` or see if it's maybe all of them.

bridge/src/main/java/io/stargate/bridge/impl/BridgeImpl.java
```diff
@@ -71,9 +71,7 @@ public BridgeImpl(
             EXECUTOR_SIZE, GrpcUtil.getThreadFactory("bridge-stargate-executor", true));
     server =
         NettyServerBuilder.forAddress(new InetSocketAddress(listenAddress, port))
-            // `Persistence` operations are done asynchronously so there isn't a need for a separate
-            // thread pool for handling gRPC callbacks in `GrpcService`.
-            .directExecutor()
+            .executor(Executors.newScheduledThreadPool(16))
```
tatu-at-datastax, line 74: I don't know the context here, but just in general 16 sounds like a relatively small thread pool?
Not sure what would be a good way to figure out an optimal setting; would be good to make it configurable I think, with a new `stargate.rpc.xxx` property.

ivansenic: We need to understand first why is better performance observed and if we are not mistaken.. Why was a direct executor chosen at the first place..

We also need to have a proper way of getting the default value for the pool size, should it be based on the number of cores or something like this.. Plus, we should have it configurable..

This is very sensitive change for me and I would like to avoid doing any damage long term..

mpenick: > Why was a direct executor chosen at the first place..

ivansenic: @mpenick OK, but what about the requests that are not jumping to another thread? For example, `DescribeKeyspace` call is not jumping to another thread right? Can this be the reason? Maybe we need to offload those call also to another thread?

mpenick: Just to track the discussion we had @ivansenic. Ivan and I talked about making another, explicitly created, thread pool for blocking requests like `DescribeKeyspace` and still continuing to use `.directExecutor()`. That way the already async requests like "execute" type request don't have to move through yet another thread.

ivansenic (changes requested):
Just blocking until @EricBorczuk pushes a new version.

bridge/src/main/java/io/stargate/bridge/impl/BridgeImpl.java
EricBorczuk: I will implement that here, cool

EricBorczuk:
Is it correct to use the executor for the bridge service during describe keyspace? That's what I ended up doing.

bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
```diff
@@ -148,7 +148,7 @@ public void executeBatch(Batch batch, StreamObserver<Response> responseObserver)
   public void describeKeyspace(
       Schema.DescribeKeyspaceQuery request,
       StreamObserver<Schema.CqlKeyspaceDescribe> responseObserver) {
-    SchemaHandler.describeKeyspace(request, persistence, responseObserver);
+    executor.execute(() -> SchemaHandler.describeKeyspace(request, persistence, responseObserver));
```
ivansenic, line 151: That looks good, but what about the `executeQueryWithSchema` method? Should suffer from same issues if the query is not executed?

ivansenic (changes requested):

bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
EricBorczuk: `executeQueryWithSchema` calls `executeQuery`, which creates a QueryHandler that uses the `executor` for `SchemaAgreementHelper` - is there something else that has to be scheduled on the executor? I thought we had said that queries themselves should remain on direct execution?

ivansenic: But only in that `else` case, two other ifs are still processed in the same thread? execute query is not call always:

https://github.com/stargate/stargate/blob/d724cc5ed85f03c6f05b9606e192655495c6e9d2/bridge/src/main/java/io/stargate/bridge/service/BridgeService.java#L91-L138

ivansenic: That said I am not sure what to actually do here, cause we don't want to offload query execution with two executor task.. Maybe we can leave as is, but that's not good imo as well..

EricBorczuk: Well, if I took the body of those two `if` clauses and used the executor to execute them instead, that would work (and wouldn't spawn tasks within tasks or anything) - I will try that

ivansenic (approved):
Yea let's go with this now..

PR #2020, opened 2022-08-10
id-path now accepts a path to a non-string value

EricBorczuk:
**What this PR does**:
Allow a user to provide an id-path that points to a non-string value.

Example:
[
   {"_id": 1},
   {"_id": "foo"},
]

batch writing the above with `id-path=_id` will now be ALLOWED, and the two documents written will have keys `"1"` and `"foo"`.

Fixes https://github.com/stargate/stargate-mongoose/issues/19

**Checklist**
- [x] Changes manually tested


EricBorczuk:
This will have to be ported to V2 as well, doing that next (should not take long cya in a few)

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -1251,16 +1251,16 @@ private BiFunction<Optional<JsonPointer>, JsonNode, String> documentIdResolver()
             .map(
                 p -> {
                   JsonNode node = jsonNode.at(p);
-                  if (!node.isTextual()) {
+                  if (!node.isValueNode() || node.isMissingNode()) {
```
mpenick, line 1254: `|| node.isMissingNode()` might not be needed because `.isValueNode()` include that check already.

```java
    public final boolean isValueNode()
    {
        switch (getNodeType()) {
            case ARRAY: case OBJECT: case MISSING:
```

EricBorczuk: I wasn't sure - thanks!

mpenick: Is `null` a valid scalar for this case? In other words should a `null` scalar be an error or is that okay?

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
```diff
@@ -1251,16 +1251,16 @@ private BiFunction<Optional<JsonPointer>, JsonNode, String> documentIdResolver()
...
                     String format =
                         String.format(
-                            "JSON document %s requires a String value at the path %s in order to resolve document ID, found %s. Batch write failed.",
+                            "JSON document %s requires a scalar value at the path %s in order to resolve document ID, found %s. Batch write failed.",
                             jsonNode, p, nodeDes);
                     throw new ErrorCodeRuntimeException(
                         ErrorCode.DOCS_API_WRITE_BATCH_INVALID_ID_PATH, format);
                   }
-                  return node.textValue();
+                  return node.asText();
```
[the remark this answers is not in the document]
mpenick: Is it going to be a problem that `"1"` and `1`, `"false"` and `false` are going to map to the same value or do these need to be differentiated?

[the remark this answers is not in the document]
ivansenic: Batch write checks for duplicate IDs, so I guess that's fine.. If the document with the ID already exists in the Db. then it's overwritten. No issues imo..

restapi/src/main/java/io/stargate/web/docsapi/service/ReactiveDocumentService.java
ivansenic: Let's avoid `null` that makes no sense..

ivansenic (approved):
* We need to add one int test where you would batch write with int, long, boolean, etc..
* Don't allow nulls as @mpenick suggested (although I am not sure if `isValueNode()` is doing null check as well)

mpenick:
Per response above shouldn't we add `if (!node.isValueNode() || node.isNull()` to the error check?

ivansenic:
> Per response above shouldn't we add `if (!node.isValueNode() || node.isNull()` to the error check?

True. cc @EricBorczuk

mpenick (approved):
Nice work.

PR #2021, opened 2022-08-10
V2 fix: id-path now accepts a path to a non-string value

EricBorczuk:
**What this PR does**:
Exactly analogous fix for https://github.com/stargate/stargate-mongoose/issues/19, see https://github.com/stargate/stargate/pull/2020


ivansenic (approved):
* Same here, let's add an INT test with batch write that has int, long, boolean as target IDs..
* Maybe better naming of the PR, so we get meaningful commit msg in the tree

mpenick (approved):
:+1:

Issue #2022, opened 2022-08-10
Update DSE to 6.8.25


tatu-at-datastax:
It looks like we need to upgrade DSE persistence from 6.8.24 to 6.8.25 to resolve a Snyk reported CVE:

https://security.snyk.io/vuln/SNYK-JAVA-ORGCODEHAUSJACKSON-534878

and get some assorted other bug fixes, as per:

https://github.com/datastax/release-notes/blob/master/DSE_6.8_Release_Notes.md

PR #2024, opened 2022-08-11
Update Stargate-builder image to use DSE 6.8.25

tatu-at-datastax:
**What this PR does**:

Follow up to #2022 needed to make CI use DSE 6.8.25 as well.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/ci/Makefile
+++ b/ci/Makefile
@@ -1,7 +1,7 @@
 HOSTNAME        := gcr.io
 PROJECT         := stargateio
 NAME            := stargate-builder
-VERSION         := v1.0.9
+VERSION         := v1.0.10
 TAG             := ${VERSION}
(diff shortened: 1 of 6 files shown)


jeffreyscarpenter (commented):
Doesn't the Dockerfile need to change as well?

https://github.com/stargate/stargate/blob/master/ci/Dockerfile#L57

tatu-at-datastax:
> Doesn't the Dockerfile need to change as well?
> 
> https://github.com/stargate/stargate/blob/master/ci/Dockerfile#L57

Yes, but that was done in the first PR when updating references, so it already has 6.8.25.

As Ivan suggested in future it'd make sense to combine both PRs into just one.

mpenick (approved):

PR #2026, opened 2022-08-12
fix release workflows input defaults

ivansenic:
**What this PR does**:

Seems that there is a bug with github inputs and default values, as when release is trigger with the tag push, the inputs still have the default values.. 

I was following the documentation where it states that context is only available in reusable workflows:

image

I am not sure if this is the solution or if we should explicitly set default to `true`.. I am a bit afraid that this will not work as well, and that bug could be is that as `publish` is of type boolean it gets default false when triggered from the push..

In addition, there was a bug in the publish itself, that made the re-running the release fail, see: https://github.com/stargate/stargate/actions/runs/2845140244

Internal issue.

### DIFF
--- a/.github/workflows/release-v2.yml
+++ b/.github/workflows/release-v2.yml
@@ -25,15 +25,13 @@ env:
 
 jobs:
 
-  # creates a new release if it's not existing
+  # resolves tag value
   # outputs the resolved release tag value in the release-tag output var
-  # outputs the upload URL in the release-upload-url output var
-  create-release:
-    name: Create release
+  resolve-tag:
+    name: Resolve tag
     runs-on: ubuntu-latest
     outputs:
       release-tag: ${{ steps.resolve_tag.outputs.tag }}
-      release-upload-url: ${{ steps.create_release.outputs.upload_url }}
     steps:
       - name: Resolve tag
         id: resolve_tag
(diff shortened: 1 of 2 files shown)


ivansenic:
cc @tatu-at-datastax

ivansenic:
OK this makes no sense, seems that `inputs` should not even be defined: https://github.com/orgs/community/discussions/26322

ivansenic:
Started a GitHub discussion about this: https://github.com/community/community/discussions/29242

ivansenic:
I also realized that re-trying v2 release, would also fail now as tag resolving is connected to the creation of the release.. In re-try the create release would fail as release already exists.. So we could separate those two steps.. For now the workaround would be to delete the created release before re-try, so at least option exists.

mpenick:
It looks like there's a follow up suggestion from your community discussion post: https://github.com/orgs/community/discussions/29242#discussioncomment-3383438

.github/workflows/release-v2.yml
```diff
@@ -10,7 +10,6 @@ on:
       publish:
         description: 'Publish docker images and JARs to OSSRH?'
         required: true
-        default: false
```
tatu-at-datastax, line 13: Would this change anything? Is the idea that default value from workflow_dispatch would somehow have "leaked" to on-push triggering? And that now default would be `null`.
From discussions it seemed more likely that instead it's just that for the most part `null` and `false` behave the same and the default on-push uses might just be that of "falsy" (undefined/null, but something that coerces to `false` in boolean context).

In a way, reversing the logic could solve this (as I mentioned in discussion). But alternatively, maybe on-push trigger could pass explicit `true` for this value -- if `inputs` entries can be passed by triggers other than "workflow_dispatch".

ivansenic:
@tatu-at-datastax @mpenick OK fixed as suggested by Tatu on the discussion, can I get an approval so I can merge & release..

tatu-at-datastax (approved):

tatu-at-datastax:
Looks good @ivansenic

PR #2027, opened 2022-08-12
Fix merge issue

olim7t:


tatu-at-datastax:
Ah! I should have thought about this; even if it's REST v1 it still needs to go through new port number scheme.
Thank you for fix!

tatu-at-datastax (approved):

mpenick (approved):
:+1: I ran into this issue the other day with a different test when merging master.

PR #2028, opened 2022-08-15
sign v2 container images with cosign

ivansenic:
**What this PR does**:
Signs the V2 docer images using cosign.. 

This is based on the `ivansenic-release-fix` branch, otherwise there would be conflicts to solve.. Note that merge target is `master`, thus there is a need for the V2 branch merging after this is in..

Internal issues

**Checklist**
- [x] Add secrets to the github actions
- [x] ~Commit public key as well, this would enable others to verify our images?~
- [x] Add password and private key to the stargate store 
- [x] sync with https://github.com/stargate/stargate/pull/2039

### DIFF
--- a/.github/workflows/release-v2.yml
+++ b/.github/workflows/release-v2.yml
@@ -304,7 +304,7 @@ jobs:
 
     strategy:
       matrix:
-        image: [coordinator-3_11, coordinator-4_0, coordinator-dse-68, restapi, graphqlapi, docsapi]
+        image: [restapi, graphqlapi, docsapi]
 
     env:
       # TODO newest version is 1.10.1, this reflects riptano action target version
--- a/.github/workflows/release-v2.yml
+++ b/.github/workflows/release-v2.yml
@@ -294,6 +294,54 @@ jobs:
           JAVA_HOME=$JAVA_17 ./mvnw clean package -DskipTests -Dquarkus.container-image.build=true -Dquarkus.container-image.push=true -Dquarkus.container-image.registry=${{ secrets.ECR_REPOSITORY }} -Dquarkus.container-image.tag=${{needs.resolve-tag.outputs.release-tag}}
           cd ../
 
+  # signs all docker images with cosign
+  # skip whole job if we did not push images
+  sign-images:
+    name: Sign container images
+    needs: [resolve-tag, publish-docker]
+    if: ${{ !inputs.skipPublish }}
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        image: [coordinator-3_11, coordinator-4_0, coordinator-dse-68, restapi, graphqlapi, docsapi]
+
+    env:
+      # TODO newest version is 1.10.1, this reflects riptano action target version
+      COSIGN_VERSION: v1.9.0
+
+    steps:
+      - name: Install Cosign
+        uses: sigstore/cosign-installer@main
+        with:
+          cosign-release: ${COSIGN_VERSION}
+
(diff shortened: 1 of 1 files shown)


ivansenic:
I am not sure if we should target the `v2.0.0` branch here?

tatu-at-datastax:
> I am not sure if we should target the `v2.0.0` branch here?

It's probably good to merge via `master` -- it is always bit confusing when workflows vary between actively developed branches?

.github/workflows/release-v2.yml
```diff
@@ -288,6 +288,54 @@ jobs:
...
+          aws-region: us-east-1
+
+      - name: Login to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v1
+
+      - name: Sign a docker image
+        shell: bash
+        env:
+          COSIGN_IMAGE: ${{ secrets.ECR_REPOSITORY }}/${{ matrix.image }}:${{ needs.resolve-tag.outputs.release-tag }}
```
ivansenic, line 327: should be fixed if `${{ secrets.ECR_REPOSITORY }}` is changed!

.github/workflows/release-v2.yml
```diff
@@ -288,6 +288,54 @@ jobs:
           JAVA_HOME=$JAVA_17 ./mvnw clean package -DskipTests -Dquarkus.container-image.build=true -Dquarkus.container-image.push=true -Dquarkus.container-image.registry=${{ secrets.ECR_REPOSITORY }} -Dquarkus.container-image.tag=${{needs.resolve-tag.outputs.release-tag}}
           cd ../
 
+  # signs all docker images with cosign
+  # skip whole job if we did not push images
+  sign-images:
+    name: Sign container images
+    needs: [resolve-tag, publish-docker]
+    if: ${{ !inputs.skipPublish }}
```
tatu-at-datastax, line 296: I only now realize that `skipPublish` is bit ambiguous wrt Maven vs Docker publish?
It looks like it would mean skipping both Maven artifact and Docker image publish, right?

tatu-at-datastax:
LGTM but hoping someone familiar with signing (if anyone?) could have a look as well.

tatu-at-datastax (approved):

.github/workflows/release-v2.yml
ivansenic: Yes this is what I also specified in the description:

```
   skipPublish:
        description: 'Skip publishing docker images and JARs to OSSRH?'
```

ivansenic:
@tatu-at-datastax I need help with resolving those two TODOs as well..

.github/workflows/release-v2.yml
```diff
@@ -294,6 +294,54 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        image: [coordinator-3_11, coordinator-4_0, coordinator-dse-68, restapi, graphqlapi, docsapi]
```
ivansenic, line 307: coordinator images are not pushed to the ecr, this needs fixing..

.github/workflows/release-v2.yml
tatu-at-datastax: Ok sorry I missed that & was only trying to remember the logic. Sounds good.

tatu-at-datastax:
> @tatu-at-datastax I need help with resolving those two TODOs as well..

I marked discussions as resolved -- is something else needed?

ivansenic:
> 

I meant two todos in the code.. never mind..

PR #2029, opened 2022-08-15
making all int test in Docs V2 to use @QuarkusIntegrationTest

ivansenic:
**What this PR does**:
Makes all integration tests in Docs V2 real end-to-end tests using `@QuarkusIntegrationTest`.. `@QuarkusIntegrationTest` supports executing tests against an already running instance of the application.. This way we can run int tests against any application setup..

All improvements:

* end-to-end tests possible
* fixes in the test resource to enable int test using the docker container of the app (active for docs api CI workflow)
* auth token read from system props, instead of using fixed token
* option to disable fixed token (this is only not to break rest and gql)
* execute tests against the running app

Internal issue.

**Checklist**
- [x] Don't break `restapi` and `graphqlapi`
- [x] Auth token instead of fixed token setup
- [x] Resource loading dismiss on `-Dquarkus.http.test-host`
- [x] Test against running app
- [x] INT tests with docker buiild

### DIFF
--- a/.github/workflows/apis-v2.yaml
+++ b/.github/workflows/apis-v2.yaml
@@ -317,10 +317,13 @@ jobs:
 
       # run finally the int tests
       # runs dedicated project with -pl, but also picks depending projects with -am
+      # container test for docsapi only for now, enable for other when ready
       - name: Integration Test
+        env:
+          CONTAINER_TEST: ${{ matrix.project == 'sgv2-docsapi' && 'true' || 'false' }}
         run: |
           cd apis/
-          ./mvnw -B verify -DskipUnitTests -pl ${{ matrix.project }} -am -P ${{ matrix.profile }} -Dtesting.containers.stargate-image=${{ matrix.image }}
+          ./mvnw -B verify -DskipUnitTests -pl ${{ matrix.project }} -am -P ${{ matrix.profile }} -Dquarkus.container-image.build=$CONTAINER_TEST -Dtesting.containers.stargate-image=${{ matrix.image }}
 
   # runs always, deletes built docker image artifacts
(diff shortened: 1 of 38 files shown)


ivansenic:
Currently there is an issue with the unit tests, as the resource used in the integration test is considered as global. I will find out the best solution for this.. For now I have opened an issue: https://github.com/quarkusio/quarkus/issues/27285

ivansenic:
Looking for reviews here:

* @EricBorczuk from now on, in docs-api when you use `@QuarkusTest` you need to use the `NoGlobalResourcesTestProfile.Impl` or make a test profile extend the `NoGlobalResourcesTestProfile`.. That's like a default profile now
* @olim7t @tatu-at-datastax Please have a deep look on this PR, I now managed to run all the integration tests in docs-api using `@QuarkusIntegrationTest`, in the CI I even test against the docker build, which is great imo.. I did everything not to break any work you are doing currently, so merging this issue **should not** force you to change anything..

.github/workflows/apis-v2.yaml
```diff
@@ -317,10 +317,13 @@ jobs:
 
       # run finally the int tests
       # runs dedicated project with -pl, but also picks depending projects with -am
+      # container test for docsapi only for now, enable for other when ready
       - name: Integration Test
+        env:
+          CONTAINER_TEST: ${{ matrix.project == 'sgv2-docsapi' && 'true' || 'false' }}
```
EricBorczuk, line 323: What does the `&& 'true' || 'false'` do here?

EricBorczuk (approved):
Looks good - will incorporate these changes on my branch once this is merged

PR #2038, opened 2022-08-16
fix tag value resolving on release workflows

ivansenic:
**What this PR does**:

Fixes after running v2 release:

* release tag not resolved correctly, somehow not noticed in v1 as create release step does accept the invalid value

Internal issue

### DIFF
--- a/.github/workflows/release-v2.yml
+++ b/.github/workflows/release-v2.yml
@@ -33,10 +33,16 @@ jobs:
     outputs:
       release-tag: ${{ steps.resolve_tag.outputs.tag }}
     steps:
+      - uses: actions/checkout@v3
+
+      - name: Set reference
+        id: vars
+        run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
+
       - name: Resolve tag
         id: resolve_tag
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release-v2.yml
```diff
@@ -33,10 +33,16 @@ jobs:
...
+
+      - name: Set reference
+        id: vars
+        run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
+
       - name: Resolve tag
         id: resolve_tag
         run: |
-          TAG=${{ inputs.tag != null && inputs.tag || github.ref }}
+          TAG=${{ inputs.tag != null && inputs.tag || steps.vars.outputs.tag }}
```
tatu-at-datastax, line 45: I may be wrong, but wouldn't this be same as simple:

    TAG=${{ inputs.tag || steps.vars.outputs.tag }}

assuming Github action follows usual "falsy" logic of scripting languages.

tatu-at-datastax (approved):

.github/workflows/release-v2.yml
ivansenic: Could be, but I don't want to have a yet another failure here.. This is working imo..

PR #25, opened 2020-09-10
Readme enhancements

csplinter:

PR #2040, opened 2022-08-17
$set support

EricBorczuk:
**What this PR does**:
Adds a builtin function `$set` which allows setting particular fields in a document. This could be combined in the future with atomic find() statements with a small amount of work.

Fixes [#25](https://github.com/stargate/stargate-mongoose/issues/25)

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/model/dto/BuiltInFunctionDto.java
+++ b/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/model/dto/BuiltInFunctionDto.java
@@ -27,12 +27,15 @@ import org.eclipse.microprofile.openapi.annotations.media.Schema;
 /** DTO for a namespace. */
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public record BuiltInFunctionDto(
-    @Schema(description = "The operation to execute", pattern = "\\$pop|\\$push", example = "$push")
+    @Schema(
+            description = "The operation to execute",
+            pattern = "\\$pop|\\$push|\\$set",
+            example = "$push")
         @NotNull(message = "`operation` is required for function execution")
         @NotBlank(message = "`operation` is required for function execution")
         @Pattern(
-            regexp = "\\$pop|\\$push",
-            message = "available built-in functions are $pop and $push")
+            regexp = "\\$pop|\\$push|\\$set",
+            message = "available built-in functions are [$pop, $push, $set]")
         String operation,
     @Schema(
             description = "The value to use for the operation",
--- a/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/function/impl/SetFunction.java
+++ b/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/function/impl/SetFunction.java
@@ -29,11 +29,6 @@ public class SetFunction implements BuiltInFunction {
   /** {@inheritDoc} */
   @Override
   public Uni<JsonNode> execute(Uni<Schema.CqlTable> table, BuiltInFunction.Data data) {
-    // convert data input into rows
-    if (!data.input().isObject()) {
-      return null;
-    }
-
     return writeDocumentsService
         .setPathsOnDocument(
             table,
@@ -44,6 +39,6 @@ public class SetFunction implements BuiltInFunction {
             data.input(),
             true,
             data.executionContext())
-        .map(__ -> mapper.createObjectNode());
+        .map(__ -> mapper.nullNode());
   }
 }
--- a/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/function/impl/SetFunction.java
+++ b/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/function/impl/SetFunction.java
@@ -3,7 +3,6 @@ package io.stargate.sgv2.docsapi.service.function.impl;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
-import io.smallrye.mutiny.Multi;
 import io.smallrye.mutiny.Uni;
 import io.stargate.bridge.proto.Schema;
 import io.stargate.sgv2.docsapi.api.v2.model.dto.DocumentResponseWrapper;
@@ -52,21 +51,22 @@ public class SetFunction implements BuiltInFunction {
       // Turn the dottedPath into expected path format
       String[] pathSegments = dottedPath.split("\\.");
       List<String> subPath = Arrays.asList(pathSegments);
+      List<String> fullPath = new ArrayList<>(data.documentPath());
+      fullPath.addAll(subPath);
       setResults.add(
           writeDocumentsService.updateSubDocument(
               table,
               data.namespace(),
               data.collection(),
               data.documentId(),
-              subPath,
+              fullPath,
               subDoc,
               false,
               data.executionContext()));
     }
-    return Multi.createFrom()
-        .iterable(setResults)
-        .collect()
-        .asList()
+    return Uni.join()
+        .all(setResults)
+        .andFailFast()
         .onItem()
         .transform(__ -> mapper.createObjectNode());
   }
--- a/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/function/impl/SetFunction.java
+++ b/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/function/impl/SetFunction.java
@@ -1,6 +1,7 @@
 package io.stargate.sgv2.docsapi.service.function.impl;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.smallrye.mutiny.Multi;
 import io.smallrye.mutiny.Uni;
@@ -24,6 +25,8 @@ import javax.inject.Inject;
 @ApplicationScoped
 public class SetFunction implements BuiltInFunction {
   @Inject WriteDocumentsService writeDocumentsService;
+
+  @Inject ObjectMapper mapper;
 
   /** {@inheritDoc} */
   @Override
@@ -65,6 +68,6 @@ public class SetFunction implements BuiltInFunction {
         .collect()
         .asList()
         .onItem()
-        .transform(__ -> null);
+        .transform(__ -> mapper.createObjectNode());
   }
 }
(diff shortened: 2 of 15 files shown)


ivansenic:
@EricBorczuk let me know when this is ready for review.. Finally the function refactoring pays off, so simple to add a new function :)

apis/sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResourceIntegrationTest.java
```diff
@@ -313,7 +320,112 @@ public void pushNoArray() {
...
+          .when()
+          .get(BASE_PATH, DEFAULT_NAMESPACE, DEFAULT_COLLECTION, DOCUMENT_ID)
+          .then()
+          .body("documentId", equalTo(DOCUMENT_ID))
+          .body("data", jsonPartEquals("object.b[1].nested", "newvalue"))
+          .statusCode(200);
+    }
+
+    @Test
+    public void setUpdateOneValueCreate() {
```
EricBorczuk, line 354: To be on par with mongo, if you `set` a field that doesn't yet exist, it will get created for you.

apis/sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResourceIntegrationTest.java
```diff
@@ -313,7 +320,112 @@ public void pushNoArray() {
...
-          .body("description", equalTo("The path provided to push to has no array, found {}."));
+          .body(
+              "description",
+              equalTo(
+                  "The path provided to push to has no array, found {\"a\":3,\"b\":[{},{\"nested\":\"value\"},[]]}."));
+    }
+
+    @Test
+    public void setUpdateOneValue() {
+      String setOperation = "{ \"b.[1].nested\": \"newvalue\" }";
```
vkarpov15, line 331: Do nested arrays work? Like `String setOperation = "{ \"b.[1].c.[2].d\": \"newvalue\" }"` ?

apis/sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResourceIntegrationTest.java
```diff
@@ -313,7 +320,112 @@ public void pushNoArray() {
...
+          .statusCode(200);
+
+      // get whole document and check the value
+      given()
+          .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, "")
+          .when()
+          .get(BASE_PATH, DEFAULT_NAMESPACE, DEFAULT_COLLECTION, DOCUMENT_ID)
+          .then()
+          .body("documentId", equalTo(DOCUMENT_ID))
+          .body("data", jsonPartEquals("object.b[1].nested", "newvalue"))
```
vkarpov15, line 398: Would be nice for this test to assert that other fields in `object.b[1]` are unaffected

vkarpov15 (commented):
Had a couple of questions, but overall looks good :+1:

apis/sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResourceIntegrationTest.java
EricBorczuk: It should, and I will add a test to that effect :)

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/model/dto/BuiltInFunctionDto.java
```diff
@@ -27,12 +27,15 @@
...
+    @Schema(
+            description = "The operation to execute",
+            pattern = "\\$pop|\\$push|\\$set",
+            example = "$push")
         @NotNull(message = "`operation` is required for function execution")
         @NotBlank(message = "`operation` is required for function execution")
         @Pattern(
-            regexp = "\\$pop|\\$push",
-            message = "available built-in functions are $pop and $push")
+            regexp = "\\$pop|\\$push|\\$set",
```
ivansenic, line 37: we should add a body example for the new function in the api documentation, same way we have for the pop and push

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/function/impl/SetFunction.java
```diff
@@ -0,0 +1,73 @@
...
+          writeDocumentsService.updateSubDocument(
+              table,
+              data.namespace(),
+              data.collection(),
+              data.documentId(),
+              fullPath,
+              subDoc,
+              false,
+              data.executionContext()));
+    }
```
ivansenic, line 66: This would fire n amount of the write requests to the bridge, which is not really optmial.. Furthermore, if one fails, then this `Uni.join` would emit an error to the user, however all the successful ops would be written and user has no idea how some of the set succeeded and how some not..

Thus, we should have only one write operation.. In order to do this, we need to merge the sub-documents together.. I know it's not that easy considering you also need to care about the `subPath`, but it's doable with a bit of gymnastics..

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/function/impl/SetFunction.java
```diff
@@ -0,0 +1,73 @@
...
+    return Uni.join()
+        .all(setResults)
+        .andFailFast()
+        .onItem()
+        .transform(__ -> mapper.createObjectNode());
```
ivansenic, line 71: should set return the updated document back or what's the idea?

ivansenic (changes requested):
I think we need to re-think the write ops against the bridge..

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/function/impl/SetFunction.java
EricBorczuk: It should return nothing, I think - but upstream, encountering a `null` causes an error to be emitted

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/function/impl/SetFunction.java
EricBorczuk: Hmmm, yeah - maybe this would merit a new method in the writeDocumentsService. I feel like this _should_ amount to just a single C* batch with a bunch of row inserts in it

ivansenic (changes requested):

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/WriteDocumentsService.java
```diff
@@ -332,6 +335,87 @@ public Uni<DocumentResponseWrapper<Void>> patchDocument(
...
+      String namespace,
+      String collection,
+      String documentId,
+      List<String> subPath,
+      JsonNode payload,
+      Integer ttl,
+      ExecutionContext context) {
+    // the payload will have a representation of all paths that need a partial update
+    return jsonSchemaManager
+        .validateJsonDocument(table, payload, true)
```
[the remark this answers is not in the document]
EricBorczuk: This is technically just checking that there is no schema assigned, since the boolean here is for `has sub path`. It will error out if the user has any sort of schema and tries to $set.

One day we'll support schema on subdocument changes...

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/function/impl/SetFunction.java
```diff
@@ -0,0 +1,49 @@
...
+        .setPathsOnDocument(
+            table,
+            data.namespace(),
+            data.collection(),
+            data.documentId(),
+            data.documentPath(),
+            data.input(),
+            true,
+            data.executionContext())
+        .map(__ -> mapper.createObjectNode());
```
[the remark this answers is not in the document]
EricBorczuk: $set shouldn't return the updated document, it should just return that the operation was successful; so I think returning NullNode is best

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/WriteDocumentsService.java
```diff
@@ -332,6 +335,87 @@ public Uni<DocumentResponseWrapper<Void>> patchDocument(
...
+    while (fieldsIterator.hasNext()) {
+      String dottedField = fieldsIterator.next();
+      List<String> fieldPaths = processSubDocumentPath(Arrays.asList(dottedField.split("\\.")));
+      List<String> fullPath = new ArrayList<>();
+      fullPath.addAll(subPathProcessed);
+      fullPath.addAll(fieldPaths);
+      allPaths.add(fullPath);
+    }
+    return allPaths;
+  }
```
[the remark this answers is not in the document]
EricBorczuk: `convertFieldsToPaths` expects an ArrayNode, whereas this is expecting an ObjectNode. Perhaps parts of it can be reused though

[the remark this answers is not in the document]
EricBorczuk: Yeah I extracted part of `convertFieldsToPaths` into its own method in DocsApiUtils, to make sure we are reusing the same logic

apis/sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResourceIntegrationTest.java
```diff
@@ -101,6 +103,8 @@ class ExecuteBuiltInFunction {
 
     public static final String POP_PAYLOAD = "{\"operation\": \"$pop\"}";
 
+    public static final String SET_PAYLOAD = "{\"operation\": \"$set\", \"value\": %s}";
+
```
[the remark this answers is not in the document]
EricBorczuk: I'll add it - to be clear, it _should_ fail in all cases; since $set is an operation on subdocuments, it is disallowed if a schema is attached.

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/WriteDocumentsService.java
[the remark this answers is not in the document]
ivansenic: Ahhh smart..

apis/sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResourceIntegrationTest.java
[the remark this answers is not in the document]
ivansenic: Good.

ivansenic (approved):
Great work :+1:

PR #2044, opened 2022-08-23
remove c2 from int tests on V2

ivansenic:
**What this PR does**:
Remove `c2` from the CI workflow. It's not having any benefits, if needed will be done outside the open-source repo.


tatu-at-datastax (approved):
Good!

PR #2050, opened 2022-08-30
enable population of the token in the swagger-ui

ivansenic:
**What this PR does**:
Enables population of the auth token in the swagger UI, by reading it from the `sg-swagger-token` cookie..

The reviewer must confirm it works by:
* opening `http://localhost:8180/swagger-ui`
* open browser dev tools
* add cookie with name `sg-swagger-token` with arbitrary value
* hit refresh on the page
* confirm that 
    * authorize box has token already (closed lock)
    * try now executes a request with the token having the value of the cookie

### DIFF
--- a/apis/sgv2-docsapi/src/main/resources/application.yaml
+++ b/apis/sgv2-docsapi/src/main/resources/application.yaml
@@ -147,3 +147,16 @@ quarkus:
     path: /swagger-ui
     always-include: true
     title: ${quarkus.application.name}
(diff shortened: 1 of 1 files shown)


EricBorczuk (approved):
I tried it out locally using the step-by-step you described, and it works!

ivansenic:
@EricBorczuk I think I would merge this and then add support for deleting the cookie after read if we need it?

PR #2067, opened 2022-09-13
 Allow executing a built in function against the root of a document 

EricBorczuk:
**What this PR does**:
Allow executing a built in function against the root of a document (before it was producing 405 Method Not Allowed because it was matching a different route)

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
+++ b/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
@@ -113,7 +113,7 @@ public class BuiltInFunctionResource {
       })
   @POST
   @Path(
-      "{collection:\\w+}/{document-id}{slash: /?}{document-path: (.*)?}/function{trailingSlash: /?}")
+      "{collection:\\w+}/{document-id: [^/]+}{slash: /?}{document-path: (.+)?}/function{trailingSlash: /?}")
   public Uni<RestResponse<Object>> executeBuiltInFunction(
       @Context UriInfo uriInfo,
       @PathParam("namespace") String namespace,
@@ -124,8 +124,12 @@ public class BuiltInFunctionResource {
       @QueryParam("raw") boolean raw,
       @NotNull(message = "payload not provided") @Valid BuiltInFunctionDto body) {
     ExecutionContext context = ExecutionContext.create(profile);
+
     List<String> subPath =
-        documentPath.stream().map(PathSegment::getPath).collect(Collectors.toList());
+        documentPath.stream()
+            .map(PathSegment::getPath)
+            .filter(p -> !p.isEmpty())
+            .collect(Collectors.toList());
 
     // call table manager for valid table
     return collectionManager
--- a/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
+++ b/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
@@ -112,7 +112,8 @@ public class BuiltInFunctionResource {
         @APIResponse(ref = OpenApiConstants.Responses.GENERAL_503),
       })
   @POST
-  @Path("{collection:\\w+}/{document-id}/{document-path:.*}/function")
+  @Path(
+      "{collection:\\w+}/{document-id}{slash: /?}{document-path: (.*)?}/function{trailingSlash: /?}")
   public Uni<RestResponse<Object>> executeBuiltInFunction(
       @Context UriInfo uriInfo,
       @PathParam("namespace") String namespace,
--- a/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
+++ b/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
@@ -25,6 +25,7 @@ import io.stargate.sgv2.docsapi.config.constants.OpenApiConstants;
 import io.stargate.sgv2.docsapi.service.ExecutionContext;
 import io.stargate.sgv2.docsapi.service.function.FunctionService;
 import io.stargate.sgv2.docsapi.service.schema.CollectionManager;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
@@ -36,11 +37,9 @@ import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
-import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.PathSegment;
 import javax.ws.rs.core.Response;
-import javax.ws.rs.core.UriInfo;
 import org.eclipse.microprofile.openapi.annotations.Operation;
 import org.eclipse.microprofile.openapi.annotations.media.Content;
 import org.eclipse.microprofile.openapi.annotations.media.ExampleObject;
@@ -69,6 +68,62 @@ public class BuiltInFunctionResource {
 
   @Operation(
       summary = "Executes a a built-in function",
+      description =
+          "Execute a built-in function (e.g. `$push` and `$pop`) against the root of this document. Performance may vary.")
+  @Parameters(
+      value = {
+        @Parameter(name = "namespace", ref = OpenApiConstants.Parameters.NAMESPACE),
+        @Parameter(name = "collection", ref = OpenApiConstants.Parameters.COLLECTION),
+        @Parameter(name = "document-id", ref = OpenApiConstants.Parameters.DOCUMENT_ID),
+        @Parameter(name = "profile", ref = OpenApiConstants.Parameters.PROFILE),
+        @Parameter(name = "raw", ref = OpenApiConstants.Parameters.RAW),
+      })
+  @RequestBody(ref = OpenApiConstants.RequestBodies.FUNCTION)
+  @APIResponses(
+      value = {
+        @APIResponse(
+            responseCode = "200",
+            description =
+                "Function is executed. Certain functions will return a value in the `data` property. Note that in case of unwrapping (`raw=true`), the response contains only the contents of the `data` property.",
+            content = {
+              @Content(
+                  schema =
+                      @org.eclipse.microprofile.openapi.annotations.media.Schema(
+                          implementation = DocumentResponseWrapper.class),
+                  examples = @ExampleObject(ref = OpenApiConstants.Examples.FUNCTION_RESPONSE))
+            }),
+        @APIResponse(
+            responseCode = "404",
+            description = "Not found.",
+            content =
+                @Content(
+                    examples = {
+                      @ExampleObject(ref = OpenApiConstants.Examples.NAMESPACE_DOES_NOT_EXIST),
+                      @ExampleObject(ref = OpenApiConstants.Examples.COLLECTION_DOES_NOT_EXIST),
+                    },
+                    schema =
+                        @org.eclipse.microprofile.openapi.annotations.media.Schema(
+                            implementation = ApiError.class))),
+        @APIResponse(ref = OpenApiConstants.Responses.GENERAL_400),
+        @APIResponse(ref = OpenApiConstants.Responses.GENERAL_401),
+        @APIResponse(ref = OpenApiConstants.Responses.GENERAL_500),
+        @APIResponse(ref = OpenApiConstants.Responses.GENERAL_503),
+      })
+  @POST
+  @Path("{collection:\\w+}/{document-id}/function")
+  public Uni<RestResponse<Object>> executeBuiltInFunction(
+      @PathParam("namespace") String namespace,
+      @PathParam("collection") String collection,
+      @PathParam("document-id") String documentId,
+      @QueryParam("profile") boolean profile,
+      @QueryParam("raw") boolean raw,
+      @NotNull(message = "payload not provided") @Valid BuiltInFunctionDto body) {
+    return executeBuiltInFunction(
+        namespace, collection, documentId, Collections.emptyList(), profile, raw, body);
+  }
+
+  @Operation(
+      summary = "Executes a a built-in function against a path in a document",
       description =
           "Execute a built-in function (e.g. `$push` and `$pop`) against a value in this document. Performance may vary.")
   @Parameters(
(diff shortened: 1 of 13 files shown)


apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
```diff
@@ -113,7 +113,7 @@ public class BuiltInFunctionResource {
       })
   @POST
   @Path(
-      "{collection:\\w+}/{document-id}{slash: /?}{document-path: (.*)?}/function{trailingSlash: /?}")
+      "{collection:\\w+}/{document-id: [^/]+}{slash: /?}{document-path: (.+)?}/function{trailingSlash: /?}")
```
tatu-at-datastax, line 116: On `document-id`: would slash actually be allowed otherwise? I assumed it usually wouldn't. But I assume that it would be (otherwise wouldn't have need to prevent it)

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
```diff
@@ -124,8 +124,12 @@ public Uni<RestResponse<Object>> executeBuiltInFunction(
       @QueryParam("raw") boolean raw,
       @NotNull(message = "payload not provided") @Valid BuiltInFunctionDto body) {
     ExecutionContext context = ExecutionContext.create(profile);
+
     List<String> subPath =
-        documentPath.stream().map(PathSegment::getPath).collect(Collectors.toList());
+        documentPath.stream()
+            .map(PathSegment::getPath)
+            .filter(p -> !p.isEmpty())
```
tatu-at-datastax, line 131: So empty path segments are allowed, just trimmed out?

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/DocumentDeleteResource.java
```diff
@@ -158,7 +158,10 @@ public Uni<RestResponse<Object>> deleteSubDocument(
       @QueryParam("profile") boolean profile) {
     ExecutionContext context = ExecutionContext.create(profile);
     List<String> subPath =
-        documentPath.stream().map(PathSegment::getPath).collect(Collectors.toList());
+        documentPath.stream()
+            .map(PathSegment::getPath)
+            .filter(p -> !p.isEmpty())
+            .collect(Collectors.toList());
```
tatu-at-datastax, line 164: Not a big deal but since it's called from a few resources, could also extract into helper method.

apis/sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResourceIntegrationTest.java
```diff
@@ -411,11 +411,9 @@ public void setUpdateManyValuesBaseDoc() {
           .get(BASE_PATH, DEFAULT_NAMESPACE, DEFAULT_COLLECTION, DOCUMENT_ID)
           .then()
           .body("documentId", equalTo(DOCUMENT_ID))
-          .body("data", jsonPartEquals("b[1].nested", "value"))
```
tatu-at-datastax: Why were these removed?

tatu-at-datastax (approved):

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
EricBorczuk: Yeah, if you do something like /api/v2/namespaces/x/collections/y/document/a//b/c, I just figured it should ignore the typo

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
EricBorczuk: Yea I noticed that because of how quarkus URI parsing works with slashes, it actually slurps everything between literal `/` for each token. So I was getting matches on `{document-id: .+}{slash: /?}{document-path: (.+)?}` that were actually matching `abcde/fghij` as the value of document-id (which is ...not right)

apis/sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResourceIntegrationTest.java
EricBorczuk: This entire test fn was added - but then I realized that because it is updating data on the root of the document, the data at the root is just different.

In the test this was copied from, I was setting data on a nested path, and then checking that both the data at the nested path previously and the new data were both present.

tatu-at-datastax: Ok

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
tatu-at-datastax: Yeah, just odd how it would include slash within if you do NOT include pattern. I would expect it to essentially use "anything but slash". But I guess not :)

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
EricBorczuk: Yeaaaa that actually might be a problem in certain cases...it would mean that we have to restrict JSON fields from having empty string as well, or else you could write JSON with an empty string as field and then never be able to read that path directly via the URI. Let me check to see how we handle this, because also I think that empty-string serves as the delimiter for where the nesting ends in the cassandra row

tatu-at-datastax: Right. To me it seems reasonable not to allow these, but I am not 100% sure how Ivan feels about that.
I could see empty String as key for some sort of special metadata, and it is probably used by someone somewhere.
Does not mean we have to support it but we need to decide one way or the other.

ivansenic (changes requested):
Few comments on my side:

1. was a trailing slash really necessary? I would say no, why was this added?
2. the whole change of the endpoints, now require an update of the path matching for the micrometer metrics.. however, with being so complex now, I have no idea how to adapt this now..

I am really opting in only fixing the `/function` stuff.. Blocking merge until we resolve this..

EricBorczuk:
The trailing-slash changes were done because Val's mongoose stuff allows either, and it made sense for this to support both if it was easy to do...but if micrometer metrics become impossible, then maybe I would cut it...

jeffreyscarpenter:
Can we rename this PR to be more informative, maybe something like "URI format fixes to enable stargate-mongoose"?

EricBorczuk:
I actually removed the trailing slash changes from this PR, so it only has one significant change. Updated the description to reflect that

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
```diff
@@ -112,7 +112,7 @@ public class BuiltInFunctionResource {
         @APIResponse(ref = OpenApiConstants.Responses.GENERAL_503),
       })
   @POST
-  @Path("{collection:\\w+}/{document-id}/{document-path:.*}/function")
+  @Path("{collection:\\w+}/{document-id: [^/]+}{slash: /?}{document-path: (.+)?}/function")
```
ivansenic, line 115: OK, but we still have to fix the micrometer template on this line:

https://github.com/stargate/stargate/blob/2384046a394613eb2d9a6d4dcea2e7712eea4f5f/apis/sgv2-docsapi/src/main/resources/application.yaml#L111

Seems like only the right hand side has to be updated to the new path spec from that `@Path`.. The best way to test this is to hit that endpoint without authentication and see what's get counted in the `/metrics`.. And then hit with the token and assure both calls are having same path tag..

Failing to do this we are risking of a metrics explosion, it's very high risk imo.. Please let me know if you have questions!

ivansenic (changes requested):

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
EricBorczuk: I feel like the left side needs to be changed, too - no? Because `/v2/namespaces/X/collections/Y/documentId/function` doesn't match the left side regex?

ivansenic (commented):
OK the metric changes are fine, tested and it works :heavy_check_mark: 

However, have a look on the openapi, the endpoint is displayed as:

```
[/v2/namespaces/{namespace}/collections/{collection}/{document-id}{slash}{document-path}/function](http://localhost:8180/swagger-ui/#/Documents/post_v2_namespaces__namespace__collections__collection___document_id__slash__document_path__function)
Executes a a built-in function
```

And built in curl is:

```
curl -X 'POST' \
  'http://localhost:8180/v2/namespaces/cycling/collections/events/id{slash}path/function' \
  -H 'accept: application/json' \
  -d '{
  "operation": "$pop"
}'
```

Is this actually the right way to develop this? Should we just simply add a new endpoint with 

```
{collection:\\w+}/{document-id}/function
```

and delegate to that method below.. I am completely confused with the slash stuff? Is slash default when targeting the root document?

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
```diff
@@ -112,9 +167,8 @@ public class BuiltInFunctionResource {
         @APIResponse(ref = OpenApiConstants.Responses.GENERAL_503),
       })
   @POST
-  @Path("{collection:\\w+}/{document-id}/{document-path:.*}/function")
+  @Path("{collection:\\w+}/{document-id}/{document-path: .+}/function")
```
ivansenic, line 170: why this change?

apis/sgv2-docsapi/src/main/resources/application.yaml
```diff
@@ -108,7 +108,7 @@ quarkus:
           /v2/namespaces/[\\w]+/collections/[\\w]+/batch=/v2/namespaces/{namespace:\\w+}/collections/{collection:\\w+}/batch,
           /v2/namespaces/[\\w]+/collections/[\\w]+/json-schema=/v2/namespaces/{namespace:\\w+}/collections/{collection:\\w+}/json-schema,
           /v2/namespaces/[\\w]+/collections/[\\w]+/upgrade=/v2/namespaces/{namespace:\\w+}/collections/{collection:\\w+}/upgrade,
-          /v2/namespaces/[\\w]+/collections/[\\w]+/(.*?)/.*/function=/v2/namespaces/{namespace:\\w+}/collections/{collection:\\w+}/{document-id}/{document-path:.*}/function,
+          /v2/namespaces/[\\w]+/collections/[\\w]+/(.*?)/?.*/function=/v2/namespaces/{namespace:\\w+}/collections/{collection:\w+}/{document-id: [^/]+}{slash: /?}{document-path: (.+)?}/function,
```
ivansenic, line 111: This is not valid anymore:

1. new endpoint added
2. other endpoint has changed `.*` to `.+`

Not sure if this is correct, but should be:

```
/v2/namespaces/[\\w]+/collections/[\\w]+/(.*?)/.*/function=/v2/namespaces/{namespace:\\w+}/collections/{collection:\\w+}/{document-id}/{document-path:.+}/function
/v2/namespaces/[\\w]+/collections/[\\w]+/(.*?)/function=/v2/namespaces/{namespace:\\w+}/collections/{collection:\\w+}/{document-id}/function
```

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
```diff
@@ -123,8 +177,12 @@ public Uni<RestResponse<Object>> executeBuiltInFunction(
       @QueryParam("raw") boolean raw,
       @NotNull(message = "payload not provided") @Valid BuiltInFunctionDto body) {
     ExecutionContext context = ExecutionContext.create(profile);
+
     List<String> subPath =
-        documentPath.stream().map(PathSegment::getPath).collect(Collectors.toList());
+        documentPath.stream()
+            .map(PathSegment::getPath)
+            .filter(p -> !p.isEmpty())
+            .collect(Collectors.toList());
```
ivansenic, line 185: can this also be reverted now to the old way or now? not sure..

ivansenic (approved):
Approved, but there are still some issues..

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
EricBorczuk: accidental! I will revert

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/api/v2/namespaces/collections/documents/BuiltInFunctionResource.java
EricBorczuk: Yeah, it can - shouldn't make a difference now

ivansenic (approved):

tatu-at-datastax:
@EricBorczuk Could you merge this now? We are planning on getting this in 2.0.0-BETA-4 build.

PR #2072, opened 2022-09-14
Change sgv2 REST API default port back to 8082

tatu-at-datastax:
**What this PR does**:

Sets default `quarkus.https.port` to `8082` for REST API to make `docker-compose` work, and should be compatible with Operator settings used for SGv2 rollout.

In future we probably want to change default of all Services/APIs to 8080, but that needs to be coordinated across components we have.

Fixes #<issue number>

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/sgv2-restapi/pom.xml
+++ b/apis/sgv2-restapi/pom.xml
@@ -12,6 +12,7 @@
     <failsafe.useModulePath>false</failsafe.useModulePath>
     <quarkus.container-image.group>stargateio</quarkus.container-image.group>
     <quarkus.container-image.name>restapi</quarkus.container-image.name>
+    <quarkus.container-image.tag>v${project.version}</quarkus.container-image.tag>
   </properties>
   <dependencies>
     <dependency>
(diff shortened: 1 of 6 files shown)


apis/sgv2-restapi/pom.xml
```diff
@@ -12,6 +12,7 @@
     <failsafe.useModulePath>false</failsafe.useModulePath>
     <quarkus.container-image.group>stargateio</quarkus.container-image.group>
     <quarkus.container-image.name>restapi</quarkus.container-image.name>
+    <quarkus.container-image.tag>v${project.version}</quarkus.container-image.tag>
```
tatu-at-datastax, line 15: Needed so the tag is `v2.0.0-BETA-2-SNAPSHOT` and not just `2.0.0-BETA-2-SNAPSHOT` (copied from `sgv2-graphqlapi/pom.xml`)

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/persistence/RestApiV2QMaterializedViewIT.java
```diff
@@ -81,6 +81,14 @@ public void getRowsFromMV() {
     assertThat(resultSet.wasApplied()).isTrue();
 
     // And then read entries using MV:
+
+    // 14-Sep-2022, tatu: Not sure why but there is a transient issue here wrt timing.
```
tatu-at-datastax, line 85: Non-elegant, but unfortunately attempts at using `@RetryableTest` were unsuccessful (does not work well together with "assumptions" since those are implemented via `AssertionException`s and taken as actual fails).

PR #2083, opened 2022-09-19
(WIP) Add retries for DataStorePropertiesConfiguration

### LINKED ISSUE #2075: `DataStorePropertiesConfiguration` needs to do retries, fail only after multiple retries
Currently Stargate V2 services will access persistence (backend) metadata during startup; this is done by class `DataStorePropertiesConfiguration`. It will only try one call to Bridge, and if that fails it gives up (logging stack trace) but continues with some arbitrary defaults which are unlikely to be correct values for the actual backend.

Instead, what should happen is that this class attempted call multiple times with some delay in-between, and if exhausting maximum time/attempts, fail to start up.

tatu-at-datastax:
** NOTE ***

Work-in-progress: needs tests, and for that little bit of refactoring to unit test the retry logic itself, without actual calls. Once that is done, ready for review.

----------

**What this PR does**:

Add configurable retries for bootstrapping of APIs/Services, wrt initial Persistence Metadata lookup via Bridge.

Fixes #2075

**Checklist**
- [ ] Changes manually tested


ivansenic:
I am questioning why is this needed at all?

tatu-at-datastax:
@ivansenic retries are typically used to work around transient connection/call failures, or if Coordinator(s) are not available temporarily during startup of service. That's the idea here.

Or are you thinking that startup should just fail, instance restarted, hoping for better outcome?

The original idea came from failure of docker-compose set up where everything was started at same time and coordinators took long enough to start so that metadata call failed. And docker-compose does not try restarting docker containers.

tatu-at-datastax:
Based on discussions with @ivansenic this approach is bit too complicated, considering multi-tenant use case does not even need any calls. Can and should replace with a simpler approach, using Quarkus-provided retry logic.

PR #2087, opened 2022-09-20
Add retry for data store properties access; new property to enable use of fallbacks

### LINKED ISSUE #2075: `DataStorePropertiesConfiguration` needs to do retries, fail only after multiple retries

tatu-at-datastax:
**What this PR does**:

Improves reliability of data store metadata lookup on startup; makes use of possible fallbacks configurable (defaulting to no fallbacks)

Fixes #2075

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/DataStoreConfig.java
+++ b/apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/DataStoreConfig.java
@@ -33,6 +33,13 @@ public interface DataStoreConfig {
   @WithDefault("${stargate.multi-tenancy.enabled}")
   boolean ignoreBridge();
 
+  /**
+   * @return If call(s) to fetch metadata fail, should we just return default settings ({@code
+   *     true}) or throw an exception ({@code false}). Defaults to {@code false}.
+   */
+  @WithDefault("false")
+  boolean useFallbacksIfCallFails();
+
   /** @return If the secondary indexes are enabled, defaults to <code>true</code>. */
   @WithDefault("true")
   boolean secondaryIndexesEnabled();
--- a/apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/DataStoreConfig.java
+++ b/apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/DataStoreConfig.java
@@ -38,7 +38,7 @@ public interface DataStoreConfig {
    *     true}) or throw an exception ({@code false}). Defaults to {@code false}.
    */
   @WithDefault("false")
-  boolean useFallbacksIfCallFails();
+  boolean bridgeFallbackEnabled();
 
   /** @return If the secondary indexes are enabled, defaults to <code>true</code>. */
   @WithDefault("true")
--- a/apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/properties/datastore/configuration/DataStorePropertiesConfiguration.java
+++ b/apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/properties/datastore/configuration/DataStorePropertiesConfiguration.java
@@ -25,6 +25,7 @@ import io.stargate.sgv2.api.common.config.DataStoreConfig;
 import io.stargate.sgv2.api.common.properties.datastore.DataStoreProperties;
 import io.stargate.sgv2.api.common.properties.datastore.impl.DataStorePropertiesImpl;
 import java.time.temporal.ChronoUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Produces;
 import org.eclipse.microprofile.faulttolerance.Retry;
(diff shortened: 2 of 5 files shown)


apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/DataStoreConfig.java
```diff
@@ -33,6 +33,13 @@ public interface DataStoreConfig {
   @WithDefault("${stargate.multi-tenancy.enabled}")
   boolean ignoreBridge();
 
+  /**
+   * @return If call(s) to fetch metadata fail, should we just return default settings ({@code
+   *     true}) or throw an exception ({@code false}). Defaults to {@code false}.
+   */
+  @WithDefault("false")
+  boolean useFallbacksIfCallFails();
```
ivansenic, line 41: maybe something like `bridgeFallbackEnabled()`, the other one is really hard to get with `-D` or evn vars

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/properties/datastore/configuration/DataStorePropertiesConfiguration.java
```diff
@@ -48,25 +50,51 @@ DataStoreProperties configuration(
...
+      delay = 5,
+      delayUnit = ChronoUnit.SECONDS,
+      maxDuration = 60,
+      durationUnit = ChronoUnit.SECONDS)
+  private Schema.SupportedFeaturesResponse fetchSupportedFeatures(
+      StargateBridgeGrpc.StargateBridgeBlockingStub bridge) {
+    try {
+      return bridge.getSupportedFeatures(Schema.SupportedFeaturesRequest.newBuilder().build());
+    } catch (Exception e) {
+      LOG.warn("Data store metadata fetch using Bridge failed, problem: {}", e.getMessage());
```
ivansenic, line 96: maybe on info level, and warn only if it completely fails

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/properties/datastore/configuration/DataStorePropertiesConfiguration.java
```diff
@@ -48,25 +50,51 @@ DataStoreProperties configuration(
...
+   * Method for fetching data store metadata: will attempt up to 5 retries, for up to 1 minute
+   * before failing.
+   */
+  @Retry(
+      maxRetries = 5,
+      delay = 5,
+      delayUnit = ChronoUnit.SECONDS,
+      maxDuration = 60,
+      durationUnit = ChronoUnit.SECONDS)
+  private Schema.SupportedFeaturesResponse fetchSupportedFeatures(
```
ivansenic, line 91: we should extend the `DataStorePropertiesConfiguration` and add a case where first call fails and second one works..

ivansenic (approved):

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/DataStoreConfig.java
tatu-at-datastax: Yeah, good point. Will change as suggested.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/properties/datastore/configuration/DataStorePropertiesConfiguration.java
tatu-at-datastax: Hmmh. I can see it going either way, but sure I can change this to `info`.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/properties/datastore/configuration/DataStorePropertiesConfiguration.java
tatu-at-datastax: @ivansenic Yes; was able to do the test based on existing mock and some googling on how to override `@Retry` defaults. It is impressive how various things can be overridden.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/properties/datastore/configuration/DataStorePropertiesConfiguration.java
```diff
@@ -48,25 +51,65 @@ DataStoreProperties configuration(
...
+      throw e;
+    }
+  }
+
+  /**
+   * Method for fetching data store metadata: will attempt up to 9 retries, for up to 1 minute
+   * before failing.
+   */
+  @Retry(
+      maxRetries = 9,
```
ivansenic, line 93: is `9` really needed? I mean even `5` seems too high to me..

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/properties/datastore/configuration/DataStorePropertiesConfiguration.java
```diff
@@ -48,25 +51,65 @@ DataStoreProperties configuration(
...
+      delay = 3,
+      delayUnit = ChronoUnit.SECONDS,
+      maxDuration = 60,
+      durationUnit = ChronoUnit.SECONDS)
+  // NOTE: Must NOT be 'private' for annotation to take effect!
+  protected Schema.SupportedFeaturesResponse fetchSupportedFeatures(
+      StargateBridgeGrpc.StargateBridgeBlockingStub bridge, AtomicInteger callCount) {
```
ivansenic, line 100: why the `AtomicInteger` here? log only purpose? I guess this is over engineered :-P

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/api/common/properties/datastore/configuration/BridgeDataStorePropertiesRetryTest.java
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Tests for verifying retry logic of {@link DataStorePropertiesConfiguration#configuration} call.
+ *
+ * <p>NOTE: properties of {@code @Retry} annotation are overridden to both reduce delay between
+ * calls AND to lower max-retry to 2 calls. See "./src/test/resources/application.yaml" for details.
+ * Override logic by Quarkus/SmallRye is explained <a
+ * href="https://quarkus.io/guides/smallrye-fault-tolerance">here</a>.
+ */
+@QuarkusTest
+public class BridgeDataStorePropertiesRetryTest extends BridgeTest {
```
ivansenic, line 35: nit: Oh man understanding this test was so hard for me.. This is why I try not to have de-duplication in the test code.. It's almost impossible to read the tests and go through the code with understanding this easily.. Just imagine how this works:

1. Start at the test case -> `dataStoreWithNoRetriesOk() `
2. OK, just delegates to the method, `dataStoreWithNCalls()`, ok that's at the bottom of the class, navigate, understand parameters, etc
3. And now this calls another method `mockGetSupportedFeaturesCall()` which is now top of the class :exploding_head:

Clear `Given -> When -> Then` structure is way better.. Can you at least kill the `mockGetSupportedFeaturesCall` and push all code to the `dataStoreWithNCalls`, there's really no reason to split this up..

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/api/common/properties/datastore/configuration/BridgeDataStorePropertiesRetryTest.java
```diff
@@ -0,0 +1,109 @@
...
+  @Test
+  public void dataStoreWithTwoRetriesFail() {
+    // Fails if 3 retries needed (test setup only allows 2)
+    try {
+      dataStoreWithNCalls(4);
+      Assertions.fail("Should not have succeeded (max 2 retries)");
+    } catch (Exception e) {
+      assertThat(e).isInstanceOf(StatusRuntimeException.class);
+      assertThat(e.getMessage()).contains("UNAVAILABLE");
+    }
```
ivansenic, line 94:
```
Throwable e = catchThrowable(() ->  dataStoreWithNCalls(4));

assertThat(e).isInstanceOf(StatusRuntimeException.class);
assertThat(e.getMessage()).contains("UNAVAILABLE");
```

ivansenic (approved):

ivansenic:
Formatting fixes required as well..

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/properties/datastore/configuration/DataStorePropertiesConfiguration.java
tatu-at-datastax: Round number like 10,  but mostly that it'd try for closer to 1 minute, but without super long delay in-between.
We can change it of course. Depends on what kind of problem/outage we'd have.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/properties/datastore/configuration/DataStorePropertiesConfiguration.java
tatu-at-datastax: Yes, just for logging. I like seeing how many retries were made, and which call is which.
I don't think the goal is to absolutely minimize code.

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/api/common/properties/datastore/configuration/BridgeDataStorePropertiesRetryTest.java
tatu-at-datastax: Ok, I am sorry I caused you headache :)
I did go through couple of rounds of refactoring, limited by this being bit new code to me (wrt mocking). I am bit surprised that it is so hard to follow, given that amount of code is not super high, but YMMV. But I am sure I can improve ordering, structure based on comments above.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/properties/datastore/configuration/DataStorePropertiesConfiguration.java
tatu-at-datastax: I'll change this back to 5 total calls (4 retries); can tune later on if need be.

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/api/common/properties/datastore/configuration/BridgeDataStorePropertiesRetryTest.java
tatu-at-datastax: @ivansenic Ok here goes again. I hope it's bit more readable now.

ivansenic:
@tatu-at-datastax I already approved this, you can ship it.. No need to wait anymore on my side..

tatu-at-datastax:
@ivansenic yes, I saw approval but since this isn't urgent wanted to give you a chance just in case you wanted to add some more suggestions. Thanks!

PR #2093, opened 2022-09-21
eliminating obsolete extra starctl scripts

jeffreyscarpenter:
Since all HTTP API Services (REST, GraphQL, Docs API) are now on Quarkus, we no longer need the `starctl-service-rest` and `starctl-service-graphql` scripts. Instead, we are favoring container-based usage in our docs. Instructions on running the API services via command line (Maven tasks) are available in the README docs for each API. 

With removal of `starctl-service-rest` and `starctl-service-graphql` scripts, the changes previously made under `v2.0.0` branch to make `starctl` into an umbrella script that starts coordinator and all API services. Therefore this PR also removes the `starctl` umbrella script and renames `starctl-coordinator` back to `starctl`. The coordinator Dockerfile is updated to reference the updated script name. There are no other dependencies, as we had not made any documentation changes discussing the extra `starctl` scripts that are deleted here.

**Checklist**
- [ ] Changes manually tested


ivansenic (approved):
Just to confirm, since git was showing diff a little bit strange, the content from `starctl-cordinator` is only moved to the `starctl` now, right?

Confirmed the startup works in a newly built docker image..

jeffreyscarpenter:
> Just to confirm, since git was showing diff a little bit strange, the content from `starctl-cordinator` is only moved to the `starctl` now, right?

Yes, that's correct

PR #2101, opened 2022-09-27
closes #2085: publish v2 quarkus-common jar to the maven repository

### LINKED ISSUE #2085: Publish `sgv2-quarkus-common` jar to Maven Central
While Stargate v1 publishes all artifacts as jars, even if many/most are not needed (Coordinators), Stargate v2 stops publishing of all APIs as jars: they are published as Docker images which makes more sense. But there is one component, `sgv2-quarkus-common` that is a library and will be needed by anyone implementing Stargate V2 APIs/Services.
It should be published as a jar, to Maven Central to be accessible by external developers.

ivansenic:
**What this PR does**:
Publishes `svg2-quarkus-common` to the Maven repository:

* uses similar setup as the base pom
* however, with the newest plugin versions
* I had to do some gymnastics to enable only `svg2-quarkus-common` in the deploy profile

Fixes #2085

### DIFF
--- a/apis/pom.xml
+++ b/apis/pom.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.stargate</groupId>
   <artifactId>sgv2-api-parent</artifactId>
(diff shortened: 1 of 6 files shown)


jeffreyscarpenter:
Just wondering if publishing the service Jars is such a bad thing, especially if we have to put in so much effort to avoid publishing them?

apis/pom.xml
```diff
@@ -162,5 +174,111 @@
...
+              <serverId>ossrh</serverId>
+              <nexusUrl>https://oss.sonatype.org/</nexusUrl>
+              <autoReleaseAfterClose>true</autoReleaseAfterClose>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
   </profiles>
+  <licenses>
```
tatu-at-datastax, line 266: Maybe move "developers", "licenses" and "scm" up towards beginning of file.

tatu-at-datastax (approved):
LGTM

tatu-at-datastax:
Looks like "Unit tests" fails with:

```
java.lang.RuntimeException: io.quarkus.bootstrap.BootstrapException: Failed to create the application model for io.stargate:sgv2-docsapi::jar:2.0.0-BETA-3-SNAPSHOTnull
```

so something missing/wrong in `pom.xml` or CI.

ivansenic:
> Just wondering if publishing the service Jars is such a bad thing, especially if we have to put in so much effort to avoid publishing them?

Who would ever use this?

ivansenic:
> Looks like "Unit tests" fails with:
> 
> ```
> java.lang.RuntimeException: io.quarkus.bootstrap.BootstrapException: Failed to create the application model for io.stargate:sgv2-docsapi::jar:2.0.0-BETA-3-SNAPSHOTnull
> ```
> 
> so something missing/wrong in `pom.xml` or CI.

Don't get this, I can not reproduce locally.. Doing exactly the same thing as in the CI.. I do assume it's something with the modules, or some caching issues maybe..

apis/pom.xml
ivansenic: where exactly would this fit?

ivansenic:
@tatu-at-datastax seems that the modules in profiles was not working, so I reverted and added a `-pl sgv2-quarkus-common` to the github workflow.. Can you please re-check again?

apis/pom.xml
tatu-at-datastax: Earlier after artifact and group ids, description; before dependencies and properties is typical.

But whatever, it's fine.

tatu-at-datastax:
> @tatu-at-datastax seems that the modules in profiles was not working, so I reverted and added a `-pl sgv2-quarkus-common` to the github workflow.. Can you please re-check again?

Was just looking at CI, seems to pass now.

PR #2107, opened 2022-09-28
Fix #2106: change semantics of `stargate.rest.enableV1` to what we need

### LINKED ISSUE #2106: Remove in-coordinator Docs API implementation, update `stargate.rest.enableV1`
In Stargate V2, all HTTP APIs are moved to stand-alone services, out of coordinator.

Since then we have removed GraphQL implementation from Coordinator, and can now remove Document API as well, having first removed Integration Tests that rely on it.

This still leaves REST API,  and for that we have `stargate.rest.enableV1` system property that was added earlier.
We need to change that a bit to allow for disabling RESTv2 part of API only, leaving RESTv1 as that is still needed for time being (for Stargate V2 deployments; new REST API only implements RESTv2 part of API).

tatu-at-datastax:
**What this PR does**:

See #2106: will basically allow only exposing RESTv1 in-coordinator API, disabling DocsAPI/RESTv2.

Fixes #2106

**Checklist**
- [ ] Changes manually tested


jeffreyscarpenter (approved):

tatu-at-datastax:
Ok, so this failed on Documents API ITs (old) since by default it's no longer included. Once I merge the other PR to kill those tests, this would pass -- same is true if just removing Documents API implementation altogether.

PR #2111, opened 2022-09-29
Fix #2106: remove SGv1 Documents API, change `stargate.rest.enableV1`

### LINKED ISSUE #2106: Remove in-coordinator Docs API implementation, update `stargate.rest.enableV1`

tatu-at-datastax:
**What this PR does**:

* Removes Stargate V1 Documents API implementation!
* Changes semantics of sysprop `stargate.rest.enableV1` to allows us only exposing RESTv1 API of in-Coordinator REST API implementation

Fixes #2106

**Checklist**
- [ ] Changes manually tested


tatu-at-datastax:
@jeffreyscarpenter @EricBorczuk So here's the big removal of Docs API of Stargate V1, instead of changing switch to allow disabling it.

PR #2113, opened 2022-09-29
Switch port numbers for in-coordinator/extracted-drop-wizard cases

tatu-at-datastax:
**What this PR does**:

Fixes port number of in-coordinator REST API for SGv2 to be same as in SGv1 (8082); switches now-obsolete DropWizard-based "new" REST API to 8083 until it can be safely removed. And modifies test base class to use that port number.

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/sgv2-restapi/src/main/resources/config.yaml
+++ b/sgv2-restapi/src/main/resources/config.yaml
@@ -3,9 +3,9 @@ server:
   applicationContextPath: /
   connector:
     type: http
-    # 11-Feb-2022, tatu: For Stargate V1, REST+Docs were both on port 8082, but for Stargate V2
-    #    will separate the two and the "new" REST API inherits 8082
-    port: 8082
+    # 11-Feb-2022, tatu: This implementation will be removed soon, but until then
+    #    has to expose port OTHER than 8082 for ITs to work
+    port: 8083
   requestLog:
     type: external
 stargate:
--- a/restapi/src/main/resources/restapi-config.yaml
+++ b/restapi/src/main/resources/restapi-config.yaml
@@ -3,9 +3,9 @@ server:
   applicationContextPath: /
   connector:
     type: http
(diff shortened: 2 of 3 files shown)


sgv2-restapi/src/main/resources/config.yaml
```diff
@@ -3,9 +3,9 @@ server:
   applicationContextPath: /
   connector:
     type: http
-    # 11-Feb-2022, tatu: For Stargate V1, REST+Docs were both on port 8082, but for Stargate V2
-    #    will separate the two and the "new" REST API inherits 8082
-    port: 8082
+    # 11-Feb-2022, tatu: This implementation will be removed soon, but until then
+    #    has to expose port OTHER than 8082 for ITs to work
+    port: 8083
```
ivansenic, line 8: Why is this change needed? What integration tests would not work? I think this is not needed imo..

ivansenic, line 8: Why is this change needed? What integration tests would not work? I think this is not needed imo..

ivansenic (commented):
This is very confusing to me.. Maybe a quick sync..

tatu-at-datastax:
@ivansenic yes, meeting now.

tatu-at-datastax:
@ivansenic Are we good now? Hoping to just merge this and proceed with eliminating parts as discussed.

sgv2-restapi/src/main/resources/config.yaml
tatu-at-datastax: (discussed in meeting -- is due to DW-based impl not yet removed due to IT dependencies to it -- will be removed once ITs deleted/converted)

jeffreyscarpenter (approved):

ivansenic:
Go for it..

PR #2114, opened 2022-09-29
align base configuration for the quarkus based apis

### LINKED ISSUE #2108: Bridge Metadata access fails during bootstrapping with `java.lang.IllegalStateException: No RESTEasy Reactive request in progress`
With #2087 we no longer default to fallback values if the initial Bridge call to fetch metadata fails. This exposed a seeming error in the implementation of the access call: failure appears to be:

```
2022-09-28 12:42:34,537 ERROR [io.qua.run.Application] (Quarkus Main Thread) Failed to start application (with profile dev): java.lang.IllegalStateException: No RESTEasy Reactive request in progress
	at org.jboss.resteasy.reactive.server.injection.ContextProducers.getContext(ContextProducers.java:138)
	at org.jboss.resteasy.reactive.server.injection.ContextProducers.securityContext(ContextProducers.java:132)
	at org.jboss.resteasy.reactive.server.injection.ContextProducers_Subclass.securityContext$$superforward1(Unknown Source)
[LONG stack trace removed, reproduce with simple `../mvnw quarkus:dev` when Cassandra and Coordinator running]
```

This is odd as none of the tests (including one added as part of PR failed), but this is likely due to:

1. Mocked version having different context/threading and not trigger this issue
2. Existing Integration Tests being configuring to specifically skip metadata access (possibly since at some point functionality was not yet implemented)

At any rate, call fails and needs to be fixed to work.

ivansenic:
**What this PR does**:
Aligns the base configuration of the Quarkus based APIs.. Currently it seems that every API defines completely different props and thus we end up in inconsistent state.. The base props should be more or less the same..

**Issues**:
Closes #2108 

**Checklist**
- [x] @tatu-at-datastax fixes micrometer match patterns (https://github.com/stargate/stargate/issues/2116)


apis/sgv2-restapi/src/main/resources/application.yaml
```diff
@@ -101,13 +132,3 @@ quarkus:
...
-
-
-# In Dev mode, gRPC server usually running on different virtual IP
-"%dev":
-  quarkus:
-    grpc:
-      clients:
-        bridge:
-          host: 127.0.0.2
-          port: 8091
```
ivansenic, line 113: @tatu-at-datastax Removed because this config applies only to your local setup.. You can add `application-dev.properties` for you locally and ignore it with git.. This way you are having a setup only for yourself..

See: https://quarkus.io/guides/config-reference#profile-aware-files

apis/sgv2-graphqlapi/src/main/resources/application.yaml
```diff
@@ -48,6 +59,19 @@ quarkus:
     dev-mode:
       force-server-start: false
 
+  # HTTP settings
+  http:
+    port: ${STARGATE_GRAPHQL_PORT:8080}
```
tatu-at-datastax, line 64: Similar to above, maybe just remove and rely on Quarkus' overrides (I don't think this is used anywhere, docker-compose was changed not to use it)

apis/sgv2-restapi/src/main/resources/application.yaml
```diff
@@ -47,10 +56,13 @@ quarkus:
     # force no grpc server start in dev as we have no grpc service
     dev-mode:
       force-server-start: false
+
+  # HTTP settings
   http:
     # 14-Sep-2022, tatu: Use the legacy port (same as SGv1) for now,
```
tatu-at-datastax, line 62: Can just remove this comment now that we decided to use different ports.

apis/sgv2-restapi/src/main/resources/application.yaml
```diff
@@ -68,10 +80,29 @@ quarkus:
         # ignore all non-application uris, as well as the custom set
         suppress-non-application-uris: true
         ignore-patterns: /,/metrics,/swagger-ui.*,.*\.html
+
+        # due to the https://github.com/quarkusio/quarkus/issues/24938
+        # we need to define uri templating on our own for now
+        # note that order is important
+        # TODO Tatu match-patterns:
```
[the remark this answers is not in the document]
tatu-at-datastax: Ok. We probably want to merge this PR first.

apis/sgv2-restapi/src/main/resources/application.yaml
tatu-at-datastax: Ok. The only thing is that ideally out-of-the-box defaults should work for default setup we use? I was assuming this was the default set up. But if things work without this that's great.

tatu-at-datastax (approved):

apis/sgv2-restapi/src/main/resources/application.yaml
ivansenic: `127.0.0.2`, how is this default?

tatu-at-datastax: It was the setting I saw it in earlier instructions and have used locally, suggesting binding of coordinator that way. Not sure what was the original reason.

So I may be mistaken here; if other APIs not using it then probably REST shouldn't either.

What I meant to ask was just "should it just work for the usual settings/workflow" and I think answer is yes.

apis/sgv2-restapi/src/main/resources/application.yaml
ivansenic: Done.

apis/sgv2-graphqlapi/src/main/resources/application.yaml
```diff
@@ -38,7 +47,9 @@ quarkus:
         expire-after-access: PT5M
         metrics-enabled: true
 
+  # properties for the gRPC clients
   grpc:
+    # all the clients' setup, only bridge
     clients:
       bridge:
         host: ${STARGATE_BRIDGE_HOST:localhost}
```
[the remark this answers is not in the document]
ivansenic: Yes, removing.

apis/sgv2-graphqlapi/src/main/resources/application.yaml
ivansenic: Removing.

ivansenic:
@EricBorczuk Please be aware that `STARGATE_GRAPHQL_PORT`, `STARGATE_BRIDGE_PORT` and `STARGATE_BRIDGE_HOST` are now removed from the config.. They could be used somewhere in the deployment artifacts, you only need to migrate to classicalQuarkus based props as in the docsapi..

PR #2115, opened 2022-09-29
relates to #2108: fix bridge interceptor context check

ivansenic:
**What this PR does**:
Fixes bug in the bridge interceptor.. The inital implementation I did was too weak.. Fixed now with explicit check..

In order to close the whole issue, we need to see what should be the default startup behavior..

Relates #2108

**Checklist**
- [x] @tatu-at-datastax final startup decision


tatu-at-datastax (approved):

tatu-at-datastax:
As per our discussion, just needs changes to defaults so that for %dev mode:

1. Override for 0 retries (need overrides for all APIs' `application.yaml` unless we can specify in `sgv2-quarkus-common` ?)
2. Enable fallback use

Setting for (1) is similar to test YAML override for sgv2-quarkus-common:

```
io:
  stargate:
    sgv2:
      api:
        common:
          properties:
            datastore:
              configuration:
                # Test override(s) for "BridgeDataStorePropertiesRetryTest" to speed up test
                DataStorePropertiesConfiguration/fetchSupportedFeatures/Retry/delay: 0
```

ivansenic:
Adding the dev setup in the #2114, seems more correct to do it there.

PR #2132, opened 2022-10-04
publish to OSSRH not to fail job on error

ivansenic:
**What this PR does**:

In this [release action](https://github.com/stargate/stargate/actions/runs/3161467691/jobs/5149553877), the first publish to OSSRH reported an error, but jars were pushed.. I added `continue-on-error: true` so that next step for pushing quarkus-commons jars runs regardless of the error or not..

### DIFF
--- a/.github/workflows/release-v2.yml
+++ b/.github/workflows/release-v2.yml
@@ -178,8 +178,11 @@ jobs:
         run: |
           cat <(echo -e "${{ secrets.OSSRH_GPG_SECRET_KEY }}") | gpg --batch --import
 
+      # coordinator jars
+      # continue on error as sometimes deploy works even if an error is reported and we want to try next step as well
       - name: Publish coordinator JARs
         if: ${{ !inputs.skipPublish }}
+        continue-on-error: true
         run: |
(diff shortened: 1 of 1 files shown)


tatu-at-datastax (approved):

tatu-at-datastax:
@ivansenic But isn't that error correct? We did publish 2.0.0-BETA-3 last Friday and Maven typically does not allow overwriting releases (releases are immutable). So I don't think it's transient failure.
Or perhaps I misread it?

ivansenic:
After a discussion with @tatu-at-datastax we agreed this is a right approach..

tatu-at-datastax:
I think that publishing still fails, unfortunately.

tatu-at-datastax:
@ivansenic Yes I think so. At first I thought it's odd those are needed, but then thought about it -- they are just basic requirements for every `pom.xml` for artifacts to publish to Maven Central. Which is stricter than installing to local repo etc.

I'll create a PR for this so you can review. Not sure how this could be easily tested, I'll see if I find something.

tatu-at-datastax:
Created #2146. Did not find Maven plug-in for verifying minimum set so maybe we just have to try out publishing. Basic `../mvnw deploy` (which would push snapshot) fails due to missing repo.

PR #2133, opened 2022-10-04
added health extension to the graphql v2

ivansenic:
**What this PR does**:
Adding missing health extension to the graphql v2.. I really don't know how we ended up having different extensions and configuration when we had a clear example on the bootstrapping and could simply use copy-pasta..


tatu-at-datastax (approved):
Oh. Did not realize this did not exist (and hence config change not sufficient). Good catch.

PR #2134, opened 2022-10-04
helm chart configuration for stargate

maheshrajamani:
**What this PR does**:
Added helm chart deployment manifest for stargate.

Fixes #

**Checklist**
- [ X] Changes manually tested


maheshrajamani:
1) Install helm.
2) Add bitnami as repo
    helm repo add bitnami https://charts.bitnami.com/bitnami
3) Start cassandra 4.0 using bitnami/cassandra. This should start cassandra and have storage port 7000 exposed as service "my-release-cassandra-headless"
    helm install my-release bitnami/cassandra
4) On deployment follow the instruction to get the cassandra user's password from the kubernetes secrets.
5) Update cassandra cluster information if needed in helm/stargate/values.yaml. Default will install for 4.0.6
5) Install stargate as below after cassandra cluster is ready from the helm folder.
    cd helm
    helm install stargate stargate
6) Wait for coordinator and service to be Ready.
7) Do portforward the following to test in local:
    kubectl port-forward --namespace default svc/stargate-auth 8081:8081 &
    kubectl port-forward --namespace default svc/stargate-graphqlapi 8080:8080 &
    kubectl port-forward --namespace default svc/stargate-docsapi 8180:8180 &
    kubectl port-forward --namespace default svc/stargate-restapi 8082:8082 &

kathirsvn:
I tested this helm installation and below are my comments:

Set up details:
MacOS, 6 cpu, 8 GB (1GB swap), Docker Desktop, Kubernetes (enabled in Docker Desktop)

- Basic tests of auth, docsapi, restapi and graphqlapi services look good after the installation.
- Tested with the default cassandra version on the distribution (which is 4.0.6)

Issues/suggestions/things to be noted:

- The installation instructions above in the comments could be moved to a README file under helm folder in the repo
- Step 6 on the README above could be bit more clear that we need to clone the repo and then do `cd helm`
- While installing stargate, docsapi and restapi pod were restarted once (we could see the RESTARTS column below). The liveliness/readiness initial delays or retry could be adjusted as discussed to make them not to restart during installation.

```
NAME                                    READY   STATUS    RESTARTS   AGE
my-release-cassandra-0                  1/1     Running   0          12m
stargate-coordinator-7d75785998-j4d5q   1/1     Running   0          8m53s
stargate-docsapi-84fc9596d-hvsv4        1/1     Running   1          8m53s
stargate-graphqlapi-65d88d85c-p5hff     1/1     Running   0          8m53s
stargate-restapi-68c5656cdd-7dmst       1/1     Running   1          8m53s
```

helm/stargate/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "2.0.0"
```
ivansenic, line 24: our tags are always in the form `v2.0.13`, do we need a leading `v` here?

helm/stargate/templates/stargate_authapi_service.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
```
ivansenic, line 1: Few general things:

* can we add a small comment at top of the each yaml file, describing the service/deployment, etc it can help others to understand everything and connect dots
* yaml files should start with `---`, or?

helm/stargate/templates/stargate_coordinator_deployment.yaml
```diff
@@ -0,0 +1,85 @@
...
+        image: "{{ .Values.image.repository311 }}:{{ .Values.image.tag }}" 
+        {{ else if eq .Values.cassandra.clusterVersion "6.8" }}
+        image: "{{ .Values.image.repositoryDse68 }}:{{ .Values.image.tag }}"
+        {{ else }}
+        image: "{{ .Values.image.repository40 }}:{{ .Values.image.tag }}"
+        {{end}}
+        ports:
+        - containerPort: 8091
+        - containerPort: 9042
+        - containerPort: 8081
```
ivansenic, line 29: Some notes here:

* this is missing the public grpc port exposed on the `8090`, it must be exposed, mapped to a service and visible to the outside via the ingress
* the `8091` is the bridge port, this must be exposed, but must not be visible to the outside!

helm/stargate/templates/stargate_coordinator_deployment.yaml
```diff
@@ -0,0 +1,85 @@
...
+        args:          
+         {{  range .Values.arguments }}
+           - {{ . }}
+         {{ end }}
+        readinessProbe:
+          httpGet:
+            path: /checker/readiness
+            port: 8084
+          timeoutSeconds: 20
+          initialDelaySeconds: 80
```
ivansenic, line 39: initial delay of 80 is too long, we need something shorter, like:

 * initial delay 20s
 * delay between checks 5s
 * max checks 20
 
this means you will wait for a maximum of 1.5 mins for the service to be up..

This is also the reason why the API services were restarting, you mark this one as healthly earliest 80s after start, meaning others need to wait at least that for a successful bridge data call..

helm/stargate/templates/stargate_coordinator_deployment.yaml
```diff
@@ -0,0 +1,85 @@
...
+            port: 8084
+          initialDelaySeconds: 80
+          timeoutSeconds: 20
+        resources:
+          limits:
+            cpu: {{ mul 4 (.Values.cpuReqMillicores | default 500) }}m
+            memory: {{add 500 (.Values.heapMB | default 1024)}}Mi
+          requests:
+            cpu: {{ mul 4 (.Values.cpuReqMillicores | default 200) }}m
+            memory: {{add 500 (.Values.heapMB | default 1024)}}Mi
```
ivansenic, line 52: any reason not to have `limits` and `requests` at the same level?

helm/stargate/templates/stargate_coordinator_service.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
```
ivansenic, line 1: this is the coordaintor-bridge service, should be named correctly

helm/stargate/templates/stargate_coordinator_deployment.yaml
```diff
@@ -0,0 +1,85 @@
...
+            -XX:+UnlockDiagnosticVMOptions
+            -XX:GuaranteedSafepointInterval=300000
+            -XX:-UseBiasedLocking
+            -XX:+DebugNonSafepoints
+            -Xss256k
+            -XX:StringTableSize=1000003
+            -XX:+UseG1GC
+            -XX:MaxGCPauseMillis=100
+            -XX:+PrintGC
+            -Djava.awt.headless=true
```
ivansenic, line 85: what about affinity, something like:
```
      # never co-locate two instances on the same node
      affinity:
        podAntiAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            - labelSelector:
                matchExpressions:
                  - key: app
                    operator: In
                    values:
                      - coordinator
              topologyKey: "kubernetes.io/hostname"
```

So replicas are not deployed to the same node.

helm/stargate/templates/stargate_docsapi_deployment.yaml
```diff
@@ -0,0 +1,51 @@
+{{ if .Values.docsapi.enabled }}
```
ivansenic, line 1: same stuff as in the coordinator:

* probes should be faster, why 80s
* limits and requests could be the same
* affinity

helm/stargate/templates/stargate_docsapi_deployment.yaml
```diff
@@ -0,0 +1,51 @@
...
+          initialDelaySeconds: 80
+          timeoutSeconds: 20
+        resources:
+          limits:
+            cpu: {{ mul 4 (.Values.docsapi.cpuReqMillicores | default 500) }}m
+            memory: {{add 500 (.Values.docsapi.memory | default 512)}}Mi
+          requests:
+            cpu: {{ mul 4 (.Values.docsapi.cpuReqMillicores | default 200) }}m
+            memory: {{add 500 (.Values.docsapi.memory | default 512)}}Mi
+        env:
```
ivansenic, line 42: would add `QUARKUS_HTTP_PORT` and fix it to the `8180` just in case we change the port in the app, the deployment continues to work.. or simply set it to the `8080` as a standarad

helm/stargate/templates/stargate_graphql_deployment.yaml
```diff
@@ -0,0 +1,51 @@
+{{ if .Values.graphqlapi.enabled }}
```
ivansenic, line 1: same comments as for the docsapi

helm/stargate/templates/stargate_restapi_deployment.yaml
```diff
@@ -0,0 +1,51 @@
+{{ if .Values.restapi.enabled }}
```
ivansenic, line 1: same comments as for the docs api

helm/stargate/values.yaml
```diff
@@ -0,0 +1,49 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  repository311: "stargateio/coordinator-3_11" 
+  repository40: "stargateio/coordinator-4_0" 
+  repositoryDse628: "stargateio/coordinator-dse-68" 
+  tag: "v2.0.0-BETA-3"
```
ivansenic, line 11: should we change this to `v2`, I think we do have the `v2` tag as well now..

helm/stargate/values.yaml
```diff
@@ -0,0 +1,49 @@
...
+cassandra:
+  clusterName: "cassandra"
+  dcName: "datacenter1"
+  rack: "rack1"
+  seed: "my-release-cassandra-headless" 
+  isDse: null
+  clusterVersion: "4.0"
+  enableCqlAuth: true
+
+cpuReqMillicores: 250
```
ivansenic, line 22: would give less cpu to the APIs then to the coordinator.. Did I understood this correctly is that the final value would be this x4? If so APis can go with the 125 by default imo..

helm/stargate/templates/stargate_coordinator_deployment.yaml
ivansenic: ah OK the diff is only for the CPU, that's fine.. why defaults here differ from the `values.yaml`?

ivansenic (changes requested):
Added few comments.. We are missing the ingress :) this would be deployed nice and then you can not route traffic to it :) Also public gRPC API is not exposed..

helm/stargate/templates/stargate_authapi_service.yaml
maheshrajamani: Instead of adding comment to the yaml file, I will add a readme file.
Not sure why we need ---, if I have multiple resource configuration in the same yaml file i have seen using --- to separate them.

helm/stargate/templates/stargate_docsapi_deployment.yaml
maheshrajamani: Why this port need to be configurable? The appropriate port needs to exposed and the service configuration has to change accordingly? Also that won't align with the documentation of stargate. Thoughts?

ivansenic: I am not saying to be configurable for the users of the chart, just to add it as ENV and then you are sure the port is fixed here in the deployment spec, no matter if we change the port in the application.yaml.. Understand what I mean?

helm/stargate/values.yaml
maheshrajamani: This change is done.

helm/stargate/templates/stargate_coordinator_deployment.yaml
```diff
@@ -0,0 +1,91 @@
...
+        - name: RACK_NAME
+          value: "{{.Values.cassandra.rack}}"
+        - name: SEED
+          value: "{{.Values.cassandra.seed}}"
+        - name: DSE
+          value: "{{.Values.cassandra.isDse}}"
+        - name: CLUSTER_VERSION
+          value: "{{.Values.cassandra.clusterVersion}}"
+        - name: BRIDGE_TOKEN
+          value: "{{.Values.cassandra.bridgeToken}}"
```
ivansenic, line 73: this is not needed, why this?

ivansenic (commented):
I would suggest:

* move this out of draft
* respond to all the comment with done / not done because of

I'll review again and we can proceed forward..

helm/stargate/templates/stargate_coordinator_deployment.yaml
maheshrajamani: Done. As far bridge port visibility, services are configured as ClusterIP so won't be visible outside the cluster.

helm/stargate/templates/stargate_coordinator_deployment.yaml
maheshrajamani: Sync'd up the values.

helm/stargate/templates/stargate_docsapi_deployment.yaml
maheshrajamani: Override of port number not needed to be in sync with stargate documentation.

helm/stargate/templates/stargate_graphql_deployment.yaml
maheshrajamani: Override of port number not needed to be in sync with stargate documentation.

helm/stargate/templates/stargate_restapi_deployment.yaml
maheshrajamani: Override of port number not needed to be in sync with stargate documentation.

helm/stargate/templates/stargate_coordinator_deployment.yaml
maheshrajamani: Done.

helm/stargate/templates/stargate_coordinator_deployment.yaml
```diff
@@ -0,0 +1,89 @@
...
+        livenessProbe:
+          httpGet:
+            path: /checker/liveness
+            port: 8084
+          timeoutSeconds: 10
+          periodSeconds: 10
+          failureThreshold: 5
+          initialDelaySeconds: 300
+        resources:
+          limits:
```
jsanda, line 53: While I think it is good to make limits configurable, it is arguable as to whether it is good to do so at least out of the box. See https://www.reddit.com/r/kubernetes/comments/all1vg/on_kubernetes_cpu_limits/ for details.

helm/stargate/templates/stargate_coordinator_deployment.yaml
```diff
@@ -0,0 +1,89 @@
...
+            cpu: {{ mul 4 (.Values.cpuReqMillicores | default 1000) }}m
```
jsanda, line 54: It seems to be fairly common across charts to define default resources as a yaml block in values.yaml and then in the template it would look something like this:

```
          resources:
            {{- toYaml .Values.resources | nindent 12 }}
```
This provides flexibility for users and it keeps the template logic simpler.

helm/stargate/templates/stargate_authapi_service.yaml
```diff
@@ -0,0 +1,18 @@
...
+kind: Service
+metadata:
+  namespace: {{ .Release.Namespace }}
+  labels:
```
jsanda, line 5: You should consider adding common/recommended labels to all objects that the chart creates. See https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/. 

In k8ssandra we use https://github.com/bitnami/charts/tree/main/bitnami/common to handle this for us. We have another chart we use on top of it called k8ssandra-common.  The `metadata` section in every template in k8ssandra charts has this:

```
metadata:
  name: {{ include "k8ssandra-common.fullname" . }}
  labels: {{ include "k8ssandra-common.labels" . | indent 4 }}

```

helm/stargate/templates/stargate_coordinator_deployment.yaml
```diff
@@ -0,0 +1,89 @@
...
+        ports:
+        - containerPort: 8091
+        - containerPort: 8090
+        - containerPort: 9042
+        - containerPort: 8081
+        args:          
+         {{  range .Values.arguments }}
+           - {{ . }}
+         {{ end }}
+        readinessProbe:
```
jsanda, line 36: I would consider making probes configurable. As I mentioned below with resources, I would simply allow for a yaml block.

helm/stargate/values.yaml
```diff
@@ -0,0 +1,42 @@
+# Default values for stargate.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
```
jsanda, line 5: Consider adding comments for chart properties. I consider values.yaml to be somewhat of an API of sorts so having it well documented is critical IMO.

helm/stargate/templates/stargate_coordinator_deployment.yaml
ivansenic: It's still there, can you push the changes?

ivansenic: Upps, looking at the wrong commit.. Resolving..

helm/stargate/values.yaml
maheshrajamani: Added comments in the values.yaml file, Also added readMe file

helm/stargate/templates/stargate_coordinator_deployment.yaml
maheshrajamani: @ivansenic 
Anti affinity is added to the deployments.
Added ingress for configuration for the http services.

helm/stargate/Chart.yaml
maheshrajamani: Have to revert back as market place expect the app version as semver.

helm/stargate/templates/stargate_coordinator_deployment.yaml
maheshrajamani: Done

helm/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Cassandra installation
+For a quick start, you can install Cassandra in your Kubernetes cluster using Helm.
+
+```shell script
+ helm install my-release bitnami/cassandra
+```
+
+### Autoscaling
+Autoscaling uses metrics server. Metrics server can be installed by executing the command:
```
jeffreyscarpenter, line 16: Does the Autoscaler scale all types of pods including Coordinator and all API Services? Would be good to note which types of pods will be scaled.

jeffreyscarpenter (approved):

kathirsvn (approved):

PR #2138, opened 2022-10-05
disabe vert.x uri validation in the v2 docker images

ivansenic:
**What this PR does**:
The `-Dvertx.disableURIValidation=true` works and disables the URI validation, but can not be defined in the `application.yaml` config.. I had no other idea, then actually adding this to the docker image.. Or should we leave this as optional in the deployment?

Another thing I noticed in the docker images is `JAVA_OPTS` and `JAVA_OPTS_APPEND`. Would defining a `JAVA_OPTS` in the container during deployment actually overwrite what I added? We need to use the `JAVA_OPTS_APPEND` then..

It would be nice to have a test for this, will try to add one maybe.

**Tasks:**

* [ ] Test
* [ ] Final agreement on the option specification

### DIFF
--- a/apis/sgv2-docsapi/src/main/docker/Dockerfile.jvm
+++ b/apis/sgv2-docsapi/src/main/docker/Dockerfile.jvm
@@ -88,6 +88,6 @@ COPY --chown=185 target/quarkus-app/quarkus/ /deployments/quarkus/
 
 EXPOSE 8080
 USER 185
-ENV JAVA_OPTS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
+ENV JAVA_OPTS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager -Dvertx.disableURIValidation=true"
 ENV JAVA_APP_JAR="/deployments/quarkus-run.jar"
 
--- a/apis/sgv2-docsapi/src/main/docker/Dockerfile.legacy-jar
+++ b/apis/sgv2-docsapi/src/main/docker/Dockerfile.legacy-jar
@@ -85,5 +85,5 @@ COPY target/*-runner.jar /deployments/quarkus-run.jar
 
 EXPOSE 8080
 USER 185
-ENV JAVA_OPTS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
+ENV JAVA_OPTS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager -Dvertx.disableURIValidation=true"
 ENV JAVA_APP_JAR="/deployments/quarkus-run.jar"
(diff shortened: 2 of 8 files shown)


ivansenic:
@EricBorczuk can you double check that this works? build a docker and fire a request
@tatu-at-datastax what do you think about this option?

EricBorczuk (approved):

ivansenic:
Need to add this to the native image profile as well

tatu-at-datastax:
@ivansenic So what is the problem this is needed to fix? Is that related to where clause issues Lorina noticed?

It is quite odd that it cannot be defined in `application.yaml`, but if so I guess this is the way to do it.
I wish it wasn't since it's both hidden and then  if running using other mechanisms (start with Maven, or `starctl` etc) need to add option for those separately. But I do not have any better ideas here.

tatu-at-datastax:
One thing I do note is this from https://github.com/quarkusio/quarkus/blob/main/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java: 

```
        /** JVM system property that disables URI validation, don't use this in production. */
        private static final String DISABLE_URI_VALIDATION_PROP_NAME = "vertx.disableURIValidation";
```

so... do we really need and want to use this in production?

ivansenic:
> 

@tatu-at-datastax Let's sync on this please..

tatu-at-datastax (changes requested):
I think I would like to discuss this first to make sure we really want to do it.
I am bit concerned that we would do this to support incorrect usage, allow violating expected escaping.
For example, I think our Postman collections use unescaped curly braces and brackets ({, }, [, ]): this is not legal usage, I think. But was allowed by Jersey in the past.

But I understand that it may be necessary, not so much wrt allowing wrong escaping but avoiding NPEs or 400s with no response. Just want to understand to trade-offs.

tatu-at-datastax:
Ok I think @ivansenic may have gone through it all but just in case I'll add what I found on client side, for Postman.

Looks like they do not want to encode anything but bare minimum of absolutely required characters; that's one way to look at things. So the problem we have has been reported as:

https://github.com/postmanlabs/postman-app-support/issues/8808

but closed as, I think, "wont fix unless there is a specific problem".

There is per-request configuration -- "Encode URL automatically" -- which can be disabled and manually encoding necessary characters (`=`, `{`, `}`) which we could use.
And then there is the question of what `curl` does wrt escaping, if anything.

tatu-at-datastax:
As per discussion, we do want to go with this approach. Changes look good but I need to build locally to validate this works.

tatu-at-datastax:
@ivansenic I tested this and confirmed that a `curl` like:

    curl -i --location --request GET 'localhost:8082/v2/keyspaces/baselines/iot/rows?where=\{\}' --header "X-Cassandra-Token: $AUTH_TOKEN"

* fails from docker image built from `v2.0.0` (with empty 400)
* passes URL validation with docker build from this PR (gives 400 for missing table i.e. hits REST API & persistence)

So I think it works as expected. I tested with C*4 backend should work the same for all.

tatu-at-datastax (approved):

PR #2143, opened 2022-10-06
Add `-r` option for docker-compose scripts for configuring default log level

### LINKED ISSUE #2141: Docker-compose scripts should allow changing default log level (from INFO to DEBUG)
The default logging level for Stargate V2 is `INFO` which is probably good for production (could even consider `WARN` if necessary to prevent extra noise), but for testing it would sometimes be useful to change that level; most likely to `DEBUG`.
I think there is probably Quarkus override to pass as ENV variable, and script should have another switch (cannot use `-l` as that means "use local version"?) for users to pass it.

We could also consider possible defaulting to DEBUG if that makes sense: current output for working set up is not super verbose.

tatu-at-datastax:
**What this PR does**:

Adds `-r` option to override API/Service default log level (INFO) with something else.

Fixes #2141

**Checklist**
- [x] Changes manually tested


tatu-at-datastax:
Note: for draft only added for Cassandra 4 scripts; if approved (and maybe changed) can make similar changes to Cass 3, DSE as well.

One thing I noticed was that DEBUG (same as FINE) level gets output VERY verbose and we might consider overriding some categories to INFO to reduce some of noise (via per-category defaults in `application.yaml`, which overrides global default).

ivansenic (approved):
This looks good, I am approve it already, although we should add same thing for the other versions as well..

Maybe it would also be interesting to enable the `JAVA_OPTS_APPEND` env variable, this way you would be able to pass any setting to the quarkus docker images.. so you could do something like:

`JAVA_OPTS_APPEND="-Dquarkus.log.leve=DEBUG"`

and this would also enable you to add more fine grained logging information for specific classes, but also control any other option and JVM params..

tatu-at-datastax:
@ivansenic yes I think something like that would make sense: switches only for simplest of options.

PR #2146, opened 2022-10-07
Fix #2132 by adding missing properties to parent pom

tatu-at-datastax:
**What this PR does**:

Resolves an issue with  #2132 by adding `name`, `description`, `url` properties so API parent pom can be published

Fixes #2132

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/pom.xml
+++ b/apis/pom.xml
@@ -3,8 +3,11 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.stargate</groupId>
   <artifactId>sgv2-api-parent</artifactId>
+  <name>Stargate V2 API parent pom</name>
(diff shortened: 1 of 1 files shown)


tatu-at-datastax:
Note: was unable to find something to validate what is needed; looked at some other parent poms that do not have this issue and they did seem to have elements that were listed missing here. I think these are just needed for `pom.xml` of every artifact that goes in Maven Central.

jeffreyscarpenter (approved):

tatu-at-datastax:
@ivansenic I knew that Maven Central has validation rules for various fields, but never had issue with parent pom so this was surprising. I think parent poms also inherit some values if they have parents of their own or something
Either way I don't know of any way to really validate what is needed without just trying to publish (release).

PR #2147, opened 2022-10-07
(WIP) Test unescaped curly brace as query parameter (PostMan)

### LINKED ISSUE #2142: Test handling of "where" clause for REST API that does NOT escape curly braces
It looks like PostMan does not escape `{` and `}` characters in query parameters, although it should. This seems to lead to an odd 400 failure with no response with REST API

This needs to be investigated; it probably makes sense to add a new Integration Test.

Note: this requires disabling of URL encoding of query parameters as per:

https://stackoverflow.com/questions/55019246/restassured-disable-url-encoding-not-working-correctly

that is, using:

   given().urlEncodingEnabled(false)....

tatu-at-datastax:
This PR attempts to add a test to reproduce issue with "where" query parameter that is not fully escaped as expected -- this is what older versions of PostMan do.

Unfortunately this version does not reproduce server-side issue; attempts so far include:

1. Trying to use RestAssured but disabling query parameter escaping
2. Use of JDK 11 `HttpClient`, building URI manually

but in both cases it is client that fails before sending, not Server: Java code / JDK really does its best to validate URI and not allow unescaped curly braces be sent.

One more way might using pre-Java-11 http client functionality.

**What this PR does**:

Tries to add test(s) to address 2142

Fixes #2142

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QPostManCompatTest.java
+++ b/apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QPostManCompatTest.java
@@ -1,0 +1,81 @@
+package io.stargate.sgv2.it;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.quarkus.test.common.QuarkusTestResource;
+import io.quarkus.test.common.ResourceArg;
+import io.quarkus.test.common.http.TestHTTPResource;
+import io.quarkus.test.junit.QuarkusIntegrationTest;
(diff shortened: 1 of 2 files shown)


ivansenic:
Would be a good idea to add this commit to the #2138 to ensure the fix there works :)

tatu-at-datastax:
Unfortunate in its current state test fails on client side before sending request.
Will see if I can make it work, would be nice to verify fix, but it's not 100% mandatory (we can verify with Postman obviously).

ivansenic:
@tatu-at-datastax Let's close this.

tatu-at-datastax:
Difficult to do, not critically needed; closing for now.

PR #2149, opened 2022-10-11
Convert DSE-backed REST API tests SG v1->v2

tatu-at-datastax:
**What this PR does**:

Convert existing DSE-specific REST API integration tests from `testing` (SG v1) into new `sgv2-restapi`

N/A

**Checklist**
- [ ] Changes manually tested


tatu-at-datastax:
Finally ready for review; after this will do another PR to remove 2 last remaining tests that run SGv2/REST/DropWizard and we should be able to drop that implementation!

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/persistence/RestApiV2QMaterializedViewIT.java
```diff
@@ -81,9 +81,9 @@ public void getRowsFromMV() throws Exception {
     // 14-Sep-2022, tatu: Not sure why but there is a transient issue here wrt timing.
     //   Locally test does not appear to ever fail, but in CI it does, with error suggesting
     //   MV has not been created or Schema metadata not (yet) updated. In absence of
-    //   better solution, let's try simple wait to give it time; 5 seconds should do it.
+    //   better solution, let's try simple wait to give it time; 3 seconds should do it.
 
-    Thread.sleep(5000L);
+    Thread.sleep(3000L);
```
ivansenic, line 86: better than `Thread.sleep` is to use `Awaitability`, there should be several examples on usage in docs api..

ivansenic (approved):

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/persistence/RestApiV2QMaterializedViewIT.java
tatu-at-datastax: Yes if there was a way to check for condition. Would love to use it but here all I get is weird failure (index not yet created etc, or whatever preceding query tried to create)

tatu-at-datastax: I guess I can change it for this particular case tho as query tested is not too complex.

tatu-at-datastax: Thank you @ivansenic I forgot it's fine to just have the thing fail and not necessarily have condition to wait. Changed as suggested, thanks!

PR #2157, opened 2022-10-14
closes #2131: fixes swagger issues in the docs api v2

### LINKED ISSUE #2131: Swagger in Docs API v2 creates a wrong curl command
We need to investigate further, but I noticed that query parameters are not correctly being mapped to the curl command:

* non defined parameters still appear
* parameters having values that are JSON objects are not encoded correctly
* we should also re-check the token, as it seems not to work once removed (token still sent with `MY_AUTH_TOKEN` value)

Labeling with REST as well, as it could be a problem there too.

ivansenic:
**What this PR does**:
Fixes all Swagger issues in the docs API:

* `fields` and `where` passed as string (JSON), unfortunately layout in Swagger is not perfect
* `document-path` param was sent wrongly, again fallback to string
* fixed example for schema attaching
* minor documentation improvements

Fixes #2131


ivansenic:
@tatu-at-datastax Not sure if this is relevant for the restapi, but you should have a look..

tatu-at-datastax (approved):

PR #2158, opened 2022-10-14
Quarkus log format is fixed to match the format defined in logback.xm…

kathirsvn:
**What this PR does**:
Quarkus log format is fixed to match the format defined in logback.xml in the previous version of apis

**Checklist**
- [X] Changes manually tested

### DIFF
--- a/apis/sgv2-quarkus-common/src/main/resources/application.yaml
+++ b/apis/sgv2-quarkus-common/src/main/resources/application.yaml
@@ -1,4 +1,4 @@
 quarkus:
   log:
     console:
-      format: "%-5p [%t] %d{yyyy-MM-dd HH:mm:ss,SSS} %F:%L - %s%e%n"
+      format: "%-5p [%t] %d{yyyy-MM-dd HH:mm:ss,SSS} %F:%L - %m%n"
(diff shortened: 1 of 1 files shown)


apis/sgv2-quarkus-common/src/main/resources/application.yaml
```diff
@@ -0,0 +1,4 @@
+quarkus:
+  log:
+    console:
+      format: "%-5p [%t] %d{yyyy-MM-dd HH:mm:ss,SSS} %F:%L - %s%e%n"
```
ivansenic, line 4: I don't think this will work? Ohh wow this does work..

ivansenic, line 4: Two things:

* `%s%e` could this simply be `%m` - `Renders the log message plus exception (if any).`
* does `%d{ISO8601}` not working?

ivansenic (approved):

apis/sgv2-quarkus-common/src/main/resources/application.yaml
kathirsvn: Yes, tested all 3 apis (docs, rest, and graphql), the log format change is reflected.

kathirsvn: Sure, changing %s%e to %m now.
ISO8601 didn't work, tested that, so had to put the date format explicitly

PR #2162, opened 2022-10-14
Remove intermediate DropWizard-based SGv2/REST API

tatu-at-datastax:
**What this PR does**:

PR removes DropWizard-based Stargate V2 REST API implementation, since we now have its replacement based on Quarkus. This also requires removes of old integration tests that were using this implementations (and cannot use Quarkus due to project structure) -- there are replacements for all these tests.

In addition formerly disabled RESTv1 test will be re-enabled after being changed to use in-coordinator "old" `restapi` implementation.

NOTE: will NOT remove SGv1 "old" REST API (`restapi`) as we need it for RESTv1 support still.

**Checklist**
- [ ] Changes manually tested


ivansenic (approved):
I am approving, but I am confused a bit with the tests in the `testing` being removed? Are those V2-related tests? Do we have same for V1 still in? Would not like to remove some V1 related tests that might have been moved to V2 in the process.. Maybe we can still use them?

tatu-at-datastax:
One other thing I remembered: at some point the intermediate DW/REST/SGv2 API had Docker build, as part of `./build_docker_images.sh`. But I think (and it looks like) this was removed already before this pr, so there should not be anything left to remove. Same goes for `starctl_` subcommands.

PR #2167, opened 2022-10-17
closes #2163: don't delete docker imgs in failed GH workflows

### LINKED ISSUE #2163: Can not re-run failing V2 API tests due to eager ("always") Docker cleanup
Sometimes one of Stargate V2 API test fails on CI; re-run for just that matrix job would usually be the way to do it.
Unfortunately clean up of Docker images in all cases (even failure):

```
  # runs always, deletes built docker image artifacts
  clean-up:
    name: Clean up
    runs-on: ubuntu-latest
    needs: int-tests
    if: ${{ always() }}
```

means that re-run attempt fails immediately. This can be fixed by only cleaning up Docker images on success.
But additionally it may be necessary to add some retention options to avoid possible accumulation of images on consistent failures.

ivansenic:
**What this PR does**:
Keep built docker images on CI failure, so we can re-run and debug locally..

Fixes #2163

### DIFF
--- a/.github/workflows/apis-v2.yaml
+++ b/.github/workflows/apis-v2.yaml
@@ -139,6 +139,7 @@ jobs:
         with:
           name: img-${{ matrix.image }}-${{ github.sha }}
           path: ${{ matrix.image }}-${{ github.sha }}.tar
+          retention-days: 3
 
   # runs unit tests for the sgv2-docsapi
   build:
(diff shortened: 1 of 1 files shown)


ivansenic:
OK as expected this works fine, docker images deletion skipped. Let me try to re-run the failed job..

ivansenic:
Re-run works as well :partying_face:

tatu-at-datastax (approved):

tatu-at-datastax:
Excellent! Nice that there is specific "retention" tag for automatic discarding.

jeffreyscarpenter (approved):
Nice, elegant change!

Issue #2150, opened 2022-10-12
The `set-output` command deprecated in GitHub actions


ivansenic:
We need to update our CI workflows.

```
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

ivansenic:
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

ivansenic:
Done in #2168 and #2169.

PR #2168, opened 2022-10-17
relates to #2150: removes set-output from the V2 GH workflows

ivansenic:
**What this PR does**:

* removes usage of the `set::output` in V2 workflows
* fixes a small caching issue in the APIs V2 CI


tatu-at-datastax:
An unrelated transient failure of SGv2/REST/C*4 (failed to d/l a dependency), re-running

PR #2178, opened 2022-10-20
Fix to skip printing the token received in the request header in the …

### LINKED ISSUE #2176: Enabling Quarkus access log must not print the access token
Seeing this when enabled:

```
INFO  [vert.x-eventloop-thread-0] 2022-10-20 12:14:56,650 JBossLoggingAccessLogReceiver.java:44 - 10.0.67.99 - [TOKEN_VALUE]:... 20/Oct/2022:12:14:56 +0000 "GET /v2/namespaces/stargate/collections HTTP/1.0" 404 85
```

kathirsvn:
**What this PR does**:

When quarkus http access log is enabled (by setting quarkus.http.access-log.enabled to true), the access log is printed based on the format setting (quarkus.http.access-log.pattern) which is set to the value ‘common’ by default. 

The ‘common’ pattern implies the format of `%h %l %u %t \"%r\" %s %b` where the field %u prints the ‘Remote user that was authenticated’ (Refer: io.quarkus.vertx.http.runtime.filters.accesslog.AccessLogHandler) and token that is passed on the X-Cassandra-Token is printed in the place of this template field. 

By adjusting the access log pattern setting by excluding %u, the token will not be printed in the log any more.

Fixes #2176 

**Checklist**
- [X] Changes manually tested

### DIFF
--- a/apis/sgv2-quarkus-common/src/main/resources/application.yaml
+++ b/apis/sgv2-quarkus-common/src/main/resources/application.yaml
@@ -2,3 +2,6 @@ quarkus:
   log:
     console:
       format: "%-5p [%t] %d{yyyy-MM-dd HH:mm:ss,SSS} %F:%L - %m%n"
+  http:
+    access-log:
+      pattern: "%h %l %t \"%r\" %s %b"


ivansenic (approved):
This works fine.. However, I think we should define this as well:

```
[quarkus.http.access-log.exclude-pattern](https://quarkus.io/guides/http-reference#quarkus-vertx-http-config-group-access-log-config_quarkus.http.access-log.exclude-pattern)

A regular expression that can be used to exclude some paths from logging.
```

And exclude:
* `/metrics`
* `/stargate`

And best is if we could actually use the `${quarkus.http.non-application-root-path}` and `${quarkus.micrometer.export.prometheus.path}`, so the regex could look like:

```
  http:
    access-log:
      pattern: "%h %l %t \"%r\" %s %b"
      exclude-pattern: ${quarkus.http.non-application-root-path}|${quarkus.micrometer.export.prometheus.path}

```

kathirsvn:
> This works fine.. However, I think we should define this as well:
> 
> ```
> [quarkus.http.access-log.exclude-pattern](https://quarkus.io/guides/http-reference#quarkus-vertx-http-config-group-access-log-config_quarkus.http.access-log.exclude-pattern)
> 
> A regular expression that can be used to exclude some paths from logging.
> ```
> 
> And exclude:
> 
> * `/metrics`
> * `/stargate`
> 
> And best is if we could actually use the `${quarkus.http.non-application-root-path}` and `${quarkus.micrometer.export.prometheus.path}`, so the regex could look like:
> 
> ```
>   http:
>     access-log:
>       pattern: "%h %l %t \"%r\" %s %b"
>       exclude-pattern: ${quarkus.http.non-application-root-path}|${quarkus.micrometer.export.prometheus.path}
> ```

Makes sense. Let me test the exclude pattern separately and create a separate PR for that.

PR #2183, opened 2022-10-20
enforce JDK 8

jeffreyscarpenter:
Use the Maven enforcer plugin to ensure usage of JDK 8 to compile.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -390,7 +390,8 @@
             <configuration>
               <rules>
                 <requireJavaVersion>
-                  <version>1.8</version>
+                  <version>[1.8,1.9)</version>
+                  <message>Project requires JDK 8</message>
                 </requireJavaVersion>
                 <requireUpperBoundDeps>
 		  <!-- If specifying includes ONLY those are checked
--- a/pom.xml
+++ b/pom.xml
@@ -389,6 +389,9 @@
             <id>enforce</id>
             <configuration>
               <rules>
+                <requireJavaVersion>
+                  <version>1.8</version>
+                </requireJavaVersion>
                 <requireUpperBoundDeps>
 		  <!-- If specifying includes ONLY those are checked
 		       (default: all)


jeffreyscarpenter:
This will give a fast fail if the user doesn't have the right JDK configured, and will have a better error message than what users would see currently - confusing messages about wrong class file version numbers. If there is agreement this is a good idea, we should do something similar on the v2 `apis` directory to enforce usage of Java 17.

pom.xml
```diff
@@ -389,6 +389,9 @@
             <id>enforce</id>
             <configuration>
               <rules>
+                <requireJavaVersion>
+                  <version>1.8</version>
+                </requireJavaVersion>
```
ivansenic, line 394: can we add same to the APIs, just with `17`?

ivansenic (commented):
This does not work for me, I still get:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.10.1:testCompile (default-testCompile) on project core: Compilation failure
[ERROR] /home/ise/workspace/datastax/stargate/stargate/core/src/test/java/io/stargate/core/activator/DependentService1.java:[18,8] cannot access java.lang.Object
[ERROR]   bad class file: /modules/java.base/java/lang/Object.class
[ERROR]     class file has wrong version 61.0, should be 53.0
[ERROR]     Please remove or make sure it appears in the correct subdirectory of the classpath.
[ERROR] 
```

pom.xml
jeffreyscarpenter: Yes, that would be the idea. This is the PR for the main branch, we'd merge this down to v2 and do another PR to add similar configuration to `apis`

jeffreyscarpenter:
@ivansenic more info? curious what command you executed and what JDK was used, and if you had done a `mvn clean` recently

ivansenic:
> @ivansenic more info? curious what command you executed and what JDK was used, and if you had done a `mvn clean` recently

`./mvnw install` with `JAVA_HOME` pointing to the Java 17

tatu-at-datastax:
@jeffreyscarpenter I think I know why. as per:

https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html

what you have means "1.8 or higher". To make it "1.8 and nothing else" need to use range:

     [1.8]

jeffreyscarpenter:
> what you have means "1.8 or higher". To make it "1.8 and nothing else" need to use range: `[1.8]`

Turns out that means exactly 1.8 only, with no allowance for any sub-version numbers. so I ended up having to do this: `[1.8,1.9)` which means greater than or equal to 1.8 but less than 1.9. This is not well documented on the enforcer site so I actually found it here: https://stackoverflow.com/questions/62465338/how-to-enforce-compilation-with-a-given-jdk-in-maven

ivansenic (approved):
Works now:

```
[WARNING] Rule 0: org.apache.maven.plugins.enforcer.RequireJavaVersion failed with message:
Project requires JDK 8
```

Just don't forget to add same for the APIs

tatu-at-datastax:
> > what you have means "1.8 or higher". To make it "1.8 and nothing else" need to use range: `[1.8]`
> 
> Turns out that means exactly 1.8 only, with no allowance for any sub-version numbers. so I ended up having to do this: `[1.8,1.9)` which means greater than or equal to 1.8 but less than 1.9. This is not well documented on the enforcer site so I actually found it here: https://stackoverflow.com/questions/62465338/how-to-enforce-compilation-with-a-given-jdk-in-maven

Good catch. Yeah, version ranges in Maven are interesting and need some trial-and-error testing....

Issue #2179, opened 2022-10-20
Bridge service is not handling runtime exceptions


ivansenic:
Bridge service currently silently ignores the `RuntimeExceptions`, if they occur in the methods that are not handled with the `MessageHandlers`. Basically:

* `executeQueryWithSchema`
* `describeKeyspace`

The proper try/catch must be included, so that any runtime exception is handled. Current situation is that client side hangs, because observer on the server side is not informed on the error.

We need to catch that hang in test first. This can be easily done, mock the `Persistence` and for example throw exception on `persistence.decorateKeyspaceName`..

In addition to this, these exceptions should be logged. Also we should log possibly all the exceptions that we catch in the `io.stargate.bridge.service.ExceptionHandler`, which is not done right now.

ivansenic:
The exception being thrown is the NPE, due to the https://github.com/stargate/stargate/issues/2181

tatu-at-datastax:
Looks like the issue with `describeKeyspace()` is that `SchemaHandler` does not catch `RuntimeException`.
It needs to; and should also use `ExceptionHandler` to properly map exception to desired error to return to client (and log if need be).

But `executeQueryWithSchema()` may already be handling things properly if I am not mistaken: I will try to add a test to verify this.

PR #2184, opened 2022-10-21
closes #2181: bridge context key not to be used in another thread

### LINKED ISSUE #2181: Bridge service uses the `Context.Key` without context propagation
The `SchemaHandler` is accessing the context key defined in the `BridgeService`:

https://github.com/stargate/stargate/blob/dbc41c04d8bb545e5d8a259b80b430ab2015e02e/bridge/src/main/java/io/stargate/bridge/service/SchemaHandler.java#L62-L63

However, the bridge service is executing that `describeKeyspace` in another executor and the context is not passed over to new thread.

ivansenic:
**What this PR does**:
Fixes grpc context access in executor.. Reproduced first with the updated test, then added a fix..

Unfortunately our tests around the `Bridge` are not good enough, and we might need to revisit those asap..

Fixes #2181


tatu-at-datastax (approved):

tatu-at-datastax:
@ivansenic Will merge to get in before my work on #2179.

PR #2187, opened 2022-10-21
Fix handling of `describeKeyspace` to catch RuntimeEcxeptions too, add a test

### LINKED ISSUE #2179: Bridge service is not handling runtime exceptions

tatu-at-datastax:
**What this PR does**:

Improves bridge handling so that:

* `describeKeyspace()` operation will catch all `Throwable`s and report them properly (not just gRPC fails)
* `queryWithSchema()` operation that fails internally will also report failure gracefully (add test(s))

Fixes #2179

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/bridge/src/main/java/io/stargate/bridge/service/ExceptionHandler.java
+++ b/bridge/src/main/java/io/stargate/bridge/service/ExceptionHandler.java
@@ -43,9 +43,9 @@ public class ExceptionHandler {
   static final Metadata.Key<QueryOuterClass.CasWriteUnknown> CAS_WRITE_UNKNOWN_KEY =
       ProtoUtils.keyForProto(QueryOuterClass.CasWriteUnknown.getDefaultInstance());
 
-  private final StreamObserver<QueryOuterClass.Response> responseObserver;
+  private final StreamObserver<?> responseObserver;
 
-  public ExceptionHandler(StreamObserver<QueryOuterClass.Response> responseObserver) {
+  public ExceptionHandler(StreamObserver<?> responseObserver) {
     this.responseObserver = responseObserver;
   }
 
@@ -68,8 +68,9 @@ public class ExceptionHandler {
    * StreamObserver#onError(Throwable)} method.
    */
   public void handleException(Throwable throwable) {
-    if (throwable instanceof CompletionException
-        || throwable instanceof MessageHandler.ExceptionWithIdempotencyInfo) {
+    if ((throwable instanceof CompletionException
+            || throwable instanceof MessageHandler.ExceptionWithIdempotencyInfo)
+        && (throwable.getCause() != null)) {
       handleException(throwable.getCause());
     } else if (throwable instanceof StatusException
         || throwable instanceof StatusRuntimeException) {
--- a/bridge/src/test/java/io/stargate/bridge/service/SchemaOperationsTest.java
+++ b/bridge/src/test/java/io/stargate/bridge/service/SchemaOperationsTest.java
@@ -264,9 +264,6 @@ public class SchemaOperationsTest extends BaseBridgeServiceTest {
     StargateBridgeGrpc.StargateBridgeBlockingStub stub = makeBlockingStub();
     when(persistence.decorateKeyspaceName(any(String.class), any())).thenReturn("my_stuff");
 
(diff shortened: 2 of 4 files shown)


tatu-at-datastax:
Main fix done, still considering whether more failures should be logged on bridge server side (probably?).

bridge/src/main/java/io/stargate/bridge/service/ExceptionHandler.java
```diff
@@ -43,9 +43,9 @@ public class ExceptionHandler {
   static final Metadata.Key<QueryOuterClass.CasWriteUnknown> CAS_WRITE_UNKNOWN_KEY =
       ProtoUtils.keyForProto(QueryOuterClass.CasWriteUnknown.getDefaultInstance());
 
-  private final StreamObserver<QueryOuterClass.Response> responseObserver;
+  private final StreamObserver<?> responseObserver;
```
tatu-at-datastax, line 46: Parameterization not used for anything. Could have changed it to something like `GeneratedMessageV3` (common base class for protobuf-generated message classes) but does not seem valuable.

tatu-at-datastax:
@ivansenic Aside from main changes didn't yet add more logging of failures: feel free to change if you want, as we need this to be merged for build. Or I can change it when I get online. And we can always change it later on either way.

bridge/src/test/java/io/stargate/bridge/service/SchemaOperationsTest.java
```diff
@@ -253,4 +256,30 @@ public void schemaTableWithMaterializedView() {
...
+
+  @Test
+  @DisplayName("Handle and propagate server-side NPE appropriately")
+  public void schemaTableWithServerSideNPE() {
+    // Given
+    StargateBridgeGrpc.StargateBridgeBlockingStub stub = makeBlockingStub();
+    when(persistence.decorateKeyspaceName(any(String.class), any())).thenReturn("my_stuff");
+
+    // Only need minimal schema for bootstrapping
+    io.stargate.db.schema.Schema schema = io.stargate.db.schema.Schema.build().build();
```
ivansenic, line 268: Codacy complains here and I guess is correct this time:

> io.stargate.db.schema.Schema schema = io.stargate.db.schema.Schema.build().build();

bridge/src/main/java/io/stargate/bridge/service/ExceptionHandler.java
ivansenic: Leave it.. seems enough for exception handling

ivansenic (changes requested):
We are missing the protection in the `io.stargate.bridge.service.BridgeService#executeQueryWithSchema`, exactly same issues can occur there if for example runtime exception occurs with persistence interaction, calling io.stargate.bridge.service.SchemaHandler#buildKeyspaceDescription, etc

Furthermore, the code in current state is complete spaghetti code. There is absolutely no rules on which components do what, separation of concerns, etc.. Let's solve this quickly and then let's have a ticket to rewrite this, including tests..

ivansenic (approved):
Approving, @tatu-at-datastax will submit another PR for other fixes..

bridge/src/test/java/io/stargate/bridge/service/SchemaOperationsTest.java
tatu-at-datastax: > Codacy complains here and I guess is correct this time:
> 
> > io.stargate.db.schema.Schema schema = io.stargate.db.schema.Schema.build().build();

Ah. Yes, will remove. Copy-paste flag. For once Codacy was right :)

Issue #1786, opened 2022-04-19
Add Docs as source API in V2


ivansenic:
We should extend the protos and add Docs as an additional source API.

@olim7t Any chance you can quickly tackle this, should be easy for you.

olim7t:
It's not as simple as adding a constant: the gRPC enum maps to `io.stargate.auth.SourceAPI` in the `authnz` module. In v1 the docs API is using `REST`. If we add a separate constant, there will probably also be impacts in authorization backends.

ivansenic:
@olim7t So should we actually go all the way and add this as a new constant? Sure for now we can simply use the `REST` as up until now, but in long term should we split this? Is this being recorded in some type of auditing logs?

olim7t:
It's not only used for logging, I think it's also an integral part of permission definitions. For example a token might be authorized to do certain operations via the GraphQL API, but not via the Documents API.

Adding the new constant to Stargate is a couple of lines of code. But the other aspect is whether the authorization service will understand what the constant means. Admittedly that's more of a deployment concern (Astra in our case), so maybe we should move the conversation to a private channel.

PR #2190, opened 2022-10-24
closes #2175: apis to report source api in the grpc metadata

### LINKED ISSUE #2175: API V2 bridge communication has wrong Source API
Getting this when trying to use the REST API V2, without cql scoped token:

```
{
  "description": "Unauthorized gRPC operation (Status.Code.PERMISSION_DENIED->401): PERMISSION_DENIED: Missing correct permission on stargate.rest_example_products: Not authorized to access stargate.rest_example_products for CREATE (from source api: cql)",
  "code": 401,
  "grpcStatus": "PERMISSION_DENIED",
  "internalTxId": null
}
```

It's either that:

* REST API does not define a source API in the default parameters
* Not using the default parameters
* Or bridge does not consider the passed source API for the auth

ivansenic:
**What this PR does**:
* adds source api as the metadata when communicating to the bridge
* adds bridge handling to read those values
* updates query handlers to support this for:
   * DSE 
   * C* 4
   * C* 3.11

Fixes #2175.

**Checklist**
- [x] Bridge changes
- [x] Query handler changes


apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/SourceApiConfiguration.java
```diff
@@ -0,0 +1,32 @@
...
+import javax.enterprise.context.ApplicationScoped;
+import javax.ws.rs.Produces;
+
+@ApplicationScoped
+public class SourceApiConfiguration {
+
+  @Produces
+  @SourceApi
+  public String sourceApi() {
+    return "rest";
```
tatu-at-datastax, line 30: Probably should be "docs"?

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/GrpcMetadataResolver.java
```diff
@@ -19,15 +19,35 @@ public class GrpcMetadataResolver {
   /** Metadata key for passing the cassandra token to the Bridge. */
   private final Metadata.Key<String> cassandraTokenKey;
 
+  /** Default metadata. */
+  private final Metadata defaultMetadata;
+
+  // TODO add validation @Pattern(regexp = "rest|graphql")  after
```
tatu-at-datastax, line 25: We probably either want to just validate legal characters, or make set of acceptable ids configurable, to allow addition of new APIs by other developers?
If it was configurable we could add our own regex for production deployment to ensure no unknown ids were to be used.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/GrpcMetadataResolver.java
```diff
@@ -19,15 +19,35 @@ public class GrpcMetadataResolver {
...
   @Inject
-  public GrpcMetadataResolver(GrpcMetadataConfig config) {
+  public GrpcMetadataResolver(GrpcMetadataConfig config, @SourceApi String sourceApi) {
     this.tenantIdKey = Metadata.Key.of(config.tenantIdKey(), Metadata.ASCII_STRING_MARSHALLER);
     this.cassandraTokenKey =
         Metadata.Key.of(config.cassandraTokenKey(), Metadata.ASCII_STRING_MARSHALLER);
+
+    // default metadata includes source api
+    Metadata.Key<String> sourceApiKey =
+        Metadata.Key.of(config.sourceApiKey(), Metadata.ASCII_STRING_MARSHALLER);
```
tatu-at-datastax, line 35: I know short-living garbage is not a big problem in general but couldn't we just create this key once (add a convenience method in `GrpcMetadataConfig` to get pre-created `Metadata.Key` instead of `String` for creating one) and reuse it? `Metadata.Key` is annotated as Immutable and hence thread-safe.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/StargateBridgeInterceptor.java
```diff
@@ -35,18 +35,22 @@ public class StargateBridgeInterceptor implements ClientInterceptor {
...
+    Metadata metadata = metadataResolver.getDefaultMetadata();
+
     // StargateRequestInfo is request scoped bean
     // there is no better way to know if if's available
     // as we can intercept outside request
     // thus ensure context is active and has instances
     InjectableContext.ContextState contextState =
         Arc.container().requestContext().getStateIfActive();
-    if (null == contextState || !hasStargateRequestBean(contextState.getContextualInstances())) {
-      return next.newCall(method, callOptions);
```
tatu-at-datastax, line 45: So this call would not pass default metadata (for now mostly SourceApi) so need to change to pass it?

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/SourceApiTestConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+import javax.ws.rs.Produces;
+
+/** Test config for the source api, so common can boot itself. */
+@ApplicationScoped
+public class SourceApiTestConfiguration {
+
+  @Produces
+  @SourceApi
+  public String sourceApi() {
+    return "rest";
```
tatu-at-datastax, line 31: For tests could perhaps use a distinct value, separate from values we use in prod?
Unless we have validation settings that prevent use of id like "test" in which case never mind.

auth-table-based-service/src/test/java/io/stargate/auth/SourceAPITest.java
```diff
@@ -0,0 +1,86 @@
...
+            Map<String, ByteBuffer> map = new HashMap<>();
+            SourceAPI.GRAPHQL.toCustomPayload(map);
+
+            SourceAPI result = SourceAPI.fromCustomPayload(map, SourceAPI.CQL);
+
+            assertThat(result).isEqualTo(SourceAPI.GRAPHQL);
+        }
+
+        @Test
+        public void emptyMap() {
```
tatu-at-datastax: I know it's implicit but maybe add a comment stating this verifies that our default SourceAPI is CQL?

authnz/src/test/java/io/stargate/auth/SourceAPITest.java
```diff
@@ -0,0 +1,79 @@
...
+      Map<String, ByteBuffer> map = new HashMap<>();
+      SourceAPI.GRAPHQL.toCustomPayload(map);
+
+      SourceAPI result = SourceAPI.fromCustomPayload(map, SourceAPI.CQL);
+
+      assertThat(result).isEqualTo(SourceAPI.GRAPHQL);
+    }
+
+    @Test
+    public void emptyMap() {
```
tatu-at-datastax, line 62: I know it's implicit here, but maybe add a comment that this verifies that the default `SourceAPI` value, if nothing explicitly passed, is CQL?

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/SourceApi.java
```diff
@@ -0,0 +1,29 @@
...
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import javax.inject.Qualifier;
+
+@Documented
+@Retention(RUNTIME)
+@Qualifier
+public @interface SourceApi {}
```
tatu-at-datastax, line 29: Not sure what to do about this (if anything), but having 2 different `SourceApi`s (one enum in auth, for valid values for our auth; another this tag interface for config) is bit confusing.
Maybe this annotation could be `SourceApiDefinition` or something?

authnz/src/main/java/io/stargate/auth/SourceAPI.java
```diff
@@ -17,11 +17,19 @@
...
 package io.stargate.auth;
 
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
 public enum SourceAPI {
```
tatu-at-datastax, line 27: I know it didn't exist before, but Javadoc here would be nice; could mention that value is passed along as custom metadata (for CQL) or headers (bridge/grpc).

authnz/src/main/java/io/stargate/auth/SourceAPI.java
```diff
@@ -31,4 +39,27 @@ public enum SourceAPI {
...
+
+  public static SourceAPI fromCustomPayload(Map<String, ByteBuffer> customPayload, SourceAPI defaultIfMissing) {
+    if (null == customPayload) {
+      return defaultIfMissing;
+    }
+
+    return Optional.ofNullable(customPayload.get(CUSTOM_PAYLOAD_KEY))
+            .flatMap(buffer -> {
+              String decode = StandardCharsets.UTF_8.decode(buffer).toString();
+              return Arrays.stream(SourceAPI.values()).filter(s -> Objects.equals(s.getName(), decode)).findFirst();
```
tatu-at-datastax: Why not just use `SourceAPI.valueOf(decode)`? To avoid exception for mismatching name?

testing/src/main/java/io/stargate/it/bridge/BridgeIntegrationTest.java
```diff
@@ -137,4 +145,14 @@ protected QueryOuterClass.QueryParameters.Builder queryParameters(
   protected QueryOuterClass.QueryParameters.Builder queryParameters(CqlIdentifier keyspace) {
     return queryParameters(keyspace, false);
   }
+
+  protected static String sourceApi() {
```
tatu-at-datastax, line 149: Since these are static and not meant for overridability, could just inline `sourceApi` value and interceptor closure.

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/SourceApiConfiguration.java
tatu-at-datastax, line 30: Looks like "rest" is used here because `AuthApi` has no enum for docs-api.
Might want to add a comment mentioning this since I first thought this was unintentional.

ivansenic: We don't have docs as the source API.. Historically, for the auth the docs and rest were grouped into `rest`.. 

There is a ticket for this, but this requires authz service implementations to respect the docs as the new source api: https://github.com/stargate/stargate/issues/1786

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/GrpcMetadataResolver.java
ivansenic: Does not make sense to be configurable, because it should no be allowed that somebody changes the source API for the REST to `graphql`.. Adding a new value is another issue and quite complicated at the moment..

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/GrpcMetadataResolver.java
ivansenic: Don't get this, this is a constructor, called only one time during init? Does exactly that by creating a single instance of the default metadata..

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/StargateBridgeInterceptor.java
ivansenic: Yup, at least sending the default metadata..

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/SourceApiTestConfiguration.java
ivansenic: But what would we gain with this? This was made so we can start the commons context in the tests, basically it's not used nor asserted anywhere..

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/SourceApi.java
ivansenic: OK.

1. Renaming this to `@SourceApiQualifier`, because it is a `Qualifier`
2. Added https://github.com/stargate/stargate/issues/2195 to remove from schema auth

authnz/src/main/java/io/stargate/auth/SourceAPI.java
ivansenic: Because source api has the field `name` that I am comparing to, would not work..

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/SourceApiConfiguration.java
ivansenic: See comment above..

tatu-at-datastax: (note: nope, since there's no matching Enum entry -- Docs API just uses REST API authz rights)

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/GrpcMetadataResolver.java
tatu-at-datastax: @ivansenic Ok sorry. Missed that context. N/m.

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/SourceApiTestConfiguration.java
tatu-at-datastax: I was thinking it'd avoid test accidentally passing. If and when not tested yeah never mind.

And since we'd need matching Enum (which I did not realize when adding this comment) not worth it either.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/SourceApi.java
tatu-at-datastax: Cool, makes sense.

authnz/src/main/java/io/stargate/auth/SourceAPI.java
tatu-at-datastax: Ah. True.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/GrpcMetadataResolver.java
tatu-at-datastax: Yeah, sorry, should have removed comment after I realized complexity of enum mapping etc.
You are absolutely right.
There is the bigger question on extensibility, but this PR is not the place for that.

PR #2194, opened 2022-10-24
Second part of fix to #2179: handle "executeQueryWithSchema()" too

### LINKED ISSUE #2179: Bridge service is not handling runtime exceptions

tatu-at-datastax:
**What this PR does**:

Adds graceful handling of RuntimeExceptions too for "BridgeService.executeQueryWithSchema()"

Fixes #2179

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
+++ b/bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
@@ -91,64 +91,61 @@ public class BridgeService extends StargateBridgeGrpc.StargateBridgeImplBase {
   public void executeQueryWithSchema(
       Schema.QueryWithSchema request,
       StreamObserver<Schema.QueryWithSchemaResponse> responseObserver) {
-    String keyspaceName = request.getKeyspaceName();
-    int keyspaceHash = request.getKeyspaceHash();
-    String decoratedName;
-    Keyspace keyspace;
-
     try {
-      decoratedName =
+      String keyspaceName = request.getKeyspaceName();
+      int keyspaceHash = request.getKeyspaceHash();
+      String decoratedName =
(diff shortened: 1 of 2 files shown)


bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
```diff
@@ -93,9 +93,17 @@ public void executeQueryWithSchema(
...
+    Keyspace keyspace;
+
+    try {
+      decoratedName =
+          persistence.decorateKeyspaceName(keyspaceName, BridgeService.HEADERS_KEY.get());
+      keyspace = persistence.schema().keyspace(decoratedName);
+    } catch (Throwable t) {
+      new ExceptionHandler(responseObserver).handleException(t);
+      return;
+    }
```
ivansenic, line 106: I am gonna include the whole method with try.catch, no need just to protect that first part.

ivansenic (approved):

bridge/src/main/java/io/stargate/bridge/service/BridgeService.java
ivansenic: Most likely also needs rebase after #2190 is merged

PR #2203, opened 2022-11-02
Allow legacy flush to be used with property

tjake:


ivansenic (approved):

ivansenic:
@tatu-at-datastax I merged this to `master`, but I see there is `main` now, how is the sync working here at the moment?

tatu-at-datastax:
@ivansenic Same as before, just using the new name: merge to `main` just like we used to merge to `v2.0.0`. So rename does not change logic; nor will change of default branch after renames.

PR #2193, opened 2022-10-24
enforce JDK 17+ for building SGv2 APIs

jeffreyscarpenter:
Use maven enforcer plugin to ensure APIs are compiled against JDK 17 or later.

**Checklist**
- [x] Changes manually tested
- [ ] Automated Tests added/updated
- [ ] Documentation added/updated
- [x] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

PR #2205, opened 2022-11-03
Help troubleshoot #2193 (IAE for null proxy message) by explicit arg validation

tatu-at-datastax:
**What this PR does**:

Adds explicit argument validation to avoid generic IllegalArgumentException from `InetSocketAddress` constructor; add information on faulty message. This is to help troubleshoot production failures we observed from logs.

N//A

**Checklist**
- [ ] Changes manually tested


cql/src/main/java/org/apache/cassandra/stargate/transport/internal/HAProxyProtocolDetectingDecoder.java
```diff
@@ -87,4 +74,27 @@ private static boolean match(byte[] prefix, ByteBuf buffer, int idx) {
     }
     return true;
   }
+
+  static class InboundHAProxyHandler extends SimpleChannelInboundHandler<HAProxyMessage> {
```
tatu-at-datastax, line 78: Extracted into separate class to create a stateless singleton, to avoid allocating instance per call. Not a big deal but seemed pointless to create throw-away instances. Main change is the validation.

maheshrajamani (approved):

Issue #1490, opened 2021-12-15
Rename `master` branch to `main`


jeffreyscarpenter:
Rename the `master` branch to `main` for diversity and inclusion reasons.

jeffreyscarpenter:
See also: https://github.com/stargate/stargate/discussions/1489

tatu-at-datastax:
So, renaming as suggested (current `master` renamed as `v1`; current `v2.0.0` renamed as `main`) is what we plan to do. This will require local changes to Stargate project, as well as likely changes to C2 sister-project. This includes:

* [TO BE WRITTEN]

jeffreyscarpenter:
@tatu-at-datastax I've created draft PRs for changes to docs and website to reflect these new branch names.

tatu-at-datastax:
I added a separate google doc with details; unfortunately this requires changes to non-OSS parts as well so can't link from this issue.

tatu-at-datastax:
Progress so far:

* `main` points to Stargate v2 (former `v2.0.0`)
* `v1` points to Stargate v1 (former `master`)

Yet to do:

* Change default branch to be `main` -- is currently `v1`

tatu-at-datastax:
Change the default branch to be `main`, task completed.

PR #2207, opened 2022-11-04
Changes to CI jobs related to #1490, first part: renaming of v2.0.0->…

tatu-at-datastax:
**What this PR does**:

Changes CI job references from "v2.0.0" to "main", as part of v1/v2 branch renaming (see #1490)

Part of fixes for #1490

**Checklist**
- [x] Changes manually tested (via CI run)

### DIFF
--- a/.github/workflows/coordinator-test.yml
+++ b/.github/workflows/coordinator-test.yml
@@ -1,4 +1,4 @@
-name: CI IT Experiment
+name: Coordinator Tests
 
 on:
   workflow_dispatch:
--- a/.github/workflows/apis-v2.yaml
+++ b/.github/workflows/apis-v2.yaml
@@ -2,13 +2,13 @@
 name: Stargate APIs V2
 
 # runs on
-# * pushes and pull requests on the v2.0.0.
+# * pushes and pull requests on the "main" (Stargate v2)
 # * manual trigger
 on:
   push:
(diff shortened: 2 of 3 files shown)


.github/workflows/coordinator-test.yml
```diff
@@ -1,4 +1,4 @@
-name: CI IT Experiment
+name: Coordinator Tests
```
tatu-at-datastax, line 1: Can define a better name but these are tests for main-level projects; for v1 it's all components (since APIs part of Coordinator), for v2 just Coordinator tests (cql, grpc)

ivansenic (approved):
Seems fine, not sure if we need any other changes in the dependent projects?

.github/workflows/coordinator-test.yml
jeffreyscarpenter: I like this name

tatu-at-datastax:
> Seems fine, not sure if we need any other changes in the dependent projects?
> 
> I also think we need to check the V1 related items and make sure that this works now from the `v1` branch. For example, I think the google based CI will not work now. We can do this in a separate PR.

Good question on other projects -- I think c2 is fine and cndb should not have deps. But there might be others. I'll need to see if docker-images repo has a dependency.

I think this first part is safer than renaming of actual `master`.

PR #2210, opened 2022-11-07
(TESTING) Rebuild branch to verify "main" works after merge from "master"

tatu-at-datastax:
**DO NOT MERGE!!!**

Temporary PR for tracking merging of commits from "v1"/"master" to "main" (Stargate V2) since our CI does not yet protect latter and can theoretically allow merging of breaking changes.


tatu-at-datastax:
Now that `main` is protected, no longer needed.

PR #2213, opened 2022-11-08
Convert SGv2/REST to Async handling with Quarkus/Mutiny

### LINKED ISSUE #1599: Change SGv2/REST api gRPC client from blocking to non-blocking/async operation
Initial implementation of REST API for Stargate V2 uses blocking gRPC client to talk to the gRPC backend ("bridge").
This could be changed to non-blocking variant to improve scaling for higher concurrency cases.

Whether work should be done before or after our framework change (see #1526) is an open question, and depends both on effort involved (is there much more work to do before change and/or lots of rework) and resources available (if someone has time to work on this before framework task gets started).

tatu-at-datastax:
**What this PR does**:

Converts REST API / SGv2 from blocking to async handling using Quarkus Mutiny-based async rest framework.

Fixes #1599

**Checklist**
- [x] Changes manually tested


ivansenic:
@tatu-at-datastax hint: I think it's best to start with refactoring a single (simplest) resource, you would then get a very clear idea how would it look at the end.

tatu-at-datastax:
@ivansenic yes, that is the idea. Not sure which one is the simplest but I plan to do what you suggest; verify one method works with IT etc. Started first by sort of unrolling use of StargateBridgeClient, will continue with that.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2UDTsResourceImpl.java
```diff
@@ -119,86 +119,88 @@ public Response createType(final String keyspaceName, final String udtAddPayload
...
           new QueryBuilder()
               .alter()
               .type(keyspaceName, typeName)
               .renameColumn(columnRenames)
               .build();
-      bridge.executeQuery(query);
+      if (addQuery != null) {
+        return executeQueryAsync(addQuery)
+            .chain(any -> executeQueryAsync(renameQuery))
+            .map(any -> RestResponse.ok());
```
tatu-at-datastax, line 187: Note: MUST use opened-up form and NOT `RestResponse::ok` -- latter will fail Integration Test for some reason that I do not understand.

tatu-at-datastax: Ah. Realized why -- `RestResponse:ok` would take in `any` as opposed to 0-arg version not taking any payload. And the problem is that returned query results are not JSON serializable.
So no mystery there.

ivansenic:
@tatu-at-datastax Great, ping me if you need help, or when you need a review..

tatu-at-datastax:
@ivansenic This is now ready for review, passes all tests, now unresolved issues wrt conversion (that I know of).
But there may be some issues or improvements I can make so interested in your feedback.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/Sgv2RowsResourceImpl.java
```diff
@@ -184,26 +184,25 @@ public Response createRow(
...
+
+              try {
+                return buildAddRowQuery(keyspaceName, tableName, payloadMap, toProtoConverter);
+              } catch (IllegalArgumentException e) {
+                throw new WebApplicationException(
+                    "Invalid path for row to create, problem: " + e.getMessage(),
+                    Status.BAD_REQUEST);
+              }
+            })
+        .map(any -> RestResponse.status(Response.Status.CREATED, payloadAsString));
```
kathirsvn, line 201: Should we handle 'onFailure' on the _queryWithTableAsync_ Uni response, since it might throw `WebApplicationException` internally when the passed function is invoked?

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2ColumnsResourceImpl.java
```diff
@@ -10,130 +11,133 @@
...
+        .map(any -> restResponseCreatedWithName(columnName));
   }
 
   @Override
-  public Response getOneColumn(
+  public Uni<RestResponse<Object>> getOneColumn(
       String keyspaceName, String tableName, String columnName, boolean raw) {
     requireNonEmptyKeyspaceAndTable(keyspaceName, tableName);
     if (isStringEmpty(columnName)) {
       throw new WebApplicationException("columnName must be provided", Response.Status.BAD_REQUEST);
```
kathirsvn, line 67: Should we return Uni<RestResponse> with error message instead of WebApplicationException?

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2IndexesResourceImpl.java
```diff
@@ -41,8 +44,11 @@ public Response getAllIndexesForTable(String keyspaceName, String tableName) {
     }
 
     // check that we're authorized for the table
-    bridge.authorizeSchemaRead(
-        SchemaReads.table(keyspaceName, tableName, Schema.SchemaRead.SourceApi.REST));
+    Uni<List<Boolean>> uni =
+        authorizeSchemaReadsAsync(
+            Arrays.asList(
+                SchemaReads.table(keyspaceName, tableName, Schema.SchemaRead.SourceApi.REST)));
+    Futures.getUninterruptibly(uni.subscribeAsCompletionStage());
```
kathirsvn, line 51: I believe we should check the result of this uni? (i.e. check all the booleans in the list)?

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java
```diff
@@ -80,6 +128,26 @@ protected QueryOuterClass.Response queryWithTable(
...
+  protected Uni<QueryOuterClass.Response> executeQueryAsync(QueryOuterClass.Query query) {
+    return requestInfo.getStargateBridge().executeQuery(query);
+  }
+
+  protected Uni<QueryOuterClass.Response> executeQueryAsync(
+      String keyspaceName,
+      String tableName,
+      Function<Optional<Schema.CqlTable>, QueryOuterClass.Query> queryProducer) {
+
+    // TODO implement optimistic queries (probably requires changes directly in SchemaManager)
```
ivansenic, line 140: What about this, should be very simple imo? And I don't think there is anything do to with the Schema Manager:

```
return findTableAsync(keyspaceName, tableName, true)
    .onItem()
    .transformToUni(table -> executeQueryAsync(queryProducer.apply(table)))
    .onFailure(t -> {// predicate for matching the optimistic schema failure}).retry().atMost(2);


```

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2ColumnsResourceImpl.java
```diff
@@ -10,130 +11,133 @@
...
-              .table(keyspaceName, tableName)
-              .addColumn(columnDef)
-              .parameters(PARAMETERS_FOR_LOCAL_QUORUM)
-              .build();
-        });
-    return Response.status(Response.Status.CREATED)
-        .entity(Collections.singletonMap("name", columnName))
-        .build();
+            })
+        .map(any -> restResponseCreatedWithName(columnName));
```
ivansenic, line 59: and is fine not to peak at the response at all, guess it's fine..

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java
```diff
@@ -203,9 +267,22 @@ protected static byte[] decodeBase64(String base64encoded) {
...
-        .build();
+  protected static Uni<RestResponse<Object>> apiErrorResponseUni(
+      Response.Status httpStatus, String failMessage) {
+    return Uni.createFrom().item(apiErrorResponse(httpStatus, failMessage));
+  }
+
+  protected static RestResponse<Object> apiErrorResponse(
+      Response.Status httpStatus, String failMessage) {
+    return RestResponse.status(httpStatus, new ApiError(failMessage, httpStatus.getStatusCode()));
+  }
```
ivansenic, line 278: Are those needed? Usually an exception mapper would do the transformation?

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java
```diff
@@ -203,9 +267,22 @@ protected static byte[] decodeBase64(String base64encoded) {
...
+
+  protected static RestResponse<Object> restResponseCreatedWithName(String createdName) {
+    return RestResponse.status(
+        Response.Status.CREATED, Collections.singletonMap("name", createdName));
+  }
+
+  protected static RestResponse<Object> restResponseOkWithName(String name) {
+    return RestResponse.status(Response.Status.OK, Collections.singletonMap("name", name));
```
ivansenic, line 286: Where is this used, becuase this will not be documented correctly in the openapi? above as well?

ivansenic (commented):
I think it's good as the first step.. Did not look into deep details of all classes.. Gave a lot of suggestions, I guess some of them can be moved to tickets.. I saw some issues that I will list here:

* the queries are not taking the consistence level defined in the configuration
* optimistic schema queries support should be solved sooner rather than later
* looking at the response types, I am pretty sure there is lot of space for improvement in the openapi
* validation can be improved all over the place
* etc

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/Sgv2RowsResourceImpl.java
tatu-at-datastax: My assumption was that exceptions would be handled by general exception mappers like `WebApplicationExceptionMapper` so there would be usually no need to handle them in flow, unless there is specific need convert particular exception with more information available.

WDYT @ivansenic ?

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2ColumnsResourceImpl.java
tatu-at-datastax: Like previous one it would go to `WebApplicationExceptionMapper` which translates it into suitable `Response`.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2IndexesResourceImpl.java
tatu-at-datastax: Yes, that sounds like something that should be done; will change.

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QSchemaUserTypeIT.java
```diff
@@ -292,27 +292,9 @@ public void udtDelete() {
...
-        // C*3: "Invalid argument for gRPC operation (INVALID_ARGUMENT->Bad Request):
-        //   INVALID_ARGUMENT:
-        //   No user type named udt_ks_udtDelete_1662051039657.test_udt_delete exists."
-        // C*4: "Invalid argument for gRPC operation (INVALID_ARGUMENT->Bad Request):
-        //   INVALID_ARGUMENT:
-        //   Type 'udt_ks_udtDelete_1662052163437.test_udt_delete' doesn't exist"
-        .matches(String.format("Invalid argument.*%s.*exist.*", typeName));
+    // Then try 2 invalid cases; first, trying to delete again: deletion now
+    // idempotent (same as other resources) due to "ifExists", so
+    givenWithAuth().when().delete(deletePath).then().statusCode(HttpStatus.SC_NO_CONTENT);
```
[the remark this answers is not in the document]
tatu-at-datastax: Since the behavior is to use "ifExists" -- same as with all other operations -- there will not be failure for retrying delete operation. This should have been changed earlier but for some reason hasn't been.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2ColumnsResourceApi.java
```diff
@@ -54,7 +55,7 @@ public interface Sgv2ColumnsResourceApi {
         @APIResponse(ref = RestOpenApiConstants.Responses.GENERAL_404),
         @APIResponse(ref = RestOpenApiConstants.Responses.GENERAL_500)
       })
-  Response getAllColumns(
+  Uni<RestResponse<Object>> getAllColumns(
```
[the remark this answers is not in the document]
tatu-at-datastax: Ok; I followed example of Docs API I think -- but that's probably not applicable. Will create a follow-up issue.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2ColumnsResourceImpl.java
```diff
@@ -10,130 +11,133 @@
...
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
+import org.jboss.resteasy.reactive.RestResponse;
 
 public class Sgv2ColumnsResourceImpl extends RestResourceBase implements Sgv2ColumnsResourceApi {
   @Override
-  public Response getAllColumns(String keyspaceName, String tableName, boolean raw) {
+  public Uni<RestResponse<Object>> getAllColumns(
+      String keyspaceName, String tableName, boolean raw) {
     requireNonEmptyKeyspaceAndTable(keyspaceName, tableName);
```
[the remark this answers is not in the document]
tatu-at-datastax: Will create follow-up issue, that does seem cleaner (although with some duplication of error messages I guess).  Will need to verify we get appropriate error response; I assume exception mapper should convert to `ApiError`.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2ColumnsResourceImpl.java
```diff
@@ -10,130 +11,133 @@
...
+        .map(t -> table2columns(t))
+        // map to wrapper if needed
+        .map(t -> raw ? t : new Sgv2RESTResponse<>(t))
+        .map(result -> RestResponse.ok(result));
   }
 
   @Override
-  public Response createColumn(
+  public Uni<RestResponse<Object>> createColumn(
       String keyspaceName, String tableName, Sgv2ColumnDefinition columnDefinition) {
```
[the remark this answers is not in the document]
tatu-at-datastax: Sounds good, will file a follow-up issue.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java
tatu-at-datastax: > What about this, should be very simple imo? And I don't think there is anything do to with the Schema Manager:
> 
> ```
> return findTableAsync(keyspaceName, tableName, true)
>     .onItem()
>     .transformToUni(table -> executeQueryAsync(queryProducer.apply(table)))
>     .onFailure(t -> {// predicate for matching the optimistic schema failure}).retry().atMost(2);
> ```

It was an existing TODO, did not look into it yet. But sounds reasonable, adding another issue :)

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2ColumnsResourceImpl.java
tatu-at-datastax: That was the existing behavior, FWTW.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java
tatu-at-datastax: In most cases, yes, but I think there was one or two cases where there is more local information available to give better error message.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java
tatu-at-datastax: I can have a look at this, I don't remember off-hand.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/Sgv2RowsResourceImpl.java
kathirsvn: @tatu-at-datastax
I believe that's at the REST API method level exceptions, where the mapper will work.
(E.g. https://github.com/stargate/stargate/blob/6d691f80deb1808acc71f10a559e123a04eb22ed/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/Sgv2RowsResourceImpl.java#L44)

But, in this case when we transform an Uni , the exception might be thrown and this will be propagated back all the way to the client as Uni with a failure, I believe. 
(https://github.com/stargate/stargate/blob/6d691f80deb1808acc71f10a559e123a04eb22ed/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java#L144)

Please see the gist I just created https://gist.github.com/kathirsvn/c0c74eee5e9f0f6637bad28416b5083e

When you run this test, you could see the RuntimeException (WebApplicationException) is not thrown by the server method  and instead, it's returned as Uni failure. And I believe, the Exception Mapper will only work when the exception is thrown by the server method.

Please correct me if my understanding is incorrect on the Exception Mapper is incorrect. I haven't tested that part specifically.

tatu-at-datastax: @kathirsvn Ok: I do not really know `Uni` handling well enough to know either way. I'll have to think about this more next week. Sounds like you may be right wrt problem; but if so I'd ideally need a reproduction that goes through call, to verify behavior before and after fix.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2ColumnsResourceImpl.java
tatu-at-datastax: Actually I think this might be resolved more elegantly by using Bean Validation annotations as suggested by Ivan.

tatu-at-datastax:
@ivansenic @kathirsvn I think all suggestions are good and I have created issues for many (and can create more for the rest). But I was hoping to merge initial changes in first if tha's ok -- Jake is submitting a REST addition, for example, and it'd be good to have baseline changed to async.
Similarly we can get more testing done sooner.

kathirsvn (approved):

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2IndexesResourceImpl.java
kathirsvn: And also, that should be chained with the next call (fetchRowsAsync) I think.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/Sgv2RowsResourceImpl.java
ivansenic: Exception mappers will take care of this.

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QSchemaUserTypeIT.java
```diff
@@ -330,7 +312,7 @@ public void udtDelete() {
         Arrays.asList("id text", "name " + typeInUse),
         Arrays.asList("id"),
         Arrays.asList());
-    response =
+    String response =
```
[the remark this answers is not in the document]
ivansenic: It does, you can do `body("path.[0].field"` etc..

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2ColumnsResourceApi.java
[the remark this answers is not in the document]
ivansenic: Yea you should never directly return `ApiError`, but let mappers do transformation from a failure..

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java
ivansenic: That's why we have our own exception, where we can push custom messages and then have a way to catch that exception and transform to the `ApiError`.. Just saying this is an option as well, ass you can always push a failure down on the reactive sequence..

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2ColumnsResourceApi.java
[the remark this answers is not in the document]
tatu-at-datastax: @ivansenic I started working on this and can improve usage in many places. But one common problematic pattern is the "raw" handling: basically having either "ResponseWrapper<ResultValue>" OR "ResultValue". Since Java doesn't really have union types it's difficult to model it.

So I will add types where applicable (there are some `Map<String,String>`s and a few `Void`s (for DELETEs)), and then registration annotations where common base type not available.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2IndexesResourceImpl.java
tatu-at-datastax: Was able to change code to do the same in right way, but Ivan pointed out this is odd usage -- no other methods check it explicitly. So will change to use one of other methods on validating access to table.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java
tatu-at-datastax: Did end up removing and using `WebApplicationException`

PR #2214, opened 2022-11-08
closes #2209: add transient retries to the grpc client

ivansenic:
**What this PR does**:
Adds retries to the grpc client.

**Which issue(s) this PR fixes**:
Fixes #2209

PR #2216, opened 2022-11-08
closes #2159: moved common configuration to sgv2-quarkus-commons 

### LINKED ISSUE #2159: Move common and shared configuration to the `quarkus-common` `application.yaml`
In the https://github.com/stargate/stargate/pull/2158 @kathirsvn added a `application.yaml` file and seems this is picked up the way.. So we should actually transfer all the common props to the `quarkus-common` project, instead of having it duplicated in API projects..

ivansenic:
**What this PR does**:

* moved all common configuration to a single place
* moved branding assets to common, so it's picked up by all services

Fixes #2159

**Tasks**:
* [ ] sync with #2214


apis/sgv2-quarkus-common/src/main/resources/application.yaml
```diff
@@ -1,7 +1,148 @@
+---
+# please keep properties in the alphabetical order
```
tatu-at-datastax, line 2: Maybe add a note that these are the defaults for all APIs, and changes here will affect all APIs (except if overridden by API's application.yaml)

tatu-at-datastax (approved):
Nice! I like improved DRYness.

PR #2217, opened 2022-11-08
closes #2201: playground to auto-inject token from headers

### LINKED ISSUE #2201: GraphQL playground must not get auth token from the token resolver
Served at `/playground`, the path is not required to be authenticated and we must not load the token value from the `CassandraTokenResolver`, as the implementation can be `PrincipalTokenResolver`.. We need to:

1. Load token from the headers, if `requestInfo.getCassandraToken` return empty; 
   *  maybe define a configuration option that defines the header name, `stargate.graphql.playground.token-header` (default `X-Cassandra-Token`)
   * Add test that confirms that in the returned HTML we have `tokenValue = "[PASSED_HEADER_VALUE]"`
3. Enable proactive authentication, so that endpoints that do not require authentication are also checked (see https://quarkus.io/guides/security-built-in-authentication#proactive-authentication)

Second option is a config option and does not require any code changes. However, we could include it in tests.. Also we should add a test for the `PlaygroundResource` that does confirm that with default settings we will be setting the token right.

ivansenic:
**What this PR does**:
Enables auto-injecting a Cassandra token to the GraphQL playground from a header, in case the token resolver can not resolve it.

* added tests
* added config class and the readme for the configuration

Fixes #2201

### DIFF
--- a/apis/sgv2-graphqlapi/CONFIGURATION.md
+++ b/apis/sgv2-graphqlapi/CONFIGURATION.md
@@ -11,7 +11,7 @@
 |--------------------------------------------|-----------|---------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `stargate.graphql.enable-default-keyspace` | `boolean` | `true`              | Whether to default to the oldest keyspace when the user accesses `/graphql`. If this is disabled, `/graphql` throws an error, and the keyspace must be provided explicitly in the path, as in `/graphql/{keyspace_name}`. |
 | `stargate.graphql.playground.enabled`      | `boolean` | `true`              | If GraphQL Playground is enabled at `/playground`.                                                                                                                                                                        |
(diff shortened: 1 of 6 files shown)


apis/sgv2-graphqlapi/CONFIGURATION.md
```diff
@@ -0,0 +1,23 @@
...
+| `stargate.graphql.enable-default-keyspace` | `boolean` | `true`              | Whether to default to the oldest keyspace when the user accesses `/graphql`. If this is disabled, `/graphql` throws an error, and the keyspace must be provided explicitly in the path, as in `/graphql/{keyspace_name}`. |
+| `stargate.graphql.playground.enabled`      | `boolean` | `true`              | If GraphQL Playground is enabled at `/playground`.                                                                                                                                                                        |
+| `stargate.graphql.playground.token-header` | `String`  | `X-Cassandra-Token` | Optional, the header name that caries the token that should auto-injected to the playground. Note that this is used as a fallback if `CassandraTokenResolver` can not resolve the token.                                  |
```
tatu-at-datastax, line 14: nit: `caries` -> `carries`

tatu-at-datastax (approved):
LGTM

PR #2218, opened 2022-11-08
closes #2185: exclude pattern for the access log

### LINKED ISSUE #2185: Quarkus Logging - Add Exclude Pattern for non application related API calls
Exclude `/metrics` `/stargate` API calls from logging as discussed in below PR

https://github.com/stargate/stargate/pull/2178#issuecomment-1287034441

ivansenic:
**What this PR does**:
Excludes `/stargate` and `/metrics` from the access log by default. When access log is activated, the logs gets filled by those as those are accessed frequently.

Fixes #2185

**Checklist**
- [ ] Rebase after #2216 is merged

### DIFF
--- a/apis/sgv2-quarkus-common/src/main/resources/application.yaml
+++ b/apis/sgv2-quarkus-common/src/main/resources/application.yaml
@@ -77,6 +77,7 @@ quarkus:
     # access log format, must be explicitly enabled
     access-log:
       pattern: "%h %l %t \"%r\" %s %b"
+      exclude-pattern: ${quarkus.http.non-application-root-path}.*|${quarkus.micrometer.export.prometheus.path}
 
     # non-application path to /stargate
     non-application-root-path: stargate


ivansenic:
@tatu-at-datastax @jeffreyscarpenter I am not sure if this is actually needed. Maybe it's good to see in the access logs that metrics are getting scraped and health endpoints are checked?

tatu-at-datastax:
@ivansenic I can see it both ways. But if we have Prometheus metrics for call counts, separate from access logs, that would fulfill same strategy? If we don't, then I can see why we would want entries in access logs.

tatu-at-datastax (approved):

ivansenic:
@tatu-at-datastax I think we also exclude some patterns in the metrics.. So for example metrics, and health points are not included. What do you think, then we completely lose a view on those?

tatu-at-datastax:
> @tatu-at-datastax I think we also exclude some patterns in the metrics.. So for example metrics, and health points are not included. What do you think, then we completely lose a view on those?

Ah. Yeah then we'd lose all visibility which seems unfortunate. I think I'd leave them on access logs for now, esp. since access logs are not enabled by default themselves.

But should we enable scraping of healthcheck, metrics in Prometheus then? And if that was done we could exclude them here.

ivansenic:
Agree here with @tatu-at-datastax , will close this for now.

PR #2225, opened 2022-11-10
Fix for #2221 -> enhanced logging for the gRPC bridge exceptions

### LINKED ISSUE #2221: gRPC bridge exception handlers should log exceptions
We should at log the exceptions that's considered server-side errors.. A good example of those exists already in the:

```
io.stargate.sgv2.restsvc.impl.GrpcExceptionMapper#grpcResponseAndLog
```

A set of status codes that do not have to be logged: `OK|CANCELLED|INVALID_ARGUMENT|NOT_FOUND|ALREADY_EXISTS|UNAUTHENTICATED|PERMISSION_DENIED`

Because there is no logging, the original exceptions are lost.

This is same as issue #2166 but for the gRPC bridge component

versaurabh:
**What this PR does**:
1. This PR logs additional details for the exceptions in the gRPC bridge, the additional details are `status code and the stacktrace` 
2. Excludes the detailed logging for status codes `OK|CANCELLED|INVALID_ARGUMENT|NOT_FOUND|ALREADY_EXISTS|UNAUTHENTICATED|PERMISSION_DENIED`

Fixes #2221

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/bridge/src/main/java/io/stargate/bridge/service/ExceptionHandler.java
+++ b/bridge/src/main/java/io/stargate/bridge/service/ExceptionHandler.java
@@ -74,8 +74,10 @@ public class ExceptionHandler {
       handleException(throwable.getCause());
     } else if (throwable instanceof StatusException
         || throwable instanceof StatusRuntimeException) {
+      LOG.error(getErrorMessage(null, throwable), throwable);
       onError(throwable);
     } else if (throwable instanceof UnhandledClientException) {
+      LOG.error(getErrorMessage(Status.UNAVAILABLE, throwable), throwable);
       onError(Status.UNAVAILABLE, throwable);
     } else if (throwable instanceof PersistenceException) {
       handlePersistenceException((PersistenceException) throwable);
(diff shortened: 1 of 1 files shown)


tatu-at-datastax (changes requested):
I think we should use same criteria for not logging entries and/or stack traces as with public gRPC service, unless there are specific reasons to deviate.
Looks like right now everything would be logged as error, including stack trace in all cases.
So should remove stack trace argument for about half cases and remove logging itself for couple of client failures.

versaurabh:
> I think we should use same criteria for not logging entries and/or stack traces as with public gRPC service, unless there are specific reasons to deviate. Looks like right now everything would be logged as error, including stack trace in all cases. So should remove stack trace argument for about half cases and remove logging itself for couple of client failures.

Yes @tatu-at-datastax, below is the impl matrix for logging and stacktraces

|  | Logged | Not Logged |
| ------------- | ------------- | ------------- |
| **Stacktrace logged**  |StatusException, StatusRuntimeException, UnhandledClientException, SERVER_ERROR, PROTOCOL_ERROR, UNPREPARED, CAS_WRITE_UNKNOWN, CONFIG_ERROR, TRUNCATE_ERROR, CDC_WRITE_FAILURE, READ_FAILURE, FUNCTION_FAILURE, WRITE_FAILURE  | ---NA--- |
| **Stacktrace not logged**  | UNAVAILABLE, IS_BOOTSTRAPPING, WRITE_TIMEOUT, READ_TIMEOUT | INVALID, SYNTAX_ERROR, BAD_CREDENTIALS, OVERLOADED, UNAUTHORIZED, ALREADY_EXISTS |

bridge/src/main/java/io/stargate/bridge/service/ExceptionHandler.java
```diff
@@ -160,6 +189,7 @@ private void handleUnavailable(UnavailableException ue) {
   }
 
   private void handleWriteTimeout(WriteTimeoutException wte) {
+    LOG.error(getErrorMessage(Status.DEADLINE_EXCEEDED, wte));
```
ivansenic, line 192: timeouts are not the same as in the public or? Can you adapt so it s same 1:1?

ivansenic (approved):

bridge/src/main/java/io/stargate/bridge/service/ExceptionHandler.java
versaurabh: timeouts are handled in the same way as in public, both public and bridge follow https://github.com/stargate/stargate/pull/2225#issuecomment-1311294546 impl matrix for logging and stacktrace

ivansenic: my bad

PR #2228, opened 2022-11-10
[WIP] search document to ensure document matches the expression

ivansenic:
**What this PR does**:
Fixes possible race condition between document candidate search and population.

Internal.

**Tasks**:
- [ ] Add test

@amorton Can you re-test plase.

### DIFF
--- a/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/ReadBridgeService.java
+++ b/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/ReadBridgeService.java
@@ -19,6 +19,8 @@ package io.stargate.sgv2.docsapi.service.query;
 
 import com.bpodgursky.jbool_expressions.Expression;
 import com.bpodgursky.jbool_expressions.Literal;
+import com.bpodgursky.jbool_expressions.eval.EvalEngine;
+import com.bpodgursky.jbool_expressions.eval.EvalRule;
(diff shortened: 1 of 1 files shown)


apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/query/ReadBridgeService.java
```diff
@@ -89,15 +93,24 @@ public Multi<RawDocument> searchDocuments(
...
+          .filter(
+              document -> {
+                Map<String, EvalRule<FilterExpression>> rules = EvalEngine.booleanRules();
+                rules.put(FilterExpression.EXPR_TYPE, new RawDocumentEvalRule(document));
+                return EvalEngine.evaluate(expression, rules);
+              })
+
+          // limit to requested page size only to stop fetching extra docs
+          .select()
+          .first(paginator.docPageSize);
```
amorton, line 113: we already select the first paginator.docPageSize on line 97, what is the second for ?

ivansenic: The first on line `97` was moved here.. Not sure I actually understand, do I have two `.first(paginator.docPageSize)` calls now?

amorton: My bad, I  missed the line removals.

PR #2252, opened 2022-11-21
Fix #2245: allow numeric timestamp for CQL `timestamp` valued columns

### LINKED ISSUE #2245: Timestamp fields are now expected to be provided as `Instant` and no more numbers
`Timestamp` in REST payload are expected to be `Instant` in V2 whereas it was `long` in `V1`. (at least test was passing)

I would recommend accepting both to avoid any regression.

tatu-at-datastax:
**What this PR does**:

Allows use of JSON integral numbers for CQL `timestamp` valued columns (same as Stargate V1 / REST)

Fixes #2245

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/ToProtoValueCodecs.java
+++ b/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/ToProtoValueCodecs.java
@@ -611,6 +611,11 @@ public class ToProtoValueCodecs {
       if (value instanceof String) {
         return protoValueFromStringified((String) value);
       }
+      // Should only accept integral numbers; Jackson will expose these as Integers and Longs
+      // (unfortunately "java.lang.Number" has no convenience methods for determining this)
+      if (value instanceof Long || value instanceof Integer) {
+        return Values.of(((Number) value).longValue());
+      }
       return cannotCoerce(value);
     }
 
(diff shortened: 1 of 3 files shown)


jeffreyscarpenter (approved):

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/ToProtoValueCodecs.java
```diff
@@ -611,6 +611,11 @@ public QueryOuterClass.Value protoValueFromStrictlyTyped(Object value) {
       if (value instanceof String) {
         return protoValueFromStringified((String) value);
       }
+      // Should only accept integral numbers; Jackson will expose these as Integers and Longs
+      // (unfortunately "java.lang.Number" has no convenience methods for determining this)
+      if (value instanceof Long || value instanceof Integer) {
+        return Values.of(((Number) value).longValue());
+      }
```
ivansenic, line 618: how can a timestamp be represented in an Integer? The max value is not enough.

ivansenic (changes requested):

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/ToProtoValueCodecs.java
tatu-at-datastax: If caller specifies a number that fits in `int` it gets bound to `Integer`, that's the only reason.
In most cases it's big enough to become `Long`.
But such small values are technically valid (from early 1970s) so should, I think, be accepted.

ivansenic (approved):

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/ToProtoValueCodecs.java
tatu-at-datastax: > OK, was not aware that early 1970s can fit into an int..

Yeah it's just because the textual representation (like `0` for 1970-01-01 UTC) fits... JSON has not types for different classes. So parser just sees a "small enough" number, without context. Coercion to `long` is done if and as needed for properties, but when read as `java.lang.Object` it'll choose smaller choice (does not try to use `Short` etc tho, just `int` and `long` and `BigInteger` for beyond).

PR #2256, opened 2022-11-22
Add explicit handling of "empty" Tuples Bridge returns for missing Tuples

### LINKED ISSUE #2246: Cannot retrieve tuples field from GET REST API V2 when value is null
Given the following data in the database:

- Table Structure
```sql
CREATE TABLE java.videos (
    genre text,
    year int,
    title text,
    formats frozen<map<text, text>>,
    tuples frozen<tuple<text, text, text>>,
    upload timestamp,
    frames list<int>,
    tags set<text>,
    PRIMARY KEY (genre, year, title)
) WITH CLUSTERING ORDER BY (year ASC, title ASC);
```

- Query
```sql
 select * from videos where genre='Sci-Fi' AND year=1990 AND title='line_update';
```

- Data 
```
 genre  | year | title       | formats | frames | tags | tuples | upload
--------+------+-------------+---------+--------+------+--------+---------------------------------
 Sci-Fi | 1990 | line_update |    null |   null | null |   null | 2022-11-18 08:16:35.243000+0000
```

When you want to retrieve the row providing the full primary key

- `GET` on `http://localhost:8082/v2/keyspaces/java/videos/Sci-Fi/1990/line_update`

The following error is thrown:

```json
{
    "description": "Internal error: failed to convert value of column #7/#8 ('tuples'), problem: Illegal Tuple representation, expected 3 values, received 0",
    "code": 500,
    "grpcStatus": null,
    "internalTxId": null
}
```

The field `tuples` is nullable and still an marshalling is expected.

tatu-at-datastax:
**What this PR does**:

Adds explicit handling (and tests) for cases where Bridge returns "empty" Tuple (that represents missing value for a Tuple column): will be converted to JSON `null`

Fixes #2246

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/FromProtoValueCodecs.java
+++ b/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/FromProtoValueCodecs.java
@@ -601,6 +601,9 @@ public class FromProtoValueCodecs {
     public Object fromProtoValue(QueryOuterClass.Value value) {
       QueryOuterClass.Collection coll = value.getCollection();
       final int len = verifyTupleLength(coll);
+      if (len == 0) {
+        return null;
+      }
       List<Object> result = new ArrayList<>(len);
       for (int i = 0; i < len; ++i) {
         result.add(elementCodecs.get(i).fromProtoValue(coll.getElements(i)));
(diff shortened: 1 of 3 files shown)


ivansenic (approved):
I am giving an approval, but I am not sure how the integration test is confirm this works now..

tatu-at-datastax:
> I am giving an approval, but I am not sure how the integration test is confirm this works now..

I added a case where new row is added without specifying `Tuple` value: this is then fetched.
Before fix it failed with exception reported; after fix does not.
Also added unit test for same conversion.

So there are tests that I think cover what is needed. But I think Cedric can validate via `curl` once this is merged.

PR #2259, opened 2022-11-23
fixing missing netty-codec dependency for dse

ivansenic:
**What this PR does**:
The `dse-db` in version `6.8.29` is not specifying the `netty-codec` as transitive dependency, thus ending up in the no class found exception. I added the dependency explicitly. 

Furthermore, I bumped the netty version to the same DSE is using, @tatu-at-datastax can you confirm this is OK.

### DIFF
--- a/persistence-dse-6.8/pom.xml
+++ b/persistence-dse-6.8/pom.xml
@@ -127,6 +127,11 @@
 
     <!-- 3rd party dependencies -->
     <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-codec</artifactId>
+      <version>4.1.78.1.dse</version>
+    </dependency>
+    <dependency>
       <groupId>org.immutables</groupId>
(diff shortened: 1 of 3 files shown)


persistence-dse-6.8/pom.xml
```diff
@@ -126,6 +126,11 @@
     </dependency>
 
     <!-- 3rd party dependencies -->
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-codec</artifactId>
+      <version>4.1.78.1.dse</version>
```
tatu-at-datastax, line 132: Yes, I believe this is correct as per:

https://github.com/datastax/release-notes/blob/master/DSE_6.8_Release_Notes.md

for DSE 6.8.29

tatu-at-datastax (approved):

tatu-at-datastax (approved):

PR #2266, opened 2022-11-23
Add CQL endpoint to restapi such that webapps can execute cql statements directly

tjake:


restapi/src/main/java/io/stargate/web/restapi/resources/v2/CqlResource.java
```diff
@@ -0,0 +1,120 @@
...
+import org.apache.cassandra.stargate.db.ConsistencyLevel;
+
+@Api(
+    produces = MediaType.APPLICATION_JSON,
+    consumes = MediaType.TEXT_PLAIN,
+    tags = {"data"})
+@Path("/v2/cql")
+@Produces(MediaType.APPLICATION_JSON)
+@Singleton
+public class CqlResource {
```
tatu-at-datastax, line 47: Do we want to add this implementation for Stargate V1 as well? This will not be used or needed with Stargate V2.
We are planning on removing most of SGv1 REST implementation in couple of weeks (from `main` at least).
The only use case would be for Astra Classic I suppose.
But leaving this out would allow leaving out `QueryBuilderImpl` too.

tjake: I assume we need it for classic,  should this be in old v2 or should I move to v1?

tatu-at-datastax: @tjake Class to modify is correct for Stargate V1 but branch wrong. So would need separate PRs here.

@jeffreyscarpenter I think we need to talk about this, wrt our refactoring and so on.

tjake: I removed the v1 code

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/CQLResource.java
```diff
@@ -0,0 +1,80 @@
...
+          final String payloadAsString) {
+    QueryOuterClass.Query query =
+        new QueryBuilder()
+            .cql(payloadAsString)
+            .parameters(parametersForPageSizeAndState(pageSizeParam, pageStateParam))
+            .build();
+    return requestInfo
+        .getStargateBridge()
+        .executeQuery(query)
+        .map(response -> convertRowsToResponse(response, false));
```
tatu-at-datastax, line 78: Could also take in `raw` query parameter for consistency with other endpoints?

polandll:
So, would you set a query, like `SELECT * FROM ks.table` to a variable and then use `/v2/cql/{var_name}`? Or would you use some delimiter like quotes, like `/v2/cql/"SELECT * FROM ks.table"`?

tjake:
It's a POST since the cql can be quite long (batch statements)

polandll:
Ah yeah - that makes more sense! Thanks @tjake

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/CQLResource.java
```diff
@@ -0,0 +1,83 @@
...
+    QueryOuterClass.Query query =
+        new QueryBuilder()
+            .cql(payloadAsString)
+            .parameters(
+                parametersForPageSizeStateAndKeyspace(pageSizeParam, pageStateParam, keyspace))
+            .build();
+    return requestInfo
+        .getStargateBridge()
+        .executeQuery(query)
+        .map(response -> convertRowsToResponse(response, raw));
```
ivansenic, line 81: what happens if the query is not valid? do we have a test that proves this would be translated into `400`?

apis/sgv2-restapi/src/main/resources/application.yaml
```diff
@@ -43,6 +43,7 @@ quarkus:
         # we need to define uri templating on our own for now
         # note that order is important
         match-patterns: |
+          /v2/cql,
```
ivansenic, line 46: Remove

* this is a wrong pattern, see that all use `=`
* we don't need this one, as it has not path based templating

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/CQLResource.java
```diff
@@ -0,0 +1,83 @@
...
+          final String keyspace,
+      @Parameter(name = "page-size", ref = RestOpenApiConstants.Parameters.PAGE_SIZE)
+          @QueryParam("page-size")
+          final int pageSizeParam,
+      @Parameter(name = "page-state", ref = RestOpenApiConstants.Parameters.PAGE_STATE)
+          @QueryParam("page-state")
+          final String pageStateParam,
+      @Parameter(name = "raw", ref = RestOpenApiConstants.Parameters.RAW) @QueryParam("raw")
+          final boolean raw,
+      @RequestBody(description = "CQL Query String", required = true)
```
ivansenic, line 70: add `@NotBlank` here with message for proper validation of the CQL existing

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QCqlIT.java
```diff
@@ -0,0 +1,39 @@
...
+
+import io.quarkus.test.common.QuarkusTestResource;
+import io.quarkus.test.junit.QuarkusIntegrationTest;
+import io.restassured.http.ContentType;
+import io.stargate.sgv2.common.testresource.StargateTestResource;
+import org.junit.jupiter.api.Test;
+
+@QuarkusIntegrationTest
+@QuarkusTestResource(StargateTestResource.class)
+public class RestApiV2QCqlIT extends RestApiV2QIntegrationTestBase {
```
ivansenic, line 13: Missing test cases:

* CQL not valid, should end up in `400`
* Body empty, validation kicks in

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QCqlIT.java
```diff
@@ -0,0 +1,39 @@
...
+            .statusCode(200)
+            .extract()
+            .asString();
+
+    assertThat(r)
+        .isEqualTo(
+            "{\"count\":1,\"data\":[{\"cluster_name\":\""
+                + StargateTestResource.getClusterName()
+                + "\"}]}");
+  }
```
ivansenic, line 38: can we use any other test that would not depend on the `cluster_name`, for example:

* create table
* insert row
* read that row and assert on the data

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/common/testresource/StargateTestResource.java
```diff
@@ -287,11 +287,11 @@ private String getStargateImage() {
     }
   }
 
-  private String getClusterName() {
+  public static String getClusterName() {
     return System.getProperty("testing.containers.cluster-name", Defaults.CLUSTER_NAME);
   }
 
-  private String getClusterVersion() {
+  public static String getClusterVersion() {
```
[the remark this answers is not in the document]
tjake: Ok

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/CQLResource.java
tjake: added test

ivansenic (approved):

tatu-at-datastax:
@tjake Good to go -- can you merge it or do I need to?

PR #2270, opened 2022-11-24
organize dependencies in the DSE

ivansenic:
**What this PR does**:
Trying to fix the issues with the DSE dependencies.

Small info about the changes:
* netty-codec removed from explicit dependencies, will be brought in by dse-db
* the netty handler must be ignored from persistance-api, that s the only way not to override the netty-codec as the handler would be introducing it in different version
* i tried to remove the explicit dep to the dse java driver, as core would bring the right one
* I tested this against main, so that all int tests are checked as well.. but moving the base to the v1

### DIFF
--- a/persistence-dse-6.8/pom.xml
+++ b/persistence-dse-6.8/pom.xml
@@ -44,6 +44,20 @@
       <artifactId>persistence-api</artifactId>
       <version>${project.version}</version>
       <scope>provided</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>com.datastax.oss</groupId>
+          <artifactId>java-driver-core</artifactId>
+        </exclusion>
+        <exclusion>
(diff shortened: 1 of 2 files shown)


persistence-dse-6.8/pom.xml
```diff
@@ -126,11 +123,6 @@
     </dependency>
 
     <!-- 3rd party dependencies -->
-    <dependency>
-      <groupId>io.netty</groupId>
-      <artifactId>netty-codec</artifactId>
-      <version>4.1.78.1.dse</version>
-    </dependency>
```
tatu-at-datastax, line 133: Good riddance! Glad we can remove this now.

tatu-at-datastax (approved):
I don't know how to validate this better than our CI/IT: looks good to me otherwise.
Handling of dependencies is tricky with DSE (and also due to OSGi complications).

PR #2271, opened 2022-11-24
setup the branding for the ECR images

ivansenic:
**What this PR does**:
Special branding for the ECR docker images.. Directly as the env var during the build time..

Internal issue.

### DIFF
--- a/.github/workflows/release-v2.yml
+++ b/.github/workflows/release-v2.yml
@@ -308,11 +308,50 @@ jobs:
         run: |
           ./build_docker_images.sh -p -t ${{needs.resolve-tag.outputs.release-tag}} -r ${{ secrets.ECR_REPOSITORY }} -a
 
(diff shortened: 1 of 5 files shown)


tatu-at-datastax:
Should this also include addition of `info-description` for GraphQL API? (it looked like one only added for REST and Docs API)

ivansenic:
> Should this also include addition of `info-description` for GraphQL API? (it looked like one only added for REST and Docs API)

GraphQL API does not have a swagger-ui, so I did not do it there.. But I could at least adapt the general application name..

jeffreyscarpenter (approved):

tatu-at-datastax:
> > Should this also include addition of `info-description` for GraphQL API? (it looked like one only added for REST and Docs API)
> 
> GraphQL API does not have a swagger-ui, so I did not do it there.. But I could at least adapt the general application name..

Ah. I always forget this. :)

PR #2272, opened 2022-11-28
gRPC exception logs for ValuesHelper

versaurabh:
**What this PR does**:
1. This PR logs additional details for the exceptions in the gRPC bridge, the additional details are `status code and the stacktrace` 
2. Excludes the detailed logging for status codes `OK|CANCELLED|INVALID_ARGUMENT|NOT_FOUND|ALREADY_EXISTS|UNAUTHENTICATED|PERMISSION_DENIED`
3. Impl matrix as mentioned in https://github.com/stargate/stargate/pull/2225#issuecomment-1311294546

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/grpc/src/main/java/io/stargate/grpc/service/ValuesHelper.java
+++ b/grpc/src/main/java/io/stargate/grpc/service/ValuesHelper.java
@@ -23,6 +23,7 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import io.grpc.Status;
 import io.grpc.StatusException;
+import io.grpc.StatusRuntimeException;
 import io.stargate.db.BoundStatement;
 import io.stargate.db.Result.Prepared;
 import io.stargate.db.Result.Rows;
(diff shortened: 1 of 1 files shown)


grpc/src/main/java/io/stargate/grpc/service/ValuesHelper.java
```diff
@@ -43,32 +44,43 @@
...
-                    columnCount, namesCount))
-            .asException();
+        StatusException statusException =
+            Status.FAILED_PRECONDITION
+                .withDescription(
+                    String.format(
+                        "Invalid number of bind names. Expected %d, but received %d",
+                        columnCount, namesCount))
+                .asException();
+        LOG.error(getErrorMessage(Status.FAILED_PRECONDITION, statusException), statusException);
```
tatu-at-datastax, line 82: This may make sense for now, to get more information, but aren't these client errors triggered by incorrect usage of gRPC client by customers? Since we alert on ERRORs this can lead to lots of noise.

grpc/src/main/java/io/stargate/grpc/service/ValuesHelper.java
```diff
@@ -189,72 +205,100 @@ public static TypeSpec convertType(ColumnType columnType) throws StatusException
...
+          builder.setUdt(udtBuilder.build());
+          break;
+        default:
+          builder.setBasic(
+              Objects.requireNonNull(
+                  TypeSpec.Basic.forNumber(columnType.id()), "Unhandled parameterized type"));
+      }
+    } catch (StatusException statusException) {
+      LOG.error(getErrorMessage(statusException.getStatus(), statusException), statusException);
+      throw statusException;
```
tatu-at-datastax, line 276: Usually logging AND rethrowing is an anti-pattern. Does log line here have enough information to be useful for us? Like contextual information on caller and so on?

tatu-at-datastax:
Ok I know this is already merged but I think we need to see how this works -- I have some concerns on logging at low-level code, in cases where we have no control over failures (as far as I understand these are mostly failures due to clients sending invalid gRPC requests).

jeffreyscarpenter:
@tatu-at-datastax it looks like this was closed, not merged. Is more discussion required @versaurabh ?

versaurabh:
hi @tatu-at-datastax @jeffreyscarpenter 
- Agreed that the logging with this PR may / would create noise, one of the questions to answer this would be: would enabling the logs in this PR allow / help us with details of the `StargateGrpcErrorBudgetBurn` errors on PD which we are getting for couple of customers. 
- I can think of enabling these logs via a system property for these customers and then may be get back to them with exact details, and then disable these logs with time per customer, wdyt?

tatu-at-datastax:
@versaurabh A question I have is that wouldn't these already be logged by caller since an exception is thrown? As long as this exception has all pertinent information caller can decide to log or not, and I think we have changed this appropriately. So we would not want to double log things? Apologies if I misunderstood logic here.

On System property, maybe not add that yet since changing value would require restart anyway. We may want to start thinking about dynamic settings however, in general.

versaurabh:
- Oh right @tatu-at-datastax, the changes done earlier for exception handling in `ExceptionHandler` would be able to log these exceptions as well via method  ` handleException`(https://github.com/stargate/stargate/blob/v1/grpc/src/main/java/io/stargate/grpc/service/ExceptionHandler.java#L58), in which case this PR would double log the exceptions -> going ahead and closing https://github.com/stargate/stargate/pull/2274 and https://github.com/stargate/stargate/pull/2273

PR #2279, opened 2022-11-30
Fix #2268: check access rights as expected (plus IT improvements)

### LINKED ISSUE #2268: We should check the result of this uni (check all the booleans in the list) in `Sgv2IndexesResourceImpl`
I believe we should check the result of this `Uni` returned by `authorizeSchemaReadsAsync()`? (i.e. check all the booleans in the list)?

And also, that should be chained with the next call (fetchRowsAsync) I think.

_Originally posted by @kathirsvn in https://github.com/stargate/stargate/pull/2213#discussion_r1029788677_

tatu-at-datastax:
**What this PR does**:

Adds handling of access checks as per #2268. Also fixes Integration Test for "get all indexes for a table": was not using REST endpoint but direct access from system talbe.

Fixes #2268

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2IndexesResourceImpl.java
+++ b/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2IndexesResourceImpl.java
@@ -22,7 +22,6 @@ import io.stargate.bridge.proto.QueryOuterClass.Query;
 import io.stargate.bridge.proto.Schema;
 import io.stargate.sgv2.api.common.cql.builder.Predicate;
 import io.stargate.sgv2.api.common.cql.builder.QueryBuilder;
-import io.stargate.sgv2.api.common.futures.Futures;
 import io.stargate.sgv2.api.common.grpc.proto.SchemaReads;
 import io.stargate.sgv2.restapi.service.models.Sgv2IndexAddRequest;
 import io.stargate.sgv2.restapi.service.resources.RestResourceBase;
@@ -43,14 +42,6 @@ public class Sgv2IndexesResourceImpl extends RestResourceBase implements Sgv2Ind
     if (isStringEmpty(tableName)) {
       throw new WebApplicationException("tableName must be provided", Status.BAD_REQUEST);
     }
-
-    // check that we're authorized for the table
-    Uni<List<Boolean>> uni =
-        authorizeSchemaReadsAsync(
-            Arrays.asList(
-                SchemaReads.table(keyspaceName, tableName, Schema.SchemaRead.SourceApi.REST)));
-    Futures.getUninterruptibly(uni.subscribeAsCompletionStage());
-
     Query query =
         new QueryBuilder()
             .select()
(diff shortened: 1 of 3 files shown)


apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2IndexesResourceImpl.java
```diff
@@ -59,7 +50,23 @@ public Uni<RestResponse<Object>> getAllIndexes(String keyspaceName, String table
...
+        .onItem()
+        .invoke(
+            accessChecks -> {
+              if (!accessChecks.get(0).booleanValue()) {
+                throw new WebApplicationException(
+                    "Not authorized for read access to table '" + tableName + "'",
+                    Status.UNAUTHORIZED);
+              }
+            })
+        .onItem()
```
ivansenic, line 67: Wait a bit.. Why are we having the `authorizeSchemaReadsAsync` method being used at all here? Imo this method should be removed, and everything should be going throught the schema manager? Would would not the `getTableAuthorized` not be used here? Imo remove the `authorizeSchemaReadsAsync` from the code and adapt where was used..

I have to say that this mix of `Async` methods existing in the resource base is very confusing.. Can you just not go directly to the schema manager? Or if you really need some extra processing, check how we created `CollectionManager` (table manager) and `NamespaceManager` (keyspace manager) that deal with tables and keyspaces..

tatu-at-datastax: @ivansenic I am not 100% sure since this is code that come from previous implementation. I was wondering about same thing but assumed there was a reason (was first contemplating just removing it); possibly since actual read is from system tables, to get information about user tables' indexes. But same applies to many other places where no check is done so I guess I'll go and change handling here to use one of other methods.

As to confusion, I don't know. It is true that they have accumulated over time due to changes in handling but being very thin layer I am not sure I'd find them major distraction.
At this point I think I'd rather not spend more time on perfecting abstractions and get this iteration completed.

tatu-at-datastax: @ivansenic Removed explicit call and using method that checks authorization as part of table access; then uses query against system table to get actual information.

Also added check for some basic failures (although not for unauthorized: wasn't sure how to trigger authorization failure easily).

ivansenic (approved):
Fine. But think about call optimization, following what s going on in this simple change was a challenge

```
-> executeQueryAsync
 -> findTableAsync
  -> getTableAsync
```
It's really hard to follow..

tatu-at-datastax:
> Fine. But think about call optimization, following what s going on in this simple change was a challenge
> 
> ```
> -> executeQueryAsync
>  -> findTableAsync
>   -> getTableAsync
> ```
> 
> It's really hard to follow..

Ok yeah, I think last part had to with `Optional` handling and checks. There is some complexity in there, probably due to conversion from blocking to non-blocking.
But looking at SchemaManager there's quite a bit of complexity there too fwtw.

PR #2282, opened 2022-11-30
Enable Github Action for Coordinator Tests for `main` pushes, PRs

### LINKED ISSUE #2281: Enable Github Action based Coordinator Test to replace GCB for Stargate V2
We should be able to enable `.github/workflows/coordinator-test.yml` on Stargate V2 active branches as well as PRs -- so far it has only been manually startable.
If and when this works well this should replace Google Cloud Build we have used so far.

We can also consider doing the same for `v1` (Stargate V1) but it as bit more testing as there are also tests for in-Coordinator APIs

tatu-at-datastax:
**What this PR does**:

Enables formerly manual-only run Coordinator Test Github Action for PRs, pushes to `main` (Stargate V2). This will allow moving these tests from GCB and unify our CI/IT handling.

Fixes #2281

**Checklist**
- [x] Changes manually tested


kathirsvn (approved):

tatu-at-datastax:
Not sure why but C*3.11 build failed while C*4 and DSE passed just fine.
Re-run immediately fails (gets cancelled).

ivansenic:
Wrote on the issue, this is better place:

* This will be fine when v2 stuff and tests are moved out of the coordinator.. So maybe wait until then? We could also use more powerful machines on the GH..
* We could split unit tests and int tests, I now this is extra work, but basically we are redoing stuff in all three runs at the moment.. Maybe the over time will not be better, but at least we would fail faster and spend less minutes..
* Another issue is that ccm must download the C* distribution on each run, that's lot of resources kinda  screwed up, maybe there is a way to use custom build images on the GH as well?

ivansenic:
> 

Ths GCB machine has 32 cores.. We could try with a bit more resources here.. And increase heap size for maven as well, this saves like 10 mins in the DSE build.. https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners

tatu-at-datastax:
Looks like bigger runners require changes to enterprise settings, so maybe first trying Maven heap size is worth it. Not sure what is the best way to do it, trying one setting (although that applies to all backends).

tatu-at-datastax:
@ivansenic I tried couple of ways to increase Maven heap size (although TBH not 100% sure what Github Actions default to), but all attempts failed without good error message. Perhaps 2gigs is too big for our runners? Or maybe I did not find the right way to increase settings.

ivansenic:
This works:

https://github.com/stargate/stargate/blob/7a4f0fd389dd9d05bc3b17389322a2e2df2382dc/.github/workflows/release-v2.yml#L24

tatu-at-datastax:
> This works:
> 
> https://github.com/stargate/stargate/blob/7a4f0fd389dd9d05bc3b17389322a2e2df2382dc/.github/workflows/release-v2.yml#L24

Thank you! Will try and see if that might work.

I am not sure what the default Maven heap settings are: usual locally would be (I thought) 2 gigs, but GH runners seem to have more limited resource settings.

tatu-at-datastax:
Unfortunately run fails without any indication on what is happening: my guess is that there isn't enough resources to give to Maven. Will try one last time with 1.5 gigs (which I think is still more than what defaults are) but running out of options for default runners.

tatu-at-datastax:
Looks like I was wrong: I may have had a problem with configs (in one case there was tab char from editor and that breaks YAML indentation). Maven heap size of 4g is accepted, but does not seem to improve times (no different between 1.5, 3 or 4 that I can see). Not sure why this is different from what Ivan saw.

Will leave this PR for a while, can return after we manage to deploy SGv2 with routing.

jeffreyscarpenter:
Isn't this now obsolete @tatu-at-datastax ?

tatu-at-datastax:
yes @jeffreyscarpenter, will close.

PR #2286, opened 2022-12-02
Add simple substitution for Maven project.version into "index.html"

### LINKED ISSUE #2285: Inject build version number of APIs into `index.html` (or another easily accessible place)
Currently there is no easy way to check the version of APIs deployed from resource endpoints or web UI.
But since the API frontpage comes from:

    sgv2-graphqlapi/src/main/resources/META-INF/resources/index.html

(etc for others)
It seems that we could perhaps embed version number there dynamically: either during release or perhaps during build, if templating is supported.

The point is that it should be easy to just check a web UI or call resource to get the version number that build has.

tatu-at-datastax:
**What this PR does**:

Injects `project.versions` from Maven build into Quarkus webapp home page "index.html" so that we can easily see version of (REST) API deployed.

Note: at first only added for REST API but can easily add to other APIs if we like the idea

Fixes #2285

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/sgv2-restapi/pom.xml
+++ b/apis/sgv2-restapi/pom.xml
@@ -139,6 +139,30 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>com.google.code.maven-replacer-plugin</groupId>
+        <artifactId>replacer</artifactId>
+        <version>1.5.3</version>
+        <executions>
+          <execution>
+            <phase>prepare-package</phase>
+            <goals>
+              <goal>replace</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <ignoreMissingFile>false</ignoreMissingFile>
+          <file>${basedir}/src/main/resources/META-INF/resources/index.html</file>
+          <outputFile>${basedir}/target/classes/META-INF/resources/index.html</outputFile>
+          <replacements>
+            <replacement>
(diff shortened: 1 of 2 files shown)


jeffreyscarpenter (approved):

ivansenic (commented):
Do we really need this? Would the git changes here be handled by the PR that is bumping the version?

tatu-at-datastax:
> Do we really need this? Would the git changes here be handled by the PR that is bumping the version?

Seems nice to have a way to see the actual build version, not mandatory of course.

Replacer simply uses active value from Maven build to modify file mechanically, not relying on git per se. As far as I know it works fine under all circumstances.

ivansenic:
> > Do we really need this? Would the git changes here be handled by the PR that is bumping the version?
> 
> Seems nice to have a way to see the actual build version, not mandatory of course.
> 
> Replacer simply uses active value from Maven build to modify file mechanically, not relying on git per se. As far as I know it works fine under all circumstances.

Yea sure, there are no constant git changes here.. Would this work if defined in commons with the regex for targetting the index.html files in all sub-modules?

ivansenic:
Wait merged, what about other APIs?

tatu-at-datastax:
@ivansenic Ah. Sorry, yes, it would make sense to update others -- will do that as follow-up.

PR #2293, opened 2022-12-06
V1 0 69 to main

tatu-at-datastax:
**What this PR does**:

Merges `v1` into `main` to resolve bogus version differences in `pom.xml`s after release

**Checklist**
- [ ] Changes manually tested


jeffreyscarpenter:
Just to confirm - the net change from these two commits appears to be the removal of a single line comment?

tatu-at-datastax:
@jeffreyscarpenter Yes. Unfortunately merging "no changes" is not possible, which is unfortunate -- what happens here is that we have a merge conflict with every `pom.xml` (due to different versions, changes), which has to be "merged". But since `main` does not change, cannot do PR without at least one actual change.

tatu-at-datastax:
Not needed since I have to release v2.0.3, closing.

PR #2302, opened 2022-12-08
Fix #2231: map exception for keyspace creation failure to more meaningful

### LINKED ISSUE #2231: Keyspace Creation with DC information KO in V2
The following request is not passing anymore.

Docker image
```
stargateio/restapi:v2
```

**METHOD**: `POST`
**URL**: `http://localhost:8082/v2/schemas/keyspaces/`
**BODY:**

```json
{ 
    "name":"ks3",
    "datacenters":[
        {"name":"dc1","replicas":1}
    ]
}
```

Error Messages:
```json
{
    "description": "Invalid state or argument(s) for gRPC operation (Status.Code.FAILED_PRECONDITION->400): FAILED_PRECONDITION: Unrecognized strategy option {dc1} passed to NetworkTopologyStrategy for keyspace ks3",
    "code": 400,
    "grpcStatus": "FAILED_PRECONDITION",
    "internalTxId": null
}
```

CURL COMMAND (token to update)
```
curl --location --request POST 'http://localhost:8082/v2/schemas/keyspaces/' \
--header 'X-Cassandra-Token: f19d846e-2b8d-4c9f-8ac5-6491eec1234e' \
--data-raw '{ 
    "name":"ks3",
    "datacenters":[
        {"name":"dc1","replicas":1}
    ]
}'
```

* It is still what's written in the swagger (cannot use the swagger-UI with a token, i got an error not sure why)

Screenshot 2022-11-13 at 6 57 52 PM

----

* Not related but `grpcStatus` and `internalTxId` should probably not be part of the public output . The exception coming from the bridge should be swallowed and converted to explicit error message*

tatu-at-datastax:
**What this PR does**:

Improves error handling to resolve #2231

Fixes #2231

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2KeyspacesResourceImpl.java
+++ b/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2KeyspacesResourceImpl.java
@@ -124,14 +124,16 @@ public class Sgv2KeyspacesResourceImpl extends RestResourceBase
     // Simple execution but resulting failure from bad DC needs to be mapped to something
     // more useful (as per [stargate#2231])
     return executeQueryAsync(query)
-        .onFailure(failure -> failureMatchesBadDC(failure, dcMap))
+        .onFailure()
         .transform(
             failure ->
-                new WebApplicationException(
-                    String.format(
-                        "Bad request: one or more of datacenters passed (%s) invalid",
-                        dcMap.keySet()),
-                    Response.Status.BAD_REQUEST))
+                failureMatchesBadDC(failure, dcMap)
+                    ? new WebApplicationException(
+                        String.format(
+                            "Bad request: one or more of datacenters passed (%s) invalid",
+                            dcMap.keySet()),
+                        Response.Status.BAD_REQUEST)
+                    : failure)
         // No real contents; can ignore ResultSet it seems and only worry about exceptions
         .map(any -> restResponseCreatedWithName(keyspaceName));
   }
--- a/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2KeyspacesResourceImpl.java
+++ b/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2KeyspacesResourceImpl.java
@@ -140,12 +140,6 @@ public class Sgv2KeyspacesResourceImpl extends RestResourceBase
     if (t instanceof StatusRuntimeException) {
       StatusRuntimeException failure = (StatusRuntimeException) t;
       Matcher m = BAD_DC_PATTERN.matcher(failure.getStatus().getDescription());
-      if (m.matches()) {
-        LOGGER.warn("KEYSPACE: match! " + m.matches() + " Group #1 = " + m.group(1));
-      } else {
-        LOGGER.warn(
-            "KEYSPACE: no match! Message to match [" + failure.getStatus().getDescription() + "]");
-      }
       return m.matches() && dcMap.containsKey(m.group(1));
     }
     return false;
--- a/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2KeyspacesResourceImpl.java
+++ b/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2KeyspacesResourceImpl.java
@@ -19,6 +19,7 @@ import com.fasterxml.jackson.core.json.JsonReadFeature;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.databind.json.JsonMapper;
+import io.grpc.StatusRuntimeException;
 import io.smallrye.mutiny.Uni;
 import io.stargate.bridge.proto.QueryOuterClass.Query;
 import io.stargate.bridge.proto.Schema;
@@ -30,9 +31,13 @@ import io.stargate.sgv2.restapi.service.models.Sgv2NameResponse;
 import io.stargate.sgv2.restapi.service.models.Sgv2RESTResponse;
 import io.stargate.sgv2.restapi.service.resources.RestResourceBase;
 import java.io.IOException;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 import org.jboss.resteasy.reactive.RestResponse;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


tatu-at-datastax:
Filed https://datastax.jira.com/browse/DSP-23015 for DSE since I think the handling is broken somewhere; possibly by DSE, but maybe Java driver or even Stargate DSE persistence backend. But seems reasonable to start with DSE and maybe get push back.
No point in filing an issue against C-3.11 I doubt that would be addressed.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2KeyspacesResourceImpl.java
```diff
@@ -100,8 +109,9 @@ public Uni<RestResponse<Sgv2NameResponse>> createKeyspace(final String payloadSt
               .ifNotExists()
               .withReplication(Replication.simpleStrategy(ksCreateDef.replicas))
               .build();
+      dcMap = Collections.emptyMap();
```
ivansenic, line 112: is this option tested now with empty map?

ivansenic (approved):

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2KeyspacesResourceImpl.java
tatu-at-datastax: There is an existing test for simple creation so this code path is covered to a degree -- but empty map is just used for convenience so that `failureMatchesBadDC()` need not to check special case.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2KeyspacesResourceImpl.java
```diff
@@ -42,6 +47,9 @@ public class Sgv2KeyspacesResourceImpl extends RestResourceBase
     implements Sgv2KeyspacesResourceApi {
   private static final Logger LOGGER = LoggerFactory.getLogger(Sgv2KeyspacesResourceImpl.class);
 
+  private static final Pattern BAD_DC_PATTERN =
```
tlasica, line 50: I think this is not correct.
If the DC name is passed without surrounding single quotes you will get this.
But if it is passed with single quotes then the message will be completely different:

```
cqlsh> create keyspace zupa WITH replication = {'class': ' = {'class': 'NetworkTopologyStrategy', dc1: 1};
NetworkTopologyStrategy  SimpleStrategy           
cqlsh> create keyspace zupa WITH replication = {'class': 'NetworkTopologyStrategy', 'xyz': 1};
ConfigurationException: Unknown data-center name 'xyz' passed to NetworkTopologyStrategy for keyspace 'zupa': it is either unknown to the configured snitch, or has no active member (known DCs: {Cassandra: 3 nodes})
cqlsh> create keyspace zupa WITH replication = {'class': 'NetworkTopologyStrategy', xyz: 1};
SyntaxException: line 1:80 no viable alternative at input ':' (... = {'class': 'NetworkTopologyStrategy', [xyz]:...)
```

tatu-at-datastax: @tlasica Keep in mind that the exception I saw was from Cassandra-4.0, not DSE. And Cassandra-3.11 did not even response with a failure. I did not realize fail messages vary that much and Stagate tests default to C-4.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2KeyspacesResourceImpl.java
```diff
@@ -111,11 +121,32 @@ public Uni<RestResponse<Sgv2NameResponse>> createKeyspace(final String payloadSt
...
+                        String.format(
+                            "Bad request: one or more of datacenters passed (%s) invalid",
+                            dcMap.keySet()),
+                        Response.Status.BAD_REQUEST)
+                    : failure)
         // No real contents; can ignore ResultSet it seems and only worry about exceptions
         .map(any -> restResponseCreatedWithName(keyspaceName));
   }
 
+  private boolean failureMatchesBadDC(Throwable t, Map<String, Integer> dcMap) {
```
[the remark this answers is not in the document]
tatu-at-datastax: @tlasica Yes, I am not happy with having to rely on failure message, but at least match is specific so there should not be false positives.

But the real problem in checking for existence is that code here has no direct access to this information: all it can do is talk to Stargate Coordinator via Bridge API that only allows queries. I didn't think DC information was available through queries (perhaps it is); but more generally it is simpler operationally to attempt an operation, only validating what can be done locally and react to failure messages.
This is one of challenges with Stargate V2, now that APIs are separate from coordinators so there is much less access to specifics of persistence backend.

I agree that it'd be great to have simple access to existing DCs to pre-validate this particular case.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2KeyspacesResourceImpl.java
tlasica: So it seems different backends can have different error message.

Maybe instead of trying to find most robust pattern it would be better to rely on the fact that this is `ConfigurationException`? And in such case I would simply pass whatever is after it. This would work well for every configuration exception.

I understand that this may be too "technical" but it will also be way easier to debug in case of a customer reported problem.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2KeyspacesResourceImpl.java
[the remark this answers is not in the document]
tlasica: I have proposed some simplified approach to just treat `ConfigurationExceptions` in the special way instead of trying to find a specific "unknown DC" one.

I am sorry, I did not realize that there is no access to the information about the DCs.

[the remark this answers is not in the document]
tatu-at-datastax: @tlasica No I appreciate your help & suggestions. Handling also changed a lot from Stargate V1 (where APIs had direct access in-process to backend) to V2 (where gRPC call is needed).

Another problem then is that API does not get `ConfigurationException` directly but something Bridge gRPC translates. Worse, Bridge handles all queries about the same (no separate end point for creating keyspaces just general "executeQuery()") so adding better translation is tricky.

I'll keep the ideas in mind and I hope at some point we can improve on this. Default handling will just pass exception message REST API gets to caller: perhaps that could suggest that DC information should be double-checked.

Finally: it's all bit moot wrt Astra too: turns out Stargate keyspace creation is blocked currently so users will actually get very different fail message from OSS Stargate :)

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2KeyspacesResourceImpl.java
tatu-at-datastax: Due to use of Bridge gRPC which handles all queries (data, schema) same way, I don't think it is possible to rely on `ConfigurationException` -- this type will not be available on REST API side.
Bridge may get this (although due to intermediate Persistence API that Stargate implements, I am not sure it gets this either).

So at this point I think this may have to do. I'd be happy to find a better way tho.

I have filed issue against DSE backend (3.11 will not really be further developed so seems pointless), so if that gets tackled this can be revisited.

PR #2304, opened 2022-12-09
revising usage of FAILED_PRECONDITION error status

jeffreyscarpenter:
We're using status message `FAILED_PRECONDITION` inappropriately in the gRPC module.

According to https://grpc.github.io/grpc-java/javadoc/io/grpc/Status.Code.html, this code should be used when the state of the system does not allow the operation but could be modified. This PR changes usage in the `Values` class to `INVALID_ARGUMENT` and usage in `ExceptionHandler` class to `INTERNAL`. 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/HAProxyProtocolDetectingDecoder.java
+++ b/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/HAProxyProtocolDetectingDecoder.java
@@ -75,6 +75,7 @@ public class HAProxyProtocolDetectingDecoder extends ByteToMessageDecoder {
     return true;
   }
 
+  @Sharable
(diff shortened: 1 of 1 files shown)


tatu-at-datastax:
What does Cassandra failure of `CONFIG_ERROR` usually relate to? If I remember correctly, trying to follow it through, this would be error of request, not that of configuration of Cassandra -- that is, a user error.
If so, why would we map to "Internal error" which (to me at least) implies something wrong in Stargate's (or Cassandra's) internal processing? Wouldn't it rather need to map to something indicating client/user error?
I think I am missing/misunderstanding something here.

ivansenic (approved):

tatu-at-datastax (approved):

PR #2306, opened 2022-12-13
Document version support and branching strategy

jeffreyscarpenter:
Adds to the project README to clarify our planned support model for major version number releases and the related branching strategy. This is written to be major version agnostic such that we could make the same change to both `v1` and `main` branches.

This addresses a portion of #2294

**Checklist**
- [ ] Changes manually tested


ivansenic:
What about the suggestion I had about the minor version increase? That this would denote that coordinator and APIs are not any more compatible.. Meaning that you can use any version of the coordinator and API that have same major and minor version without issues?

jeffreyscarpenter:
> What about the suggestion I had about the minor version increase?

Good suggestion, added a paragraph

tatu-at-datastax (approved):

PR #2307, opened 2022-12-15
Fix for C2-1835

versaurabh:
**What this PR does**:

Enforces guardrail checks by Stargate for Astra classic & AD4D

- via making Stargate pass the actual username to the backend DSE for Astra classic and AD4D clusters
- the backend DSE for Astra classic & AD4D clusters use `AstraAuthenticator` which own Authn/z for the the actual username identified by the token (via https://github.com/riptano/cloud-ondemand/pull/4186)

**Deployment ordering**
1. The changes in this PR should be deployed _after_ the corresponding changes in `cloud-ondemand` https://github.com/riptano/cloud-ondemand/pull/4186

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
+++ b/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
@@ -55,7 +55,6 @@ import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import org.apache.cassandra.auth.AuthenticatedUser;
-import org.apache.cassandra.auth.CassandraRoleManager;
 import org.apache.cassandra.auth.IAuthContext;
 import org.apache.cassandra.auth.RoleResource;
 import org.apache.cassandra.auth.user.UserRolesAndPermissions;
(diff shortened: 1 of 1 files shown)


jeffreyscarpenter:
looks like CI failures on this PR are causes by formatting errors, which you can correct by running 

`mvn com.coveo:fmt-maven-plugin:format`

ivansenic (changes requested):
Is this backward compatible change? I see this as a very good change and I am all :+1: for it, but I am afraid if this is actually work in the clusters that already exist? Are we certain that all the client names in the tokens are actually having a correct user in the DB (with all permissions needed).

Also I am not sure if this is actually dismissing the concept of the `TRANSITIONAL_AUTH`? Maybe a safer solution would be here to do something like:

```
 if (user.isFromExternalAuth()) {
   if (USE_TRANSITIONAL_AUTH) {
         loginSingle = this.clientState.login(EXTERNAL_AUTH_USER);
   } else {
         loginSingle = this.clientState.login(new ExternalAuthenticatedUser(user.name()));
   }
}
```

But then this can cause the different deployment configs, so could be even more confusing in a way..

@jeffreyscarpenter wdyt?

I am marking with *Request changes* until we figure it out..

jeffreyscarpenter:
@kathirsvn might have better perspective to comment on Ivan's concerns?

versaurabh:
> the client names in the tokens are actually having a correct

hi @ivansenic AFAIK, each token is guaranteed to be present in the OPA backend alongwith the permissions info over there. 
For Astra instance, the DSE authn/z is uses OPA backend via https://github.com/riptano/dse-astra-auth-lib/

jeffreyscarpenter:
@kathirsvn can you take a look at this and recommend next step?

PR #2315, opened 2022-12-20
no default int test profile in the sgv2 apis

ivansenic:
**What this PR does**:
We realized that having a maven profile `cassandra-40` activated by default, bounds any child module also to have this profile as a default one. There are situations where child modules are made for only a specific backend and thus we need to relax this a bit. Unfortunately making a profile not default in a child module is not work. Thus, a solution:

* `cassandra-40` is not marked as default anymore
* properties from `cassandra-40` are now set directly in the pom and can be overwritten by any child module in what ever fashion wanted (profile, props, etc)

### DIFF
--- a/apis/pom.xml
+++ b/apis/pom.xml
@@ -27,6 +27,17 @@
     <skipTests>false</skipTests>
     <skipUnitTests>${skipTests}</skipUnitTests>
     <surefire-plugin.version>3.0.0-M7</surefire-plugin.version>
+    <!--
+      Integration test properties, defaults to cassandra-40
+      Please maintain default values for int-test in sync with io.stargate.sgv2.docsapi.testresource.StargateTestResource.Defaults
+     -->
+    <stargate.int-test.cassandra.image>cassandra</stargate.int-test.cassandra.image>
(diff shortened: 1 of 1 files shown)


ivansenic:
@tatu-at-datastax can you check this and see if there is any better option that this one..

apis/pom.xml
```diff
@@ -193,19 +204,6 @@
   <profiles>
     <profile>
       <id>cassandra-40</id>
```
ivansenic, line 206: kept this for backward compatibility.. in case any command uses `-P cassandra-40`, I don't want to break it..

tatu-at-datastax: Yeah makes sense to me. Could add a comment explaining that it's left for backwards compatibility, used to be default.

tatu-at-datastax:
> @tatu-at-datastax can you check this and see if there is any better option that this one..

This sounds like a reasonable way to do it, I can't think of anything obviously better. So let's do this.

I think this is only needed for SGv2 and not v1 so makes sense to against `main`.

tatu-at-datastax (approved):

apis/pom.xml
ivansenic: Comment added.

tatu-at-datastax (approved):

PR #2319, opened 2022-12-21
Make inclusion of /v2/cql optional for REST API, with sysprop `stargate.rest.cql.disabled` (use Quarkus 2.16 `@EndpointDisabled`)

### LINKED ISSUE #2316: Add a way to opt-in to new `/v2/cql` endpoint (via REST API) -- using Quarkus `@EndpointDisabled` via `stargate.rest.cql.disabled`
(follow-up to #2313 )

Since it is possible that only some users may want to expose the new "raw CQL" endpoint for REST API, let's make its inclusion dynamic, based on a feature flag (Quarkus property that can be set via system props and/or env variables).

tatu-at-datastax:
**What this PR does**:

Makes the new CQL endpoint be optional included, depending on value of Boolean system property `stargate.rest.cql.enabled` (and its ENV variable counterpart, `STARGATE_REST_CQL_ENABLED`): value of `true` enables; any other value disables.
Default is that the endpoint is disabled (not included)

NOTE: should use new `@EndpointDisabled` Quarkus 2.16 adds, once 2.16.0 released (see https://github.com/quarkusio/quarkus/pull/30051 for details)

Fixes #2316

**Checklist**
- [x] Changes manually tested


tatu-at-datastax:
Unfortunately approach does not work: CQL-endpoint tests should be failing now, they are not.
Looks like things do not work due to either bug or limitation wrt handling of `@LookupIfProperty` annotation.

I filed https://github.com/quarkusio/quarkus/issues/30022 to see if this might be considered a bug or just limitation.
Otherwise not sure if and how this could be achieved.

tatu-at-datastax:
May decide to change to just fail at endpoint level -- but for now will not proceed until we know this is actually needed.

ivansenic:
Let s wait for https://github.com/quarkusio/quarkus/pull/30051

tatu-at-datastax:
Note: has to wait until https://github.com/quarkusio/quarkus/pull/30051 (now merged) gets released as part of Quarkus 2.16.0

apis/sgv2-restapi/CONFIGURATION.md
```diff
@@ -0,0 +1,22 @@
+# Configuration Guide
+
+> **IMPORTANT:** Please check the [Stargate Common Configuration](../sgv2-quarkus-common/CONFIGURATION.md) for properties shared between all Stargate V2 APIs.
+
+## Stargate REST API Configuration
```
ivansenic, line 5: is there no other restapi prop in the whole project?

ivansenic (approved):
Perfect :ship:

apis/sgv2-restapi/CONFIGURATION.md
tatu-at-datastax: Let me go check, good point.

tatu-at-datastax: I don't see any, looking at `application.yaml`s under `sgv2-restapi`. There is at least one for SGv1 REST API (in-coordinator one) for inclusion of RESTv1 but nothing for SGv2 REST portion that I remember.

I'll keep this in mind if I mind something and make sure to update this page.

jeffreyscarpenter: Make sure to also update the README, which currently says:

```
The REST API currently does not define any additional properties beyond these common properties.
```

tatu-at-datastax: > Make sure to also update the README, which currently says:
> 
> ```
> The REST API currently does not define any additional properties beyond these common properties.
> ```

Changed. Thanks!

PR #2324, opened 2022-12-23
closes #2321: delete keyspaces on the integration test end

### LINKED ISSUE #2321: Delete keyspaces after each integration test
We leave a lot of keyspaces after running integration tests. Although this is making the tests faster, we should actually consider deleting them, so that we don't end up with warnings like:

```
WARN  [s26-admin-0] 2022-12-22 16:50:06,008 CqlRequestHandler.java:398 - Query '[0 values] CREATE KEYSPACE IF NOT EXISTS "ks9907982729542956" WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1}' generated server side warning(s): Cluster already contains 52 keyspaces. Having a large number of keyspaces will significantly slow down schema dependent cluster operations.
```

ivansenic:
**What this PR does**:
Running all tests on a single env creates a lot of keyspaces, thus delete after each test. CI times should remain more or less the same.

Fixes #2321

### DIFF
--- a/apis/sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/v2/DocsApiIntegrationTest.java
+++ b/apis/sgv2-docsapi/src/test/java/io/stargate/sgv2/docsapi/api/v2/DocsApiIntegrationTest.java
@@ -117,4 +117,24 @@ public abstract class DocsApiIntegrationTest {
               });
     }
   }
+
+  @Nested
+  @Order(Integer.MAX_VALUE)
+  class CleanUp {
+
+    @Test
+    public void deleteNamespace() {
+      createNamespace()
+          .ifPresent(
+              namespace -> {
+                // delete
+                given()
+                    .contentType(ContentType.JSON)
+                    .header(HttpConstants.AUTHENTICATION_TOKEN_HEADER_NAME, getAuthToken())
+                    .delete(NamespacesResource.BASE_PATH + "/{namespace}", namespace)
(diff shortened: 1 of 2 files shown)


jeffreyscarpenter (approved):

tatu-at-datastax (approved):

ivansenic:
@tatu-at-datastax Would this also be enough for rest tests?

tatu-at-datastax:
> @tatu-at-datastax Would this also be enough for rest tests?

I was wondering about that, it'd be good to clean up if number of keyspaces becomes an issue (which apparently it does).

But it would need changes I think since REST tests create separate keyspace for every test method (not just per class) and so either deletion should be after every test, or iterating over all keyspaces (with given prefix?) after all tests are run.

ivansenic:
@tatu-at-datastax I am going to merge this one and then create a separate issue for rest tests.. I am not sure why would you need a keyspace per test method, is this a requirement?

tatu-at-datastax:
> @tatu-at-datastax I am going to merge this one and then create a separate issue for rest tests.. I am not sure why would you need a keyspace per test method, is this a requirement?

It is inherited from existing tests (that is, it has always been done this way since I started): it allows all tests start from an empty dedicated keyspace without having to clean up after test. For some tests this doesn't really matter (row CRUD ones create test-specific tables anyway), for others it does (creating tables, testing to see keyspace only has table test method created and nothing else).

PR #2330, opened 2022-12-27
relates to #2195: remove api source from schema read

### LINKED ISSUE #2195: Remove source api from schema reads
Use the value passed in the metadata and intercepted by the bridge.

ivansenic:
**What this PR does**:

Fixes #2195

**Checklist**
- [x] APIs updates
- [x] Ensure deprecated works
- [ ] Talk about breaking changes (`2.1.0-SNAPSHOT` maybe?)
- [x] Add tests (insane that there was no single test for this)


jeffreyscarpenter (approved):

ivansenic:
@jeffreyscarpenter this is still a draft.. some more changes are coming..

ivansenic:
@jeffreyscarpenter I made a field in the protos deprecated.. My intention is that I don't remove it immediately, but in one of the future releases (although the field is effectively ignore now). We should discuss the bridge breaking code changes and how can we better "announce" them using versioning. Thus I was thinking that once we break bridge api, we increase minor version?

jeffreyscarpenter:
> @jeffreyscarpenter I made a field in the protos deprecated.. My intention is that I don't remove it immediately, but in one of the future releases (although the field is effectively ignore now). We should discuss the bridge breaking code changes and how can we better "announce" them using versioning. Thus I was thinking that once we break bridge api, we increase minor version?

I agree with the approach you describe - deprecating fields we don't intend to use, and incrementing minor version based on bridge API changes. However, I also want to say that I don't think we need to be in any particular hurry to remove the deprecated field and force a breaking API change. If nothing is hurt by having the deprecated field remain unused, why rush?

Issue #2192, opened 2022-10-24
Add deadline for the client side gRPC in the V2


ivansenic:
* Deadlines should be a `quarkus.grpc.clients` prop configuration
* We need to confirm that once occurred, we map it correctly and send back the proper JSON back to the use

ivansenic:
Marking as done, deadlines can be define by property and status code is corrected in #2320.

PR #2331, opened 2022-12-27
add default grpc client deadline of 10sec

ivansenic:
**What this PR does**:
Checks if all INT tests are green if client deadline is used.

Relates to #2192.


ivansenic:
Seems that this is stub deadline and not the call deadline.. The documentation on Quarkus was wrong. We need to decide if we want to implement our own timeouts? Then we can add them using the `CallOptions`.

ivansenic:
Closing this, as this was a test only to showcase the issue: https://github.com/quarkusio/quarkus/issues/30082

PR #2333, opened 2022-12-28
ux improvements to docker compose configs

jeffreyscarpenter:
Usability improvements to docker compose scripts

- incorporated health checks into container definitions so that we don't have to rely on the `start_` scripts invoking `docker compose` multiple times and scraping log files to know when it's safe to start next containers (inspiration from @ivansenic)
- based on this, users who ignore READMEs and just run `docker compose up` will have a good default experience. The `start_` scripts are now described as "convenience" scripts in the READMEs.
- environment now defaults to latest minor version tag for Stargate, Cassandra 3.11, and 4.0 images (minor version tags not available for DSE)

**Checklist**
- [x] Changes manually tested
- [ ] Automated Tests added/updated (planning to add automated tests that exercise these in a separate PR)


docker-compose/README.md
```diff
@@ -37,6 +37,31 @@ To enable access to these ports from outside the Docker network, you'll need to
 ## Docker Images
 Stargate Docker images are publicly available on [Docker Hub](https://hub.docker.com/r/stargateio/). To learn how the images are built or to build your own local versions, see the [coordinator node developer guide](../DEV_GUIDE.md) or API Service developer guides (under the [apis](apis) directory).
 
+The fastest way to build your own local images involves the following steps:
+
+* Local build of coordinator images
+  * Make sure you are in the repository root directory and have `JAVA_HOME` set to point to a JDK 1.8 installation
+  * Do a local build of jar files for coordinator:
+    ```
+    ./mvnw clean install -DskipTests
```
ivansenic, line 46: missing `-P dse`, otherwise the dse will not be built, and the image building in the next step will not contain needed jars, which would make image not startable

docker-compose/README.md
```diff
@@ -37,6 +37,31 @@ To enable access to these ports from outside the Docker network, you'll need to
...
+    ```
+  * Generate docker images (image tag will default to the Stargate version specified in the `pom.xml`):
+    ```
+    ./mvnw clean install -DskipTests
```
ivansenic, line 50: you don't need another another build before running the docker build script

docker-compose/cassandra-3.11/.env
```diff
@@ -1,4 +1,5 @@
 COMPOSE_PROJECT_NAME=cass311-stargate
-CASSTAG=3.11.13
+CASSTAG=3.11
```
ivansenic, line 2: just to confirm `3.11` always points to latest? This is good, but could end up into situation where docker is started with the non-supported version. for example coordinator supports `3.11.13` and C* moves to `3.11.14`.. But maybe this is also fine, especially if we can have this in a CI that could run stuff on top of this?

docker-compose/cassandra-3.11/README.md
```diff
@@ -1,27 +1,38 @@
-This directory provides two ways to start Stargate with Cassandra 3.11 using docker-compose.
+# Docker Compose scripts for Stargate with Cassandra 3.11
```
ivansenic, line 1: Do you think it would be valuable to add a debugging info like this one to all readmes:

```
## Debugging

If you want to debug a docker container, add following debug arg to the `JAVA_OPTS` of the target container:

`-agentlib:jdwp=transport=dt_socket,address=host.docker.internal:8000,server=n,suspend=n`

Then create a `LIsten to remote JVM` debugging run configuration in IntelliJ that points to the port 8000.
First start the debugger and then hit `docker-compose up`.
```

Well, this is for coordinator mainly, not sure anymore

docker-compose/cassandra-3.11/README.md
```diff
@@ -31,25 +42,22 @@ Both scripts support the following options:
...
-* The `.env` file defines variables for the docker compose project name (`COMPOSE_PROJECT_NAME`),
-  and the Cassandra Docker image tag to use (`CASSTAG`).
+* The `.env` file defines variables for the docker compose project name (`COMPOSE_PROJECT_NAME`), the Cassandra Docker image tag to use (`CASSTAG`), and the Stargate Docker image tag to use (`SGTAG`).
 
-* The Stargate Docker image tag to use (`SGTAG`) defaults to the current version as defined in the
-  top level project `pom.xml` file. It can be overridden with the `-t` option on either script:
+* When using the convenience scripts, the Stargate Docker image (`SGTAG`) used for the `-l` option is the current (snapshot) version as defined in the top level project `pom.xml` file. It can be overridden with the `-t` option on either script:
 
-  `./start_cass_311.sh -t v2.0.0-BETA-2`
+  `./start_cass_311.sh -t v2.0.1`
```
ivansenic, line 51: I would say we don't need the start scripts anymore? Ah OK, there are something there like request logs etc, but this would all be part of the `.env` now?

docker-compose/cassandra-3.11/docker-compose.yml
```diff
@@ -10,34 +10,55 @@ services:
...
+      retries: 10
+
   cassandra-3:
     image: cassandra:${CASSTAG}
     networks:
       - stargate
     mem_limit: 2G
     depends_on:
-      - cassandra-2
+      cassandra-2:
```
ivansenic, line 44: * can't we wait on `cassandra-1` only, would be way faster to startup
* and actually, I think 2 storage nodes are perfectly fine here, why are we adding a third one?

ivansenic (approved):
Added few suggestions, if you apply them, please apply to all the backends (I only added comments to 3.11, but it applies everywhere)..

I confirmed I can run the docker-compose with C* 4.0 without any issues nor prerequisites..

docker-compose/README.md
jeffreyscarpenter: good catch, thanks

docker-compose/README.md
jeffreyscarpenter: yeah, copy-paste error

docker-compose/cassandra-3.11/.env
jeffreyscarpenter: By definition a Cassandra `3.11.14` node should be backward compatible. We can revisit if this proves not to be the case. My thought was that we don't remember to update this file regularly when Cassandra patch version increases, so let's take that off the table.

docker-compose/cassandra-3.11/README.md
jeffreyscarpenter: They provide value as a convenience, especially the `-l` option.

docker-compose/cassandra-3.11/docker-compose.yml
jeffreyscarpenter: I believe the best practice is to add a single node to a cluster at a time which is why `cassandra-3` depends on `cassandra-2`.

I preserved the convention we've traditionally had in docker compose files of having a three node cluster. I don't have a strong opinion about that.

ivansenic (approved):

PR #2335, opened 2022-12-29
organize shared production and dev props

ivansenic:
**What this PR does**:
Tries to organize the shared production/dev properties in one place (common project).


apis/sgv2-quarkus-common/src/main/resources/application.yaml
```diff
@@ -152,3 +152,15 @@ quarkus:
               datastore:
                 configuration:
                   DataStorePropertiesConfiguration/fetchSupportedFeatures/Retry/maxRetries: 0
+
+# shared production properties
+"%prod":
```
ivansenic, line 157: I am actually not sure if I like the mixup of the production and development properties in one file.. I am used to have

* `application-dev.yaml`
* `application-prod.yaml` 

which clearly separates those.. The ``%dev` and `%prod` notation from quarkus is mixing everything in one file.. Should we change?

apis/sgv2-quarkus-common/src/main/resources/application.yaml
```diff
@@ -152,3 +152,15 @@ quarkus:
...
+
+  quarkus:
+
+    # OpenTelemetry configuration
+    opentelemetry:
```
ivansenic, line 162: agreed to disable open telemetry by default in production @kathirsvn

apis/sgv2-quarkus-common/src/main/resources/application.yaml
tatu-at-datastax: I agree that it gets bit messy. Assuming `application.yaml` has shared defaults, and overrides would be in `application-prod.yaml` (or -dev or -test, depending), that seems a bit better to me.
But I can live with either approach.

apis/sgv2-quarkus-common/src/main/resources/application.yaml
tatu-at-datastax: How is it disabled here? Isn't this just setting sampler, but leaving it enabled (same as before change)?

ivansenic: It's not, we want to do it by default.. For now in prod it's disabled..

ivansenic: Actually, I can configure this without a `prod` profile at all..

tatu-at-datastax: Ah. I like this, cool.

tatu-at-datastax (approved):

PR #2342, opened 2023-01-04
Jeff/fix postman ci

jeffreyscarpenter:
various attempts at getting this workflow to execute. Still getting socket close errors on test runs

https://github.com/stargate/stargate/actions/workflows/postman-docker.yaml


.github/workflows/postman-docker.yaml
```diff
@@ -22,38 +22,56 @@ jobs:
...
           - backend: cassandra-311
             path: docker-compose/cassandra-3.11
           - backend: dse-68
             path: docker-compose/dse-6.8
-          - collection: docsapi-library
-            collection-id: 12949543-caba2a02-6559-486a-9e4a-d5c0791fd296
-            environment-id: 17930693-2f355f7b-123d-4bca-9a9c-7d82222ca49d
+          #- collection: docsapi-library
+          #  collection-id: 12949543-caba2a02-6559-486a-9e4a-d5c0791fd296
+          #  environment-id: 12949543-2e78cf27-bd8c-43f2-909f-70a2b87d65fe
```
jeffreyscarpenter, line 35: Temporarily commenting this one out while I work with Lorina on it running successfully to completion (it has extra keyspace creation commands that cause errors).

.github/workflows/postman-docker.yaml
```diff
@@ -22,38 +22,56 @@ jobs:
...
           - collection: graphqlapi-library
             collection-id: 17930693-65da5c64-561a-449b-a0e8-0318575f6871
-            environment-id: 17930693-2f355f7b-123d-4bca-9a9c-7d82222ca49d
+            environment-id: 12949543-2e78cf27-bd8c-43f2-909f-70a2b87d65fe
```
jeffreyscarpenter, line 38: use correct environment "Stargate v2 OSS API Environment LIBRARY"

.github/workflows/postman-docker.yaml
```diff
@@ -22,38 +22,56 @@ jobs:
...
         run: |
           cd ${{ matrix.path }}
           docker compose -f docker-compose-dev-mode.yml up -d
-      - name: Run API test
-        run: |
-          postman collection run ${{ matrix.collection-id }} -e ${{ matrix.environment-id }}
+
+      #- name: Run API test
+      #  run: |
+      #    postman collection run ${{ matrix.collection-id }} -e ${{ matrix.environment-id }} --verbose --timeout-request 10000
```
jeffreyscarpenter, line 61: Currently commented out while I try Newman (OSS CLI for Postman). 
I have tried various settings. Verbose flag allowed me to see the socket close message. Increasing timeout seems not to fix the problem

.github/workflows/postman-docker.yaml
```diff
@@ -22,38 +22,56 @@ jobs:
...
+      - name: Run API Test with Newman
+        id: run-newman
+        uses: anthonyvscode/newman-action@v1
+        with:
+          apiKey: ${{ secrets.POSTMAN_API_KEY }}
+          collection: ${{ matrix.collection-id }}
+          environment: ${{ matrix.environment-id }}
+          reporters: cli
+
+      - name: Output summary to console
```
jeffreyscarpenter, line 72: [Documentation](https://github.com/marketplace/actions/newman-cli-postman-action) says to include this but it doesn't do anything. Should remove.

docker-compose/cassandra-3.11/docker-compose-dev-mode.yml
```diff
@@ -8,7 +8,6 @@ services:
     ports:
       - "9042:9042"
       - "8081:8081"
-      - "8084:8084"
```
jeffreyscarpenter, line 11: GH Action fails saying the port is already in use. Must be something running in that environment but was not able to determine what. So I just removed this port mapping temporarily to make progress.

docker-compose/cassandra-3.11/docker-compose-dev-mode.yml
```diff
@@ -36,9 +35,9 @@ services:
       - "8082:8082"
     mem_limit: 2G
     environment:
-      - QUARKUS_HTTP_CORS=true
-      - QUARKUS_HTTP_CORS_ORIGINS=*
-      - QUARKUS_HTTP_CORS_METHODS=POST,GET,PUT,OPTIONS,DELETE
+      #- QUARKUS_HTTP_CORS=true
+      #- QUARKUS_HTTP_CORS_ORIGINS=*
+      #- QUARKUS_HTTP_CORS_METHODS=POST,GET,PUT,OPTIONS,DELETE
```
jeffreyscarpenter, line 40: Also tried disabling the cors filtering to see if that was the cause of the socket close, but does not seem to be the issue.

jeffreyscarpenter:
@ivansenic when I run the GH action on my Mac desktop using [Act](https://github.com/nektos/act), the Postman CLI is able to run the collections with success (aside from an error in the docs collection which we are working through).

However when running in GitHub, every call to one of the API services gets a socket close error. You'll see things in the action output like `read ECONNRESET` or `socket hang up`. See for example https://github.com/stargate/stargate/actions/runs/3834690686/jobs/6527308192#step:6

Interestingly, the initial call to the Auth API on the coordinator succeeds which makes me suspect there's something Quarkus related.

.github/workflows/postman-docker.yaml
```diff
@@ -22,38 +22,56 @@ jobs:
...
           docker compose -f docker-compose-dev-mode.yml up -d
-      - name: Run API test
-        run: |
-          postman collection run ${{ matrix.collection-id }} -e ${{ matrix.environment-id }}
+
+      #- name: Run API test
+      #  run: |
+      #    postman collection run ${{ matrix.collection-id }} -e ${{ matrix.environment-id }} --verbose --timeout-request 10000
+
+      - name: Run API Test with Newman
```
jeffreyscarpenter, line 63: Switched over to trying this at @ivansenic suggestion. Seems to be pretty equivalent to Postman CLI in how it works. Get the same socket close errors though.

ivansenic:
@jeffreyscarpenter But where do you see that the every call fails, as  I see many calls are `200` but then assertion fails.. They said here that content-length header should be set: https://stackoverflow.com/a/63489537.. I am also thinking we should only expose needed ports for this test, maybe there is something with that (but doubth it).. I am pritty sure it's postman itself.. We are already running the INT tests in this way for a long time, there's no chance there's something with Quarkus imo..

jeffreyscarpenter:
@ivansenic I have updated the PR to just use curl to access the REST API, still fails. I don't think the issue is only with Postman. There is something going on in terms of how the endpoints are being exposed.

ivansenic:
everything works fine on my machine.. can you try with `ubuntu-20.04`, maybe something is off in the `ubuntu-22.04`?

jeffreyscarpenter:
I figured out why the test was failing, it was a timing issue. Postman was calling the APIs before they were fully initialized.

This is a pretty messy PR with lots of experimentation do I am closing this and doing a new clean pr #2427

PR #2344, opened 2023-01-05
(WIP/Blocked) Update Ubuntu version from deprecated 18.x for "coordinator-test" GH Action (v1)

tatu-at-datastax:
**What this PR does**:

Updates ref to deprecated Ubuntu version (18.04) with newer (20.04)

n/a

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/.github/workflows/coordinator-test.yml
+++ b/.github/workflows/coordinator-test.yml
@@ -6,8 +6,8 @@ on:
 jobs:
   integration-test:
     name: Integration Test
-    # CCM installation trickier on 20.04, alas; Python 2.x dep etc problematic
-    runs-on: ubuntu-18.04
+    # 05-Jan-2023, tatu: CCM install tested with this version, current ubuntu-latest:
+    runs-on: ubuntu-22.04
     strategy:
       # Defaults to "true" but let's let all runs finish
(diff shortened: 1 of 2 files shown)


.github/workflows/coordinator-test.yml
```diff
@@ -2,12 +2,14 @@ name: Coordinator Tests
 
 on:
   workflow_dispatch:
+  pull_request:
+    branches: [ "v1" ]
```
ivansenic, line 6: why only v1?

tatu-at-datastax: True, when enabling should be for `main` as well (and right now, will not matter). I added this for now to trigger CI runs for changes themselves.

NOTE: I converted this back to Draft since unfortunately CI/ITs start to fail will upgrade due to something going wrong (exception claim "too many Stargate nodes" which perhaps suggests some leak).

jeffreyscarpenter: Interesting. I did a separate experiment under #2378 and am seeing the same issue.

ivansenic:
Clsoing in favor of https://github.com/stargate/stargate/pull/2378.

@tatu-at-datastax OK?

tatu-at-datastax:
@ivansenic yes, makes sense. Thank you!

PR #2351, opened 2023-01-06
Avoid SchemaReadCheck optional for Row CRUD (leave for Schema access)

### LINKED ISSUE #2349: SGv2 REST API should NOT do `authorizeSchemaRead` for regular DML CRUD operations
As noted by @kathirsvn it looks like table metadata access for basic GET/POST/PUT/DELETE (CRUD) row operations invokes check for schema read access:

```
And regarding authorizeSchemaRead:
  findTableAsync method in executeQueryAsync of RestResourceBase gets the table information after  executing authorizeSchemaRead
```

This should not be necessary since authorization for actual Table being accessed will be performed as part of query processing: trying to invoke it before query will incur additional request/response overhead for all operations (whereas authz for query is done on Bridge side as part of query processing).

Authorization is still needed for Schema (DDL) access, just not for DML.

tatu-at-datastax:
**What this PR does**:

Changes schema-read-access check handling for table metadata access so that:

1. It WILL be performed when accessing Schema info directly (tables, indexes)
2. It WILL NOT be performed for Row CRUD: access check will be performed as part of actual query execution

The reason being that in either case query execution will still perform checks; but in case of schema operations (DDL) we can afford "double-checking" to catch issue earlier, despite higher overall latency.

Issue was found during performance testing

Fixes #2349

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java
+++ b/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java
@@ -71,15 +71,15 @@ public abstract class RestResourceBase {
   }
 
   protected Uni<Schema.CqlTable> getTableAsync(
-      String keyspaceName, String tableName, boolean checkIfAuthorized) {
-    return checkIfAuthorized
+      String keyspaceName, String tableName, boolean checkAuthzForTableMetadata) {
+    return checkAuthzForTableMetadata
         ? schemaManager.getTableAuthorized(keyspaceName, tableName, MISSING_KEYSPACE)
         : schemaManager.getTable(keyspaceName, tableName, MISSING_KEYSPACE);
   }
 
   protected Uni<Schema.CqlTable> getTableAsyncCheckExistence(
-      String keyspaceName, String tableName, boolean checkIfAuthorized, Response.Status failCode) {
-    return getTableAsync(keyspaceName, tableName, checkIfAuthorized)
+      String keyspaceName, String tableName, boolean checkAuthzForTableMetadata, Response.Status failCode) {
+    return getTableAsync(keyspaceName, tableName, checkAuthzForTableMetadata)
         .onItem()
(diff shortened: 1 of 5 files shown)


kathirsvn (approved):
LGTM

ivansenic (approved):
Do we need any tests that would actually confirm what kind of the bridge access is performed in certain situations? I know that integration tests remain the same and that's fine, but having such a major change without a single test changed or added brings up a red flag for me..

tatu-at-datastax:
> Do we need any tests that would actually confirm what kind of the bridge access is performed in certain situations? I know that integration tests remain the same and that's fine, but having such a major change without a single test changed or added brings up a red flag for me..

I agree that ideally we should have this, verification of a sequence of bridge calls.
But right now I don't think there are existing tests to build on so it is quite a big undertaking.
I will file an issue for adding such test(s) for REST API.

ivansenic:
> > Do we need any tests that would actually confirm what kind of the bridge access is performed in certain situations? I know that integration tests remain the same and that's fine, but having such a major change without a single test changed or added brings up a red flag for me..
> 
> I agree that ideally we should have this, verification of a sequence of bridge calls. But right now I don't think there are existing tests to build on so it is quite a big undertaking. I will file an issue for adding such test(s) for REST API.

I think there is a set of very good examples how you can write those tests in docsapi..

tatu-at-datastax:
> > > Do we need any tests that would actually confirm what kind of the bridge access is performed in certain situations? I know that integration tests remain the same and that's fine, but having such a major change without a single test changed or added brings up a red flag for me..
> > 
> > 
> > I agree that ideally we should have this, verification of a sequence of bridge calls. But right now I don't think there are existing tests to build on so it is quite a big undertaking. I will file an issue for adding such test(s) for REST API.
> 
> I think there is a set of very good examples how you can write those tests in docsapi..

Hmmh. Ok, that's good. I'll see if I can find one.

ivansenic:
@tatu-at-datastax Exactly

PR #2352, opened 2023-01-09
add client deadline to bridge interceptor

ivansenic:
**What this PR does**:
Adds client deadlines for each RPC call to the bridge. Configurable via the property.

Internal.

### DIFF
--- a/apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/GrpcConfig.java
+++ b/apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/GrpcConfig.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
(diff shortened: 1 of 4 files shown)


ivansenic:
@kathirsvn this is the Quarkus issue about the stub deadline: https://github.com/quarkusio/quarkus/issues/30082

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/GrpcConfig.java
```diff
@@ -0,0 +1,32 @@
...
+import io.smallrye.config.WithDefault;
+import java.time.Duration;
+import java.util.Optional;
+
+/** Configuration for the gRPC calls to the Bridge. */
+@ConfigMapping(prefix = "stargate.grpc")
+public interface GrpcConfig {
+
+  /** @return Optional deadline duration for the each RPC to the bridge. Defaults to 30 seconds. */
+  @WithDefault("PT30S")
```
tatu-at-datastax, line 30: Isn't 30 seconds kind of high?
I guess it is lower than what the ultimate default would be (or maybe there was no deadline specified at all).

Would think that something like 10 seconds might be better but maybe we might have some data to suggest something less than 30 seconds that is rarely seen.

tatu-at-datastax (approved):
Have one question on default value for grpc call deadline, but no strong objections to just merging. Makes sense to me & great to have a test to verify.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/GrpcConfig.java
ivansenic: * 30 seconds is what we agreed to be the value for now
* is it too much, maybe, but
   * better that not having at all (yes it's unlimited)
   * or to get timeout from the sidecar (60 sec default on nginx)
* and value is configurable, so you can set whatever you want when starting

tatu-at-datastax: Ok.

PR #2355, opened 2023-01-09
closes #1977: adds support for optimistic queries in the common

### LINKED ISSUE #1977: Implement optimistic bridge queries in `sgv2-quarkus-common`
Original non-Quarkus version: #1876.
We need to do the same in the `sgv2-quarkus-common` module introduced in #1965.

Note that only the REST API uses optimistic queries at the moment. So:
- if the REST API uses `sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/StargateBridgeClient.java`, then it needs to be done there (there is a TODO in `StargateBridgeClientImpl`).
- if the REST API uses the Mutiny gRPC API directly, then it needs to be done at that level, and the method can be removed in the interface above.

ivansenic:
**What this PR does**:
Enables the optimistic query execution against the bridge. 

The changes added to the `SchemaManager` are allowing to get a keyspace without hash validation. I added minimal changes there and not all methods have support for this atm. We can add later if needed.

The new service is fetching the keyspace from the `SchemaManager` without hash validation. This means if keyspace is hashed, we will simply return the cached version. If it's not cached, we will execute a bridge call nevertheless.

Once we have the keyspace, we find the table and pass it to the query function. We execute optimistic call and handle it in a way:
* No keyspace on bridge, going for missing keyspace result
* New keyspace, re-execute with new keyspace in mind
* None of the above, return the results :)

Usage should be:

```
optimisticQueryService.executeOptimistic(
  "keyspace", 
  "table", 
  k -> Mono.create.fromFailure(new MissingKeyspaceException(..)),
  t -> {
    if (null == t) {
      return Mono.create().fromFailure(new MissingTableException(..));
    }
   // impl query
   return Query.newBuilder()....build();
  }
```

Fixes #1977

**Checklist**
- [x] Add tests


apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/optimistic/OptimisticQueryService.java
```diff
@@ -0,0 +1,179 @@
...
+  @Inject private StargateRequestInfo requestInfo;
+
+  /**
+   * Executes the optimistic query, by fetching the keyspace with authorization from the @{@link
+   * SchemaManager}.
+   *
+   * @param keyspace Keyspace name.
+   * @param table Table name.
+   * @param missingKeyspace Function in case the keyspace in case it's not existing. Usually there
+   *     to * provide a failure.
```
tatu-at-datastax, line 47: nit: extra `*` in there :)

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/schema/SchemaManager.java
```diff
@@ -410,8 +453,8 @@ private Uni<Schema.CqlKeyspaceDescribe> getKeyspaceInternal(
                     .flatMap(
                         keyspace -> {
                           // if it was not cached before, we can simply return
-                          // no need to check
-                          if (!cached) {
+                          // same if we don't want to validate hash
```
tatu-at-datastax, line 456: Ok, logic here and above looks very complicated (and I know it already existed and is not changed here): first cache is accessed only to see if there is a cached entry and one that is not null (why would it be `null`? Do we cache "no such keyspace" entries?).
This produces `Boolean` (have in cache, true/false), instead of retaining Keyspace entry -- requiring separate cache lookup for some reason.
But then on line 407/450 we make call to `fetchKeyspace(ks, tenant, bridge)` which ... calls Bridge doesn't it?

So doesn't this still make actual call even if there's cached keyspace entry?

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/optimistic/OptimisticQueryService.java
```diff
@@ -0,0 +1,179 @@
...
+   * @param cqlUni Uni providing the keyspace.
+   * @param table Table name.
+   * @param missingKeyspace Supplier in case the keyspace in case it's not existing. Usually there
+   *     to provide a failure.
+   * @param queryFunction Function that creates the query from the {@link Schema.CqlTable}. Note
+   *     that the table can be <code>null</code>.
+   * @return Response when optimistic query is executed correctly.
+   */
+  public Uni<QueryOuterClass.Response> executeOptimistic(
+      Uni<Schema.CqlKeyspaceDescribe> cqlUni,
```
tatu-at-datastax, line 98: Maybe use name like `keyspace` or `uniKeyspace` instead?

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/optimistic/OptimisticQueryService.java
```diff
@@ -0,0 +1,179 @@
...
+      Function<Schema.CqlTable, Uni<QueryOuterClass.Query>> queryFunction) {
+    return response -> {
+
+      // * if no keyspace, calls the missing key space function
+      // * if there is new keyspace, execute again with that keyspace, use same handler
+      // * if everything is fine, map to result
+      if (response.hasNoKeyspace()) {
+        return missingKeyspace.get();
+      } else if (response.hasNewKeyspace()) {
+        // note that this is endlessly trying to execute the query in case there is always a changes
```
tatu-at-datastax, line 136: I think I'd be most worried about bug that indicates new keyspace returned. But to try to handle infinite repetition would probably require passing "number of attempts" via bridge to keep track... so not sure how to avoid.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/schema/SchemaManager.java
tatu-at-datastax: Looking further into this, `fetchKeyspace()` apparently does indirectly access cache due to annotations.
So I guess it would avoid call if cache entry exists.

tatu-at-datastax (approved):
Ok that `SchemaManager.getKeyspaceInternal()` is really complicated sequence of things (and already exists, not made worse here) and I had hard time understanding it. But I think I got it finally; there are at least 2 places where `describeKeyspace()` is called (either with hash to allow omitting returning schema if hashes match; or without hash to always get it), and then there's separate "look to see if I have it cached" separate from "get it from cache" (with seemingly possible race condition, but not very harmful one).

It would be great if that could be simplified, but I am not sure if that is possible.

Anyway, yeah, let's get this merged.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/schema/SchemaManager.java
ivansenic: Yes, unfortunately we do access cache twice.. That's because there is no way to know if the item was already in the cache or not.. And yes this can be improved :) Add a fix :+1:

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/optimistic/OptimisticQueryService.java
ivansenic: Yes, same thinking here, that's why I added the comment there.. Adding retry count is pritty simple, however, my main question would be what to do when we hit that last retry and we get a new keyspace then? Some kind of retries exhausted mapper than in addition?

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/schema/SchemaManager.java
ivansenic: Fixed in https://github.com/stargate/stargate/pull/2355/commits/6494741439a60be3bb2244fb3885d5543e72959b

I think it was worth it:

* decrease cache access
* and actually fix that extra fetch in error case; as you can see in the commit in case of an error when calling the bridge, this error will also be cached.. however old implementation counted errors as `cached=true`, so next time we try we go fetch the keyspace (because last try was error) and plus validate hash (because error was cached)..

ivansenic:
@tatu-at-datastax I added requested fixes, and implemented tests.. If you want to have another look before I merge?

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/optimistic/OptimisticQueryService.java
tatu-at-datastax: Right, good question. I think there should be an exception for failing to sync schema, plus count of retries that was used.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/optimistic/OptimisticQueryService.java
```diff
@@ -0,0 +1,182 @@
...
+        // keyspace
+        return executeWithKeyspace(response.getNewKeyspace(), table, queryFunction)
```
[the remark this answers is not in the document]
tatu-at-datastax: Ahh. That could explain number of cache misses.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/schema/SchemaManager.java
```diff
@@ -381,81 +424,95 @@ public Uni<Boolean> authorizeInternal(StargateBridge bridge, Schema.SchemaRead s
...
+                // otherwise push null uni down
+                return Uni.createFrom().nullItem();
               }
             })
+
+        // if it was not cached in first place, fetch from the keyspace from bridge
+        // and return with cached false
+        .onItem()
+        .ifNull()
+        .switchTo(() -> fetchKeyspace(keyspaceName, tenantId, bridge).map(k -> Tuple2.of(k, false)))
```
tatu-at-datastax, line 464: Good -- I much prefer this version to the earlier version! Much easier to read, follow what is happening. Plus avoids extra cache access too,

tatu-at-datastax (commented):
I like the changes! If you want to make further changes that's fine, but I'll approve this so it can be merged as-is as well.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/optimistic/OptimisticQueryService.java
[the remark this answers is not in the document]
ivansenic: No I mean in this new implementation.. But seems I need to move than the whole logic to the SchemaManager itself.. But that's fine, maybe then we don't even have to have those new methods in the SchemaManager, but only the impl stuff for the query with schema..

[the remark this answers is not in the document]
tatu-at-datastax: Ok looks good now.

PR #2332, opened 2022-12-28
update APIs to test against C* 4.0.7

ivansenic:
**What this PR does**:
Updates int test setup to use 4.0.7. Needs merging of `v1` and `main`. Followup work after #2329. 

**Checklist**
- [ ] Rebase after merge occurs..

PR #2251, opened 2022-11-21
dse version upgraded to 6.8.29

versaurabh:
**What this PR does**:
- Upgrades DSE version to 6.8.29

**Checklist**
- [ ] Changes manually tested
- [x] Automated Tests added/updated
- [ ] Documentation added/updated
- [x] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

PR #2357, opened 2023-01-09
Bump DSE version to 6.8.31

msmygit:
**What this PR does**: Bump DSE default version in environment file to `6.8.31`.

**Which issue(s) this PR fixes**: n/a
Fixes #<issue number>: n/a

**Checklist**
- [x] Changes manually tested


tatu-at-datastax:
Hi @msmygit! Thank you for submitting this PR. But there is much more than is needed to upgrade DSE version Stargate uses.

The best way to see what is needed is to check

    persistence-dse-6.8/README.md

and maybe have a look at a recent PR for upgrading DSE dependency, like #2251.

In addition, PR needs to be against `v1` branch, since we need it to be updated for both "old" Stargate v1 (which is still needed for many environments) and then merged (Stargate team can do this) into `main` for Stargate v2.

I hope this helps.

msmygit:
Hi @tatu-at-datastax, I've created https://github.com/stargate/stargate/pull/2359 to address the `v1` branch changes and have updated this PR to address other items.

testing/pom.xml
```diff
@@ -361,6 +361,10 @@
                   <goal>verify</goal>
                 </goals>
                 <configuration>
+                  <systemPropertyVariables>
+                    <ccm.version>3.11.14</ccm.version>
+                    <stargate.test.nodes>1</stargate.test.nodes>
+                  </systemPropertyVariables>
```
ivansenic, line 367: why is this added, we did not have this before?

persistence-cassandra-3.11/pom.xml
```diff
@@ -10,13 +10,13 @@
...
-    <!-- 3.11.13 depends on 3.0.1: https://mvnrepository.com/artifact/org.apache.cassandra/cassandra-all/3.11.13 -->
-    <cassandra.version>3.11.13</cassandra.version>
+    <!-- 3.11.14 depends on 3.11.3: https://mvnrepository.com/artifact/org.apache.cassandra/cassandra-all/3.11.14 -->
+    <cassandra.version>3.11.14</cassandra.version>
     <!--
       The driver used internally by cassandra-all for UDFs (must match the version declared in
       cassandra-all's POM).
     -->
-    <cassandra.bundled-driver.version>3.0.1</cassandra.bundled-driver.version>
+    <cassandra.bundled-driver.version>3.11.3</cassandra.bundled-driver.version>
```
ivansenic, line 19: This is not a correct change:

image

apis/pom.xml
```diff
@@ -32,7 +32,7 @@
       Please maintain default values for int-test in sync with io.stargate.sgv2.docsapi.testresource.StargateTestResource.Defaults
```
ivansenic, line 32: versions in the APIs are updated in the separate PR, targeting the `main` branch.. That's done after this PR is merged..

ivansenic (changes requested):
Great stuff, thanks for contribution..

* First of all, the Cassandra 4.x is already bumped to version `4.0.7`
* That's already on the `v1` branch that this PR should anyway target
* The builder image is also updated on the `v1` branch, thus you need to bump version again, to `v1.0.15`
   * Are you able to build and push the new image?

persistence-cassandra-3.11/pom.xml
ivansenic: same for the 4.x btw, but I guess we want no 4.x related changes in this PR anyway

apis/pom.xml
msmygit: Reverted the version here to `4.0.4`.

testing/pom.xml
msmygit: Both OSS C* 4.0.x and DSE 6.8.x had this and I added it here for OSS C* to make sure it is consistent with other versions thinking it might have been missing in the first place.

persistence-cassandra-3.11/pom.xml
msmygit: I've fixed these versions based on the above.

testing/pom.xml
msmygit: > The builder image is also updated on the v1 branch, thus you need to bump version again, to v1.0.15

GCR image version on `v1` branch prior to #2359 was at `1.0.14` and on `main` branch prior to this PR at `1.0.13`. Should we keep the images versions on both brances at the same version number? In that case, was that missed on the `main` branch earlier?

> Are you able to build and push the new image?

I didn't have access to GCB and hence Jeff was able to help with that in testing #2359 yesterday.

ivansenic:
OK let me try to explain this in detail. 

* The version updates are done in the `v1` branch
   * this has already done in #2359, with additional fixes in https://github.com/stargate/stargate/pull/2360
* Then we need to merge `main` and `v1` so that changes done in `v1` are also applied on the `main` branch   
* Once this is done, we only need to update the versions in the `/apis/*` (you correctly changed two needed files in this PR); however, we need no other changes!

I suggest you close this PR and open a new one for this purpose.. This new one should be similar to https://github.com/stargate/stargate/pull/2332

PR #2360, opened 2023-01-10
Get the version of cassandra bundled driver version correct

msmygit:
**What this PR does**: Get the version of cassandra bundled driver version correct

**Which issue(s) this PR fixes**: n/a
Fixes #<issue number>: n/a

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -53,8 +53,8 @@ RUN git clone --branch master --single-branch https://github.com/riptano/ccm.git
 
 # Create clusters to pre-download necessary artifacts
 RUN ccm create -v 4.0.7 stargate_40 && ccm remove stargate_40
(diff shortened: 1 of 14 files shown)


ivansenic (approved):
This looks good now.. What about that test spec in `pom.xml`, we should revert that as well imo

msmygit:
> This looks good now.. What about that test spec in `pom.xml`, we should revert that as well imo

We've discussed and decided to leave the `testing/pom.xml` changes to OSS C* 3.11 system test properties.

tatu-at-datastax:
Ok so this is partial revert of changes done earlier for DSE-6.8.31, right?
I wish I had more strongly asked about version update since I vaguely recalled there being diff versioning scheme for drivers vs Cassandra backend.
Thank you @ivansenic for catching this.

PR #2361, opened 2023-01-10
Update APIs to test against OSS C* 3.11.14, 4.0.7 and DSE 6.8.31

msmygit:
Updates apis to use OSS C* `3.11.14` & `4.0.7` and DSE `6.8.31`. Updates int test setup to use OSS C* `4.0.7`. Needs merging of `v1` and `main`. Followup work after:

- [x] PR #2370

**Checklist**
- [x] Rebase after merge occurs


tatu-at-datastax (approved):

jeffreyscarpenter:
What is the status of this PR?

tatu-at-datastax:
> What is the status of this PR?

Probably blocked pending DSE-6.8.31 investigation? However, I wonder if on short term we should try to undo DSE upgrade but keep C-3.11/4.0 updates? Would that be problematic with GCP builders tho?

jeffreyscarpenter:
> > What is the status of this PR?
> 
> Probably blocked pending DSE-6.8.31 investigation? However, I wonder if on short term we should try to undo DSE upgrade but keep C-3.11/4.0 updates? Would that be problematic with GCP builders tho?

We'd create a new version of the builder referencing the releases. The reason for creating a new builder version is to have the correct C*/DSE versions cached in the image to reduce CI run time.

tatu-at-datastax:
Ok I will create v1->main merge PR now.

tatu-at-datastax (approved):

tatu-at-datastax:
@jeffreyscarpenter Looks like CI fails due to inability to access `dse-core` `6.8.31` for Coordinator build. I am guessing this is because @msmygit does not have access somehow? Unless there is some other transient problem here?

ivansenic:
Closing as not needed anymore, changes already in the `main`.

PR #2363, opened 2023-01-10
closes #2209: implement retriable bridge calls

ivansenic:
**What this PR does**:
Adds our own configuration of the retriable bridge calls. By default, retries are enabled and done only once in case response status code in `UNAVAILABLE`.

From now, bridge must be received from `StargateRequestInfo`, except when you know what are you doing :)

**Which issue(s) this PR fixes**:
* Fixes #2209.
* Makes #2047 as won't fix.
* Makes #2353 as won't fix.

**Checklist**
- [x] Tests with retries
- [x] Configuration documentation

PR #2371, opened 2023-01-11
update quarkus to 2.15.3

ivansenic:
**What this PR does**:
Updates Quarkus to v2.15.3.. Hopefully we decrease chance of getting vulnerabilities and gets updating to next major version easier.

* I added small improvement for the retriable bridge test that I had in my mind after merging the #2363

### DIFF
--- a/apis/pom.xml
+++ b/apis/pom.xml
@@ -15,7 +15,7 @@
     <module>sgv2-restapi</module>
   </modules>
   <properties>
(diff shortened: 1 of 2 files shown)


tatu-at-datastax (approved):
Update looks good; there is 1 graphql IT failure but expecting it to be transient (retrying it).
Makes sense to upgrade. We'll want 2.16 when it lands, to get conditional inclusion for CQL endpoint.

Bit difficult to diff test changes but I assume they are fine.

PR #2373, opened 2023-01-12
Use Optimistic Queries for REST API

### LINKED ISSUE #2265: Use "Optimistic Query" functionality from common in SGv2/REST API
(from #2213 code review)

`RestResourceBase` method

```
  protected Uni<QueryOuterClass.Response> executeQueryAsync(
```

has a TODO note 

```
    // TODO implement optimistic queries (probably requires changes directly in SchemaManager)
```

which might be doable as:

```
return findTableAsync(keyspaceName, tableName, true)
    .onItem()
    .transformToUni(table -> executeQueryAsync(queryProducer.apply(table)))
    .onFailure(t -> {// predicate for matching the optimistic schema failure}).retry().atMost(2);
```

tatu-at-datastax:
**What this PR does**:

Uses newly added (#1977) Optimistic Query functionality to improve REST API performance.

Fixes #2265

**Checklist**
- [ ] Changes manually tested


apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java
```diff
@@ -44,6 +43,13 @@ public abstract class RestResourceBase {
   private static final Function<String, Uni<? extends Schema.CqlKeyspaceDescribe>>
       MISSING_KEYSPACE = ks -> Uni.createFrom().nullItem();
 
+  private static final Function<String, Uni<? extends QueryOuterClass.Response>>
+      MISSING_KEYSPACE_RESPONSE =
+          ks -> {
```
ivansenic, line 48: Instead of throwing you should do 

`return Mono.createFrom().failure(exception)`

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java
```diff
@@ -112,44 +112,29 @@ protected Uni<QueryOuterClass.Response> queryWithTableAsync(
...
-          return queryProducer.apply(table);
-        });
+
+    Function<Schema.CqlTable, Uni<QueryOuterClass.Query>> queryProducerUni =
+        table -> {
+          if (table == null) {
+            throw new WebApplicationException(
+                String.format("Table '%s' not found (in keyspace '%s')", tableName, keyspaceName),
+                Response.Status.BAD_REQUEST);
+          }
```
ivansenic, line 122: same here, return mono with failure, don't throw

ivansenic (commented):
Not sure what's going on in the tests.. Maybe I made a mistake, I am missing the test when the keyspace exists and it s not pre-cached, let me check that

ivansenic:
OK the issue is that the mappers for missing keyspace/table in`SchemaManager` are not made for explicit throwing.. And this should be fixed as mentioned in the review..

However, I realized that somebody might assume that this is possible and thus pushed 2 commits to avoid calling those functions until they are needed.. So that we don't hit this again in the future..

@tatu-at-datastax cc

ivansenic:
Tests are still failing, but I can not reproduce any of those using Swagger..

ivansenic:
@tatu-at-datastax OK the new issue is the following:

* you create keyspace and that's getting cached
* you create a new table, and this one is not in the keyspace anymore
* you fetched keyspace and table is reported as missing, without rechecking the keyspace

Adding a fix..

ivansenic:
@tatu-at-datastax OK down to 4 failures.. Can this be actually wrong tests?

tatu-at-datastax:
@ivansenic At least 3 of 4 fails are due to changing error message: these are actually improvements as now bad keyspace is directly reported, instead of claiming table doesn't exist. 4th one I need to look into more, not yet sure.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2IndexesResourceImpl.java
```diff
@@ -80,22 +80,34 @@ public Uni<RestResponse<Map<String, Object>>> addIndex(
...
-                throw new WebApplicationException(
-                    String.format("Index '%s' not found.", indexName), Status.NOT_FOUND);
+    final Query dropQuery =
+        new QueryBuilder().drop().index(keyspaceName, indexName).ifExists(ifExists).build();
+    if (ifExists) {
+      return executeQueryAsync(dropQuery).map(any -> RestResponse.noContent());
+    }
+
+    // Use direct query for checking existence
+    final Query existenceQuery =
```
tatu-at-datastax, line 90: Same query as for GET INDEX except all we care about is one specific index. So select "kind" field.

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QIntegrationTestBase.java
```diff
@@ -151,11 +152,11 @@ public void cleanupPerClass() {
    */
 
   public String testKeyspaceName() {
-    return testKeyspaceName;
+    return Objects.requireNonNull(testKeyspaceName, "'testKeyspaceName' not initialized");
```
tatu-at-datastax, line 155: Added because running of individual test from IDE for classes that only use per-class keyspaces will otherwise result in weird query execution failure ("No keyspace 'null' found").

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QSchemaColumnsIT.java
```diff
@@ -253,9 +253,7 @@ public void columnUpdateBadKeyspace() {
             HttpStatus.SC_BAD_REQUEST);
     ApiError apiError = readJsonAs(response, ApiError.class);
     assertThat(apiError.code()).isEqualTo(HttpStatus.SC_BAD_REQUEST);
-    // 02-Sep-2022, tatu: Error message is bit misleading but it is what it is; verify:
-    assertThat(apiError.description())
-        .matches("Table 'table' not found.*keyspace.*" + badKeyspace + ".*");
```
tatu-at-datastax, line 258: One bonus improvement is that we get better error checking for keyspace argument.

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QSchemaIndexesIT.java
```diff
@@ -177,10 +177,19 @@ public void indexDrop() {
     // And back to "no indexes"
     indexes = findAllIndexesFromSystemSchema(testKeyspaceName(), tableName);
     assertThat(indexes).isEmpty();
+  }
 
-    // And now an invalid case (could extract into separate test method in future
-    indexName = "no_such_index";
-    deletePath = endpointPathForIndexDelete(testKeyspaceName(), tableName, indexName);
+  @Test
+  public void indexDropNoSuchIndex() {
```
tatu-at-datastax, line 183: As per comment makes sense to separate out happy case from failure.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2IndexesResourceImpl.java
ivansenic: I am not understanding this. Why in case `ifExsist=false` we can not go through optimistic path?

I am also not understanding why we need to check the index first? and how is index not in the `CqlTable` schema.. Going for `system_schema` seems wrong imo.. Keyspace can not be prefixed with tenant and this will fail in some cases?

ivansenic (changes requested):
This index fix will not work in all envs.. We need to find alternative..

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2IndexesResourceImpl.java
tatu-at-datastax: So, why check? That's the way code was constructed when I started working: but I _think_ it was meant to give better user experience: perhaps deletion would simply succeed? I don't know the original reasoning.
Perhaps it'd be better simply to try to execute the query and see what happens.

Checking `system_schema` is used for other endpoints of Index access which is why I thought going that route would work (as it seems to).

But I can also change code not to do any validation and try to get failure from persistence backend.

The big question is indeed why `CqlTable` does not have/get updated information. Is it possible index information would not be part of hash calculation?

tatu-at-datastax: @ivansenic I rewrote handling again to use Optimistic Query path but removing explicit validation by caller -- persistence backend will do that anyway, reporting it as expected. This works and gives expected fails on missing index. I think this makes sense even if we figure out the issue with `CqlTable` not containing index information.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2IndexesResourceImpl.java
```diff
@@ -80,23 +80,15 @@ public Uni<RestResponse<Map<String, Object>>> addIndex(
...
-                throw new WebApplicationException(
-                    String.format("Index '%s' not found.", indexName), Status.NOT_FOUND);
-              }
-              return new QueryBuilder()
-                  .drop()
-                  .index(keyspaceName, indexName)
-                  .ifExists(ifExists)
-                  .build();
-            })
+            table ->
```
ivansenic, line 90: but this should still ensure table does exist before firing that query?

ivansenic: But table existence is not verified, see that null check for `table` that was on line `91`? Fix this and lets merge this?

tatu-at-datastax:
> I am not sure I am following anymore.. If there is an issue with the table hash wrt to the indexes, we should fix this asap imo.. Did you check the bridge? would this actually help here, I see in the new code there is no more check if the index exists as before?

Right, the thing about information not being updated on bridge is strange, but I think that regardless of that, checking for existence on client side seems unnecessary. But it is true this was done in the original code in SGv1 (when information was always directly accessible).

I can try to see why schema information for indexes seems not to be propagated.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2IndexesResourceImpl.java
tatu-at-datastax: > But table existence is not verified, see that null check for `table` that was on line `91`? Fix this and lets merge this?

Ok thanks I'll have a look there. I forgot that checking for `Table` not being null is left to handlers, whereas keyspace non-existence is

tatu-at-datastax: Adding a test to verify, but Table existence is actually checked in helper method in `RestResourceBase`, so that is covered. But adding a test for this makes sense, good suggestion.

ivansenic (approved):

PR #2378, opened 2023-01-14
use setup-python action in coordinator-test CI

jeffreyscarpenter:
The CI tests for the coordinator are running on an ubuntu 18.04 image that is no longer supported by GitHub Actions. GitHub has started enforcing periodic brownouts of the 18.04 image that cause CI runs to fail. We need to update

However there is a challenge in updating to newer ubuntu since we have been using Python 2 to run the ccm tool to start Cassandra nodes. Python 2 is not installed by default in newer ubuntu images, and our current approach to setting up ccm no longer works.

This PR upgrades the CI workflow to run on the latest ubuntu. The workflow installs Python 3 using the [GitHub recommended approach](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python) and then ccm (yes, you can pip install ccm on Python 3 and it works).


jeffreyscarpenter:
Interesting. Tests fail with repeated messages: 

```
java.lang.AssertionError: Tests using too many Stargate nodes (20 maximum)
```

.github/workflows/coordinator-test.yml
```diff
@@ -6,16 +6,25 @@ on:
...
       fail-fast: false
       matrix:
         backend: [ cassandra-40, cassandra-311, dse-68 ]
+        include:
+          - name: cassandra-40
+            ccm-version: -v 4.0.7
+          - name: cassandra-311
+            ccm-version: -v 3.11.14
+          - name: dse-68
+            ccm-version: -v 6.8.31 --dse
```
kathirsvn, line 22: dse version should be 6.8.32?

jeffreyscarpenter: I think there's a way to do this without specifying any specific versions in this file. Pushed another commit to give it a try.

.github/workflows/coordinator-test.yml
```diff
@@ -4,6 +4,53 @@ on:
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.x'
+      - uses: actions/setup-java@v3
+        name: Setup Java JDK
+        with:
+          distribution: 'temurin'
+          java-version: 8.0.332
```
tatu-at-datastax, line 21: We can use this, but latest seems to be 362. fwtw.

jeffreyscarpenter: I think I should move this to just be `8` so we always get latest.

ivansenic:
So is this running now fine?

tatu-at-datastax:
> So is this running now fine?

I am not sure since it's still only manually triggered (workflow_dispatch). So CI passing doesn't tell much.

.github/workflows/coordinator-test.yml
```diff
@@ -1,31 +1,120 @@
...
+        run: |
+          ./mvnw -B -ntp clean test -P dse
+
   integration-test:
-    name: Integration Test
-    # CCM installation trickier on 20.04, alas; Python 2.x dep etc problematic
-    runs-on: ubuntu-18.04
+    name: Integration tests
+    runs-on:
+      labels: stargate-coord-runner
```
ivansenic, line 69: speed improvements with 2 vs 8 core github runner

```
c 4
 * 2 cores: 54m47s
 * 8 cores: 50m43s

c 3.11
 * 2 cores: 51m31s  
 * 8 cores: 52m18s

dse
 * 2 cores: 1h24m
 * 8 cores: 1h10m 
```

I don't think there is any significant difference here

ivansenic (approved):
Merging this on green run.

PR #2379, opened 2023-01-16
v2 ci without snapshot caching

ivansenic:
**What this PR does**:
Trying to see if CI would work without snapshots cache.. we anyway use the maven cache, so I guess this could be enough..

### DIFF
--- a/.github/workflows/apis-v2.yaml
+++ b/.github/workflows/apis-v2.yaml
@@ -60,18 +60,6 @@ jobs:
           </settings>
           EOF
 
-      # ensure snapshots from this build are used in next jobs
-      - name: Cache Snapshots
-        uses: actions/cache@v3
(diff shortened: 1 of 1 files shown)


ivansenic:
@jeffreyscarpenter maybe this would help the cache..

tatu-at-datastax (approved):
I can't comment on what effect this will have but makes sense since we are running low on cache space.
But what about next step (Cache Libs)? I guess those are required for building persistence backends?

ivansenic:
@jeffreyscarpenter @tatu-at-datastax My solution is wrong. The build-coordinator is not caching anything.. Thus, I can not relay on the setup-java caching.

Let me try to avoid maven cache in setup java altogether.. But I doubt this would work anyway.. I would also need to delete any existing caching to actually correctly test this..

PR #2385, opened 2023-01-17
Fix #2384: revert DSE update (back to 6.8.29)

### LINKED ISSUE #2384: Revert partial DSE update for SGv1 to get back to 6.8.29 (undo of DSE changes from #2359)
Update of DSE persistence to use 6.8.31 was incomplete in #2359, and cannot be completed quite yet (there is investigation on version compatibility). In the meantime we need to make sure `v1` can be built with consistent set of components, so let's make sure all references are to 6.8.29.

tatu-at-datastax:
**What this PR does**:

As per #2384 revert DSE persistence dependency to 6.8.29.

Fixes #2384

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/ci/Makefile
+++ b/ci/Makefile
@@ -1,7 +1,7 @@
 HOSTNAME        := gcr.io
 PROJECT         := stargateio
 NAME            := stargate-builder
-VERSION         := v1.0.15
+VERSION         := v1.0.16
 TAG             := ${VERSION}
 IMG             := ${NAME}:${TAG}
 IMG_GCP         := ${HOSTNAME}/${PROJECT}/${IMG}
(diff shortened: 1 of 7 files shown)


jeffreyscarpenter (approved):

ci/Makefile
```diff
@@ -1,7 +1,7 @@
 HOSTNAME        := gcr.io
 PROJECT         := stargateio
 NAME            := stargate-builder
-VERSION         := v1.0.15
+VERSION         := v1.0.16
```
jeffreyscarpenter, line 4: Not sure we need to rev the builder, we can just use v1.0.14?

jeffreyscarpenter: but looks like it's already done so never mind

tatu-at-datastax: Well, I built new one just in case; wasn't 100% sure, since 3.11 version changed.

PR #2386, opened 2023-01-17
Docker-based IT for persistence/coordinator

jeffreyscarpenter:
Update IT framework including `ExternalStorage` to be Docker-based instead of using ccm. 

**Checklist**
- [ ] Changes manually tested


jeffreyscarpenter:
Initial push does not pass CIs - sharing for discussion purposes. 

The `ExternalStorage` class is updated to have a new implementation of the `Cluster` class alongside the `CcmCluster` there is now a `DockerCluster`. The challenge is getting the networking correct using TestContainers to allow the Cassandra node to know how to reach back from the Docker network to the host network in order to gossip. 

Rather than trying to figure out this networking, the better path is probably to update the `StargateExtension` to also be able to run Stargate nodes in Docker. This will require building local copies of Docker containers as a prerequisite.

persistence-test/src/main/java/io/stargate/it/storage/ExternalStorage.java
```diff
@@ -497,4 +499,141 @@ public void start() {
...
+      super(spec);
+      this.initSite = context.getUniqueId();
+
+      ImmutableList.Builder<GenericContainer> nodes = ImmutableList.builder();
+      for (int i = 0; i < spec.nodes(); i++) {
+        nodes.add(createCassandraContainer(i));
+      }
+      this.nodes = nodes.build();
+
+      Testcontainers.exposeHostPorts(7000);
```
ivansenic, line 531: maybe this is why it's not working, I think you also need `7001` for gossip?

persistence-dse-6.8/pom.xml
```diff
@@ -10,7 +10,7 @@
   <groupId>io.stargate.db.dse</groupId>
   <artifactId>persistence-dse-6.8</artifactId>
   <properties>
-    <dse.version>6.8.29</dse.version>
+    <dse.version>6.8.31</dse.version>
```
ivansenic, line 13: maybe we should not touch the version

persistence-test/src/main/java/io/stargate/it/storage/ExternalStorage.java
```diff
@@ -497,4 +499,141 @@ public void start() {
...
+      }
+    }
+
+    private GenericContainer<?> createCassandraContainer(int num) {
+      String image =
+          isDse()
+              ? DSE_REPOSITORY + ":" + DSE_VERSION
+              : CASSANDRA_REPOSITORY + ":" + CASSANDRA_VERSION;
+
+      GenericContainer<?> container =
```
ivansenic, line 610: * what about the seeds? like set seed address for the ones that have the `num > 0`..
* I think you should also define the alias for each node, like `cassandra-0`, `cassandra-1`, etc

persistence-test/src/main/java/io/stargate/it/storage/ExternalStorage.java
```diff
@@ -497,4 +499,141 @@ public void start() {
...
+          new GenericContainer<>(image)
+              .withEnv("HEAP_NEWSIZE", "512M")
+              .withEnv("MAX_HEAP_SIZE", "2048M")
+              .withEnv("CASSANDRA_CGROUP_MEMORY_LIMIT", "true")
+              .withEnv("CASSANDRA_BROADCAST_ADDRESS", broadcastAddress)
```
ivansenic, line 615: what's this doing?

ivansenic (commented):
The tests on the GCB will never work, we can not use testcontainers there.. What about a local execution or the GitHub workers?

jeffreyscarpenter:
After additional investigation, I don’t think the approach I am attempting here will work, to have Stargate nodes on a host machine work with Cassandra nodes in Docker, at least not without some changes to Stargate (and perhaps not even then). 

When you set the seed port on Stargate, it uses that value in two ways: 1) that’s the port uses for gossiping to hosts on the seed address list. 2) it listens on that same port for incoming gossip requests. The implications of this: when Cassandra is started in Docker by Testcontainers, port 7000 is exposed and mapped to a random port on the host machine, which is what we have to set as the seed port for Stargate. We'd need to expose this port to Docker using `Testcontainers.exposeHostPorts()`, but that is only with reference to 127.0.0.1, not the loopback addresses the StargateExtension typically uses.  

Based on this, I tried starting just a single Stargate on 127.0.0.1. The problem with this is that the Cassandra container does not know how to connect back to 127.0.0.1 - that means localhost for the container, not the host machine. I tried setting the `stargate.broadcast_address=host.testcontainers.internal` on the Stargate node so that this would be passed to the Cassandra node as the hostname to gossip back to, but the Stargate node fails on startup because it doesn't recognize that name.

I don't see a way to make this hybrid environment work. We could try running both Stargate and Cassandra in containers. This would mean that we'd need to first build the Stargate container. This would be possible, but I'm not sure it's worth it. 

I started exploring this configuration originally because we were having issues with using ccm in GitHub actions (and Google Cloud Build, which we've now discontinued using) based on Python 2. We were able to get ccm running on Python 3 for the [coordinator CI workflow](https://github.com/stargate/stargate/blob/main/.github/workflows/coordinator-test.yml) in a simple, reliable way., so I think the original issue is now moot. I'm comfortable continuing to use ccm for coordinator integration tests for now. Closing this issue.

PR #2387, opened 2023-01-17
Sync v1 changes to v2: C*3.11 and C*4.0 upgrades.

### LINKED ISSUE #2384: Revert partial DSE update for SGv1 to get back to 6.8.29 (undo of DSE changes from #2359)

tatu-at-datastax:
**What this PR does**:

Merges v1 changes (update C*3.11 and C*4) to v2 (main)

Fixes #2384 (part)

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,6 +1,6 @@
 steps:
   - id: 'pull image'
-    name: 'gcr.io/stargateio/stargate-builder:v1.0.15'
+    name: 'gcr.io/stargateio/stargate-builder:v1.0.16'
     entrypoint: 'bash'
     args: [ '-c', 'echo pulled builder' ]
   - id: 'fetch_secret'
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,6 +1,6 @@
 steps:
   - id: 'pull image'
-    name: 'gcr.io/stargateio/stargate-builder:v1.0.13'
+    name: 'gcr.io/stargateio/stargate-builder:v1.0.15'
     entrypoint: 'bash'
     args: [ '-c', 'echo pulled builder' ]
   - id: 'fetch_secret'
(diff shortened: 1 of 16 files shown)


cloudbuild.yaml
```diff
@@ -1,6 +1,6 @@
 steps:
   - id: 'pull image'
-    name: 'gcr.io/stargateio/stargate-builder:v1.0.13'
+    name: 'gcr.io/stargateio/stargate-builder:v1.0.15'
```
ivansenic, line 3: everywhere else is `v1.0.16`

pom.xml
```diff
@@ -16,7 +16,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.scm.id>github</project.scm.id>
     <doclint>none</doclint>
-    <driver.version>4.14.1</driver.version>
+    <driver.version>4.15.0</driver.version>
```
ivansenic, line 19: where s this coming from?

ivansenic (commented):
Is this a merge commit? What's the idea?

tatu-at-datastax:
> Is this a merge commit? What's the idea?

Yes, merge commit.

pom.xml
tatu-at-datastax: From v1. Updates other than DSE retained -- this included. Should not make a difference (I tried with both versions and DSE 6.8.30/.31 had same issue).

cloudbuild.yaml
tatu-at-datastax: Ah. thanks will fix.

jeffreyscarpenter (approved):

PR #2393, opened 2023-01-20
Add minimal support for DynamicCompositeType in persistence backends

### LINKED ISSUE #2144: Prevent Schema update failure on Stargate due to `CompositeType` / `DynamicCompositeType` inclusion in table definition
CompositeType are not supported in stargate schema.

Create table as below:

```
CREATE TABLE IF NOT EXISTS jtest.composite(
     "name" text,
     "compTest" 'org.apache.cassandra.db.marshal.DynamicCompositeType(B=>org.apache.cassandra.db.marshal.ReversedType(org.apache.cassandra.db.marshal.BytesType),D=>org.apache.cassandra.db.marshal.ReversedType(org.apache.cassandra.db.marshal.DoubleTyp)',
     "shard" int,
     PRIMARY KEY(name))
```

Error happened while updating the schema

```
java.lang.IllegalArgumentException: Unknown type mapping for class org.apache.cassandra.db.marshal.DynamicCompositeType
        at com.datastax.oss.driver.shaded.guava.common.base.Preconditions.checkArgument(Preconditions.java:216)
        at io.stargate.db.dse.impl.Conversion.getTypeFromInternal(Conversion.java:667)
        at io.stargate.db.dse.impl.SchemaConverter.columnType(SchemaConverter.java:87)
        at io.stargate.db.dse.impl.SchemaConverter.columnType(SchemaConverter.java:26)
        at io.stargate.db.datastore.common.AbstractCassandraSchemaConverter.convertColumn(AbstractCassandraSchemaConverter.java:176)
        at io.stargate.db.datastore.common.AbstractCassandraSchemaConverter.lambda$convertColumns$1(AbstractCassandraSchemaConverter.java:168)
```

tatu-at-datastax:
**What this PR does**:

Adds minimal support for `DynamicCompositeType` such that creation of Tables with this type via CQL does NOT cause hard failure in schema handler. No support added for actual usage.

Fixes #2144

**Checklist**
- [x] Changes manually tested


tatu-at-datastax:
Uhhh. Looks like Java cassandra driver does not really recognize DynamicCompositeType: it appears to have no support whatsoever. Same appears to be true for native protocol, used by Stargate persistence.
This presents a problem in mapping type.
So far I have been able to avoid failure from schema sync aspect; and CRUD to/from table with DCT works, but only as long as DCT column is not referenced.

persistence-api/src/main/java/io/stargate/db/schema/Column.java
```diff
@@ -259,6 +260,18 @@ public enum Type implements ColumnType {
     Decimal(6, BigDecimal.class, false, "Variable-precision decimal, supports integers and floats"),
     Double(7, Double.class, false, "64-bit IEEE-754 floating point"),
     Duration(21, CqlDuration.class, false, "128 bit encoded duration with nanosecond precision"),
+    /**
+     * Need a placeholder for DynamicCompositeType: while not usable via APIs can potentially be
+     * created via CQL leading to problems (see {@href
+     * https://github.com/stargate/stargate/issues/2144}).
+     */
+    DynamicComposite(48, LinkedHashMap.class, false, "Legacy data type") {
```
tatu-at-datastax, line 268: `LinkedHashMap` only used to be unique (not used by other types). Same true for `id`; `48` undefined.
Obviously not correct choices at this point.

tatu-at-datastax: Needed to change to "real" type, can just create a placeholder similar to, say, `Point` (see below).

tatu-at-datastax:
I think this will do; prevents failure system failure from schema-sync as Dynamic Composite Type is now recognized.
If actual usage was to be supported that requires more future work, including investigation on what exactly would be needed.

JeremiahDJordan (approved):
Looks good to me.  Matches what was done so that Point/LineString/Polygon don't crash the schema.

JeremiahDJordan:
Should we also open a new ticket to add full support?  I wonder if we should also add CompositeType here as well.

tatu-at-datastax:
> Should we also open a new ticket to add full support? I wonder if we should also add CompositeType here as well.

Yes, I will create a follow-up ticket for possible full support (it is being requested)

I could extend this to "plain" (?) `CompositeType` too, if I can find how to test it. Let me see.

tatu-at-datastax:
@JeremiahDJordan Added handling for `CompositeType` as well, along with tests; same behavior as far as I can see.

kathirsvn (approved):

tatu-at-datastax:
Tests green, will merge.

JeremiahDJordan:
> 

👍

PR #2404, opened 2023-01-25
set location of jemalloc library

jeffreyscarpenter:
As described in https://github.com/stargate/docker-images/issues/128, Stargate nodes output a warning message on startup indicating that the jemalloc library cannot be located. Enabling access to this library enables more optimal memory allocation for better performance. The PR https://github.com/stargate/docker-images/pull/129 ensures the library is installed in Stargate Docker images, while this PR adds logic borrowed from the `cassandra` startup script to correctly set the location of the library on both Linux / Mac platforms. Similar changes will be required for Stargate v2

**Checklist**
- [x] Changes manually tested


starctl
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e
+set -x
```
tatu-at-datastax, line 2: So instead of failing eagerly, print out each command to execute? Is removal of former intentional?

tatu-at-datastax (approved):

starctl
jeffreyscarpenter: this was for debugging, reverting the change

ivansenic:
@jeffreyscarpenter so running `./starctl` on my machine with these changes does not work at all.. Command just exists..

PR #2412, opened 2023-01-31
adds user agent grpc tag provider

ivansenic:
**What this PR does**:
Implements grpc tag provider that ads a grpc user-agent tag.

* [ ] ~base on v1~

### DIFF
--- a/grpc/src/main/java/io/stargate/grpc/metrics/api/GrpcMetricsTagProvider.java
+++ b/grpc/src/main/java/io/stargate/grpc/metrics/api/GrpcMetricsTagProvider.java
@@ -17,6 +17,7 @@
 
 package io.stargate.grpc.metrics.api;
 
+import io.grpc.Metadata;
 import io.grpc.ServerCall;
 import io.micrometer.core.instrument.Tags;
 
@@ -29,7 +30,8 @@ public interface GrpcMetricsTagProvider {
    * Prometheus does not allow different tags from the single process.
    *
    * @param call Server-side gRPC call
+   * @param requestHeaders Call metadata
    * @return Tags
    */
-  Tags getCallTags(ServerCall<?, ?> call);
+  Tags getCallTags(ServerCall<?, ?> call, Metadata requestHeaders);
 }
--- a/grpc/src/main/java/io/stargate/grpc/metrics/api/UserAgentTagProvider.java
+++ b/grpc/src/main/java/io/stargate/grpc/metrics/api/UserAgentTagProvider.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.stargate.grpc.metrics.api;
+
+import io.grpc.Metadata;
+import io.grpc.ServerCall;
+import io.grpc.internal.GrpcUtil;
+import io.micrometer.core.instrument.Tags;
+import io.stargate.core.metrics.StargateMetricConstants;
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+/**
+ * Extension of the {@link GrpcMetricsTagProvider} that can extract user agent information from the
+ * request metadata.
+ *
+ * <p>To avoid metric explosion, we are processing the user agent value in the following way:
+ *
+ * <ol>
+ *   <li><code>grpc-java-netty/1.51.0</code> results in <code>grpc-java</code>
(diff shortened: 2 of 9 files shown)


tatu-at-datastax:
> Should this be targeting `v1` rather?

IMO If it is easy enough that would make sense. If lots of work, no (but I assume code is about same so should not be tons)?

grpc/src/main/java/io/stargate/grpc/metrics/api/UserAgentTagProvider.java
```diff
@@ -0,0 +1,92 @@
...
+    String[] agentParts = userAgent.split(" ");
+    String tagValue =
+        Arrays.stream(agentParts)
+
+            // remove version from agent name
+            // input ex. grpc-java-netty/1.51.0, my-agent/1.0.0
+            .map(
+                agentAndVersion -> {
+                  int versionIndex = agentAndVersion.indexOf('/');
+                  if (versionIndex > 0) {
```
tatu-at-datastax, line 67: Is this guaranteed to always have version as second part as per spec? Or should it try to validate that (f.ex) first character is number (no need to do thorough validation, just needs to look like version)

grpc/src/main/java/io/stargate/grpc/metrics/api/GrpcMetricsTagProvider.java
```diff
@@ -29,7 +30,8 @@ public interface GrpcMetricsTagProvider {
    * Prometheus does not allow different tags from the single process.
    *
    * @param call Server-side gRPC call
+   * @param requestHeaders Call metadata
    * @return Tags
    */
-  Tags getCallTags(ServerCall<?, ?> call);
+  Tags getCallTags(ServerCall<?, ?> call, Metadata requestHeaders);
```
tatu-at-datastax, line 36: I guess/hope there are no external developers implementing this interface; if there were, this would be backwards-incompatible change (could probably make it work by deprecating old method, having new method have default implementation that called old method).

This may also mean that this should only be done in 2.x? (there may be Stargate 1.x developers that do this)

tatu-at-datastax (approved):
Had couple of questions but overall looks good.

tatu-at-datastax:
> > Should this be targeting `v1` rather?
> 
> IMO If it is easy enough that would make sense. If lots of work, no (but I assume code is about same so should not be tons)?

Looking at code, maybe we should not backport this: if anyone implements tag provider, interface has changed?
(I am not sure if that is actually possible thing to do, maybe I am over-analysing this?)

jeffreyscarpenter:
> > > Should this be targeting `v1` rather?
> > 
> > 
> > IMO If it is easy enough that would make sense. If lots of work, no (but I assume code is about same so should not be tons)?
> 
> Looking at code, maybe we should not backport this: if anyone implements tag provider, interface has changed? (I am not sure if that is actually possible thing to do, maybe I am over-analysing this?)

I agree with not backporting to `v1`

jeffreyscarpenter:
I am ok with the changes here, but I wonder if we should actually consider a different approach. The HTTP APIs have `INFO` logging provided by the framework that includes Apache-style log messages, for example:

```
172.25... - - [01/Feb/2023:20:30:29 +0000] "GET /v2/keyspaces/... HTTP/1.0" 200 29 "-" "Go-http-client/2.0" 22
```

Is there a similar log message provided by `grpc-java` that we could enable, or could we implement one ourselves? This would preserve the original user-agent string and keep the cardinality of metric tagging low.

tatu-at-datastax:
> I am ok with the changes here, but I wonder if we should actually consider a different approach. The HTTP APIs have `INFO` logging provided by the framework that includes Apache-style log messages, for example:
> 
> ```
> 172.25... - - [01/Feb/2023:20:30:29 +0000] "GET /v2/keyspaces/... HTTP/1.0" 200 29 "-" "Go-http-client/2.0" 22
> ```

I thought access-log was not enabled for Quarkus/SGv2? But if it is enabled it'd definitely make sense to include cleaned up user-agent there.

grpc/src/main/java/io/stargate/grpc/metrics/api/GrpcMetricsTagProvider.java
ivansenic: Wait what are we talking about here.. This is interface? It's fully backward compatible, when someone updates the SG version he will need to change the signature of the existing interfaces.. That has nothing to do with backward compatibility.. Every time we update Quarkus there is some change that needs to be applied, that's normal.. 

It would not be backward compatible if this change would cause the existing SG installations not to be usable anymore, or break in some point.. This is not the case here.. Let's not mix things up..

ivansenic:
@tatu-at-datastax @jeffreyscarpenter waiting until we figure out if this should be done using the metrics or not..

grpc/src/main/java/io/stargate/grpc/metrics/api/GrpcMetricsTagProvider.java
tatu-at-datastax: In practice not a big deal, but conceptually this would be breaking change -- for minor version, Quarkus upgrade actually should never require a code change. That's what semantic version says. It happens given Quarkus is still sort of young project, but compilation breaking changes really aren't supposed to go in minor (or esp. patch versions).

The way to handle this -- if it mattered, and I agree that it doesn't :) -- would be to leave old version as deprecated, redirect to new (or vice versa).

ivansenic:
@jeffreyscarpenter can you provide a review here.. I aligned the impl with the http.. The user agent tagging is not active by default..

jeffreyscarpenter (approved):

PR #2415, opened 2023-02-02
Externalizing Cassandra and Coordinator timeout values for test container

maheshrajamani:
Default value will remain the same as it is now.

**Checklist**
- [ x] Changes manually tested

### DIFF
--- a/apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/common/testresource/StargateTestResource.java
+++ b/apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/common/testresource/StargateTestResource.java
@@ -81,6 +81,10 @@ public class StargateTestResource
     String CLUSTER_VERSION = "4.0";
 
     String CLUSTER_DSE = null;
+
+    long CASSANDRA_STARTUP_TIMEOUT = 2;
+    long COORDINATOR_STARTUP_TIMEOUT = 3;
+
   }
 
   private static final Logger LOG = LoggerFactory.getLogger(StargateTestResource.class);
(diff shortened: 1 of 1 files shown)


apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/common/testresource/StargateTestResource.java
```diff
@@ -81,6 +81,9 @@ interface Defaults {
     String CLUSTER_VERSION = "4.0";
 
     String CLUSTER_DSE = null;
+
+    long CASSANDRA_STARTUP_TIMEOUT = 2;
```
tatu-at-datastax, line 85: These are in minutes? Would make sense to add a comment mentioning this?

tatu-at-datastax (approved):
Makes sense, but I think this should be documented somewhere (if we have a README or something)? Maybe `sgv2-quarkus-common` needs `TESTING.md`.

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/common/testresource/StargateTestResource.java
tatu-at-datastax: Actually, thinking about this some more: should we require it in minutes, or would it make more sense to use seconds value? To allow more granular configurability (90 seconds etc)

maheshrajamani: This is timeout value for starting up the service, so not sure if we need seconds level granularity. Added comments to for the defaults.

maheshrajamani:
> Makes sense, but I think this should be documented somewhere (if we have a README or something)? Maybe `sgv2-quarkus-common` needs `TESTING.md`.

Yes we need these System property documented somewhere. Will create a issue for this and complete it.

ivansenic:
@maheshrajamani +1 for the documentation.. Just don't forget to document the props from the `IntegrationTestUtils` those are far more important than these container ones.. Anyway a nice `TESTING.md` would be really helpful..

PR #2416, opened 2023-02-03
fix(sec): upgrade org.eclipse.jetty:jetty-http to 11.0.10

neostfox:
### What happened？
There are 1 security vulnerabilities found in org.eclipse.jetty:jetty-http 9.4.49.v20220914
- [CVE-2022-2047](https://www.oscs1024.com/hd/CVE-2022-2047)

### What did I do？
Upgrade org.eclipse.jetty:jetty-http from 9.4.49.v20220914 to 11.0.10 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@
       Note that this is currently aligned with the Jetty version that DropWizard transitively depends on.
       It's probably a good idea to keep the two versions in sync.
     -->
-    <jetty.version>9.4.49.v20220914</jetty.version>
+    <jetty.version>11.0.10</jetty.version>
     <!-- Logging framework likewise in sync -->
     <slf4j.version>1.7.36</slf4j.version>
     <logback.version>1.2.11</logback.version>


tatu-at-datastax:
First of all @neostfox, thank you for informing us about CVE & proposing a patch.

Unfortunately, since Jetty is dependency is brought via DropWizard it is unlikely to be possible to blindly just upgrade Jetty by two major versions. DW depends on specific version (or at least specific major version), so we need to see what latest DW is compatible with.
Upgrade also needs to go via `v1` branch, but that's a smaller issue and I can handle that.

tatu-at-datastax:
Created #2417 as replacement, leaving this open just to see what CI would have thought (likely to fail)

tatu-at-datastax:
Fails already as Jetty 11.x seems to require Java 11: coordinator still with Java 8 mostly due to OSGi (new APIs require Java 17).

Closing, following up with #2417.

tatu-at-datastax:
I am also bit puzzled by this on CVE:

```In Eclipse Jetty versions 9.4.0 thru 9.4.46, ...```

As we were already on 9.4.49, this CVE doesn't even appear applicable?

PR #2418, opened 2023-02-03
Add helper method for logging in Starter.java (updated) 

### LINKED ISSUE #1303: Add helper method for logging in Starter.java

jeffreyscarpenter:
Update to previous attempt to resolve #1303 that addresses the review feedback on #1348 

Fixes #1303

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/stargate-starter/src/main/java/io/stargate/starter/Starter.java
+++ b/stargate-starter/src/main/java/io/stargate/starter/Starter.java
@@ -417,7 +417,7 @@ public class Starter {
         setStargateProperties();
       }
     } catch (Exception e) {
-      logerr(e.getMessage(), Level.WARNING);
+      logerr(e.getMessage(), Level.SEVERE);
       System.exit(2);
     }
 
@@ -425,7 +425,7 @@ public class Starter {
     framework = new Felix(felixConfig());
     framework.init();
 
-    logerr("JAR DIR: " + JAR_DIRECTORY, Level.WARNING);
+    log("JAR DIR: " + JAR_DIRECTORY, Level.INFO);
 
     // Install bundles
     context = framework.getBundleContext();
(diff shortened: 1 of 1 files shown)


stargate-starter/src/main/java/io/stargate/starter/Starter.java
```diff
@@ -307,6 +308,14 @@ public Starter(
     this.disableMBeanRegistration = true;
   }
 
+  private static void log(String message, Level level) {
+    System.out.println("[" + level.getName() + "] " + message);
+  }
+
+  private static void logerr(String message, Level level) {
+    System.err.println("[" + level.getName() + "] " + message);
```
kathirsvn, line 316: Should we consider passing the throwable exception 'e' here and print the stack trace like below, instead of just the message? 
`e.printStackTrace(System.err);`
Relying just on the message (and when that being some generic message) might sometimes make it hard to find which part of the code caused the exception and the stack trace will give a quick idea with the line number.

tatu-at-datastax: Usually I worry about verbosity, but here I agree that stack trace would make lots of sense. These should be rare and real failures.

jeffreyscarpenter: sure, and these are things that prevent startup

tatu-at-datastax (commented):
LGTM but I think @kathirsvn's suggestion wrt Exception is good to consider. Letting him approve.

kathirsvn (approved):
LGTM, exception trace logging is a nice to have and not a blocker definitely.

PR #2422, opened 2023-02-06
trying to fix flaky deadline test

ivansenic:
**What this PR does**:
Trying to fix flaky deadline test.. Couldn't reproduce it.. I improved a bit a shutdown of the executor service in the `BridgeTest`, increased the timeout, moved this to common.. Not sure really why it fails on the action..


tatu-at-datastax:
CI-only failures are really nasty :-/

tatu-at-datastax (approved):
Well, can't hurt and hopefully helps. Looks good to me.

PR #2424, opened 2023-02-06
sign images with additional tags during the release

ivansenic:
**What this PR does**:
During the release we sign images push to the ECR, but only the ones for the version that's being released. The images pushed with additional tags (`v2`) are not signed. This PR fixes that.

### DIFF
--- a/.github/workflows/release-v2.yml
+++ b/.github/workflows/release-v2.yml
@@ -456,10 +456,13 @@ jobs:
         id: login-ecr
         uses: aws-actions/amazon-ecr-login@v1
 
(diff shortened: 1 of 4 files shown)


kathirsvn (approved):
LGTM

ivansenic:
@kathirsvn Ohh man I can not believe that we both did not catch this bug: https://github.com/stargate/stargate/pull/2437

kathirsvn:
> @kathirsvn Ohh man I can not believe that we both did not catch this bug: #2437

yeah ... untestable as well, since it's in the release workflow.

PR #2425, opened 2023-02-06
Fix workflow typo to address #2423

### LINKED ISSUE #2423: Release workflow contains wrong vars
Just noticed this issue:

https://github.com/stargate/stargate/blob/13ab4e1ddfb3a6b306cfde8e3d5bdcd42a8ee225/.github/workflows/release-v2.yml#L62

Seems that after the update, the correct variable names were not handled correctly. Each plugin has it's own vars, and they are clearly here misused.

Note this would fail the release workflow until fixed.

jeffreyscarpenter:
Fixes #2423

### DIFF
--- a/.github/workflows/release-v2.yml
+++ b/.github/workflows/release-v2.yml
@@ -59,7 +59,7 @@ jobs:
         id: create_release
         uses: softprops/action-gh-release@v1
         with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ secrets.GITHUB_TOKEN }}
           tag_name: ${{needs.resolve-tag.outputs.release-tag}}
           name: Release ${{needs.resolve-tag.outputs.release-tag}}
           draft: false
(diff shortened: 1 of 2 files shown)


.github/workflows/release-v2.yml
```diff
@@ -59,7 +59,7 @@ jobs:
         id: create_release
         uses: softprops/action-gh-release@v1
         with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ secrets.GITHUB_TOKEN }}
```
tatu-at-datastax, line 62: Lovely. So there's very little standardization/common practices in naming of configuration properties across actions :-(

jeffreyscarpenter: I tried to test using that `act` tool using the dryrun mode but it didn't pick up on the error.

ivansenic:
Good now we can release

PR #2427, opened 2023-02-07
Fix postman-docker workflow

jeffreyscarpenter:
Second attempt to fix issues with workflow that runs postman collections against docker-compose configurations

**Checklist**
- [x] Changes manually tested


docker-compose/cassandra-3.11/docker-compose-dev-mode.yml
```diff
@@ -8,7 +8,6 @@ services:
     ports:
       - "9042:9042"
       - "8081:8081"
-      - "8084:8084"
```
tatu-at-datastax, line 11: So this is the... metrics endpoint in Coordinator? And one we don't really need to expose.

tatu-at-datastax (approved):

docker-compose/cassandra-3.11/docker-compose-dev-mode.yml
jeffreyscarpenter: something else is using this port when running in the GitHub Actions environment, no idea what. So yeah I just removed it and moved on.

PR #2439, opened 2023-02-10
always remove docker images in CI for APIs

jeffreyscarpenter:
Reduce usage of resources within GitHub account by always deleting docker images built during runs of the APIs CI workflow.

Rationale: we are storing docker images of the coordinator according to the commit sha that triggered the CI. While there is caching in this workflow of coordinator Jars from the maven build, the docker images are generated and stored as GH artifacts every time this workflow is executed. If there is a test failure, we are currently not deleting these images, which ties up storage for the retention period (currently set for 3 days). 

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/.github/workflows/apis-v2.yaml
+++ b/.github/workflows/apis-v2.yaml
@@ -305,11 +305,11 @@ jobs:
           cd apis/
           ./mvnw -B -ntp verify -DskipUnitTests -pl ${{ matrix.project }} -am -P ${{ matrix.profile }} -Dquarkus.container-image.build=$CONTAINER_TEST -Dtesting.containers.stargate-image=${{ matrix.image }}
 
-  # runs only on success, deletes built docker image artifacts
-  # keep docker images for 3 days so we can re-run the workflows
+  # always delete built docker image artifacts
   clean-up:
     name: Clean up
     runs-on: ubuntu-latest
+    if: always()
     needs: int-tests
 
     steps:


tatu-at-datastax (approved):

jeffreyscarpenter:
One of the integration tests in the matrix failed with an odd error. `Stargate APIs V2 / Integration tests (sgv2-graphqlapi, dse-68)`

```Run actions/download-artifact@v3
Starting download for img-coordinator-dse-6[8](https://github.com/stargate/stargate/actions/runs/4148525986/jobs/7177653816#step:6:9)-88dc7a[9](https://github.com/stargate/stargate/actions/runs/4148525986/jobs/7177653816#step:6:10)1b38fd865b6913a99935c2a723c200dc5
Error: Unable to find any artifacts for the associated workflow
```

This seems odd because the other two tests in the matrix that use the DSE coordinator image succeeded. Could there be a timing error where GitHub Actions is considering `int_tests` completed prematurely?

ivansenic (changes requested):
I would not do this. There is no chance to retry int tests if we always delete the docker images. The tests depends on docker images, and thus to retry you would need to run everything again. Makes no sense imo.. Can we simply just decrease the rentention?

tatu-at-datastax:
> I would not do this. There is no chance to retry int tests if we always delete the docker images. The tests depends on docker images, and thus to retry you would need to run everything again. Makes no sense imo.. Can we simply just decrease the rentention?

Ah! This is the part I forgot -- so not just debugging failures but ability to retry individual matrix test run without having to re-trigger all ITs.

I agree then, reducing retention to something lower would make sense.

jeffreyscarpenter:
Closing since we do not have consensus on approach. I will create a separate PR to reduce retention time.

PR #2442, opened 2023-02-13
Fix #2438: rename back and forth of (id) column failing

### LINKED ISSUE #2438: Error in REST API on schema change
Automated CI tests "Stargate Postman Collections" that were recently created for Postman collections are failing for the REST collection. This collection has a section where there is a request to rename a column followed immediately by a request to revert back to the original name.

This is the pertinent output from the Postman run:

```
↳ Change the column firstname to first
  PUT localhost:8082/v2/schemas/keyspaces/users_keyspace/tables/users/columns/firstname  
  200 OK ★ 153ms time ★ 474B↑ 101B↓ size ★ 10↑ 2↓ headers ★ 0 cookies
  ┌ ↑ raw ★ 60B
  │ {
  │       "name": "first",
  │       "typeDefinition": "varchar"
  │ }
  └ 
  ┌ ↓ application/json ★ text ★ json ★ utf8 ★ 16B
  │ {"name":"first"}
  └
  prepare   wait    dns-lookup   tcp-handshake   transfer-start   download   process   total 
  1ms       349µs   (cache)      (cache)         150ms            1ms        94µs      154ms 

  ✓  Status code is 200
  ✓  Verify column name

↳ Change the column first back to firstname
  PUT localhost:8082/v2/schemas/keyspaces/users_keyspace/tables/users/columns/first  
  400 Bad Request ★ 19ms time ★ 474B↑ 203B↓ size ★ 10↑ 2↓ headers ★ 0 cookies
  ┌ ↑ raw ★ 64B
  │ {
  │       "name": "firstname",
  │       "typeDefinition": "varchar"
  │ }
  └ 
  ┌ ↓ application/json ★ text ★ json ★ utf8 ★ 108B
  │ {"description":"Column 'first' not found in table 'use
  │ rs'","code":400,"grpcStatus":null,"internalTxId":null}
  └
  prepare   wait    dns-lookup   tcp-handshake   transfer-start   download   process   total 
  1ms       367µs   (cache)      (cache)         16ms             1ms        59µs      20ms  

  1. Status code is 200
  2. Verify column name
 ```

Run logs available under
https://github.com/stargate/stargate/actions/workflows/postman-docker.yaml

This may be a race condition, as this is not observed when the test is run on a local machine. However, even when a 1s delay between requests was added (`--delay-request 1000` option on `postman` command), the error still occurs - see run # 28 initiated from the `jeff/add-postman-delay` branch.

### PR DESCRIPTION
**Pull request by tatu-at-datastax**
**What this PR does**:

Will make sequence of rename-to-new-name-rename-back not fail, by using blocking access instead of optimistic query (`queryWithSchema()`).

And after doing that will do the same for all REST/Schema operations to prevent similar issues with table, index CRUD, not just columns.

Fixes #2438

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java
+++ b/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java
@@ -41,10 +41,16 @@ public abstract class RestResourceBase {
   protected static final int DEFAULT_PAGE_SIZE = 100;
 
   private static final Function<String, Uni<? extends Schema.CqlKeyspaceDescribe>>
-      MISSING_KEYSPACE = ks -> Uni.createFrom().nullItem();
+      MISSING_KEYSPACE =
+          ks ->
+              Uni.createFrom()
+                  .failure(
+                      new WebApplicationException(
+                          String.format("Keyspace '%s' not found", ks),
+                          Response.Status.BAD_REQUEST));
 
   private static final Function<String, Uni<? extends QueryOuterClass.Response>>
-      MISSING_KEYSPACE_RESPONSE =
+      MISSING_KEYSPACE_AS_RESPONSE =
           ks ->
               Uni.createFrom()
                   .failure(
@@ -130,10 +136,10 @@ public abstract class RestResourceBase {
 
     if (checkAuthzForTableMetadata) {
       return schemaManager.queryWithSchemaAuthorized(
-          keyspaceName, tableName, MISSING_KEYSPACE_RESPONSE, queryProducerUni);
+          keyspaceName, tableName, MISSING_KEYSPACE_AS_RESPONSE, queryProducerUni);
     }
     return schemaManager.queryWithSchema(
-        keyspaceName, tableName, MISSING_KEYSPACE_RESPONSE, queryProducerUni);
+        keyspaceName, tableName, MISSING_KEYSPACE_AS_RESPONSE, queryProducerUni);
   }
 
   protected Uni<QueryOuterClass.Response> executeQueryAsync(QueryOuterClass.Query query) {
--- a/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2ColumnsResourceImpl.java
+++ b/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2ColumnsResourceImpl.java
@@ -32,26 +32,23 @@ public class Sgv2ColumnsResourceImpl extends RestResourceBase implements Sgv2Col
   public Uni<RestResponse<Sgv2NameResponse>> createColumn(
       String keyspaceName, String tableName, Sgv2ColumnDefinition columnDefinition) {
     final String columnName = columnDefinition.name();
-    return queryWithTableAsync(
-            keyspaceName,
-            tableName,
-            true,
-            (tableDef) -> {
-              Column.Kind kind =
-                  columnDefinition.isStatic() ? Column.Kind.STATIC : Column.Kind.REGULAR;
-              Column columnDef =
-                  ImmutableColumn.builder()
-                      .name(columnName)
-                      .kind(kind)
-                      .type(columnDefinition.typeDefinition())
-                      .build();
-              return new QueryBuilder()
-                  .alter()
-                  .table(keyspaceName, tableName)
-                  .addColumn(columnDef)
-                  .parameters(PARAMETERS_FOR_LOCAL_QUORUM)
-                  .build();
-            })
+    final Column columnDef =
+        ImmutableColumn.builder()
+            .name(columnName)
+            .kind(columnDefinition.isStatic() ? Column.Kind.STATIC : Column.Kind.REGULAR)
+            .type(columnDefinition.typeDefinition())
+            .build();
+    return getTableAsyncCheckExistence(keyspaceName, tableName, true, Response.Status.BAD_REQUEST)
+        // We only want to validate keyspace, existence of table, before attempting insertion so:
+        .flatMap(
+            tableDef ->
+                executeQueryAsync(
+                    new QueryBuilder()
+                        .alter()
+                        .table(keyspaceName, tableName)
+                        .addColumn(columnDef)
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by tatu-at-datastax** [comment]
@ivansenic I think I may be doing something wrong with update here so that perhaps update query does not actually get executed at all. I was wondering if you can see something obviously wrong. This because there are 2 new test failures with change and while there are many things that could cause that, one is that maybe rename query itself is not being executed.

File: apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2ColumnsResourceImpl.java [hunk 1]
```diff
@@ -81,25 +81,28 @@ public Uni<RestResponse<Sgv2NameResponse>> updateColumn(
...
-              return new QueryBuilder()
-                  .alter()
-                  .table(keyspaceName, tableName)
-                  .renameColumn(columnName, newName)
-                  .parameters(PARAMETERS_FOR_LOCAL_QUORUM)
-                  .build();
+              return column;
             })
-        .map(any -> restResponseOkWithName(newName));
+        .map(
```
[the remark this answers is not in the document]
tatu-at-datastax: Thank you @ivansenic. I guess I was relying too much on types matching, but `Uni<?>` can be nested for sure.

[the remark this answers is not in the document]
tatu-at-datastax: @ivansenic cool, this resolved 2 failures. One remaining one is due to different exception for bad keyspace (says "no such table" for invalid/missing keyspace), and the new test is still failing but now at a later point.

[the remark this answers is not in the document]
tatu-at-datastax: Was able to resolve this by changing "missing keyspace" case to produce failure, not `null` table.
This actually improves some of existing misleading failure messages too.

tatu-at-datastax:
Need to see how CI does: fix to handling of bad/non-existent keyspace may change reported failure in ITs (improvement but existing test verify existing message). Will also verify that the column endpoint is not using optimistic query for other operations.

tatu-at-datastax:
Converted all 3 REST/Schema endpoints that were using optimistic query, to use one that gets schema instead.
Now only data (row CRUD) uses optimistic query.

jeffreyscarpenter (approved):

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java
```diff
@@ -41,10 +41,16 @@ public abstract class RestResourceBase {
...
+          ks ->
+              Uni.createFrom()
+                  .failure(
+                      new WebApplicationException(
+                          String.format("Keyspace '%s' not found", ks),
+                          Response.Status.BAD_REQUEST));
 
   private static final Function<String, Uni<? extends QueryOuterClass.Response>>
-      MISSING_KEYSPACE_RESPONSE =
+      MISSING_KEYSPACE_AS_RESPONSE =
```
ivansenic, line 53: where is this used?

ivansenic: aha this is typing problem, uf so bad..

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2TablesResourceImpl.java
```diff
@@ -123,31 +123,29 @@ public Uni<RestResponse<Sgv2NameResponse>> createTable(
...
-                  .table(keyspaceName, tableName)
-                  .withDefaultTTL(options.defaultTimeToLive())
-                  .parameters(PARAMETERS_FOR_LOCAL_QUORUM)
-                  .build();
-            })
+    Sgv2Table.TableOptions options = tableUpdate.tableOptions();
+    List<?> clusteringExpressions = options.clusteringExpression();
+    if (clusteringExpressions != null && !clusteringExpressions.isEmpty()) {
+      throw new WebApplicationException(
+          "Cannot update the clustering order of a table", Status.BAD_REQUEST);
```
ivansenic, line 130:
```
return Uni.createFrom().failure(new WebApplicationException(
          "Cannot update the clustering order of a table", Status.BAD_REQUEST));
```

Always prefer this that throwing exceptions in the JVM

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2TablesResourceImpl.java
```diff
@@ -123,31 +123,29 @@ public Uni<RestResponse<Sgv2NameResponse>> createTable(
...
+    }
+    final Integer defaultTTL = options.defaultTimeToLive();
+    // 09-Dec-2021, tatu: Seems bit odd but this is the way SGv1/RESTv2 checks it,
+    //    probably since this is the only thing that can actually be changed:
+    if (defaultTTL == null) {
+      throw new WebApplicationException("No update provided for defaultTTL", Status.BAD_REQUEST);
+    }
```
ivansenic, line 137: same here

ivansenic (approved):

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/RestResourceBase.java
tatu-at-datastax: yes, exactly. Was trying to think of whether to just cast it (since type not really needed) but decided to keep separate for now.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/schemas/Sgv2TablesResourceImpl.java
tatu-at-datastax: Ok I can change that.

PR #2457, opened 2023-02-23
automated release support: changelog and event dispatching 

ivansenic:
**What this PR does**:

Adds first step for release automation:
* changelog generation as part of the bump version PR
    * continues on error, just in case we endup in some API limits or so
* dispatches events about releases to numerous repositories
    * includes release train and version   
    * does not fail fast so matrix executes all of the items
    * still missing `C2_PAT` variable

Based on `v1`, will propagate to main with a merge.

Internal.

### DIFF
--- a/.github/workflows/release-v2.yml
+++ b/.github/workflows/release-v2.yml
@@ -541,6 +541,7 @@ jobs:
     name: Dispatch release event
     needs: [ 'resolve-tag', 'build', 'publish-ossrh', 'publish-docker']
     if: ${{ always() }}
+    continue-on-error: true
     runs-on: ubuntu-latest
 
     strategy:
@@ -561,4 +562,4 @@ jobs:
           token: ${{ matrix.token }}
           repository: ${{ matrix.repo }}
           event-type: stargate-v2-release
-          client-payload: '{"release-train": "v2", "version": "${{ needs.resolve-tag.outputs.release-tag}}"}'
+          client-payload: '{"version": "${{ needs.resolve-tag.outputs.release-tag}}"}'
--- a/.github/workflows/release-v2.yml
+++ b/.github/workflows/release-v2.yml
@@ -550,7 +550,7 @@ jobs:
         repo: [ 'stargate/jsonapi', 'riptano/c2']
         include:
           - repo: stargate/jsonapi
-            token: ${{ secrets.JSONAPI_PAT }}
+            token: ${{ secrets.SG_EVENTS_PAT }}
 
           - repo: riptano/c2
             token: ${{ secrets.C2_PAT}}
--- a/.github/workflows/release-v2.yml
+++ b/.github/workflows/release-v2.yml
@@ -507,6 +507,12 @@ jobs:
           </settings>
           EOF
 
+      - name: Generate changelog
+        continue-on-error: true
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: ./update_changelog.sh
+
       - name: Update version numbers (coordinator)
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release-v2.yml
```diff
@@ -529,3 +535,30 @@ jobs:
           branch-suffix: "short-commit-hash"
           base: "main"
           labels: "stargate-v2"
+
+  # dispatch release event
+  dispatch:
+    name: Dispatch release event
+    needs: [ 'resolve-tag', 'build', 'publish-ossrh', 'publish-docker']
+    if: ${{ always() }}
```
ivansenic, line 543: I think there is value in running even if some parts of the workflow fail?

.github/workflows/release-v2.yml
```diff
@@ -529,3 +535,30 @@ jobs:
...
+            token: ${{ secrets.C2_PAT}}
+
+    steps:
+      - name: Repository dispatch
+        uses: peter-evans/repository-dispatch@v2
+        with:
+          token: ${{ matrix.token }}
+          repository: ${{ matrix.repo }}
+          event-type: stargate-v2-release
+          client-payload: '{"release-train": "v2", "version": "${{ needs.resolve-tag.outputs.release-tag}}"}'
```
ivansenic, line 564: no need for release train, events are different..

.github/workflows/release.yml
```diff
@@ -176,3 +182,30 @@ jobs:
...
+  dispatch:
+    name: Dispatch release event
+    needs: [ 'create-release', 'build', 'publish']
+    if: ${{ always() }}
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        repo: [ 'stargate/stargate', 'riptano/c2']
```
jeffreyscarpenter, line 196: Let's add stargate/docker-images

.github/workflows/release-v2.yml
jeffreyscarpenter: I see. Because event type is `stargate-v2-release` that conveys enough info? So you are going to push a change to remove `release-train` from the payload?

jeffreyscarpenter (changes requested):

.github/workflows/release-v2.yml
ivansenic: Yup.

.github/workflows/release-v2.yml
```diff
@@ -529,3 +535,30 @@ jobs:
           branch-suffix: "short-commit-hash"
           base: "main"
           labels: "stargate-v2"
+
+  # dispatch release event
+  dispatch:
+    name: Dispatch release event
```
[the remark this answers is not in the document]
ivansenic: Done.

.github/workflows/release.yml
ivansenic: Done.

jeffreyscarpenter (approved):
Looks good

tatu-at-datastax:
Sweet! Automatic generation of changelog should simplify workflow a lot (it takes a while). And the idea is then that this allows other repositories to run workflows as a result of Stargate release? Like run integration tests, maybe bump their dependency version, or even (in future) maybe create release of their own? (or at least PR for update)?
That sounds very useful.

ivansenic:
> Sweet! Automatic generation of changelog should simplify workflow a lot (it takes a while). And the idea is then that this allows other repositories to run workflows as a result of Stargate release? Like run integration tests, maybe bump their dependency version, or even (in future) maybe create release of their own? (or at least PR for update)? That sounds very useful.

Yup.. Already implemented everything :)

PR #2458, opened 2023-02-23
adds workflow for automatic merge v1 into main

ivansenic:
**What this PR does**:

Adds workflow that merges v1 into main and creates a PR. Runs on `stargate-v1-release` event or manual trigger.

Internal

### DIFF
--- a/.github/workflows/merge-main-v1.yaml
+++ b/.github/workflows/merge-main-v1.yaml
@@ -1,0 +1,44 @@
+# @author Ivan Senic
+name: Merge main/v1
+
+# runs on
+# * stargate v1 release event
+# * manual trigger
+on:
+  repository_dispatch:
+    types: [ stargate-v1-release ]
+  workflow_dispatch:
+
+jobs:
+
+  merge:
+    name: Merge branches
+    runs-on: ubuntu-latest
+
+    steps:
+
+      # always checkout main, but fetch complete history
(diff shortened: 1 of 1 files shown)


.github/workflows/merge-main-v1.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Run git merge
+        run: git merge --no-commit -s ours origin/v1
+
+      # commit and create pr
+      - name: Create pull-request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: Merging branch v1 into main
+          branch: create-pull-request/merge-main-v1
```
jeffreyscarpenter, line 36: Do we need to make this branch name different each time? In most cases the branch gets deleted once the PR is merged, but what if we don't merge the PR quickly enough for some reason?

ivansenic: it is changed, it's gonna be `create-pull-request/merge-main-v1-aaabbb`, as `branch-suffix: short-commit-hash` defines.

jeffreyscarpenter: interesting, thanks.

jeffreyscarpenter (approved):

PR #2459, opened 2023-02-23
fix merge main/v1 workflow

ivansenic:
**What this PR does**:

* Needs commit
* Workflow fails on the merge with following:
  ```
  Committer identity unknown
  
  *** Please tell me who you are.
  
  Run
  
    git config --global user.email "you@example.com"
    git config --global user.name "Your Name"
  ```

Solution from https://github.com/actions/checkout#push-a-commit-using-the-built-in-token.

### DIFF
--- a/.github/workflows/merge-main-v1.yaml
+++ b/.github/workflows/merge-main-v1.yaml
@@ -25,7 +25,10 @@ jobs:
 
       # run merge without a commit (will be added in final step)
       - name: Run git merge
-        run: git merge --no-commit -s ours origin/v1
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git merge --no-commit -s ours origin/v1
 
       # commit and create pr
       - name: Create pull-request
--- a/.github/workflows/merge-main-v1.yaml
+++ b/.github/workflows/merge-main-v1.yaml
@@ -28,7 +28,7 @@ jobs:
         run: |
           git config user.name github-actions
           git config user.email github-actions@github.com
-          git merge --no-commit -s ours origin/v1
+          git merge -s ours origin/v1
 
       # commit and create pr
       - name: Create pull-request


ivansenic:
Works now:
 
* success action: https://github.com/stargate/stargate/actions/runs/4254689855
* created PR: https://github.com/stargate/stargate/pull/2460

ivansenic:
@jeffreyscarpenter can you approve?

.github/workflows/merge-main-v1.yaml
```diff
@@ -25,7 +25,10 @@ jobs:
 
       # run merge without a commit (will be added in final step)
       - name: Run git merge
-        run: git merge --no-commit -s ours origin/v1
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git merge -s ours origin/v1
```
ivansenic, line 31: should we deifne better commit message then the default: `Merge remote-tracking branch 'origin/v1'`

tatu-at-datastax: qq: is this needed because of bogus version conflict resolution? (1.x vs 2.x version difference in pom.xml)?

tatu-at-datastax: and yes, we could probably use better message, indicating something like 

"v1->main merge from GH action by [user]"

if possible?

ivansenic: by [user]? who is the user?

tatu-at-datastax: User who triggered the workflow. If possible to determine.

ivansenic: Impossible. Plus the main purpose of this is to react on the event of the SG v1 release and create merge automatically.. So this can not be done.

Any other proposal?

tatu-at-datastax: Ok yeah if not doable then can't do.

Mostly I'd just like to know these PRs came from workflow action -- so whatever static description is fine.
Even the original one is ok, it has enough finrmation.

tatu-at-datastax (approved):

PR #2481, opened 2023-03-07
speed up coord integration tests by ordering tests

ivansenic:
**What this PR does**:
Tries to speed up tests, by ordering them in a way that tests:

* that need custom ccm cluster runs first
* that need custom stargate runs last

This will ensure minimum amount of restarts of both Stargate and CCM nodes.


tatu-at-datastax (approved):
Sounds good, hopefully improves speed as expected. Would be good to have a note explaining use of ordering somewhere but looks like there is no `testing/README.md` to add it to.

ivansenic:
Seems like it's 5 mins saved app.. We decreased the amount of Stargate restarts from 30 to 14.. These 14 are needed, as this is how many tests need custom Stargate context.. Still, better than nothing..

tatu-at-datastax:
> Seems like it's 5 mins saved app.. We decreased the amount of Stargate restarts from 30 to 14.. These 14 are needed, as this is how many tests need custom Stargate context.. Still, better than nothing..

Agreed, it's a Win.

PR #2491, opened 2023-03-13
move files for dependency checking to coordinator directory

jeffreyscarpenter:
A couple of files related to dependency/vulnerability checks were not moved to the `coordinator/` directory so the weekly workflow is failing.

https://github.com/stargate/stargate/actions/workflows/dependency-check.yml

PR #2478, opened 2023-03-07
closes #2455: coordinator folder refactoring final

ivansenic:
**What this PR does**:

I tried different things, but for now I would like to close the #2455 with the more or less same structure as now. So no shared pom, the projects can be opened independently or added to a single workspace in IDE.

What's done:

* Updated maven wrappers 
* Tried to align poms, make it more clear, readable and developer friendly (not sure if I managed)
* Minimal adaptation of the documents

I noticed some issues, will try to comment on some, but there are others that we should create tickets for:

* source code for quarkus commons is not published
* source code plugin in coordinator has outdated version
* 

**Which issue(s) this PR fixes**:
Fixes #2455

PR #2463, opened 2023-02-28
relates to #2455: create dedicated coordinator folder

ivansenic:
**:warning:  WARNING REVIEWERS: This is sensitive change, please invest considerable amount of time to review all changes carefully. In addition, think what could have been forgotten and should be also adapted.**

### What this PR does:

* Moves all coordinator modules under `coordinator/` folder
* Adapts workflows and CI to the new structure

### Follow-up tasks:

To make review process easier and to make git recognize all the moved files I left some tasks for a second follow-up PR:

* create joined root
* use single maven wrapper
* update documentation

### Which issue(s) this PR fixes:
Relates to #2455.

PR #2492, opened 2023-03-14
Move coordinator to separate folder on `v1`

ivansenic:
**What this PR does**:

Applies following PRs on the `v1` branch as well, so we have also coordinator files under dedicated folder:

* https://github.com/stargate/stargate/pull/2463
* https://github.com/stargate/stargate/pull/2478
* https://github.com/stargate/stargate/pull/2491
* applies some manual changes


ivansenic:
@jeffreyscarpenter @tatu-at-datastax This was pretty easy to do. However, doing a merge test now showed a lot of conflicts. However, I think that can be resolved once and then we would be fine. Need to check that though.

tatu-at-datastax (approved):

ivansenic:
Hmm, I think I am going to redo this one. I don't feel too confident here because I used the commits from `main` and applied them here.. Would feel more if I would do it still step by step directly on v1, confirming that there are no hidden changes..

tatu-at-datastax:
> Hmm, I think I am going to redo this one. I don't feel too confident here because I used the commits from `main` and applied them on `v1`.. Would feel more if I would do it still step by step directly on v1, confirming that there are no hidden changes..

Ok, I think that's a good idea, just in case. +1.

PR #2517, opened 2023-03-29
allow specification of custom grpc retry policy

ivansenic:
**What this PR does**:

Allows project that depend on the `sgv2-quarkus-common` to provide their own `custom` grpc retry policy.

Internal issue.

### DIFF
--- a/apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/GrpcConfig.java
+++ b/apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/GrpcConfig.java
@@ -59,7 +59,8 @@ public interface GrpcConfig {
      *
      * If unset, noop policy will be used (never retries).
      *
-     * @return The type of the {@link io.stargate.sgv2.api.common.grpc.retries.GrpcRetryPredicate} used.
+     * @return The type of the {@link io.stargate.sgv2.api.common.grpc.retries.GrpcRetryPredicate}
+     *     used.
      */
     @WithDefault("status-codes")
     Optional<@Pattern(regexp = "status-codes|custom") String> policy();
--- a/apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/GrpcConfig.java
+++ b/apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/GrpcConfig.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Optional;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
 import javax.validation.constraints.Positive;
 
 /** Configuration for the gRPC calls to the Bridge. */
@@ -48,8 +49,25 @@ public interface GrpcConfig {
     boolean enabled();
 
     /**
+     * What type of retry policy to use.
+     *
+     * <ol>
+     *   <li><code>status-codes</code> - based on status codes (see {@link
+     *       io.stargate.sgv2.api.common.grpc.retries.impl.StatusCodesRetryPredicate}}
+     *   <li><code>custom</code> - allows configuring custom policy
+     * </ol>
+     *
+     * If unset, noop policy will be used (never retries).
+     *
+     * @return The type of the {@link io.stargate.sgv2.api.common.grpc.retries.GrpcRetryPredicate} used.
+     */
+    @WithDefault("status-codes")
+    Optional<@Pattern(regexp = "status-codes|custom") String> policy();
+
+    /**
      * @return List of status codes to execute retries for. Defaults to <code>UNAVAILABLE</code>, as
-     *     this code means that the request never reached the bridge and it should be safe to retry.
+     *     this code means that the request never reached the bridge or C* responded with
+     *     Unavailable exception, thus it should be safe to retry.
      */
     @WithDefault("UNAVAILABLE")
     @NotNull
--- a/apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/GrpcConfig.java
+++ b/apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/GrpcConfig.java
@@ -63,7 +63,9 @@ public interface GrpcConfig {
      *     used.
      */
     @WithDefault("status-codes")
-    Optional<@Pattern(regexp = "status-codes|custom") String> policy();
+    @NotNull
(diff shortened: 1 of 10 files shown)


apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/GrpcConfig.java
```diff
@@ -47,9 +48,27 @@ interface Retries {
...
+     *   <li><code>custom</code> - allows configuring custom policy
+     * </ol>
+     *
+     * If unset, noop policy will be used (never retries).
+     *
+     * @return The type of the {@link io.stargate.sgv2.api.common.grpc.retries.GrpcRetryPredicate}
+     *     used.
+     */
+    @WithDefault("status-codes")
+    Optional<@Pattern(regexp = "status-codes|custom") String> policy();
```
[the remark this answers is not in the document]
ivansenic: Sure we can remove the optional :+1:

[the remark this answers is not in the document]
tatu-at-datastax: Looking at `GrpcRetriesConfiguration` there is value `noop` so I think that needs to be allowed?

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/retries/configuration/GrpcRetriesConfiguration.java
```diff
@@ -0,0 +1,31 @@
...
+  GrpcRetryPredicate statusCodes(GrpcConfig config) {
+    List<Status.Code> statusCodes = config.retries().statusCodes();
+    return new StatusCodesRetryPredicate(statusCodes);
+  }
+
+  @Produces
+  @ApplicationScoped
+  @LookupIfProperty(
+      name = "stargate.grpc.retries.policy",
+      stringValue = "noop",
```
tatu-at-datastax, line 26: As per my comment in `GrpcCongif`, I think this value needs to be allowed in regexp.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/retries/impl/StatusCodesRetryPredicate.java
```diff
@@ -0,0 +1,25 @@
...
+
+import io.grpc.Status;
+import io.grpc.StatusRuntimeException;
+import io.stargate.sgv2.api.common.grpc.retries.GrpcRetryPredicate;
+import java.util.List;
+
+/** {@link GrpcRetryPredicate} based on status code matching. */
+public class StatusCodesRetryPredicate implements GrpcRetryPredicate {
+
+  private final List<Status.Code> statusCodes;
```
tatu-at-datastax, line 11: Should be a `Set` since we do lookup for `contains`.

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/api/common/grpc/RetriableStargateBridgeTest.java
```diff
@@ -37,191 +37,141 @@
...
-          .isInstanceOfSatisfying(
-              StatusRuntimeException.class,
-              e -> assertThat(e.getStatus()).isEqualTo(Status.UNIMPLEMENTED));
-
-      // verify one call only
-      verify(bridgeService).executeQuery(eq(request), any());
+    @Override
+    public Map<String, String> getConfigOverrides() {
+      return ImmutableMap.<String, String>builder()
+          .put("stargate.grpc.retries.status-codes", "UNAVAILABLE,NOT_FOUND")
```
tatu-at-datastax, line 51: I guess this also tests that default policy is "status-codes". Could alternatively specify if explicitly here as documentation. Or maybe just add a comment it's the default (in case reader is wondering about it)

tatu-at-datastax (changes requested):
I think choice "noop" needs to be included in the validation regexp.
One of tests could add it as a choice (f.ex one where retries are disabled).

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/retries/configuration/GrpcRetriesConfiguration.java
ivansenic: see above, not needed anymore

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/grpc/retries/impl/StatusCodesRetryPredicate.java
ivansenic: Yup makes sense :+1:

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/api/common/grpc/RetriableStargateBridgeTest.java
ivansenic: adding a comment

ivansenic:
@tatu-at-datastax fixed everything if you want to have another look

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/config/GrpcConfig.java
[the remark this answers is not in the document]
tatu-at-datastax: Ah ok. So the idea is to use enabled/disabled instead of specific "enabled but does nothing" option. Makes sense.

tatu-at-datastax:
> @tatu-at-datastax fixed everything if you want to have another look

Looks like it was missing 2 small changes (adding a comment, changing `List` to `Set`)?
(but had the changes to policy changes).

ivansenic:
> > @tatu-at-datastax fixed everything if you want to have another look
> 
> Looks like it was missing 2 small changes (adding a comment, changing `List` to `Set`)? (but includes changes to policy configuration).

Yup somehow missed it, fixed in: https://github.com/stargate/stargate/pull/2521

tatu-at-datastax:
> > > @tatu-at-datastax fixed everything if you want to have another look
> > 
> > 
> > Looks like it was missing 2 small changes (adding a comment, changing `List` to `Set`)? (but includes changes to policy configuration).
> 
> Yup somehow missed it, fixed in: #2521

Ok np, thanks for doing PR.

PR #2530, opened 2023-04-06
Create dependabot.yml

jeffreyscarpenter:
I've been looking at our current approach using a homegrown workflow running maven commands to track dependency updates and vulnerabilities. After reading the dependabot docs in more detail I think it has the potential to be a better solution, with support for checking more than just Java dependencies, and better automation to help us keep current more quickly. I'd like to give it a shot on this repo and then see if we want to also enable on other repos.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+updates:
+- package-ecosystem: maven
+  # should pick up both apis and coordinators poms
+  directory: "/"
+  schedule:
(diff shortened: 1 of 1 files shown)


tatu-at-datastax (approved):

ivansenic (approved):
Cool.. Can we get rid of that maven commands you mentioned? So we don't have both approaches in place?

PR #2537, opened 2023-04-13
Fix #2536: problem with Schema Hash calculation for List/Map/Set columns

### LINKED ISSUE #2536: SchemaHashable calculation unstable for List/Map/Set CQL Column types
Looks like calculation of hashes used for Schema hashing is unstable for structured Column types: Map, List and Set. This results in different Coordinator nodes calculating different hash for Keyspace instances, resulting in occasional nasty failure when API node thinks different coordinators have different schema versions and tries in quick succession to update to a stable state.

tatu-at-datastax:
**What this PR does**:

Fixes SchemaHash calculation for List/Map/Set columns.

Fixes #2536

**Checklist**
- [ ] Changes manually tested


coordinator/persistence-api/src/main/java/io/stargate/db/schema/ParameterizedType.java
```diff
@@ -157,7 +157,7 @@ public Column.ColumnType fieldType(String name) {
     @Value.Auxiliary
     public int schemaHashCode() {
       return SchemaHashable.combine(
-          SchemaHashable.hashCode(isFrozen()), SchemaHashable.hashCode(parameters()));
+          SchemaHashable.hashCode(isFrozen()), SchemaHashable.hash(parameters()));
```
tatu-at-datastax, line 160: This is important! Former lead to `list.hashCode()` calculation; latter calculates over `List<SchemaHashable>`.
One reason why it probably has not a great idea to provide general `SchemaHashable.hashCode(Object)` method in the first place....

coordinator/persistence-api/src/main/java/io/stargate/db/schema/SchemaHashable.java
```diff
@@ -49,11 +54,19 @@ static int hashCode(boolean b) {
 
   static int enumHashCode(Enum<?> object) {
     if (object == null) {
-      return 0;
+      return -1;
     }
     return object.name().hashCode();
   }
 
+  static int hashCode(Map<String, String> map) {
```
tatu-at-datastax, line 62: Adding explicit methods to reduce need to call generic `hashCode(Object)` which is dangerous and should be removed eventually.

coordinator/persistence-api/src/main/java/io/stargate/db/schema/SchemaHashable.java
```diff
@@ -24,17 +25,21 @@ public interface SchemaHashable {
 
   static int hashCode(SchemaHashable object) {
     if (object == null) {
-      return 0;
+      return -1;
```
tatu-at-datastax, line 28: Changed just because value `0` has often special meaning as "no hash calculated". Doesn't greatly matter.

coordinator/persistence-api/src/main/java/io/stargate/db/schema/SchemaHashable.java
```diff
@@ -24,17 +25,21 @@ public interface SchemaHashable {
...
   }
 
+  // 12-Apr-2023, tatu: This is still dangerous; should get rid of it. But reduce usage for now
   static int hashCode(Object object) {
     // It's still possible that something like `Optional<Enum>` would make it through. So that's
     // something to look out for.
-    if (object instanceof Enum || object instanceof SchemaHashable) {
+    if (object instanceof Enum
+        || object instanceof SchemaHashable
+        || object instanceof Collection) {
```
tatu-at-datastax, line 39: Preventing calls with Collections (Lists, Sets); would have caught the bug we had.

coordinator/persistence-api/src/main/java/io/stargate/db/schema/SchemaHashable.java
```diff
@@ -24,17 +25,21 @@ public interface SchemaHashable {
 
   static int hashCode(SchemaHashable object) {
     if (object == null) {
-      return 0;
+      return -1;
     }
     return object.schemaHashCode();
   }
 
+  // 12-Apr-2023, tatu: This is still dangerous; should get rid of it. But reduce usage for now
```
jeffreyscarpenter, line 33: Why not eliminate the method now?

coordinator/persistence-api/src/main/java/io/stargate/db/schema/SchemaHashable.java
```diff
@@ -24,17 +25,21 @@ public interface SchemaHashable {
...
   static int hashCode(Object object) {
     // It's still possible that something like `Optional<Enum>` would make it through. So that's
     // something to look out for.
-    if (object instanceof Enum || object instanceof SchemaHashable) {
+    if (object instanceof Enum
+        || object instanceof SchemaHashable
+        || object instanceof Collection
+        || object instanceof Map) {
```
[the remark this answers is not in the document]
tatu-at-datastax: Perhaps surprisingly it is not; different interface (Map is not a Collection<Map.Entry<...>>)

jeffreyscarpenter (approved):

ivansenic:
@tatu-at-datastax @jeffreyscarpenter this also only on `main`? are we abandoning v1?

tatu-at-datastax:
@ivansenic Bridge and optimistic queries were only added in v2. So while we are moving away from v1, this is not due to that.

PR #2548, opened 2023-04-18
Fix #2547, upgrade Jetty test dep (only relevant wrt SGv1 endpoints)

tatu-at-datastax:
**What this PR does**:

Updates Jetty-core dependency (used by DropWizard, used by SGv1 endpoints)

**Checklist**
- [x] Changes manually tested
- [] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

### DIFF
--- a/coordinator/pom.xml
+++ b/coordinator/pom.xml
@@ -55,12 +55,12 @@
     <doclint>none</doclint>
     <driver.version>4.15.0</driver.version>
     <!-- First DropWizard version and things it directly depends on that make sense to sync -->
-    <dropwizard.version>2.1.6</dropwizard.version>
+    <dropwizard.version>2.0.35</dropwizard.version>
     <!--
       Note that this is currently aligned with the metrics version that DropWizard transitively depends on.
       It's probably a good idea to keep the two versions in sync.
     -->
-    <dropwizard.metrics.version>4.2.18</dropwizard.metrics.version>
+    <dropwizard.metrics.version>4.1.36</dropwizard.metrics.version>
     <!--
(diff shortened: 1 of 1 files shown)


coordinator/pom.xml
```diff
@@ -63,9 +63,9 @@
     <dropwizard.metrics.version>4.1.36</dropwizard.metrics.version>
     <!--
       Note that this is currently aligned with the Jetty version that DropWizard transitively depends on.
-      It's probably a good idea to keep the two versions in sync.
+      It's probably a good idea to keep the two versions in sync (or close)
     -->
-    <jetty.version>9.4.50.v20221201</jetty.version>
+    <jetty.version>9.4.51.v20230217</jetty.version>
```
ivansenic, line 68: I would actually also like to see Dropwizard being bumped to `2.1.16`, which actually uses this version.. That being said, we could then invest time in bumping the dependencies across the project.. For example, logback could be bumped one version, maybe something else as well..

tatu-at-datastax: I can try to see if that works fine, sure.

tatu-at-datastax: @ivansenic Unfortunately it's not plug-in replacement without changes. I could spend more time figuring it out, but we are trying to get rid of DropWizard so for this change there's not much benefit.

ivansenic (approved):

coordinator/pom.xml
tatu-at-datastax: Ok yes I will file an issue.

PR #2553, opened 2023-04-24
Fix #2551: upgrade DropWizard (2.0.35 -> 2.1.6)

### LINKED ISSUE #2551: Update DropWizard from 2.0.x to 2.1.x
(note: follow-up to #2547 / #2548)

We are currently depending on Dropwizard 2.0.35 (last 2.0.x version) for Coordinator (and until removed, SGv1 endpoints).
There are newer branches: 2.1.x being actively maintained, stable; and 3.x / 4.x for likely bigger changes.
See https://mvnrepository.com/artifact/io.dropwizard/dropwizard-core for more details
We should upgrade to 2.1.x to make sure we get security updates; if easy enough we may consider bigger upgrade too.
It is unclear if 2.0.x branch will get any updates.

Initial testing showed that 2.1.x upgrade is not lift-and-shift (as CI failed); it may be we just need to upgrade other dependencies, or perhaps need small (?) changes to code for deprecated methods.

tatu-at-datastax:
**What this PR does**:

Upgrades DropWizard from 2.0.x to 2.1.x to remain on supported version.

Fixes #2551

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/coordinator/core/src/main/java/io/stargate/core/metrics/impl/PrefixingMetricRegistry.java
+++ b/coordinator/core/src/main/java/io/stargate/core/metrics/impl/PrefixingMetricRegistry.java
@@ -108,7 +108,7 @@ public class PrefixingMetricRegistry extends MetricRegistry {
 
   @Override
   @SuppressWarnings("rawtypes")
-  public Gauge gauge(String name, MetricSupplier<Gauge> supplier) {
+  public <T extends Gauge> T gauge(String name, MetricSupplier<T> supplier) {
     return backingRegistry.gauge(addPrefix(name), supplier);
   }
 
--- a/coordinator/pom.xml
+++ b/coordinator/pom.xml
@@ -55,20 +55,20 @@
     <doclint>none</doclint>
     <driver.version>4.15.0</driver.version>
     <!-- First DropWizard version and things it directly depends on that make sense to sync -->
-    <dropwizard.version>2.0.35</dropwizard.version>
+    <dropwizard.version>2.1.6</dropwizard.version>
     <!--
       Note that this is currently aligned with the metrics version that DropWizard transitively depends on.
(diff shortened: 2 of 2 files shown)


coordinator/core/src/main/java/io/stargate/core/metrics/impl/PrefixingMetricRegistry.java
```diff
@@ -108,7 +108,7 @@ public Timer timer(String name, MetricSupplier<Timer> supplier) {
 
   @Override
   @SuppressWarnings("rawtypes")
-  public Gauge gauge(String name, MetricSupplier<Gauge> supplier) {
+  public <T extends Gauge> T gauge(String name, MetricSupplier<T> supplier) {
```
tatu-at-datastax, line 111: Signature change in "dropwizard-metrics" package: binary-compatible, source-incompatible

coordinator/pom.xml
```diff
@@ -613,14 +613,11 @@
         <artifactId>metrics-graphite</artifactId>
         <version>${dropwizard.metrics.version}</version>
       </dependency>
-      <!-- 30-Mar-2022, tatu: temporarily force newer jersey-common
-	   (over what DW asks) to get vuln fixes; make sure we have enforcer rules to alert if this
-	   becomes less than latest
-	  -->
+      <!-- 24-Apr-2023, tatu: Sync with DropWizard 2.0.6 again -->
```
ivansenic, line 616: you mean `2.1.6`?

ivansenic (approved):
Awesome @tatu-at-datastax

coordinator/pom.xml
tatu-at-datastax: Yes, good catch.

PR #2564, opened 2023-04-28
Cql metrics driver tag

maheshrajamani:
**What this PR does**:
Cql client metrics by driver name and version.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/coordinator/cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
+++ b/coordinator/cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
@@ -290,7 +290,12 @@ public final class ClientMetrics {
               .map(clientInfoTagProvider::getClientInfoTags)
               .orElse(Tags.empty());
 
-      requestsProcessed = meterRegistry.counter(REQUESTS_PROCESSED_METRIC, tags);
+      Tags tagsByDriver =
+          Optional.ofNullable(clientInfo)
+              .map(clientInfoTagProvider::getClientInfoTagsByDriver)
+              .orElse(Tags.empty());
+
+      requestsProcessed = meterRegistry.counter(REQUESTS_PROCESSED_METRIC, tagsByDriver);
       requestsDiscarded = meterRegistry.counter(REQUESTS_DISCARDED_METRIC, tags);
       authSuccess = meterRegistry.counter(AUTH_SUCCESS_METRIC, tags);
       authFailure = meterRegistry.counter(AUTH_FAILURE_METRIC, tags);
--- a/coordinator/persistence-api/src/main/java/io/stargate/db/metrics/api/ClientInfoMetricsTagProvider.java
+++ b/coordinator/persistence-api/src/main/java/io/stargate/db/metrics/api/ClientInfoMetricsTagProvider.java
@@ -46,15 +46,15 @@ public interface ClientInfoMetricsTagProvider {
    * @return Tags
    */
   default Tags getClientInfoTags(ClientInfo clientInfo) {
-    final DriverInfo driverInfo = clientInfo.driverInfo().get();
     List<Tag> tags = new ArrayList<>(2);
-    if (driverInfo != null) {
+    if (clientInfo.driverInfo().isPresent()) {
+      DriverInfo driverInfo = clientInfo.driverInfo().get();
       tags.add(Tag.of("driverName", driverInfo.name()));
-      String driverVersion = driverInfo.version().get();
-      if (driverVersion == null) {
-        driverVersion = "unknown";
+      if (driverInfo.version().isPresent()) {
+        tags.add(Tag.of("driverVersion", driverInfo.version().get()));
+      } else {
+        tags.add(Tag.of("driverVersion", "unknown"));
       }
-      tags.add(Tag.of("driverVersion", driverVersion));
     } else {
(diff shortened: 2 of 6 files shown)


tatu-at-datastax:
@maheshrajamani Looks like ITs fail -- transient issue? (all 3 seem to fail)

maheshrajamani:
> @maheshrajamani Looks like ITs fail -- transient issue? (all 3 seem to fail)

Had to fix an IT. Can you check the now? One IT failed with timeout error and trying to rerun it.

tatu-at-datastax:
> 

I had some timeout issues similarly last week. Not sure how common they are; no specific indicator to recognize them.

coordinator/testing-services/src/main/java/io/stargate/testing/metrics/FixedClientInfoTagProvider.java
```diff
@@ -28,9 +31,21 @@ public class FixedClientInfoTagProvider implements ClientInfoMetricsTagProvider
 
   public static final String TAG_KEY = "clientInfo";
   public static final String TAG_VALUE = "fixed";
+  public static final String TAG_KEY_DRIVER_NAME = "driverName";
+  public static final String TAG_VALUE_DRIVER_NAME = "java";
```
tatu-at-datastax, line 35: Since this is used for testing, maybe use a name that indicates it's for testing? Like `java-test-client` or something.

tatu-at-datastax (approved):
LGTM. My only concern would be if cardinality of driver names or versions could be problematic wrt metrics, but probably that is not yet problematic.

coordinator/persistence-api/src/main/java/io/stargate/db/metrics/api/ClientInfoMetricsTagProvider.java
```diff
@@ -28,9 +32,35 @@
...
+   *
+   * <p>Note that the implementation must return constant amount of tags for any input. IF the
+   * client info is populated with driver info, the tags should be populated with the driver name
+   * and version. Otherwise, the tags should be populated with "unknown" values.
+   *
+   * @param clientInfo {@link ClientInfo}
+   * @return Tags
+   */
+  default Tags getClientInfoTagsByDriver(ClientInfo clientInfo) {
+    List<Tag> tags = new ArrayList<>(2);
```
ivansenic, line 52: you forgot to include the `getClientInfoTags(ClientInfo clientInfo)` here, you are completely omitting it and having only two tags now?

coordinator/persistence-api/src/main/java/io/stargate/db/metrics/api/ClientInfoMetricsTagProvider.java
```diff
@@ -28,9 +32,35 @@
...
+    if (clientInfo.driverInfo().isPresent()) {
+      DriverInfo driverInfo = clientInfo.driverInfo().get();
+      tags.add(Tag.of("driverName", driverInfo.name()));
+      tags.add(Tag.of("driverVersion", driverInfo.version().orElse("unknown")));
+    } else {
+      tags.add(Tag.of("driverName", "unknown"));
+      tags.add(Tag.of("driverVersion", "unknown"));
```
ivansenic, line 59: * for the unknown value we should use the constant `StargateMetricConstants.UNKNOWN`
* `driverName` and `driverVersion` could be the public constants of this class, so implementing classes can override those if needed

coordinator/persistence-api/src/main/java/io/stargate/db/metrics/api/ClientInfoMetricsTagProvider.java
```diff
@@ -28,9 +32,35 @@
...
 
-  /** Returns default interface implementation, which returns empty tags for all methods. */
+  /**
+   * Returns default interface implementation, which returns empty tags or with driver info for all
+   * methods.
+   */
   ClientInfoMetricsTagProvider DEFAULT = new ClientInfoMetricsTagProvider() {};
 
+  /**
+   * Returns tags for a {@link ClientInfo}.
```
ivansenic, line 42: `Returns tags for a {@link ClientInfo} enriched with the driver tags if they are available`

coordinator/cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -290,7 +290,12 @@ public ConnectionMetricsImpl(ClientInfo clientInfo) {
               .map(clientInfoTagProvider::getClientInfoTags)
               .orElse(Tags.empty());
 
-      requestsProcessed = meterRegistry.counter(REQUESTS_PROCESSED_METRIC, tags);
+      Tags tagsByDriver =
```
ivansenic, line 293: why is tags and tags by driver separated now? Is it just to keep in in a single metric?

ivansenic (commented):
Except this one small issue, this looks good.. However, I don't think this will work and do what we want it to do.. You can ping me for explanation..

coordinator/cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
jeffreyscarpenter: This was at my request in order to cause the least disruption and potential metrics expansion. I'm not sure we ever look at the other metrics.

PR #2565, opened 2023-05-02
Add default tags

maheshrajamani:
**What this PR does**:
Add default tags based on client info to tags by driver.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/coordinator/cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
+++ b/coordinator/cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
@@ -278,7 +278,7 @@ public final class ClientMetrics {
   private class ConnectionMetricsImpl implements ConnectionMetrics {
 
     private final Tags tags;
-    private final Counter requestsProcessed;
+    private Counter requestsProcessed;
     private final Counter requestsDiscarded;
     private final Counter authSuccess;
     private final Counter authFailure;
@@ -325,6 +325,12 @@ public final class ClientMetrics {
     @Override
     public void markAuthError() {
       authError.increment();
+    }
+
+    @Override
+    public void updateDriverInfo(ClientInfo clientInfo) {
+      Tags tagsByDriver = clientInfoTagProvider.getClientInfoTagsByDriver(clientInfo);
+      requestsProcessed = meterRegistry.counter(REQUESTS_PROCESSED_METRIC, tagsByDriver);
     }
 
     @Override
--- a/coordinator/cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
+++ b/coordinator/cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
@@ -290,7 +290,12 @@ public final class ClientMetrics {
               .map(clientInfoTagProvider::getClientInfoTags)
               .orElse(Tags.empty());
 
-      requestsProcessed = meterRegistry.counter(REQUESTS_PROCESSED_METRIC, tags);
+      Tags tagsByDriver =
+          Optional.ofNullable(clientInfo)
+              .map(clientInfoTagProvider::getClientInfoTagsByDriver)
+              .orElse(Tags.empty());
+
+      requestsProcessed = meterRegistry.counter(REQUESTS_PROCESSED_METRIC, tagsByDriver);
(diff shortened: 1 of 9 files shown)


coordinator/persistence-api/src/main/java/io/stargate/db/metrics/api/ClientInfoMetricsTagProvider.java
```diff
@@ -38,6 +38,9 @@ public interface ClientInfoMetricsTagProvider {
    */
   ClientInfoMetricsTagProvider DEFAULT = new ClientInfoMetricsTagProvider() {};
 
+  public static final String TAG_KEY_DRIVER_NAME = "driverName";
+  public static final String TAG_KEY_DRIVER_VERSION = "driverVersion";
+  public static final String TAG_VALUE_UNKNOWN = "unknown";
```
ivansenic, line 43: remove, use the `StargateMetricsConstants.UNKNOWN`

coordinator/persistence-api/src/main/java/io/stargate/db/metrics/api/ClientInfoMetricsTagProvider.java
```diff
@@ -49,15 +52,17 @@ public interface ClientInfoMetricsTagProvider {
...
-      tags.add(Tag.of("driverVersion", driverInfo.version().orElse("unknown")));
+      tags.add(Tag.of(TAG_KEY_DRIVER_NAME, driverInfo.name()));
+      tags.add(Tag.of(TAG_KEY_DRIVER_VERSION, driverInfo.version().orElse(TAG_VALUE_UNKNOWN)));
     } else {
-      tags.add(Tag.of("driverName", "unknown"));
-      tags.add(Tag.of("driverVersion", "unknown"));
+      tags.add(Tag.of(TAG_KEY_DRIVER_NAME, TAG_VALUE_UNKNOWN));
+      tags.add(Tag.of(TAG_KEY_DRIVER_VERSION, TAG_VALUE_UNKNOWN));
     }
+    defaultTags.stream().forEach(tags::add);
```
ivansenic, line 65: No need to create a list here.. Simply use:

```
    Tags defaultTags = getClientInfoTags(clientInfo);
    if (clientInfo.driverInfo().isPresent()) {
      DriverInfo driverInfo = clientInfo.driverInfo().get();
      return defaultTags
          .and(TAG_KEY_DRIVER_NAME, driverInfo.name())
          .and(TAG_KEY_DRIVER_VERSION, driverInfo.version().orElse(StargateMetricConstants.UNKNOWN));
    } else {
      return defaultTags
              .and(TAG_KEY_DRIVER_NAME, StargateMetricConstants.UNKNOWN)
              .and(TAG_KEY_DRIVER_VERSION, StargateMetricConstants.UNKNOWN));
    }
```

ivansenic (approved):
See comments pls..

coordinator/cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -278,7 +278,7 @@ private static String metric(String name) {
   private class ConnectionMetricsImpl implements ConnectionMetrics {
 
     private final Tags tags;
-    private final Counter requestsProcessed;
+    private Counter requestsProcessed;
```
maheshrajamani, line 281: @ivansenic Connection metrics is created when new connection is created and driver info is available only when startup message is sent. So driver based tag metrics is refreshed on start up message.

coordinator/cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
```diff
@@ -327,6 +327,12 @@ public void markAuthError() {
       authError.increment();
     }
 
+    @Override
+    public void updateDriverInfo(ClientInfo clientInfo) {
+      Tags tagsByDriver = clientInfoTagProvider.getClientInfoTagsByDriver(clientInfo);
```
ivansenic, line 332: the only thing that worries me here is that line `296` above.. since the tags can change now, if the `clientInfo` is `null` in the constructor, we would register a metric with the empty tags and then with non-empty tags.. I am not sure if this is a problem.. and if this situation can happen at all.. but I also see not too many options here to fix..

ivansenic (approved):
Very nice :+1:

PR #2594, opened 2023-05-24
Add minimal column type for Vector values

### LINKED ISSUE #2593: Add support for experimental "VectorType" in `persistence-api`
Somewhat similar to CompositeType support (see https://github.com/stargate/stargate/pull/2393), there is need to add extension point for backend(s) to support upcoming experimental Vector type.
It is (at this point) a custom type (with no dedicated id) and will be supported by just one of backends. But for Stargate to handle it, there needs to at least be `Column`.

tatu-at-datastax:
**What this PR does**:

Fixes #2593

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/coordinator/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/ColumnUtils.java
+++ b/coordinator/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/ColumnUtils.java
@@ -98,9 +98,12 @@ public class ColumnUtils {
           encodeType(t, dest, version);
         }
         break;
+      case Composite: // Fallthrough intended
+      case DynamicComposite: // Fallthrough intended
+      case LineString: // Fallthrough intended
       case Point: // Fallthrough intended
       case Polygon: // Fallthrough intended
-      case LineString:
+      case Vector:
         // For custom types the class name of the type follows the type ID (which is zero)
         CBUtil.writeAsciiString(type.marshalTypeName(), dest);
         break;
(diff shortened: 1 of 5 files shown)


tatu-at-datastax:
@JeremiahDJordan Have been trying to figure out if this really is all that is needed from Stargate persistence-api side and I think it might be; binding to actual datatype is from `Conversion.java` of actual persistence implementation, where new `Column.Type.Vector` would be mapped to CC defined `org.apache.cassandra.db.marshal.VectorType` which has conversion functionality.

But I feel I may be omitting something as it all seems bit too easy.

tatu-at-datastax:
Ahhhh. Actually, I see code in `org.apache.cassandra.stargate.transport.internal.ColumnUtils` of `cql` module where some more work is probably needed.

tatu-at-datastax:
Ok additions to `ColumnUtils` was the ticket; could uncomment (Dynamic)CompositeType test as well.
This should be enough to be used by CC backend I think.

JeremiahDJordan:
Thanks!  I might be able to try this out tonight, but most likely in the morning.

tatu-at-datastax:
@JeremiahDJordan I think changes here alone won't work without changes to backend implementation fwtw.

tatu-at-datastax:
Note: to test would need to use something like:

```
create KEYSPACE tmp WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1};
use tmp;
create table test (pk int, str_val text, val float vector[3], PRIMARY KEY(pk));
INSERT INTO test (pk, str_val, val) VALUES (0, 'A', [1.0, 2.0, 3.0]);
select * from test;
```

tatu-at-datastax:
Replaced by #2597, needs handling of `dimensions`. Closing.

PR #2600, opened 2023-05-26
Remove SERVER_VERSION and PRODUCT_TYPE assertions

### LINKED ISSUE #2599: Remove SERVER_VERSION and PRODUCT_TYPE assertions
In the `SupportedOptionsTest`, we check the SERVER_VERSION and PRODUCT_TYPE if the backend is of, type DSE. In some of the variants of DSE, these are not available, so we need to remove the assertions for these two properties.

kathirsvn:
**What this PR does**:
In the `SupportedOptionsTest`, we check the SERVER_VERSION and PRODUCT_TYPE if the backend is of type DSE. In some of the variants of DSE, these are not available, hence removing them from the assertions.

Fixes #2599 

**Checklist**
- [X] Changes manually tested

### DIFF
--- a/coordinator/testing/src/main/java/io/stargate/it/cql/SupportedOptionsTest.java
+++ b/coordinator/testing/src/main/java/io/stargate/it/cql/SupportedOptionsTest.java
@@ -26,8 +26,6 @@ public class SupportedOptionsTest extends BaseIntegrationTest {
     assertThat(options).containsKey("COMPRESSION");
     if (backend.isDse()) {
       assertThat(options).containsKey("PAGE_UNIT");
-      assertThat(options).containsKey("SERVER_VERSION");
-      assertThat(options).containsKey("PRODUCT_TYPE");
       assertThat(options).containsKey("EMULATE_DBAAS_DEFAULTS");
     }
   }


coordinator/testing/src/main/java/io/stargate/it/cql/SupportedOptionsTest.java
```diff
@@ -26,8 +26,6 @@ public void supportedOptionsTest(CqlSession session) {
     assertThat(options).containsKey("COMPRESSION");
     if (backend.isDse()) {
       assertThat(options).containsKey("PAGE_UNIT");
-      assertThat(options).containsKey("SERVER_VERSION");
-      assertThat(options).containsKey("PRODUCT_TYPE");
```
ivansenic, line 30: do we have any other option? would it be better to said that CC is not DSE?

ivansenic (changes requested):
Don't like this solution, we could hit other bumbs on the road.. Can we shortly talk about this @kathirsvn ?

kathirsvn:
@tatu-at-datastax I agree. Reliably identifying the distinct backends and adding/skipping tests according would be the ideal solution. The reason, why I thought of removing the assertions is that, not all variants of DSE are going to have those fields.
@ivansenic I see you already made changes to skip those tests in CC, so we can close this PR now.

PR #2606, opened 2023-06-05
make sure DSE persistence jar is available for Dockerfile

### LINKED ISSUE #2605: DSE docker image fails with no persistence backend found error
When testing with a locally built copy of the DSE docker image, @tatu-at-datastax reported seeing this error message:

```
[SEVERE] Unexpected error: No persistence backend found for dse 6.8
java.lang.IllegalArgumentException: No persistence backend found for dse 6.8
```

jeffreyscarpenter:
Fixes #2605.

It's possible to generate a malformed local coordinator docker image using the `build_docker_images.sh` script if you forget to include the DSE package when doing a maven build (i.e. `mvn install -P dse`). If you omit the `-P dse` the `persistence-dse` JAR is not built, and then when you build the docker image the file is missing.

This PR adds a check to the `build_docker_images.sh` script to make sure the `persistence-dse` JAR is present.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/coordinator/build_docker_images.sh
+++ b/coordinator/build_docker_images.sh
@@ -55,6 +55,12 @@ echo "Building version $SGTAG"
 # Persistence images
 #
 
+# Verify existence of persistence jars
+if [ ! -f ./stargate-lib/persistence-dse*.jar ]; then
+  echo "No matching DSE Persistence jar found: ./stargate-lib/persistence-dse*.jar"
+  exit 1
+fi
+
 # Create a temp directory under current directory to use as a staging area for image creation
 # This is a workaround since Docker COPY command provides no way to exclude files.
 # We could build a base image and then remove files, but this is wasteful


tatu-at-datastax (approved):

tatu-at-datastax:
> I tested both options, with jar and without.. Without gives me warning 👍 With the jar the script now complains with:
> 
> ```
> ./build_docker_images.sh 
> Building version vWarning: JAVA_HOME environment variable is not set.
> 2.0.16-SNAPSHOT
> error: "docker buildx build" requires exactly 1 argument.
> See 'docker buildx build --help'.
> 
> Usage:  docker buildx build [OPTIONS] PATH | URL | -
> 
> Start a build
> ```
> 
> And I don't know why since there were no changes in the other parts..

Not sure how or why, but this works fine for me from this branch, with:

```
./mvnw clean install -Pdse -DskipTests
./build_docker_images.sh 
```

jeffreyscarpenter:
@ivansenic what was the environment in which you ran this? Just wondering why `JAVA_HOME` would not be set?

I believe the `Warning: JAVA_HOME environment variable is not set.` is coming from the maven command to retrieve the version number from the pom file. It appears that the warning message is being interpreted as part of the output so the version number is this string that includes spaces and a newline:

```
vWarning: JAVA_HOME environment variable is not set.
2.0.16-SNAPSHOT
```

so the first docker command ends up being something like

```
docker buildx build --target coordinator-4_0 --build-arg LIBDIR="./tmp-11282" -t stargateio/coordinator-4_0:vWarning: JAVA_HOME environment variable is not set.<newline>
``` 

Perhaps you could try running the `build_docker_images.sh` on the main branch to verify if this is really a problem unique to PR? 

Another thing to try: provide the tag explicitly: `./build_docker_images.sh -t v2.0.16-SNAPSHOT`

ivansenic (approved):
@jeffreyscarpenter yes that was it, the command outputs 

```
$ ./mvnw help:evaluate -Dexpression=project.version -q -DforceStdout
Warning: JAVA_HOME environment variable is not set.
2.0.16-SNAPSHOT
```

Not sure if needed, but maybe smth like this would help.. But then again it adds a risk of wrong regex which I generated with gpt :smile: :

```
$ ./mvnw help:evaluate -Dexpression=project.version -q -DforceStdout | sed -n '/[0-9]\+\(\.[0-9]\+\)*/p'
2.0.16-SNAPSHOT
```

PR #2608, opened 2023-06-07
Docker compose improvements for DSE on Mac M1 machines

jeffreyscarpenter:
updates to docker compose scripts for DSE: 
- increase memory allocation for DSE nodes to 2.5 GB
- reduce from 3 nodes to 2
- add platform option to force usage of amd64 architecture for DSE nodes, since DSE does not currently publish images with arm64 architecture. This enables running on Macs with M1.

Also adds recommended minimum Docker resource settings to READMEs for all supported backends.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/docker-compose/dse-6.8/docker-compose.yml
+++ b/docker-compose/dse-6.8/docker-compose.yml
@@ -3,9 +3,10 @@ version: '2'
 services:
   dse-1:
     image: datastax/dse-server:${DSETAG}
+    platform: linux/amd64
     networks:
       - stargate
-    mem_limit: 2G
+    mem_limit: 2560M
     cap_add:
       - IPC_LOCK
     ulimits:
@@ -24,9 +25,10 @@ services:
 
   dse-2:
     image: datastax/dse-server:${DSETAG}
+    platform: linux/amd64
     networks:
       - stargate
-    mem_limit: 2G
+    mem_limit: 2560M
     cap_add:
       - IPC_LOCK
     ulimits:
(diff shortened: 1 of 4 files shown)


docker-compose/dse-6.8/docker-compose.yml
```diff
@@ -3,9 +3,10 @@ version: '2'
 services:
   dse-1:
     image: datastax/dse-server:${DSETAG}
+    platform: linux/amd64
     networks:
       - stargate
-    mem_limit: 2G
+    mem_limit: 2560M
```
jeffreyscarpenter, line 9: On my M1 macbook pro, DSE nodes fail silently during startup without this memory increase. Not clear if this is due to upgrading DSE version or my machine architecture, but this doesn't harm anything.

docker-compose/dse-6.8/docker-compose.yml
```diff
@@ -3,9 +3,10 @@ version: '2'
 services:
   dse-1:
     image: datastax/dse-server:${DSETAG}
+    platform: linux/amd64
```
jeffreyscarpenter, line 6: enforcing platform architecture since this is the only architecture currently supported for DSE. Users will likely also require enablement of Docker's Rosetta emulation features to run this image on M1 macs.

docker-compose/dse-6.8/docker-compose.yml
```diff
@@ -47,30 +49,6 @@ services:
       timeout: 10s
       retries: 10
 
-  dse-3:
```
jeffreyscarpenter, line 50: To compensate for additional memory on DSE nodes, reducing cluster size from 3 to 2.

docker-compose/dse-6.8/docker-compose.yml
ivansenic, line 9: Maybe better idea would be to stick to 2G and decrease the max heap to 1G, currently `MAX_HEAP_SIZE=1536M`

docker-compose/dse-6.8/docker-compose.yml
ivansenic: Totally agree with this

ivansenic (approved):

docker-compose/dse-6.8/docker-compose.yml
jeffreyscarpenter: I tried this and the DSE containers ended up failing about 13 minutes after I started them. So probably this is cutting it too close.

tatu-at-datastax: Good that you managed to verify it at any rate. Definitely prefer working over non-working, but also good we know how close it is.

PR #2617, opened 2023-06-14
Remove use of errorprone static analyzer plugin (see #2616 for rationale)

### LINKED ISSUE #2616: Remove "error-prone" plug-in
Currently our build uses errorprone (https://github.com/google/error-prone) for some static validation, but we are not really looking much into it -- it just adds noise to build process. Worse, the version we use is old, and attempts at upgrading failed when they were attempted. This version does not work with JDK 11 (it's one of couple of blockers, even if smaller on).
Finally, the original reason for considering removal is the log pollution: since "errorprone" output has token "error", it adds noise when searching for ERROR level log entries.

While it would probably be possible to figure out upgrade path I think it's better to first remove it and then consider what static analysis too -- if any -- we'd like to use. Ideally it should be something less intrusive (errorprone requires its own compiler for `maven-compiler-plugin`) and something team finds valuable.

tatu-at-datastax:
**What this PR does**:

Removes use of "errorprone" static analyzer Java plug-in, related annotations, to help clean up logs and remove one (minor) blocker to allow use of JDK 11 for eventually building and running coordinators.

Fixes #2616

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/coordinator/graphqlapi/pom.xml
+++ b/coordinator/graphqlapi/pom.xml
@@ -116,10 +116,6 @@
       <version>0.9.1</version>
     </dependency>
     <dependency>
-      <groupId>com.google.errorprone</groupId>
-      <artifactId>error_prone_annotations</artifactId>
-    </dependency>
-    <dependency>
       <groupId>com.apollographql.federation</groupId>
       <artifactId>federation-graphql-java-support</artifactId>
       <version>2.0.3</version>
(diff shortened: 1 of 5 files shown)


jeffreyscarpenter (approved):
looks good

tatu-at-datastax:
I assume this is not controversial based on earlier discussions (I think "meh" was the attitude wrt error-prone), so will merge since this removes a minor obstacle for JDK 11 builds.

PR #2629, opened 2023-06-16
add java exports to stargate extension cmd line

jeffreyscarpenter:
Should allow ITs to pass. ITs use StargateExtension which builds its own command line. This PR incorporates the `add-exports` flag previously added to `starctl`


tatu-at-datastax (approved):

jeffreyscarpenter:
merged prior to full test completion in the interest of speed as we bootstrap this branch.

PR #2631, opened 2023-06-19
closes #2623: extend prepared with info with var indexes

### LINKED ISSUE #2623: NPE in `persistence-dse-6.8` class `Conversion.java` (line 580)
Looks like there is an NPE from DSE-6.8 backend, on line 580 (method `toResultInternal()`), observed in a production setting. 

Partial stack trace:

```
ERROR [CoreThread-1] 2023-06-15 08:37:02,753 Conversion.java:609 - Unexpected error thrown by persistence
java.lang.NullPointerException: null
        at io.stargate.db.dse.impl.Conversion.toResultInternal(Conversion.java:580)
        at io.stargate.db.dse.impl.Conversion.toResult(Conversion.java:554)
        at io.stargate.db.dse.impl.DsePersistence$DseConnection.lambda$executeRequest$1(DsePersistence.java:574)
        at io.reactivex.internal.operators.single.SingleMap$MapSingleObserver.onSuccess(SingleMap.java:57)
```

Line itself (if line numbers from stack trace match latest sources) would be:

```
            toResultMetadata(prepared.resultMetadata, null),
```

within:

```
      case PREPARED:
        ResultMessage.Prepared prepared = (ResultMessage.Prepared) resultMessage;
        PreparedWithInfo preparedWithInfo = (PreparedWithInfo) prepared;
        QueryHandler.Prepared preparedStatement =
            QueryProcessor.instance.getPrepared(prepared.statementId);
        return new Result.Prepared(
            Conversion.toExternal(prepared.statementId),
            Conversion.toExternal(prepared.resultMetadataId),
            toResultMetadata(prepared.resultMetadata, null),
            toPreparedMetadata(
                prepared.metadata.names,
                preparedStatement.statement.getPartitionKeyBindVariableIndexes()),
            preparedWithInfo.isIdempotent(),
            preparedWithInfo.isUseKeyspace());
```

but that doesn't make sense -- maybe line numbers are off, and they are not running from latest in `v1`? -- since if `prepared` was `null` it should already fail on line 576?!

**EDIT**: earlier fix #2552 was for seemingly similar (NPE) issue.

ivansenic:
**What this PR does**:
Removes the call to `QueryProcessor.instance.getPrepared` in the conversion. To do so, we need to extend the `PreparedWithInfo` with the data directly in the query processor.

Note that this fix is done only for DSE, as the Cassandra persistence implementations are already using the same solution. How and why was this not applied to DSE, I don't know.

Basing on `v1`.

Fixes #2623

### DIFF
--- a/coordinator/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java
+++ b/coordinator/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/Conversion.java
@@ -22,10 +22,8 @@ import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 import org.apache.cassandra.cql3.ColumnSpecification;
 import org.apache.cassandra.cql3.PageSize;
-import org.apache.cassandra.cql3.QueryHandler;
 import org.apache.cassandra.cql3.QueryOptions;
 import org.apache.cassandra.cql3.QueryOptions.PagingOptions;
-import org.apache.cassandra.cql3.QueryProcessor;
 import org.apache.cassandra.cql3.ResultSet;
 import org.apache.cassandra.cql3.statements.BatchStatement;
 import org.apache.cassandra.db.Clustering;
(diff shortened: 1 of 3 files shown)


tatu-at-datastax (approved):
Ok I like this, clean and straight-forward!

PR #2642, opened 2023-06-21
Fix DSE ITs for Java 11

jeffreyscarpenter:
For Stargate 2.1, ITs using DSE fail with Java 11 because DSE fails on startup with the following error message:

```
java.lang.ClassNotFoundException: jdk.internal.misc.VM not found by io.stargate.db.dse
```

In #2628 @tatu-at-datastax made some progress on getting ITs to run for Stargate v2.1 branch by adding the `add-extends` directive to load specific modules. It appears that for DSE tests to pass we actually need deeper access to non-public classes. This PR changes `add-extends` to `add-opens` which allows the private access that DSE requires.

**Checklist**
- [x] Changes manually tested
- [x] Automated Tests added/updated
- [ ] Documentation added/updated
- [x] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

PR #2643, opened 2023-06-22
Fix DSE ITs for Java 11 (second attempt)

jeffreyscarpenter:
see #2642


tatu-at-datastax:
@jeffreyscarpenter So this is all that was needed? :-D

Wrt other options, add-opens/add-extends; yeah, I played with minimal set needed and had come to conclusion that for initial passing what I had was enough. But obviously always possible something more might be needed.

PR #2649, opened 2023-06-27
preserve path element encoding on REST API

jeffreyscarpenter:
preserve path element encoding on REST API

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/Sgv2RowsResourceApi.java
+++ b/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/Sgv2RowsResourceApi.java
@@ -127,6 +127,7 @@ public interface Sgv2RowsResourceApi {
           final String tableName,
       @Parameter(name = "primaryKey", ref = RestOpenApiConstants.Parameters.PRIMARY_KEY)
           @PathParam("primaryKey")
+          @Encoded
           List<PathSegment> path,
       @Parameter(name = "fields", ref = RestOpenApiConstants.Parameters.FIELDS)
           @QueryParam("fields")
@@ -237,6 +238,7 @@ public interface Sgv2RowsResourceApi {
           final String tableName,
       @Parameter(name = "primaryKey", ref = RestOpenApiConstants.Parameters.PRIMARY_KEY)
           @PathParam("primaryKey")
+          @Encoded
           List<PathSegment> path,
       @Parameter(name = "raw", ref = RestOpenApiConstants.Parameters.RAW) @QueryParam("raw")
           final boolean raw,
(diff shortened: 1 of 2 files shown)


jeffreyscarpenter:
needs test case added

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/Sgv2RowsResourceApi.java
```diff
@@ -127,6 +127,7 @@ Uni<RestResponse<Object>> getRows(
           final String tableName,
       @Parameter(name = "primaryKey", ref = RestOpenApiConstants.Parameters.PRIMARY_KEY)
           @PathParam("primaryKey")
+          @Encoded
```
ivansenic, line 130: I am not sure what is this doing.. I also see test fail..

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/Sgv2RowsResourceApi.java
```diff
@@ -30,6 +30,7 @@
  */
 @ApplicationScoped
 @Path("/v2/keyspaces/{keyspaceName}/{tableName}")
+@Encoded
```
ivansenic, line 33: * What about the other resource classes?
* And what happens now if I send encoded keyspace and table name, I guess this fails on the bridge call as expected

jeffreyscarpenter:
I modified tests but not thoroughly enough. Current version does not pass a manual test with postman. Still investigating.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/Sgv2RowsResourceApi.java
tatu-at-datastax: I suspect this might be the right annotation to use, but placement may be wrong: I think it would belong only next to:

```
         @PathParam("primaryKey")
          List<PathSegment> path,
```

and/or anything needed wrt annotations should be there. We do not want to prevent decoding of all parameter (query, path) values.

tatu-at-datastax: FWTW, semantics for `@Encoded` are explained f.ex here:

https://docs.jboss.org/resteasy/docs/1.0.0.GA/userguide/html/_Encoded_and_encoding.html

jeffreyscarpenter: I tried applying the annotation to just the path on my first commit on this branch but it did not pass tests.

jeffreyscarpenter:
Closing in favor of #2658

PR #2651, opened 2023-06-30
Map data in REST API changed back to previous format for compatibility

### LINKED ISSUE #2577: Possible Serialization incompatibility for `Map`s between SGv2/SGv1 for REST API (v2)
(from user comment)

For table definition

```
CREATE TABLE sensor_data.sensors_by_network (
    network text,
    sensor text,
    characteristics map<text, text>,
    latitude decimal,
    longitude decimal,
    PRIMARY KEY (network, sensor)
)
```

we get results like Stargate V2 / RestV2:

```
{
    "count": 3,
    "data": [
        {
            "network": "forest-net",
            "sensor": "s1001",
            "characteristics": {
                "accuracy": "medium",
                "sensitivity": "high"
            },
            "latitude": 30.526503,
            "longitude": -95.582815
        },
       ...
}
```

vs Stargate V1 / RESTv2

```
{
    "count": 3,
    "data": [
        {
            "sensor": "s1001",
            "characteristics": [
                {
                    "key": "accuracy",
                    "value": "medium"
                },
                {
                    "key": "sensitivity",
                    "value": "high"
                }
            ],
            "network": "forest-net",
            "latitude": "30.526503",
            "longitude": "-95.582815"
        },
       ...
}
```

so `characteristics` property seems to be serialized differently.
Looks like Stargate V1 / RESTv2 may be wrapping `map` valued properties in an array somehow.

kathirsvn:
**What this PR does**:
Fixes https://github.com/stargate/stargate/issues/2577

Changes to accept and return map data in the old format in the REST API

For example
```json
"characteristics": [
                {
                    "key": "accuracy",
                    "value": "medium"
                },
                {
                    "key": "sensitivity",
                    "value": "high"
                }
            ]
```

instead of 
```json
"characteristics": {
                "accuracy": "medium",
                "sensitivity": "high"
            }
```

Fixes #2577 

**Checklist**
- [X] Changes manually tested

### DIFF
--- a/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/StringifiedValueUtil.java
+++ b/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/StringifiedValueUtil.java
@@ -1,9 +1,7 @@
 package io.stargate.sgv2.restapi.grpc;
 
 import io.stargate.bridge.proto.QueryOuterClass;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 /**
  * Helper class that deals with "Stringified" variants of structured values.
@@ -12,6 +10,8 @@ import java.util.Map;
  * io.stargate.web.resources.Converters}), unchanged.
  */
 public class StringifiedValueUtil {
+  private static final String[] entriesToCheck = new String[] {"key", "value"};
+
   public static void decodeStringifiedCollection(
       String value,
       ToProtoValueCodec elementCodec,
@@ -90,12 +90,12 @@ public class StringifiedValueUtil {
     if (idx >= value.length()) {
       throw new IllegalArgumentException(
           String.format(
-              "Invalid Map value '%s': at character %d expecting '{' but got EOF", value, idx));
+              "Invalid Map value '%s': at character %d expecting '[' but got EOF", value, idx));
     }
-    if (value.charAt(idx++) != '{') {
+    if (value.charAt(idx++) != '[') {
       throw new IllegalArgumentException(
           String.format(
-              "Invalid Map value '%s': at character %d expecting '{' but got '%c'",
+              "Invalid Map value '%s': at character %d expecting '[' but got '%c'",
               value, idx, value.charAt(idx)));
     }
 
@@ -104,24 +104,35 @@ public class StringifiedValueUtil {
     if (idx >= value.length()) {
       throw new IllegalArgumentException(
           String.format(
-              "Invalid Map value '%s': at character %d expecting element or '}' but got EOF",
+              "Invalid Map value '%s': at character %d expecting element or ']' but got EOF",
               value, idx));
     }
-    if (value.charAt(idx) == '}') {
+    if (value.charAt(idx) == ']') {
       return;
     }
-
+    int e = 0;
     while (idx < value.length()) {
-      int n = skipCqlValue(value, idx);
+      idx = skipSpaces(value, idx);
+      if (e % 2 == 0) {
+        if (value.charAt(idx) == '{') {
+          idx++;
+          idx = skipSpaces(value, idx);
+        } else {
+          throw new IllegalArgumentException(
+              String.format(
+                  "Invalid map value '%s': invalid CQL value at character %d", value, idx));
+        }
+      }
+      if (value.charAt(idx) == '}') {
+        return;
+      }
+      int n = skipKeyName(value, idx, entriesToCheck[e % 2]);
       if (n < 0) {
         throw new IllegalArgumentException(
             String.format("Invalid map value '%s': invalid CQL value at character %d", value, idx));
       }
-
-      QueryOuterClass.Value k =
-          keyCodec.protoValueFromStringified(handleSingleQuotes(value.substring(idx, n)));
+      // skip reading "key"
       idx = n;
-
       idx = skipSpaces(value, idx);
       if (idx >= value.length()) {
         throw new IllegalArgumentException(
@@ -143,10 +154,11 @@ public class StringifiedValueUtil {
       }
 
       QueryOuterClass.Value v =
-          valueCodec.protoValueFromStringified(handleSingleQuotes(value.substring(idx, n)));
+          e % 2 == 0
+              ? keyCodec.protoValueFromStringified(handleSingleQuotes(value.substring(idx, n)))
+              : valueCodec.protoValueFromStringified(handleSingleQuotes(value.substring(idx, n)));
       idx = n;
 
-      results.add(k);
       results.add(v);
 
       idx = skipSpaces(value, idx);
@@ -157,6 +169,10 @@ public class StringifiedValueUtil {
                 value, idx));
       }
       if (value.charAt(idx) == '}') {
+        idx++;
+        idx = skipSpaces(value, idx);
+      }
+      if (value.charAt(idx) == ']') {
         return;
       }
       if (value.charAt(idx++) != ',') {
@@ -167,9 +183,28 @@ public class StringifiedValueUtil {
       }
 
       idx = skipSpaces(value, idx);
+      e++;
     }
-    throw new IllegalArgumentException(
-        String.format("Invalid map value '%s': missing closing '}'", value));
+  }
+
+  private static int skipKeyName(String value, int idx, String keyName) {
+    while (idx < value.length() && (Character.isWhitespace(value.charAt(idx)))) {
+      idx++;
+    }
+    if (idx + keyName.length() + 2 > value.length()) {
+      return -1;
+    }
+    if (!value.startsWith("\"" + keyName + "\"", idx)
+        && !value.startsWith("'" + keyName + "'", idx)) {
+      return -1;
+    }
+    idx += keyName.length() + 2;
+    if (idx >= value.length()) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Invalid Map value '%s': at character %d expecting ':' but got EOF", value, idx));
+    }
+    return idx;
   }
 
   // Quite a bit of similarities with Collection handler but not enough to
--- a/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/ToProtoValueCodecs.java
+++ b/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/ToProtoValueCodecs.java
@@ -13,13 +13,7 @@ import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalTime;
 import java.time.format.DateTimeFormatter;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
 
 public class ToProtoValueCodecs {
   /**
@@ -804,12 +798,13 @@ public class ToProtoValueCodecs {
 
     @Override
     public QueryOuterClass.Value protoValueFromStrictlyTyped(Object mapValue) {
-      if (mapValue instanceof Map<?, ?>) {
+      if (mapValue instanceof List<?>) {
         // Maps are actually encoded as Collections where keys and values are interleaved
         List<QueryOuterClass.Value> elements = new ArrayList<>();
-        for (Map.Entry<?, ?> entry : ((Map<?, ?>) mapValue).entrySet()) {
-          elements.add(keyCodec.protoValueFromStrictlyTyped(entry.getKey()));
-          elements.add(valueCodec.protoValueFromStrictlyTyped(entry.getValue()));
+        for (Object entry : (List<?>) mapValue) {
+          Map<?, ?> entryNode = (Map<?, ?>) entry;
+          elements.add(keyCodec.protoValueFromStrictlyTyped(entryNode.get("key")));
+          elements.add(valueCodec.protoValueFromStrictlyTyped(entryNode.get("value")));
         }
         return Values.of(elements);
       }
--- a/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/FromProtoValueCodecs.java
+++ b/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/FromProtoValueCodecs.java
@@ -13,6 +13,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 
@@ -558,11 +559,12 @@ public class FromProtoValueCodecs {
     public Object fromProtoValue(QueryOuterClass.Value value) {
       QueryOuterClass.Collection coll = value.getCollection();
       int len = verifyMapLength(coll);
-      Map<Object, Object> result = new LinkedHashMap<>(len);
+      List<Object> result = new LinkedList<>();
       for (int i = 0; i < len; i += 2) {
-        result.put(
-            keyCodec.fromProtoValue(coll.getElements(i)),
-            valueCodec.fromProtoValue(coll.getElements(i + 1)));
+        Map<Object, Object> mapEntry = new LinkedHashMap<>();
+        mapEntry.put("key", keyCodec.fromProtoValue(coll.getElements(i)));
+        mapEntry.put("value", valueCodec.fromProtoValue(coll.getElements(i + 1)));
+        result.add(mapEntry);
       }
       return result;
     }
@@ -571,13 +573,14 @@ public class FromProtoValueCodecs {
     public JsonNode jsonNodeFrom(QueryOuterClass.Value value) {
       QueryOuterClass.Collection coll = value.getCollection();
       int len = verifyMapLength(coll);
-      ObjectNode map = jsonNodeFactory.objectNode();
+      ArrayNode list = jsonNodeFactory.arrayNode();
       for (int i = 0; i < len; i += 2) {
-        map.set(
-            keyCodec.jsonNodeFrom(coll.getElements(i)).asText(),
-            valueCodec.jsonNodeFrom(coll.getElements(i + 1)));
+        ObjectNode mapEntry = jsonNodeFactory.objectNode();
+        mapEntry.set("key", keyCodec.jsonNodeFrom(coll.getElements(i)));
+        mapEntry.set("value", valueCodec.jsonNodeFrom(coll.getElements(i + 1)));
+        list.add(mapEntry);
       }
-      return map;
+      return list;
     }
 
     private int verifyMapLength(QueryOuterClass.Collection mapValue) {
--- a/apis/sgv2-restapi/src/test/java/io/stargate/sgv2/restapi/grpc/FromProtoConverterTest.java
+++ b/apis/sgv2-restapi/src/test/java/io/stargate/sgv2/restapi/grpc/FromProtoConverterTest.java
@@ -1,21 +1,18 @@
 package io.stargate.sgv2.restapi.grpc;
 
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.jupiter.params.provider.Arguments.arguments;
-
 import io.stargate.bridge.grpc.CqlDuration;
 import io.stargate.bridge.grpc.Values;
 import io.stargate.bridge.proto.QueryOuterClass;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.LinkedHashSet;
-import java.util.Map;
-import java.util.Set;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.*;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
 
 public class FromProtoConverterTest {
   private static final String TEST_COLUMN = "test_column";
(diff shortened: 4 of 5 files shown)


kathirsvn:
@tatu-at-datastax 
I'm not sure in which scenario, `jsonNodeFrom` method of `MapCodec` in [apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/FromProtoValueCodecs.java](https://github.com/stargate/stargate/pull/2651/files#diff-90f201673652b4195c9612947e67b33ed609434c977f641362804d8078fba147) is invoked. Testing of that block is pending.

tatu-at-datastax:
One quick note on feature itself: if choosing the old serialization, data can not be deserialized back. This is not necessarily problematic (existing "raw" option has similar trade-offs).

apis/sgv2-restapi/CONFIGURATION.md
```diff
@@ -8,9 +8,10 @@
...
+| `stargate.rest.cql.disabled`      | `boolean` | `true`  | Whether /v2/cql endpoint should be disabled or not.                                                                                                                                                                                                                                                                                                                                                              |
+| `stargate.rest.optimize-map-data` | `boolean` | `true`  | When enabled, the map data will be of the below format for example:<br/>```characteristics: { "accuracy" : "medium" }``` <br/>when disabled:<br/>```characteristics: [ { "key" : "accuracy", "value" : "medium" } ]```<br/>_This flag can be overridden in the runtime using `optimize_map` flag in the data read and write APIs. Please check the API specification in the swagger-ui for API specific options_ |
```
tatu-at-datastax, line 14: I'd probably name it bit differently, as either "compact" (default: true) or "wrap-entries" (default: false)?

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/config/constants/RestApiConstants.java
```diff
@@ -0,0 +1,6 @@
+package io.stargate.sgv2.restapi.config.constants;
+
+public interface RestApiConstants {
```
tatu-at-datastax, line 3: Javadoc?

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/config/constants/RestOpenApiConstants.java
```diff
@@ -44,6 +44,7 @@ interface Parameters {
     String RAW = "raw";
     String TABLE_NAME = "tableName";
     String SORT = "sort";
+    String OPTIMIZE_MAP = "optimize_map";
```
tatu-at-datastax, line 47: Beside naming choice, shouldn't this be capitalized as `optimizeMap` (see `tableName` above)?

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/BridgeProtoValueConverters.java
```diff
@@ -27,15 +27,16 @@ public static BridgeProtoValueConverters instance() {
     return INSTANCE;
   }
 
-  public FromProtoConverter fromProtoConverter(List<QueryOuterClass.ColumnSpec> columns) {
+  public FromProtoConverter fromProtoConverter(
+      List<QueryOuterClass.ColumnSpec> columns, boolean optimizeMapData) {
```
tatu-at-datastax, line 31: Instead of adding separate flag, would it make sense to pass a config object, as we may well get more options in future?
Maybe that's for future expansion if and when that is needed I guess.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/Sgv2RowsResourceApi.java
```diff
@@ -96,7 +96,10 @@ Uni<RestResponse<Object>> getRowWithWhere(
       @Parameter(name = "raw", ref = RestOpenApiConstants.Parameters.RAW) @QueryParam("raw")
           final boolean raw,
       @Parameter(name = "sort", ref = RestOpenApiConstants.Parameters.SORT) @QueryParam("sort")
-          final String sort);
+          final String sort,
+      @Parameter(name = "optimize_map", ref = RestOpenApiConstants.Parameters.OPTIMIZE_MAP)
+          @QueryParam("optimize_map")
+          final boolean optimizeMap);
```
tatu-at-datastax, line 102: This should be `Boolean` since otherwise there will always be a value and defaults are never used?
If using wrapper, `null` value (missing query parameter) would use the default.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/Sgv2RowsResourceApi.java
```diff
@@ -140,7 +143,10 @@ Uni<RestResponse<Object>> getRows(
...
```
tatu-at-datastax, line 149: Same here and other places

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/config/constants/RestApiConstants.java
kathirsvn: Added now

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/config/constants/RestOpenApiConstants.java
kathirsvn: Fixed

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/BridgeProtoValueConverters.java
kathirsvn: I'm working on this. Can we call the config object as `RequestConfig`  or `RequestParams` since it's per request?

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/Sgv2RowsResourceApi.java
kathirsvn: Good catch. Fixed this now in all the API entry points and also started adding integration tests now, which also caught this issue. I'll add more integration tests and ll update.

For the integration tests, I'm following the pattern that you used for CQL enabled/disabled flag tests. Hope that's fine. Please review the Integration tests model I'm setting for this as well when you get a chance.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/Sgv2RowsResourceApi.java
kathirsvn: Fixed

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/CQLResource.java
```diff
@@ -85,6 +86,9 @@ public Uni<RestResponse<Object>> cqlQuery(
     return requestInfo
         .getStargateBridge()
         .executeQuery(query)
-        .map(response -> convertRowsToResponse(response, raw));
+        .map(
+            response ->
+                convertRowsToResponse(
+                    response, raw, optimizeMap || restApiConfig.optimizeMapData()));
```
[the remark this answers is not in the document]
kathirsvn: Fixed

kathirsvn:
> Ok, I can see the need for this, assuming there was existing production usage that expected "key"/"value" wrapping that was missed when evaluating usage.
> 
> Some things I'd like to discuss:
> 
> 1. We need to support serialization into JSON using old notation, but should we support deserialization? And if we do, I think there might be a problem if ser/deser must match -- so "key"/"value" were to be used either both on input and output or neither. But I thought (maybe I am wrong, need to double-check) that input used to NOT expect wrapping but output always included it? Alternatively would be to just ignore setting on deserialization and use the "optimized" style anyway; that'd be simpler code-wise too.
> 2. Naming of the feature: I guess it's all over the place so maybe "optimize" is fine, but if we wanted other name, now would be the time.
> 3. Handling of the new setting -- do we want to carry it through the API, or would it make sense to refactor to try to reduce either passing itself, or use config object to allow passing multiple settings?

1. The current version of the REST API is handling the map data in both input and the output in the same manner (i.e. with `key` and `value` fields separately. With the API level and the request level flag that we will be adding, it's going to be consistent with that behavior by default and a user can send the flag appropriately in read or write as per the requirement and override the default behavior
2. Naming : I agree, this might require a change, ll do it at the end once we finalize all in this PR. Should call that as `compactMapData` on the request level and `COMPACT_MAP_DATA` on the QUARKUS config level?
3. Handling of the setting : I tried my best to pass the flag only to the `MapCodec`. Passing a config object also sounds good, considering any flag that we might add in the future. So, I'm open to ideas on this.

kathirsvn:
The next step is to fill the empty implementation of tests in `RestApiV2QMapTestsImplIT.java`

apis/sgv2-restapi/CONFIGURATION.md
kathirsvn: This is now changed to `compactMapData` as discussed.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/BridgeProtoValueConverters.java
kathirsvn: Made changes to pass `RequestParams` config object from API end points so that we can add fields in the future if required for any new requirements

apis/sgv2-restapi/CONFIGURATION.md
```diff
@@ -8,9 +8,10 @@
...
+| `stargate.rest.cql.disabled`     | `boolean` | `true`  | Whether /v2/cql endpoint should be disabled or not.                                                                                                                                                                                                                                                                                                                                                                |
+| `stargate.rest.compact-map-data` | `boolean` | `true`  | When enabled, the map data will be of the below format for example:<br/>```characteristics: { "accuracy" : "medium" }``` <br/>when disabled:<br/>```characteristics: [ { "key" : "accuracy", "value" : "medium" } ]```<br/>_This flag can be overridden in the runtime using `compactMapData` flag in the data read and write APIs. Please check the API specification in the swagger-ui for API specific options_ |
```
tatu-at-datastax, line 14: Does this default to `true`? I thought we were thinking of defaulting to `false`.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/StringifiedValueUtil.java
```diff
@@ -172,6 +172,132 @@ public static void decodeStringifiedMap(
...
+  public static void decodeStringifiedMap(
+      String value,
+      ToProtoValueCodec keyCodec,
+      ToProtoValueCodec valueCodec,
+      Collection<QueryOuterClass.Value> results) {
+    int idx = skipSpaces(value, 0);
+    if (idx >= value.length()) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Invalid Map value '%s': at character %d expecting '[' but got EOF", value, idx));
```
tatu-at-datastax, line 184: Could the exception message indicate that this is "non-compact" map, so when we see decode problems it's clear which mode value was being decoded in? Similarly for existing code, indicating "compact" map representation being decoded.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/Sgv2RowsResourceImpl.java
```diff
@@ -47,13 +36,15 @@ public Uni<RestResponse<Object>> getRowWithWhere(
     } catch (IllegalArgumentException e) {
       throw invalidSortParameterException(e);
     }
-
+    final boolean compactMapData = compactMap != null ? compactMap : restApiConfig.compactMapData();
```
tatu-at-datastax, line 39: Maybe extract these 3 lines into a helper method (passing `restApiConfig`, Boolean `compactMap`, build & returns `ImmutableRequestParams`)?

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QIntegrationTestBase.java
```diff
@@ -247,6 +230,23 @@ protected Map<String, Object> map(
...
+  protected Map<String, Object> map(
+      String key1,
+      Object value1,
+      String key2,
+      Object value2,
+      String key3,
+      Object value3,
+      String key4,
+      Object value4) {
+    Map<String, Object> map = new LinkedHashMap<>();
```
tatu-at-datastax, line 242: Since we use JDK 17, could also use `Map.of(....)`

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QIntegrationTestBase.java
```diff
@@ -490,9 +500,25 @@ protected List<Map<String, String>> insertRows(
...
   protected List<Map<String, Object>> insertTypedRows(
       String keyspaceName, String tableName, List<Map<String, Object>> rows) {
+    return insertTypedRowsInternal(keyspaceName, tableName, rows, null);
+  }
+
+  protected List<Map<String, Object>> insertTypedRows(
+      String keyspaceName,
+      String tableName,
+      List<Map<String, Object>> rows,
+      Boolean compactMapData) {
```
tatu-at-datastax, line 510: Should probably pass primitive `boolean`, no need to pass `null`?

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QIntegrationTestBase.java
```diff
@@ -529,6 +582,25 @@ protected List<Map<String, Object>> findAllRowsAsList(String keyspaceName, Strin
     return readJsonAs(response, LIST_OF_MAPS_TYPE);
   }
 
+  protected JsonNode findRowsAsJsonNode(
+      String keyspaceName, String tableName, Boolean compactMapData, Object... primaryKeys) {
+    final String path = endpointPathForRowByPK(keyspaceName, tableName, primaryKeys);
+    RequestSpecification rquest = givenWithAuth();
```
tatu-at-datastax, line 588: rquest -> request

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QIntegrationTestBase.java
```diff
@@ -681,4 +767,131 @@ protected void setupMixedClusteringTestCase(String keyspaceName, String tableNam
...
+  protected String updateRowReturnResponse(String updatePath, boolean raw, Map<?, ?> payload) {
+    return updateRowReturnResponse(updatePath, raw, payload, HttpStatus.SC_OK);
+  }
+
+  protected String updateRowReturnResponse(
+      String updatePath,
+      boolean raw,
+      Map<?, ?> payloadMap,
+      int expectedStatus,
+      Boolean compactMapData) {
```
tatu-at-datastax, line 785: Boolean -> boolean?

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QMapCompactDisabledIT.java
```diff
@@ -0,0 +1,126 @@
...
+import io.stargate.sgv2.common.testresource.StargateTestResource;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+@QuarkusIntegrationTest
+@QuarkusTestResource(StargateTestResource.class)
+@TestProfile(RestApiV2QMapCompactDisabledIT.Profile.class)
+public class RestApiV2QMapCompactDisabledIT extends RestApiV2QIntegrationTestBase
+    implements RestApiV2QMapTests {
+  private static final boolean SERVER_FLAG = false;
```
tatu-at-datastax, line 16: Could use Javadoc comment -- and maybe rename if it's for compact-map-data setting.

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QMapTests.java
```diff
@@ -0,0 +1,31 @@
+package io.stargate.sgv2.it;
+
+public interface RestApiV2QMapTests {
```
tatu-at-datastax, line 3: Not sure why there's API/impl if it covers both tests -- adding Javadoc would be good, explaining the reasoning (it's for different default setting?)

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/ToProtoValueCodecs.java
```diff
@@ -804,22 +817,38 @@ public ToProtoValueCodec getValueCodec() {
...
+            elements.add(valueCodec.protoValueFromStrictlyTyped(entry.getValue()));
+          }
+          return Values.of(elements);
+        }
+      } else {
+        if (mapValue instanceof List<?>) {
+          List<QueryOuterClass.Value> elements = new ArrayList<>();
+          for (Object entry : (List<?>) mapValue) {
+            Map<?, ?> entryNode = (Map<?, ?>) entry;
+            elements.add(keyCodec.protoValueFromStrictlyTyped(entryNode.get("key")));
```
tatu-at-datastax, line 835: I think code here should validate, strictly that the entry has exactly two entries; one for `key`, one for `value`, and nothing else; and throw appropriate exceptions for missing or extra entries.
This would make debugging easier, otherwise I think we'd just get NPEs if one or both missing, and extra values silently ignored.

tatu-at-datastax (approved):
Looks good. Added some suggestions, mostly adding javadocs, and some additional validation.

kathirsvn:
> Looks good. Added some suggestions, mostly adding javadocs, and some additional validation.

Thank you @tatu-at-datastax. I'll incorporate the review comments and ll merge.

apis/sgv2-restapi/CONFIGURATION.md
kathirsvn: This will be `true` by default in the OSS version and this will be turned off using the flag on our deployment.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/Sgv2RowsResourceImpl.java
kathirsvn: Done

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QIntegrationTestBase.java
kathirsvn: There is a null check here in `getInsertRowRequest` method, where the request query param is either set or not set based on this Boolean wrapper object.

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QIntegrationTestBase.java
kathirsvn: This needs to be Boolean. There's a null check down stream.

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QIntegrationTestBase.java
kathirsvn: Removed the new method and used `Map.of`

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QMapCompactDisabledIT.java
kathirsvn: Changed the name and updated the JavaDoc as well

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QMapCompactIT.java
```diff
@@ -0,0 +1,115 @@
...
+import io.quarkus.test.common.QuarkusTestResource;
+import io.quarkus.test.junit.QuarkusIntegrationTest;
+import io.stargate.sgv2.common.testresource.StargateTestResource;
+import org.junit.jupiter.api.Test;
+
+@QuarkusIntegrationTest
+@QuarkusTestResource(StargateTestResource.class)
+public class RestApiV2QMapCompactIT extends RestApiV2QIntegrationTestBase
+    implements RestApiV2QMapTests {
+  private static final boolean SERVER_FLAG = true;
```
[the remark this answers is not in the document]
kathirsvn: Changed the name and updated the JavaDoc as well

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2QMapTests.java
kathirsvn: Added java doc

apis/sgv2-restapi/CONFIGURATION.md
tatu-at-datastax: Oh ok.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/ToProtoValueCodecs.java
kathirsvn: Agreed. Handled and added relevant test cases as well in this commit [aebff81](https://github.com/stargate/stargate/pull/2651/commits/aebff812897456ac0c5008df9385a12d81ad0465).

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/StringifiedValueUtil.java
```diff
@@ -275,6 +275,10 @@ public static void decodeStringifiedMap(
         idx = skipSpaces(value, idx);
       }
       if (value.charAt(idx) == ']') {
+        if (results.size() % 2 != 0) {
+          throw new IllegalArgumentException(
+              String.format("Invalid Non-Compact Map value: '%s'. Missing key or value.", value));
```
tatu-at-datastax, line 280: Wouldn't this mean it must be missing value, not key... ?

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/ToProtoValueCodecs.java
```diff
@@ -832,6 +832,12 @@ public QueryOuterClass.Value protoValueFromStrictlyTyped(Object mapValue) {
           List<QueryOuterClass.Value> elements = new ArrayList<>();
           for (Object entry : (List<?>) mapValue) {
             Map<?, ?> entryNode = (Map<?, ?>) entry;
+            if (entryNode.size() != 2) {
+              throw new IllegalArgumentException("Map entry must have exactly 2 elements");
+            }
+            if (!entryNode.containsKey("key") || !entryNode.containsKey("value")) {
+              throw new IllegalArgumentException("Map entry missing key or value");
```
tatu-at-datastax, line 839: Maybe just separate checks so you can specifically say if it is missing key or missing value

tatu-at-datastax (approved):

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/ToProtoValueCodecs.java
kathirsvn: Sure, I split the check now.

PR #2677, opened 2023-07-14
Reduce Guava usage/dependencies to help against transitive Guava dep conflicts

tatu-at-datastax:
**What this PR does**:

Reduces usage of Guava classes, to help reduce likelihood of conflicts between transitively included Guava versions; eventually needed to remove use of OSGi.

N/A

**Checklist**
- [x] Changes manually tested


jeffreyscarpenter (approved):
Looks good to me. How many Guava dependencies are left? (Are you planning a follow up PR to do more?)

tatu-at-datastax:
@jeffreyscarpenter Only handful, from fewer modules. I'll have to think about whether to use shaded Guava from java-driver; that is being used from tests at least. But yes, I think I will do another round at a later point.

PR #2678, opened 2023-07-14
Vector search integration to Stargate bridge 

maheshrajamani:
**What this PR does**:
Added vector type codec to stargate bridge
Added vsearch function to api query builder.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/coordinator/bridge/src/main/java/io/stargate/bridge/codec/VectorCodec.java
+++ b/coordinator/bridge/src/main/java/io/stargate/bridge/codec/VectorCodec.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright The Stargate Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.stargate.bridge.codec;
+
+import com.datastax.oss.driver.api.core.type.codec.TypeCodecs;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import io.stargate.bridge.grpc.Values;
+import io.stargate.bridge.proto.QueryOuterClass.Collection;
+import io.stargate.bridge.proto.QueryOuterClass.Value;
+import io.stargate.db.schema.Column;
+import io.stargate.db.schema.Column.ColumnType;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class VectorCodec extends CompositeCodec {
+
+  @Override
+  public ByteBuffer encode(@NonNull Value value, @NonNull ColumnType type) {
+    if (value.hasNull()) return null;
+    if (!value.hasCollection()) {
+      throw new IllegalArgumentException("Expected collection of float type");
+    }
+    Collection collection = value.getCollection();
+    int elementCount = collection.getElementsCount();
+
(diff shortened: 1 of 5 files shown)


coordinator/bridge/src/main/java/io/stargate/bridge/codec/VectorCodec.java
```diff
@@ -0,0 +1,88 @@
...
+  @Override
+  public ByteBuffer encode(@NonNull Value value, @NonNull ColumnType type) {
+    if (value.hasNull()) return null;
+    if (!value.hasCollection()) {
+      throw new IllegalArgumentException("Expected collection of float type");
+    }
+    Collection collection = value.getCollection();
+    int elementCount = collection.getElementsCount();
+
+    ByteBuffer[] encodedElements = new ByteBuffer[elementCount];
```
tatu-at-datastax, line 40: Seems very inefficient to construct separate `ByteBuffer` for every element, and then in the end create appended version. Instead of allocating one bigger `ByteBuffer` and appending directly into that. 
Would it be possible to change this? (I am guessing code is copied from driver or so).

tatu-at-datastax:
Ok looks good overall; added one note about inefficient handling that may be worth addressing.

But aside from that, some questions/comments:

1. Looks like we are basically encoding vector values as Collection (List) of floats, to avoid having to introduce new type? I guess that avoids having to change `protoc` definition (to add new type)
2. Should we make similar change to `grpc` side as well? Or maybe it makes sense to do this first, then refer to this PR in "add vector support to gRPC" to help copy over changes
3. I think there are plans for OSS Cassandra to support vectors of any/all types (or at least more than `float[]`) -- do you think we need to worry about that yet? It looks like value codecs may be flexible enough to make support possible without breakage (and error out if wrong content types encountered)

coordinator/bridge/src/main/java/io/stargate/bridge/codec/VectorCodec.java
maheshrajamani: Not sure how to avoid multiple ByteBuffer creation, we could avoid creating the array. But would keep it like this so the code will look similar to what's done in driver code [here](https://github.com/datastax/java-driver/blob/8101fbf9ec91d952e90022606090a4be9f24aace/core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/VectorCodec.java#L69). Let me know if this needs to be changed.

maheshrajamani:
> 1. Looks like we are basically encoding vector values as Collection (List) of floats, to avoid having to introduce new type? I guess that avoids having to change `protoc` definition (to add new type)

We use collection for most non primitive types on the bridge like map, tuple etc. So used the same for this.

> 2\. Should we make similar change to `grpc` side as well? Or maybe it makes sense to do this first, then refer to this PR in "add vector support to gRPC" to help copy over changes

Yes, gprc has separate set of codec same code can be copied, I saw tickets in your name for this. We should do a separate PR.

> 3\. I think there are plans for OSS Cassandra to support vectors of any/all types (or at least more than `float[]`) -- do you think we need to worry about that yet? It looks like value codecs may be flexible enough to make support possible without breakage (and error out if wrong content types encountered)

The VectorType class in Stargate don't have the parameterized type. Until then we can't do it here.

coordinator/bridge/src/main/java/io/stargate/bridge/codec/VectorCodec.java
tatu-at-datastax: Ok. If this is due to specific constraint like API of encoders not allowing encoding into existing `ByteBuffer` (but always allocating new one), then it is what it is. I was thinking encoder should be able to just add to given ByteBuffer but I guess that's not the case.

I agree it's good to have code similar to the original and they probably had a good reason to do it that way, even if it appears pretty wasteful.

So no need to change it, will approve.

tatu-at-datastax (approved):
LGTM

tatu-at-datastax:
> The VectorType class in Stargate don't have the parameterized type. Until then we can't do it here.

Understood. I was more wondering if there is anything here we should avoid doing to make it easier to change -- but I can't think of any specific problem, as we are checking types to ensure we don't blindly encode/decoce, so if underlying type did vary, we'd get specific exception message (and not like data corruption or wrong number of entries).

PR #2699, opened 2023-07-27
Add Vector-via-cql tests for Coordinator

### LINKED ISSUE #2698: Add CQL tests for `dse-next` backend to verify basic Vector functionality over cql
Now that we have "dse-next" backend with vector search capabilities, let's add simple ITs for verifying that basic vector functionality works over cql:

* Ability to create tables with vector column(s)
* Ability to insert rows with vector values
* Ability to perform vector search

Initial version should work more as smoke test / sanity check with just simple verification. More test cases may be added over time to cover regressions and such.
(that is, we are not validating full vector functionality but rather that basic functionality is exposed via cql).

tatu-at-datastax:
**What this PR does**:

Adds basic tests to verify Vector functionality access using CQL for dse-next backend

Fixes #2698

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/coordinator/persistence-test/src/main/java/io/stargate/it/storage/ExternalStorage.java
+++ b/coordinator/persistence-test/src/main/java/io/stargate/it/storage/ExternalStorage.java
@@ -192,6 +192,11 @@ public class ExternalStorage extends ExternalResource<ClusterSpec, ExternalStora
     public int errorsDetected() {
       return errorsInTest.get();
     }
+
+    @Override
+    public boolean supportsVSearch() {
+      return Boolean.getBoolean("stargate.test.backend.vsearch");
+    }
   }
 
   public static class CcmCluster extends Cluster {
--- a/coordinator/pom.xml
+++ b/coordinator/pom.xml
@@ -55,7 +55,7 @@
     <!-- Release -->
     <project.scm.id>github</project.scm.id>
     <doclint>none</doclint>
-    <driver.version>4.16.0</driver.version>
+    <driver.version>4.17.0</driver.version>
     <!-- First DropWizard version and things it directly depends on that make sense to sync -->
     <dropwizard.version>2.1.6</dropwizard.version>
     <!--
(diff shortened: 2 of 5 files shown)


tatu-at-datastax:
As expected, passes on `dse-next`, fails on other backends, due to test that runs against all backends currently. Need to figure out how to check for vector support (lack thereof) to skip tests on such backends.

coordinator/pom.xml
```diff
@@ -55,7 +55,7 @@
     <!-- Release -->
     <project.scm.id>github</project.scm.id>
     <doclint>none</doclint>
-    <driver.version>4.16.0</driver.version>
+    <driver.version>4.17.0</driver.version>
```
tatu-at-datastax, line 58: Needed to get access to vector-aware Result/Row accessors.

coordinator/testing/pom.xml
```diff
@@ -365,6 +365,7 @@
                     <ccm.version>4.0.10</ccm.version>
                     <stargate.test.nodes>1</stargate.test.nodes>
                     <stargate.test.persistence.module>persistence-cassandra-4.0</stargate.test.persistence.module>
+                    <stargate.test.backend.vsearch>false</stargate.test.backend.vsearch>
```
tatu-at-datastax, line 368: No clean way to auto-detect via driver or ccm so just add another system property for tests

coordinator/persistence-test/src/main/java/io/stargate/it/storage/ExternalStorage.java
```diff
@@ -192,6 +192,11 @@ public int testsSkipped() {
     public int errorsDetected() {
       return errorsInTest.get();
     }
+
+    @Override
+    public boolean supportsVSearch() {
+      return Boolean.getBoolean("stargate.test.backend.vsearch");
```
jeffreyscarpenter, line 198: defaults to false?

jeffreyscarpenter (approved):
Great addition!

coordinator/persistence-test/src/main/java/io/stargate/it/storage/ExternalStorage.java
tatu-at-datastax: Yes.

PR #2700, opened 2023-07-28
Add `size()` method for `ColumnType`, implemented by `VectorType`

tatu-at-datastax:
**What this PR does**:

As per title exposes VectorType `size`

N/A

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/coordinator/persistence-api/src/main/java/io/stargate/db/schema/VectorType.java
+++ b/coordinator/persistence-api/src/main/java/io/stargate/db/schema/VectorType.java
@@ -31,6 +31,11 @@ public class VectorType extends Column.DelegatingColumnType {
   }
 
   @Override
+  public int size() {
+    return dimensions;
+  }
+
+  @Override
   public int schemaHashCode() {
     return delegate.schemaHashCode() ^ SchemaHashable.hashCode(elementTypeName) + dimensions;
   }
--- a/coordinator/persistence-api/src/main/java/io/stargate/db/schema/Column.java
+++ b/coordinator/persistence-api/src/main/java/io/stargate/db/schema/Column.java
@@ -145,6 +145,10 @@ public abstract class Column implements SchemaEntity, Comparable<Column> {
       return frozen(true);
     }
 
+    default int size() {
+      throw new UnsupportedOperationException("ColumnType.size() not supported by " + rawType());
+    }
+
     default ColumnType of(ColumnType... types) {
       throw new UnsupportedOperationException("ColumnType.of() not supported by " + rawType());
     }


tatu-at-datastax:
@maheshrajamani I think this should work for size.

Supporting element metadata gets bit trickier, although it is possible. One challenge is that right now CNDB relies on Stargate 2.0.17 (or such) definition of `VectorType` so we have to keep interface compatible. It is possible to do that I think, but I wanted to check whether you absolutely need access to type parameter currently or not; if not, we could defer this for later point when there is actual support for different element types and we can be confident we know API and usage details.

maheshrajamani:
> One challenge is that right now CNDB relies on Stargate 2.0.17 (or such) definition of `VectorType` so we have to keep interface compatible.

It's in the Conversion.java that changes the value from Cassandra VectorType to Stargate VectorType.

coordinator/persistence-api/src/main/java/io/stargate/db/schema/VectorType.java
```diff
@@ -30,6 +30,11 @@ public String marshalTypeName() {
     return marshalTypeName;
   }
 
+  @Override
```
maheshrajamani, line 33: We need access to element type, with Cassandra 5.0 vector type is adding other data type support.

tatu-at-datastax: Yes, eventually. But is it needed now? As in, do we know how it'd be used? My only concern is with API, translating element type needs to be done carefully.

tatu-at-datastax:
> > One challenge is that right now CNDB relies on Stargate 2.0.17 (or such) definition of `VectorType` so we have to keep interface compatible.
> 
> It's in the Conversion.java that changes the value from Cassandra VectorType to Stargate VectorType.

Yes. But what I mean is that there is another backend implementation based on `dse-db-all` that CNDB team maintains in `vsearch` branch; and it depends on definition from `persistence-api`.
So we cannot change `Column.ColumnType` / `VectorType` implementations in an incompatible way: can add new factory method(s) (`of(...)`) in `VectorType` as long as retaining existing one.

And the challenge is that of handling of conversion of `Column.ColumnType` passed in. I guess I can try to make that work, as we will be testing it.

coordinator/bridge/src/test/java/io/stargate/bridge/codec/ValueCodecTest.java
```diff
@@ -415,22 +416,24 @@ public static Stream<Arguments> invalidListValues() {
   }
 
   public static Stream<Arguments> vectorValues() {
+    Column.ColumnType vectorType = VectorType.of("org.apache.cassandra.db.marshal.FloatType", 4);
```
tatu-at-datastax, line 419: I _think_ this is what `Conversion` produces (via cassandraType.getClass().getName()). Test does not (and can not) validate it so String passed can be anything. But test cannot depend on backend-specific element type either.

PR #2706, opened 2023-08-02
Fix #2653: add vector search CRUD support for REST API

### LINKED ISSUE #2653: Add vector CRUD support to REST API, happy path tests
Since we added support for the experimental `vector` CQL data type in #2597, we can take advantage of this to add vector support to the REST API. This will also require documentation updates.

tatu-at-datastax:
**What this PR does**:

Adds support for basic CRUD operations on Vector fields for REST API (but not yet vector search)

Fixes #2653

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/BridgeProtoTypeTranslator.java
+++ b/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/BridgeProtoTypeTranslator.java
@@ -75,6 +75,19 @@ public class BridgeProtoTypeTranslator {
           frozen = udtType.getFrozen();
           break;
         }
+      case VECTOR:
+        {
+          QueryOuterClass.TypeSpec.Vector vectorType = type.getVector();
+          desc =
+              "vector<"
+                  + cqlTypeFromBridgeTypeSpec(vectorType.getElement(), false)
+                  + ", "
+                  + vectorType.getSize()
+                  + ">";
+          frozen = false;
+          break;
+        }
+
       case SPEC_NOT_SET:
       default:
         throw new IllegalArgumentException("Undefined/unrecognized TypeSpec: " + type);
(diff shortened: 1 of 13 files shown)


tatu-at-datastax:
Work so far:

* Fixed a bug in `QueryBuilder` that was quoting number part of `vector<float, 1024>` type declaration (so table creation works)
* Add mapping of `VectorType` to gRPC protos (so table metadata GET works)

Next to be done

* Conversion from JSON to gRPC payload ("ToProtobuf") codec -- needed for INSERTs
* Conversion from  gRPC to JSON payload ("FromProtobuf") codec -- needed for SELECTs.

tatu-at-datastax:
Support for INSERTs, SELECTs added (happy-path only -- will create follow-up for more test cases)

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2VectorTestIT.java
```diff
@@ -0,0 +1,132 @@
...
+    createVectorTable(testKeyspaceName(), tableName);
+  }
+
+  /*
+  /////////////////////////////////////////////////////////////////////////
+  // Tests: Create, fail
+  /////////////////////////////////////////////////////////////////////////
+   */
+
+  // TODO: Add test trying to create vector column with non-float element
```
jeffreyscarpenter, line 51: should we try to complete these now?

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/grpc/BridgeProtoTypeTranslator.java
```diff
@@ -75,6 +75,19 @@ public static String cqlTypeFromBridgeTypeSpec(
           frozen = udtType.getFrozen();
           break;
         }
+      case VECTOR:
```
jeffreyscarpenter, line 78: I guess I didn't realize how much of the type translation code is specific to the REST API, vs having something reusable in sgv2-quarkus-common. Is there a lot of duplicated code between REST API and JSON API?

jeffreyscarpenter (approved):
No showstoppers but wondering if we should try to add more of the test cases currently labeled TODO.

apis/sgv2-restapi/src/test/java/io/stargate/sgv2/it/RestApiV2VectorTestIT.java
tatu-at-datastax: These are markers for #2715 so I am thinking of that as follow-up.

tatu-at-datastax:
> No showstoppers but wondering if we should try to add more of the test cases currently labeled TODO.

As per my other comment, those were added on purpose for #2715  :)

PR #2713, opened 2023-08-02
Fix #2712: make QueryBuilder avoid quoting numbers

### LINKED ISSUE #2712: Quarkus-common library `QueryBuilderImpl` incorrectly quotes size parameter of Vector type
Looks like type declaration

    vector<float, 5>

gets incorrectly quoted as

    vector<float, "5">

leading to failures at trying to construct tables with vector types.

This is because `CqlStrings.doubleQuoteUdts()` quotes any scalar terminal other than one of known scalar types (or pre-quoted terms).
One simple fix would be to recognize numbers as something not to quote.

tatu-at-datastax:
**What this PR does**:

Prevents quoting of numbers in type declarations like:

```
   vector<float, 5>
```

where currently token `5` would be quoted as '"5"` as it is not a known type token.

Fixes #2712

**Checklist**
- [x] Changes manually tested


maheshrajamani (approved):
LGTM

PR #2716, opened 2023-08-03
cassandra 4.0 docker compose updates

jeffreyscarpenter:
While doing some local testing with the Cassandra 4.0 docker compose, I observed an error running the `start_cass_40.sh` script. The Cassandra container was failing with a message to the effect that if you set `MAX_HEAP_SIZE` you must also set `HEAP_NEWSIZE`. This seems to be something introduced more recently? This PR updates the configuration to set both parameters and also removes the third Cassandra container as it doesn't really add much but consumes more resources.

**Checklist**
- [x] Changes manually tested


Yuqi-Du (approved):
This change works in my local environment. Made some api tests.

PR #2729, opened 2023-08-17
DocsAPI: Debug log to log the batch size when a doc is translated 

### LINKED ISSUE #2728: Log DocsAPI Batch extrapolation size
For debugging purposes, add logging around how much a Doc has been translated into, by the DOCS API before sending it to the Bridge service.

kathirsvn:
**What this PR does**:
Adds a debug log to log the batch size when a doc is translated by the DOCS API

Fixes #2728 

**Checklist**
- [X] Changes manually tested

### DIFF
--- a/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/WriteBridgeService.java
+++ b/apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/WriteBridgeService.java
@@ -592,7 +592,9 @@ public class WriteBridgeService {
                             .setValue(queriesConfig.consistency().writes())));
     batchQueries.forEach(batch::addQueries);
     Batch batchBuilt = batch.build();
-    logger.debug("Batch payload size : {}", batchBuilt.getSerializedSize());
+    if (logger.isDebugEnabled()) {
+      logger.debug("Batch payload size : {}", batchBuilt.getSerializedSize());
+    }
     return bridge.executeBatch(batchBuilt).map(QueryOuterClass.Response::getResultSet);
   }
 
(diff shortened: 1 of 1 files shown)


Yuqi-Du (approved):

apis/sgv2-docsapi/src/main/java/io/stargate/sgv2/docsapi/service/write/WriteBridgeService.java
```diff
@@ -599,8 +591,11 @@ private Uni<ResultSet> executeBatch(
                         QueryOuterClass.ConsistencyValue.newBuilder()
                             .setValue(queriesConfig.consistency().writes())));
     batchQueries.forEach(batch::addQueries);
-
-    return bridge.executeBatch(batch.build()).map(QueryOuterClass.Response::getResultSet);
+    Batch batchBuilt = batch.build();
+    if (logger.isDebugEnabled()) {
+      logger.debug("Batch payload size : {}", batchBuilt.getSerializedSize());
```
tatu-at-datastax, line 596: I know this is already merged, but I think this would have benefit from additional contextual information like number of requests in batch, and indication it's for Documents API -- plain "Batch payload size" is bit too generic (there's class name in log of course if one starts digging).

kathirsvn: @tatu-at-datastax I think there is a task planned around including tenant in all log messages and I believe that will add more context. An indication for docs api might not be needed, since this log is in the docs api and not in the commons. I agree with including the number of statements in the batch and I'll create a separate PR for that.

tatu-at-datastax: Good points, thank you!

Yeah, wrt Docs API I was thinking something like "Docs api batch payload..." or whatever wording to indicate it's needed to update a Document. But I let you decide what makes most sense.

kathirsvn: Makes sense. I'll change the log message as well.

PR #2748, opened 2023-08-25
Fixes #2747: add legacy /ping endpoint, tests to verify.

### LINKED ISSUE #2747: Add missing (/api/rest)`/` "ping" endpoint along with existing (/api/rest)`/health`
Looks like Stargate V2 needs to add something that in-Coordinator health-check had: secondary "ping" endpoint at root level. Looks like some existing installations (Astra DB) require this as well as "regular" health-check endpoint

tatu-at-datastax:
**What this PR does**:

Adds `/ping` endpoint (looks like Quarkus doesn't allow overriding `/` dynamically?) for SLA checker to call.

Fixes #2747

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/HealthResource.java
+++ b/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/HealthResource.java
@@ -24,7 +24,7 @@ import jakarta.ws.rs.core.Response;
 
 /** Legacy Health endpoint used by some routers to check if the service is up and running. */
 @Path("/")
-@Produces(MediaType.MEDIA_TYPE_WILDCARD)
+@Produces(MediaType.TEXT_PLAIN)
 @Singleton
 public class HealthResource {
   @GET
--- a/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/HealthResource.java
+++ b/apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/HealthResource.java
@@ -22,13 +22,20 @@ import jakarta.ws.rs.Produces;
 import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
 
+/** Legacy Health endpoint used by some routers to check if the service is up and running. */
 @Path("/")
-@Produces(MediaType.APPLICATION_JSON)
+@Produces(MediaType.MEDIA_TYPE_WILDCARD)
 @Singleton
 public class HealthResource {
   @GET
   @Path("/health")
   public Response health() {
-    return Response.status(Response.Status.OK).entity("UP").build();
+    return Response.status(Response.Status.OK).entity("UP [REST]").build();
+  }
+
+  @GET
+  @Path("/ping")
+  public Response ping() {
+    return Response.status(Response.Status.OK).entity("It's Alive [REST]").build();
   }
 }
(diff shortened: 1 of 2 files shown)


apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/HealthResource.java
```diff
@@ -22,13 +22,20 @@
 import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
 
+/** Legacy Health endpoint used by some routers to check if the service is up and running. */
 @Path("/")
-@Produces(MediaType.APPLICATION_JSON)
+@Produces(MediaType.TEXT_PLAIN)
```
tatu-at-datastax, line 27: Was copied along from v1, but since we do not produce JSON but plain text seems more correct.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/HealthResource.java
```diff
@@ -22,13 +22,20 @@
...
 @Singleton
 public class HealthResource {
   @GET
-  @Path("/health")
```
tatu-at-datastax, line 30: JAX-RS does not differentiate between leading slash and missing, other frameworks (Vertex) may.
Seems more correct not to indicate absolute path but related to class annotation.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/HealthResource.java
```diff
@@ -22,13 +22,20 @@
...
+  @Path("health")
   public Response health() {
-    return Response.status(Response.Status.OK).entity("UP").build();
+    return Response.status(Response.Status.OK).entity("UP [REST]").build();
```
tatu-at-datastax, line 33: I think various checkers do not validate contents, so for diagnostics/debugging purposes seems useful indicate source API. I am open to using some other notation but to make starting with "UP " makes sense for backwards compatibility.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/HealthResource.java
```diff
@@ -22,13 +22,20 @@
...
 
+/** Legacy Health endpoint used by some routers to check if the service is up and running. */
 @Path("/")
-@Produces(MediaType.APPLICATION_JSON)
+@Produces(MediaType.TEXT_PLAIN)
 @Singleton
 public class HealthResource {
   @GET
-  @Path("/health")
+  @Path("health")
```
tatu-at-datastax, line 31: JAX-RS does not differentiate between leading slash and missing, other frameworks (Vertex) may.
Seems more correct not to indicate absolute path but related to class annotation.

apis/sgv2-restapi/src/main/java/io/stargate/sgv2/restapi/service/resources/HealthResource.java
```diff
@@ -22,13 +22,20 @@
...
   public Response health() {
     return Response.status(Response.Status.OK).entity("UP").build();
   }
+
+  @GET
```
tatu-at-datastax, line 36: Adding at `/ping` because Quarkus reserves `/` for contents of `index.html`; was not able to find a way to override dynamically.
Not 100% sure we will be using this endpoint but at least it is defined should we want to do that.

tatu-at-datastax:
Will not pursue this (see #2747 for details); closing.

PR #2751, opened 2023-08-31
revert cql protocol v5

jeffreyscarpenter:
revert commit to add cql protocol v5 support as it seems to be causing issues for some clients. Hope to add it back in after we resolve issues. 

**Checklist**
- [x] Changes manually tested


tatu-at-datastax (approved):
Cannot really say much other than didn't notice anything obviosly off. But it's a huge diff (since original was).
And since we need to revert it ASAP, LGTM.

PR #2754, opened 2023-09-01
API controlled logging changes

### LINKED ISSUE #2730: Log Quarkus API Request Body for debugging
The reactive resteasy Quarkus plugin that we use for building the REST API doesn't have a way to log the payload received in a request and this makes debugging hard at times.

kathirsvn:
**What this PR does**:
Changes to support logging request url, body, status code based on tenant, methods, request url path, path prefixes response error codes, etc.

Fixes #2730 

**Checklist**
- [X] Changes manually tested

### DIFF
--- a/apis/sgv2-quarkus-common/CONFIGURATION.md
+++ b/apis/sgv2-quarkus-common/CONFIGURATION.md
@@ -76,7 +76,25 @@
 | `stargate.queries.consistency.schema-changes` | `String` | `LOCAL_QUORUM` | Consistency level to use for C* queries that are performing the schema changes.  |
 | `stargate.queries.consistency.writes`         | `String` | `LOCAL_QUORUM` | Consistency level to use for C* queries that are inserting or updating the data. |
 | `stargate.queries.consistency.reads`          | `String` | `LOCAL_QUORUM` | Consistency level to use for C* queries that are reading the data.               |
-| `stargate.queries.serial-consistency`         | `String` | `SERIAL`      | Serial consistency level to be used for C* queries.                              |
+| `stargate.queries.serial-consistency`         | `String` | `SERIAL`       | Serial consistency level to be used for C* queries.                              |
+
+### Additional Logging Configuration
+*Configuration mapping for the additional logging that are not covered by Quarkus out of the box such as request body, defined by [LoggingConfig.java](src/main/java/io/stargate/sgv2/api/common/config/LoggingConfig.java).*
+
+> NOTE: In order to use these features, you need to set the log level of `io.stargate.sgv2.api.common.logging.LoggingFilter` to `DEBUG` by doing below:
+```shell
+export QUARKUS_LOG_CATEGORY__IO_STARGATE_SGV2_API_COMMON_LOGGING_LOGGINGFILTER__LEVEL=DEBUG
+```
+
+| Property                                            | Type      | Default | Description                                                                                              |
+|-----------------------------------------------------|-----------|---------|----------------------------------------------------------------------------------------------------------|
+| `stargate.api.logging.enabled`                      | `boolean` | `false` | API level request logging is enabled.                                                                    |
+| `stargate.api.logging.enabled-tenants`              | `Set`     | `all`   | If set, only the requests from the tenants listed here will be logged. If empty, all tenants are logged. |
+| `stargate.api.logging.enabled-paths`                | `Set`     | `all`   | If set, only the requests to the paths listed here will be logged. If empty, all paths are logged.       |
+| `stargate.api.logging.enabled-path-prefixes`        | `Set`     | `*`     | If set, only the requests to the paths that start with the prefixes listed here will be logged.          |
+| `stargate.api.logging.enabled-error-codes`          | `Set`     | `all`   | If set, only the requests that return the error codes listed here will be logged.                        |
+| `stargate.api.logging.enabled-methods`              | `Set`     | `all`   | If set, only the requests that use the methods listed here will be logged.                               |
+| `stargate.api.logging.request-body-logging-enabled` | `boolean` | `false` | If the request body logging is enabled.                                                                  |
 
 ## Stargate Development Configuration
 
--- a/apis/sgv2-quarkus-common/CONFIGURATION.md
+++ b/apis/sgv2-quarkus-common/CONFIGURATION.md
@@ -86,15 +86,15 @@
 export QUARKUS_LOG_CATEGORY__IO_STARGATE_SGV2_API_COMMON_LOGGING_LOGGINGFILTER__LEVEL=DEBUG
 ```
 
-| Property                                            | Type      | Default | Description                                                                                              |
-|-----------------------------------------------------|-----------|---------|----------------------------------------------------------------------------------------------------------|
-| `stargate.api.logging.enabled`                      | `boolean` | `false` | API level request logging is enabled.                                                                    |
-| `stargate.api.logging.enabled-tenants`              | `Set`     | `all`   | If set, only the requests from the tenants listed here will be logged. If empty, all tenants are logged. |
-| `stargate.api.logging.enabled-paths`                | `Set`     | `all`   | If set, only the requests to the paths listed here will be logged. If empty, all paths are logged.       |
-| `stargate.api.logging.enabled-path-prefixes`        | `Set`     | `*`     | If set, only the requests to the paths that start with the prefixes listed here will be logged.          |
-| `stargate.api.logging.enabled-error-codes`          | `Set`     | `all`   | If set, only the requests that return the error codes listed here will be logged.                        |
-| `stargate.api.logging.enabled-methods`              | `Set`     | `all`   | If set, only the requests that use the methods listed here will be logged.                               |
-| `stargate.api.logging.request-body-logging-enabled` | `boolean` | `false` | If the request body logging is enabled.                                                                  |
+| Property                                            | Type      | Default | Description                                                                                                               |
+|-----------------------------------------------------|-----------|---------|---------------------------------------------------------------------------------------------------------------------------|
+| `stargate.api.logging.enabled`                      | `boolean` | `false` | API level request logging is enabled.                                                                                     |
+| `stargate.api.logging.enabled-tenants`              | `Set`     | `all`   | If set to value other than `all`, only the requests from the tenants listed here will be logged.                          |
+| `stargate.api.logging.enabled-paths`                | `Set`     | `all`   | If set to value other than `all`, only the requests to the paths listed here will be logged.                              |
+| `stargate.api.logging.enabled-path-prefixes`        | `Set`     | `all`   | If set to value other than `all`, only the requests to the paths that start with the prefixes listed here will be logged. |
+| `stargate.api.logging.enabled-error-codes`          | `Set`     | `all`   | If set to value other than `all`, only the requests that return the error codes listed here will be logged.               |
+| `stargate.api.logging.enabled-methods`              | `Set`     | `all`   | If set to value other than `all`, only the requests that use the methods listed here will be logged.                      |
+| `stargate.api.logging.request-body-logging-enabled` | `boolean` | `false` | If the request body logging is enabled.                                                                                   |
 
 ## Stargate Development Configuration
 
--- a/apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/logging/LoggingFilter.java
+++ b/apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/logging/LoggingFilter.java
@@ -19,7 +19,8 @@ package io.stargate.sgv2.api.common.logging;
 
 import static io.stargate.sgv2.api.common.config.constants.LoggingConstants.*;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Charsets;
+import com.google.common.io.CharStreams;
 import io.stargate.sgv2.api.common.StargateRequestInfo;
 import io.stargate.sgv2.api.common.config.LoggingConfig;
 import jakarta.enterprise.context.ApplicationScoped;
@@ -28,6 +29,7 @@ import jakarta.ws.rs.container.ContainerRequestContext;
 import jakarta.ws.rs.container.ContainerResponseContext;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.util.Collections;
 import java.util.Set;
 import org.jboss.resteasy.reactive.server.ServerResponseFilter;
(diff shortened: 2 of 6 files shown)


kathirsvn:
As an example, we can control logging of the requests with below params to the REST, Docs, GraphQL, APIs
```
-Dstargate.api.logging.enabled=true 
-Dquarkus.log.category.\"io.stargate.sgv2.api.common.logging.LoggingFilter\".level=DEBUG 
-Dstargate.api.logging.request-body-logging-enabled=true
-Dstargate.api.logging.enabled-path-prefixes=/v2/schemas/keyspaces/killrvideo
```

We have more options such as below. Please refer to the CONFIGURATION.md file for more info.
```
-Dstargate.api.logging.enabled-error-codes=400
-Dstargate.api.logging.enabled-tenants=test-id-1
-Dstargate.api.logging.enabled-methods=POST,DELETE
```

apis/sgv2-quarkus-common/CONFIGURATION.md
```diff
@@ -76,7 +76,25 @@
...
+> NOTE: In order to use these features, you need to set the log level of `io.stargate.sgv2.api.common.logging.LoggingFilter` to `DEBUG` by doing below:
+```shell
+export QUARKUS_LOG_CATEGORY__IO_STARGATE_SGV2_API_COMMON_LOGGING_LOGGINGFILTER__LEVEL=DEBUG
+```
+
+| Property                                            | Type      | Default | Description                                                                                              |
+|-----------------------------------------------------|-----------|---------|----------------------------------------------------------------------------------------------------------|
+| `stargate.api.logging.enabled`                      | `boolean` | `false` | API level request logging is enabled.                                                                    |
+| `stargate.api.logging.enabled-tenants`              | `Set`     | `all`   | If set, only the requests from the tenants listed here will be logged. If empty, all tenants are logged. |
+| `stargate.api.logging.enabled-paths`                | `Set`     | `all`   | If set, only the requests to the paths listed here will be logged. If empty, all paths are logged.       |
```
tatu-at-datastax, line 93: Is "all" then same as empty? (default is "all")

apis/sgv2-quarkus-common/CONFIGURATION.md
```diff
@@ -76,7 +76,25 @@
...
+| `stargate.api.logging.enabled-path-prefixes`        | `Set`     | `*`     | If set, only the requests to the paths that start with the prefixes listed here will be logged.          |
```
tatu-at-datastax, line 94: Is this a list of comma-separated Strings (I am guessing that's what `Set` means).
Are Strings exact matches, or regular expressions?

apis/sgv2-quarkus-common/CONFIGURATION.md
tatu-at-datastax: Looking at later definitions, it sounds like constant `"all"` means what "empy" refers to -- or "all" and "" are aliases. Maybe better to say "If set to value other than 'all', ..."?

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/logging/LoggingFilter.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  private String getRequestBody(ContainerRequestContext requestContext) {
+    try {
+      InputStream inputStream = requestContext.getEntityStream();
+      if (inputStream == null || !inputStream.markSupported()) {
+        return "";
+      }
+      inputStream.reset();
+      return objectMapper.readTree(inputStream).toString();
```
tatu-at-datastax, line 95: What is the point of decoding JSON, then re-encoding? Wouldn't it be easier to just read contents as String (assuming UTF-8 encoding)?

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/logging/LoggingFilter.java
```diff
@@ -0,0 +1,147 @@
...
+      return "";
+    }
+  }
+
+  private boolean isLoggingAllowed(
+      ContainerRequestContext requestContext, ContainerResponseContext responseContext) {
+    if (!loggingConfig.enabled()) {
+      return false;
+    }
+    Set<String> allowedTenants =
```
tatu-at-datastax, line 107: We could (and I think should) extract "contains all" setting when constructing filter, to by-pass construction of single "all" String set for the common case.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/logging/LoggingFilter.java
```diff
@@ -0,0 +1,147 @@
...
+        loggingConfig.enabledTenants().orElse(Collections.singleton(ALL_TENANTS));
+    if (!allowedTenants.contains(ALL_TENANTS)
+        && !allowedTenants.contains(requestInfo.getTenantId().orElse(""))) {
+      return false;
+    }
+    Set<String> allowedPaths =
+        loggingConfig.enabledPaths().orElse(Collections.singleton(ALL_PATHS));
```
tatu-at-datastax, line 114: Same here, can check inclusion is "all" or empty String, store in a flag when constructing filter, use that instead of checking for inclusion.

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/api/common/logging/LoggingFilterTest.java
```diff
@@ -0,0 +1,504 @@
...
+        new LoggingFilter(new ObjectMapper(), stargateRequestInfo, loggingConfig);
+    Method isAllowedMethod =
+        loggingFilter
+            .getClass()
+            .getDeclaredMethod(
+                "isLoggingAllowed", ContainerRequestContext.class, ContainerResponseContext.class);
+    isAllowedMethod.setAccessible(true);
+    boolean isAllowed =
+        (boolean) isAllowedMethod.invoke(loggingFilter, requestContext, responseContext);
+    assert !isAllowed;
```
tatu-at-datastax, line 50: Should use assert method from junit instead of Java language assert keyword?

tatu-at-datastax (commented):
Looks pretty good, got some questions that I'd like more info on before approving.

apis/sgv2-quarkus-common/CONFIGURATION.md
kathirsvn: I mean all the tenants by 'all' here. So, the 'all' in `enabled-paths` or leaving it not given means, log requests to all paths when `stargate.api.logging.enabled` is set to `true` and then we can control on which path to be logged by setting paths explicitly.

And all these parameter's values are treated as AND condition. 

**For example:**
If below is the configuration, then this means that all the `POST` requests to the path `/api/rest/v2/keyspaces` from the tenant `tenant-id1-1` that fails with `400` error will be logged along with its body.

```properties
stargate.api.logging.enabled-tenants=tenant-id-1
stargate.api.logging.enabled-paths=/api/rest/v2/keyspaces
stargate.api.logging.enabled-error-codes=400
stargate.api.logging.enabled-methods=POST
stargate.api.logging.request-body-logging-enabled=true
```

If the `stargate.api.logging.enabled-tenants=tenant-id-1` is not set, then it will be set to `all` by default, which means that all the `POST` requests to the path `/api/rest/v2/keyspaces` from **_any tenant_** that fails with `400` error will be logged along with its body.

apis/sgv2-quarkus-common/CONFIGURATION.md
kathirsvn: Yes, its the list of comma separated strings and they are treated as exact matches.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/logging/LoggingFilter.java
kathirsvn: Good catch. I originally thought of pretty printing the request body, which is available in the `readTree` method of `ObjectMapper`. Then I changed it to simple `toString` as it was logged in multiline which will make debugging harder if splunk parses them as separate lines. Let me change this to simple read from `InputStream` as string.

kathirsvn: Fixed in [f13a3eb](https://github.com/stargate/stargate/pull/2754/commits/f13a3eb82770bfaa7dc902f694f4285946909fe9)

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/logging/LoggingFilter.java
kathirsvn: Sorry, could you please explain it a little bit more. I'm not creating the filter up front.

apis/sgv2-quarkus-common/CONFIGURATION.md
tatu-at-datastax: I guess I am just wondering why have both "" and "all" and not just, say, "".

I do understand the logic itself, nothing wrong with that (AND is what I'd expect and I think most users too).

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/logging/LoggingFilter.java
tatu-at-datastax: You could determine whether "all" is included as an option (or if value is missing) right when `LoggingFilter` is constructed -- it will never change. So it does not need to be checked dynamically here, can check once and store as `boolean includeAllTenants` (or whatever).

apis/sgv2-quarkus-common/CONFIGURATION.md
tatu-at-datastax: In that case, shouldn't it be `"all"` for patc-prefixes too and not `*`?

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/logging/LoggingFilter.java
tatu-at-datastax: I guess it does not matter that much, slight performance overhead. Approving.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/logging/LoggingFilter.java
```diff
@@ -0,0 +1,145 @@
...
+  private static final Logger logger = org.slf4j.LoggerFactory.getLogger(LoggingFilter.class);
+
+  /** The request info bean. */
+  private final StargateRequestInfo requestInfo;
+  /** The configuration for logging. */
+  private final LoggingConfig loggingConfig;
+
+  /** Default constructor. */
+  @Inject
+  public LoggingFilter(StargateRequestInfo requestInfo, LoggingConfig loggingConfig) {
```
tatu-at-datastax, line 50: @kathirsvn Does this (injection of `StargateRequestInfo` mean that a new filter is created for every request)? I think it must mean that... is there no way around needing to get that injected? And instead get if from `ContainerRequestContext`?

apis/sgv2-quarkus-common/CONFIGURATION.md
kathirsvn: The `Default` column has what is the default value when it is not provided, which says 'all'. Having empty string "" as the default value will be confusing I believe.

apis/sgv2-quarkus-common/CONFIGURATION.md
kathirsvn: Makes sense. Let me change this too, to 'all'

apis/sgv2-quarkus-common/CONFIGURATION.md
kathirsvn: @tatu-at-datastax I fixed the config description as well as you suggested.

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/api/common/logging/LoggingFilterTest.java
kathirsvn: Fixed in [06db8de](https://github.com/stargate/stargate/pull/2754/commits/06db8de3d2aa336086250226249c76497bcbe1e2)

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/logging/LoggingFilter.java
kathirsvn: @tatu-at-datastax Great question! this discovers another issue I think.

`StargateRequestInfo` is `RequestScoped` and `LoggingFilter` is `ApplicationScoped`. So, if the `LoggingFilter` is designed to be shared across requests, I'm not sure how the request specific `StargateRequestInfo` ll be passed on cleanly.

I think we should either change `LoggingFilter` to `RequestScoped` or we should remove support for tenant filter in this feature, because parsing tenant from `ContainerRequestContext` or `ContainerResponseContext` is not possible I believe. I could see we use `RoutingContext` for resolving tenant info in `TenantResolver`

kathirsvn: Also, some of the Filter optimization can be done if `LoggingFilter` is going to be created only once in the applications life cycle.

kathirsvn: I will do some more debugging on what information we can get from `ContainerRequestContext` and `ContainerResponseContext` and see if getting tenant id be still achieved while leaving `LoggingFilter` as `ApplicationScoped`

tatu-at-datastax: On mis-matching scopes: Quarkus use proxies to handle this case, as I recall -- so marking it as "Application-scoped" does not fail, but requires use of proxy to delegate from single instance to per-request implementations. Or such.

But I think that since Tenant Id is not easily accessible (maybe it's injected by another filter? I am not sure how that is done), we probably just need to accept per-request construction.
LMK what you find!

tatu-at-datastax (approved):

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/logging/LoggingFilter.java
kathirsvn: @tatu-at-datastax 
I agree on the proxies part. I just debugged this and I can confirm that using proxies, the request scoped beans are updated on the ApplicationScoped instances.

Though the constructor of `LoggingFilter` is invoked only once on the first request, the `StargateRequestInfo` keeps changing when the `@ServerResponseFilter` method is invoked. i.e. The request scoped bean is changing as expected for each request. I changed the token between requests and I could see it changing when it reaches the filter method on that instance, so we are good, no need to change the filter to `RequestScoped`

Also, there is a predecessor for this design. `TenantRequestMetricsFilter` (which is `ApplicationScoped`) works the same way. So, I'm merging this.

PR #2760, opened 2023-09-07
Fix toResultMetadata conversion; only store first columnCount columns in list

michaeljmarshall:
**What this PR does**:

When converting `org.apache.cassandra.cql3.ResultSet.ResultMetadata` to `Result.ResultMetadata`, we should use the first metadata's `getColumnCount()` method to configure the number of columns in the `Result.ResultMetadata`.

This fix is important in cases where a result includes columns that the coordinator needed for ordering, but the client did not select. Currently, ordering columns are sent to clients because the use the count of columns instead of the `getColumnCount()`. (The naming does not seem ideal here...)

Here are some relevant references:

https://github.com/datastax/cassandra/blob/eb3632fdd23bae0bf06cf84ad023d4a3206c5970/src/java/org/apache/cassandra/cql3/ResultSet.java#L264-L280

https://github.com/datastax/cassandra/blob/020dbe2daa0f669c720d7ed57d8bf455fa218f48/src/java/org/apache/cassandra/cql3/ResultSet.java#L174-L185

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/coordinator/bridge/src/main/java/io/stargate/bridge/service/QueryHandler.java
+++ b/coordinator/bridge/src/main/java/io/stargate/bridge/service/QueryHandler.java
@@ -122,6 +122,7 @@ public class QueryHandler extends MessageHandler<Query, Prepared> {
           Result.Rows rows = (Result.Rows) result;
 
           if (enrichResponse) {
+            // TODO raw use of column. Is this correct?
             RowDecorator rowDecorator =
                 connection.makeRowDecorator(TableName.of(rows.resultMetadata.columns));
             responseBuilder.setResultSet(
--- a/coordinator/bridge/src/main/java/io/stargate/bridge/service/QueryHandler.java
+++ b/coordinator/bridge/src/main/java/io/stargate/bridge/service/QueryHandler.java
@@ -122,7 +122,6 @@ public class QueryHandler extends MessageHandler<Query, Prepared> {
           Result.Rows rows = (Result.Rows) result;
 
           if (enrichResponse) {
-            // TODO raw use of column. Is this correct?
             RowDecorator rowDecorator =
                 connection.makeRowDecorator(TableName.of(rows.resultMetadata.columns));
             responseBuilder.setResultSet(
--- a/coordinator/bridge/src/main/java/io/stargate/bridge/service/ValuesHelper.java
+++ b/coordinator/bridge/src/main/java/io/stargate/bridge/service/ValuesHelper.java
@@ -175,7 +175,8 @@ public class ValuesHelper {
       QueryOuterClass.ResumeMode resumeMode)
       throws StatusException {
     final List<Column> columns = rows.resultMetadata.columns;
-    final int columnCount = columns.size();
+    // 08-Sep-2023: As per https://github.com/stargate/stargate/pull/2760 may get extra columns:
+    final int columnCount = rows.resultMetadata.columnCount;
 
     ResultSet.Builder resultSetBuilder = ResultSet.newBuilder();
     if (!skipMetadata) {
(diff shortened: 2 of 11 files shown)


tatu-at-datastax:
@michaeljmarshall I think need for change makes sense based on description. 2 questions:

1. It'd be great to have tests to verify fix. What kind of query would exhibit the problem?
2. Is it guaranteed that the not-returned columns are sorted after ones for which values are returned? (or maybe expand a bit on logic that occurs here, wrt columns, values -- sounds like there are more columns specified than values per row returned)

tatu-at-datastax:
Hmmh. We both fixed formatting... not sure why fail occurred.

tatu-at-datastax:
Close+re-open just to force re-triggering of tests with latest (not sure why sometimes needed).

tatu-at-datastax:
Ah. Needed to use `./mvnw -Pdse fmt:format` to get DSE-6.8 part re-formatted. Re-running CI now.

michaeljmarshall:
>     1. It'd be great to have tests to verify fix. What kind of query would exhibit the problem?

I agree. I'll take a look at the repo to see where a test might make sense. There are two queries that will be improved by this fix. The first is vector ANN queries. The second is described by the original optimization that required this feature: https://issues.apache.org/jira/browse/CASSANDRA-4911.

>     2. Is it guaranteed that the not-returned columns are sorted after ones for which values are returned? (or maybe expand a bit on logic that occurs here, wrt columns, values -- sounds like there are more columns specified than values per row returned)

Yes, this is part of the feature added as part of https://issues.apache.org/jira/browse/CASSANDRA-4911. The extra columns are used for coordinator side ordering before sending to the client. The implementation from the C* code shows this guarantee:

https://github.com/datastax/cassandra/blob/020dbe2daa0f669c720d7ed57d8bf455fa218f48/src/java/org/apache/cassandra/cql3/ResultSet.java#L174-L185

tatu-at-datastax:
Failing test -- fails with an extra column before change -- added.
Appears to fail with the fix too so likely needs some more work.

michaeljmarshall:
Thanks for adding the test. I am seeing some clear encoding issues. I need to fix my local env to get tests working. I will be back online in a few hours to keep pushing this through.

coordinator/bridge/src/main/java/io/stargate/bridge/service/QueryHandler.java
```diff
@@ -122,6 +122,7 @@ protected CompletionStage<BridgeService.ResponseAndTraceId> buildResponse(Result
           Result.Rows rows = (Result.Rows) result;
 
           if (enrichResponse) {
+            // TODO raw use of column. Is this correct?
             RowDecorator rowDecorator =
                 connection.makeRowDecorator(TableName.of(rows.resultMetadata.columns));
```
michaeljmarshall, line 127: Tracking down one of the uses, it looks like we need to adjust this or we could hit the exception here:

https://github.com/stargate/stargate/blob/8c1208a461e68e3068fadb576f63ac812d78a1bf/coordinator/persistence-api/src/main/java/io/stargate/db/AbstractRowDecorator.java#L35-L52

michaeljmarshall:
@tatu-at-datastax - I found further evidence that the columns in the list can be filtered out. The only case when the `columns` list is increased in the `ResultSet` is when C* calls a method named `addNonSerializedColumns`.

https://github.com/datastax/cassandra/blob/ccb09c546b744cca6f684586a237cf690bb93f10/src/java/org/apache/cassandra/cql3/ResultSet.java#L209-L213

https://github.com/datastax/cassandra/blob/ccb09c546b744cca6f684586a237cf690bb93f10/src/java/org/apache/cassandra/cql3/ResultSet.java#L288-L299

At this point, I am confident we have the right fix. Once tests pass, I think we should merge and release.

tatu-at-datastax (approved):

PR #2779, opened 2023-09-19
compile and test with JDK 20

jeffreyscarpenter:
Draft PR using JDK 20

### DIFF
--- a/.github/workflows/apis-v21.yaml
+++ b/.github/workflows/apis-v21.yaml
@@ -39,11 +39,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up JDK 17
+      - name: Set up JDK 20
         uses: actions/setup-java@v3
         with:
           distribution: 'temurin'
-          java-version: '17'
+          java-version: '20'
 
       - name: Setup Maven
         env:
@@ -193,11 +193,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up JDK 17
+      - name: Set up JDK 20
         uses: actions/setup-java@v3
         with:
           distribution: 'temurin'
-          java-version: '17'
+          java-version: '20'
           cache: maven
 
       - name: Setup Maven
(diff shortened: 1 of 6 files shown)


tatu-at-datastax:
Temporarily changed to JDK 17 to see if that's an easier intermediate target -- does not seem to make a big difference unfortunately.

JeremiahDJordan:
> Temporarily changed to JDK 17 to see if that's an easier intermediate target -- does not seem to make a big difference unfortunately.

Yeah, 17 is actually where all the issues come from.  Going from 17 to 20 is a no-op from what I have seen in C* land.

JeremiahDJordan:
You might try compile with 11 and run tests with 20?  Not sure if that is possible with the Stargate build.

tatu-at-datastax:
> > Temporarily changed to JDK 17 to see if that's an easier intermediate target -- does not seem to make a big difference unfortunately.
> 
> Yeah, 17 is actually where all the issues come from. Going from 17 to 20 is a no-op from what I have seen in C* land.

Interesting. Did not know 20 might have fewer compatibility issues, was expecting reverse based on history. Will change to 20 then.

JeremiahDJordan:
> Did not know 20 might have fewer compatibility issues

It is not that it has fewer, it is just that it does not have any new ones over 17.

tatu-at-datastax:
@JeremiahDJordan The issue with Duzzt+QueryBuilder is strange in that somehow `apis/` (Quarkus-based, separate from Coordinato) has no problem running annotation pre-processing to create similar `QueryBuilder` on that side -- yet in-Coordinator one fails with various different root causes, seemingly related to StringTemplate (ST4), module/classpath access, version of ST4. With older one Duzzt 0.1.0 depends on, there's

```
[ERROR]   Caused by: java.lang.IllegalAccessException: class org.stringtemplate.v4.misc.ObjectModelAdaptor cannot access class com.sun.tools.javac.code.Symbol$MethodSymbol (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.code to unnamed module @2d85487
```

which I thought would be resolved by adding

```
      <plugin>
        <artifactId>maven-compiler-plugin</artifactId>
        <configuration>
          <compilerArgs>
            <arg>--add-opens</arg>
            <arg>jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED</arg>
          </compilerArgs>
        </configuration>
```

but has no effect it seems; is used (verified by adding bogus directive).

tatu-at-datastax:
Superceded by #2783 (which itself is blocked), closing.

PR #2782, opened 2023-09-21
Fix #2781: Coordinator QueryBuilder removal, attempt 2

### LINKED ISSUE #2781: Remove use of Duzzt-library as a pre-cursor to JDK 20 (runtime) upgrade
Looks like Duzzt library is not compatible with JDKs 17+, so it makes sense to get rid of its use.
One way to do that seems to be to use `-Impl` types, modify them; this works only when our use is limited (now that in-Coordinator APIs are removed except for Auth API).

Also it turns out that JPMS (module system) is blocking lots of Reflection access to JDK classes so need to add lots of JDK access overrides as well.

tatu-at-datastax:
**What this PR does**:

Changes `QueryBuilderImpl`, `SchemBuilderImpl` to be usable without code generation.

Fixes #2781

**Checklist**
- [x] Changes manually tested


tatu-at-datastax:
Rats. Wrong target, re-do.

PR #2783, opened 2023-09-21
(TEST) Change to run Coordinator tests on JDK 20

### LINKED ISSUE #2781: Remove use of Duzzt-library as a pre-cursor to JDK 20 (runtime) upgrade

tatu-at-datastax:
**What this PR does**:

Changes `QueryBuilderImpl`, `SchemaBuilderImpl` to be usable without Duzzt-based code generation.

Fixes #2781

**Checklist**
- [x] Changes manually tested


coordinator/persistence-api/pom.xml
```diff
@@ -30,13 +30,6 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
-      <!-- To upgrade StringTemplate version Duzzt uses, easiest to force here
```
tatu-at-datastax, line 33: Since Duzzt itself removed, no need to use (or override version of ) StringTemplate library.

coordinator/persistence-api/src/test/java/io/stargate/db/query/builder/BuiltSelectTest.java
```diff
@@ -69,57 +59,6 @@ public void testSelectColumnsNoWhere() {
...
-    BoundSelect select = checkedCast(query.bind());
-
-    assertBoundQuery(select, expectedQueryString);
-
-    assertThat(select.isStarSelect()).isFalse();
-    assertThat(names(select.selectedColumns())).isEqualTo(asSet("k2", "v1", "v2"));
-  }
-
-  public static Stream<Arguments> functionsToTest() {
-    Supplier<QueryBuilder.QueryBuilder__20> base = () -> newBuilder().select().column("k2", "v1");
```
tatu-at-datastax, line 86: This is the reason test needs to be removed -- refers to internal names of generated classes, and since generation removed now not easy to convert. But test itself not very valuable due to limited usage of this QueryBuilder variant.

coordinator/persistence-common/src/main/java/io/stargate/db/datastore/common/util/UserDefinedFunctionHelper.java
```diff
@@ -39,18 +40,32 @@ public static void fixCompilerClassLoader() {
...
       Field targetField = targetClass.getDeclaredField("insecureClassLoader");
-      targetField.setAccessible(true);
 
-      Field modifiers = Field.class.getDeclaredField("modifiers");
-      modifiers.setAccessible(true);
-      modifiers.setInt(targetField, targetField.getModifiers() & ~Modifier.FINAL);
-
-      targetField.set(null, targetClass.getClassLoader());
+      // As per https://stackoverflow.com/questions/74723932/java-17-reflection-issue
+      // we must use more complicated handling on JDK 12+
```
tatu-at-datastax, line 45: ... alas, does not seem to work on JDK 17 so instead remove calls altogether.

coordinator/pom.xml
```diff
@@ -43,15 +43,15 @@
...
+    <maven.compiler.target>20</maven.compiler.target>
     <!-- Encoding -->
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <!-- OSGi -->
     <osgi.bundle.version>${project.version}</osgi.bundle.version>
-    <!-- JDK 11 settings -->
-    <jdk11.module-exports>--add-exports java.base/jdk.internal.ref=ALL-UNNAMED --add-exports java.base/jdk.internal.misc=ALL-UNNAMED</jdk11.module-exports>
+    <!-- JDK 9+ JPMS settings -->
+    <jdk11.module-exports>--add-opens java.base/java.io=ALL-UNNAMED --add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.nio=ALL-UNNAMED --add-opens java.base/java.nio.file.spi=ALL-UNNAMED --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.util.concurrent=ALL-UNNAMED --add-opens java.base/java.util.concurrent.atomic=ALL-UNNAMED --add-opens java.base/jdk.internal.ref=ALL-UNNAMED --add-opens java.base/jdk.internal.misc=ALL-UNNAMED --add-opens java.base/sun.nio.ch=ALL-UNNAMED</jdk11.module-exports>
```
tatu-at-datastax, line 54: This is quite close to set of exports added in CNDB/vsearch. Using `--add-opens` since some reflection usage needs that (which gives more access than `--add-exports`). Probably not all entries strictly needed but were at some point in time and lots of work to run through the whole test suite to weed out ones not actually needed.

coordinator/pom.xml
```diff
@@ -93,18 +93,21 @@
       -->
     <netty.version>4.1.75.Final</netty.version>
     <netty-boringssl.version>2.0.51.Final</netty-boringssl.version>
-    <jacoco.version>0.8.8</jacoco.version>
+    <jacoco.version>0.8.10</jacoco.version>
```
tatu-at-datastax, line 96: Was necessary to prevent lots of noise about failed instrumentation (although I don't think we actually use code coverage reports for anything).

coordinator/persistence-common/src/main/java/io/stargate/db/datastore/common/util/UserDefinedFunctionHelper.java
tatu-at-datastax: But as per #2788, commenting out breaks DSE-6.8, DSE-next and C-4.0 backends. Issue, whatever it is, is indeed fixed in C-4.1.

tatu-at-datastax:
No longer needed, closing.

PR #2788, opened 2023-09-27
Remove calls to UserDefinedFunctionHelper.fixCompilerClassLoader() to test if needed

tatu-at-datastax:
**What this PR does**:

As per title, remove call to something that fails on JDK 17+ without extensive set of workarounds (or maybe even then).
Let run CI to see if any of the tests show workaround was needed.

N/A

**Checklist**
- [x] Changes manually tested


tatu-at-datastax:
Looks like work-around IS needed for:

* C-4.0
* DSE-6.8
* DSE-next (based on C-4.0.7)

but IS NOT needed for

* C-4.1

(as per code comments suggested).

Issue #2812, opened 2023-10-10
add our own logic expression classes


Yuqi-Du:
Jbool library support logic operations, but will use hashcomparator to sort all child expression.
which leads to JSON API unit tests fail (ValidatingStargateBridge)

tatu-at-datastax:
Could you elaborate a bit more on what is happening (why do we have test failures), and about suggested solution? (replace `Jbool_expressions` with straight-forward custom implementation or... ?)

Yuqi-Du:
Thanks, Tatu, I have attached explanations here  https://github.com/stargate/stargate/pull/2815

PR #2815, opened 2023-10-11
querybuilder and, or support 

Yuqi-Du:
**What this PR does**:
paired with [ JSONAPI PR ](https://github.com/stargate/jsonapi/pull/565)
- handle null for where(expression), insure we still build a cql in JSONAPI. eg. when filter body is empty, we find everything.
- remove jbool library, wrote our own logical expression classes
remove dependency of Jbool, although this library supports And.of(xx), Or.of(xx), it will store sub-expression 'randomly' (use hashcomparator). Which means if we have And.of(A,B), then in reality, it may store [B,A] instead of [A,B]. This is fine for all the functionality part. However, in the unit tests of JSONAPI, we use ValidatingStargateBridge and need to specify an expected cql, but the cql we generated is actually on-deterministic. Thus, random order of expression will make it hard to pass the unit tests.
Mahesh suggested to pass our customized comparator when using Jbool, but it is hard for some situations, eg, if you have And.of (And.of(A), And.of(B), Or.of(C) ) or even more nested situation, the expected order is complex. 
Finally, I wrote our Logical expression classes in Quarkus-common module to keep expression in order as when it was created. The implementation is straightforward, it stores expressions in the order of creation.

- remove parenthesis when there is only one sub-expression, 
eg, 
And.of(A) -> A
OR.of(A,B) -> (A or B)
And.of(Or.of(A,B,C)) -> (A or B or C)

Fixes [2812]( https://github.com/stargate/stargate/issues/2812)  [2789]( https://github.com/stargate/stargate/issues/2789) 

**Checklist**
- [x] Changes manually tested


amorton:
I am -1 on this commit - please do not merge it. 

We do not want logic handling like this in the service, we want the logic handling to be in Cassandra. Generate a CQL query that represents the logic the caller asked for, if we need to optimise we optimise Cassandra. This is a code design decision. 

I am also -1 on including the https://github.com/bpodgursky/jbool_expressions that explicitly says: 

"jbool_expressions is very much in-development, and is in no way, shape, or form guaranteed to be stable or bug-free. Bugs, suggestions, or pull requests are all very welcome."

and has not been updated since sept 2022

amorton (changes requested):
see comment, -1 in the conversation.

Yuqi-Du:
Thanks Aaron, Mahesh.

Probably Aaron misunderstood that we are reading and doing boolean operation in JSON API. We actually use it to pass along the logic information and finally generate a CQL query that represents the logic the caller asked for.

This library was used in docs api, and used with $in implementation of JSONAPI.

It is indeed that this library is not maintained, and we will look for an alternative.

tatu-at-datastax:
I don't think it is entirely correct to say that library is not maintained: disclaimer is just the usual "hey there could be bugs", I wouldn't read much to it.
That there hasn't been changes in 1 year is another thing, but looking at scope of the library that probably just means it's mature for its scope (since it's quite narrow in scope).
There are only 4 open issues, only 1 or 2 that are possible flaws.

I have no problem looking for replacement but I wouldn't be too alarmed, given that it seems to  be quite stable and working for us.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/cql/ExpressionUtils.java
```diff
@@ -0,0 +1,33 @@
+package io.stargate.sgv2.api.common.cql;
+
+import com.bpodgursky.jbool_expressions.And;
+import com.bpodgursky.jbool_expressions.Expression;
+import com.bpodgursky.jbool_expressions.Or;
+import java.util.List;
+
+public class ExpressionUtils<K> {
```
tatu-at-datastax, line 8: Add a simple javadoc explaining its set of utility methods for constructing logical expressions for use by QueryBuilder.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/cql/ExpressionUtils.java
```diff
@@ -0,0 +1,33 @@
...
+
+  public static <K> And<K> OrderedAndOf(Expression<K>... expressions) {
```
tatu-at-datastax, line 10: Is there need for non-ordered one? If not, just use `andOf()` and `orOf()`. Can clarify in Javadoc(s) that expressions are constructed in specified order.

tatu-at-datastax, line 10: On naming: should follow Java naming convention, so even if leaving "ordered" in there, should be

    public static .... orderedAndOf(...)

(that is, start with lower-case letter)

tatu-at-datastax:
@Yuqi-Du Unit test fail on CI, you need to run `./mvnw fmt:format` -- I think there's at least one unnecessary `import` statement that'll remove.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/cql/builder/QueryBuilderImpl.java
```diff
@@ -547,13 +547,8 @@ public void where(Collection<? extends BuiltCondition> where) {
...
   @DSLAction(autoVarArgs = false)
   public void where(Expression<BuiltCondition> whereExpression) {
-    Set<BuiltCondition> conditions = whereExpression.getAllK();
-    for (BuiltCondition condition : conditions) {
-      Variable<BuiltCondition> newConditionExpression = Variable.of(bindGrpcValues(condition));
-      Map<BuiltCondition, Expression<BuiltCondition>> conditionExpressionMap =
-          Collections.singletonMap(condition, newConditionExpression);
-      whereExpression =
-          whereExpression.replaceVars(conditionExpressionMap, new ExprFactory.Default<>());
+    if (whereExpression == null) {
```
tatu-at-datastax, line 550: or change to 

```
if (whereExpression != null) {
  setWhereExpression(whereExpression);
}
```

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/cql/builder/QueryBuilderImpl.java
```diff
@@ -1336,10 +1331,13 @@ private String updateQuery() {
   }
 
   private void appendWheres(StringBuilder builder) {
+    // JSON API fully relly on Expression<BuildCondition> instead of List<BuildCondition>
```
tatu-at-datastax, line 1334: relly -> rely

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/api/common/cql/builder/QueryBuilderValuesTest.java
```diff
@@ -131,40 +133,41 @@ public void shouldBindValueModifiers() {
...
   @Test
   public void shouldBindDirectWhereValues() {
-    Query query =
-        new QueryBuilder()
-            .select()
-            .from("ks", "tbl")
-            .where("c1", Predicate.EQ, INT_VALUE1)
-            .where("c2", Predicate.EQ, TEXT_VALUE)
-            .build();
+    Expression<BuiltCondition> expression =
```
tatu-at-datastax, line 136: Would old code not work here? If it would, maybe instead add a new method with new code, but leave old test as-is? This way we would verify backwards-compatibility for other APIs?

If this is not possible, does this mean that existing APIs query building code needs to be changed (I assume not, given that CI passes)

tatu-at-datastax:
@Yuqi-Du One bigger question on tests: it seems to me that it'd be better to add new tests that use ANDs (instead of implicit construction), if possible -- that way old way of building queries that other APIs use would still be covered. As well as new way of building queries that JSON API will use.

Yuqi-Du:
> @Yuqi-Du One bigger question on tests: it seems to me that it'd be better to add new tests that use ANDs (instead of implicit construction), if possible -- that way old way of building queries that other APIs use would still be covered. As well as new way of building queries that JSON API will use.

Thanks Tatu.
I did not delete the old way of building queries that old apis used. 
Added more tests for new building queries.

apis/sgv2-quarkus-common/src/test/java/io/stargate/sgv2/api/common/cql/builder/QueryBuilderValuesTest.java
Yuqi-Du: yes, CI passes, no impact to old apis.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/cql/builder/QueryBuilderImpl.java
Yuqi-Du: done

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/cql/builder/QueryBuilderImpl.java
Yuqi-Du: done

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/cql/ExpressionUtils.java
Yuqi-Du: done

Yuqi-Du: done

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/cql/ExpressionUtils.java
Yuqi-Du: done

tatu-at-datastax: Hmmh. I don't see any for the class itself? (Javadoc comment as in /** .... */ )

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/cql/ExpressionUtils.java
```diff
@@ -0,0 +1,71 @@
...
+import com.bpodgursky.jbool_expressions.Expression;
+import com.bpodgursky.jbool_expressions.Or;
+import java.util.List;
+
+/**
+ * Convenience expression builder
+ *
+ * <p>when construct jbool expression without specifying a comparator, it will use hashComparator by
+ * default which will cause the order of expression indeterminate, and cause JSONAPI unit tests
+ * failure By using this ExpressionUtils class, we pass a default comparator to make expression
```
tatu-at-datastax, line 28: make -> keep

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/cql/ExpressionUtils.java
tatu-at-datastax: Ok good!

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/cql/builder/QueryBuilderImpl.java
```diff
@@ -1336,10 +1329,13 @@ private String updateQuery() {
   }
 
   private void appendWheres(StringBuilder builder) {
+    // JSON API fully rely on Expression<BuildCondition> instead of List<BuildCondition>
     if (this.whereExpression != null) {
       appendConditions(this.whereExpression, " WHERE ", builder);
       return;
     }
+
+    // keep wheres, since Docs API still use it, although in deprecate plan
```
tatu-at-datastax, line 1338: Just include "keep wheres since Docs API still uses it", no need to mention deprecation here.

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/cql/ExpressionUtils.java
Yuqi-Du: solved

apis/sgv2-quarkus-common/src/main/java/io/stargate/sgv2/api/common/cql/builder/QueryBuilderImpl.java
Yuqi-Du: solved

tatu-at-datastax (approved):
LGTM

amorton (approved):

PR #2796, opened 2023-09-29
Fixes #2795: update Docker base image versions for Docs/GraphQL/REST API

tatu-at-datastax:
**What this PR does**:

Updates base images from 5 month old to latest for `jvm` and `native` images of Docs/GraphQL/REST APIs.

**Which issue(s) this PR fixes**:
Fixes #2795

**Checklist**
- [x] Changes manually tested
- [ ] Automated Tests added/updated
- [ ] Documentation added/updated
- [x] CLA Signed: [DataStax CLA](https://cla.datastax.com/)

PR #2826, opened 2023-10-28
ensure JVM heap memory settings applied in API images

jeffreyscarpenter:
Upgrading the base Docker images used for APIs in #2796 caused a regression in the v2.0.20 and v2.0.21 releases where JVM settings including max heap size via `JAVA_MAX_MEM_RATIO` are no longer honored. The issue is that the Dockerfiles for our API images use the `JAVA_OPTS` environment variable, when apparently we should be using `JAVA_OPTS_APPEND`. If you set `JAVA_OPTS`, other environment variables such as `JAVA_MAX_MEM_RATIO` are ignored. 

From this issue link : https://access.redhat.com/solutions/7031183
```
It is expected that if JAVA_OPTS is used, it includes all JVM options and system properties, and any S2I properties (e.g. JAVA_MAX_MEM_RATIO) will be ignored.

If it is desired to use S2I properties (e.g. JAVA_MAX_MEM_RATIO) and/or add to some base set of JVM options and/or system properties, then JAVA_OPTS_APPEND should be used.

Some builds do not strictly adhere to this (e.g. JAVA_OPTS does not do a full replacement); however, these are regressions that should not be relied upon and are fixed in later release.
```

Apparently we are now using a new enough base image that enforcement is now tripping us up.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/apis/sgv2-docsapi/src/main/docker/Dockerfile.jvm
+++ b/apis/sgv2-docsapi/src/main/docker/Dockerfile.jvm
@@ -90,6 +90,6 @@ COPY --chown=185 target/quarkus-app/quarkus/ /deployments/quarkus/
 
 EXPOSE 8080
 USER 185
-ENV JAVA_OPTS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager -Dvertx.disableURIValidation=true"
+ENV JAVA_OPTS_APPEND="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager -Dvertx.disableURIValidation=true"
 ENV JAVA_APP_JAR="/deployments/quarkus-run.jar"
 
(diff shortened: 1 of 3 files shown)


jeffreyscarpenter:
Verification procedure: 
- in stargate repo, check out this branch
- follow instructions to build coordinator and APIs locally
- then run `cd docker-compose/cassandra-4.0, ./start_cass_40_dev_mode.sh -l`
- access http://localhost:8082/metrics/ to view REST API metrics
- search for `jvm_memory_max_bytes`, you should see something like this:

```
jvm_memory_max_bytes{area="heap",id="G1 Old Gen",module="sgv2-restapi",} 1.610612736E9
```

This value equates to 75% of the 2GB allocated which is the desired value.




Jeff Carpenter
  [31 minutes ago](https://datastax.slack.com/archives/G01393WG1M1/p1698461249130179?thread_ts=1698449481.996419&cid=G01393WG1M1)
that’s 1.5 GB


Jeff Carpenter
  [31 minutes ago](https://datastax.slack.com/archives/G01393WG1M1/p1698461268556499?thread_ts=1698449481.996419&cid=G01393WG1M1)
if you do the same thing but with v2.0.21 you see the .5 gb

kathirsvn (approved):

tatu-at-datastax (approved):

PR #2854, opened 2023-12-05
Setting tracingId in PersistenceException

kathirsvn:
**What this PR does**:
Setting tracingId from the backend result in PersistenceException and in turn in ErrorMessage in case of an error and when the tracing is ON

**Checklist**
- [X] Changes manually tested


coordinator/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
```diff
@@ -564,8 +565,11 @@ private <T extends Result> CompletableFuture<T> executeRequest(
                       // There is only 2 types of response that can come out: either a
                       // ResultMessage (which itself can of different kind), or an ErrorMessage.
                       if (response instanceof ErrorMessage) {
-                        throw convertExceptionWithWarnings(
-                            (Throwable) ((ErrorMessage) response).error);
+                        PersistenceException pe =
+                            convertExceptionWithWarnings(
+                                (Throwable) ((ErrorMessage) response).error);
+                        pe.setTracingId(MessageHandler.getTracingId(response));
```
amorton, line 571: Why do we need the MessageHandler ? Can we simply call response.getTracingId ?

kathirsvn: `getTracingId()` of `Message.Response` is package protected so I ended up creating a class in that package and access that method. It was not accessible outside the package.

coordinator/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/ErrorMessage.java
```diff
@@ -461,6 +461,7 @@ public static ErrorMessage fromException(
         ProtocolVersion forcedProtocolVersion = ((ProtocolException) e).getForcedProtocolVersion();
         if (forcedProtocolVersion != null) message.forcedProtocolVersion = forcedProtocolVersion;
       }
+      message.tracingId = ((PersistenceException) e).tracingId();
```
tatu-at-datastax, line 464: nit: use `setTracingId()` instead of direct assign?

coordinator/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
tatu-at-datastax: I guess that is necessary for now, but split packages will become problematic with later JDKs.
But then again if that's the only way to force access for now, need to do so and worry about problem later on
(we won't need to support C* pre-5, or DSE 6.8 in future)

tatu-at-datastax (commented):
LGTM but I'll let @amorton approve.

coordinator/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/messages/ErrorMessage.java
tatu-at-datastax: Looks like method is not accessible, but can be made so; will push change.

coordinator/persistence-dse-6.8/src/main/java/io/stargate/db/dse/impl/DsePersistence.java
tatu-at-datastax: Ok, this is only needed with Cassandra-4.0 (and probably 4.1 for SG 2.1) -- 3.11/DSE-6.8 have method `public` so can be accessed directly. Will push changes.

coordinator/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Message.java
```diff
@@ -290,7 +290,7 @@ protected Response(Type type) {
       if (type.direction != Direction.RESPONSE) throw new IllegalArgumentException();
     }
 
-    Message setTracingId(UUID tracingId) {
+    public Message setTracingId(UUID tracingId) {
```
tatu-at-datastax, line 293: Needs to be `public` to call by persistence backends.

tatu-at-datastax:
@kathirsvn This was merged today in `main` (v2.x), merged to `v2.1` (and patched via #2857 to apply to C-4.1, dse-next). But I think there'd be some work needed for `cndb` persistence backend.

Issue #2875, opened 2024-01-22
GraphQL mutations with @atomic directive and TTL are not persisted


marksurnin:
Hi Stargate team,

We've encountered a potential bug in Stargate v2. Atomic mutations (logged batches) with any number of mutations that use TTL are not persisted. Moreover, Stargate responds with `"applied": true` for those mutations, which is misleading. I've reproduced this in the v2 images `stargateio/coordinator-4_0:v2` and `stargateio/graphqlapi:v2`, as well as on the v2.1 branch.

Request (using the standard data_endpoint_auth keyspace):
```
mutation insert_atomic_ttl @atomic {
  m1: inserttoken(
    value: {
      auth_token: "12735332-b914-11ee-a506-0242ac120002"
      username: "2a"
    }
    options:{ttl:86400}
  ) {
    applied
  }
  m2: inserttoken(
    value: {
      auth_token: "15cdf1e0-b914-11ee-a506-0242ac120002"
      username: "2b"
    }
    options:{ttl:86400}
  ) {
    applied
  }
}
```

Response
```
{
  "data": {
    "m1": {
      "applied": true
    },
    "m2": {
      "applied": true
    }
  }
}
```

Subsequently querying data_endpoint_auth.token table shows that no rows have been inserted. Note that rows are persisted correctly as long as I use _either_ `@atomic` directive or `ttl` in the mutations.

I haven't found the root cause of the failed mutation but it's concerning that `applied: true` and it's not persisted. This used to work in v1. Do you know what could be causing this? Thanks!

marksurnin:
Some troubleshooting findings:
- I'm aware of special handling of `@atomic` mutations in [MutationFetcher.java](https://github.com/stargate/stargate/blob/9193a19a888ae602a2cbbe66511c0db203038982/apis/sgv2-graphqlapi/src/main/java/io/stargate/sgv2/graphql/schema/cqlfirst/dml/fetchers/MutationFetcher.java#L59C1-L63C6). However, this kicks in if the number of selections (i.e. mutations is > 1) but I've reproduced this behavior even in a single mutation with `@atomic` and TTL.
[...]

PR #2893, opened 2024-02-12
Fix the write timestamp in atomic GraphQL mutations

marksurnin:
**What this PR does**:
Set `timestamp` and `now_in_seconds` in batch parameters only if they have been passed via query parameters. `com.google.protobuf.Int64Value` values in [query.proto](https://github.com/stargate/stargate/blob/0326abb5a7c9fe22e1257d2a0f074b91562da0d7/coordinator/bridge-proto/proto/query.proto#L576-L585) are serialized to 0 if they are unset, which causes the bug described below.

Fixes https://github.com/stargate/stargate/issues/2875

As described in this [comment](https://github.com/stargate/stargate/issues/2875#issuecomment-1919549571), Stargate currently sets the write timestamp of all `@atomic` batch mutations to Unix time 0. 

Consequently, the TTL is calculated as (Unix time 0 + TTL). Therefore, even when using the [maximum TTL](https://github.com/apache/cassandra/blob/2a4cd36475de3eb47207cd88d2d472b876c6816d/src/java/org/apache/cassandra/cql3/Attributes.java#L47) of 20 years, Cassandra interprets these rows as having already expired and they cannot be retrieved.

**Checklist**
- [x] Changes manually tested
- [ ] Documentation added/updated – N/A?


apis/sgv2-graphqlapi/src/test/java/io/stargate/sgv2/graphql/schema/cqlfirst/dml/fetchers/MutationFetcherTest.java
```diff
@@ -146,4 +167,45 @@ public void mutationAtomicMultipleSelectionWithDifferentOptionsFailTest() {
...
+
+    if (Strings.isNullOrEmpty(ttl)) {
+      assertBookAndAuthorInserts(batch);
+    } else {
+      assertBookAndAuthorInsertsWithTTL(batch, Integer.parseInt(ttl));
+    }
+
+    // Batch timestamp and nowInSeconds parameters should be unset to avoid using them in
+    // BatchHandler's `makeParameters`
+    assertFalse(batch.getParameters().hasTimestamp());
```
tub, line 204: Might be more idiomatic to use assertThat(batch.getParameters().hasTimestamp()).isFalse() here, given that we're already using assertJ

kathirsvn:
Thanks for the PR @marksurnin. 
The changes LGTM! 
Can we have this PR created against the `main` branch please? since this fix should be applied on both `v2` (`main`) and `v2.1`. We can pull this change to `v2.1` once it's merged with the `main`.

tatu-at-datastax:
+1 for PR against `main` (2.x) so that we can get this deployed to all of our production systems (there's a split)

kathirsvn:
@marksurnin may be you will have to rebase your fork or create new PR? Right now, its showing a lot of diff (around 644 files as diff)

kathirsvn (approved):
LGTM

marksurnin:
@kathirsvn the coordinator build is failing to download dependencies from DataStax's internal Maven repo. I don't build `persistence-dse-next` or `persistence-dse-6.8` locally so unsure how to proceed.

```Error:  Failed to execute goal on project persistence-dse-6.8: Could not resolve dependencies for project io.stargate.db.dse:persistence-dse-6.8:jar:2.0.25-SNAPSHOT: The following artifacts could not be resolved: io.netty:netty-codec:jar:4.1.100.1.dse, io.netty:netty-common:jar:4.1.100.1.dse, io.netty:netty-buffer:jar:4.1.100.1.dse, io.netty:netty-transport:jar:4.1.100.1.dse, io.netty:netty-resolver:jar:4.1.100.1.dse: Could not transfer artifact io.netty:netty-codec:jar:4.1.100.1.dse from/to artifactory-snapshots (https://repo.datastax.com/datastax-snapshots-local): status code: 401, reason phrase:  (401) -> [Help 1]```

**Comment by tatu-at-datastax** [comment]
@marksurnin This is likely due to access problems: DSE jar and some of its dependencies (like netty with backported fixes) are only available from DataStax internal Maven repo. I was hoping we could trigger re-run and that'd pass but that does not seem to be the case (I tried and it appears to fail still).

**Comment by kathirsvn** [comment]
Yes, I just triggered another re-run and it failed. Let me follow up on this.

**Comment by kathirsvn** [comment]
As per suggestion from @jeffreyscarpenter, I tried recreating this PR [here](https://github.com/stargate/stargate/pull/2895) and the build step goes through fine now. So, it looks like some access issue (i.e. my account seems to have read access to the datastax internal repo but @marksurnin account doesn't).

Now, we are getting below IT error
https://github.com/stargate/stargate/actions/runs/7907256239/job/21584210196?pr=2895
```
Results:
[INFO] 
Error:  Failures: 
Error:    AtomicBatchIntegrationTest.successfulBatchWithTTL:318 
expected: null
 but was: [{"extensions"={"classification"="DataFetchingException"}, "locations"=[{"column"=3, "line"=2}], "message"="Exception while fetching data (/foo) : Column 'v_writetime' does not exist", "path"=["foo"]}]
```

Probably, we will have to read the writetime and ttl as in `getWriteTime` method in `io.stargate.sgv2.graphql.integration.graphqlfirst.AtomicIntegrationTest`

**Comment by marksurnin** [comment]
> As per suggestion from @jeffreyscarpenter, I tried recreating this PR [here](https://github.com/stargate/stargate/pull/2895) and the build step goes through fine now. So, it looks like some access issue (i.e. my account seems to have read access to the datastax internal repo but @marksurnin account doesn't).


@kathirsvn could you please try pushing an empty commit to my [branch](https://github.com/marksurnin/stargate/tree/u/marksurnin/fix_atomic_batch_write_timestamp)? Perhaps the maven repo authentication would work this way.

I see that the credentials are read from secrets [here](https://github.com/stargate/stargate/blob/9e1528fd9c218686bef3c509ba81d86e3e533f4e/.github/workflows/apis-v2.yaml#L50-L51) but I don't know where the account's organization is used to check access.

> 
> Now, we are getting below IT error https://github.com/stargate/stargate/actions/runs/7907256239/job/21584210196?pr=2895
> 
> ```
> Results:
> [INFO] 
> Error:  Failures: 
> Error:    AtomicBatchIntegrationTest.successfulBatchWithTTL:318 
> expected: null
>  but was: [{"extensions"={"classification"="DataFetchingException"}, "locations"=[{"column"=3, "line"=2}], "message"="Exception while fetching data (/foo) : Column 'v_writetime' does not exist", "path"=["foo"]}]
> ```
> 
> Probably, we will have to read the writetime and ttl as in `getWriteTime` method in `io.stargate.sgv2.graphql.integration.graphqlfirst.AtomicIntegrationTest`

Secondly, I can update the tests to read writetime and TTL via CQL but it's concerning that it's failing with `Column 'v_writetime' does not exist`. These itests pass locally. The field that [AtomicBatchIntegrationTest](https://github.com/stargate/stargate/pull/2893/files#diff-7509ed795cce88afdfc883d1f04137177fc336628a4b04f8078463040005c578R316) is retrieving is
```
v_writetime: _bigint_function(name:"writetime", args: ["v"])
```
which is a standard use-case of an aggregation function. `v_writetime` a GraphQL alias, not a column. Do you know if the integration test image lacks aggregation function support or if something could be off with aliases?

**Comment by tatu-at-datastax** [comment]
@marksurnin I think the issue is that forks by developers without push rights to original repo cannot access secrets; this to prevent potential security problems.
But I guess it'd be easy enough to see whether commit to PR by someone with access would let CI run access secrets and see.

**Comment by kathirsvn** [comment]
@marksurnin I accepted the invitation to your fork and just pushed an empty commit to see if coordinator build CI job works fine now, let's see.

**Comment by kathirsvn** [comment]
> @marksurnin I accepted the invitation to your fork and just pushed an empty commit to see if coordinator build CI job works fine now, let's see.

Cool, it looks like the coordinator build job is good now. Let's check the aggregate function problem (i.e. v_writetime) next.

**Comment by kathirsvn** [comment]
> > @marksurnin I accepted the invitation to your fork and just pushed an empty commit to see if coordinator build CI job works fine now, let's see.
> 
> Cool, it looks like the coordinator build job is good now. Let's check the aggregate function problem (i.e. v_writetime) next.

When I run the test `AtomicBatchIntegrationTest.successfulBatchWithTTL` on my local, it fails with below error

```
Expected :null
Actual   :[{"extensions"={"classification"="DataFetchingException"}, "locations"=[{"column"=3, "line"=2}], "message"="Exception while fetching data (/foo) : Column 'v_writetime' does not exist", "path"=["foo"]}]
```

dav3yblaz3 (approved):
ok for test

PR #2958, opened 2024-06-25
Fix #2956: change Docker base images for API/native (to resolve CVEs)

### LINKED ISSUE #2956: Update Stargate 2.0/2.1 API native base images from `redhat/ubi8-mnimal:8.8` to `redhat/ubi8/openjdk-17-runtime:1.17`
Scanning tools complain about the current base image version in `sgv2-restapi/src/main/docker/Dockerfile.native` (and Docs/GraphQL) being

redhat/ubi8/ubi-minimal:8.8

but it looks like `sgv2-restapi/src/main/docker/Dockerfile.jvm` instead has

redhat/ubi8/openjdk-17-runtime:1.17

we should either bump former to 8.9, or, more likely change to same as "jvm" one
(we do not actually use native images for anything, based on observed poor performance)

tatu-at-datastax:
**What this PR does**:

Update Stargate v2 REST/GraphQL/DocsAPI native variants' Docker base images to ones without Snyk-reported vulnerabilities.
For some reason native images use different base: will try PR to see if use of ".jvm" images' base images just works as-is, to unify handling.

Base image change appears to work: actual image size likely grow, going from minimal to "full" base.

Fixes #2956

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/apis/sgv2-docsapi/src/main/docker/Dockerfile.native
+++ b/apis/sgv2-docsapi/src/main/docker/Dockerfile.native
@@ -14,7 +14,7 @@
 # docker run -i --rm -p 8180:8180 quarkus/sgv2-docsapi
 #
 ###
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.8
+FROM registry.access.redhat.com/ubi8/openjdk-17-runtime:1.17
 WORKDIR /work/
 RUN chown 1001 /work \
     && chmod "g+rwX" /work \
--- a/apis/sgv2-graphqlapi/src/main/docker/Dockerfile.native
+++ b/apis/sgv2-graphqlapi/src/main/docker/Dockerfile.native
@@ -14,7 +14,7 @@
 # docker run -i --rm -p 8080:8080 quarkus/sgv2-graphqlapi-quarkus
 #
 ###
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.8
+FROM registry.access.redhat.com/ubi8/openjdk-17-runtime:1.17
 WORKDIR /work/
 RUN chown 1001 /work \
     && chmod "g+rwX" /work \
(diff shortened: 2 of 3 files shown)


jeffreyscarpenter (approved):
This looks fine but can you add a note on the change to image size? I'm guessing it is going up since we are no longer using a "minimal" image?

tatu-at-datastax:
@jeffreyscarpenter Probably true. What would be the best way to figure out size difference? For jars I can look at `apis/sgv2-docsapi/target/` but Docker images go to... registry?

jeffreyscarpenter:
@tatu-at-datastax i was thinking just build the images locally and run `docker images`

tatu-at-datastax:
@jeffreyscarpenter Ah. Run the Maven build, see what the result is. Makes sense, I can do that.

tatu-at-datastax:
Default build only produces ".jvm" images, I think, where indicated size is about 400 MB locally (yet 160MB at DockerHub, odd).

But looking at base images themselves:

* Minimal, 8.8 (https://hub.docker.com/layers/redhat/ubi8-minimal/8.8-1072.1697626218/images/sha256-a49924d9d685a35b2d0817ffe9c84f3429d97e9ad29ed3816c389f45564c9e19?context=explore) has size of about 40 MB (compressed)
* Full, 17 (https://catalog.redhat.com/software/containers/ubi8/openjdk-17/618bdbf34ae3739687568813?architecture=amd64&image=65369e291095f6679ad488e1) has size of about 140 MB (compressed)

so there would likely be size increase. I am unable to see how ".native" images are built, however, and there seems to be no tags at DockerHub to differentiate
(but let me look maybe it's different image name)

tatu-at-datastax:
Given that unification has some value, and that we don't actually deploy native images anywhere, will proceed and merge.

PR #2998, opened 2024-09-11
CNDB-9895: Allow excluding protocol versions from supported versions advertised to the client

jacek-lewandowski:
Adds a system property which can limit the maximum support protocol version.

### DIFF
--- a/coordinator/persistence-api/src/main/java/org/apache/cassandra/stargate/transport/ProtocolVersion.java
+++ b/coordinator/persistence-api/src/main/java/org/apache/cassandra/stargate/transport/ProtocolVersion.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.function.IntPredicate;
 import java.util.stream.Collectors;
 
 /**
@@ -51,6 +52,8 @@ public enum ProtocolVersion implements Comparable<ProtocolVersion> {
 
   /** Set this to true for beta versions */
   private final boolean beta;
+
+  public static final IntPredicate ALLOW_ALL_FILTER = v -> true;
 
   ProtocolVersion(int num, String descr, boolean beta) {
     this.num = num;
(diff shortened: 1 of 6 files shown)


coordinator/persistence-api/src/main/java/org/apache/cassandra/stargate/transport/ProtocolVersion.java
```diff
@@ -84,19 +90,60 @@ public enum ProtocolVersion implements Comparable<ProtocolVersion> {
...
+    try {
+      maxProtocolVersion = Integer.getInteger(MAX_PROTOCOL_VERSION_PROP, Integer.MAX_VALUE);
+    } catch (NumberFormatException e) {
+      throw new ConfigurationException(
+          "Invalid integer value for "
+              + MAX_PROTOCOL_VERSION_PROP
+              + ": "
+              + System.getProperty(MAX_PROTOCOL_VERSION_PROP));
+    }
+    if (maxProtocolVersion < 1)
```
maheshrajamani, line 113: The condition needs to be (maxProtocolVersion < 3), so there is at least 1 supported version. Supported OS versions are {V3, V4, V5, V6}.

maheshrajamani (commented):
There is CI failure, can you check?

coordinator/persistence-api/src/main/java/org/apache/cassandra/stargate/transport/ProtocolVersion.java
```diff
@@ -114,6 +124,11 @@ public static List<ProtocolVersion> supportedVersionsStartingWith(
   }
 
   public static ProtocolVersion decode(int versionNum, boolean allowOlderProtocols) {
```
maheshrajamani, line 126: Can we remove this method, don't think this should be used from anywhere.

maheshrajamani (approved):
Overall looks good and approving the change. The decode method in ProtocolVersion class without the protocol filter should be removed.

PR #3002, opened 2024-09-24
amc-844 - usage of Histogram Metric to measure Reads Size with SelectStatements

maxospiquante:
**What this PR does**:
This creates a new Metric  `persistence_dse_[clusterVersion]_org_apache_cassandra_metrics_client_request_read_size_histogram` exposed under xxx:8084/metrics

This PR also contains an associated test which test the right presence of this Metric in Metrics list.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/coordinator/testing/src/main/java/io/stargate/it/http/MetricsTest.java
+++ b/coordinator/testing/src/main/java/io/stargate/it/http/MetricsTest.java
@@ -17,6 +17,7 @@
 package io.stargate.it.http;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assumptions.assumeThat;
 import static org.awaitility.Awaitility.await;
 
 import io.stargate.it.BaseIntegrationTest;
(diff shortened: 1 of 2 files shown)


coordinator/testing/src/main/java/io/stargate/it/http/MetricsTest.java
```diff
@@ -184,6 +185,22 @@ public void dropwizardMetricsPersistence() throws IOException {
     assertThat(lines).isNotEmpty();
   }
 
+  @Test
+  public void dropwizardMetricReadSizeIsPresent() throws IOException {
```
tatu-at-datastax, line 189: nit: I know it's following existing naming, but since SG v2 does not use DropWizard any more (but quarkus), should just drop it and lave "metricReadSizeIsPresent()".

tatu-at-datastax:
What is the issue this PR fixes, and that explain what is being added and why?
(yes, adding new metric, histogram, but details, reasoning etc)

coordinator/testing/src/main/java/io/stargate/it/http/MetricsTest.java
maxospiquante: The new Metric `metricReadSizehistogram` is based on Histogram class from `com.codahale.metrics. MetricRegistry` package, coming from `io.dropwizard.metrics:metrics-core:4.2.18`maven dependancy. Anyway I can update the method name, sure.

maxospiquante:
> What is the issue this PR fixes, and that explain what is being added and why? (yes, adding new metric, histogram, but details, reasoning etc)
As this Metric is missing from Cassandra metrics, the idea is to add it on Stargate side.

PR #3004, opened 2024-09-27
Update CQL message encoder/decoder to respect frame header flags ordering as per native protocol specs

### LINKED ISSUE #3003: CQL ProtocolDecoder/ProtocolEecoder doesn't adhere to native protocol specs when it comes to frame header flags ordering.
I encountered a problem when utilizing custom payload in native CQL protocol: https://github.com/datastax/cassandra/blob/2ec96ae669ad506b30a0489acf077b615fc67870/doc/native_protocol_v4.spec#L141

The problem manifest itself on a downstream stargate user when decoding the response with a compliant decoder - for example a java driver that uses the java native protocol from: https://github.com/datastax/native-protocol/tree/78e62fd7c93afb2500c7db103f4bf7c9dfd6662d

The problem is the encoding order is significant, and it is different between the [encoder](https://github.com/stargate/stargate/blob/a2b4dd3353cadf8a814afde81db95fa163f19d2d/coordinator/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Message.java#L414-L425) and [decoder](https://github.com/datastax/native-protocol/blob/78e62fd7c93afb2500c7db103f4bf7c9dfd6662d/src/main/java/com/datastax/oss/protocol/internal/FrameCodec.java#L322-L334). 

I beilve the problem was untice in the past because the `customPayload` was absent from the C* responses - even if it set it would go through if it was the only flag. But if both `customPayload` & say `warnings` coexists, a java client would observe the below error:

```
java.lang.IndexOutOfBoundsException: readerIndex(54) + length(65554) exceeds writerIndex(125): UnpooledSlicedByteBuf(ridx: 54, widx: 125, cap: 125/125, unwrapped: PooledUnsafeDirectByteBuf(ridx: 125, widx: 125, cap: 147))

	at com.datastax.oss.driver.shaded.netty.buffer.AbstractByteBuf.checkReadableBytes0(AbstractByteBuf.java:1442)
```

aymkhalil:
**What this PR does**:
* Change the order of encoding/decoding frame header flags to respect the order of native protocal specs (V4 or later) as specified [here](https://github.com/datastax/cassandra/blob/2ec96ae669ad506b30a0489acf077b615fc67870/doc/native_protocol_v4.spec#L125) 
* Also exposed `customPayload` on `ResultMesasge` to make it available for a downstream client

Fixes #3003

**Checklist**
- [X] Changes manually tested

### DIFF
--- a/coordinator/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Message.java
+++ b/coordinator/cql/src/main/java/org/apache/cassandra/stargate/transport/internal/Message.java
@@ -174,7 +174,7 @@ public abstract class Message {
   protected Connection connection;
   private int streamId;
   private long sourceFrameBodySizeInBytes;
-  private Map<String, ByteBuffer> customPayload;
+  protected Map<String, ByteBuffer> customPayload;
   protected ProtocolVersion forcedProtocolVersion = null;
 
   protected Message(Type type) {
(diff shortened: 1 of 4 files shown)


jeffreyscarpenter (commented):
Changes look good, but should this be based off of the v2.1 branch since that is what cndb is now using? @aymkhalil

tatu-at-datastax (approved):

tatu-at-datastax:
Looks like CI fails due to access issue. I will try to "fork" this PR into my own; that should work around the problem.

tatu-at-datastax:
Created #3009 off of this and it seems to avoid access problem. If so, will close this, merge that one. Unfortunately we do not seem to have a better way to handle "external" PRs at this point.

PR #3011, opened 2024-10-03
CNDB-11144 Allow to run on JDK22+ (update OSGI runtime)

jacek-lewandowski:
**What this PR does**:
- Update OSGI core from 6.0.0 to 8.0.0
- Update Apache Felix from 6.0.2 to 7.0.5
- Upgrade Mockito
- Allow BUILDING on JDK up to JDK23

Fixes #<issue number>

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/.github/workflows/coordinator-test-v21.yml
+++ b/.github/workflows/coordinator-test-v21.yml
@@ -57,12 +57,11 @@ jobs:
         with:
           python-version: '3.12'
 
-      - uses: actions/@v4
+      - uses: actions/setup-java@v4
         name: Setup Java JDK
         with:
           distribution: 'temurin'
-          # Some Maven Plugin requires Java 17
-          java-version: 17
+          java-version: 11
           cache: 'maven'
 
       - name: Setup Maven
@@ -160,8 +159,7 @@ jobs:
         name: Setup Java JDK
         with:
           distribution: 'temurin'
-          # Some Maven Plugin requires Java 17
-          java-version: 17
+          java-version: 11
           cache: 'maven'
 
       - name: Setup Maven
--- a/.github/workflows/coordinator-test-v21.yml
+++ b/.github/workflows/coordinator-test-v21.yml
@@ -61,7 +61,7 @@ jobs:
         name: Setup Java JDK
         with:
           distribution: 'temurin'
-          java-version: 22
+          java-version: 17
           cache: 'maven'
 
       - name: Setup Maven
@@ -159,7 +159,7 @@ jobs:
         name: Setup Java JDK
         with:
           distribution: 'temurin'
-          java-version: 22
+          java-version: 17
           cache: 'maven'
 
       - name: Setup Maven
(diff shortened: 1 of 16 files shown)


coordinator/persistence-dse-next/pom.xml
```diff
@@ -18,7 +18,7 @@
     <!-- Latest as of 2023-12-05: 4.0.11-b259738f492f -->
     <!-- Latest as of 2024-01-24: 4.0.11-0248d170a615 -->
     <!-- Latest as of 2024-04-29: 4.0.11-591d171ac9c9 -->
-    <cassandra.version>4.0.11-591d171ac9c9</cassandra.version>
+    <cassandra.version>4.0.11-5696964ff12e</cassandra.version>
```
tatu-at-datastax, line 21: When changing this version, please add a comment on date when update was made (to give an idea of how old DSE version is)

coordinator/pom.xml
```diff
@@ -43,8 +43,8 @@
   <!-- Shared properties -->
   <properties>
     <!-- Compiler -->
-    <maven.compiler.source>11</maven.compiler.source>
-    <maven.compiler.target>11</maven.compiler.target>
+    <maven.compiler.source>17</maven.compiler.source>
```
eolivelli, line 46: I don't think that we need to change this in order to make it run on JDK22
we can keep it built for JDK11, this way we reduce the risks of subtle behavior changes and also it won't change the developer enviroment setup for existing developers

WDYT ?

coordinator/persistence-dse-next/pom.xml
eolivelli: do we need to update Cassandra ? IIUC we are only updating the dependencies in order to make it work on modern JDKs, the current version of Cassandra work decently on JDK22

.github/workflows/coordinator-test-v21.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         name: Setup Java JDK
         with:
           distribution: 'temurin'
-          java-version: 11
+          java-version: 17
```
eolivelli, line 64: I would keep this file and ensure that everything keeps working on jdk11 and create a second workload that builds on jdk22

eolivelli (commented):
i have left a couple of comments, in general I think that it is better to keep supporting jdk11, this way it is easier for people to port changes/do hotfixes

coordinator/bridge/src/test/java/io/stargate/bridge/service/BaseBridgeServiceTest.java
```diff
@@ -73,7 +72,7 @@ public class BaseBridgeServiceTest {
   protected @Mock Persistence persistence;
   protected @Mock AuthorizationService authorizationService;
 
-  protected Connection connection = spy(mock(Connection.class));
+  protected Connection connection = mock(Connection.class);
```
eolivelli, line 75: the new version of Mockito errored with "IllegalArgument Please don't pass mock here. Spy is not allowed on mock"

eolivelli (approved):

tatu-at-datastax (approved):

PR #3080, opened 2025-04-15
Try to fix CI problems

tatu-at-datastax:
**What this PR does**:

Addresses 2 observed problems with CI

N/A

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/coordinator/persistence-test/src/main/java/io/stargate/it/exec/ProcessRunner.java
+++ b/coordinator/persistence-test/src/main/java/io/stargate/it/exec/ProcessRunner.java
@@ -159,7 +159,7 @@ public class ProcessRunner {
 
     @Override
     public void onProcessFailed(ExecuteException e) {
-      LOG.info("{} process {}/{} failed with exception: {}", kind, generation, node, e);
+      LOG.info("{} process {}/{} failed with exception: {}", kind, generation, node, String.valueOf(e));
       cleanup();
       ready.completeExceptionally(e);
       exit.countDown();
--- a/coordinator/persistence-test/src/main/java/io/stargate/it/exec/ProcessRunner.java
+++ b/coordinator/persistence-test/src/main/java/io/stargate/it/exec/ProcessRunner.java
@@ -159,7 +159,8 @@ public class ProcessRunner {
 
     @Override
     public void onProcessFailed(ExecuteException e) {
-      LOG.info("{} process {}/{} failed with exception: {}", kind, generation, node, String.valueOf(e));
+      LOG.info(
+          "{} process {}/{} failed with exception: {}", kind, generation, node, String.valueOf(e));
       cleanup();
       ready.completeExceptionally(e);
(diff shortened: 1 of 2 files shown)


coordinator/persistence-test/src/main/java/io/stargate/it/exec/ProcessRunner.java
```diff
@@ -159,7 +159,7 @@ public void onProcessComplete(int exitValue) {
 
     @Override
     public void onProcessFailed(ExecuteException e) {
-      LOG.info("{} process {}/{} failed with exception: {}", kind, generation, node, e);
+      LOG.info("{} process {}/{} failed with exception: {}", kind, generation, node, String.valueOf(e));
```
tatu-at-datastax, line 162: Not sure why but I see these in logs:

```
2025-04-15T16:33:51.0618251Z INFO  [process-runner-stargate-0-0] 2025-04-15 16:33:51,042 ProcessRunner.java:162 - Stargate process 0/0 failed with exception: {}
```

so trying to see if method called is different when last arg is `Throwable` (before change) vs `String` with change.

coordinator/pom.xml
```diff
@@ -734,6 +734,11 @@
         <artifactId>logback-core</artifactId>
         <version>${logback.version}</version>
       </dependency>
+      <dependency>
+        <groupId>ch.qos.logback</groupId>
+        <artifactId>logback-access</artifactId>
+        <version>${logback.version}</version>
+      </dependency>
```
tatu-at-datastax, line 741: Noticed from IDEA there was a dependency to older version of `logback-access` than `logback-core`. And logs had:

    2025-04-15T16:40:08.8618576Z INFO  [Exec Stream Pumper] 2025-04-15 16:40:08,792 ProcessRunner.java:78 - stargate1-0> Caused by: java.lang.ClassNotFoundException: ch.qos.logback.core.hook.DelayingShutdownHook not found by io.stargate.db.dse_next [1]

which could be due to old(er) version of `logback-core`. Not sure this will fix it, but will try.

tatu-at-datastax: Does not change things; ClassNotFoundException remains.

coordinator/persistence-test/src/main/java/io/stargate/it/exec/ProcessRunner.java
tatu-at-datastax: Seems to fix the issue!

coordinator/pom.xml
tatu-at-datastax: Not really sure where logback 1.2.x comes through: it could be due to dropwizard-metrics 4.2.18 (not latest, but all 4.2.x have same dep) which depends on 1.2.11... but we have managed-dependency force version so really unclear.
Even jars under `coordinator/stargate-lib` have 1.3.15 version.

PR #3083, opened 2025-04-17
Disabling dse-next Coordinator tests (failing with 2h timeout)

tatu-at-datastax:
**What this PR does**:

Disables "dse-next" Coordinator ITs from CI, to unblock our pipeline: this combo always fails with 2h timeout. Unable to find the root cause, ruled out possibility was caused by changes we made.
Leading candidate for root cause is Github runner change for `ubuntu-latest`, changing from 22.04 to 24.04 around Dec-2024/Jan-2025), somehow causing failure (but just for this backend).

N/A

**Checklist**
- [ ] Changes manually tested

### DIFF
--- a/.github/workflows/coordinator-test-v21.yml
+++ b/.github/workflows/coordinator-test-v21.yml
@@ -18,7 +18,7 @@ on:
       - 'coordinator/**pom.xml'
       - 'coordinator/**/src/**'
       - 'coordinator/**mvn**'
-      - '.github/workflows/coordinator-test.yml'
+      - '.github/workflows/coordinator*.yml'
 
   pull_request:
     branches: [ "v2.1" ]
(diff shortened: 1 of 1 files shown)


Hazel-Datastax (approved):

.github/workflows/coordinator-test-v21.yml
```diff
@@ -18,15 +18,15 @@ on:
       - 'coordinator/**pom.xml'
       - 'coordinator/**/src/**'
       - 'coordinator/**mvn**'
-      - '.github/workflows/coordinator-test.yml'
+      - '.github/workflows/coordinator*.yml'
```
tatu-at-datastax, line 21: Just so this workflow is run for changes to workflow file itself

.github/workflows/coordinator-test-v21.yml
```diff
@@ -131,9 +133,9 @@ jobs:
             build-profile: -P dse
             test-profile: -P dse,it-dse-v4-6.8
 
-          - backend: dse-next
-            build-profile: -P dse-next
-            test-profile: -P dse-next,it-dse-next
+#          - backend: dse-next
```
tatu-at-datastax, line 136: Must also comment these, otherwise `dse-next` will be added to matrix (won't remove yet just in case we want to re-enable in near future; if not can remove for good)

PR #3091, opened 2025-04-30
Fix several issues with CQL ClientMetrics

jkni:
**What this PR does**:
* In PreV5Handlers, avoid double-pausing connections. This prevent metric drift from the underlying number of paused connections.
* In ClientMetrics, change PerFrame metrics to PerMessage metrics. The PerFrame metrics are currently never incremented, and in V5, it makes more sense to expose these on a per-message level. In pre-V5, the protocol is frame per message.
* In pre-v5 protocol, actually increment total bytes/bytes per message metrics.
* In CQLMessageHandler, Cassandra metrics are being updated for ClientMetrics/ClientMessageSizeMetrics. Switch to Stargate metrics.
* Update tests to improve coverage for the above issues.
* Add error logging to previously silent exception path intended to be unreachable.

**Checklist**
- [x] Changes manually tested

### DIFF
--- a/coordinator/cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
+++ b/coordinator/cql/src/main/java/org/apache/cassandra/stargate/metrics/ClientMetrics.java
@@ -91,8 +91,8 @@ public final class ClientMetrics {
   private Counter unknownException;
   private Counter totalBytesRead;
   private Counter totalBytesWritten;
-  private DistributionSummary bytesReceivedPerFrame;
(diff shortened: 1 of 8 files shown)


tatu-at-datastax:
@maheshrajamani I had a quick look and things make sense but do not really know internals very well -- I recall you working on protocol v5 support so perhaps you could sanity check as well?

tatu-at-datastax (approved):
Not an expert on this but makes sense to me. Will see if I can have another team mate to have a quick look before merging.

**EDIT**: @Yuqi-Du had a peek as well. Will now merge.

PR #3119, opened 2025-08-27
CNDB-14889: add handling for `DateRangeType` (follow-up to CNDB-14199)

tatu-at-datastax:
**What this PR does**:

Adds minimal support for `DateRangeType`, to avoid crashes for table schema contain columns of that type.

CNDB-14889

**Checklist**
- [ ] Changes manually tested


tatu-at-datastax:
Closing since we cannot properly test the change: `DateRange` type is not supported by "vanilla" DSE (dse-db) but only `dse-search` extension, and our CI does not test such backend. Fix would probably work but at this point we don't have confidence in verification.

cc @ekaterinadimitrova2

